PR #8, opened 2025-03-11
removing ds store, adding gitignore

dmitry-openai:
Removing DS_Store and adding gitignore


rm-openai:
While you're here, can you remove the egg/pycache too? there are a bunch of pycache directories that slipped in

rm-openai (approved):

PR #9, opened 2025-03-11
add __pycache__ to .gitignore and remove compiled Python files

Jai0401:
add __pycache__ to .gitignore and remove compiled Python files


rm-openai:
I'm taking care of this separately - appreciate the PR though!

PR #10, opened 2025-03-11
fix(example): Replace non-existent model with a valid OpenAI model

hironow:
This pull request resolves the following issue:

```
openai.NotFoundError: Error code: 404 - {'error': {'message': 'The model computer-use-preview-2025-02-04 does not exist or you do not have access to it.', 'type': 'invalid_request_error', 'param': None, 'code': 'model_not_found'}}
```

The example has been updated to reference an existing and accessible OpenAI model.  
This change ensures the example code functions correctly when executed.


hironow:
Already updated, thanks! I'll close this PR.

PR #4, opened 2025-03-11
Cleanup  still remaining __pycache__ and *.egg-info directories and fix typo in README 

notV3NOM:
Removed  __pycache__ and *.egg-info directories

PR #11, opened 2025-03-11
Delete pycache/egg files + fix computer use model

rm-openai:
These slipped in because I forgot to add the gitignore initially. Remove them.

Also fixed computer use model


sbera77:
I noticed that the __pycache__ folders haven't been completely removed yet . I previously submitted a PR #4 that covers all of them

PR #13, opened 2025-03-11
fix: correct typo in README regarding tool responses

Jai0401:
"We process the tool calls (if any) and append the tool responses messsages."
→ "We process the tool calls (if any) and append the tool response messages." (typo: "messsages" → "messages")
"If the response has a final output (see below for the more on this), we return it and end the loop."
→ "If the response has a final output (see below for more on this), we return it and end the loop." (remove "the")


sbera77:
Hi, the changes in your PR appears to duplicate the modifications already proposed in PR #4

Jai0401:
Hi @notV3NOM , PR #4 seems quite cluttered with duplicate or overlapping changes.

Jai0401:
Hi @rm-openai, could you please review my PR when you have a moment?

rm-openai (approved):

PR #22, opened 2025-03-11
Add memory example using mem0

Dev-Khant:
- This example demonstrates a customer service application with memory enabled using [[mem0](https://mem0.ai/)](https://mem0.ai/).
- You can obtain your mem0 API key from: https://app.mem0.ai/dashboard/api-keys.
- For more details on how to use core mem0 functions, refer to the official docs: https://docs.mem0.ai/platform/quickstart.

### DIFF
--- a/examples/customer_service/memory_enabled.py
+++ b/examples/customer_service/memory_enabled.py
@@ -1,0 +1,122 @@
+from __future__ import annotations as _annotations
+
+import os
+import asyncio
+
+from pydantic import BaseModel
+try:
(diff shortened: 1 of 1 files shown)


rm-openai:
Thanks for the PR. For now, we are not adding code that requires new dependencies unless they are universally useful. We may revisit this stance in the future. Until then, perhaps you could add an example in a separate repo or in a separate integration guide to Mem0?

(And yes I know the PR doesn't add an explicit dependency, but implicit ones count too)

PR #24, opened 2025-03-11
Run `make format`

alexmojaki:


rm-openai:
rolling this into #27 - bc some of these files are beng deleted!

PR #25, opened 2025-03-11
Stronger tracing tests with `inline-snapshot`

alexmojaki:
`assert len(spans) == 12` is a very weak assertion. This PR asserts the exported traces and spans more precisely in a readable tree format. And when the format of an exported trace/span changes (e.g. a new key is added to every span), you can use `pytest --inline-snapshot=fix` to update all relevant tests automatically. See https://15r10nk.github.io/inline-snapshot/latest/ for more info.


alexmojaki:
There are more places where this can be done (I just did a find and replace on `spans = fetch_ordered_spans()`), and the simpler testing logic can be cleaned up, but this demonstrates the idea.

rm-openai:
This is a great PR, thank you - hadn't seen this library before. Looks like there are some test failures/typecheck failures. I'm happy to take over and send out an updated PR, or let me know if you feel like updating things.

PR #39, opened 2025-03-12
Update quickstart.md

matthewford:
Add an example that works before the guard, so people don't think the task just failed.


dmitry-openai (approved):
LGTM, thank you!

PR #50, opened 2025-03-12
perf: Add screenshot caching to reduce data transfer

tongshu1943:
# Optimization: Screenshot Caching to Reduce Data Transfer

## Feature Description
Added a screenshot caching mechanism to avoid repeatedly transferring the same screenshot data, improving performance and data transfer efficiency.

## Specific Improvements
1. Added `screenshot_hash` abstract method to `Computer` and `AsyncComputer` interfaces
2. Implemented caching logic in `ComputerAction` class using hash values to detect duplicate screenshots
3. Implemented `screenshot_hash` method in `LocalPlaywrightComputer`
4. Added tests to ensure the caching mechanism works correctly
5. Fixed type checking and linting issues

## Technical Implementation
- Uses hash algorithm to detect duplicate screenshots, avoiding redundant transmission of the same image data
- Maintains backward compatibility with existing API, not affecting existing code
- Added comprehensive test coverage to ensure functionality
- Clean code implementation without unnecessary comments

## Testing
- Added dedicated test file `tests/test_screenshot_cache.py` to verify the caching mechanism
- All type checks and lint checks pass


rm-openai (changes requested):
Thanks for the PR. I don't grok the change fully, a couple of quick questions/comments:
1. The screenshot URL is a base64 encoded image. It's not a URL like "foo.com/bar.jpg". So I don't get how hashing it helps here?
2. I think the right way to do this is via the Responses API - to use a reference ID and just send that up. 
3. This makes an implicit tradeoff between bandwidth and memory usage, which may not be the right tradeoff for most cases.

PR #55, opened 2025-03-12
fix: count `spans` and `traces` correctly

### LINKED ISSUE #53: Traces and Spans not updated in the SDK
## Issue

In the [`export`](https://github.com/openai/openai-agents-python/blob/main/src/agents/tracing/processors.py#L73) method in `processor.py` the following code lines do not update the `traces` and `spans` variables -

```python
traces: list[dict[str, Any]] = []
spans: list[dict[str, Any]] = []

data = [item.export() for item in items if item.export()]
payload = {"data": data}

headers = {
    "Authorization": f"Bearer {self.api_key}",
    "Content-Type": "application/json",
    "OpenAI-Beta": "traces=v1",
}
```

Hence, in the debug logs you won't see the correct values when this code is run -

```python
# Exponential backoff loop
attempt = 0
delay = self.base_delay
while True:
    attempt += 1
    try:
        response = self._client.post(url=self.endpoint, headers=headers, json=payload)

        # If the response is successful, break out of the loop
        if response.status_code < 300:
            logger.debug(f"Exported {len(traces)} traces, {len(spans)} spans")
            return
```

## Solution

Updating the above code snippet with the following solves the issue -

```python
traces: list[dict[str, Any]] = []
spans: list[dict[str, Any]] = []

# Categorize items into traces and spans
for item in items:
    if hasattr(item, 'export') and callable(item.export):
        export_data = item.export()
        if export_data:
            if isinstance(item, Trace):
                traces.append(export_data)
            else:
                spans.append(export_data)

data = traces + spans
payload = {"data": data}
```

the-praxs:
Closes #53 

Counts the number of `spans` and `traces` correctly to show the correct number in the debug logs.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -81,7 +81,17 @@ class BackendSpanExporter(TracingExporter):
         traces: list[dict[str, Any]] = []
         spans: list[dict[str, Any]] = []
 
-        data = [item.export() for item in items if item.export()]
+        # Categorize items into traces and spans
+        for item in items:
+            if hasattr(item, 'export') and callable(item.export):
+                export_data = item.export()
+                if export_data:
+                    if isinstance(item, Trace):
+                        traces.append(export_data)
+                    else:
+                        spans.append(export_data)
+
+        data = traces + spans
         payload = {"data": data}
 
         headers = {
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -78,9 +78,6 @@ class BackendSpanExporter(TracingExporter):
             logger.warning("OPENAI_API_KEY is not set, skipping trace export")
             return
 
-        traces: list[dict[str, Any]] = []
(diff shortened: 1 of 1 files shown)


src/agents/tracing/processors.py
```diff
@@ -81,7 +81,17 @@ def export(self, items: list[Trace | Span[Any]]) -> None:
         traces: list[dict[str, Any]] = []
         spans: list[dict[str, Any]] = []
 
-        data = [item.export() for item in items if item.export()]
+        # Categorize items into traces and spans
```
rm-openai: I'd rather just log `len(items)` on line 113 and delete lines 81/82 instead of doing this extra work for little benefit.

rm-openai (changes requested):

src/agents/tracing/processors.py
dot-agi: Done

rm-openai (approved):

Issue #63, opened 2025-03-12
AttributeError: 'CompletionUsage' object has no attribute 'input_tokens'


LeonG7:
When I tried non-streaming output, the code worked fine. but I tried to stream output, the code reported an error

```python
## examples/agent_patterns/routing.py


import asyncio
import uuid

from openai.types.responses import ResponseContentPartDoneEvent, ResponseTextDeltaEvent

from agents import Agent, RawResponsesStreamEvent, Runner, TResponseInputItem, trace

"""
This example shows the handoffs/routing pattern. The triage agent receives the first message, and
then hands off to the appropriate agent based on the language of the request. Responses are
streamed to the user.
"""

french_agent = Agent(
    name="french_agent",
    instructions="You only speak French",
    model=OpenAIChatCompletionsModel(
        model="qwen-max",
        openai_client=external_client,
    ),
    model_settings=ModelSettings(temperature=0.6),
)

spanish_agent = Agent(
    name="spanish_agent",
    instructions="You only speak Spanish",
    model=OpenAIChatCompletionsModel(
        model="qwen-max",
        openai_client=external_client,
    ),
    model_settings=ModelSettings(temperature=0.6),
)

english_agent = Agent(
    name="english_agent",
    instructions="You only speak English",
    model=OpenAIChatCompletionsModel(
        model="qwen-max",
        openai_client=external_client,
    ),
    model_settings=ModelSettings(temperature=0.6),
)

triage_agent = Agent(
    name="triage_agent",
    instructions="Handoff to the appropriate agent based on the language of the request.",
    handoffs=[french_agent, spanish_agent, english_agent],
    model=OpenAIChatCompletionsModel(
        model="qwen-max",
        openai_client=external_client,
    ),
    model_settings=ModelSettings(temperature=0.6),
)


async def main():
    # We'll create an ID for this conversation, so we can link each trace
    conversation_id = str(uuid.uuid4().hex[:16])

    msg = input("Hi! We speak French, Spanish and English. How can I help? ")
    agent = triage_agent
    inputs: list[TResponseInputItem] = [{"content": msg, "role": "user"}]

    while True:
        # Each conversation turn is a single trace. Normally, each input from the user would be an
        # API request to your app, and you can wrap the request in a trace()
        with trace("Routing example", group_id=conversation_id):
            result = Runner.run_streamed(
                agent,
                input=inputs,
            )
            async for event in result.stream_events():
                if not isinstance(event, RawResponsesStreamEvent):
[...]

PR #61, opened 2025-03-12
fix(usage): error usage format in streaming

Kaikaikaifang:
image

In the `stream_response`, forgot to modify the format of usage.

## Why this error occur

in `_run_single_turn_stream` (`run.py`), treat the usage as `ResponseUsage`, but in fact the usage is `CompletionUsage`, because forgot to modify the format of usage in `stream_response` (`openai_chatcompletions.py`).

fix #63

PR #65, opened 2025-03-12
Fix token usage attribute names in stream_response for streamed outputs

zky001:
This PR fixes an issue in the streaming response processing where the code incorrectly accessed non-existent attributes (input_tokens and output_tokens) on the Usage object. Instead, the correct property names (prompt_tokens and completion_tokens) should be used. see https://github.com/openai/openai-python/blob/v1.66.2/src/openai/types/completion_usage.py
fix bug metioned in #63

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -651,8 +651,8 @@ class Runner:
                 usage = (
                     Usage(
                         requests=1,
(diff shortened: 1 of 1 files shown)


Kaikaikaifang:
This PR #61 has made a more comprehensive fix for this issue.

PR #74, opened 2025-03-12
Fixes #71: Added support for max_tokens in ModelSettings

### LINKED ISSUE #71: max_tokens is not an accepted parameter
Out of the Documentation, there is no reference of max_tokens. the ModelSettings does not accept the max_tokens parameter..

This becomes a problem especially when using anthropic models as they dont assume a max tokens value and need one to get passed.

s44002:
Fixes #71

This PR adds support for `max_tokens` in `ModelSettings`. This was previously missing, causing issues with models that require an explicit max_tokens value.


rm-openai:
Apologies, I didn't see this issue/PR in time and implemented it myself via #105

PR #81, opened 2025-03-12
refactor: reduce unnecessary task cancel

gouzil:
Removed duplicate calls to `self._output_guardrails_task.cancel()` in the `_cleanup_tasks` method.


rm-openai:
apologies i cleaned this up myself - thanks for the pr though!

PR #90, opened 2025-03-12
perf: implement screenshot caching via Responses API reference IDs

tongshu1943:
## Design Approach
This implementation leverages the OpenAI Responses API reference ID mechanism to optimize screenshot handling in the ComputerTool. The system uses reference IDs to point to previously sent images, significantly reducing bandwidth usage and memory footprint.

## Implementation Details
- Modified `Converter._convert_tool` to set the include parameter "computer_call_output.output.image_url" for ComputerTool instances
- Added test coverage to verify the reference ID functionality works correctly
- Updated existing tests to maintain compatibility with the new include parameter while preserving original design intent

## Technical Considerations
- The OpenAI Responses API automatically handles reference management on the server side
- This approach shifts the storage burden from client memory to the API service
- The implementation maintains backward compatibility with existing code
- Clean code principles were followed with minimal changes to the codebase

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -366,7 +366,7 @@ class Converter:
                 "display_width": tool.computer.dimensions[0],
                 "display_height": tool.computer.dimensions[1],
             }
-            includes = None
+            includes = "computer_call_output.output.image_url"
 
         else:
             raise UserError(f"Unknown tool type: {type(tool)}, tool")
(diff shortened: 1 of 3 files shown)


tongshu1943:
@rm-openai Thank you for your valuable suggestions! 🙏 I have updated the implementation to use the Responses API for caching, as per your recommendation. This approach simplifies the logic and improves efficiency.

Please take a moment to review the changes and let me know if there are any further adjustments or improvements needed.

Looking forward to your feedback!

PR #100, opened 2025-03-12
Create CONTRIBUTING.md

sumitaryal:


rm-openai:
We'll follow up with this from openai directly

PR #103, opened 2025-03-12
Remove duplicated code

rm-openai:


rm-openai:
We'll follow up with this from openai directly

PR #137, opened 2025-03-13
fix(#94): handle special parameters (self/cls) correctly with context parameter

### LINKED ISSUE #94: Add `method_tool` Functionality

```
@function_tool
    """Get a random number."""
    return random.randint(0, 100)

agent = Agent(
    name="Assistant",
    tools=[get_random_number]
)
print(result.final_output)
```

But the following doesn't:

```
class Tools:
    @function_tool
        """Get a random number."""
        return random.randint(0, 100)

tools = Tools()

agent = Agent(
    name="Assistant",
)
print(result.final_output)
```

```
```

Shehryar718:
Fixes #94 

## Description
- Fix issue where special parameters (self/cls) were being incorrectly filtered
- Add tests to verify context parameter works correctly with special parameters
- Ensure schema only contains actual function parameters, not special ones

## Testing
- Added comprehensive test cases:
  - Instance methods with context
  - Class methods with context
  - Static methods with context
- Verified special parameters (self/cls) are properly handled
- Confirmed schema generation excludes special parameters

## Related Issue
This PR addresses [issue #94](https://github.com/openai/openai-agents-python/issues/94)

## Checklist
- [x] Tests added/updated and passing locally
- [x] Code follows project style guidelines
- [x] Documentation updated (if required)
- [x] Issue linked in the PR description


seratch:
If we still want to support #94 use case, having this change may be worth considering. What do you think? @rm-openai If it's not a priority, we can close both the issue and this PR.

PR #143, opened 2025-03-13
perf: Optimize has_more_than_n_keys function and BatchTraceProcessor

tongshu1943:
## Feature Description
Implemented performance optimizations in two key areas of the codebase:

1. Replaced inefficient key counting method in `has_more_than_n_keys` function with a more efficient approach
2. Optimized the `BatchTraceProcessor` class to improve trace and span export performance

## Specific Improvements
1. Modified `has_more_than_n_keys` function to use `len(obj)` directly instead of iterating through keys
2. Optimized `_export_batches` method to reduce queue checks and improve batching logic
3. Added tests to verify the optimized functions work correctly

## Technical Implementation
- Dictionary key counting: Using `len(obj)` directly is more efficient than iterating through all keys
- Trace batch processing: Reduced the number of queue checks and improved the batching logic
- Both optimizations maintain existing behavior without introducing regressions
- Added comprehensive test coverage for the optimized functions

## Testing
- Added test file `test_has_more_than_n_keys.py` to verify the optimized function
- All tests pass, confirming the optimizations work correctly
- All type checks and lint checks pass

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -159,9 +159,4 @@ def is_list(obj: object) -> TypeGuard[list[object]]:
 
 
 def has_more_than_n_keys(obj: dict[str, object], n: int) -> bool:
-    i = 0
-    for _ in obj.keys():
-        i += 1
-        if i > n:
-            return True
-    return False
+    return len(obj) > n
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for sending this pull request. We haven't observed performance issues here, so let us close this PR now.

PR #147, opened 2025-03-13
feat: Add Graphviz-based agent visualization functionality

MartinEBravo:
This pull request introduces functionality for visualizing agent structures using Graphviz. The changes include adding a new dependency, implementing functions to generate and draw graphs, and adding tests for these functions.

New functionality for visualizing agent structures:

* Added `graphviz` as a new dependency in `pyproject.toml`.
* Implemented functions in `src/agents/visualizations.py` to generate and draw graphs for agents using Graphviz. These functions include `get_main_graph`, `get_all_nodes`, `get_all_edges`, and `draw_graph`.

Testing the new visualization functionality:

* Added tests in `tests/test_visualizations.py` to verify the correctness of the graph generation and drawing functions. The tests cover `get_main_graph`, `get_all_nodes`, `get_all_edges`, and `draw_graph`.

For example, given the following code:

```python
from agents import Agent, function_tool
from agents.visualizations import draw_graph


@function_tool
    return f"The weather in {city} is sunny."


spanish_agent = Agent(
    name="Spanish agent",
    instructions="You only speak Spanish.",
)

english_agent = Agent(
    name="English agent",
    instructions="You only speak English",
)

triage_agent = Agent(
    name="Triage agent",
    instructions="Handoff to the appropriate agent based on the language of the request.",
    handoffs=[spanish_agent, english_agent],
    tools=[get_weather],
)


draw_graph(triage_agent)
```

Generates the following image:

Screenshot 2025-03-13 at 18 36 23


tests/test_visualizations.py
```diff
@@ -0,0 +1,111 @@
+from unittest.mock import Mock
+
+import graphviz
+import pytest
+
+from src.agents.agent import Agent
```
rm-openai:
```suggestion
from agents import Agent
```

rm-openai (changes requested):
This is such a great PR/idea. Couple of small things to get it in:
1. Could you please add this to the `agents/src/extensions` directory? So the import would be `from agents.extensions import visualization`
2. Could you please add an optional dependency group? So you can do `pip install openai-agents[visualization]` to install it, and if you don't, there's an import error.

Thank you!

tests/test_visualizations.py
MartinEBravo: Fixed!

src/agents/extensions/visualization.py
```diff
@@ -0,0 +1,119 @@
...
+    for tool in agent.tools:
+        parts.append(
+            f'"{tool.name}" [label="{tool.name}", shape=ellipse, style=filled, '
+            f"fillcolor=lightgreen, width=0.5, height=0.3];"
+        )
+
+    # Bigger handoffs (rounded box, yellow)
+    for handoff in agent.handoffs:
+        parts.append(
+            f'"{handoff.name}" [label="{handoff.name}", shape=box, style=filled, style=rounded, '
```
rm-openai: I think you want `handoff.agent_name` not `handoff.name`, if it's a `Handoff` instance

src/agents/extensions/visualization.py
```diff
@@ -0,0 +1,119 @@
...
+            f"fillcolor=lightyellow, width=1.5, height=0.8];"
+        )
+        parts.append(get_all_nodes(handoff))
```
rm-openai: I don't think this is right - the handoffs list is `list[Handoff | Agent]`. So the node might be an Agent, but it might also be a `Handoff` object, where you don't statically know which agent is being handed off to. You can use `handoff.agent_name` for that

rm-openai (changes requested):
Looks like tests are failing. I think you can add an optional dependency to pyproject.toml
```
[project.optional-dependencies]
visualization=["graphviz"]
```

src/agents/extensions/visualization.py
MartinEBravo: Added isinstance check.

src/agents/extensions/visualization.py
MartinEBravo: Added isinstance check.

pyproject.toml
```diff
@@ -49,6 +49,12 @@ dev = [
     "playwright==1.50.0",
     "inline-snapshot>=0.20.7",
 ]
+
+[project.optional-dependencies]
+visualization = [
+    "graphviz>=0.17",
+]
+
```
rm-openai: you might need a rebase, as `[project.optional-dependencies]` exists already. Also perhaps `viz` is a shorter optional dependency heh

rm-openai: Yeah looks like this lack of rebase is causing all the checks to fail.

rm-openai (changes requested):
THink you just need a rebase and should be good to go

rm-openai:
Oof unfortunately checks still failing. I tried to update them but you may need to run `make tests`, `make mypy` etc locally.

MartinEBravo:
Ok, now it cannot fail.

rm-openai (approved):

PR #151, opened 2025-03-13
Fix typo on Agent documentation

caike:
Argument name is not description but handoff_description


rm-openai (changes requested):
looks like lint is failing

PR #155, opened 2025-03-14
Fix typos in event names and comments across multiple files

apeltop:
I hope it helps!

### DIFF
--- a/src/agents/_run_impl.py
+++ b/src/agents/_run_impl.py
@@ -632,7 +632,7 @@ class RunImpl:
             elif isinstance(item, HandoffCallItem):
                 event = RunItemStreamEvent(item=item, name="handoff_requested")
             elif isinstance(item, HandoffOutputItem):
-                event = RunItemStreamEvent(item=item, name="handoff_occured")
+                event = RunItemStreamEvent(item=item, name="handoff_occurred")
             elif isinstance(item, ToolCallItem):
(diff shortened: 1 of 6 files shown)


src/agents/stream_events.py
```diff
@@ -31,7 +31,7 @@ class RunItemStreamEvent:
     name: Literal[
         "message_output_created",
         "handoff_requested",
-        "handoff_occured",
+        "handoff_occurred",
```
rm-openai: unfortunately this would be a breaking change, since people may have written code that discriminates on the `name`. Maybe revert this and add a comment saying it's misspelled but we can't change it?

apeltop:
Sorry for the late response. If there is anything awkward or wrong in the sentences in the comment, please tell me.

src/agents/_run_impl.py
```diff
@@ -720,7 +720,7 @@ def stream_step_result_to_queue(
             elif isinstance(item, HandoffCallItem):
                 event = RunItemStreamEvent(item=item, name="handoff_requested")
             elif isinstance(item, HandoffOutputItem):
-                event = RunItemStreamEvent(item=item, name="handoff_occured")
+                event = RunItemStreamEvent(item=item, name="handoff_occurred")
```
seratch, line 723: this needs to be reverted too

PR #176, opened 2025-03-15
Add PDF extraction example agent with verification

### LINKED ISSUE #1234: Agent-level MCP server tool filtering
### Question
I saw that agent level tool filtering was planned but then removed in the tool filter PR [#861](https://github.com/openai/openai-agents-python/pull/861).
In my multi-agent system, I want to reuse my MCP servers but also limit each agent to the tools that are relevant to it. Is there a way to do this with the current implementation?

crjaensch:
### Checks

### DIFF
--- a/examples/extract_doc/pdf_extraction_agent.py
+++ b/examples/extract_doc/pdf_extraction_agent.py
@@ -1,0 +1,243 @@
+#!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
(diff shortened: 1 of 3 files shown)


rm-openai (changes requested):
Thank you for this PR! We're trying to keep the examples here universally helpful and dependency light. We're internally discussing the best way to add examples like this that use dependencies like reportlab, and will get back to you soon!

akhilsmokie (approved):

PR #180, opened 2025-03-16
Update README.md

leonprou:
I was trying to install with pip 3.8 but it was not finding out the repo. Finally I understood that I need to upgrade my python version.


rm-openai (changes requested):
Hey @leonprou - thanks for the PR! My main hesitation adding this is that Python 3.8 is already end-of-life (https://devguide.python.org/versions/). We support all non-EOL versions. DO you still think it would be useful to add this?

PR #182, opened 2025-03-16
    Add Firecrawl integration for web scraping and information extraction

aparupganguly:
Just wanted to integrate Firecrawl as an Example option. 🔥 
Let me know if any changes need to be made!


rm-openai (changes requested):
Thank you for this PR! We're trying to keep the examples here universally helpful and dependency light. We're internally discussing the best way to add examples like this that use dependencies like firecrawl, and will get back to you soon!

aparupganguly:
> Thank you for this PR! We're trying to keep the examples here universally helpful and dependency light. We're internally discussing the best way to add examples like this that use dependencies like firecrawl, and will get back to you soon!

That’s awesome and really exciting! Looking forward to the integration..

PR #186, opened 2025-03-17
fix reasoning order in examples

heartkilla:
I think we need to put reasoning-related outputs first to encourage CoT reasoning.


rm-openai (approved):
makes sense

PR #198, opened 2025-03-17
docs: rm extra parentheses

baskaryan:


rm-openai (changes requested):
already implemented in main

PR #203, opened 2025-03-17
Introduce tool_use_behavior on agents

### LINKED ISSUE #117: Use tool output as agent output
Is there some equivalent to AutoGen's `reflect_on_tool_use=False`? That is: Some way for an agent to use the output of one of its tools as its own output, without needing another roundtrip to the LLM?

Worked example:
```python
# Untested code, just illustrative

class LoremModel(BaseModel):
    paragraph: str

@function_tool
def generate_lorem(n: int) -> List[LoremModel]:
    '''Generate `n` paragraphs of lorem ipsum'''
    # omitted

agent = Agent(
    name='Lorem Agent',
    tools=[generate_lorem],
    output_type=List[LoremModel],
)

await Runner.run(agent, 'Generate 1000 paragraphs of lorem ipsum')
```

I believe the way this works is:
 1. Prompt gets sent up to the LLM
 2. LLM responds requesting invocation of `generate_lorem(1000)`
 3. We execute that function and now have 1000 paragraphs of lorem ipsum in memory
 4. We send these 1000 paragraphs back up to the LLM
 5. The LLM generates a final response to the user prompt, which is just those same paragraphs of lorem ipsum again.

In practice we already have our answer after step 3, but then have this extra roundtrip in steps 4&5, which can get expensive in terms of both tokens and time if the output from our tool is large.

**Is there functionality or design pattern to work around this? I think it would really help push the composability of agents even further than you already have with this SDK!**

_Real world usecase_: I have an agent whose goal it is to retrieve a chunky piece of content. It has ~5 tools at its disposal. The "main" tool actually retrieves the data, the other tools help the agent determine the correct parameters with which to invoke the main tool. I would like to then use this "data retrieval agent" inside other agents which can then use this dataset to do interesting things.

I can work around this by making my "data retrieval agent" only return the correct parameters with which to call the main tool, and leave the _actual_ retrieval to other agents, but then it feels like my agent isn't encapsulating a nice isolated chunk of work.

Love the library so far, thanks for your work!

rm-openai:
## Context
By default, the outputs of tools are sent to the LLM again. The LLM gets to read the outputs, and produce a new response. There are cases where this is not desired:
1. Every tool results in another round trip, and sometimes the output of the tool is enough.
2. If you force tool use (via model settings `tool_choice=required`), then the agent will just infinite loop.

This enables you to have different behavior, e.g. use the first tool output as the final output, or write a custom function to process tool results and potentially produce an output.

## Test plan
Added new tests and ran existing tests
Also added examples.

Closes #117

### DIFF
--- a/src/agents/_run_impl.py
+++ b/src/agents/_run_impl.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import asyncio
+import inspect
+from collections.abc import Awaitable
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 
 from openai.types.responses import (
     ResponseComputerToolCall,
@@ -25,7 +27,7 @@ from openai.types.responses.response_computer_tool_call import (
 from openai.types.responses.response_input_param import ComputerCallOutput
 from openai.types.responses.response_reasoning_item import ResponseReasoningItem
 
-from .agent import Agent
+from .agent import Agent, ToolsToFinalOutputResult
 from .agent_output import AgentOutputSchema
 from .computer import AsyncComputer, Computer
 from .exceptions import AgentsException, ModelBehaviorError, UserError
@@ -48,7 +50,7 @@ from .logger import logger
 from .models.interface import ModelTracing
 from .run_context import RunContextWrapper, TContext
 from .stream_events import RunItemStreamEvent, StreamEvent
-from .tool import ComputerTool, FunctionTool
+from .tool import ComputerTool, FunctionTool, FunctionToolResult
 from .tracing import (
     SpanError,
     Trace,
@@ -69,6 +71,8 @@ class QueueCompleteSentinel:
 
 
 QUEUE_COMPLETE_SENTINEL = QueueCompleteSentinel()
+
+_NOT_FINAL_OUTPUT = ToolsToFinalOutputResult(is_final_output=False, final_output=None)
 
 
 @dataclass
@@ -199,7 +203,7 @@ class RunImpl:
                 config=run_config,
             ),
         )
-        new_step_items.extend(function_results)
+        new_step_items.extend([result.run_item for result in function_results])
         new_step_items.extend(computer_results)
 
         # Second, check if there are any handoffs
@@ -214,6 +218,36 @@ class RunImpl:
                 hooks=hooks,
                 context_wrapper=context_wrapper,
                 run_config=run_config,
+            )
+
+        # Third, we'll check if the tool use should result in a final output
+        check_tool_use = await cls._check_for_final_output_from_tools(
+            agent=agent,
+            tool_results=function_results,
+            context_wrapper=context_wrapper,
+            config=run_config,
+        )
+
+        if check_tool_use.is_final_output:
+            # If the output type is str, then let's just stringify it
+            if not agent.output_type or agent.output_type is str:
+                check_tool_use.final_output = str(check_tool_use.final_output)
+
+            if check_tool_use.final_output is None:
+                logger.error(
+                    "Model returned a final output of None. Not raising an error because we assume"
+                    "you know what you're doing."
+                )
+
+            return await cls.execute_final_output(
+                agent=agent,
+                original_input=original_input,
+                new_response=new_response,
+                pre_step_items=pre_step_items,
+                new_step_items=new_step_items,
+                final_output=check_tool_use.final_output,
+                hooks=hooks,
+                context_wrapper=context_wrapper,
             )
 
         # Now we can check if the model also produced a final output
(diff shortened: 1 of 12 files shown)


src/agents/_run_impl.py
```diff
@@ -646,6 +678,39 @@ def stream_step_result_to_queue(
...
+        elif callable(agent.tool_use_behavior):
+            if inspect.iscoroutinefunction(agent.tool_use_behavior):
+                return await cast(
+                    Awaitable[ToolsToFinalOutputResult],
+                    agent.tool_use_behavior(context_wrapper, tool_results),
+                )
+            else:
+                return cast(
+                    ToolsToFinalOutputResult, agent.tool_use_behavior(context_wrapper, tool_results)
+                )
```
TimoVink, line 723: Perhaps a try/except/rethrow around the call to the user-provided function with a useful error message?

Was just playing with this branch and even with this function it was stopping after 1 turn.
```python
def custom_tool_use_behavior(context, results):
    return ToolsToFinalOutputResult(is_final_output=False, final_output=None)
```
... turns out it was a silly mistake on my part where I had forgotten to import `ToolsToFinalOutputResults`, but there were no obvious user-visible errors that I could find -- even with the `openai.agents` logger set to DEBUG. It just looked as-if the agent had decided it was done even though `is_final_result` was `False`.

src/agents/agent.py
```diff
@@ -95,6 +119,22 @@ class Agent(Generic[TContext]):
...
+        and gets to respond.
+    - "stop_on_first_tool": The output of the first tool call is used as the final output. This
+        means that the LLM does not process the result of the tool call. In addition, its
+    - A function: If you pass a function, it will be called with the run context and the list of
+      tool results. It must return a `ToolToFinalOutputResult`, which determines whether the tool
+      calls result in a final output.
+
+      NOTE: This configuration is specific to FunctionTools. Hosted tools, such as file search,
+      web search, etc are always processed by the LLM.
+    """
```
TimoVink, line 144: This completely supports my use cases. Thanks!

Minor suggestion: In my case an agent's goal is to invoke one particular tool, but needs to use a few others tools to get there. I imagine this may be a fairly common usecase.

If you agree, there's two ways the library could make that usecase even easier to handle:
 (1) Similar to the `ModelSettings.tool_choice` parameter, allow `tool_use_behaviour` to be a `str` that matches a tool name. And the behaviour would be "stop after the named tool has been called".
 (2) Or, provide a factory function similar to the below:

```python
def stop_after_tool_use(tool: str | FunctionTool) -> ToolsToFinalOutputFunction:
    if isinstance(tool, FunctionTool):
        tool = tool.name
    
    def custom_tool_use_behavior(context, results):
        for result in results:
            if tool == result.tool.name:
                return ToolsToFinalOutputResult(is_final_output=True, final_output=result.output)
        return ToolsToFinalOutputResult(is_final_output=False, final_output=None)

    return custom_tool_use_behavior

# Usage example:
@function_tool
def get_foo_id(user_input):
   ...

@function_tool
def get_bar_id(user_input):
    ...

@function_tool
def get_large_dataset(foo_id, bar_id):
    ...

agent = Agent(
    name="Foo Agent",
    instructions="Using your tools, retrieve the requested dataset.",
    tools=[get_foo_id, get_bar_id, get_large_dataset],
    tool_use_behavior=stop_after_tool_use(get_large_dataset),
)
```

rm-openai: good suggestion

src/agents/_run_impl.py
rm-openai: oh weird! Yeah I'll add that. Thanks

src/agents/agent.py
```diff
@@ -157,3 +197,5 @@ async def get_system_prompt(self, run_context: RunContextWrapper[TContext]) -> s
             logger.error(f"Instructions must be a string or a function, got {self.instructions}")
 
         return None
+
+        return None
```
stevenheidel: double return

src/agents/agent.py
```diff
@@ -22,6 +24,28 @@
...
 
+@dataclass
+class ToolsToFinalOutputResult:
+    is_final_output: bool
+    """Whether this is the final output. If False, the LLM will run again and receive the tool call
+    output.
+    """
+
+    final_output: Any | None = None
+    """The final output. Can be None if `is_final_output` is False, otherwise must match the
```
stevenheidel, line 35: what happens when this is set and `is_final_output` is False?

rm-openai: `is_final_output` is the arbiter there (i.e. the final output is ignored)

stevenheidel: if the `final_output` only makes sense when `is_final_output=True` - could we collapse this into just `final_output: Any | None` ?

rm-openai: I have a slight preference for the way it is - because None is technically a valid `Any` and there's no way to differentiate between None as in "no final output" and None as in "the final output is literally None"

stevenheidel: good point

stevenheidel (approved):

PR #204, opened 2025-03-17
Update tracing docs

rm-openai:
Update the README to point to the list of tracing processors. The docs will contain all the external processors, so the readme doesn't get too long.


dmitry-openai (approved):
LGTM

PR #211, opened 2025-03-18
Improve note section for Structure output

EnggQasim:
Improve note section for Structure output


rm-openai:

PR #231, opened 2025-03-18
openai-agents with google colab and Gemini

EnggQasim:
anyone can practice with Google Gemini free


EnggQasim:
@rm-openai [2c1a684](https://github.com/openai/openai-agents-python/pull/231/commits/2c1a6849d2d8e9e863717c0da32332bf93222e17) 
**Tracing Example with local computer and Google gemini**

rm-openai:
Hey thanks for the PR, but I don't think we want to have copies for jupyter here. We have one demonstrative example, but this is too much overhead to maintain.

PR #244, opened 2025-03-19
Update tool.py bullets lines

EnggQasim:
Update tool.py bullets lines


rm-openai:
I'd rather not have newlines here, its harder to read.

Issue #252, opened 2025-03-20
Enhance `on_tool_start` Hook to Include Tool Call Arguments


yanmxa:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) - Yes
- **Have you searched for related issues?** Others may have had similar requesrs - Yes

### Describe the feature

Currently, the `on_tool_start` hook in the Agents SDK only provides access to the Agent, Tool (definition), and context, but does not include the actual tool call arguments.

This limitation makes it difficult to trace and debug the agent's lifecycle, as there's no way to inspect the exact parameters passed to the tool before execution.

**Proposed Enhancement:**

Modify the `on_tool_start` hook definition to include the tool call arguments, providing full visibility into the tool invocation process.

- Current Definition

```python
async def on_tool_start(
    self,
    context: RunContextWrapper[TContext],
    agent: Agent[TContext],
    tool: Tool,
) -> None:
    """Called before a tool is invoked."""
    pass

```
- Suggested Update

```python
async def on_tool_start(
    self,
    context: RunContextWrapper[TContext],
    agent: Agent[TContext],
    tool: Tool,
    arguments: str,  # Include the actual tool call arguments
) -> None:
    """Called before a tool is invoked."""
    pass
```

**Why This Improvement Matters:**

1. Improved Debugging & Tracing: Developers can log and analyze tool inputs before execution.
2. Greater Flexibility: Hooks can validate or preprocess tool arguments before they are used.
3. More Context for Hooks: Allows hooks to monitor execution flows with full tool call details.

MrJarnould:
Until the linked PR is accepted, is there any other way to obtain the tool call arguments using hooks before the agent calls the tool?

vanhtuan0409:
> Until the linked PR is accepted, is there any other way to obtain the tool call arguments using hooks before the agent calls the tool?

Some workaround I think of is either using LLM tracing, put logging in tool implement or hijack the network call

yanmxa:
@MrJarnould @vanhtuan0409 Yes, I do it in the tool calling internal for now. But it's more reasonable to put this in the tracing hook :)

MrJarnould:
I hope your PR gets reviewed soon!

Kalai-off:
Any idea when this enhancement will be released ?

yanmxa:
@rm-openai Could you please take a look at this issue? Thanks!

BarAshkenazi:
[...]

PR #253, opened 2025-03-20
Add tool call parameters for `on_tool_start` hook

yanmxa:
Resolve: https://github.com/openai/openai-agents-python/issues/252

### DIFF
--- a/src/agents/_run_impl.py
+++ b/src/agents/_run_impl.py
@@ -398,9 +398,9 @@ class RunImpl:
                     span_fn.span_data.input = tool_call.arguments
                 try:
                     _, _, result = await asyncio.gather(
-                        hooks.on_tool_start(context_wrapper, agent, func_tool),
+                        hooks.on_tool_start(context_wrapper, agent, func_tool, tool_call.arguments),
                         (
-                            agent.hooks.on_tool_start(context_wrapper, agent, func_tool)
+                            agent.hooks.on_tool_start(context_wrapper, agent, func_tool, tool_call.arguments)
                             if agent.hooks
                             else _coro.noop_coroutine()
                         ),
--- a/examples/basic/agent_lifecycle_example.py
+++ b/examples/basic/agent_lifecycle_example.py
@@ -28,7 +28,7 @@ class CustomAgentHooks(AgentHooks):
             f"### ({self.display_name}) {self.event_counter}: Agent {source.name} handed off to {agent.name}"
         )
 
-    async def on_tool_start(self, context: RunContextWrapper, agent: Agent, tool: Tool) -> None:
+    async def on_tool_start(self, context: RunContextWrapper, agent: Agent, tool: Tool, args: str) -> None:
         self.event_counter += 1
         print(
             f"### ({self.display_name}) {self.event_counter}: Agent {agent.name} started tool {tool.name}"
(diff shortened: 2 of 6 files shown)


src/agents/_run_impl.py
```diff
@@ -398,9 +398,9 @@ async def run_single_tool(
                     span_fn.span_data.input = tool_call.arguments
                 try:
                     _, _, result = await asyncio.gather(
-                        hooks.on_tool_start(context_wrapper, agent, func_tool),
+                        hooks.on_tool_start(context_wrapper, agent, func_tool, tool_call.arguments),
                         (
-                            agent.hooks.on_tool_start(context_wrapper, agent, func_tool)
+                            agent.hooks.on_tool_start(context_wrapper, agent, func_tool, tool_call.arguments)
                             if agent.hooks
```
mini-peanut, line 565: Since using only arguments as a parameter is too granular, and adding tool_call.id in the future would lead to parameter inflation at the root level, plus tool_call.id is actually needed, would it be better to pass the entire tool_call object as a parameter instead?"

louis-sanna-dev:
@yanmxa thanks for doing this work, very useful!

src/agents/_run_impl.py
yanmxa: @mini-peanut Thanks for your review! I've updated the hook to use `Action`, which not only includes the `Tool` but also contains the instance information for a tool call, including the parameters and `tool_call.id`.

wesleytao:
much needed feature!

MattMorgis:
Bump on this, could use this merged 🙏

mattmorgis:
I learned that if you use [`Runner.run_streamed()`](https://openai.github.io/openai-agents-python/ref/run/#agents.run.Runner.run_streamed), they provide a [`RunItemStreamEvent`](https://openai.github.io/openai-agents-python/ref/stream_events/#agents.stream_events.RunItemStreamEvent), one of which is type `tool_called`. It gives you a full `RunItem` which has the tool arguments. I think this is the preferred approach because it uses SSE from the responses api. 

Example: 
```python 
result = Runner.run_streamed(
    agent,
    input,
)
    if event.type == "run_item_stream_event":
        if event.name == "tool_called":
            print(
                f"-- Tool {event.item.raw_item.name} was called with args: {event.item.raw_item.arguments} "
            )
        elif event.name == "reasoning_item_created":
            summary = event.item.raw_item.summary
            # there’s at least one summary part
            # o-series models will send this every response, it is sometimes blank
            if summary:
                text = summary[0].text
                print(f"-- Reasoning item created: {text}")
                print("--")
        elif event.name == "message_output_created":
            print(event.item.raw_item.content[0].text)
        else:
            pass  # Ignore other event types
```

yanmxa:
Thanks @rm-openai\!

I tend to favor the breaking change approach because:

1. **Adding a new method (like `on_tool_start_with_args()`) creates API confusion** - developers would choose between two similar hooks, and we'd maintain both forever

2. **Simple migration** - users just replace `tool` with `action`, then access tool via `action.function_tool`

3. **Clean long-term design** - one unified hook that provides complete information

The community feedback shows this is needed. A clean break now might be better than API fragmentation.

What do you think?

ludovicoserrani-nautes:
@rm-openai do you have any ETA on this PR?

ihower:
@seratch This PR can be closed because it’s already implemented at https://github.com/openai/openai-agents-python/pull/1765

seratch:
@ihower Ah, yes! Thanks for your effort 🙇

PR #255, opened 2025-03-20
Adds example for financial agent

shyamal-anadkat:
This example shows how you might compose a richer financial research agent using the Agents SDK. The pattern is similar to the `research_bot` example, but with more specialized sub‑agents and a verification step.

The flow is:

1. **Planning**: A planner agent turns the end user’s request into a list of search terms relevant to financial analysis – recent news, earnings calls, corporate filings, industry commentary, etc.
2. **Search**: A search agent uses the built‑in `WebSearchTool` to retrieve terse summaries for each search term. (You could also add `FileSearchTool` if you have indexed PDFs or 10‑Ks.)
3. **Sub‑analysts**: Additional agents (e.g. a fundamentals analyst and a risk analyst) are exposed as tools so the writer can call them inline and incorporate their outputs.
4. **Writing**: A writer agent brings together the search snippets and any sub‑analyst summaries into a long‑form markdown report plus a short executive summary.
5. **Verification**: A final verifier agent audits the report for obvious inconsistencies or missing sourcing.

### DIFF
--- a/examples/financial_research_agent/README.md
+++ b/examples/financial_research_agent/README.md
@@ -1,0 +1,38 @@
+# Financial Research Agent Example
+
+This example shows how you might compose a richer financial research agent using the Agents SDK. The pattern is similar to the `research_bot` example, but with more specialized sub‑agents and a verification step.
+
+The flow is:
+
+1. **Planning**: A planner agent turns the end user’s request into a list of search terms relevant to financial analysis – recent news, earnings calls, corporate filings, industry commentary, etc.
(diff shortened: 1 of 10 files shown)


rm-openai (approved):

schatekar:
Thank you for putting together this example. It clarifies a few things for me. 

I have a quick question, though. I came here hoping to find out how to get agents to coordinate with each other to complete a task. But what I am seeing is a lot of Python code that invokes individual agents one after another. It this is the recommended approach to orchestrate between agents? Or is it possible to make use of `as_tool` method to get agents to use other agents as toos to complete task?

PR #261, opened 2025-03-20
Remove redundant weaker tracing assertions

alexmojaki:
Following up on https://github.com/openai/openai-agents-python/pull/25, this removes uses of `fetch_traces` and `fetch_ordered_spans` where there's already a stronger assertion using `fetch_normalized_spans`. This is to help move towards the stronger style as much as possible, since people are still adding weaker assertions such as [this](https://github.com/openai/openai-agents-python/blob/7a0ca7930e31e5d12f7785dfcb0a2f123739b21a/tests/test_concurrency.py#L59-L61) in https://github.com/openai/openai-agents-python/pull/91. The next step will be to find remaining uses of `fetch_ordered_spans` that can be replaced.


rm-openai (approved):
thank you!

PR #263, opened 2025-03-20
Fix potential infinite tool call loop by resetting tool_choice after …

mini-peanut:
# Fix potential infinite tool call loop by resetting tool_choice after tool execution

## Summary

This PR fixes an issue where setting `tool_choice` to "required" or a specific function name could cause models to get stuck in an infinite tool call loop.

When `tool_choice` is set to force tool usage, this setting persists across model invocations. This PR automatically resets `tool_choice` to "auto" after tool execution, allowing the model to decide whether to make additional tool calls in subsequent turns.

Unlike using `tool_use_behavior="stop_on_first_tool"`, this approach lets the model continue processing tool results while preventing forced repeated tool calls.

## Test plan

- Added tests to verify tool_choice reset behavior for both agent and run_config settings
- Added integration test to verify the solution prevents infinite loops
- All tests pass

## Checks

- [x] I've added new tests for the fix
- [x] I've updated the relevant documentation (added comment in code)

### DIFF
--- a/src/agents/_run_impl.py
+++ b/src/agents/_run_impl.py
@@ -47,6 +47,7 @@ from .items import (
 )
 from .lifecycle import RunHooks
 from .logger import logger
+from .model_settings import ModelSettings
 from .models.interface import ModelTracing
 from .run_context import RunContextWrapper, TContext
 from .stream_events import RunItemStreamEvent, StreamEvent
@@ -205,6 +206,37 @@ class RunImpl:
         )
         new_step_items.extend([result.run_item for result in function_results])
         new_step_items.extend(computer_results)
+
+        # Reset tool_choice to "auto" after tool execution to prevent infinite loops
+        if (processed_response.functions or processed_response.computer_actions):
+            # Reset agent's model_settings
+            if agent.model_settings.tool_choice == "required" or isinstance(agent.model_settings.tool_choice, str):
+                # Create a new model_settings to avoid modifying the original shared instance
+                agent.model_settings = ModelSettings(
+                    temperature=agent.model_settings.temperature,
+                    top_p=agent.model_settings.top_p,
+                    frequency_penalty=agent.model_settings.frequency_penalty,
+                    presence_penalty=agent.model_settings.presence_penalty,
+                    tool_choice="auto",  # Reset to auto
+                    parallel_tool_calls=agent.model_settings.parallel_tool_calls,
+                    truncation=agent.model_settings.truncation,
+                    max_tokens=agent.model_settings.max_tokens,
+                )
+            
+            # Also reset run_config's model_settings if it exists
+            if run_config.model_settings and (run_config.model_settings.tool_choice == "required" or 
+                                             isinstance(run_config.model_settings.tool_choice, str)):
+                # Create a new model_settings for run_config
+                run_config.model_settings = ModelSettings(
+                    temperature=run_config.model_settings.temperature,
+                    top_p=run_config.model_settings.top_p,
+                    frequency_penalty=run_config.model_settings.frequency_penalty,
+                    presence_penalty=run_config.model_settings.presence_penalty,
+                    tool_choice="auto",  # Reset to auto
+                    parallel_tool_calls=run_config.model_settings.parallel_tool_calls,
+                    truncation=run_config.model_settings.truncation,
+                    max_tokens=run_config.model_settings.max_tokens,
+                )
 
         # Second, check if there are any handoffs
         if run_handoffs := processed_response.handoffs:
--- a/tests/test_prevent_infinite_tool_loop.py
+++ b/tests/test_prevent_infinite_tool_loop.py
@@ -54,7 +54,7 @@ class MockModel:
         )
 
 
-class TestToolChoiceReset:
+class TestPreventInfiniteToolLoop:
     async def test_tool_choice_resets_after_call(self):
         """Test that tool_choice is reset to 'auto' after tool call when set to 'required'"""
         # Create an agent with tool_choice="required"
--- a/tests/test_tool_choice_reset.py
+++ b/tests/test_tool_choice_reset.py
@@ -54,7 +54,7 @@ class MockModel:
         )
 
 
-class TestPreventInfiniteToolLoop:
+class TestToolChoiceReset:
     async def test_tool_choice_resets_after_call(self):
         """Test that tool_choice is reset to 'auto' after tool call when set to 'required'"""
         # Create an agent with tool_choice="required"
--- a/tests/test_tool_choice_reset.py
+++ b/tests/test_tool_choice_reset.py
@@ -1,0 +1,303 @@
+from unittest import mock
+import asyncio
+import json
+from typing import List
+
+from agents import Agent, ModelSettings, RunConfig, function_tool, Runner
+from agents.models.interface import ModelResponse
+from agents.items import Usage
+from openai.types.responses.response_function_tool_call import ResponseFunctionToolCall
+
+
+@function_tool
+def echo(text: str) -> str:
+    """Echo the input text"""
+    return text
+
+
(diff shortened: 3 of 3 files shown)


rm-openai:
This is a good idea! What do you think about making it a configurable param, default to `reset=True`?

mini-peanut:
> This is a good idea! What do you think about making it a configurable param, default to `reset=True`?

@rm-openai Thanks for the feedback! I considered adding a config parameter, but wonder if it might add complexity without clear use cases. Most users would want to prevent infinite loops by default, and those with specific needs could already implement custom behaviors through the existing API.

Unless you have specific scenarios in mind where maintaining forced tool calls is beneficial, perhaps the simpler approach is better?

rm-openai:
@mini-peanut, yeah one use case I had in mind was this:

Setup:
```
agent = Agent(
  instructions="Use the find_company tool to find the company info. Then use the search_directory tool to get the CEO's email.",
  tools=[find_company, search_directory],
  tool_choice="required",
  tool_use_behavior={"stop_at_tool_names": "search_directory"},
```

If we reset `tool_choice`, then we can't trust the Agent to reliably call the second tool.

Thoughts?

mini-peanut:
> @mini-peanut, yeah one use case I had in mind was this:
> 
> Setup:
> 
> ```
> agent = Agent(
>   instructions="Use the find_company tool to find the company info. Then use the search_directory tool to get the CEO's email.",
>   tools=[find_company, search_directory],
>   tool_choice="required",
>   tool_use_behavior={"stop_at_tool_names": "search_directory"},
> ```
> 
> If we reset `tool_choice`, then we can't trust the Agent to reliably call the second tool.
> 
> Thoughts?

@rm-openai Thanks for sharing that use case. I'd like to refine my approach to focus on the specific problem we're solving.

**The Problem:** Setting `tool_choice` to "required" or a specific function name can inadvertently cause infinite loops.

**Core Hypothesis:** When a user forces a single specific function call, they rarely intend for that same function to be repeatedly called in an infinite loop. This differs from intentional sequential calling of different functions.

**Problem Scenario:** This issue typically manifests in two specific cases:
1. When `tool_choice` is set to a specific function name, causing the same function to be called repeatedly
2. When `tool_choice="required"` with only one available tool, which functionally behaves the same way

**Concerns with Adding a Configuration Parameter:**
Users with legitimate sequential tool usage would need to explicitly set `reset_tool_choice_after_use` to `False`.

**Targeted Solution:** We can address these specific scenarios without disrupting legitimate use cases:
```python
# Only reset in the problematic scenarios where loops are likely unintentional
if (isinstance(tool_choice, str) and tool_choice not in ["auto", "required", "none"]) or 
   (tool_choice == "required" and len(tools) == 1):
    # Reset to "auto"
```

This approach precisely targets the infinite loop problem without affecting the multi-tool sequential calling pattern you described, and without requiring additional configuration.

rm-openai (dismissed):

rm-openai:
lgtm - but would you mind fixing lint/typechecking please? can't merge without that

src/agents/_run_impl.py
```diff
@@ -206,6 +207,37 @@ async def execute_tools_and_side_effects(
         new_step_items.extend([result.run_item for result in function_results])
         new_step_items.extend(computer_results)
 
+        # Reset tool_choice to "auto" after tool execution to prevent infinite loops
+        if (processed_response.functions or processed_response.computer_actions):
+            # Reset agent's model_settings
+            if agent.model_settings.tool_choice == "required" or isinstance(agent.model_settings.tool_choice, str):
+                # Create a new model_settings to avoid modifying the original shared instance
+                agent.model_settings = ModelSettings(
```
baderalfahad: I ran this code and it looks like the changes persist across agent runs? Not sure if this is the intended behavior.

src/agents/_run_impl.py
```diff
@@ -206,6 +207,37 @@ async def execute_tools_and_side_effects(
...
+                    parallel_tool_calls=agent.model_settings.parallel_tool_calls,
+                    truncation=agent.model_settings.truncation,
+                    max_tokens=agent.model_settings.max_tokens,
+                )
+            
+            # Also reset run_config's model_settings if it exists
+            if run_config.model_settings and (run_config.model_settings.tool_choice == "required" or 
+                                             isinstance(run_config.model_settings.tool_choice, str)):
+                # Create a new model_settings for run_config
+                run_config.model_settings = ModelSettings(
```
baderalfahad: Perhaps you could use `dataclasses.replace` here instead?

src/agents/_run_impl.py
mini-peanut: Yes, the changes to tool_choice do persist across agent runs. This is the intended behavior. When tool_choice is set to "required" or a specific function name, it's automatically reset to "auto" after a tool call to prevent infinite loops.

src/agents/_run_impl.py
mini-peanut: Great suggestion! Your recommendation to use dataclasses.replace() has been implemented.

mini-peanut:
@rm-openai Fixed, and the code should pass the checks. Thanks for your patience

rm-openai:
Unfortunately looks like typechecking is still not passing

src/agents/_run_impl.py
```diff
@@ -206,6 +208,25 @@ async def execute_tools_and_side_effects(
         new_step_items.extend([result.run_item for result in function_results])
         new_step_items.extend(computer_results)
 
+        # Reset tool_choice to "auto" after tool execution to prevent infinite loops
+        if processed_response.functions or processed_response.computer_actions:
+            tools = agent.tools
+            # Only reset in the problematic scenarios where loops are likely unintentional
+            if cls._should_reset_tool_choice(agent.model_settings, tools):
+                agent.model_settings = dataclasses.replace(
```
baderalfahad: As mentioned earlier, it looks like the original agent instance is being modified. This means the agent may not be guaranteed to use the tool in subsequent runs. This may not be the intended behavior.

Example:
```
agent = Agent(
    ...
    tools=[custom_tool],
    model_settings=ModelSettings(tool_choice="required"),
)

await Runner.run(agent, "first run")   # starts run with tool_choice="required"
await Runner.run(agent, "second run")  # starts run with tool_choice="auto"
```

src/agents/_run_impl.py
```diff
@@ -206,6 +208,25 @@ async def execute_tools_and_side_effects(
...
+                    agent.model_settings,
+                    tool_choice="auto"
+                )
+
+            if (
```
baderalfahad, line 215: If modifying the agent model settings is necessary, this should ideally only be done after checking the run config first. This is because the run config overrides agent-specific settings for a run.

src/agents/_run_impl.py
baderalfahad: @rm-openai Thoughts?

rm-openai: hmm that's true, does feel like an issue with this solution. Maybe we should save it as internal state.

mini-peanut: @baderalfahad  @rm-openai I've fixed the issue by creating new copies of the agent and model settings objects instead of modifying the original instances. The linting errors have also been resolved. This should pass the type checking now.

src/agents/_run_impl.py
mini-peanut: Good point. These two operations should actually be parallel rather than sequential, but putting the run_config check first does make more sense. I've made this change in the latest commit.

rm-openai (approved):

rm-openai:
I'm merging this because it's mostly great. I think it will need a couple of followups:
1. Instead of copying the agent, we should do internal bookkeping of the resets
2. I still think this should be configurable
3. I'm not sure it makes sense to reset the RunConfig ModelSettings. 

I'll follow up with all of those!

PR #269, opened 2025-03-20
chore: add missing openai.fm model names to `TTSModelSettings.voice` type

hironow:
## changes

Include two missing model names ( `ballad` , `verse` ) from [openai.fm](https://www.openai.fm/) in the voice type definition to ensure coverage of all TTS voices.

*Note that a simultaneous update of the dependency module is required for full compatibility—this is the reason why tests are failing.*
https://github.com/openai/openai-python/blob/6d0ecdd8ecbface903cf93c7571398b90b803b0b/src/openai/resources/audio/speech.py#L56


rm-openai:
looks like tests are failing?

hironow:
Thanks for your review! @rm-openai 

I realized that I omitted some details in the description.
It appears that the test failures are due to this process, as explained [here](https://github.com/openai/openai-python/blob/f66d2e6fdc51c4528c99bb25a8fbca6f9b9b872d/src/openai/resources/audio/speech.py#L56
) (occurred type mismatching):

To clarify, the changes work together with the related PR in openai-python(deps repo) as follows:

1. The model type is updated in openai-python (deps repo and https://github.com/openai/openai-python/pull/2237 ).
2. The updated version is then used as a dependency in openai-agents-python (this repo).
3. Finally, the model type is updated in openai-agents-python (this repo and this PR).

Please note that first, the addition of the model type on the openai-python side must be completed. After that, when openai-agents-python uses the updated version (step 2), there is a concern that the typecheck might still fail.

I appreciate your attention to this matter!

rm-openai:
Makes sense. I think we should probably update this param in the agents SDK to be a `str | Literal`, so that you can pass arbitrary model names

hironow:
That makes sense. However, even with the parameter updated to `str | Literal`, passing a string now causes an error. Should we wait for an update on the openai-python side, or use a `type: ignore` workaround for now?

suggestion [here](https://github.com/hironow/openai-agents-python/blob/2f6d8b56415610e251ccbd1defbfe105efd0a548/src/agents/voice/models/openai_tts.py#L44):
```py
voice=settings.voice or DEFAULT_VOICE, # type: ignore
```

rm-openai:
yes, we'll be updating the openai SDK soon

hironow:
`openai-python` repository has been updated, and the test failures have been resolved—just wanted to share that with you! @rm-openai 

It seems that the dependency on `numpy` changed with the version update, so I’ve added numpy using `uv add`.

hironow:
closing this as the issue has already been addressed in a better way elsewhere.

PR #288, opened 2025-03-21
Replace remaining uses of `fetch_ordered_spans` and `fetch_traces` for stronger tests

alexmojaki:
Following https://github.com/openai/openai-agents-python/pull/261


rm-openai (approved):
🙏

PR #289, opened 2025-03-21
Read tracing API data lazily

rm-openai:
A common complaint is that people load their API keys using dotenv:
```
import agents # imports tracing by default

load_dotenv()
```

This causes issues, because by the time you load your dotenv, the batch processor has already read the API key.

Let's lazy load it, so it's deferred to when the processor actually _needs_ the key.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -5,6 +5,7 @@ import queue
 import random
 import threading
 import time
+from functools import cached_property
 from typing import Any
 
 import httpx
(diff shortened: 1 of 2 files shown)


src/agents/tracing/processors.py
```diff
@@ -70,10 +72,24 @@ def set_api_key(self, api_key: str):
...
+        if self._lazy_read_complete:
+            return
+
+        self._lazy_read_complete = True
+        if not self.api_key:
+            self.api_key = os.environ.get("OPENAI_API_KEY")
+        if not self.organization:
+            self.organization = os.environ.get("OPENAI_ORG_ID")
+        if not self.project:
+            self.project = os.environ.get("OPENAI_PROJECT_ID")
```
stevenheidel: consider using built-in functools.cached_property from python 3.8 rather than managing state yourself:

```suggestion
    @cached_property
    def api_key(self):
        return self._api_key or os.environ.get("OPENAI_API_KEY")
    
    @cached_property
    def organization(self):
        return self._organization or os.environ.get("OPENAI_ORG_ID")

    @cached_property
    def project(self):
        return self._project or os.environ.get("OPENAI_PROJECT_ID")
```

stevenheidel (approved):

Issue #291, opened 2025-03-21
circular import issue in streamed voice example


gaviral:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) ✅
- **Have you searched for related issues?** Others may have faced similar issues. ✅

### Describe the bug

```
uv run examples/voice/streamed/main.py


Traceback (most recent call last):
  File "/Users/aviralgarg/code/openai-agents-python/examples/voice/streamed/main.py", line 14, in <module>
    from agents.voice import StreamedAudioInput, VoicePipeline
  File "/Users/aviralgarg/code/examples/voice/streamed/main.py", line 5, in <module>
    from agents import Agent, Runner, TResponseInputItem, function_tool
ImportError: cannot import name 'Agent' from partially initialized module 'agents' (most likely due to a circular import) (/Users/aviralgarg/code/examples/voice/streamed/agents.py)

```

circular import issue.

fix:

```
mv examples/voice/streamed/agents.py examples/voice/streamed/my_workflow.py
```

and replace `from .agents import MyWorkflow` with `from my_workflow import MyWorkflow` in examples/voice/streamed/main.py

### Debug information
- Agents SDK version: v0.0.6
- Python version 3.11.11

### Repro steps

```
uv run examples/voice/streamed/main.py
```

### Expected behavior

voice streamed example should have run.

rm-openai:
thanks for fixing.

PR #292, opened 2025-03-21
Fix circular dependency in voice streamed example by renaming agents #291

gaviral:
Fix circular dependency in voice streamed example by renaming agents.py to my_workflow.py

Fix circular dependency in voice streamed example by renaming agents #291


rm-openai (approved):

rm-openai:
Looks like some checks are failing

gaviral:
> Looks like some checks are failing

on it.

rm-openai (approved):

PR #297, opened 2025-03-22
Refactor: Add explicit TypeAlias for VoiceStreamEvent using modern Py…

EnggQasim:
…thon typing |Update events.py

This PR improves type hint clarity and consistency by defining `VoiceStreamEvent` as an explicit `TypeAlias` using Python 3.10+ syntax.

### 🔧 What was changed:
```python
VoiceStreamEvent: TypeAlias = (
    VoiceStreamEventAudio | VoiceStreamEventLifecycle | VoiceStreamEventError
)
```

### 🧠 Why:
- ✅ **Explicit Type Alias**: Clearly indicates that `VoiceStreamEvent` is a type alias, not a runtime variable (per [PEP 613](https://peps.python.org/pep-0613/)).
- ✅ **Improved Readability**: Uses the modern `|` union syntax, which is more concise and readable than `Union[...]`.
- ✅ **Better Tooling Support**: Helps static type checkers (like MyPy and Pyright) understand developer intent, improving accuracy and IDE support.

### 📌 Note:
This change does not affect runtime behavior. It purely enhances type safety and code quality for developers and static analyzers.

### DIFF
--- a/src/agents/voice/events.py
+++ b/src/agents/voice/events.py
@@ -41,7 +41,6 @@ class VoiceStreamEventError:
     """The type of event."""
 
 
-VoiceStreamEvent: TypeAlias = Union[
-    VoiceStreamEventAudio, VoiceStreamEventLifecycle, VoiceStreamEventError
-]
+VoiceStreamEvent: TypeAlias = ( VoiceStreamEventAudio | VoiceStreamEventLifecycle | VoiceStreamEventError )
+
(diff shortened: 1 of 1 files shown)


rm-openai:
We can't use some of this stuff because the repo is 3.9+.

Btw I've noticed a lot of PRs from you - appreciate the effort, but not all of them are aligned with the requirements of this repo (eg this one breaks 3.9 compat). It might make sense to open issues first, and focus on ones that have broad value!

PR #125, opened 2025-03-13
Added LiteLLM support

pontus-devoteam:
Added Lite LLM support to ensure we can cover more providers to be used when building agents.

PR #318, opened 2025-03-24
Add LiteLLM support in line with AsyncOpenAI and with reproducible examples

vmayoral:
This PR is a follow up from #125, which unsuccessfully delivered LiteLLM support. See conversation there for further context.

Given the late architectural updates, there's no real need for any relevant code changes to support LiteLLM. Correspondingly, this PR delivers a working example under the `examples/model_providers` support to easy adoption for others. In particular, the example leverages `AsyncOpenAI` proposed extensions, as well as a default generic configuration demonstrating basic use with SOTA models.

Empirical testing of proposed changes:

```bash
(env) alias@brain:~/Alias/research/openai-agents-python (main*) $ python3 examples/agent_patterns/litellm.py  # with gpt-4o
Code calls itself now,  
Infinite loops echoing,  
Depth finds its own end.

(env) alias@brain:~/Alias/research/openai-agents-python (main*) $ python3 examples/agent_patterns/litellm.py   # with claude-3-7
# Recursion Haiku

Function calls itself,
Diving deeper, then returns—
Loops within themselves.

(env) alias@brain:~/Alias/research/openai-agents-python (main*) $ python3 examples/agent_patterns/litellm.py    # with qwen2.5:14b
Function calls itself,
Depth of thought in endless loop,
Circle spins to start.
```


pyproject.toml
```diff
@@ -7,7 +7,7 @@ requires-python = ">=3.9"
 license = "MIT"
 authors = [{ name = "OpenAI", email = "support@openai.com" }]
 dependencies = [
-    "openai>=1.66.5",
+    "openai>=1.68.2",
```
mjio, line 10: I think this should not be part of the PR, as it has nothing to do with the added example.

vmayoral: In my local testing, it was neccessary, otherwise it wouldn't work with `1.66.5`.

Happy to change it though if that gets this through.

PR #320, opened 2025-03-24
[1/n] Add MCP types to the SDK

rm-openai:
### Summary:
1. Add the MCP dep for python 3.10, since it doesn't support 3.9 and below
2. Create MCPServer, which is the agents SDK representation of an MCP server
3. Create implementations for HTTP-SSE and StdIO servers, directly copying the [MCP SDK example](https://github.com/modelcontextprotocol/python-sdk/blob/main/examples/clients/simple-chatbot/mcp_simple_chatbot/main.py)
4. Add a util to transform MCP tools into Agent SDK tools

Note: I added optional caching support to the servers. That way, if you happen to know a server's tools don't change, you can just cache them.

### Test Plan:

Checks pass. I added tests at the end of the stack.

--- 

#324
#322
#321
-> #320
#319

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -1,0 +1,269 @@
+from __future__ import annotations
+
+import abc
+import asyncio
+from contextlib import AbstractAsyncContextManager, AsyncExitStack
+from pathlib import Path
+from typing import Any, Literal
+
+from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
+from mcp import ClientSession, StdioServerParameters, Tool as MCPTool, stdio_client
+from mcp.client.sse import sse_client
+from mcp.types import CallToolResult, JSONRPCMessage
+from typing_extensions import NotRequired, TypedDict
+
+from ..exceptions import UserError
+from ..logger import logger
+
+
+class MCPServer(abc.ABC):
+    """Base class for Model Context Protocol servers."""
+
+    @abc.abstractmethod
+    async def connect(self):
+        """Connect to the server. For example, this might mean spawning a subprocess or
+        opening a network connection. The server is expected to remain connected until
+        `cleanup()` is called.
+        """
+        pass
+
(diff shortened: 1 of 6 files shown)


src/agents/mcp/util.py
```diff
@@ -0,0 +1,96 @@
...
+from ..run_context import RunContextWrapper
+from ..tool import FunctionTool, Tool
+
+if TYPE_CHECKING:
+    from mcp.types import Tool as MCPTool
+
+    from .server import MCPServer
+
+
+class MCPUtil:
```
stevenheidel, line 17: this is the same as the other MCPUtil class

src/agents/mcp/server.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @abc.abstractmethod
+    async def list_tools(self) -> list[MCPTool]:
+        """List the tools available on the server."""
+        pass
+
+    @abc.abstractmethod
+    async def call_tool(self, tool_name: str, arguments: dict[str, Any] | None) -> CallToolResult:
+        """Invoke a tool on the server."""
+        pass
```
stevenheidel, line 45: what do you think about making these just return our `Tool` class? ie. the MCPServer is responsible for doing the conversion

this is an MCPServer class specifically for integrating with our SDK, so it can use our types

src/agents/mcp/util.py
rm-openai: Yeah sorry deleted in the end of stack.

src/agents/mcp/server.py
rm-openai: Yeah good question. I considered that, but I actually think MCP servers should be shared between frameworks. I talked to @dsp-ant and plan to make a PR to the MCP repo to make this part of their SDK, so then this import would just be an alias to `mcp.client.MCPServer`.

stevenheidel (dismissed):

PR #333, opened 2025-03-25
Fix parallel_tool_calls when False

apeccaud:
Currently, when we set `parallel_tool_calls=False` in the `model_settings`, the responses API is called with `parallel_tool_calls == NotGiven`, which defaults to true on the Response API side (https://platform.openai.com/docs/api-reference/responses/create#responses-create-parallel_tool_calls).

This PR attempts to fix that.

```
agent = Agent(
    ...,
    model_settings=ModelSettings(
        parallel_tool_calls=False,
    ),
)
```


rm-openai (approved):

rm-openai:
Thanks for fixing!

PR #335, opened 2025-03-25
Make the reset behavior on tool use configurable

rm-openai:
## Summary:

#263 added this behavior. The goal was to prevent infinite loops when tool choice was set. The key change I'm making is:
1. Making it configurable on the agent.
2. Doing bookkeeping in the Runner to track this, to prevent mutating agents.
3. Not resetting the global tool choice in RunConfig.

## Test Plan:
Unit tests.
.

### DIFF
--- a/src/agents/_run_impl.py
+++ b/src/agents/_run_impl.py
@@ -4,7 +4,7 @@ import asyncio
 import dataclasses
 import inspect
 from collections.abc import Awaitable
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, cast
 
 from openai.types.responses import (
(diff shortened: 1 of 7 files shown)


src/agents/_run_impl.py
```diff
@@ -77,6 +77,22 @@ class QueueCompleteSentinel:
 _NOT_FINAL_OUTPUT = ToolsToFinalOutputResult(is_final_output=False, final_output=None)
 
 
+@dataclass
+class AgentToolUseTracker:
+    data: list[tuple[Agent, list[str]]] = field(default_factory=list)
```
stevenheidel: better name than `data`? i couldn't figure out what this is

src/agents/models/openai_responses.py
```diff
@@ -208,8 +208,10 @@ async def _fetch_response(
         list_input = ItemHelpers.input_to_new_input_list(input)
 
         parallel_tool_calls = (
```
stevenheidel, line 210: this would be easier to read as an if/elif/else block rather than this expression

stevenheidel (dismissed):

src/agents/_run_impl.py
rm-openai: will rename

PR #339, opened 2025-03-25
triage agent tool choice requried + respond to user

jhills20:


docs/examples.md
```diff
@@ -7,14 +7,14 @@ Check out a variety of sample implementations of the SDK in the examples section
 
 - **agent_patterns:**
   Examples in this category illustrate common agent design patterns, such as
-  
+
```
rm-openai, line 10: Can you add links to these? Makes it more useful that way

examples/customer_service/main.py
```diff
@@ -122,7 +128,12 @@ async def on_seat_booking_handoff(context: RunContextWrapper[AirlineAgentContext
     handoffs=[
         faq_agent,
         handoff(agent=seat_booking_agent, on_handoff=on_seat_booking_handoff),
+        respond_to_user
     ],
+    tools=[respond_to_user],
+    model_settings=ModelSettings(tool_choice="required"),
+    tool_use_behavior={"stop_at_tool_names": ["respond_to_user"]},
```
rm-openai, line 135: do we need this any more after #335?

rm-openai (changes requested):
checks are failing too

docs/examples.md
jhills20: yes good call

PR #343, opened 2025-03-25
Adding Git MCP server example

dmitry-openai:
- Adding new example showing a single agent interacting with local MCP server for git


rm-openai (approved):
LGTM - can you add a README similar to the other MCP example?

rm-openai (approved):

PR #354, opened 2025-03-26
MCP example for SSE

rm-openai:
Simple example with a local SSE server.


dmitry-openai (approved):
TY for the cache/formatting fixes also

PR #361, opened 2025-03-27
add Self type

linpan:


rm-openai (changes requested):
i think cleanup actually returns None not Self?

stevenheidel:
also typing.Self is only 3.11+

PR #385, opened 2025-03-28
Update tracing.md - Added Okahu-Monocle

hocokahu:
Monocle is an open source tracing and eval for GenAI apps.


hocokahu:
@rm-openai Assuming this needs your review, Rohan?

PR #386, opened 2025-03-28
Fix incorrect slot name on `SpeechSpanData` 

tcdent:
Slot had inconsistent name. 

`"first_byte_at"` -> `"first_content_at"`


rm-openai (approved):
thanks!

PR #387, opened 2025-03-28
Support Recursive Nodes in OpenAI Agents Draw Graph Visualization Extension

NicoHwpo:
This PR enhances the OpenAI Agents Graph extension by adding support for recursive nodes and cycles. Currently, the extension does not properly graph an agent when it is recursive or has a path that loops back to itself.  

#### **Changes Introduced:**  
- Implemented an integrated method to track visited nodes (handoffs) to prevent infinite loops.  
- Updated the graphing logic to correctly represent recursive nodes with return arrows.  

This change improves the visualization of agents with cyclic or recursive behaviors, making the graph more accurate and comprehensive.

### DIFF
--- a/src/agents/extensions/visualization.py
+++ b/src/agents/extensions/visualization.py
@@ -6,7 +6,6 @@ from agents import Agent
 from agents.handoffs import Handoff
 from agents.tool import Tool
 
-
 def get_main_graph(agent: Agent) -> str:
     """
     Generates the main graph structure in DOT format for the given agent.
@@ -31,7 +30,7 @@ def get_main_graph(agent: Agent) -> str:
     return "".join(parts)
 
 
-def get_all_nodes(agent: Agent, parent: Optional[Agent] = None) -> str:
+def get_all_nodes(agent: Agent, parent: Optional[Agent] = None, visited=None) -> str:
     """
     Recursively generates the nodes for the given agent and its handoffs in DOT format.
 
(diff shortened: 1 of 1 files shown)


MartinEBravo:
Awesome.

src/agents/extensions/visualization.py
```diff
@@ -41,47 +40,46 @@ def get_all_nodes(agent: Agent, parent: Optional[Agent] = None) -> str:
     Returns:
         str: The DOT format string representing the nodes.
     """
+    if visited is None:
+        visited = set()
+    
+    if agent.name in visited:
```
rm-openai, line 46: multiple agents can share the same name, perhaps we can use the object instead?

src/agents/extensions/visualization.py
```diff
@@ -41,47 +40,46 @@ def get_all_nodes(agent: Agent, parent: Optional[Agent] = None) -> str:
...
-        '"__end__" [label="__end__", shape=ellipse, style=filled, '
-        "fillcolor=lightblue, width=0.5, height=0.3];"
-    )
-    # Ensure parent agent node is colored
+    # Start and end nodes
     if not parent:
         parts.append(
-            f'"{agent.name}" [label="{agent.name}", shape=box, style=filled, '
-            "fillcolor=lightyellow, width=1.5, height=0.8];"
+            '"__start__" [label="__start__", shape=ellipse, style=filled, '
```
rm-openai, line 55: I know this isn't directly related to the PR, but it would be cool if we could first build a graph representation, and then second turn that into a visualization. That way, you could have different graph formats (graphviz, mermaid etc), and share code between them,

rm-openai (changes requested):
Generally looks good, can you please add some tests for this?

PR #394, opened 2025-03-29
Avoid to run the Agent first step before Input guardrails finished

rach:
The guardrail may trigger a Tripewire but the agent still run

In v0.0.7, the code runs the input guardrail task and the agent's first step together asynchronously. 

If the guardrail is in place to avoid a side effect from the Agent run, the agent may still do something unexpected.

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -212,15 +212,14 @@ class Runner:
                     )
 
                     if current_turn == 1:
-                        input_guardrail_results, turn_result = await asyncio.gather(
-                            cls._run_input_guardrails(
+                        input_guardrail_results = await cls._run_input_guardrails(
                                 starting_agent,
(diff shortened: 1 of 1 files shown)


rm-openai:
the current behavior is the intended/documented behavior. If you want to run the guardrail _before_ the agent runs, you can just do that via python code:
```
triggered = await do_something()
if not triggered:
  await Runner.run(...)
```

rach:
For my own understanding. What was the rational to have the agent start before the guardrail is done?
I can work around it, now that I know the behavior.

rm-openai:
@rach for latency. If the guardrail doesn't fail, then you will have made a bunch of progress on the actual agent. RUnning (guardrail, agent) in parallel is much faster than serially

Issue #404, opened 2025-03-31
Support strict schemas for MCP tools


jezell:
Strict tool calling significantly improves OpenAI's ability to send proper payloads to JSON tools. Unfortunately, MCP tools always use non strict tool calling in the Agents SDK, increasing the likelihood that bad data will be sent over the wire. Not all MCP schemas play nicely with strict tool callings current limitations, but when the schema is supported, but it would be nice to be able to opt into strict tool calling when the schema is supported.

rm-openai:
Should be good to go in the next version (0.0.8) @jezell

PR #414, opened 2025-04-01
Convert MCP schemas to strict where possible

rm-openai:
## Summary:
Towards #404. I made this configurable because it's not clear this is always a good thing to do. I also made it default to False because I'm not sure if this could cause errors.

If it works out well, we can switch the default in the future as a small breaking changes

## Test Plan:
Unit tests

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -6,7 +6,7 @@ from collections.abc import Awaitable
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Callable, Generic, Literal, cast
 
-from typing_extensions import TypeAlias, TypedDict
+from typing_extensions import NotRequired, TypeAlias, TypedDict
 
 from .guardrail import InputGuardrail, OutputGuardrail
 from .handoffs import Handoff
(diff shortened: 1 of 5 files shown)


pakrym-oai (approved):

src/agents/agent.py
```diff
@@ -53,6 +53,15 @@ class StopAtTools(TypedDict):
     """A list of tool names, any of which will stop the agent from running further."""
 
 
+class MCPConfig(TypedDict):
+    """Configuration for MCP servers."""
+
+    convert_schemas_to_strict: NotRequired[bool]
```
pakrym-oai, line 59: It slightly feels like the wrong place to do this. IMO the MCP server should be responsible to declaring itself as strict.

rm-openai: Assuming you mean the actual MCP server, then yes I agree. That is a more involved change though, since MCP servers adhere to a published spec.

I didn't add it to the MCP server class in the repo, because I think those should be very thin wrappers around the actual MCP server, in order to be reused in other places outside the agents SDK internals.  

Gonna merge this - but if you think we should implement it differently ping me and we can revert/re-do it!

PR #431, opened 2025-04-03
Add metadata to ModelSettings

drewy-openai:
Allows metadata to be supplied to the `responses.create` call. This can, for example, be used to connect response logs to traces groups by passing the group_id.

### DIFF
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -537,6 +537,7 @@ class OpenAIChatCompletionsModel(Model):
             stream_options={"include_usage": True} if stream else NOT_GIVEN,
             store=store,
             extra_headers=_HEADERS,
+            metadata=model_settings.metadata,
         )
 
         if isinstance(ret, ChatCompletion):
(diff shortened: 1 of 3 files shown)


rm-openai (approved):

src/agents/models/openai_chatcompletions.py
```diff
@@ -537,6 +537,7 @@ async def _fetch_response(
             stream_options={"include_usage": True} if stream else NOT_GIVEN,
             store=store,
             extra_headers=_HEADERS,
+            metadata=model_settings.metadata,
```
RahulVerma989, line 540: It should be metadata=self._non_null_or_not_given(model_settings.metadata)

src/agents/models/openai_responses.py
```diff
@@ -247,6 +247,7 @@ async def _fetch_response(
             extra_headers=_HEADERS,
             text=response_format,
             store=self._non_null_or_not_given(model_settings.store),
+            metadata=model_settings.metadata,
```
RahulVerma989, line 250: It should be metadata=self._non_null_or_not_given(model_settings.metadata)

PR #445, opened 2025-04-06
[Visualization Extension] Enhance agent graph visualization to prevent infinite recursion, add mermaid graph

ashish-dahal:
Improves agent graph visualization logic:
- Safely handles recursive agent handoffs during visualization
- Adds a test case for recursive handoff loops
- Cleans up redundant code and improves readability in the visualization test

Improves on PR #387, addressing pending change requests.

### DIFF
--- a/src/agents/extensions/visualization.py
+++ b/src/agents/extensions/visualization.py
@@ -1,10 +1,11 @@
-from typing import Optional, Set, Dict, List, TypeVar, Generic
+import abc
+import base64
+import warnings
 from dataclasses import dataclass
 from enum import Enum
-import warnings
-import abc
+from typing import Generic, Optional, TypeVar
+
 import graphviz
-import base64
 import requests
 
 from agents import Agent
@@ -41,18 +42,18 @@ class Edge:
 
 class Graph:
     def __init__(self):
-        self.nodes: Dict[str, Node] = {}
-        self.edges: List[Edge] = []
+        self.nodes: dict[str, Node] = {}
+        self.edges: list[Edge] = []
 
     def add_node(self, node: Node) -> None:
         self.nodes[node.id] = node
 
     def add_edge(self, edge: Edge) -> None:
         """Add an edge to the graph.
-        
+
         Args:
             edge (Edge): The edge to add.
-            
+
         Raises:
             ValueError: If the source or target node does not exist in the graph.
         """
@@ -64,10 +65,10 @@ class Graph:
 
     def has_node(self, node_id: str) -> bool:
         """Check if a node exists in the graph.
-        
+
         Args:
             node_id (str): The ID of the node to check.
-            
+
         Returns:
             bool: True if the node exists, False otherwise.
         """
@@ -75,10 +76,10 @@ class Graph:
 
     def get_node(self, node_id: str) -> Optional[Node]:
         """Get a node from the graph.
-        
+
         Args:
             node_id (str): The ID of the node to get.
-            
+
         Returns:
             Optional[Node]: The node if it exists, None otherwise.
         """
@@ -87,14 +88,14 @@ class Graph:
 
 class GraphBuilder:
     def __init__(self):
-        self._visited: Set[int] = set()
-        
+        self._visited: set[int] = set()
+
     def build_from_agent(self, agent: Agent) -> Graph:
         """Build a graph from an agent.
-        
+
         Args:
             agent (Agent): The agent to build the graph from.
-            
+
         Returns:
             Graph: The built graph.
         """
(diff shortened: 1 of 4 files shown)


ashish-dahal:
Hey @rm-openai I noticed the original PR hasn’t had updates, so I’ve created a follow-up here with additional fixes and improvements. Happy to adjust things if needed.

ashish-dahal:
Addressed https://github.com/openai/openai-agents-python/pull/387#discussion_r2027363814 and https://github.com/openai/openai-agents-python/pull/387#discussion_r2027368216

src/agents/extensions/visualization.py
```diff
@@ -1,137 +1,426 @@
...
+@dataclass(frozen=True)
+class Node:
+    id: str
+    label: str
+    type: NodeType
+
+
+@dataclass(frozen=True)
+class Edge:
+    source: str
```
rm-openai: shouldn't these be `Node` not `str`?

src/agents/extensions/visualization.py
```diff
@@ -1,137 +1,426 @@
...
+        graph.add_node(Node("__start__", "__start__", NodeType.START))
+        graph.add_node(Node("__end__", "__end__", NodeType.END))
+
+        self._add_agent_nodes_and_edges(agent, None, graph)
+        return graph
+
+    def _add_agent_nodes_and_edges(
+        self,
+        agent: Agent,
+        parent: Optional[Agent],
```
rm-openai, line 114: can you use Agent | None instead?

src/agents/extensions/visualization.py
```diff
@@ -1,137 +1,426 @@
...
+        graph: Graph,
+    ) -> None:
+        # Add agent node
+        graph.add_node(Node(agent.name, agent.name, NodeType.AGENT))
```
rm-openai: using `agent.name` as the ID isn't safe unfortunately, since agents can have the same name. You'd need to do an instance equality check everywhere instead.

src/agents/extensions/visualization.py
```diff
@@ -1,137 +1,426 @@
...
+            webbrowser.open(f"file://{os.path.abspath(self.filename)}.png")
+        else:
+            temp_dir = tempfile.gettempdir()
+            temp_path = os.path.join(temp_dir, next(tempfile._get_candidate_names()))
+            self.renderer.save(self.rendered_graph, temp_path)
+            webbrowser.open(f"file://{os.path.abspath(temp_path)}.png")
+
+
+def draw_graph(
+    agent: Agent, filename: Optional[str] = None, renderer: str = "graphviz"
```
rm-openai: `renderer` should be either a Literal or an enum here

Also please add `str | None = None` instead of Optional

rm-openai (changes requested):
Thanks, PR is looking great! Some changes requested inline

src/agents/extensions/visualization.py
ashish-dahal: It works when agent.handoffs have Agent instances. but apparently it's a bit tricky when Handoff objects are passed. It only has the target agent's name ([handoff.agent_name](https://github.com/openai/openai-agents-python/blob/d089886f01d46e3a4cd550d50a68d15a56cbba78/src/agents/handoffs.py#L78)), not an Agent instance. The instance itself only gets resolved at runtime with [handoff.on_invoke_handoff](https://github.com/openai/openai-agents-python/blob/d089886f01d46e3a4cd550d50a68d15a56cbba78/src/agents/handoffs.py#L70).  so we can't consistently do instance equality checks to link Handoff instances back to the Agent instances at build time if we consider cases with recursive handoffs. The only option I see is to resolve this using agent names but mention the caveat in the docs. any suggestions?

rm-openai (changes requested):
same as before

ashish-dahal:
should be looking good now

src/agents/extensions/visualization.py
rm-openai: same comment

src/agents/extensions/visualization.py
```diff
@@ -1,137 +1,436 @@
...
+            filename (str): The name of the file to save the graph as.
+        """
+        # Encode the graph to base64
+        graphbytes = rendered.encode("utf8")
+        base64_bytes = base64.urlsafe_b64encode(graphbytes)
+        base64_string = base64_bytes.decode("ascii")
+
+        # Get the image from mermaid.ink
+        response = requests.get(f"https://mermaid.ink/img/{base64_string}")
+        response.raise_for_status()
```
rm-openai, line 278: I don't love this - relying on a 3rd party service that could return anything or leak information.

I'd actually prefer if we got rid of the Renderer base class, and instead had:
```
class GraphVizRenderer:
  @classmethod 
  def render_to_file(...)

  @classmethod 
  def render_to_object(...)

class MermaidRenderer:
  @classmethod 
  def render_mermaid_ink_link(...)

  @classmethod 
  def render_mermaid_text(...)
```

rm-openai (commented):
one more change sorry!

src/agents/extensions/visualization.py
ashish-dahal: yep, on it!

Issue #415, opened 2025-04-02
Add `reasoning_content` to ChatCompletions


xsoutline:
https://github.com/openai/openai-agents-python/blob/01f5e86ea546905c3d935b6f25c9b074acbf85d6/src/agents/models/openai_chatcompletions.py#L210-L213

在以上代码的位置需要添加处理reasoning_content的流式响应信息事件,或者添加能够添加自定义事件处理程序
示例：
```python
                delta = chunk.choices[0].delta

                # 尝试获取 reasoning_content（模型思考部分），防止属性不存在使用 getattr
                reasoning = getattr(delta, "reasoning_content", None)
                if reasoning:
                    print(reasoning, end='', flush=True)

                # Handle text
                if delta.content:
```

weixiewen:
While...delta is from ChoiceDelta, which is from openai types chat...

Ddper:
Hi @rm-openai I create a PR to solve this issue. 
If using reasoning model, it will yield reasoning_content before yield content.  It'll yield a title before yield the real content so the user can split the content and reasoning content. Another way is add a `<think></think>` tag.
Need your opinion.

Grjia:
@Ddper Did you give up

seratch:
The latest code should handle the property properly. If there is still anything missing, please feel free to create a new issue with details.

PR #494, opened 2025-04-14
add reasoning content to ChatCompletions

Ddper:
fix issue https://github.com/openai/openai-agents-python/issues/415

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -48,6 +48,9 @@ class ChatCmplStreamHandler:
         usage: CompletionUsage | None = None
         state = StreamingState()
 
+        is_reasoning_model = False
+        emit_reasoning_content = False
+        emit_content = False
         async for chunk in stream:
             if not state.started:
                 state.started = True
@@ -62,9 +65,16 @@ class ChatCmplStreamHandler:
                 continue
 
             delta = chunk.choices[0].delta
+            reasoning_content = None
+            content = None
+            if hasattr(delta, "reasoning_content"):
+                reasoning_content = delta.reasoning_content
+                is_reasoning_model = True
+            if hasattr(delta, "content"):
+                content = delta.content
 
             # Handle text
-            if delta.content:
+            if reasoning_content or content:
                 if not state.text_content_index_and_output:
                     # Initialize a content tracker for streaming text
                     state.text_content_index_and_output = (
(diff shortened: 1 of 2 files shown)


src/agents/models/chatcmpl_stream_handler.py
```diff
@@ -100,16 +110,59 @@ async def handle_stream(
...
-                    type="response.output_text.delta",
-                )
-                # Accumulate the text into the response part
-                state.text_content_index_and_output[1].text += delta.content
+
+                if reasoning_content is not None:
+                    if not emit_reasoning_content:
+                        emit_reasoning_content = True
+
+                        reasoning_content_title = "# reasoning content\n\n"
```
rm-openai, line 118: this doesn't seem right - why hardcode?

src/agents/models/chatcmpl_stream_handler.py
```diff
@@ -100,16 +110,59 @@ async def handle_stream(
...
+                        output_index=0,
+                        type="response.output_text.delta",
+                    )
+                    # Accumulate the text into the response part
+                    state.text_content_index_and_output[1].text += reasoning_content
+
+                if content is not None:
+                    if not emit_content and is_reasoning_model:
+                        emit_content = True
+                        content_title = "\n\n# content\n\n"
```
rm-openai, line 144: same here?

src/agents/models/chatcmpl_stream_handler.py
```diff
@@ -62,9 +65,16 @@ async def handle_stream(
                 continue
 
             delta = chunk.choices[0].delta
+            reasoning_content = None
+            content = None
+            if hasattr(delta, "reasoning_content"):
```
rm-openai, line 70: when would this be true?

rm-openai (changes requested):
Looks reasonable - can you please add:
1. Unit tests
2. A working example that shows this?

src/agents/models/chatcmpl_stream_handler.py
Ddper: When using reasoning model like deepseek-reasoner
[deepseek reasoning model](https://api-docs.deepseek.com/guides/reasoning_model)

Ddper:
> Looks reasonable - can you please add:
> 
> 1. Unit tests
> 2. A working example that shows this?

OK

src/agents/models/chatcmpl_stream_handler.py
Ddper: It's a markdown title for splitting the content and reasoning content. It's a constant value so have to hardcode.

The whole output are like below:
`
# reasoning content

Okay, I need to write a haiku about recursion in programming. Let's start by recalling what a haiku is. It's a three-line poem with a 5-7-5 syllable structure. The first line has five syllables, the second seven, and the third five again. The theme here is recursion in programming, so I should focus on elements that capture the essence of recursion.
.....
# content

**Haiku on Recursion:**

Function calls itself,  
Base case breaks the looping chain—  
Stack grows, then falls back.
`

Another way is use `<think></think>`
The whole output would be like this:
`
<think>
Okay, I need to write a haiku about recursion in programming. Let's start by recalling what a haiku is. It's a three-line poem with a 5-7-5 syllable structure. The first line has five syllables, the second seven, and the third five again. The theme here is recursion in programming, so I should focus on elements that capture the essence of recursion.
.....
</think>

**Haiku on Recursion:**

Function calls itself,  
Base case breaks the looping chain—  
Stack grows, then falls back.
`

Ddper: It's a markdown title for splitting the content and reasoning content. It's a constant value so have to hardcode.

The whole output are like below:

```
# reasoning content

Okay, I need to write a haiku about recursion in programming. Let's start by recalling what a haiku is. It's a three-line poem with a 5-7-5 syllable structure. The first line has five syllables, the second seven, and the third five again. The theme here is recursion in programming, so I should focus on elements that capture the essence of recursion.
.....
# content

**Haiku on Recursion:**

Function calls itself,  
Base case breaks the looping chain—  
Stack grows, then falls back.

```

Another way is use `<think></think>`
The whole output would be like this:

```

<think>
Okay, I need to write a haiku about recursion in programming. Let's start by recalling what a haiku is. It's a three-line poem with a 5-7-5 syllable structure. The first line has five syllables, the second seven, and the third five again. The theme here is recursion in programming, so I should focus on elements that capture the essence of recursion.
.....
</think>

**Haiku on Recursion:**

Function calls itself,  
Base case breaks the looping chain—  
Stack grows, then falls back.

```

Which way do you prefer?

rm-openai: I think neither? IMO it would be better to emit a separate item for reasoning. For example, I was trying something like this in #581. What do you think?

Ddper: Yeah, I agree with you.
I just found out openai-python 1.7.6 already added the types needed for emit reasoning content.

```
    ResponseReasoningItem,
    ResponseReasoningSummaryTextDeltaEvent,
    ResponseReasoningSummaryPartAddedEvent,
    ResponseReasoningSummaryPartDoneEvent,
    ResponseReasoningSummaryTextDoneEvent
)
```
So we can emit ResponseReasoningSummaryTextDeltaEvent for reasoning content or create some class like ResponseReasoningTextDeltaEvent in this repo?
What do you think?

dongxiaohuang:
any updates?

Issue #490, opened 2025-04-13
Clarify Python Version Compatibility


kimmeoungjun:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) Yes
- **Have you searched for related issues? Others may have had similar requests** Yes

### Question
Hi, I was trying to run the following command:

```
python -m examples.customer_service.main
```

I'm using **Python 3.9.13**, and when I run the script this way, I get the following error:

```
Traceback (most recent call last):
  File "/Users/kimmeoungjun/openai-agents-python/env/lib/python3.9/site-packages/pydantic/_internal/_typing_extra.py", line 466, in _eval_type_backport
    return _eval_type(value, globalns, localns, type_params)
  File "/Users/kimmeoungjun/openai-agents-python/env/lib/python3.9/site-packages/pydantic/_internal/_typing_extra.py", line 500, in _eval_type
    return typing._eval_type(  # type: ignore
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/typing.py", line 292, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/typing.py", line 554, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/kimmeoungjun/openai-agents-python/examples/customer_service/main.py", line 28, in <module>
    class AirlineAgentContext(BaseModel):
  File "/Users/kimmeoungjun/openai-agents-python/env/lib/python3.9/site-packages/pydantic/_internal/_model_construction.py", line 221, in __new__
    set_model_fields(cls, config_wrapper=config_wrapper, ns_resolver=ns_resolver)
  File "/Users/kimmeoungjun/openai-agents-python/env/lib/python3.9/site-packages/pydantic/_internal/_model_construction.py", line 544, in set_model_fields
[...]

PR #496, opened 2025-04-14
Fix type annotations examples

DanieleMorotti:
I fixed the type annotations errors for pydantic objects in some examples as noted in #490 .

When running `make lint` the following error occurs "UP007 Use `X | Y` for type annotations". If you know how to ignore it I will be happy to edit the pull request, thanks.

### DIFF
--- a/examples/agent_patterns/output_guardrails.py
+++ b/examples/agent_patterns/output_guardrails.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 import json
-from typing import Union
 
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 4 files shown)


rm-openai (changes requested):
Actually you can just do `from __future__ import annotations` and these type errors will go away on 3.9

DanieleMorotti:
Yes, I thought that was the solution, but it doesn't work when using `|` within a pydantic `BaseModel` class, maybe due to how Pydantic works.

For example, I got the following error:
```
TypeError: Unable to evaluate type annotation 'str | None'. If you are making use of the new typing syntax (unions using `|` since Python 3.10 or builtins subscripting since Python 3.9), you should either replace the use of new syntax with the existing `typing` constructs or install the `eval_type_backport` package.
```

Let me know if I'm wrong and I have to change the pull request.

rm-openai (approved):

rm-openai:
cool - can you fix lint please?

rm-openai:
oh wait lint is complaining about the exact thing i talked about hmm

rm-openai:
I think you can do what it says and add https://pypi.org/project/eval-type-backport/, based on https://github.com/pydantic/pydantic/issues/7873

DanieleMorotti:
I added the library to the dependencies as suggested, it should work now, tests passed

Issue #487, opened 2025-04-13
Add the possibility to add extra body fields in the RunConfig or Agents


lucas-trueffles:
The standard openai package it is possible to add additional fields to the request body. This is particularly useful if you have some kind of proxy which can process additional data for metrics, tracing or FinOps.

Here an example how to do it in the openai package:
````python
import openai
client = openai.OpenAI()

response = client.chat.completions.create(model="gpt-4o", messages = [
    {
        "role": "user",
        "content": "Write a short poem about GitHub"
    }
], 
# it is possible to send additional body fields with "extra_body"
    extra_body={
      "field1": "value 1"
    }
) 
````

Unfortunately I wasn't able to find a way to mimic this behaviour.

I would suggest to add the feature to provide extra body fields in the RunConfig:
`result = Runner.run_sync(agent, input="Say this is a test", run_config=RunConfig(extra_body={"field1":"value 1"}))`

Maybe it would also benefit if you can set extra body fields on a per agent basis. 
`agent = Agent(name="Assistant", instructions="You are a helpful assistant", extra_body={"field1":"value 1"})`
These will then be added on top of the extra fields which have been set in the RunConfig.

rm-openai:
Makes sense. I think it belongs on ModelSettings; PR welcome!

DanieleMorotti:
I submitted a PR #500 , also adding `extra_query` parameter. Tell me if this implementation is okay for you.

rm-openai:
Thanks @DanieleMorotti for fixing this!

PR #500, opened 2025-04-14
Add extra request parameters `extra_query` and `extra_body`

DanieleMorotti:
Added the possibility to pass `extra_query` and `extra_body` parameters when sending a request.
In this implementation I added the attributes to `ModelSettings` as suggested by @rm-openai in #487 .

I'll be happy to add some tests if you have any suggestions.


rm-openai (approved):

rm-openai:
thank you!

Issue #495, opened 2025-04-14
Output Guardrails while Streaming


adhishthite:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have had similar requests

### Question
I have exposed an Agent as a FastAPI endpoint:

```python
@router.post("/v1/agent/stream")
async def query(request: Request, payload: QueryInput):
    """
    Stream a query response from thr RAG Agent using Server-Sent Events.

    Args:
        request (Request): The incoming FastAPI request object
        payload (QueryInput): The query input containing the user's question and chat ID

    Returns:
        StreamingResponse: Server-sent events stream containing the agent's response

    Raises:
        HTTPException: 500 error if there's an issue processing the request
    """
    try:
        # Create streaming response using SSE
        return StreamingResponse(
            stream_sse_response(payload, request),
            media_type="text/event-stream",  # Set media type for SSE
        )
    except Exception as e:
        # Handle any errors during processing
        raise HTTPException(status_code=500, detail=str(e))
```

Here's my Agent

```python
rag_agent = Agent(
    name="RAG Agent",
    instructions=AGENT_PROMPT,  # System prompt defining agent behavior
    tools=[search_knowledgebase, get_current_datetime],  # Available tools for the agent
    model_settings=ModelSettings(
        temperature=0,  # Use deterministic responses
        parallel_tool_calls=True,  # Allow concurrent tool execution
        max_tokens=4096,  # Maximum response length
    ),
    input_guardrails=[agent_input_guardrail],
    output_guardrails=[agent_output_guardrail],
)
async def stream_sse_response(payload: QueryInput, request):
    """
    Streams agent responses using Server-Sent Events (SSE) format.

    Handles both standalone queries and conversations with chat history.
    Monitors client connection and gracefully handles disconnections.

    Args:
        payload (QueryInput): Contains the user query and optional chat history
        request: FastAPI request object for connection monitoring

    Yields:
        SSE-formatted text chunks from the agent's response

    Raises:
        Exception: Logs and yields any errors during streaming
    """

    with trace("RAG API Stream"):
        try:
            # Process chat history if available
            if payload.chat_history and len(payload.chat_history) > 0:
                # Convert chat messages to dict format
                formatted_history = [msg.model_dump() for msg in payload.chat_history]
                # Append current query to history
                input_data = formatted_history + [
                    {"role": "user", "content": payload.query}
                ]
[...]

PR #505, opened 2025-04-14
Example for streaming guardrails

rm-openai:
An example for the question in the issue attached - how to run guardrails during streaming.

Towards #495.

### DIFF
--- a/examples/agent_patterns/streaming_guardrails.py
+++ b/examples/agent_patterns/streaming_guardrails.py
@@ -1,0 +1,93 @@
+from __future__ import annotations
+
+import asyncio
+
+from openai.types.responses import ResponseTextDeltaEvent
+from pydantic import BaseModel, Field
+
+from agents import Agent, Runner
+
+"""
+This example shows how to use guardrails as the model is streaming. Output guardrails run after the
+final output has been generated; this example runs guardails every N tokens, allowing for early
+termination if bad output is detected.
+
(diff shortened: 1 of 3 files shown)


examples/agent_patterns/streaming_guardrails.py
```diff
@@ -0,0 +1,87 @@
...
+    async for event in result.stream_events():
+        if event.type == "raw_response_event" and isinstance(event.data, ResponseTextDeltaEvent):
+            print(event.data.delta, end="", flush=True)
+            current_text += event.data.delta
+
+            # Check if it's time to run the guardrail check
+            # Note that we don't run the guardrail check if there's already a task running. An
+            # alternate implementation is to have N guardrails running, or cancel the previous
+            # one.
+            if len(current_text) >= next_guardrail_check_len and not guardrail_task:
```
pakrym-oai, line 72: If the generation model is much faster than the guardrail and the guardrail doesn't trigger on the first few chunk we won't run the entire response through the guardrail not firing where it should.

pakrym-oai (commented):
This guardrail implementation seems not fully reliable.

examples/agent_patterns/streaming_guardrails.py
rm-openai: fair, added a final check as well

pakrym-oai (approved):

Issue #497, opened 2025-04-14
How to set max_turn for agents set as tools.


yongkangzhao:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) 
YES, seems max_turn is only available to be set at Runner. but with the implementation of agents as tools, I don't get access to the Runner of the tool agents.
- **Have you searched for related issues?** Others may have had similar requests 
YES couldn't find any

### Question
Describe your question. Provide details if available.
How to set max_turn when agents are being used as tools.
I have serveral agents as tool use case. and the main agent can set Runner.run(max_turn), but I couldn't figure out where so inject max_turn limits for agents created as tools.

exactly like this example here: https://github.com/openai/openai-agents-python/blob/main/examples/agent_patterns/agents_as_tools.py

if it is not already implemented, I would suggest having a parameters for Agent, so that each agent can have a different max_turn limits, while Runner.run max_turn can override it.

thanks!

rm-openai:
Just added some docs for this here: https://github.com/openai/openai-agents-python/pull/504

yongkangzhao:
I see, but does this also applies for MCP tools? Since I'm also using tools obtained from MCP servers, I don't get to modify the tool implementations.

rm-openai:
Well MCP tools don't have a max_turns param - what are you trying to achieve with MCP tools?

yongkangzhao:
For example,

Browser control MCP tool/agents often runs more than 10 turns. 

Each time the main agent gives another agent as tool a small task to finish on its own， and just report back the results. That way my main agent only sees the result and so my total context is reduced a bit, and so the agent can focus on accomplishing the big task without derailing its attention during the process. 


https://github.com/openai/openai-agents-python/blob/e8f14da473775ca6a5f3d5208e21759bdeaf5040/src/agents/run.py#L47

I see that the default max turn is hard coded here, can we modify it so that this can be override with some kind, environment variables perhaps?

rm-openai:
@yongkangzhao I think I'm not totally following.
- If the agent is being used in code *you* wrote, then you can pass `max_turns=99999` (or whatever else you like)
- If the agent is being used in an MCP server you wrote, same as the first point, you can pass `max_turns=whatever`
- If the agent is written by someone else inside their MCP server, then you'll need to ask the person who wrote that MCP server to set max_turns.

[...]

PR #510, opened 2025-04-14
added a logic to check environment variable to override DEFAULT_MAX_TURNS

yongkangzhao:
Extra flexibility on DEFAULT_MAX_TURNS without breaking backward compatibility.
related to #497


rm-openai:
no longer needed, see issue discussion

yongkangzhao:
Still could be nice to have?

Issue #515, opened 2025-04-15
RunResultStreaming.stream_events() raises ValueError when reset_current=True in async environments


JilinJL:
### Please read this first

- ✅ **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-python/)
- ✅ **Have you searched for related issues?** Yes, no existing issue was found.

---

### Describe the bug

When using the `RunResultStreaming.stream_events()` method within a streaming API endpoint in frameworks like FastAPI (or any ASGI server), the following exception is raised **at the end of the stream**:

>  _ValueError: <Token var=<ContextVar name='current_trace'> at 0x...> was created in a different Context_

This error originates from `self._trace.finish(reset_current=True)` being executed across async task boundaries, which violates how `ContextVar` tokens must be reset in the same context they were set.

---
### Debug information

- **Agents SDK version**: `v0.0.3`
- **Python version**: `3.12.2`
- **Framework**: `FastAPI + Uvicorn`
---
My codes:

```python
# the agent_func code
      if stream:
          return Runner.run_streamed(
              starting_agent=agent,
              input=input_items,
          ).stream_events()
```
``` python
# Here is the server layer:
@doc_agent_router.post("/agent/run")
async def run_agent(request: AgentRequest):
    try:
        agent_func = agent_map.get(request.mode)
        if not agent_func:
            raise ValueError("Invalid mode.")
            agent_kwargs = {
                "query": request.query,
                "original_content": request.content,
                "stream": request.stream
            }

        print("\n params :\n", agent_kwargs)
        result = await agent_func(**agent_kwargs)

        if request.stream:
            async def event_stream() -> AsyncGenerator[str, None]:
                async for chunk in result:
                    yield sse_event(chunk)
                yield sse_done()
            
            return StreamingResponse(
                event_stream(), 
                media_type="text/event-stream"
            )
        else:
            return BaseResponse(code=200, data=result)
    except Exception as e:
        raise HTTPException(status_code=500, detail=f"SLevel Internal server error: {str(e)}")
```

```python
# stream_events
# .venv/lib/python3.12/site-packages/agents/result.py
@dataclass
class RunResultStreaming(RunResultBase):
    """The result of an agent run in streaming mode. You can use the `stream_events` method to
    receive semantic events as they are generated.

    The streaming method will raise:
    - A MaxTurnsExceeded exception if the agent exceeds the max_turns limit.
    - A GuardrailTripwireTriggered exception if a guardrail is tripped.
[...]

PR #516, opened 2025-04-15
fix: prevent ContextVar reset error in async streaming context.

JilinJL:
### Summary

This PR fixes a runtime error caused by an unexpected `reset_current` keyword argument passed to `Trace.finish()` within the `RunResultStreaming.stream_events()` method.

### What was the issue?

Currently, `stream_events()` calls:
refer #515

### Solution
Wrap the .finish(reset_current=True) call in a try/except block to fallback to a safe reset mode:
```python
if self._trace:
    try:
        self._trace.finish(reset_current=True)
    except ValueError:
        logger.warning("ContextVar reset failed — switching to reset_current=False")
        self._trace.finish(reset_current=False)
```

### 🧪 Testing
Tested with a minimal FastAPI + StreamingResponse setup. After this patch:

- The stream completes successfully.
- No ValueError is raised.
- Trace finalization still works.

### 📦 Affected file
`openai-agents-python\src\agents\result.py`

💬 Related Issue
This PR resolves the bug described in the issue: #515

### DIFF
--- a/src/agents/result.py
+++ b/src/agents/result.py
@@ -186,7 +186,12 @@ class RunResultStreaming(RunResultBase):
             self._event_queue.task_done()
 
         if self._trace:
-            self._trace.finish(reset_current=True)
+            try:
+                self._trace.finish(reset_current=True)
+            except ValueError:
+                logger.warning("ContextVar reset failed — switching to reset_current=False")
+                self._trace.finish(reset_current=False)
+
 
         self._cleanup_tasks()
 


JilinJL (commented):
here

rm-openai:
IMO the loud exception is probably a good thing? Catches obvious errors. WDYT?

JilinJL:
> IMO the loud exception is probably a good thing? Catches obvious errors. WDYT?

I realize now that I don’t understand this area as well as I thought — my knowledge is still quite shallow.

I tried to be clever and submitted a PR, thinking I could grab a quick contribution, but clearly I misunderstood how it actually works.

Thanks a lot for the correction and your patience — I’ve learned from this.

PR #519, opened 2025-04-15
Only include stream_options when streaming

rm-openai:


src/agents/models/openai_chatcompletions.py
```diff
@@ -591,8 +593,11 @@ def get_store_param(cls, client: AsyncOpenAI, model_settings: ModelSettings) ->
 
     @classmethod
     def get_stream_options_param(
-        cls, client: AsyncOpenAI, model_settings: ModelSettings
+        cls, client: AsyncOpenAI, model_settings: ModelSettings, stream: bool
```
pakrym-oai, line 596: uber nit: isn't it easier to do `_Converter.get_stream_options_param() if stream else None` in the method above?

pakrym-oai (approved):

src/agents/models/openai_chatcompletions.py
rm-openai: IMO no because this way the logic for stream_options is all in one place

PR #524, opened 2025-04-15
Litellm integration

rm-openai:
litellm is a library that abstracts away details/differences for a lot of model providers. Adding an extension, so that any provider can easily be integrated.

---
* #532
* __->__ #524

### DIFF
--- a/examples/model_providers/litellm_provider.py
+++ b/examples/model_providers/litellm_provider.py
@@ -1,0 +1,55 @@
+from __future__ import annotations
+
+import asyncio
+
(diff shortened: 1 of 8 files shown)


pakrym-oai (approved):

vmayoral:
@rm-openai csn you please clarify why did you reimplement litellm integration instead of building on top of https://github.com/openai/openai-agents-python/pull/318

are community contributions not welcome for some reason? I put down some time to contribute this back and I’d like technical clarification. Why is the currently used approach better?

PR #525, opened 2025-04-16
Show repo name/data in docs

rm-openai:
Easy linking back to the repo, plus some social proof (stars/forks etc).


seratch (approved):
I like this! Confirmed how it works

JilinJL:
good

PR #534, opened 2025-04-17
RFC: automatically use litellm if possible

rm-openai:
## Summary
This replaces the default model provider with a `MultiProvider`, which has the logic:
- if the model name starts with `openai/` or doesn't contain "/", use OpenAI
- if the model name starts with `litellm/`, use LiteLLM to use the appropriate model provider.

It's also extensible, so users can create their own mappings. I also imagine that if we natively supported Anthropic/Gemini etc, we can add it to MultiProvider to make it work.

The goal is that it should be really easy to use any model provider. Today if you pass `model="gpt-4.1"`, it works great. But `model="claude-sonnet-3.7"` doesn't. If we can make it that easy, it's a win for devx.

I'm not entirely sure if this is a good idea - is it too magical? Is the API too reliant on litellm? Comments welcome.

## Test plan

For now, the example. Will add unit tests if we agree its worth mergin.

### DIFF
--- a/src/agents/extensions/models/litellm_provider.py
+++ b/src/agents/extensions/models/litellm_provider.py
@@ -1,0 +1,21 @@
+from ...models.interface import Model, ModelProvider
+from .litellm_model import LitellmModel
+
+DEFAULT_MODEL: str = "gpt-4.1"
+
+
+class LitellmProvider(ModelProvider):
+    """A ModelProvider that uses LiteLLM to route to any model provider. You can use it via:
+    ```python
+    Runner.run(agent, input, run_config=RunConfig(model_provider=LitellmProvider()))
+    ```
+    See supported models here: [litellm models](https://docs.litellm.ai/docs/providers).
+
+    NOTE: API keys must be set via environment variables. If you're using models that require
+    additional configuration (e.g. Azure API base or version), those must also be set via the
+    environment variables that LiteLLM expects. If you have more advanced needs, we recommend
+    copy-pasting this class and making any modifications you need.
+    """
+
+    def get_model(self, model_name: str | None) -> Model:
+        return LitellmModel(model_name or DEFAULT_MODEL)
--- a/src/agents/models/multi_provider.py
+++ b/src/agents/models/multi_provider.py
@@ -54,7 +54,7 @@ class MultiProviderMap:
 class MultiProvider(ModelProvider):
     """This ModelProvider maps to a Model based on the prefix of the model name. By default, the
     mapping is:
-    - "openai/" prefix or no prefix -> OpenAIProvider. e.g. "openai/gpt-4.1"
+    - "openai/" prefix or no prefix -> OpenAIProvider. e.g. "openai/gpt-4.1", "gpt-4.1"
     - "litellm/" prefix -> LitellmProvider. e.g. "litellm/openai/gpt-4.1"
 
     You can override or customize this mapping.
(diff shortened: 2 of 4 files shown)


src/agents/extensions/models/litellm_provider.py
```diff
@@ -0,0 +1,21 @@
+from ...models.interface import Model, ModelProvider
+from .litellm_model import LitellmModel
+
+DEFAULT_MODEL: str = "gpt-4.1"
```
seratch, line 4: Would importing https://github.com/openai/openai-agents-python/blob/v0.0.11/src/agents/models/openai_provider.py#L11 instead be better? Also, huge 👍 to switching from gtp-4o to gpt-4.1

src/agents/models/multi_provider.py
```diff
@@ -0,0 +1,144 @@
...
+            return None, None
+        elif "/" in model_name:
+            prefix, model_name = model_name.split("/", 1)
+            return prefix, model_name
+        else:
+            return None, model_name
+
+    def _create_fallback_provider(self, prefix: str) -> ModelProvider:
+        if prefix == "litellm":
+            from ..extensions.models.litellm_provider import LitellmProvider
```
seratch, line 113: I like this dynamic import here! 

nice-to-have: on the LitellmProvider side, having try/except clause for loading litellm module and raising a more user-friendly error message than exposing the missing litellm could make dev experience better

seratch (commented):
I personally really like this idea!

src/agents/models/multi_provider.py
rm-openai: @seratch - LitellmProvider will try to import LitellmModel, which will display the nice error message: https://github.com/openai/openai-agents-python/blob/main/src/agents/extensions/models/litellm_model.py#L16-L19

src/agents/extensions/models/litellm_provider.py
rm-openai: Feels like a minor breaking change. Though probably fine! Good call.

src/agents/models/multi_provider.py
seratch: Nice, you’re already ahead!

src/agents/models/multi_provider.py
```diff
@@ -0,0 +1,144 @@
+from __future__ import annotations
+
+from openai import AsyncOpenAI
+
+from ..exceptions import UserError
+from .interface import Model, ModelProvider
+from .openai_provider import OpenAIProvider
+
+
+class MultiProviderMap:
```
yihuang, line 10: do we need this class to encapsulate the simple operations on plain dict?

src/agents/models/multi_provider.py
```diff
@@ -0,0 +1,144 @@
...
+        Args:
+            prefix: The prefix of the model name e.g. "openai" or "my_prefix".
+        """
+        del self._mapping[prefix]
+
+
+class MultiProvider(ModelProvider):
+    """This ModelProvider maps to a Model based on the prefix of the model name. By default, the
+    mapping is:
+    - "openai/" prefix or no prefix -> OpenAIProvider. e.g. "openai/gpt-4.1"
```
stevenheidel:
```suggestion
    - "openai/" prefix or no prefix -> OpenAIProvider. e.g. "openai/gpt-4.1", "gpt-4.1"
```

src/agents/extensions/models/litellm_provider.py
seratch: @rm-openai Perhaps you're already aware of this, but switching to gpt-4.1 might break existing CUA apps (the tool is not yet available with 4.1 while web_search_preview works with 4.1), so indeed switching the default model could be a breaking change

PR #546, opened 2025-04-18
feat: support in-process fastmcp server

yihuang:
fastmcp has an interesting feature to connect a [in-process mcp server](https://github.com/jlowin/fastmcp/blob/main/src/fastmcp/client/transports.py#L379), it's quite convenient.
It unifies function calling and mcp server in a sense.

### DIFF
--- a/examples/mcp/in_process/README.md
+++ b/examples/mcp/in_process/README.md
@@ -10,5 +10,5 @@ uv run python -m examples.mcp.in_process.main
 
 ## Details
 
-The example uses the `MCPServerSse` class from `agents.mcp`. The server runs in a sub-process at `https://localhost:8000/sse`.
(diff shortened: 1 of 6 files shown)


examples/mcp/in_process/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Run the example via:
+
+```
+uv run python -m examples.mcp.in_process.main
+```
+
+## Details
+
+The example uses the `MCPServerSse` class from `agents.mcp`. The server runs in a sub-process at `https://localhost:8000/sse`.
```
yihuang:
```suggestion
The example uses the `FastMCPServer` class from `agents.mcp.fastmcp`. The server is invoked in-process like a function call.
```

PR #552, opened 2025-04-19
Add optional max_turns parameter to Agent.as_tool

dspencej:
## Description

This PR adds an optional `max_turns` parameter to the `Agent.as_tool` method, allowing users to override the default turn limit (10) when running an agent as a tool.

## Summary of Changes

- Added `max_turns: int = DEFAULT_MAX_TURNS` as an optional parameter in `as_tool`.
- Passed `max_turns` to `Runner.run` during tool invocation.


dspencej:
Missed circular import. Will adjust with default of 10.

PR #553, opened 2025-04-20
Examples for image inputs

rm-openai:


pakrym-oai (approved):

cjescudero:
Nice and useful examples!

Do you know if there is a limit size of the image as base64? I suppose that the problem will be the context window size.

pakrym-oai:
The limit is 20MB per image https://platform.openai.com/docs/guides/images/image-generation?api-mode=responses&format=base64-encoded#image-input-requirements

ceeyangwen:
Hi, @rm-openai is there an example how agent output an image. I try serval ways, but couldn't implement it.

PR #555, opened 2025-04-20
Added cached_tokens to the usage monitoring.

Christianm9000:
**1.**: Added cached tokens to the Usage class and "add" function.
**2.**: Extracted cached_tokens from the LLM response and appended it to the Usage object.
**Note**: This change is only made when pairing the Agents SDK with the responses API (not the chat-completions).

cached_tokens is now accessible via the llm_output.raw_responses[n].usage, where n is the response index.

### DIFF
--- a/src/agents/extensions/models/litellm_model.py
+++ b/src/agents/extensions/models/litellm_model.py
@@ -106,6 +106,7 @@ class LitellmModel(Model):
                     Usage(
                         requests=1,
                         input_tokens=response_usage.prompt_tokens,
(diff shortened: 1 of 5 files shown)


rm-openai (commented):
@Christianm9000 - this looks great, but any reason you didn't add it to ChatCompletions (and also the litellm implementation)?

Christianm9000:
@rm-openai - Apologies, I was on a tight project deadline, so the pull request only reflected my needs. I have made the changes to both the ChatCompletions and LiteLLM in the new commit:)

PR #577, opened 2025-04-23
Make the TTS voices type exportable

mangiucugna:
When using the voice agent in typed code, it is suboptimal and error prone to type the TTS voice variables in your code independently.

With this commit we are making the type exportable so that developers can just use that and be future-proof.

Example of usage in code:

```
DEFAULT_TTS_VOICE: TTSModelSettings.TTSVoice = "alloy"

...

tts_voice: TTSModelSettings.TTSVoice = DEFAULT_TTS_VOICE 

...

output = await VoicePipeline(
  workflow=workflow,
  config=VoicePipelineConfig(
  tts_settings=TTSModelSettings(
    buffer_size=512,
    transform_data=transform_data,
    voice=tts_voice,
    instructions=tts_instructions,
  ))
).run(audio_input)
```

### DIFF
--- a/src/agents/voice/__init__.py
+++ b/src/agents/voice/__init__.py
@@ -31,6 +31,7 @@ __all__ = [
     "STTModelSettings",
     "TTSModel",
     "TTSModelSettings",
+    "TTSVoice",
     "VoiceModelProvider",
     "StreamedAudioResult",
     "SingleAgentVoiceWorkflow",
--- a/src/agents/voice/__init__.py
+++ b/src/agents/voice/__init__.py
@@ -7,6 +7,7 @@ from .model import (
     STTModelSettings,
     TTSModel,
     TTSModelSettings,
+    TTSVoice,
     VoiceModelProvider,
 )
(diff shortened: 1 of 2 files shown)


rohan-mehta:
@mangiucugna this looks great - can you move it to a top level type instead of inside the class please?

mangiucugna:
@rohan-mehta yes of course makes more sense 😅

Ok now with the new changes this is how it will work:


```
from agents.voice import TTSVoice
DEFAULT_TTS_VOICE: TTSVoice = "alloy"

...

tts_voice: TTSVoice = DEFAULT_TTS_VOICE 

...

output = await VoicePipeline(
  workflow=workflow,
  config=VoicePipelineConfig(
  tts_settings=TTSModelSettings(
    buffer_size=512,
    transform_data=transform_data,
    voice=tts_voice,
    instructions=tts_instructions,
  ))
).run(audio_input)
```

I have renamed the type to TTSVoice because is more inline with the rest of types in the library (and python in general)

rm-openai (approved):

Issue #574, opened 2025-04-23
Canceling the stream from result.stream_events()


handrew:
Below is the example from the docs. I'm trying to `break` out of the `result.stream_events()` async for. I can break out, but the agent will continue to take actions. It would be great if there was some way for us to stop the events entirely, because the agent will continue wreaking havoc on my things. 

```python
import asyncio
from openai.types.responses import ResponseTextDeltaEvent
from agents import Agent, Runner

async def main():
    agent = Agent(
        name="Joker",
        instructions="You are a helpful assistant.",
    )

    result = Runner.run_streamed(agent, input="Please tell me 5 jokes.")
    async for event in result.stream_events():
        if event.type == "raw_response_event" and isinstance(event.data, ResponseTextDeltaEvent):
            print(event.data.delta, end="", flush=True)


if __name__ == "__main__":
    asyncio.run(main())
```

rm-openai:
Yeah makes sense, we should add this

handrew:
@rm-openai Any way I can help?

rm-openai:
@handrew PR would be great! Should be able to add `cancel()` to `RunResultStreaming` and just cancel all the `asyncio.task`s. Some tests would be great too

handrew:
@rm-openai Alrighty I gave it a shot. I'm a bit of a noob to OSS contribution so please bear with me. https://github.com/openai/openai-agents-python/pull/579

handrew:
Thanks @rm-openai !

PR #579, opened 2025-04-23
Allow cancel out of the streaming result

handrew:
Fix for #574 

@rm-openai  I'm not sure how to add a test within the repo but I have pasted a test script below that seems to work 

```python
import asyncio
from openai.types.responses import ResponseTextDeltaEvent

async def main():
    agent = Agent(
        name="Joker",
    )

    result = Runner.run_streamed(agent, input="Please tell me 5 jokes.")
    num_visible_event = 0
        if event.type == "raw_response_event" and isinstance(event.data, ResponseTextDeltaEvent):
            num_visible_event += 1
            print(num_visible_event)
            if num_visible_event == 3:
                result.cancel()


if __name__ == "__main__":
    asyncio.run(main())
````

### DIFF
--- a/src/agents/result.py
+++ b/src/agents/result.py
@@ -152,6 +152,17 @@ class RunResultStreaming(RunResultBase):
         """
         return self.current_agent
 
+    def cancel(self) -> None:
+        """Cancels the streaming run, stopping all background tasks and marking the run as complete."""
+        self._cleanup_tasks()  # Cancel all running tasks
+        self.is_complete = True  # Mark the run as complete to stop event streaming
+
+        # Optionally, clear the event queue to prevent processing stale events
+        while not self._event_queue.empty():
+            self._event_queue.get_nowait()
+        while not self._input_guardrail_queue.empty():
+            self._input_guardrail_queue.get_nowait()
+
     async def stream_events(self) -> AsyncIterator[StreamEvent]:
         """Stream deltas for new items as they are generated. We're using the types from the
         OpenAI Responses API, so these are semantic events: each event has a `type` field that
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/agents/result.py [hunk 1]
```diff
@@ -174,6 +185,7 @@ async def stream_events(self) -> AsyncIterator[StreamEvent]:
             try:
                 item = await self._event_queue.get()
             except asyncio.CancelledError:
+                self.cancel()  # Ensure tasks are cleaned up if the coroutine is cancelled
```
rm-openai: is this necessary? cleanup still occurs outside of the loop

rm-openai (commented):
@handrew looks good! You can add unit tests to the `tests/` directory. If you use codex/cursor/windsurf/chatgpt, they can suggest tests and probably even help you write them. You can run the tests locally via `make tests`

src/agents/result.py
handrew: I suppose not, though the AI thought it was 😅

handrew:
Thanks for the tip @rm-openai , I'll be sure to do that in the future, pretty neat. Anyway, added the test!

rm-openai:
@handrew Test looks good, but it doesn't look like the test is passing - seems like its an infinite loop. You may need to use `FakeModel`, since there's no network access in unit tests.

 Also lint is failing - could you fix plz?

handrew:
> @handrew Test looks good, but it doesn't look like the test is passing - seems like its an infinite loop. You may need to use `FakeModel`, since there's no network access in unit tests.
> 
> Also lint is failing - could you fix plz?

Used `FakeModel`, good call. It only yields two events so I test cancelling after one. Also linted.

Thanks @rm-openai , this was fun. Any idea when I might be able to `--upgrade` and get this fix :)?

src/agents/result.py
handrew: Removed this

rm-openai:
@handrew I'm merging the PR now. Will probably cut a new release tomorrow morning, so should be able to upgrade then. 

You can also temporarily directly install from the latest code in this repo if you want, which will include your change
```
pip uninstall openai-agents
pip install git+https://github.com/openai/openai-agents-python.git
```

rm-openai (approved):

PR #580, opened 2025-04-23
Prevent MCP ClientSession hang

njbrake:
Per https://modelcontextprotocol.io/specification/draft/basic/lifecycle#timeouts

"Implementations SHOULD establish timeouts for all sent requests, to prevent hung connections and resource exhaustion. When the request has not received a success or error response within the timeout period, the sender SHOULD issue a cancellation notification for that request and stop waiting for a response.

SDKs and other middleware SHOULD allow these timeouts to be configured on a per-request basis."

I picked 5 seconds since that's the default for SSE

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import abc
 import asyncio
 from contextlib import AbstractAsyncContextManager, AsyncExitStack
+from datetime import timedelta
 from pathlib import Path
 from typing import Any, Literal
 
@@ -69,6 +70,9 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
         self._cleanup_lock: asyncio.Lock = asyncio.Lock()
         self.cache_tools_list = cache_tools_list
 
+        self.timeout: float = 5
+        """The timeout for the MCP ClientSession. Defaults to 5 seconds."""
+
         # The cache is always dirty at startup, so that we fetch tools at least once
         self._cache_dirty = True
         self._tools_list: list[MCPTool] | None = None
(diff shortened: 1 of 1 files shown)


rm-openai:
@njbrake - thanks. Couple of quick questions:
1. This is different than the timeout on the SSE server? Any idea why there are two different places for timeout configuration?
2. Based on the spec - should we make this configurable? Just add `client_session_timeout_seconds: int | None = 5` and use that?

njbrake:
Thanks for the quick response @rm-openai. 

> @njbrake - thanks. Couple of quick questions:
> 
>     1. This is different than the timeout on the SSE server? Any idea why there are two different places for timeout configuration?

Based on my understanding of the MCP python sdk, yep this is different because this ClientSession applies to both SSE and Stdio sessions. The current `timeout` and `sse_read_timeout` parameters only apply to the `sse_client` so they're not involved with a Stdio client, resulting in a ♾️ hang if the Stdio server dies.

The parameter that I'm adding to mcp.ClientSession is passed to https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/shared/session.py#L246-L263 
which looks like it's the ClientSession equivalent HTTP timeout just like https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/sse.py#L25

> 
>     2. Based on the spec - should we make this configurable? Just add `client_session_timeout_seconds: int | None = 5` and use that?

Sounds good to me 👍

njbrake:
@rm-openai  Ok gave a shot at implementing the timeout as you suggested, let me know what you think 🙏

rm-openai (approved):

njbrake:
@rm-openai  ah sorry, I see some linting etc failing, one sec I'll fix it up

PR #582, opened 2025-04-24
Create to_json_dict for ModelSettings

rm-openai:
Now that `ModelSettings` has `Reasoning`, a non-primitive object, `dataclasses.as_dict()` wont work. It will raise an error when you try to serialize (e.g. for tracing). This ensures the object is actually serializable.


pakrym-oai (approved):
the mix of data classes and pydantic makes me sad

rm-openai:
yeah :(

Issue #587, opened 2025-04-24
LiteLLM extension crashes with run_streamed


theimpostor:
### Describe the question
Running the LiteLLM sample from the [doc](https://openai.github.io/openai-agents-python/models/litellm/) works. However modifying the sample to use the run_streamed API to log responses, tool calls etc results in an exception.

### Debug information
- Agents SDK version: v0.0.12
- Python version: 3.13

### Repro steps
```python
#!/usr/bin/env -S uv run --script

# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "openai-agents[litellm]",
# ]
# ///

from __future__ import annotations

import asyncio

from agents import Agent, ItemHelpers, Runner, function_tool, set_tracing_disabled
from agents.extensions.models.litellm_model import LitellmModel

set_tracing_disabled(disabled=True)

@function_tool
def get_weather(city: str):
    print(f"[debug] getting weather for {city}")
    return f"The weather in {city} is sunny."


async def main(model: str, api_key: str):
    agent = Agent(
        name="Assistant",
        instructions="You only respond in haikus.",
        model=LitellmModel(model=model, api_key=api_key),
        tools=[get_weather],
    )

    result = Runner.run_streamed(agent, "What's the weather in Tokyo?")
    async for event in result.stream_events():
        # We'll ignore the raw responses event deltas
        if event.type == "raw_response_event":
            continue
        # When the agent updates, print that
        elif event.type == "agent_updated_stream_event":
            print(f"Agent updated: {event.new_agent.name}")
            continue
        # When items are generated, print them
        elif event.type == "run_item_stream_event":
            if event.item.type == "tool_call_item":
                print(f"-- Tool was called: {event.item.raw_item}")
            elif event.item.type == "tool_call_output_item":
                print(f"-- Tool output: {event.item.output}")
            elif event.item.type == "message_output_item":
                print(f"-- Message output:\n {ItemHelpers.text_message_output(event.item)}")
            else:
                pass  # Ignore other event types

    print(result.final_output)


if __name__ == "__main__":
    # First try to get model/api key from args
    import argparse

    parser = argparse.ArgumentParser()
    parser.add_argument("--model", type=str, required=False)
    parser.add_argument("--api-key", type=str, required=False)
    args = parser.parse_args()

    model = args.model
    if not model:
        model = input("Enter a model name for Litellm: ")

    api_key = args.api_key
    if not api_key:
        api_key = input("Enter an API key for Litellm: ")

[...]

PR #589, opened 2025-04-24
Fix stream error using LiteLLM

DanieleMorotti:
In response to issue #587 , I implemented a solution to first check if `refusal` and `usage` attributes exist in the `delta` object.

I added a unit test similar to `test_openai_chatcompletions_stream.py`.

Let me know if I should change something.


rm-openai (approved):
awesome thank you!

rm-openai:
Ah @DanieleMorotti you'll need to move the test into `tests/models/`

devv-shayan:
was working on :) but thanks @DanieleMorotti

PR #591, opened 2025-04-24
Allow previous_response_id to be passed to input guardrails

rm-openai:
Input guardrails are expected to run on the full input, so passing the `previous_response_id` along. For back-compat, introducing a new signature for the guardrail. Using a dataclass for the inputs so we can extend in the future easily.

### DIFF
--- a/src/agents/__init__.py
+++ b/src/agents/__init__.py
@@ -18,7 +18,10 @@ from .exceptions import (
 )
 from .guardrail import (
     GuardrailFunctionOutput,
+    InputGuardailInputs,
     InputGuardrail,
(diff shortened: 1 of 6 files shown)


pakrym-oai:
Did you consider adding a field to RunContextWrapper instead?

rm-openai:
@pakrym-oai yes. Decided not to, because RunContextWrapper is the same across multiple LLM calls in a single run (and hence the previous response ID changes)

rm-openai:
After offline discussion: `previous_response_id` is spreading all over, and we should do this better. Workaround for end-users for now, is to add the ID to the context and use it in guardrails.

Issue #630, opened 2025-04-30
no attribute error occurs while calling MCP


Czhang0727:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- Yes
- **Have you searched for related issues?** Others may have faced similar issues.
Yes

### Describe the bug

When we are using local model ( ollama for example ) the result not necessary contains content here, in my case, it was result[0].text

I added a local patch 
```python
  # The MCP tool result is a list of content items, whereas OpenAI tool outputs are a single
  # string. We'll try to convert.
  logger.info(f"Tool result content: ", result[0].text)
  if len(result.content) == 1:
      tool_output = result.content[0].model_dump_json()
  elif len(result.content) > 1:
```
```bash
 logger.info(f"Tool result content: ", result[0].text)
Message: 'Tool result content: '
Arguments: ('{"content": "{\\"ids\\": [[\\"caf665b7-cb2e-428d-9139-ff6da8687731\\", \\"f3d34cf7-d263-4dd4-b84f-b63f4258b43c\\", \\"4d649cfa-2c79-497b-8e27-bae7c8925311\\", \\"f2677a99-99ff-49ad-be52-b1e078a9b56a\\", \\"b463d689-bebb-4752-9e4f-f42395f977db\\"]], \\"embeddings\\": null, \\"documents\\": [[\\"The movie begin
```

  File "/opt/homebrew/Caskroom/miniconda/base/lib/python3.11/site-packages/agents/mcp/util.py", line 116, in invoke_mcp_tool
    if len(result.content) == 1:
           ^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'content'

### Debug information
- Agents SDK version: (v0.0.13`)
- Python version (Python 3.11.5)

### Repro steps

Ideally provide a minimal python script that can be run to reproduce the bug.
Please find my repo here https://github.com/Czhang0727/object-rag
You can reproduce by running server/agents_components/evaluation/run_router_demo.py

### Expected behavior
We should not fail the call, the result should deliver to final output

Czhang0727:
A simple solution could be to assign text to content. 
Or, we can do MCP version check ( or maybe framework check ) for better data aggregation

Czhang0727:
Dummy proposed solution
```
        # The MCP tool result is a list of content items, whereas OpenAI tool outputs are a single
        # string. We'll try to convert.
[...]

PR #631, opened 2025-04-30
[MCP][Utils] Add support for FastMCP processing

Czhang0727:
Fixing issue #630 

Problem:
In FastMCP, the structure did not have array data inside of result, instead, the result itself is the array itself. And the text field is used.

Add a type check, if we found it an array, simply merge all results with the content. Please find the fixed test project here: https://github.com/Czhang0727/object-rag

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -113,13 +113,16 @@ class MCPUtil:
 
         # The MCP tool result is a list of content items, whereas OpenAI tool outputs are a single
         # string. We'll try to convert.
-        if len(result.content) == 1:
-            tool_output = result.content[0].model_dump_json()
-        elif len(result.content) > 1:
-            tool_output = json.dumps([item.model_dump() for item in result.content])
+        if type(result) == list:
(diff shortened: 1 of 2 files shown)


Czhang0727:
seems no one is reviewing this change, plz let me know if there is any updates

rm-openai:
Sorry was out on vacation! Will take a closer look soon.

rm-openai:
Not totally understanding the issue here. When would the return type not have `content`?

Czhang0727:
In the current code, we are expecting

result = { 
    content: [content_object]
}


in fastMCP, I got a response like

result = [content_object]

rm-openai:
@Czhang0727 The spec is pretty clear that the response must be `result: { content: [...] } }` (https://modelcontextprotocol.io/specification/2025-03-26/server/tools#calling-tools).

So is it possible your server is not compliant?

Czhang0727:
hmm, in that case, fastMCP may have some issue, let me check the behavior and get back to the thread

PR #639, opened 2025-05-02
fix: add ensure_ascii=False to json.dumps for correct Unicode output

KatHaruto:
Set `ensure_ascii=False` in `json.dumps(...)` to allow proper Unicode character output (e.g., Japanese).

Without this, non-ASCII characters are escaped (e.g., 日本語 → \u65e5\u672c\u8a9e), which makes logs or exports harder to read.


seratch:
Thanks for the suggestion! You’re absolutely right. Non-English text in debug logs isn’t very developer-friendly at the moment. Generally speaking, as long as all input text data is UTF-8 encoded, this shouldn’t introduce any issues. In particular, the changes to the debug logging should be safe. 

That said, the changes in the following files could potentially affect behavior (though I think it’s unlikely). Still, it would be good to test with a few patterns just to be sure:
- src/agents/models/chatcmpl_converter.py
- src/agents/mcp/util.py

Also, the CI tests detect an issue with older Python versions. Could you resolve it?

KatHaruto:
@seratch 
Thank you for the review

Regarding the following two files:

- src/agents/models/chatcmpl_converter.py
- src/agents/mcp/util.py

I realized afterward that the changes I made to them were unrelated to the main goal of this PR, which is to improve debug logs. I’ve removed those changes.

I’ve also resolved the CI test failure on older Python versions.

KatHaruto:
@rm-openai 
Would you mind reviewing this PR？

seratch (approved):
LGTM

Issue #159, opened 2025-03-14
How to utilize vision of LLM?


mocheng:
## Question

How to utilize vision capability of LLM with OpenAI Agent SDK?

The API should support specifying image URL or local image path or base64 string. The documentation seems lacking this feature.

Code example of "Describe uploaded image" would be appreciated.

DanieleMorotti:
Hi, I tested the following code and it works both with a url and base64 images. As always, you can use OpenAI code to encode an image to base64:
```python
import base64

# Function to encode the image
def encode_image(image_path):
    with open(image_path, "rb") as image_file:
        return base64.b64encode(image_file.read()).decode("utf-8")

# Path to your image
image_path = "image.jpeg"

# Getting the Base64 string
base64_image = encode_image(image_path)
```

And then you can pass the base64 image or a url:
```python
from agents import Agent, Runner, ModelSettings

agent = Agent(
    name="Assistant",
    model="gpt-4o-mini",
    model_settings=ModelSettings(temperature=0.4, max_tokens=1024),
    instructions="Given an input image you will generate the description of the image in the style specified by the user."
)

result = await Runner.run(agent, input=[
    {
        "role": "user",
        "content": [
            {"type": "input_text", "text": "Describe this image with an haiku."},
            {
                "type": "input_image",
                "image_url": f"data:image/jpeg;base64,{base64_image}", # or your url "https://..."
            },
        ]
    }
])
print(result.final_output)
```

I don't know if this is the best available method, but I hope this may be useful.

mocheng:
@DanieleMorotti  It works. Thanks!

xin-ran-w:
I still have a question about whether an MCP server can return an image as the result. I want to let the gpt-4o see how I can do this.

DanieleMorotti:
I first tried to adopt `tool_use_behavior={"stop_at_tool_name": [...]}` but it didn't work for me.

The only idea that comes to my mind is to pass a function as `tool_use_behaviour`, where you select the output of the tool and return it as a final output.
```python
from agents import Agent, Runner, ToolsToFinalOutputResult

def stop_at_tool(context, tools_resp):
    res = tools_resp[0].output
    return ToolsToFinalOutputResult(is_final_output=True, final_output=res)

mn_agent = Agent(
        name="Image descriptor agent",
        model="gpt-4o",
        instructions="You have to describe the image requested by the user",
        model_settings=ModelSettings(temperature=0.3, max_tokens=2048),
[...]

PR #685, opened 2025-05-12
Fixed a bug for "detail" attribute in input image

DanieleMorotti:
When an input image is given as input, the code tries to access the 'detail' key, that may not be present as noted in #159.

With this pull request, now it tries to access the key, otherwise set the value to `None`.
@pakrym-oai  or @rm-openai let me know if you want any changes.


s4093120:
Hopefully it get's merged quickly. I've tested it out as well and it fixed my error and it works perfectly.

rm-openai (approved):

Issue #704, opened 2025-05-15
MCP Server instructions are currently ignored


bocytko:
### Please read this first

- [x] **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- [x] **Have you searched for related issues?** Others may have had similar requests

### Question
MCP Servers [define instructions](https://github.com/modelcontextprotocol/python-sdk/blob/v1.7.x/src/mcp/server/lowlevel/server.py#L136), for example the [aws-documentation-mcp-server](https://github.com/awslabs/mcp/blob/b61e292c24343fe577d6bbdbf07eb059820642c2/src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server.py#L40-L61).

Instructions are currently ignored in the Agent execution flow. It only reads mcp tools and converts these to LLM tool descriptions. Is this an explicit design decision?

rm-openai:
Not explicit, I didn't realize this existed. From the docs, looks like instructions is optional and it says 
> For example, this information MAY be added to the system prompt.

We could add a way for you to fetch the instructions, and then you can choose to append that to the system prompt. Open to other ideas too.

bocytko:
> We could add a way for you to fetch the instructions, and then you can choose to append that to the system prompt.

This sounds intuitive enough to me. Shall I file a separate feature request?

rm-openai:
Nope, we can use this issue. PR also welcome!

DanieleMorotti:
Hi @bocytko, if you haven’t started on this yet I’d be happy to implement this feature.
I found that the `instructions` are included in the `instructions` field of the session‐initialization response. We could store it on the server and inject them into the prompt where needed.

bocytko:
@DanieleMorotti please go ahead. Thank you in advance!

DanieleMorotti:
@rm-openai I can do something like this:
```python
server_result = await session.initialize()
self.mcp_instructions = server_result.instructions or None
```

in `src/agents/mcp/server.py` at line 125. But how / where to pass this variable only for the tools related to a specific server? In the `FunctionTool` object?

rm-openai:
@DanieleMorotti wdym? I don't think you need to do anything else. Once you do that, users will be able to do this:
```
async with MCPServerStreamableHttp(...) as server:
  agent = Agent(..., mcp_servers=[server], instructions=f"...MCP instructions: {server.instructions}")
```

DanieleMorotti:
[...]

PR #706, opened 2025-05-16
Added mcp 'instructions' attribute to the server

DanieleMorotti:
Added the `instructions` attribute to the MCP servers to solve #704 .

Let me know if you want to add an example to the documentation.

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -124,7 +124,7 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
                 )
             )
             server_result = await session.initialize()
-            self.instructions = server_result.instructions if hasattr(server_result, "instructions") else None
+            self.instructions = getattr(server_result, "instructions", None)
             self.session = session
         except Exception as e:
             logger.error(f"Error initializing MCP server: {e}")
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -124,7 +124,7 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
                 )
             )
             server_result = await session.initialize()
-            self.instructions = server_result.instructions or None
+            self.instructions = server_result.instructions if hasattr(server_result, "instructions") else None
             self.session = session
         except Exception as e:
             logger.error(f"Error initializing MCP server: {e}")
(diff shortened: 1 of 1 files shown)


src/agents/mcp/server.py
```diff
@@ -122,7 +123,8 @@ async def connect(self):
                     else None,
                 )
             )
-            await session.initialize()
+            server_result = await session.initialize()
+            self.instructions = getattr(server_result, "instructions", None)
```
rm-openai: why use getattr? I see `initialize()` returns `types.InitializeResult`, which has `instructions: str | None = None`

In fact, we should probably just set `self.server_initialize_result`. That way devs can read other info too, e.g. protocol version or capabilities

rm-openai (changes requested):

src/agents/mcp/server.py
DanieleMorotti: As you prefer, the issue was about `instructions` so I read that value, but now I set the variable as you suggested.

rm-openai (approved):
awesome thank you

cruzanstx:
is this in the released version?  

uv pip freeze | grep openai
openai==1.84.0
openai-agents==0.0.16

```
(Pdb) dir(mcp_server)
['__abstractmethods__', '__aenter__', '__aexit__', '__annotations__', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__slots__', '__str__', '__subclasshook__', '__weakref__', '_abc_impl', '_cache_dirty', '_cleanup_lock', '_name', '_tools_list', 'cache_tools_list', 'call_tool', 'cleanup', 'client_session_timeout_seconds', 'connect', 'create_streams', 'exit_stack', 'invalidate_tools_cache', 'list_tools', 'name', 'params', 'session']
(Pdb) mcp_server.instructions
*** AttributeError: 'MCPServerSse' object has no attribute 'instructions'
```

rm-openai:
@cruzanstx i just released a new version v0.0.17

cruzanstx:
@rm-openai thanks much for your hard work sir!

PR #720, opened 2025-05-20
Make Runner an abstract base class

pakrym-oai:
Allows replacing the runner and trace provider logic.


src/agents/run.py
```diff
@@ -107,7 +125,48 @@ class RunConfig:
     """
 
 
-class Runner:
+class Runner(abc.ABC):
+    @abc.abstractmethod
+    async def _run_impl(
```
rm-openai: if these are meant to be overriden, we should remove the _ and document them

src/agents/run.py
```diff
@@ -107,6 +109,54 @@ class RunConfig:
...
+    """The maximum number of turns to run for."""
+
+    hooks: RunHooks[TContext] | None
+    """Lifecycle hooks for the run."""
+
+    run_config: RunConfig | None
+    """Run configuration."""
+
+    previous_response_id: str | None
+    """The ID of the previous response, if any."""
```
rm-openai: might wanna add `NotRequired[...]` to each? Otherwise you'll be forced to set each to None manually right?

src/agents/run.py
```diff
@@ -107,6 +109,54 @@ class RunConfig:
...
+
+
+class AgentRunner(abc.ABC):
```
rm-openai: Can you add a comment saying this is not part of public interface, and that you should not use/subclass it? (We'll change that in the future once we feel better about it)

src/agents/run.py
```diff
@@ -145,11 +192,134 @@ async def run(
...
+            input,
+            context=context,
+            max_turns=max_turns,
+            hooks=hooks,
+            run_config=run_config,
+            previous_response_id=previous_response_id,
+        )
+
+
+class DefaultAgentRunner(AgentRunner):
```
rm-openai: what's the point of both DefaultAgentRunner and AgentRunner? COuld just merge them?

src/agents/tracing/provider.py
```diff
@@ -0,0 +1,294 @@
+from __future__ import annotations
```
rm-openai, line 1: if possible can you try and mark as a "move" so blame is preserved?

rm-openai (approved):

src/agents/run.py
pakrym-oai: merged

src/agents/run.py
pakrym-oai: Added

src/agents/run.py
pakrym-oai: not there anymore

PR #726, opened 2025-05-20
DRAFT: Dev/add usage details to Usage class

WJPBProjects:
PR to enhance the `Usage` object and related logic, to support more granular token accounting, matching the details available in the [OpenAI Responses API](https://platform.openai.com/docs/api-reference/responses) . Specifically, it:

- Adds `input_tokens_details` and `output_tokens_details` fields to the `Usage` dataclass, storing detailed token breakdowns (e.g., `cached_tokens`, `reasoning_tokens`).
- Flows this change through
- Updates and extends tests to match
- Adds a test for the Usage.add method

### Motivation
- Aligns the SDK’s usage with the latest OpenAI responses API Usage object
- Supports downstream use cases that require fine-grained token usage data (e.g., billing, analytics, optimization) requested by startups

### DIFF
--- a/src/agents/usage.py
+++ b/src/agents/usage.py
@@ -1,4 +1,17 @@
 from dataclasses import dataclass
+from typing import TypeVar
+
+from openai.types.responses.response_usage import InputTokensDetails, OutputTokensDetails
+
+T = TypeVar("T", bound="InputTokensDetails | OutputTokensDetails")
+
+
+def add_numeric_fields(current: T, other: T) -> None:
+    for field in current.__dataclass_fields__:
+        v1 = getattr(current, field, 0)
+        v2 = getattr(other, field, 0)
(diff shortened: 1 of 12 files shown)


src/agents/usage.py
```diff
@@ -1,4 +1,6 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+
+from openai.types.responses.response_usage import InputTokensDetails, OutputTokensDetails
```
rm-openai, line 3: instead of reusing these type perhaps we can just inline the details? e.g.
```
cached_tokens: int | None
...
```

rm-openai (changes requested):
generally looks good, one quick change request

rm-openai (approved):
seems fine actually

src/agents/usage.py
WJPBProjects: Discussed - keeping to match the Responses API

PR #728, opened 2025-05-21
add get_tool_call_name method to extract tool name from ToolCallItem

syedmuneeb321:
Adds a utility method get_tool_call_name to extract the name of a tool from a ToolCallItem. Useful for identifying which tool was invoked during agent execution.

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -248,3 +248,8 @@ class ItemHelpers:
             "output": output,
             "type": "function_call_output",
         }
+    
+    @classmethod
+    def get_tool_call_name(cls, tool_call: ToolCallItem) -> str:
+        """Returns the tool name from a ToolCallItem."""
+        return tool_call.raw_item.name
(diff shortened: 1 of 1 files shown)


src/agents/items.py
```diff
@@ -248,3 +248,15 @@ def tool_call_output_item(
             "output": output,
             "type": "function_call_output",
         }
+    
+    @classmethod
+    def get_tool_call_name(cls, tool_call: ToolCallItem) -> str:
+        """Returns the tool name from a ToolCallItem."""
+        return tool_call.raw_item.name
```
rm-openai, line 255: can we just add this to `ToolCallItem` instead? so you can just do `item.tool_name`

syedmuneeb321: But I used it inside ItemHelper, like how you made different methods that take out the agent’s message.

rm-openai: Yeah, those are for Responses API`items`, to which we cannot add methods because they are not part of this SDK (they are in the openai sdk). Doing `item.tool_name` is far better devx than `ItemHelper.get_tool_call_name(item)`

PR #731, opened 2025-05-21
Hosted MCP support

rm-openai:
---
* #732
* __->__ #731


src/agents/tool.py
```diff
@@ -308,3 +358,13 @@ def decorator(real_func: ToolFunction[...]) -> FunctionTool:
         return _create_function_tool(real_func)
 
     return decorator
+    return decorator
```
pakrym-oai, line 361: Pretty sure the first one should work :)

src/agents/_run_impl.py
```diff
@@ -375,6 +409,34 @@ def process_model_response(
                 computer_actions.append(
                     ToolRunComputerAction(tool_call=output, computer_tool=computer_tool)
                 )
+            elif isinstance(output, McpApprovalRequest):
```
pakrym-oai, line 412: nit: inconsistent MCP casing

rm-openai: Comes from the SDK unfortunately. I gave the feedback but it wasnt taken into account

pakrym-oai: I guess that's the SDK

pakrym-oai (approved):

PR #732, opened 2025-05-21
Add support for local shell, image generator, code interpreter tools

rm-openai:


src/agents/_run_impl.py
```diff
@@ -1021,3 +1088,54 @@ async def _get_screenshot_async(
...
+                else _coro.noop_coroutine()
+            ),
+        )
+
+        return ToolCallOutputItem(
+            agent=agent,
+            output=output,
+            raw_item={
+                "type": "local_shell_call_output",
+                "id": call.tool_call.call_id,
```
pakrym-oai, line 1137: Isn't this parameter call_id

pakrym-oai (approved):

src/agents/_run_impl.py
rm-openai: that param doesn't exist any more, idk why. I need to double check.

PR #742, opened 2025-05-22
fix(generation_span): Include `input` when `tracing.include_data()`.

daavoo:


daavoo:
It seems that the data is included but is not available `on_span_start`

Issue #719, opened 2025-05-20
Make intermediate results available when `MaxTurnExceededException` is thrown


chiehmin-wei:
### Describe the feature
I want to access intermediate results when `MaxTurnExceededException` is thrown.

I have a workflow broken into 2 phases.
Phase 1 is investigation.
Phase 2 is writing up a report based on the investigation results.

I'm feeding the results of phase 1 into phase 2 like so:
```
result = Runner.run(investigation_agent, max_turns=5)
report = Runner.run(report_agent, result.to_input_list())
```

I want to limit the number of turns in Phase 1.
If a `MaxTurnExceededException` is thrown in Phase 1, I still want to make use of the available information in Phase 2.
Currently this doesn't seem possible to do. I have no way to access the intermediate results (tool calls, etc.) if I reach max turns in Phase 1.

rm-openai:
Yes we should indeed do this. Will try to get to it soon - as always, PR welcome from anyone

DanieleMorotti:
I can have a look if @chiehmin-wei is not already working on it. How would you return / save the partial results?
In the `Runner` object?

rm-openai:
I think partial results are already built up in Runner, as you described; when the exception is raised, you can add a RunErrorDetails object with all that data

DanieleMorotti:
Okay, so I have to create a new class `RunErrorDetails` and then pass the data in the `MaxTurnsExceeded` exception? Obviously adding this optional attribute to the exception

rm-openai:
yup!

seratch:
Resolved by #743

PR #743, opened 2025-05-22
Added RunErrorDetails object for MaxTurnsExceeded exception

DanieleMorotti:
### Summary

Introduced the `RunErrorDetails` object to get partial results from a run interrupted by `MaxTurnsExceeded` exception. In this proposal the `RunErrorDetails` object contains all the fields from `RunResult` with `final_output` set to `None` and `output_guardrail_results` set to an empty list. We can decide to return less information.

@rm-openai At the moment the exception doesn't return the `RunErrorDetails` object for the streaming mode. Do you have any suggestions on how to deal with it? In the `_check_errors` function of `agents/result.py` file.

### Test plan

I have not implemented any tests currently, but if needed I can implement a basic test to retrieve partial data.

### Issue number

This PR is an attempt to solve issue #719 

### Checks

- [✅ ] I've added new tests (if relevant)
- [ ✅] I've run `make lint` and `make format`
- [ ✅] I've made sure tests pass


src/agents/exceptions.py
```diff
@@ -1,7 +1,8 @@
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
```
rm-openai: instead of Optiona, you can `from __future__ import annotations` then do `error_details: ErrorDetails | None`

src/agents/result.py
```diff
@@ -244,3 +248,15 @@ def _cleanup_tasks(self):
 
     def __str__(self) -> str:
         return pretty_print_run_result_streaming(self)
+
+@dataclass
+class RunErrorDetails(RunResultBase):
```
rm-openai: I don't think it makes sense to subclass RunResultBase. The error details and result can/should diverge - for example, `final_output` doesn't make sense in error details, because by definition there is no error.

src/agents/run.py
```diff
@@ -208,7 +208,20 @@ async def run(
...
+                        run_error_details = RunErrorDetails(
+                            input=original_input,
+                            new_items=generated_items,
+                            raw_responses=model_responses,
+                            final_output=None,
+                            input_guardrail_results=input_guardrail_results,
+                            output_guardrail_results=[],
+                            context_wrapper=context_wrapper,
+                            _last_agent=current_agent
+                        )
```
rm-openai: it might make more sense to create this in the outermost `try-catch`. You'll catch all exceptions there, so you can capture and re-raise

rm-openai (changes requested):

DanieleMorotti:
Hi, thanks for the suggestions. I should have implemented the requested changes, let me know if I understood correctly or I have to correct the code

DanieleMorotti:
I took some code from the draft PR and adapted it to my previous commits. Codex nailed it! Congratulations on its release.

With these latest changes, I’ve added:
- Error details for all exceptions
- Streaming support
- Some basic unit tests

Let me know if now it may be ok or we still need some changes.

src/agents/result.py
```diff
@@ -208,29 +216,79 @@ async def stream_events(self) -> AsyncIterator[StreamEvent]:
...
-            self._stored_exception = MaxTurnsExceeded(f"Max turns ({self.max_turns}) exceeded")
+            max_turns_exc = MaxTurnsExceeded(f"Max turns ({self.max_turns}) exceeded")
+            max_turns_exc.run_data = RunErrorDetails(
+                input=self.input,
+                new_items=self.new_items,
+                raw_responses=self.raw_responses,
+                last_agent=self.current_agent,
+                context_wrapper=self.context_wrapper,
+                input_guardrail_results=self.input_guardrail_results,
+                output_guardrail_results=self.output_guardrail_results,
```
rm-openai: this code seems ~identical to lines 236-243 and the ones before, thoughts on extracting into `self._create_error_details`?

rm-openai (changes requested):
LGTM except one minor request!

src/agents/result.py
DanieleMorotti: Yes, I agree. I also noticed it but I was waiting for your feedback. I'll change this.

Do I also need to add an explanation of these changes to the documentation?

chiehmin-wei:
thanks @DanieleMorotti for implementing this
cc @rm-openai seems like the PR is almost ready to merge? will i need to wait until the next library release to use it?

rm-openai:
> thanks @DanieleMorotti for implementing this cc @rm-openai seems like the PR is almost ready to merge? will i need to wait until the next library release to use it?

yup, will cut a new release soon

PR #752, opened 2025-05-24
Add Session Memory for Automatic Conversation History Management

knowsuchagency:
# Add Session Memory for Automatic Conversation History Management

## Overview

This PR introduces **Session Memory**, a new core feature that automatically maintains conversation history across multiple agent runs, eliminating the need to manually handle `.to_input_list()` between turns.

### 🧠 Automatic Memory Management
- **Zero-effort conversation continuity**: Agents automatically remember previous context without manual state management
- **Session-based organization**: Each conversation is isolated by unique session IDs
- **Seamless integration**: Works with existing `Runner.run()`, `Runner.run_sync()`, and `Runner.run_streamed()` methods

### 💾 Flexible Storage Options
- **In-memory SQLite**: Perfect for temporary conversations during development
- **Persistent SQLite**: File-based storage for conversations that survive application restarts
- **Custom implementations**: Easy-to-implement protocol for Redis, PostgreSQL, or other backends

### 🔧 Simple API
```python
# Before: Manual conversation management
result1 = await Runner.run(agent, "What's the weather?")
new_input = result1.to_input_list() + [{"role": "user", "content": "How about tomorrow?"}]
result2 = await Runner.run(agent, new_input)

# After: Automatic with Session Memory
memory = SQLiteSessionMemory()
config = RunConfig(memory=memory, session_id="user_123")

result1 = await Runner.run(agent, "What's the weather?", run_config=config)
result2 = await Runner.run(agent, "How about tomorrow?", run_config=config)  # Remembers context automatically
```

### Core Implementation
- **`SessionMemory` Protocol**: Interface for implementing custom memory backends
- **`SQLiteSessionMemory` Class**: Production-ready SQLite implementation with thread-safe operations
- **`RunConfig` Integration**: New `memory` and `session_id` parameters for seamless usage

### Documentation
- **Comprehensive guide**: New `docs/session_memory.md` with examples, best practices, and API reference
- **Updated running agents docs**: Enhanced conversation management section
- **API reference**: Full mkdocstrings integration for auto-generated documentation

### Examples
- **Complete working example**: `examples/basic/session_memory_example.py` demonstrating real-world usage
- **Multiple conversation patterns**: Shows how to manage different user sessions

### For Developers
- **Reduces boilerplate**: No more manual `.to_input_list()` management
- **Prevents memory leaks**: Automatic cleanup and organized storage
- **Easier debugging**: Clear conversation history tracking
- **Production ready**: Thread-safe, validated, and tested

### For Applications
- **Better user experience**: Seamless conversation continuity
- **Scalable architecture**: Support for multiple concurrent conversations
- **Flexible deployment**: In-memory for development, persistent for production
- **Multi-agent support**: Same conversation history can be shared across different agents

### Basic Usage
```python
from agents import Agent, Runner, RunConfig, SQLiteSessionMemory

agent = Agent(name="Assistant", instructions="Reply concisely.")
memory = SQLiteSessionMemory()
config = RunConfig(memory=memory, session_id="conversation_123")

# Conversation flows naturally
await Runner.run(agent, "Hi, I'm planning a trip to Japan", run_config=config)
await Runner.run(agent, "What's the best time to visit?", run_config=config)
await Runner.run(agent, "How about cherry blossom season?", run_config=config)
```

### Multiple Sessions
```python
# Different users get separate conversation histories
user1_config = RunConfig(memory=memory, session_id="user_alice")
user2_config = RunConfig(memory=memory, session_id="user_bob")

# Completely isolated conversations
await Runner.run(agent, "I like pizza", run_config=user1_config)
await Runner.run(agent, "I like sushi", run_config=user2_config)
```

### Custom Memory Backend
```python
class RedisSessionMemory:
    async def get_messages(self, session_id: str) -> List[dict]: ...
    async def add_messages(self, session_id: str, messages: List[dict]) -> None: ...
    async def clear_session(self, session_id: str) -> None: ...

# Drop-in replacement for any storage backend
config = RunConfig(memory=RedisSessionMemory(), session_id="session_123")
```

## Technical Details

- **Thread-safe SQLite operations** with connection pooling
- **Automatic schema management** with proper indexing
- **JSON serialization** for message storage
- **Graceful error handling** with detailed validation
- **Memory-efficient** conversation retrieval and storage

## Breaking Changes

None. This is a purely additive feature that doesn't affect existing functionality.

## Documentation

- Updated core concepts in `docs/index.md` to highlight Session Memory as a key primitive
- New comprehensive guide at `docs/session_memory.md`
- Enhanced `docs/running_agents.md` with automatic vs manual conversation management
- Full API reference integration via `docs/ref/memory.md`

Session Memory represents a significant quality-of-life improvement for building conversational AI applications with the Agents SDK, making it easier than ever to create stateful, context-aware agent interactions.


norton120:
This seems like an obvious thing we should have. Maybe even extend this eventually to add simple rag to the session, that's more boilerplate I'd love not to have to build. 

👍 ➕

src/agents/memory/session.py
```diff
@@ -0,0 +1,331 @@
...
+import threading
+from pathlib import Path
+from typing import Any, TYPE_CHECKING, Protocol, runtime_checkable
+
+if TYPE_CHECKING:
+    from ..items import TResponseInputItem
+
+
+@runtime_checkable
+class Session(Protocol):
```
rm-openai, line 16: nit, I prefer to use `ABC` instead, because it is better checked by type checkers, its  is more readable (because it's clear when reading the code that the implementation is for a the interface). Also, isinstance checks are faster with ABC.

src/agents/memory/session.py
```diff
@@ -0,0 +1,331 @@
...
+    """Protocol for session implementations.
+
+    Session stores conversation history for a specific session, allowing
+    agents to maintain context without requiring explicit manual memory management.
+    """
+
+    session_id: str
+
+    async def get_messages(self, amount: int | None = None) -> list[TResponseInputItem]:
```
rm-openai: I think `count` or `limit` is probably better?

src/agents/memory/session.py
```diff
@@ -0,0 +1,331 @@
...
+
+    async def add_messages(self, messages: list[TResponseInputItem]) -> None:
+        """Add new messages to the conversation history.
+
+        Args:
+            messages: List of input items to add to the history
+        """
+        ...
+
+    async def pop_message(self) -> TResponseInputItem | None:
```
rm-openai: what the usecase for popping messages?

rm-openai (commented):
This is a nice PR, simple interface. Going to have a couple more people weigh in as well. @damon-openai @dkundel-openai wdyt?

dkundel-openai:
I like the overall concept and design. Only concern on my side would be confusion if someone passes both a session and an array of inputs (as opposed to just a string like in the examples). 
But supportive of this

src/agents/memory/session.py
knowsuchagency: It would be useful for making edits to the last user message or generating a new assistant response to the last user message.

src/agents/memory/session.py
knowsuchagency: Agreed; I'll update the code

src/agents/memory/session.py
knowsuchagency: I would also go with ABC if it were meant only for internal use. The advantage of using Protocol is third-party implementations of our Session interface don't need to depend on openai-agents. We could even offer both -- a `Session` (Protocol) and `SessionABC`, with the protocol as the primary interface since ABCs automatically satisfy matching Protocols.

knowsuchagency:
That's an excellent point @dkundel-openai. I think it's better to raise an exception if a user tries passing both a list and a session. It would be confusing whether the list input should be appended to the existing session or replace it entirely. I'd rather be forced to choose between the two as a user.

knowsuchagency:
I made it so using a session with list input raises an exception. There's a test to verify the exception gets raised.

src/agents/memory/session.py
knowsuchagency: In the latest version, there's both. We use the Session Protocol for type hints, but define a SessionABC that the SQLiteSession inherits from

rm-openai:
@knowsuchagency also wanted to get your thoughts on an alternative API. 
```
with SQLSession(...) as session:
  session.add_input(...) # Add some initial input 
  result = await Runner.run(..., input=session.inputs()) # Run the runner 
  session.add_input(result) # Update the session with new inputs from the Runner
  // .. could do session.{add,remove,extend,replace,etc)() 
  result = await Runner.run(..., input=session.inputs())
```

My rationale for this is:
1. It keeps the API of `Runner` simpler (e.g. no conflict between the list input and session, etc)
2. It makes the impl of `Runner` much simpler as well, no need to manage the state
3. Session is clearly stateful and Runner is clearly not. 

WDYT?
  

Also, can we update the docs/code comments to say something like "Beta, feedback welcome"? I'd like to make it clear that the session API is subject to change, since it's brand new and i'm sure people will have feedback

knowsuchagency:
@rm-openai I think it's a good idea to mention the API is subject to change pre-1.0

I'm not sure I agree the Runner is stateless, though. It runs a loop and accumulates messages culminating in a `final_output` for the user. 

It seems natural managing conversation history via our Session interface would be part of the loop.

rm-openai:
Yeah I just worry about the Runner becoming unmanageably complicated and needing to take care of a lot of stuff. Separating pieces out where possible is useful. 

For example, we're looking into resumability, where you could pause the Runner and serialize it's state for later. If we also had to serialize the session state, that's one more thing that could go wrong.

So currently still leaning towards keeping it as a separate helper. Will think on it a bit more.

If you/anyone else has convincing arguments in either direction, would love to hear them.

knowsuchagency:
It makes sense not to overcomplicate the runner, but the goal for the PR/issue is to simplify managing conversation history. 

If the user has to explicitly `session.add_input(...) ... Runner.run(..., input=session.inputs())` in their application code, it defeats the purpose.

damvolkov:
Guys, this implementation looks very good and simple interface. I've been using in the meanwhile instead this one, because the versatility of openai agents hooks allow me to do it and it worked super nice (notice that I used mem0ai API REST server as memory backend, but could be  anyone if we create a proper interface):

````import asyncio
import logging
import json
from typing import TypeVar, Generic

from agents import Agent,  RunContextWrapper, Runner, TResponseInputItem, Tool
from agents.lifecycle import RunHooks
from agents.run import RunResult, RunConfig, RunResultStreaming
from mem0 import AsyncMemoryClient
import randomname
from fastapi.encoders import jsonable_encoder
from agents.tool import FunctionTool

from server.mem0ai.models import Message, MemoryCreate, MemoryData, MemoryCreateResponse, MemoryListResponse, Metadata, EntityType, HandoffData


DEFAULT_MAX_TURNS = 10


TMemoryContext = TypeVar('TMemoryContext', bound=BaseModel)


class MemoryManager:
    def __init__(self, client: AsyncMemoryClient):
        self.client = client

    async def _add_memories_task(self, user_id: str, messages: list[Message], metadata: Metadata | None = None) -> list[Message]:
        """Add memories to storage with duplicate check."""
        logging.info(f"Adding {len(messages)} memories for user {user_id}")
        
        # Check for duplicates by getting the last memory
        try:
            last_memories = await self.client.get_all(user_id=user_id, limit=1)
            if last_memories.get('results'):
                last_memory_content = last_memories['results'][0]['memory']
                # Check if the last message content matches any of the new messages
                for message in messages:
                    if message.content == last_memory_content:
                        logging.warning(f"Duplicate memory detected for user {user_id}: '{message.content[:50]}...' - skipping")
                        return []
        except Exception as e:
            logging.debug(f"Could not check for duplicates: {e}")

        memory_create = MemoryCreate(
            user_id=user_id,
            messages=messages,
            metadata=metadata
        )
        res = await self.client.add(**memory_create.model_dump(mode="json", exclude_none=True))
        response = MemoryCreateResponse.model_validate(res)
        return [Message(role=m.role, content=m.memory) for m in response.results]

    async def add_memories(
        self, 
        user_id: str, 
        messages: Message | list[Message], 
        metadata: Metadata | None = None,
        background: bool = True
    ) -> list[Message] | None:
        """Add single message or list of messages to memory."""
        # Convert single message to list
        if isinstance(messages, Message):
            messages = [messages]
        
        if background:
            asyncio.create_task(self._add_memories_task(user_id, messages, metadata))
            return None
        return await self._add_memories_task(user_id, messages, metadata)

    async def get_memory(self, memory_id: str) -> Message:
        """Get single memory by ID."""
        logging.info(f"Getting memory for {memory_id}")
        res = await self.client.get(memory_id=memory_id)
        memory = MemoryData.model_validate(res)
        return Message(role=memory.role, content=memory.memory)
    
    async def get_memories(
        self, 
        user_id: str, 
        limit: int | None = None, 
        filter_by: dict | None = None
    ) -> list[Message]:
        """Get memories with optional limit and filtering."""
        logging.info(f"Getting memories for user {user_id} with limit={limit}, filter_by={filter_by}")
        
        # Get memories with optional limit
        params = {"user_id": user_id}
        if limit:
            params["limit"] = limit
            
        res = await self.client.get_all(**params)
        response = MemoryListResponse.model_validate(res)
        
        # Apply filtering if specified
        if filter_by:
            response = response.filter_by(filter_by)
        
        return [Message(role=m.role, content=m.memory) for m in response.results]


class MemoryContext(BaseModel):
    user_id: str = randomname.get_name()
    thread_id: str = "default_thread"
    user_input: str | list[TResponseInputItem] = ""
    metadata: dict | None = None


@dataclass
class MemoryAgent(Agent[TMemoryContext], Generic[TMemoryContext]):
    memory_manager: MemoryManager | None = None


class MemoryHooks(RunHooks[TMemoryContext], Generic[TMemoryContext]):

    async def on_agent_start(
        self, 
        context: RunContextWrapper[MemoryContext], 
        agent: MemoryAgent
    ) -> None:
        print(f"🔍 AGENT START: {agent.name}, Input: {str(context.context.user_input)[:100]}...")
        message = Message(role="user", content=str(context.context.user_input))
        metadata = Metadata(
            entity=EntityType.AGENT, 
            structured=False, 
            io="input", 
            name=agent.name
        )
        await agent.memory_manager.add_memories(
            user_id=context.context.user_id, 
            messages=message, 
            metadata=metadata,
            background=False
        )

    async def on_agent_end(
        self,
        context: RunContextWrapper[MemoryContext],
        agent: MemoryAgent,
        output: str | BaseModel
    ) -> None:
        print(f"🔍 AGENT END: {agent.name}, Output: {str(output)[:100]}...")
        content = output.model_dump_json() if isinstance(output, BaseModel) else str(output)
        message = Message(role="assistant", content=content)
        metadata = Metadata(
            entity=EntityType.AGENT, 
            structured=isinstance(output, BaseModel), 
            io="output", 
            name=agent.name
        )
        await agent.memory_manager.add_memories(
            user_id=context.context.user_id, 
            messages=message, 
            metadata=metadata,
            background=False
        )

    async def on_handoff(
        self,
        context: RunContextWrapper[MemoryContext],
        from_agent: MemoryAgent,
        to_agent: MemoryAgent,
    ) -> None:
        print(f"🔄 HANDOFF: {from_agent.name} → {to_agent.name}")
        handoff_data = HandoffData(from_agent=from_agent.name, to_agent=to_agent.name)
        handoff_message = Message(role="system", content=handoff_data.model_dump_json())
        metadata = Metadata(
            entity=EntityType.HANDOFF, 
            structured=True, 
            io="output", 
            name=f"{from_agent.name}->{to_agent.name}"
        )
        await from_agent.memory_manager.add_memories(
            user_id=context.context.user_id, 
            messages=handoff_message, 
            metadata=metadata,
            background=True
        )

    async def on_tool_start(
        self,
        context: RunContextWrapper[MemoryContext],
        agent: MemoryAgent,
        tool: FunctionTool,
    ) -> None:
        print(f"🔧 TOOL START: {agent.name} calling '{tool.name}'")
        tool_input_schema = {
            "tool_name": tool.name,
            "description": tool.description,
            "params_schema": tool.params_json_schema
        }
        
        tool_start_message = Message(role="system", content=json.dumps(tool_input_schema))
        metadata = Metadata(
            entity=EntityType.TOOL, 
            structured=True, 
            io="input", 
            name=tool.name
        )
        await agent.memory_manager.add_memories(
            user_id=context.context.user_id, 
            messages=tool_start_message, 
            metadata=metadata,
            background=False
        )

    async def on_tool_end(
        self,
        context: RunContextWrapper[MemoryContext],
        agent: MemoryAgent,
        tool: FunctionTool,
        result: str | BaseModel,
    ) -> None:
        print(f"🔧 TOOL END: {agent.name} finished '{tool.name}' → {result}")
        logging.info(f"Tool {tool.name} finished by {agent.name} with result {result}")
        tool_result_data = {
            "tool_name": tool.name,
            "result": result.model_dump_json() if isinstance(result, BaseModel) else str(result),
            "success": True
        }
        tool_end_message = Message(role="system", content=json.dumps(tool_result_data))
        metadata = Metadata(
            entity=EntityType.TOOL, 
            structured=isinstance(result, BaseModel), 
            io="output", 
            name=tool.name
        )
        await agent.memory_manager.add_memories(
            user_id=context.context.user_id, 
            messages=tool_end_message, 
            metadata=metadata,
            background=False
        )


class MemoryRunner(Runner):

    @staticmethod
    async def _inject_previous_memories(
        user_id: str,
        input: str | list[TResponseInputItem],
        memory_manager: MemoryManager,
        memories = [m.model_dump(include={"role", "content"}) for m in await memory_manager.get_memories(user_id)]
        return memories + [{"role": "user", "content": input}]

    @classmethod
    async def run(
        cls,
        starting_agent: MemoryAgent[MemoryContext],
        input: str | list[TResponseInputItem],
        *,
        context: MemoryContext,
        max_turns: int = DEFAULT_MAX_TURNS,
        hooks: MemoryHooks[MemoryContext] | None = None,
        run_config: RunConfig | None = None,
        previous_response_id: str | None = None,
    ) -> RunResult:
        user_input =  await cls._inject_previous_memories(user_id=context.user_id, input=input, memory_manager=starting_agent.memory_manager)
        return await super().run(starting_agent, input=user_input, context=context, max_turns=max_turns, hooks=hooks, run_config=run_config, previous_response_id=previous_response_id)

    @classmethod
    def run_sync(
        cls,
        starting_agent: MemoryAgent[MemoryContext],
        input: str | list[TResponseInputItem],
        *,
        context: MemoryContext,
        max_turns: int = DEFAULT_MAX_TURNS,
        hooks: MemoryHooks[MemoryContext] | None = None,
        run_config: RunConfig | None = None,
        previous_response_id: str | None = None,
    ) -> RunResult:
        user_input = asyncio.get_event_loop().run_until_complete(cls._inject_previous_memories(user_id=context.user_id, input=input, memory_manager=starting_agent.memory_manager))
        return super().run_sync(starting_agent, input=user_input, context=context, max_turns=max_turns, hooks=hooks, run_config=run_config, previous_response_id=previous_response_id)
    
    @classmethod
    def run_streamed(
        cls,
        starting_agent: MemoryAgent[MemoryContext],
        input: str | list[TResponseInputItem],
        *,
        context: MemoryContext,
        max_turns: int = DEFAULT_MAX_TURNS,
        hooks: MemoryHooks[MemoryContext] | None = None,
        run_config: RunConfig | None = None,
        previous_response_id: str | None = None,
    ) -> RunResultStreaming:
        user_input = asyncio.get_event_loop().run_until_complete(cls._inject_previous_memories(user_id=context.user_id, input=input, memory_manager=starting_agent.memory_manager))
        return super().run_streamed(starting_agent, input=user_input, context=context, max_turns=max_turns, hooks=hooks, run_config=run_config, previous_response_id=previous_response_id)
``

**Comment by 5101good** [comment]
> Guys, this implementation looks very good and simple interface. I've been using in the meanwhile instead this one, because the versatility of openai agents hooks allow me to do it and it worked super nice (notice that I used mem0ai API REST server as memory backend, but could be anyone if we create a proper interface):
> 
> ```
> import logging
> import json
> from typing import TypeVar, Generic
> from dataclasses import dataclass
> 
> from agents import Agent,  RunContextWrapper, Runner, TResponseInputItem, Tool
> from agents.lifecycle import RunHooks
> from agents.run import RunResult, RunConfig, RunResultStreaming
> from mem0 import AsyncMemoryClient
> from pydantic import BaseModel
> import randomname
> from fastapi.encoders import jsonable_encoder
> from agents.tool import FunctionTool
> 
> from server.mem0ai.models import Message, MemoryCreate, MemoryData, MemoryCreateResponse, MemoryListResponse, Metadata, EntityType, HandoffData
> 
> 
> DEFAULT_MAX_TURNS = 10
> 
> 
> TMemoryContext = TypeVar('TMemoryContext', bound=BaseModel)
> 
> 
> class MemoryManager:
>     def __init__(self, client: AsyncMemoryClient):
>         self.client = client
> 
>     async def _add_memories_task(self, user_id: str, messages: list[Message], metadata: Metadata | None = None) -> list[Message]:
>         """Add memories to storage with duplicate check."""
>         logging.info(f"Adding {len(messages)} memories for user {user_id}")
>         
>         # Check for duplicates by getting the last memory
>         try:
>             last_memories = await self.client.get_all(user_id=user_id, limit=1)
>             if last_memories.get('results'):
>                 last_memory_content = last_memories['results'][0]['memory']
>                 # Check if the last message content matches any of the new messages
>                 for message in messages:
>                     if message.content == last_memory_content:
>                         logging.warning(f"Duplicate memory detected for user {user_id}: '{message.content[:50]}...' - skipping")
>                         return []
>         except Exception as e:
>             logging.debug(f"Could not check for duplicates: {e}")
> 
>         memory_create = MemoryCreate(
>             user_id=user_id,
>             messages=messages,
>             metadata=metadata
>         )
>         res = await self.client.add(**memory_create.model_dump(mode="json", exclude_none=True))
>         response = MemoryCreateResponse.model_validate(res)
>         return [Message(role=m.role, content=m.memory) for m in response.results]
> 
>     async def add_memories(
>         self, 
>         user_id: str, 
>         messages: Message | list[Message], 
>         metadata: Metadata | None = None,
>         background: bool = True
>     ) -> list[Message] | None:
>         """Add single message or list of messages to memory."""
>         # Convert single message to list
>         if isinstance(messages, Message):
>             messages = [messages]
>         
>         if background:
>             asyncio.create_task(self._add_memories_task(user_id, messages, metadata))
>             return None
>         return await self._add_memories_task(user_id, messages, metadata)
> 
>     async def get_memory(self, memory_id: str) -> Message:
>         """Get single memory by ID."""
>         logging.info(f"Getting memory for {memory_id}")
>         res = await self.client.get(memory_id=memory_id)
>         memory = MemoryData.model_validate(res)
>         return Message(role=memory.role, content=memory.memory)
>     
>     async def get_memories(
>         self, 
>         user_id: str, 
>         limit: int | None = None, 
>         filter_by: dict | None = None
>     ) -> list[Message]:
>         """Get memories with optional limit and filtering."""
>         logging.info(f"Getting memories for user {user_id} with limit={limit}, filter_by={filter_by}")
>         
>         # Get memories with optional limit
>         params = {"user_id": user_id}
>         if limit:
>             params["limit"] = limit
>             
>         res = await self.client.get_all(**params)
>         response = MemoryListResponse.model_validate(res)
>         
>         # Apply filtering if specified
>         if filter_by:
>             response = response.filter_by(filter_by)
>         
>         return [Message(role=m.role, content=m.memory) for m in response.results]
> 
> 
> class MemoryContext(BaseModel):
>     user_id: str = randomname.get_name()
>     thread_id: str = "default_thread"
>     user_input: str | list[TResponseInputItem] = ""
>     metadata: dict | None = None
> 
> 
> @dataclass
> class MemoryAgent(Agent[TMemoryContext], Generic[TMemoryContext]):
>     memory_manager: MemoryManager | None = None
> 
> 
> class MemoryHooks(RunHooks[TMemoryContext], Generic[TMemoryContext]):
> 
>     async def on_agent_start(
>         self, 
>         context: RunContextWrapper[MemoryContext], 
>         agent: MemoryAgent
>     ) -> None:
>         print(f"🔍 AGENT START: {agent.name}, Input: {str(context.context.user_input)[:100]}...")
>         message = Message(role="user", content=str(context.context.user_input))
>         metadata = Metadata(
>             entity=EntityType.AGENT, 
>             structured=False, 
>             io="input", 
>             name=agent.name
>         )
>         await agent.memory_manager.add_memories(
>             user_id=context.context.user_id, 
>             messages=message, 
>             metadata=metadata,
>             background=False
>         )
> 
>     async def on_agent_end(
>         self,
>         context: RunContextWrapper[MemoryContext],
>         agent: MemoryAgent,
>         output: str | BaseModel
>     ) -> None:
>         print(f"🔍 AGENT END: {agent.name}, Output: {str(output)[:100]}...")
>         content = output.model_dump_json() if isinstance(output, BaseModel) else str(output)
>         message = Message(role="assistant", content=content)
>         metadata = Metadata(
>             entity=EntityType.AGENT, 
>             structured=isinstance(output, BaseModel), 
>             io="output", 
>             name=agent.name
>         )
>         await agent.memory_manager.add_memories(
>             user_id=context.context.user_id, 
>             messages=message, 
>             metadata=metadata,
>             background=False
>         )
> 
>     async def on_handoff(
>         self,
>         context: RunContextWrapper[MemoryContext],
>         from_agent: MemoryAgent,
>         to_agent: MemoryAgent,
>     ) -> None:
>         print(f"🔄 HANDOFF: {from_agent.name} → {to_agent.name}")
>         handoff_data = HandoffData(from_agent=from_agent.name, to_agent=to_agent.name)
>         handoff_message = Message(role="system", content=handoff_data.model_dump_json())
>         metadata = Metadata(
>             entity=EntityType.HANDOFF, 
>             structured=True, 
>             io="output", 
>             name=f"{from_agent.name}->{to_agent.name}"
>         )
>         await from_agent.memory_manager.add_memories(
>             user_id=context.context.user_id, 
>             messages=handoff_message, 
>             metadata=metadata,
>             background=True
>         )
> 
>     async def on_tool_start(
>         self,
>         context: RunContextWrapper[MemoryContext],
>         agent: MemoryAgent,
>         tool: FunctionTool,
>     ) -> None:
>         print(f"🔧 TOOL START: {agent.name} calling '{tool.name}'")
>         tool_input_schema = {
>             "tool_name": tool.name,
>             "description": tool.description,
>             "params_schema": tool.params_json_schema
>         }
>         
>         tool_start_message = Message(role="system", content=json.dumps(tool_input_schema))
>         metadata = Metadata(
>             entity=EntityType.TOOL, 
>             structured=True, 
>             io="input", 
>             name=tool.name
>         )
>         await agent.memory_manager.add_memories(
>             user_id=context.context.user_id, 
>             messages=tool_start_message, 
>             metadata=metadata,
>             background=False
>         )
> 
>     async def on_tool_end(
>         self,
>         context: RunContextWrapper[MemoryContext],
>         agent: MemoryAgent,
>         tool: FunctionTool,
>         result: str | BaseModel,
>     ) -> None:
>         print(f"🔧 TOOL END: {agent.name} finished '{tool.name}' → {result}")
>         logging.info(f"Tool {tool.name} finished by {agent.name} with result {result}")
>         tool_result_data = {
>             "tool_name": tool.name,
>             "result": result.model_dump_json() if isinstance(result, BaseModel) else str(result),
>             "success": True
>         }
>         tool_end_message = Message(role="system", content=json.dumps(tool_result_data))
>         metadata = Metadata(
>             entity=EntityType.TOOL, 
>             structured=isinstance(result, BaseModel), 
>             io="output", 
>             name=tool.name
>         )
>         await agent.memory_manager.add_memories(
>             user_id=context.context.user_id, 
>             messages=tool_end_message, 
>             metadata=metadata,
>             background=False
>         )
> 
> 
> class MemoryRunner(Runner):
> 
>     @staticmethod
>     async def _inject_previous_memories(
>         user_id: str,
>         input: str | list[TResponseInputItem],
>         memory_manager: MemoryManager,
>     ) -> list[TResponseInputItem]:
>         memories = [m.model_dump(include={"role", "content"}) for m in await memory_manager.get_memories(user_id)]
>         return memories + [{"role": "user", "content": input}]
> 
>     @classmethod
>     async def run(
>         cls,
>         starting_agent: MemoryAgent[MemoryContext],
>         input: str | list[TResponseInputItem],
>         *,
>         context: MemoryContext,
>         max_turns: int = DEFAULT_MAX_TURNS,
>         hooks: MemoryHooks[MemoryContext] | None = None,
>         run_config: RunConfig | None = None,
>         previous_response_id: str | None = None,
>     ) -> RunResult:
>         user_input =  await cls._inject_previous_memories(user_id=context.user_id, input=input, memory_manager=starting_agent.memory_manager)
>         return await super().run(starting_agent, input=user_input, context=context, max_turns=max_turns, hooks=hooks, run_config=run_config, previous_response_id=previous_response_id)
> 
>     @classmethod
>     def run_sync(
>         cls,
>         starting_agent: MemoryAgent[MemoryContext],
>         input: str | list[TResponseInputItem],
>         *,
>         context: MemoryContext,
>         max_turns: int = DEFAULT_MAX_TURNS,
>         hooks: MemoryHooks[MemoryContext] | None = None,
>         run_config: RunConfig | None = None,
>         previous_response_id: str | None = None,
>     ) -> RunResult:
>         user_input = asyncio.get_event_loop().run_until_complete(cls._inject_previous_memories(user_id=context.user_id, input=input, memory_manager=starting_agent.memory_manager))
>         return super().run_sync(starting_agent, input=user_input, context=context, max_turns=max_turns, hooks=hooks, run_config=run_config, previous_response_id=previous_response_id)
>     
>     @classmethod
>     def run_streamed(
>         cls,
>         starting_agent: MemoryAgent[MemoryContext],
>         input: str | list[TResponseInputItem],
>         *,
>         context: MemoryContext,
>         max_turns: int = DEFAULT_MAX_TURNS,
>         hooks: MemoryHooks[MemoryContext] | None = None,
>         run_config: RunConfig | None = None,
>         previous_response_id: str | None = None,
>     ) -> RunResultStreaming:
>         user_input = asyncio.get_event_loop().run_until_complete(cls._inject_previous_memories(user_id=context.user_id, input=input, memory_manager=starting_agent.memory_manager))
>         return super().run_streamed(starting_agent, input=user_input, context=context, max_turns=max_turns, hooks=hooks, run_config=run_config, previous_response_id=previous_response_id)
> ``
> ```

You've implemented excellent memory functionality, and I believe most agents SDK users currently implement similar solutions themselves, including me.

However, the issue author's goal should be eliminating the mission for users to manage memory manually, or at least simplifying it by having the Runner handle it automatically. This would avoid users having to manually filter and record memories through tracing. I personally think memory is essential for a fully functional agent framework.

**Comment by sibblegp** [comment]
Why aren't sessions managed on OpenAI's API side with just a session_id? That would be so much simpler than having us manage them on the client side.

This is a useless PR. I just dump .to_input_list() into Firebase. Then I fetch it. Literally two lines of code.

Sure, if you want to use SQLLite, this is great, but no way would I take the time to implement a custom version when I can do the same thing in 2 lines of code.

**Comment by knowsuchagency** [comment]
The point of the PR is to provide the interface for session management services. SQLite is provided as a default implementation for convenience.

**Comment by sibblegp** [comment]
@knowsuchagency It just seems a conversation_id we could pass along with Runner which would retrieve context and previous messages from OpenAI so we don't have to send it every time would be easier. Not sure if your API has this capability yet though.

**Comment by knowsuchagency** [comment]
The inspiration is [Google's ADK sessions](https://google.github.io/adk-docs/sessions/session/) where they provide an `InMemorySessionService` and a `VertexAiSessionService`. 

The goal is to provide scaffolding for OpenAI (and others) to provide the same functionality in the most intuitive way possible.

**Comment by sibblegp** [comment]
Nice work, but honestly this could be achieved so easily if the OpenAI API provided something as simple as "session_id" and managed it all on their end.

I already handle this and all I do is serialize .to_result_list() into and out of firebase. It's super simple.

**Comment by rm-openai** [comment]
@knowsuchagency sorry for the long delay and thanks fore the discussion. I think this should be good to merge. Going to give it a thorough review right now - if you can fix the lint, will merge today. Thanks for the idea+implementation!

File: src/agents/run.py [hunk 4]
```diff
@@ -621,154 +657,202 @@ async def _start_streaming(
...
-                    handoff_names = [
-                        h.agent_name
-                        for h in await cls._get_handoffs(current_agent, context_wrapper)
-                    ]
-                    if output_schema := cls._get_output_schema(current_agent):
-                        output_type_name = output_schema.name()
-                    else:
-                        output_type_name = "str"
+            if streamed_result.trace:
+                streamed_result.trace.start(mark_as_current=True)
```
rm-openai: any reason you moved this here?

src/agents/run.py
```diff
@@ -621,154 +657,202 @@ async def _start_streaming(
...
+        session: Session | None,
     ):
-        if streamed_result.trace:
-            streamed_result.trace.start(mark_as_current=True)
-
         current_span: Span[AgentSpanData] | None = None
-        current_agent = starting_agent
-        current_turn = 0
-        should_run_agent_start_hooks = True
-        tool_use_tracker = AgentToolUseTracker()
```
rm-openai: possible to reduce the number of moves/unrelated changes? Would be nice to preserve blame and make code review easier

src/agents/memory/session.py
```diff
@@ -0,0 +1,381 @@
...
+    agents to maintain context without requiring explicit manual memory management.
+
+    This ABC is intended for internal use and as a base class for concrete implementations.
+    Third-party libraries should implement the Session protocol instead.
+    """
+
+    session_id: str
+
+    @abstractmethod
+    async def get_messages(self, limit: int | None = None) -> list[TResponseInputItem]:
```
rm-openai: when would `limit` be passed? Who decides what the limit is?

Also, thoughts on changing this to `get_items`? Messages is a super overloaded term

src/agents/memory/session.py
```diff
@@ -0,0 +1,381 @@
...
+    async def add_messages(self, messages: list[TResponseInputItem]) -> None:
+        """Add new messages to the conversation history.
+
+        Args:
+            messages: List of input items to add to the history
+        """
+        ...
+
+    @abstractmethod
+    async def pop_message(self) -> TResponseInputItem | None:
```
rm-openai: s/messages/items/g?

src/agents/run.py
knowsuchagency: I think there was a regression in the test suite that made it necessary, but I don't remember what the specific issue was

src/agents/memory/session.py
knowsuchagency: agreed items is better. I'll make the change!

src/agents/memory/session.py
knowsuchagency: The main benefit is when you need to render only the last N items in the conversation. But I could also see it being useful for managing the overall size of the context windows i.e. having a sliding window of messages

knowsuchagency:
I was able to lower the size of the diff in run.py in the last commit by 25 lines. Let me know if there are any other changes to make thanks @rm-openai !

src/agents/run.py
```diff
@@ -621,154 +653,200 @@ async def _start_streaming(
...
-                        current_span.finish(reset_current=True)
-                        current_span = None
-                        should_run_agent_start_hooks = True
-                        streamed_result._event_queue.put_nowait(
-                            AgentUpdatedStreamEvent(new_agent=current_agent)
+            streamed_result._event_queue.put_nowait(
+                AgentUpdatedStreamEvent(new_agent=current_agent)
+            )
+
+            try:
```
rm-openai: the nested try is making it really hard to read heh

src/agents/run.py
rm-openai: hmm ok do you think you could move it back and see if necessary?

src/agents/run.py
```diff
@@ -621,154 +653,200 @@ async def _start_streaming(
...
-                # agent changes, or if the agent loop ends.
-                if current_span is None:
-                    handoff_names = [
-                        h.agent_name
-                        for h in await cls._get_handoffs(current_agent, context_wrapper)
-                    ]
-                    if output_schema := cls._get_output_schema(current_agent):
-                        output_type_name = output_schema.name()
-                    else:
-                        output_type_name = "str"
```
rm-openai: same for moving this stuff back

src/agents/run.py
```diff
@@ -621,154 +653,200 @@ async def _start_streaming(
...
-                        streamed_result._event_queue.put_nowait(
-                            AgentUpdatedStreamEvent(new_agent=current_agent)
+            streamed_result._event_queue.put_nowait(
+                AgentUpdatedStreamEvent(new_agent=current_agent)
+            )
+
+            try:
+                while True:
+                    if streamed_result.is_complete:
+                        break
```
rm-openai: feels odd that any of this needs to change, since the only real change should be to prepare the session input at the start + write it at the end?

rm-openai (changes requested):
still seems like a ton of unrelated changes?

src/agents/run.py
knowsuchagency: removed

knowsuchagency:
The changes in run.py should be a lot more straightforward now @rm-openai. The business logic is largely the same with minor edits to accommodate the new Session interface. LMK if you notice any other necessary edits, thanks!

.gitignore
```diff
@@ -143,3 +143,4 @@ cython_debug/
 # PyPI configuration file
 .pypirc
 .aider*
+.claude/settings.local.json
```
slyapustin: @knowsuchagency That seems not to be related to the PR.

src/agents/memory/session.py
```diff
@@ -0,0 +1,381 @@
...
+
+                    cursor = conn.execute(
+                        f"""
+                        SELECT message_data FROM {self.messages_table} 
+                        WHERE session_id = ? 
+                        ORDER BY created_at ASC
+                        LIMIT ? OFFSET ?
+                    """,
+                        (self.session_id, limit, offset),
+                    )
```
slyapustin: @knowsuchagency I think it will be cleaner and faster to use DESC and then just reverse result:

```
cursor = conn.execute(
    f"""
    SELECT message_data FROM {self.messages_table}
    WHERE session_id = ?
    ORDER BY created_at DESC
    LIMIT ?
    """,
    (self.session_id, limit),
)
rows = cursor.fetchall()

# To get them in chronological order (oldest to newest), reverse the list:
messages = list(reversed(rows))

```

src/agents/memory/session.py
```diff
@@ -0,0 +1,378 @@
...
+                    (self.session_id,),
+                )
+
+                result = cursor.fetchone()
+                if result:
+                    message_id, message_data = result
+                    try:
+                        item = json.loads(message_data)
+                        # Delete the item by ID
+                        conn.execute(
```
slyapustin: @knowsuchagency You are deleting that record from DB in both cases.
Is there any reason to keep DB delete call in the try/except block? 
I would suggest to only keep JSON decoding part in the try/except block.

Also, I think you can use single SQL query to delete record and get row back, using [RETURNING](https://www.sqlite.org/lang_returning.html):
> DELETE FROM your_table WHERE id = ? RETURNING *;

knowsuchagency: Good call, thanks. Updated.

rm-openai:
@knowsuchagency this is good to merge, just need to fix the lint/types. You can just copy the last 2 commits from this: https://github.com/openai/openai-agents-python/pull/1040

(I can't add to this PR unfortunately)

sebastian-montero:
Thank you for this @knowsuchagency  !! Looking forward to seeing it merged.

knowsuchagency:
Done thanks @rm-openai !

rm-openai:
Awesome work @knowsuchagency, thanks for bearing with the long review process and getting this in! Excited for people to use it.

knowsuchagency:
Thanks for all the excellent feedback and patience as well @rm-openai !! I agree it will be a great feature for users 😃

knowsuchagency:
Thanks for your review also @slyapustin !

PR #759, opened 2025-05-26
Changed a mistake

Ham-Gholami:
Line 152: gaurdrail_agent -> agent

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -149,7 +149,7 @@ history_tutor_agent = Agent(
 
 
 async def homework_guardrail(ctx, agent, input_data):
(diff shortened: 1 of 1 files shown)


Rehan-Ul-Haq:
The previous code `result = await Runner.run(guardrail_agent, input_data, context=ctx.context)` is correct. This run is used inside `homework_guardrail` function and which is further passed to the triage agent in`input_guardrails`.  This run will run in parallel to the main `result = await Runner.run(triage_agent, "who was the first president of the united states?")` and `result = await Runner.run(triage_agent, "what is life")`.    
Therefore, the previous code is correct. No need to change `guardrail_agent` to `agent`.  Please review.

Issue #559, opened 2025-04-21
function call can not get call_id


goooodle:
### 

return FunctionTool(
        name=tool_name,
        description=tool_desc,
        params_json_schema=params_schema,
        on_invoke_tool=create_run_function(service_key, service_name),
        strict_json_schema=is_service_tool,
    )

def create_run_function(service_key: str, service_name: str):
      async def run_function(context, params_str: str) -> str:

      #当执行多个相同方法名,但是参数不同时,这里无法定位到是哪次调用, 因为该方法无法获取call_id
      #When multiple methods with the same name but different parameters are executed, it is impossible to 
          determine which call it is here because the method cannot obtain the call_id

        params = json.loads(params_str)

        try:
            result = await call_service(service_key=service_key, params=params)
           # context.context[call_id]["invoke_status"] = "success"
            return json.dumps(result, ensure_ascii=False)
        except Exception as e:
           # context.context[call_id]["invoke_status"] = "error"
            return f"Error executing tool: {str(e)}"

return run_function

### Question
当执行多个相同方法名,但是参数不同时,这里无法定位到是哪次调用, 因为该方法无法获取call_id
When multiple methods with the same name but different parameters are executed, it is impossible to determine which call it is here because the method cannot obtain the call_id

rm-openai:
Yeah, from the point of view of this SDK, multiple functions with the same name is an error. LLMs will generally have trouble with functions named the same; what's the use case?

goooodle:
```
async def mock_execute(service_key, params:str):
    return '400'

def create_run_function(service_key: str, service_name: str):
    async def run_function(context, params_str: str) -> str:
        try:
           result = await mock_execute(service_key=service_key, params=params_str)
           # context.context[call_id]["invoke_status"] = "success"   // log process status, but cant not get call_id
           return result
        except Exception as e:
           # context.context[call_id]["invoke_status"] = "error"  // log process status, but cant not get call_id
          return f"Error executing tool: {str(e)}"
    return run_function


class FunctionArgs(BaseModel):
    product_name: str

async def main():
    tools = []
    tools.append(FunctionTool(
        name='query_price',
        description='query price',
        params_json_schema=FunctionArgs.model_json_schema(),
        on_invoke_tool=create_run_function('query_price', 'query price'),
        strict_json_schema=True,
    ))

    agent = Agent(
        name="gpt-4.1",
        instructions="You are a helpful assistant.",
        tools=tools
    )

[...]

PR #766, opened 2025-05-27
Added support for passing tool_call_id via the RunContextWrapper

niv-hertz:
This PR fixes issue: https://github.com/openai/openai-agents-python/issues/559

By adding the tool_call_id to the RunContextWrapper prior to calling tools. This gives the ability to access the tool_call_id in the implementation of the tool.

### DIFF
--- a/src/agents/tool_context.py
+++ b/src/agents/tool_context.py
@@ -1,16 +1,18 @@
-from dataclasses import KW_ONLY, dataclass, fields
+from dataclasses import fields
 from typing import Any
 
 from .run_context import RunContextWrapper, TContext
 
 
-@dataclass
 class ToolContext(RunContextWrapper[TContext]):
     """The context of a tool call."""
 
-    _: KW_ONLY
     tool_call_id: str
     """The ID of the tool call."""
+
+    def __init__(self, *args: Any, tool_call_id: str, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self.tool_call_id = tool_call_id
 
     @classmethod
     def from_agent_context(
--- a/src/agents/tool_context.py
+++ b/src/agents/tool_context.py
@@ -1,0 +1,26 @@
+from dataclasses import KW_ONLY, dataclass, fields
+from typing import Any
+
+from .run_context import RunContextWrapper, TContext
+
+
+@dataclass
+class ToolContext(RunContextWrapper[TContext]):
+    """The context of a tool call."""
+
+    _: KW_ONLY
+    tool_call_id: str
+    """The ID of the tool call."""
+
+    @classmethod
+    def from_agent_context(
+        cls, context: RunContextWrapper[TContext], tool_call_id: str
+    ) -> "ToolContext":
+        """
+        Create a ToolContext from a RunContextWrapper.
+        """
+        # Grab the names of the RunContextWrapper's init=True fields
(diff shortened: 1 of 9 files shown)


niv-hertz:
@rm-openai  - This addressed the issue we discussed in another thread, would love you review :)

rm-openai (changes requested):
Sorry for delayed review. Was thinking about it, and it feels a bit bad to have `tool_call_id` be nullable.

What if instead we had `ToolContext` _subclass_ RunContextWrapper. That way, existing usecases would just work since the downcast is fine. And `tool_call_id` would be non-null always?

niv-hertz:
> Sorry for delayed review. Was thinking about it, and it feels a bit bad to have `tool_call_id` be nullable.
> 
> What if instead we had `ToolContext` _subclass_ RunContextWrapper. That way, existing usecases would just work since the downcast is fine. And `tool_call_id` would be non-null always?

@rm-openai - I added a `ToolContext` object such that new users will be able to create tool functions that receive the `ToolContext` instead of the `AgentContextWrapper`. I also modified the internal `FunctionTool` class to only receive the `ToolContext` object in the `on_invoke_tool` params, so this change will break the type checking of anyone who uses the `FunctionTool` class directly instead of the wrappers. It seemed reasonable to me, let me know if you think otherwise.

niv-hertz:
@rm-openai - bumping this 😄

src/agents/tool_context.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import Any
+
+from .run_context import RunContextWrapper, TContext
+
+
+@dataclass
+class ToolContext(RunContextWrapper[TContext]):
+    """The context of a tool call."""
+
+    _: KW_ONLY
```
rm-openai: unfortunately this is python 3.10+ only i think. And we need 3.9. Maybe just remove it?

rm-openai (approved):
Looks good, please just fix the typecheck etc and I can merge

src/agents/tool_context.py
niv-hertz: @rm-openai  - Without it, I have to make the tool_call_id a keyword-like argument, that means I have to give it a default. It's a bit less natural, but I can default it to an empty string.

rm-openai: Yeah it's super annoying. You could do this:
```
def _assert_must_pass_tool_call_id() -> str:
    raise ValueError("tool_call_id must be passed to ToolRunContext")


@dataclass
class ToolRunContext(RunContextWrapper[TContext]):
    tool_call_id: str = field(default_factory=_assert_must_pass_tool_call_id)
```

That way, this will work:
```
foo = ToolRunContext(context=None, tool_call_id="123")
```

and this will raise at runtime:
```
bar = ToolRunContext(context=None)
```

rm-openai: s/ValueError/UserError

niv-hertz: Nevermind, I found another work-around by removing the dataclass decorator definition, it's ready now :)

niv-hertz: Actually, I like your solution better, implementing it, one sec :)

niv-hertz: Fixed it

rm-openai:
awesome thanks for implementing this!

PR #801, opened 2025-06-02
Ensure item.model_dump only contains JSON serializable types

westhood:
The EmbeddedResource from MCP tool call contains a field with type AnyUrl that is not JSON-serializable. To avoid this exception, use item.model_dump(mode="json") to ensure a JSON-serializable return value.


westhood:
without this patch，the agent framework can not work with mcp tool with `Embedded Resources` reponses, like https://github.com/mark3labs/mcp-filesystem-server

rm-openai (approved):

rm-openai:
thanks!

PR #804, opened 2025-06-02
Only start tracing worker thread on first span/trace

### LINKED ISSUE #796: Background thread is created on library import
### Describe the bug
A background thread is created when the library is imported. This thread is used for tracing and having tracing disabled has no effect on this, as it is evaluated after the thread creation.

This is in my opinion quite bad and it causes issues in our environment, where we use process-based parallelization and require no extra threads to be running at the startup of the program. I am working around this by importing anything from the `agents` library inside my functions, not in the global file scope. Not ideal though.

### Debug information
- Agents SDK version: 0.0.16
- Python version: Python 3.13

### Repro steps

```
import threading
assert threading.active_count() == 1
import agents
assert threading.active_count() == 1
```

### Expected behavior
My expectation would be, that the thread will be created either explicitly, or when needed - for example when the first `Runner.run` is called, or some tracing functionality is explicitly used for the first time.

rm-openai:
Closes #796. Shouldn't start a busy waiting thread if there aren't any traces.

Test plan
```
import threading
assert threading.active_count() == 1
import agents
assert threading.active_count() == 1
```

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -188,10 +188,27 @@ class BatchTraceProcessor(TracingProcessor):
         # Track when we next *must* perform a scheduled export
         self._next_export_time = time.time() + self._schedule_delay
 
-        self._worker_thread = threading.Thread(target=self._run, daemon=True)
-        self._worker_thread.start()
+        # We lazily start the background worker thread the first time a span/trace is queued.
+        self._worker_thread: threading.Thread | None = None
+        self._thread_start_lock = threading.Lock()
+
+    def _ensure_thread_started(self) -> None:
+        # Fast path without holding the lock
+        if self._worker_thread and self._worker_thread.is_alive():
+            return
+
+        # Double-checked locking to avoid starting multiple threads
+        with self._thread_start_lock:
+            if self._worker_thread and self._worker_thread.is_alive():
+                return
+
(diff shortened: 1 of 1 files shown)


src/agents/tracing/processors.py
```diff
@@ -188,10 +188,27 @@ def __init__(
         # Track when we next *must* perform a scheduled export
         self._next_export_time = time.time() + self._schedule_delay
 
-        self._worker_thread = threading.Thread(target=self._run, daemon=True)
-        self._worker_thread.start()
+        # We lazily start the background worker thread the first time a span/trace is queued.
+        self._worker_thread: threading.Thread | None = None
```
pakrym-oai, line 192: Why not start a task + asyncio q, then you don't have an extra thread to worry about and you can start the task unconditionally

rm-openai: Hmm I was thinking:
1. small overhead in cases where there are no traces (the unconditional task would still be scheduled every poll_interval seconds, even though it doesnt have work)
2. It wouldn't work super well in totally sync contexts 
3. cleaning up the asyncio task is kinda annoying imo

lmk if you feel like I should change, happy to.

pakrym-oai: No strong opinions

PR #842, opened 2025-06-10
Allow arbitrary kwargs in model

rm-openai:
Sometimes users want to provide parameters specific to a model provider. This is an escape hatch.

### DIFF
--- a/src/agents/model_settings.py
+++ b/src/agents/model_settings.py
@@ -73,6 +73,11 @@ class ModelSettings:
     """Additional headers to provide with the request.
     Defaults to None if not provided."""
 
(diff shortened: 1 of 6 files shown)


src/agents/model_settings.py
```diff
@@ -73,6 +73,11 @@ class ModelSettings:
     """Additional headers to provide with the request.
     Defaults to None if not provided."""
 
+    kwargs: dict[str, Any] | None = None
```
pakrym-oai: nit: friendlier name maybe? `extra_params` ?

pakrym-oai (approved):

Issue #851, opened 2025-06-12
Filter MCPServerStdio server tools


esandoval30:
# Add Tool Filtering Capability to MCPServerStdio

## Problem Statement
Currently, MCPServerStdio doesn't provide a mechanism to filter which tools are exposed from a given MCP Server. This means all tools from the server are always available, which may not be desirable in cases where we want to limit tool access for security or simplicity reasons.

## Current Implementation
The current MCPServerStdio initialization only allows for basic configuration:

```python
MCPServerStdio(
    params: MCPServerStdioParams,
    cache_tools_list: bool = False,
    name: str | None = None,
    client_session_timeout_seconds: float | None = 5,
)
```

## Desired Feature
Add the ability to filter tools similar to Google ADK's implementation:

```python
MCPServerStdio(
    params: MCPServerStdioParams,
    cache_tools_list: bool = False,
    name: str | None = None,
    client_session_timeout_seconds: float | None = 5,
    tool_filter: list[str] | None = None  # New parameter
)
```

## Use Case Example
```python
# Only expose specific tools from the MCP server
server = MCPServerStdio(
    params={
        "command": "uvx",
        "args": ["mcp-neo4j-cypher@0.2.1"],
        "env": {...}
    },
    tool_filter=['get-neo4j-schema', 'read-neo4j-cypher']  # Only expose these tools
)
```

## Benefits
1. Enhanced security by limiting tool access
2. Simplified tool interfaces for specific use cases
3. Better control over agent capabilities

## Implementation Considerations
- The `tool_filter` parameter should be optional
- When not provided, maintain current behavior of exposing all tools
- Add validation to ensure filtered tools exist on the server
- Consider adding warning logs when requested tools don't exist

## References
- [Google ADK MCP Tools Documentation](https://google.github.io/adk-docs/tools/mcp-tools/#step-1-define-your-agent-with-mcptoolset)

devtalker:
I have same issue, see [830](https://github.com/openai/openai-agents-python/issues/830)

DanieleMorotti:
Hi, considering that this feature has been requested multiple times, I've implemented a PR (#854 ) enabling the filtering of visible tools from an MCP server.
@rm-openai, if you have time, I'd appreciate your feedback. Let me know if you'd like to include this implementation, otherwise I'll close the PR. Thanks!

seratch:
Resolved by #861

PR #854, opened 2025-06-12
Added mcp tool filtering and unit test

### LINKED ISSUE #830: Add tool filtering/selection capability for MCP Servers
### Please read this first

- **Have you read the docs?** Yes, I have read the [Agents SDK docs](https://openai.github.io/openai-agents-python/), particularly the MCP documentation section.
- **Have you searched for related issues?** Yes, I have searched for related issues and haven't found similar feature requests.

### Describe the feature

**Feature Request: Add tool filtering/selection capability for MCP Servers**

**Problem Description:**
Currently, the Agent SDK automatically fetches all available tools from MCP servers without the ability to select only specific tools. This creates inconveniences in the following scenarios:

1. MCP servers provide numerous tools, but the Agent only needs to use a subset of them. In many cases, only a subset of tools is needed, and unwanted tools occupy LLM context and affect the accuracy of Agent tool selection.
2. Different Agents need to use different tool subsets from the same MCP server
3. Security concerns require limiting the scope of tools an Agent can access
4. Avoiding tool name conflicts (when multiple servers have tools with the same name)

DanieleMorotti:
### Summary

Considering multiple requests to add this feature #830, #851 , I added a parameter to MCP servers to filter out some tools.
Let me know if you want to implement something in a different way and if I have to update the documentation.

### Test plan

I added a unit test to check that only the allowed tools are returned by the server.

### Issue number

It should close #830 and #851 .

### Checks

- [ ✅] I've added new tests (if relevant)
- [✅ ] I've run `make lint` and `make format`
- [✅ ] I've made sure tests pass

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -57,7 +57,10 @@ class MCPServer(abc.ABC):
 class _MCPServerWithClientSession(MCPServer, abc.ABC):
     """Base class for MCP servers that use a `ClientSession` to communicate with the server."""
 
-    def __init__(self, cache_tools_list: bool, client_session_timeout_seconds: float | None):
+    def __init__(
+        self, cache_tools_list: bool, client_session_timeout_seconds: float | None,
+        allowed_tools: list[str] | None = None
+    ):
         """
         Args:
             cache_tools_list: Whether to cache the tools list. If `True`, the tools list will be
@@ -68,6 +71,8 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
             (by avoiding a round-trip to the server every time).
 
             client_session_timeout_seconds: the read timeout passed to the MCP ClientSession.
+
+            allowed_tools: the names of the tools from the server that can be accessed.
         """
         self.session: ClientSession | None = None
         self.exit_stack: AsyncExitStack = AsyncExitStack()
@@ -80,6 +85,7 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
         # The cache is always dirty at startup, so that we fetch tools at least once
         self._cache_dirty = True
         self._tools_list: list[MCPTool] | None = None
+        self.allowed_tools = allowed_tools
 
     @abc.abstractmethod
     def create_streams(
@@ -145,6 +151,13 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
 
         # Fetch the tools from the server
         self._tools_list = (await self.session.list_tools()).tools
+        # Filter out tools that should not be available
+        if self.allowed_tools:
+            self._tools_list = [
+                tool for tool in self._tools_list
+                if tool.name in self.allowed_tools
+            ]
+
         return self._tools_list
 
     async def call_tool(self, tool_name: str, arguments: dict[str, Any] | None) -> CallToolResult:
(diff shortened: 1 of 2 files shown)


devtalker:
Hi @DanieleMorotti! 👋

What a coincidence! I was also working on the exact same feature and just submitted [PR #861](https://github.com/openai/openai-agents-python/pull/861). It's really interesting that we both identified this need and decided to tackle it simultaneously - great minds think alike! 😄

I really appreciate your initiative in addressing issues #830 and #851. After looking at both approaches, I'd love to get your thoughts on the different implementation strategies we've taken.

From what I can see, your solution focuses on server-level filtering (which is definitely the right direction!). In my implementation, I tried to provide a bit more flexibility by offering a **dual-layer filtering approach**:

**Server-level filtering** (similar to your approach):
```python
server = MCPServerStdio(
    params={...},
    allowed_tools=["read_file", "write_file"],
    excluded_tools=["delete_file"]
)
```

**Agent-level filtering** (additional layer):
```python
agent = Agent(
    name="Assistant", 
    mcp_servers=[server1, server2],
    mcp_config={
        "allowed_tools": {
            "server1": ["tool1", "tool2"]
        },
        "excluded_tools": {
            "server2": ["dangerous_tool"] 
        }
    }
)
```

The idea is that server-level filtering controls what tools the server exposes, while agent-level filtering allows fine-grained control across multiple servers without modifying server configurations.

I also included support for both `allowed_tools` (whitelist) and `excluded_tools` (blacklist) at both layers, plus comprehensive tests and documentation.

What do you think about this approach? I'd really value your feedback - maybe there are aspects of your implementation that could improve mine, or perhaps we could combine the best of both approaches? 

Would you be interested in taking a look at [PR #861](https://github.com/openai/openai-agents-python/pull/861) and sharing your thoughts? I'm definitely open to collaboration and would love to hear your perspective! 🙏

DanieleMorotti:
Hi @devtalker ! 

Thanks for your detailed reply and the work you put into PR #861 , it's indeed a funny coincidence we tackled this simultaneously! 😄

You're exactly right, my approach was focused purely on server-level filtering. However, your idea of adding flexibility at the agent-level certainly provides more granular control, which could be beneficial in some scenarios.

I'm not 100% convinced about having both allowed_tools and excluded_tools simultaneously. It may be better to provide a single option, but that's only my opinion.

In any case, let's wait to hear @rm-openai  thoughts, in order to adopt the best approach to move forward.

devtalker:
Hi @DanieleMorotti,

Thanks again for the feedback and being open to chat about it! 😊

I get where you're coming from with `allowed_tools` and `excluded_tools` feeling a bit redundant, but I’ve seen this dual approach used pretty commonly in access control systems — like feature flags or API permissions. It just gives users more flexibility depending on the use case.

For example:

- If you only want a few safe tools, `allowed_tools` is perfect.
- But if most tools are fine and you just want to block a couple risky ones, `excluded_tools` makes life easier.

It’s especially helpful when working with multiple servers or evolving configs over time — no need to rewrite everything just to tighten things up a bit.

Of course, we could go with just one option if needed, but I’d lean toward keeping both unless there's a strong reason not to.

I’ll be looking forward to hearing what @rm-openai thinks too — hopefully they can help us decide the best way forward!

rm-openai:
@DanieleMorotti @devtalker my personal opinion is a mix of your approaches:
1. I prefer restricting it to MCP for now, since thats the source of dynamic tools
2. Would like both allow/block as well as a callable, to allow truly dynamic filtering

devtalker:
@rm-openai Thank you for the feedback! 😊 I agree with your suggestions. I'll update my PR to incorporate these changes.

DanieleMorotti:
I’m closing this PR, even though I opened it first, as @devtalker 's solution appears to be more complete.

Issue #847, opened 2025-06-11
Support automatic "back" handoffs to orchestrating Agents


niv-hertz:
Currently, `handoffs` are a one-way ticket. Once an Agent handed off control to another Agent, it will never regain control. The only way to return control to the previous agent, is to have some circular hand off mechanism (which would require a circular reference between the two agents, which is far from ideal). There are scenarios, where you don't even need to call an LLM to determine that you want the main agent to regain control, assume this architecture:

<img width="603" alt="Image" src="https://github.com/user-attachments/assets/14e9dbe1-32ee-4689-812a-7b88573a736d" />

A simple user question may require more than one agent to run, Ideally, you'll want Agent A. to do the work, go back to the main agent, and then call Agent B. The current solution to this, is to use Agent A and Agent B as tools, but that's not ideal for a couple of reasons (Streaming is not easily supported, context is limited, cannot limit the max turns). 

We can generate a similar architecture with Handoffs, by implementing an optional `back_handoff` parameter, that will result in an automatic Handoff back to the main agent, whenever one of the sub-agents complete their task.

niv-hertz:
@rm-openai  

I'll be happy to implement this, I thought about adding an optional parameter to the Handoff definition, that controls the `back_handoff` logic. 

The implementation itself will be in the `Runner` - once an handoff with `back_handoff` definition is invoked, it will retain a reference to the previous agent (as a stack). Once a `NextStepFinalOutput` output item, is returned by `_run_single_turn`, check the stack to see if any handoff with `back_handoff` was performed, if so, hand it back to that agent instead of ending the loop.

rm-openai:
Following up from the PR, this is still on my radar - I'm hoping to get to it soon! Apologies for the delay.

norika1207-lab:
One practical way to make this easier to evaluate is to turn the report into a small contract test around the runtime boundary, not only the user-facing symptom.

For `automatic back handoffs to orchestrating agents`, I would check three things: first, the failing input is preserved as a minimal fixture; second, the SDK emits the same observable event/result shape before and after the fix except for the intended change; third, the trace or serialized session contains enough metadata to explain what happened after a resume or retry.

The part that matters for agent systems is durability: if a tool call, memory write, handoff, or structured-output step fails, downstream code should not have to infer state from natural-language output. A focused regression test around that boundary would make the issue much easier to close with confidence.

Issue #858, opened 2025-06-12
Mixture of Tool-Calling and Handoff


DykeF:
<html><head></head><body><h3>Feature Request: “Mixture of Tool-Calling and Handoff” </h3>
<p><strong>Is your feature request related to a problem? Please describe.</strong><br>
I have a setup with 1 main agent and multiple tools (which can be other agents, or deterministic agent workflows)</p>
<p><strong>Note:</strong> The teacher agent situation below is just an example. I think there are many use cases, where there should be one main agent, aggregating information from multiple tools (some of them which could be other specialized agents and others search-tools etc) where these tools have access to the full previous history.
<ul>
<li>
<p><strong>Main agent</strong> – runs on a light-weight/cheap model, drives the dialogue.</p>
</li>
<li>
<p><strong>Teacher agent</strong> – runs on <code inline="">o3</code>, aggregates external knowledge, does heavy reasoning.</p>
</li>
</ul>
<p>I’d like to expose the <em>entire</em> conversation context (all user/assistant messages <strong>and</strong> prior tool results generated in the current run) to the Teacher <strong>while still treating it as a function-tool call</strong> so that control returns to the main agent afterwards.</p>
<p>Today I have two partial options:</p>

Option | Drawback
-- | --
Handoff | Teacher gets full history ✔️ …but it takes over the thread, so the main agent never speaks again.
Function tool | Main agent continues ✔️ …yet the Teacher only sees the synthetic input the first agent passes, forcing my small model to re-summarise / concatenate everything (token heavy, lossy).

<p>I think it would be great to have something in between where tools (which can be agents, something different or deterministic workflows of agents) have access to the current history of the "main" agent.</p>

<p>Thanks for considering!</p></body></html>

rm-openai:
Makes sense. discussing in #869

seratch:
Thank you so much again for sharing this idea. We don't plan to add this at least in the short term, so let us close it now.

PR #869, opened 2025-06-15
Added support for "return" handoffs (#1)

niv-hertz:
Related to issue: https://github.com/openai/openai-agents-python/issues/847

### DIFF
--- a/src/agents/_run_impl.py
+++ b/src/agents/_run_impl.py
@@ -177,6 +177,11 @@ class NextStepHandoff:
 
 
 @dataclass
+class NextStepHandoffReturnControl:
+    previous_agent: Agent[Any]
+
+
+@dataclass
 class NextStepFinalOutput:
     output: Any
 
@@ -201,7 +206,9 @@ class SingleStepResult:
     new_step_items: list[RunItem]
     """Items generated during this current step."""
 
-    next_step: NextStepHandoff | NextStepFinalOutput | NextStepRunAgain
+    next_step: (
+        NextStepHandoff | NextStepFinalOutput | NextStepRunAgain | NextStepHandoffReturnControl
+    )
     """The next step to take."""
 
     @property
@@ -238,6 +245,7 @@ class RunImpl:
         hooks: RunHooks[TContext],
         context_wrapper: RunContextWrapper[TContext],
         run_config: RunConfig,
+        previous_agents: list[Agent],
     ) -> SingleStepResult:
         # Make a copy of the generated items
         pre_step_items = list(pre_step_items)
@@ -286,6 +294,7 @@ class RunImpl:
                 hooks=hooks,
                 context_wrapper=context_wrapper,
                 run_config=run_config,
+                previous_agents=previous_agents,
             )
 
         # Next, we'll check if the tool use should result in a final output
@@ -316,6 +325,7 @@ class RunImpl:
                 final_output=check_tool_use.final_output,
                 hooks=hooks,
                 context_wrapper=context_wrapper,
+                previous_agents=previous_agents,
             )
 
         # Now we can check if the model also produced a final output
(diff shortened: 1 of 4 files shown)


niv-hertz:
Could also solve the problem specified in: https://github.com/openai/openai-agents-python/issues/858

@rm-openai  - Would love your review

rm-openai (changes requested):
I thought about this for a while, and ultimately I think this doesn't belong in handoffs because the definition of a handoff is something that takes over control. Whether it chooses to return control should be up to the new agent.

That said, the problem identified is a real one. I think the right way to do it is via a FunctionTool that also receives the full conversation history. i.e. you do something like
```
@function_tool
def my_function(context, history, ... other args):
```

That function could then use an agent or not, but either way it has access to the conversation history.

Thoughts?

EDIT: Also, this should be really easy now bc of the `ToolContext` you added - can just add a `history` field with all the prev items in there.

niv-hertz:
@rm-openai  - Generally, I agree - this was actually my goto as well. There are, however two gaps with that implementation:
1. This doesn't affect the turns counting. Which you may want to control on a general scope, instead of on a per-function call level.
2. There's no easy support for streaming. One of the huge benefits of handing off, is that streaming is pretty much covered for you.

If you have any good solution for either of these with the function calls \ agent as tool approach, I'd love your input :)

EDIT: I also thought about using the existing handoffs mechanism, and as you suggested, giving the new agent the choice of whether to return control or not. But that seemed unnatural to me for two reasons:
1. The sub-agent is not aware of the wider instructions that the orchestrator has, so the choice of whether to return control or not, is not one it can properly make.
2. The Handoff object requires a reference to the agent (in the `_on_invoke_handoff` function), and the agent also contains a reference to the Handoff object. So if you want two agents to be able to handoff to each other, you're forced to implement some sort of circular reference (or hack around it) which is not good practice.

EDIT 2: Another option is to support something parallel to Handoffs, like `SupervisedHandoff`/`Assignment`/`Sub-Workers`

niv-hertz:
@rm-openai  @akhilsmokie 

bumping this :)

niv-hertz:
@rm-openai @akhilsmokie  - Bumping this :)

sibblegp:
I assumed this was how it worked from the start and was surprised to find it didn't. This is a great addition.

How do we set it up after this is implemented to make sure sub-agents hand back off to the overall agent?

Also, does this slow things down since presumably the top agent is doing some AI API calls of its own? That would of course not be ideal.

rm-openai:
@niv-hertz 
The proposal makes sense and feels important. Do you mind if I take a bit of time and try and implement it myself? I think it's going to have some edge cases and documentation updates and it would overall probably be faster - happy to add you as co-author on the PR and get your feedback before merging.

Also happy to go back-and-forth on this PR, though that may take a little bit longer. LMK what you think!

sibblegp:
@rm-openai I agree this is important. Is there any way I can help?

niv-hertz:
@rm-openai  - I'm definitely happy with you implementing it :)
If you need any helping hand, i'll obviously be happy to assist!

niv-hertz:
@rm-openai  - I'll close this PR for now, please let me know if you need any assistance or want to go back-and-forth on this PR :)

built-by-as:
any movement here?

dhruvrnaik:
Any updates on this feature @rm-openai? 
I think it is easy to get the input_history through context or session object if you define the function tool (instead of Agent.as_tool), but I am wondering what is the easiest way to get the pre_function_call_items ( same thing as pre_handoff_items ) in the function definition?

federicobassetto:
Also asking if there are updates on this front

sibblegp:
So, I learned that the designers of the SDK mean that once an item is handed off, it is handled by that agent and completed by it. They don't mean for it to return back up the stack. You can make other handoffs so as to create a flow, but it's pretty restricted.

Their official guidelines on if you want the AI to pick and chose from a bunch of agents and trade off between them is to use them .as_tool() and list them in tools, not handoffs.

It's a bit confusing but it does work if you use them all as tools instead of handoffs. Then the AI can decide which agent to use, then another, then maybe another, and so forth.

Issue #845, opened 2025-06-11
Inconsistent timeout and sse_read_timeout Types in MCPServerSseParams and MCPServerStreamableHttpParams


yinglj:
# Inconsistent `timeout` and `sse_read_timeout` Types in `MCPServerSseParams` and `MCPServerStreamableHttpParams`

## Description
In the `server.py` file of the `openai-agents-python` project (version 1.9.3), the `timeout` and `sse_read_timeout` parameters in `MCPServerSseParams` and `MCPServerStreamableHttpParams` have inconsistent types, causing confusion and potential runtime errors:

- **`MCPServerSseParams`**:
  - `timeout: NotRequired[float]`
  - `sse_read_timeout: NotRequired[float]`
- **`MCPServerStreamableHttpParams`**:
  - `timeout: NotRequired[timedelta]`
  - `sse_read_timeout: NotRequired[timedelta]`

This inconsistency requires developers to handle different types (`float` vs `timedelta`) when configuring servers, which can lead to errors. For example, passing a `float` timeout to `MCPServerStreamableHttp` causes a `TaskGroup` exception in `streamablehttp_client` (e.g., `unhandled errors in a TaskGroup (1 sub-exception)`), as observed in logs:

```
2025-06-11 19:17:20,478 server.py:132 - mcp_client - [MainThread:5432] - ERROR - Error initializing MCP server: unhandled errors in a TaskGroup (1 sub-exception)
```

The lack of clear error messages from `streamablehttp_client` further complicates debugging.

## Steps to Reproduce
1. Configure an `MCPServerStreamableHttp` with a `float` timeout (e.g., `timeout=30`).
2. Call `server.connect()` in a context like `McpServerMgr.initialize_servers`.
3. Observe a `TaskGroup` exception due to type mismatch in `streamablehttp_client`.

## Expected Behavior
Both `MCPServerSseParams` and `MCPServerStreamableHttpParams` should use the same type for `timeout` and `sse_read_timeout` (preferably `float`, as it’s more intuitive for configuration files). The `streamablehttp_client` should either handle `float` internally or provide clear error messages for type mismatches.

## Proposed Solution
1. Unify the types in `MCPServerSseParams` and `MCPServerStreamableHttpParams` to `float` for `timeout` and `sse_read_timeout`.
2. In `MCPServerStreamableHttp.create_streams`, convert `float` to `timedelta` internally if required by `streamablehttp_client`.
[...]

PR #874, opened 2025-06-16
adopted float instead of timedelta for timeout parameters

DanieleMorotti:
I replaced the `timedelta` parameters for MCP timeouts with `float` values, addressing issue #845 .

Given that the MCP official repository has incorporated these changes in [this PR](https://github.com/modelcontextprotocol/python-sdk/pull/941), updating the MCP version in openai-agents and specifying the timeouts as floats should be enough.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
     "typing-extensions>=4.12.2, <5",
     "requests>=2.0, <3",
     "types-requests>=2.0, <3",
-    "mcp>=1.8.0, <2; python_version >= '3.10'",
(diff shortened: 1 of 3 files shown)


rm-openai:
Hmm this would be a breaking change for existing users right? Do you think we can leave backwards compatibility, is that possible?

DanieleMorotti:
If I'm not wrong, it doesn't seem a breaking change, it’s a type-hinting issue, because `streamablehttp_client` can take parameters of both types as input and it will not fail.

I added both `timedelta` and `float` as possible types, in order to match the `mcp` package’s expected inputs.

rm-openai (approved):
cool thank you

Issue #862, opened 2025-06-13
Best practices for integrating Gradio's Agent and Tool


Sucran:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
Yes
- **Have you searched for related issues?** Others may have had similar requests
No, nothing about gradio framework.

### Describe the feature
What is the feature you're requesting? How would it work? Please provide examples and details if possible.

当我构建一个流式输出的Planer + Multi-Actor的应用基于Agent SDK时，Agent SDK框架内部handoff和工具调用并没有很友好的接口可以获取或自动输出。

需要有一个example，展示如何集成integrating Gradio's Agent and Tool 的卡片输出，像是这个例子https://huggingface.co/spaces/ysharma/Gemini2-Flash-Thinking/blob/main/app.py 会输出 思考过程的卡片一样。

必须承认的是 smolagents + gradio 的集成比较好，但是smolagents是非异步的，而且是codeact，而非loop pattern，很容易就识别不到mcp工具。

Sucran:
I'm woking on this problem, and if I work out, then I will made one pr on this.

rm-openai:
Also related to #864?

Sucran:
@rm-openai  No related to #864, this issue just focus on how to display agent's stream events on Gradio ui, and #864 focus on how to expose an official way (like api of sdk) of streaming tool.

Sucran:
> I'm woking on this problem, and if I work out, then I will made one pr on this.

I'm woking on define a EventHandler in the form of  chain of responsibility design pattern. I think this is the easiest way to distinguish diff events and solve this problem.

Sucran:
> [@rm-openai](https://github.com/rm-openai) No related to [#864](https://github.com/openai/openai-agents-python/issues/864), this issue just focus on how to display agent's stream events on Gradio ui, and [#864](https://github.com/openai/openai-agents-python/issues/864) focus on how to expose an official way (like api of sdk) of streaming tool.

#864 shows nested events is missing because of sub_agents wrap using as_tools. If not wrapped by as_tools, I think it works well.

Issue #855, opened 2025-06-12
AG-UI Support


stefanoamorelli:
I'm using `openai-agents-python` with `CopilotKit` to build agent UIs. Adding AG-UI support makes it easy to connect structured agent output to `CopilotKit`'s `React` components for real-time step tracing, tool inspection, and human-in-the-loop interactions.

Would be happy to contribute to support this.

rm-openai:
Not super familiar with AG-UI. Can you say more about what an integration would look like?

stefanoamorelli:
@rm-openai AG-UI is basically a standardized protocol for AI agents to talk to frontends, instead of every agent framework having its own event format, AG-UI defines 16 common event types that common frontends (such as [Copilot kit](https://github.com/CopilotKit/CopilotKit)) can understand.

Integration would be pretty straightforward, I'd suggest adding an adapter that converts our current streaming events to AG-UI format. It would reside in `agents-extensions/src/ag-ui`.

Happy to draft a PR if this interests you!

Sucran:
@stefanoamorelli @rm-openai Hi, I think we encountered the similar issue here, see #862. I had raise a pr #888, and @stefanoamorelli you can see, I also converts our current streaming events to Gradio ChatMessage, which likely what you said.

But I still confused how we can build an abstraction to adapter different ui framework. 
Also, I face some issues in streaming events conversion:
1. mcp tool call was redirected to tool call, which cannot distinguish mcp or not
2. custom model's reasoning phase (like deepseek-r1) nearly no adaption in both openai lib or litellm lib, and reasoning events are hard to adapt.
3. handoff events are lack off infos to show the reason why other agents need to be handoff.

I'm still learning the source code to understand how to solve this 3 issues.

devv-shayan:
is there any update on the integration?
 @stefanoamorelli @Sucran

Sucran:
@devv-shayan nope, check it out about #888 , seratch says "We focus on demonstrating how to use this SDK's key features in examples/ directory, so I don't think we'll add examples with third-party tools, at least in the short term. ", I think AG-UI or gradio will not support in short term. But you can hack it out by yourself, it's not so difficult.

seratch:
As mentioned in this thread, we don't have plans to add this. While we'll be closing this issue, sharing the demos available in different repos would be greatly appreciated.

PR #888, opened 2025-06-17
add example of gradio in streaming agent with mcp cases

Sucran:
# add example of gradio in streaming agent with mcp cases for issue #862

see issue #862 

This PR commit an example which shows a best practice about using gradio and agents sdk to build a streaming app with mcp tools.

## Edit points

- event_handler.py : an chain of responsibility implemention to process all stream events defined in agents sdk
- main.py : an implementation of gradio app and two agents based on deepseek-v3 and gradio mcp tool
- README.md and demo.png: show how this examples works.
- pyproject.toml : add gradio into dev dependency list and gradio 5.x version need ruff>=0.9.3 and python >= 3.10.
- uv.lock: sync change to pyproject.toml

## Features
1. deepseek model integration examples
2. gradio ChatInterface examples with tool and agent Integrations
3. chain of responsibility implemention examples to process all stream events defined in agents sdk


Sucran:
@rm-openai hi, Can you please help me review the code? Thank you very much

Sucran:
@rm-openai hi，the conflict is solved. I am thinking, is it better to integrate Gradio in the Extension way mentioned in #855? would love to know your opinion.
Then I have considered using RunHooks and AgentHooks to implement it, but I found that it does not support inference events. Do you think Hooks would be better?

Sucran:
@rm-openai can give a review or any advice?

seratch:
We focus on demonstrating how to use this SDK's key features in `examples/` directory, so I don't think we'll add examples with third-party tools, at least in the short term. For now, I'd suggest creating your own repository to host the demo app; that would be appreciated.

Issue #905, opened 2025-06-19
Support Anthropic Prompt Caching


serialx:
Unlike OpenAI APIs, Anthropic models require explicit cache control blocks for the caching to work. And the way they achieve this is by adding an additional key to the ChatCompletionMessageParam like:

```json
    "system": [
      {
        "type": "text",
        "text": "You are an AI assistant tasked with analyzing literary works. Your goal is to provide insightful commentary on themes, characters, and writing style.\n"
      },
      {
        "type": "text",
        "text": "<the entire contents of Pride and Prejudice>",
        "cache_control": {"type": "ephemeral"}
      }
    ],
    "messages": [
      {
        "role": "user",
        "content": "Analyze the major themes in Pride and Prejudice."
      }
    ]
``` 

ref: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching

It seems it is not trivial to add this feature in the codebase since it does not comply with the type system.

hanil-jihun:
I need it, too

serialx:
I'm closing this issue since there is another way to solve this: using custom model
Here's my code if anyone is interested: https://gist.github.com/serialx/31dd60db626b3dcb646dcd8369fd2552

PR #908, opened 2025-06-20
feat: Support Anthropic Claude prompt caching key "cache_control"

serialx:
Anthropic models require explicit cache control blocks for the caching to work. And the way they achieve this is by adding an additional key to the text block like:

```json
    "system": [
      {
        "type": "text",
        "text": "You are an AI assistant tasked with analyzing literary works. Your goal is to provide insightful commentary on themes, characters, and writing style.\n"
      },
      {
        "type": "text",
        "text": "<the entire contents of Pride and Prejudice>",
        "cache_control": {"type": "ephemeral"}
      }
    ],
    "messages": [
      {
        "role": "user",
        "content": "Analyze the major themes in Pride and Prejudice."
      }
    ]
``` 

ref: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching

Since "cache_control" key is not in official OpenAI API specs, there are various type errors I needed to maneuver through. Code is not ideal, but I tried to create a patch with minimal impact.

With this patch, we can inject cache control keys to the input like:

```python
input_items: list[TResponseInputItem] = []
input_items.append({"role": "user", "content": "Hello, world", "cache_control": {"type": "ephemeral"}})
result = await Runner.run(current_agent, input_items)
```

This solves #905

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -290,6 +290,16 @@ class Converter:
             return current_assistant_msg
 
         for item in items:
+            # Anthropic Claude needs explicit prompt cache directive
+            # But "cache_control" key is not supported by OpenAI models
+            # So we need to remove it from the item and add it manually to the message
+            # Check if item is dict-like and try to get cache_control key
+            cache_control = None
+            if hasattr(item, "get") and hasattr(item, "__contains__"):
+                cache_control = item.get("cache_control", None)
+                if "cache_control" in item:
+                    del item["cache_control"]  # type: ignore[typeddict-item]
+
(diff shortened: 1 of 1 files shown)


serialx:
I've solved this with a custom model that inherits `LitellmModel`. Closing because this is a very inelegant approach.

serialx:
Here's my code if anyone is interested: https://gist.github.com/serialx/31dd60db626b3dcb646dcd8369fd2552

ZXTFINAL:
> Here's my code if anyone is interested: https://gist.github.com/serialx/31dd60db626b3dcb646dcd8369fd2552

Thanks! It helps a lot.

PR #922, opened 2025-06-23
Add on_start support to VoiceWorkflowBase and VoicePipeline

### LINKED ISSUE #488: Add Intro message function for VoicePipeline
### Please read this first

### Describe the feature
While working with the VoicePipeline, I encountered a limitation with the current design. 
In the [existing pipeline structure](https://github.com/openai/openai-agents-python/blob/d0693a192e92057bfae7cc9d621a6a68e76a111b/src/agents/voice/pipeline.py#L48), the voice agent only initiates communication **after** audio input is received and transcribed, as evidenced by the design of the `run` function:

```python
class VoicePipeline:
    async def _run_multi_turn(self, audio_input: StreamedAudioInput) -> StreamedAudioResult:
        ...
                    async for input_text in transcription_session.transcribe_turns():
                        result = self.workflow.run(input_text)
```

This implementation requires the workflow.run function to be called only after the transcription of the user's audio input has begun, preventing the voice agent from starting the conversation proactively. 
This design is not ideal for scenarios where the agent should initiate a conversation or deliver an introductory message before any user input is provided.

To address this, I propose adding a dedicated function to the VoicePipeline class, such as:
```python
async def initiate_intro_audio_stream(self, intro_message: str | None = None) -> StreamedAudioResult:
```

This function would allow developers to trigger an initial intro audio stream, enabling the voice agent to communicate with users immediately upon startup. This enhancement would make the VoicePipeline more flexible and user-friendly, particularly for applications that rely on the agent to deliver greetings or instructions proactively.

If this feature already exists or if there's an alternative approach, please let me know.

vrtnis:
I added an optional `on_start()` method to `VoiceWorkflowBase`, allowing voice workflows to emit an initial TTS message before any user input. This aligns with the usage pattern suggested in #488 

The `VoicePipeline._run_multi_turn()` method was updated to call this hook and stream any yielded messages via TTS before transcription begins. No changes are required to existing workflows, as the method defaults to returning an empty `AsyncIterator[str]`.

Resolves #488  and thus improves UX for voice agents that need to greet or instruct users proactively. All tests pass under Python 3.12. Let me know if you'd like an example workflow or docs added in a follow-up!


src/agents/voice/pipeline.py
```diff
@@ -125,6 +125,13 @@ async def _run_multi_turn(self, audio_input: StreamedAudioInput) -> StreamedAudi
                 self._get_tts_model(), self.config.tts_settings, self.config
             )
 
+            if hasattr(self.workflow, "on_start"):
```
rm-openai: instead of hasattr, just invoke the method. It's already implemented as a no-op?

vrtnis: thanks @rm-openai - you are absolutely right. the `hasattr` check isn't necessary. i've removed it and pushed an update

vrtnis:
hey @rm-openai.. wondering if you could take a quick look. i’ve made the requested change... please let me know if there’s anything else needed here prior to merge..

PR #925, opened 2025-06-23
Add is_enabled to handoffs

rm-openai:


pakrym-oai (approved):

maininformer:
Hello, curious as to why this was reverted, is there another way to do this? I.e. how to make the agent would 100% call a tool before handoff?

PR #927, opened 2025-06-24
Enable developers to initialize built-in tools in a simpler way

seratch:
This pull request enhances the tool data classes to enable developers to omit the required "type" property. Passing the property works too, so this does not bring any breaking changes. (for context, I received feedback from a customer that passing the similar name when initializing these classes looks quite redundant)


seratch:
reverted the changes to the examples as it requires skip comment for mypy; there may be something else we can consider. let me close this approach for now

PR #931, opened 2025-06-24
Ensure that input_guardrails can block tools from running

### LINKED ISSUE #889: FileSearchTool runs despite InputGuardrailTripwireTriggered

**Expected behavior:**

rm-openai:
Resolves #889

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -96,6 +96,12 @@ class InputGuardrail(Generic[TContext]):
     function's name.
     """
 
+    block_tool_calls: bool = True
+    """Whether this guardrail should block tool calls until it completes. If any input guardrail
+    has this set to True, tool execution will be delayed until all blocking guardrails finish.
+    Defaults to True for backwards compatibility and safety.
+    """
+
     def get_name(self) -> str:
         if self.name:
             return self.name
@@ -208,6 +214,7 @@ def input_guardrail(
 def input_guardrail(
     *,
     name: str | None = None,
+    block_tool_calls: bool = True,
 ) -> Callable[
     [_InputGuardrailFuncSync[TContext_co] | _InputGuardrailFuncAsync[TContext_co]],
     InputGuardrail[TContext_co],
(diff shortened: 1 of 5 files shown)


seratch:
@rm-openai do you think this approach is still good to go? If not, we may want to think of a solution for #889 and #991

lionelchg:
Hi @seratch, just checked out on the PR and tried the following script (a shorter version of the quickstart script) - it still does not block the call.

- this is the script:
```python
from agents import Agent, InputGuardrail, GuardrailFunctionOutput, Runner
from agents.exceptions import InputGuardrailTripwireTriggered
import asyncio

class HomeworkOutput(BaseModel):
    is_homework: bool
    reasoning: str

guardrail_agent = Agent(
    name="Guardrail check",
    instructions="Check if the user is asking about homework.",
    output_type=HomeworkOutput,
)

math_tutor_agent = Agent(
    name="Math Tutor",
    instructions="You provide help with math problems. Explain your reasoning at each step and include examples",
)

async def homework_guardrail(ctx, agent, input_data):
    result = await Runner.run(guardrail_agent, input_data, context=ctx.context)
    final_output = result.final_output_as(HomeworkOutput)
        output_info=final_output,
        tripwire_triggered=not final_output.is_homework,
    )

triage_agent = Agent(
    name="Triage Agent",
    instructions="You determine which agent to use based on the user's homework question",
    handoffs=[math_tutor_agent],
    input_guardrails=[
        InputGuardrail(guardrail_function=homework_guardrail),
    ],
)

async def main():
    # Example: General/philosophical question
    try:
        result = await Runner.run(triage_agent, "What is the meaning of life?")
        print(result.final_output)
        print("Guardrail blocked this input:", e)

if __name__ == "__main__":
    asyncio.run(main())
```

- this is my screenshot of the logs (you can see that the second call passes even though `is_homework` is False):
Screenshot 2025-08-30 at 01 01 04

I might try a fix this week-end will keep you posted!

illia-botscrew:
@rm-openai do you plan to finish this support of blocking tools from running when input_guardrail is processing the input?

This PR and the other PR that was trying to add support for this: https://github.com/openai/openai-agents-python/pull/1622 are closed.

PR #935, opened 2025-06-25
Fix #604 Chat Completion model raises runtime error when response.choices is empty

### LINKED ISSUE #604: [Bug]: SDK crashes when `choices` is `None` (provider-error payload)
### Please read this first

- **Have you read the docs?**  [Agents SDK docs](https://openai.github.io/openai-agents-python/) [**Yes**]
- **Have you searched for related issues?**  [**Yes, see the following**]
- #380 

---

### Describe the bug

When the upstream provider responds with an *error* payload (`{"error": …}`), the Agents SDK still returns a `ChatCompletion`-shaped object whose `choices` field is `None`.  

`openai_chatcompletions.py` immediately dereferences `response.choices[0]` inside its debug logger, producing

```
TypeError: 'NoneType' object is not subscriptable
```

The exception is raised **inside** the SDK, before the calling application sees the provider error.

---

### Debug information

| Item | Value |
|------|-------|
| Agents SDK version | v0.0.13 |
| Python version | 3.12.9 |
| OS | macOS 15.4 (arm64) |

---

### Repro steps

```python
# mockup_main.py
from agents import Runner
from agents.models import OpenAIChatCompletionsModel

# A key / proxy that will reliably return {"error": …}
    model="gpt-3.5-turbo",
    api_key="sk-dummy",
    base_url="https://openrouter.ai/v1",
)

runner = Runner(model=model)

# ⇢ Raises TypeError in the SDK before we can handle the error payload
runner.run("Ping?")
```

Run:

```bash
export AGENTS_LOGGING_LEVEL=DEBUG   # optional, shows failing log line
python repro.py
```

**Observed output**

```
TypeError: 'NoneType' object is not subscriptable
```

(full traceback shows the line in `openai_chatcompletions.py` that
logs `response.choices[0]`).

---

### Expected behavior

The SDK should detect that `response.choices` is missing and raise a clear domain-specific exception (e.g. `ProviderError`) containing the provider’s error message, instead of crashing with an internal `TypeError`.  

All downstream code paths would then have a chance to handle the error or retry gracefully.

seratch:
This pull request resolves #604; just checking the existence of the data and avoiding the runtime exception should make sense.

### DIFF
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -91,7 +91,7 @@ class OpenAIChatCompletionsModel(Model):
                     )
                 else:
                     finish_reason = first_choice.finish_reason if first_choice else "-"
-                    logger.debug("LLM resp had no message. finish_reason: %s", finish_reason)
+                    logger.debug(f"LLM resp had no message. finish_reason: {finish_reason}")
 
             usage = (
                 Usage(
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -90,10 +90,8 @@ class OpenAIChatCompletionsModel(Model):
                         json.dumps(message.model_dump(), indent=2),
                     )
                 else:
-                    logger.debug(
-                        "LLM resp had no message. finish_reason: %s",
-                        first_choice.finish_reason,
-                    )
+                    finish_reason = first_choice.finish_reason if first_choice else "-"
+                    logger.debug("LLM resp had no message. finish_reason: %s", finish_reason)
 
             usage = (
                 Usage(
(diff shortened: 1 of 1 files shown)


seratch:
now this is ready for reviews

src/agents/models/openai_chatcompletions.py
```diff
@@ -86,10 +90,8 @@ async def get_response(
                         json.dumps(message.model_dump(), indent=2),
                     )
                 else:
-                    logger.debug(
-                        "LLM resp had no message. finish_reason: %s",
-                        first_choice.finish_reason,
-                    )
+                    finish_reason = first_choice.finish_reason if first_choice else "-"
+                    logger.debug("LLM resp had no message. finish_reason: %s", finish_reason)
```
rm-openai: better to use f-strings, can you switch to that and then merge?

PR #938, opened 2025-06-25
Annotating the openai.Omit type so that ModelSettings can be serialized by pydantic

tconley1428:
Because `openai.Omit` is not a type which can be serialized by `pydantic`, it produces difficulty in consistently serializing/deserializing types across the agents SDK, many of which are `pydantic` types. This adds an annotation to enable `pydantic` to serialize `Omit`, in particular in the case of `ModelSettings` which contains `Omit` in its `extra_headers`

### DIFF
--- a/src/agents/model_settings.py
+++ b/src/agents/model_settings.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import dataclasses
 from collections.abc import Mapping
 from dataclasses import dataclass, fields, replace
-from typing import Annotated, Any, Literal
+from typing import Annotated, Any, Literal, Union
 
 from openai import Omit as _Omit
 from openai._types import Body, Query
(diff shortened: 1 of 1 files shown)


tconley1428:
Turns out that while this supports creating the core schema, it doesn't yet actually allow serialization of an `Omit`

seratch (approved):
LGTM

src/agents/model_settings.py
```diff
@@ -1,14 +1,50 @@
...
+                    core_schema.is_instance_schema(_Omit),
+                    from_none_schema,
+                ]
+            ),
+            serialization=core_schema.plain_serializer_function_ser_schema(
+                lambda instance: None
+            ),
+        )
+Omit = Annotated[_Omit, _OmitTypeAnnotation]
+Headers: TypeAlias = Mapping[str, Union[str, Omit]]
```
rm-openai, line 47: @tconley1428 can you confirm that requests to the openAI API still work when headers are set like this?

tconley1428: I can confirm that our normal requests work, but I don't I know if they actually have additional headers (I think a user agent is added by default?), and I'm pretty sure they don't actually have an instance of `Omit`

PR #943, opened 2025-06-26
logger support utf8 on DEBUG level

EeyoreLee:
Some of my input is in Chinese, which makes the verbose log unreadable when it's enabled. I just want to add support for this part.


seratch:
Thank you so much for sending this, and I do understand this could be inconvenient. We already received the same contribution https://github.com/openai/openai-agents-python/pull/639 and it will be merged sometime soon. So, let me close this one.  Thanks again for taking the time for this!

PR #950, opened 2025-06-27
Fix #944 by updating the models document page to cover extra_args

### LINKED ISSUE #944: Add "user" property from Reponses API for caching/tracking and abuse issues
OpenAI docks state:

> A stable identifier for your end-users. Used to boost cache hit rates by better bucketing similar requests and to help OpenAI detect and prevent abuse.

https://platform.openai.com/docs/api-reference/responses/object#responses/object-user
https://platform.openai.com/docs/guides/safety-best-practices#end-user-ids

seratch:
This pull request resolves #944


docs/ja/models.md
```diff
@@ -1,106 +0,0 @@
-# モデル
```
seratch, line 1: this file is no longer used

docs/models/index.md
```diff
@@ -93,6 +93,22 @@ english_agent = Agent(
 )
 ```
 
+Also, when you use OpenAI's Responses API, [there are a few other optional parameters](https://platform.openai.com/docs/api-reference/responses/create) (e.g., `user`, `service_tier`, and so on). If they are not available at the top level, you can use `extra_args` to pass them as well.
```
seratch, line 96: These lines are the essential changes in this PR

seratch (commented):
left two comments for reviewers

PR #953, opened 2025-06-27
Import Path Inconsistency

DanielHashmi:
There's an inconsistent import style in the tracing module where one import uses an absolute path in line number 3 while all others use relative imports


seratch:
The change looks fine but the lint job is failing

PR #958, opened 2025-06-27
Fix incorrect argument description in `on_trace_end` docstring

shirazkk:
The docstring for the `on_trace_end` method in the `TracingProcessor` class incorrectly describes the argument `trace`. The current description states that the argument is the trace that started, but it should refer to the trace that has finished. This commit updates the docstring to clarify that the argument represents the trace that has completed, ensuring accuracy and clarity.

Changes:
- Updated the argument description for `trace` in `on_trace_end` to clarify it refers to the trace that finished.


seratch (approved):
LGTM

PR #971, opened 2025-06-30
Add logprobs to ModelSettings

zain:
Now that logprobs are available in the Responses API, I thought it might be nice to have it be available for agents.


docs/ja/models/index.md
```diff
@@ -114,6 +114,20 @@ english_agent = Agent(
 )
 ```
 
+Responses API でトークンの対数確率を取得したい場合は、
```
seratch: We generate translated docs using script, so please revert this change.

docs/models/index.md
```diff
@@ -109,6 +109,20 @@ english_agent = Agent(
 )
 ```
 
+You can also request token log probabilities when using the Responses API by
```
seratch: Once this property is available at the top level, it will be listed [here](https://openai.github.io/openai-agents-python/ref/model_settings/#agents.model_settings.ModelSettings), which is linked from this page. So, could you revert this one too?

src/agents/model_settings.py
```diff
@@ -107,6 +108,10 @@ class ModelSettings:
     """Additional output data to include in the model response.
     [include parameter](https://platform.openai.com/docs/api-reference/responses/create#responses-create-include)"""
 
+    top_logprobs: int | None = None
```
seratch, line 120: this would be used in not only responses, but LiteLLM and Chat Completions, so can you update the following files as well?
- src/agents/models/openai_chatcompletions.py
- src/agents/extensions/models/litellm_model.py

seratch (changes requested):

zain:
@seratch all done!

seratch (approved):
LGTM

Issue #831, opened 2025-06-09
separate stream items for "tool_call_item" and "tool_call_output_item"


gdisk:
In the following example, the tool call event and the tool call output event are triggered simultaneously,and I believe it's more reasonable to separate them.

### test code

```python
import logging

logging.basicConfig(
    format="%(asctime)s.%(msecs)03d - %(levelname)s - %(message)s",
    datefmt="%Y-%m-%d %H:%M:%S",
)
logging.getLogger(__name__).setLevel(logging.DEBUG)
logger = logging.getLogger(__name__)

from agents import Runner, Agent, ItemHelpers, function_tool
import asyncio
import random


@function_tool
def how_many_jokes() -> int:
    await asyncio.sleep(3)
    return random.randint(1, 2)


async def main():
    agent = Agent(
        name="Joker",
        instructions="First call the `how_many_jokes` tool, then tell that many jokes.",
        tools=[how_many_jokes],
    )

    result = Runner.run_streamed(
        agent,
        input="Hello",
    )
    logger.debug("=== Run starting ===")
    async for event in result.stream_events():
        # We'll ignore the raw responses event deltas
        if event.type == "raw_response_event":
            continue
        elif event.type == "agent_updated_stream_event":
            logger.debug(f"Agent updated: {event.new_agent.name}")
        elif event.type == "run_item_stream_event":
            if event.item.type == "tool_call_item":
                logger.debug("-- Tool was called")
            elif event.item.type == "tool_call_output_item":
                logger.debug(f"-- Tool output: {event.item.output}")
            elif event.item.type == "message_output_item":
                logger.debug(
                    f"-- Message output:\n {ItemHelpers.text_message_output(event.item)}"
                )

    logger.debug("=== Run complete ===")


if __name__ == "__main__":
    asyncio.run(main())

```

### current output

```text
2025-06-09 17:02:38.588 - DEBUG - === Run starting ===
2025-06-09 17:02:38.588 - DEBUG - Agent updated: Joker
2025-06-09 17:02:39.770 - DEBUG - -- Tool was called
2025-06-09 17:02:39.770 - DEBUG - -- Tool output: 1
2025-06-09 17:02:40.788 - DEBUG - -- Message output:
 Here's a joke for you:

Why don't skeletons fight each other?

They don't have the guts!
2025-06-09 17:02:40.788 - DEBUG - === Run complete ===
```
[...]

PR #974, opened 2025-06-30
Feat: separate tool_call_item and tool_call_output_item in stream events

gdisk:
### Summary
During my analysis of the streaming API (run_streamed) results, I noticed that tool_call_item and tool_call_output_item events are currently being emitted concurrently upon tool call completion (as evidenced in [ #831](https://github.com/openai/openai-agents-python/issues/831)). This implementation conflates what should logically be distinct events. The current PR addresses this by properly separating these event triggers to better reflect the actual workflow.
### Test plan
The test file is created in `tests` named `test_stream_events.py`.Run the test script below to test.

```bash
pytest -s test_stream_events.py
```

The test result is:

```text
======================================================================== test session starts ======================================================================== 
platform win32 -- Python 3.12.10, pytest-8.3.5, pluggy-1.5.0
rootdir: D:\moon\projects\openai-agents-python
configfile: pyproject.toml
plugins: anyio-4.9.0, inline-snapshot-0.22.3, asyncio-0.26.0, mock-3.14.0
asyncio: mode=Mode.AUTO, asyncio_default_fixture_loop_scope=session, asyncio_default_test_loop_scope=function
collected 1 item                                                                                                                                                      

tests\test_stream_events.py === Run starting ===
Agent updated: Joker
-- Message output:
 a_message
-- Tool was called at 1751271106971851300
-- Tool output: success! at 1751271109987313900
-- Message output:
 done
=== Run complete ===
.

========================================================================= 1 passed in 3.07s =========================================================================
```

### Issue number

[ #831](https://github.com/openai/openai-agents-python/issues/831)

### Checks


legeRise:
hi i was just trying to get an effect of  calling a tool  with its name displays with a spinner and then after sometime the output 
and
noticed both events are emitting at the same time

so is it fixed Now? can i use it

gdisk:
> hi i was just trying to get an effect of calling a tool with its name displays with a spinner and then after sometime the output and noticed both events are emitting at the same time
> 
> so is it fixed Now? can i use it

Yeah, just check out this PR and run the tests to verify.

legeRise:
> > hi i was just trying to get an effect of calling a tool with its name displays with a spinner and then after sometime the output and noticed both events are emitting at the same time
> > so is it fixed Now? can i use it
> 
> Yeah, just check out this PR and run the tests to verify.

thank you so much.  it works. hope it gets merged soon

sunilpandey5:
Waiting for it to get merged soon

mattmorgis:
@gdisk thank you for this PR 🙏 hoping they merge soon 

cc/ @rm-openai in case you haven't seen yet.

tests/test_stream_events.py
```diff
@@ -0,0 +1,66 @@
...
+            # Second turn: text message
+            [get_text_message("done")],
+        ]
+    )
+
+    result = Runner.run_streamed(
+        agent,
+        input="Hello",
+    )
+    print("=== Run starting ===")
```
rm-openai: can you get rid of these `print`s plz

tests/test_stream_events.py
```diff
@@ -0,0 +1,66 @@
...
+            elif event.item.type == "tool_call_output_item":
+                tool_call_end_time = time.time_ns()
+                print(f"-- Tool output: {event.item.output} at {tool_call_end_time}")
+            elif event.item.type == "message_output_item":
+                print(
+                    f"-- Message output:\n {ItemHelpers.text_message_output(event.item)}"
+                )
+
+    print("=== Run complete ===")
+    assert tool_call_start_time > 0, "tool_call_item was not observed"
```
rm-openai, line 51: might be better to instead capture all the events and do a snapshot() to ensure they are in the right order + params

tests/test_stream_events.py
gdisk: > can you get rid of these `print`s plz

Sorry, i don`t understand what does the `rid` mean for the `prints`

tests/test_stream_events.py
gdisk: > might be better to instead capture all the events and do a snapshot() to ensure they are in the right order + params

The order of these events is correct, but `tool_call_item` is triggered immediately after `tool_call_output_item` (upon tool call completion), not at the start of the tool call.

tests/test_stream_events.py
seratch: It should mean removing these print statements in this test code.

rm-openai (approved):

PR #977, opened 2025-06-30
Fix #976 MCP filtering, make agent and run ctx optional

### LINKED ISSUE #976: [BUG]: MCP Server unecessarily requires use of Agent and RunContext
### Describe the bug

Found while working on https://github.com/mozilla-ai/any-agent/issues/543

PR https://github.com/openai/openai-agents-python/pull/861 added support for tool filtering. I'm in favor of a tool filtering feature being added, but the way it was implemented now requires that you access any tools from an MCP after creating the Agent, which is an unnecessary new design restriction.

Our library already had logic to filter tools, 

which first listed the tools https://github.com/mozilla-ai/any-agent/blob/main/src/any_agent/tools/mcp/frameworks/openai.py#L47

and then filtered them https://github.com/mozilla-ai/any-agent/blob/main/src/any_agent/tools/mcp/mcp_connection.py#L40.

This was all done before passing the list of tools to the Agent. 

### Debug information
- Agents SDK version: 0.1.0
- Python version 3.13

### Repro steps

MCP list_tools now requires a run context and an agent. 

### Expected behavior

I'd prefer if this filtering feature was implemented without the change to the parameters passed into list_tools. 

However, at a minimum (and maybe at least for the short term) could they be optional parameters since they're only used when using the built in openai filtering logic?

njbrake:
This is a patch to address the most pressing issue of #976 while we discuss whether or not these parameters should exist in `list_tools` or should be migrated elsewhere.

Resolves #976

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -52,8 +52,8 @@ class MCPServer(abc.ABC):
     @abc.abstractmethod
     async def list_tools(
         self,
-        run_context: RunContextWrapper[Any],
-        agent: Agent[Any],
+        run_context: RunContextWrapper[Any] | None = None,
+        agent: Agent[Any] | None = None,
     ) -> list[MCPTool]:
         """List the tools available on the server."""
         pass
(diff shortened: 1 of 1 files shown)


seratch (approved):
LGTM

seratch:
@rh-openai This one is reasonable and looks good to me, but can you take a look before merging this, just in case?

wjykim:
There's something wrong since this code was updated. Static tool filtering does not work (Please pay attention to the situation of function **create_static_tool_filter**).

wjykim:
> There's something wrong since this code was updated. Static tool filtering does not work (Please pay attention to the situation of function **create_static_tool_filter**).

```
if run_context is None or agent is None:
    raise UserError("run_context and agent are required for dynamic tool filtering")
```
 Since run_context and agent can be None, why add the exception here? It may stop the process of static tool filtering.

PR #997, opened 2025-07-02
Support OPENAI_BASE_URL in tracing

tabletcorry:
Agent tracing was hardcoded against the API endpoint, so any client using a proxy or alternate route to the OpenAI AI endpoint would encounter errors.

So this modify the tracing init code to use `OPENAI_BASE_URL` when present, if the local environment hasn't overridden anything.

Note: Because this code path is generally executed at import time, the environment variable needs to be set prior to importing any openai-agents code.


tabletcorry:
Nod, I searched around a fair bit to see if OpenAI had any variable or name for the tracing endpoint, and couldn't find one, so I decided to piggyback instead.

Given the lasting nature of an env var name like this, it seems like something y'all should define, rather than me.

tabletcorry:
Also since the token is shared for API access and tracing access, I am not sure how a proxy would work that doesn't support both.

It would be somewhat odd to give someone a real OpenAI token, then tell them to use a proxy with a different token instead.

seratch:
@rm-openai Do you think `OPENAI_TRACES_INGEST_ENDPOINT` env variable (or anything like that) can work? Or do prefer asking devs to instantiate `TracingExporter` with the endpoint string in code? I don't have strong opinion here.

rm-openai:
Yeah let's add a URL method similar to `set_tracing_export_api_key()`.

PR #1002, opened 2025-07-03
fix(hooks): handle kwargs in hook calls

Quratulain-bilal:
🔧 Pull Request: Consistent Callback Signatures Using **kwargs
📌 Summary
This PR aims to improve consistency and usability in the OpenAI Agents SDK's hook system by making all hook callbacks accept **kwargs instead of a mix of positional and keyword arguments.

✅ Problem
Currently, the SDK defines five lifecycle hooks:

on_agent_start

on_agent_end

on_tool_start

on_tool_end

on_handoff

However, only on_handoff uses **kwargs, while the others rely on positional arguments. This inconsistency creates two major issues:

1. Readability and Developer Experience
Developers subclassing AgentHooks or RunnerHooks have to remember the exact order of positional arguments, which is error-prone and less intuitive.

2. Maintainability and Flexibility
Using **kwargs allows for more flexible hook definitions in the future without breaking existing subclasses.

✅ Solution
This PR refactors all hook callback method signatures to accept **kwargs, following the pattern already used by on_handoff. Example before and after:

# Before (current SDK)
hooks.on_agent_start(context_wrapper, agent)

# After (this PR)
hooks.on_agent_start(context=context_wrapper, agent=agent)

Note: The internal calling code (in _run_impl.py, run.py, etc.) already passes named arguments, so this change does not affect functionality—only the developer interface.

✅ Benefits
Improved consistency across all hooks

Easier to implement custom hooks

Cleaner and more maintainable API

Prevents potential bugs due to wrong argument order

🧪 Tests
✅ Manual testing was done by subclassing both AgentHooks and RunnerHooks and overriding all lifecycle methods using **kwargs.

✅ No breaking change to core logic or execution flow.

🧠 Notes
This change does not impact the internal logic or performance but focuses on developer ergonomics and interface consistency, which is crucial for open-source adoption and contribution.

### DIFF
--- a/src/agents/_run_impl.py
+++ b/src/agents/_run_impl.py
@@ -553,9 +553,9 @@ class RunImpl:
                     span_fn.span_data.input = tool_call.arguments
                 try:
                     _, _, result = await asyncio.gather(
-                        hooks.on_tool_start(tool_context, agent, func_tool),
+                        hooks.on_tool_start(context=tool_context, agent=agent, tool=func_tool),
                         (
-                            agent.hooks.on_tool_start(tool_context, agent, func_tool)
+                            agent.hooks.on_tool_start(context=tool_context, agent=agent, tool=func_tool)
                             if agent.hooks
                             else _coro.noop_coroutine()
                         ),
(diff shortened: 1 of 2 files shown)


seratch:
It seems the CI builds are failing

seratch (changes requested):
the test failures need to be resolved first

Awais68:
outstanding changes

PR #1003, opened 2025-07-03
Fix: Ensure output_guardrail provides function name instead of defaulting to None

romankhan26:
This PR improves the output_guardrail feature by ensuring that a valid function name is always provided. Previously, if the name attribute was missing, it defaulted to None. This change introduces a fallback name to avoid unexpected behavior or errors in downstream processes.

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -310,7 +310,8 @@ def output_guardrail(
     def decorator(
         f: _OutputGuardrailFuncSync[TContext_co] | _OutputGuardrailFuncAsync[TContext_co],
     ) -> OutputGuardrail[TContext_co]:
-        return OutputGuardrail(guardrail_function=f, name=name)
+        #Updating default name None to the function name of guardrail_function if no name is provided.
+        return OutputGuardrail(guardrail_function=f, name=name if name else f.__name__)
 
     if func is not None:
         # Decorator was used without parentheses


rm-openai (approved):

romankhan26:
@rm-openai 
If we create output_guardrail using the decorator without parentheses, it does not assign a name to the guardrail based on the function's name. Instead, the name remains None until the end (guardrail_results). However, if we use parentheses in the decorator and explicitly assign a name, it sets the name as expected. This code ensures that if we do not explicitly define the name in the decorator, the function name will be used as the guardrail name instead of leaving it as None.

Moreover, The thing you mentioned about OutputGuardrail.get_name() doing the same is true, but the issue is that get_name() is not used in the decorator. As a result, the name remains None until we explicitly define it, like @output_guardrail(name='guardrail_name').

rm-openai:
That makes sense, but what's the underlying problem that causes?

romankhan26:
I was not getting the name of the guardrails in my result.input_guardrail_results

seratch:
Can you fix the lint error if you have time for it?

```
Run make lint
uv run ruff check
src/agents/guardrail.py:313:101: E501 Line too long (102 > 100)
    |
311 |         f: _OutputGuardrailFuncSync[TContext_co] | _OutputGuardrailFuncAsync[TContext_co],
312 |     ) -> OutputGuardrail[TContext_co]:
313 |         #Updating default name None to the function name of guardrail_function if no name is provided.
    |                                                                                                     ^^ E501
314 |         return OutputGuardrail(guardrail_function=f, name=name if name else f.__name__)
    |

Found 1 error.
make: *** [Makefile:[12](https://github.com/openai/openai-agents-python/actions/runs/16059046857/job/45559636014?pr=1003#step:5:13): lint] Error 1
```

romankhan26:
@seratch 
Fixed it!

Issue #1011, opened 2025-07-05
Add list_prompts / get_prompt to MCP server implementations


amri369:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have had similar requests

### Describe the feature
The existing implementation of _MCPServerWithClientSession doesn't support prompts fetching from servers. This feature is needed to access to MCP Servers Prompts using Openai Agents SDK MCP Servers implementation.

amri369:
https://github.com/openai/openai-agents-python/pull/1013

PR #1010, opened 2025-07-05
Support mcp prompts

thoo:
### Summary

Add MCP prompt support to enable user-controlled agent instruction generation. This enhancement allows MCP servers to provide prompts that can be used to dynamically generate agent instructions, enabling more flexible and user-controlled agent behavior.

**Key changes:**
- Added abstract `list_prompts()` and `get_prompt()` methods to base `MCPServer` class
- Implemented prompt support in `MCPServerStdio` and `MCPServerStreamableHttp` classes
- Added comprehensive test suite for prompt functionality
- Created example MCP prompt server with working demonstration
- Updated documentation with prompt usage examples

**Note:** This PR implements MCP prompt support only. Resource support is not included in this implementation.

### Test plan

- **Unit tests**: Added 11 comprehensive tests in `tests/mcp/test_prompt_server.py` covering:
  - Prompt listing and retrieval
  - Argument formatting and validation
  - Agent integration with prompt-generated instructions
  - Streaming and non-streaming scenarios
  - Error handling for missing prompts
- **Integration tests**: Updated existing MCP test suite to handle new abstract methods
- **Example verification**: Created working example with MCP prompt server and client
- **All tests pass**: 450/450 tests passing after adding optional dependencies

### Issue number

Partially addresses #544 (prompts only, resources not implemented)

### Checks

- [x] I've added new tests (if relevant)
- [x] I've added/updated the relevant documentation
- [x] I've run `make lint` and `make format`
- [x] I've made sure tests pass

PR #1013, opened 2025-07-05
Add get_prompt and list_prompts to MCPServer

amri369:
Enable prompts listing and fetching from MCP servers.
Issue #1011

### DIFF
--- a/examples/mcp/streamablehttp_example/main.py
+++ b/examples/mcp/streamablehttp_example/main.py
@@ -46,6 +46,12 @@ async def main():
     ) as server:
         trace_id = gen_trace_id()
         with trace(workflow_name="Streamable HTTP Example", trace_id=trace_id):
+            # List available prompts
+            prompts = await server.list_prompts()
+            print(f"Prompts list -> {prompts}")
+            system_prompt = await server.get_prompt("system_prompt")
+            instructions = system_prompt.messages[0].content.text
+            print(f"instructions -> {instructions}")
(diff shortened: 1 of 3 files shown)


docs/scripts/translate_docs.py
```diff
@@ -266,7 +266,9 @@ def translate_single_source_file(file_path: str) -> None:
 
 def main():
     parser = argparse.ArgumentParser(description="Translate documentation files")
-    parser.add_argument("--file", type=str, help="Specific file to translate (relative to docs directory)")
+    parser.add_argument(
```
seratch: overall, this pull request looks good to me but minimizing unrelated changes like this one would be appreciated

seratch (changes requested):

seratch:
@rh-openai overall, this pull request looks good to me, but let me know if you have some thoughts on this feature addition.

docs/scripts/translate_docs.py
amri369: @seratch : Many thanks for the feedback. I removed the unnecessary changes as requested.

seratch:
#1010 is aiming to resolve the same issue and the PR has unit tests and document changes. If you don't mind, could you collaborate with the PR author to discuss the best solution for this? I don't have further time to be online today but am happy to check again later this week.

rm-openai:
I took a look at #1010 and that seemed good so merged it - lmk if there are further additions you need!

PR #1033, opened 2025-07-08
docs: mention 'name' is required in Agent constructor . fix: mention 'name' as required in Agent init doc

maneeshanif:
## What

Updated the docs under `docs/agents.md` to clarify that the `name` argument is required when initializing an `Agent`.

## Why

Without this clarification, users get a confusing error:
`TypeError: Agent.__init__() missing 1 required positional argument: 'name'`

This matches the actual constructor in the source code and helps future devs avoid frustration.

## Reference

- Source: [`agent.py`](https://github.com/openai/openai-agents-python/blob/main/src/agents/agent.py)

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -6,6 +6,7 @@ Agents are the core building block in your apps. An agent is a large language mo
 
 The most common properties of an agent you'll configure are:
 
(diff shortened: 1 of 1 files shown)


seratch (approved):
LGTM

Issue #992, opened 2025-07-02
FunctionTool ignores strict_json_schema parameter


habema:
### Describe the bug
The documentation for creating custom `FunctionTool` objects directly (without using the `@function_tool` decorator) shows an example that fails with OpenAI's Responses API due to missing `additionalProperties: false` in the JSON schema.

The example from the docs:
```python
tool = FunctionTool(
    name="process_user",
    description="Processes extracted user data",
    params_json_schema=FunctionArgs.model_json_schema(),  # This line is the problem
    on_invoke_tool=run_function,
)
```

This results in the error:
```
Error code: 400 - {'error': {'message': "Invalid schema for function 'process_user': In context=(), 'additionalProperties' is required to be supplied and to be false.", 'type': 'invalid_request_error', 'param': 'tools[0].parameters', 'code': 'invalid_function_parameters'}}
```

### Debug information
- Agents SDK version: 0.1.0
- Python version: 3.10.16

### Repro steps

```python
from typing import Any
from pydantic import BaseModel
from agents import RunContextWrapper, FunctionTool, Agent, Runner

class FunctionArgs(BaseModel):
    username: str
    age: int

async def run_function(ctx: RunContextWrapper[Any], args: str) -> str:
    parsed = FunctionArgs.model_validate_json(args)
    return f"{parsed.username} is {parsed.age} years old"

# This fails with OpenAI 400 error
tool = FunctionTool(
    name="process_user",
    description="Processes extracted user data",
    params_json_schema=FunctionArgs.model_json_schema(),  # Missing strict schema enforcement
    on_invoke_tool=run_function,
)

agent = Agent(
    name="Test Agent",
    instructions="You are a test agent",
    tools=[tool]
)

# This will fail when the agent tries to use the tool
result = Runner.run_sync(agent, "Process user data for John who is 30 years old")
```

**Working fix (not documented):**
```python
from agents.strict_schema import ensure_strict_json_schema

tool = FunctionTool(
    name="process_user", 
    description="Processes extracted user data",
    params_json_schema=ensure_strict_json_schema(FunctionArgs.model_json_schema()),  # Fix
    on_invoke_tool=run_function,
)
```

### Expected behavior
Either:
1. The documentation example should work as-is, OR
2. The documentation should clearly show that `ensure_strict_json_schema()` must be called on Pydantic schemas when creating `FunctionTool` objects directly, OR  
[...]

PR #1041, opened 2025-07-08
fix: Apply strict JSON schema validation in FunctionTool constructor

habema:
## Summary
Fixes an issue where directly created `FunctionTool` objects fail with OpenAI's Responses API due to missing `additionalProperties: false` in the JSON schema, while the `@function_tool` decorator works correctly.

## Problem
The documentation example for creating `FunctionTool` objects directly fails with:
```
Error code: 400 - {'error': {'message': "Invalid schema for function 'process_user': In context=(), 'additionalProperties' is required to be supplied and to be false.", 'type': 'invalid_request_error', 'param': 'tools[0].parameters', 'code': 'invalid_function_parameters'}}
```

This creates an inconsistency between `FunctionTool` and `@function_tool` behavior, both of which have `strict_json_schema=True` by default.

## Solution
- Added `__post_init__` method to `FunctionTool` dataclass
- Automatically applies `ensure_strict_json_schema()` when `strict_json_schema=True`
- Makes behavior consistent with `@function_tool` decorator
- Maintains backward compatibility

## Testing
The fix can be verified by running the reproduction case from the issue:

```python
from typing import Any
from agents import RunContextWrapper, FunctionTool, Agent, Runner

class FunctionArgs(BaseModel):
    username: str
    age: int

async def run_function(ctx: RunContextWrapper[Any], args: str) -> str:
    parsed = FunctionArgs.model_validate_json(args)
    return f"{parsed.username} is {parsed.age} years old"

# This now works without manual ensure_strict_json_schema() call
tool = FunctionTool(
    name="process_user",
    description="Processes extracted user data",
    params_json_schema=FunctionArgs.model_json_schema(),
    on_invoke_tool=run_function,
)

agent = Agent(
    name="Test Agent",
    instructions="You are a test agent",
    tools=[tool]
)

result = Runner.run_sync(agent, "Process user data for John who is 30 years old")
```

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -24,6 +24,7 @@ from .function_schema import DocstringStyle, function_schema
 from .items import RunItem
 from .logger import logger
 from .run_context import RunContextWrapper
+from .strict_schema import ensure_strict_json_schema
 from .tool_context import ToolContext
 from .tracing import SpanError
 from .util import _error_tracing
@@ -92,6 +93,10 @@ class FunctionTool:
     """Whether the tool is enabled. Either a bool or a Callable that takes the run context and agent
     and returns whether the tool is enabled. You can use this to dynamically enable/disable a tool
     based on your context/state."""
+
+    def __post_init__(self):
+        if self.strict_json_schema:
+            self.params_json_schema = ensure_strict_json_schema(self.params_json_schema)
 
 
 @dataclass


habema:
Linked Issue: https://github.com/openai/openai-agents-python/issues/992

seratch (approved):
LGTM

rm-openai (approved):

salah9003:
من هنا إلى الأعلى 👍🏿 👍🏿

3marwanmoraly3:
From here to the top 👍🏿 👍🏿 👍🏿

PR #1045, opened 2025-07-09
Introduce RealtimeAgent

rm-openai:
---
* #1046
* __->__ #1045
* #1044


sibblegp:
Just wanted to say thank you for getting this done. It's already so close and the progress every day is amazing.

Also commenting so I can become a participant in this PR.

PR #1048, opened 2025-07-10
Add automation script to generate mkdocstrings files

seratch:
This pull request adds a new script for docs, which generates missing `ref/**/*.md` files. The script can be executed when you run `make build-docs` command. The script does not do:
- overwrite the existing ones
- create files for _XXX.py and `__init__.py`

Note that the title part is generated like `tool_context` to `Tool Context`. `openai_provider` will be `Openai Provider`, so some of them needs a little manual work to adjust.

The direct need is to add `tool_context.md` for https://github.com/openai/openai-agents-python/pull/1043 but it should be useful for future maintenance.


rm-openai (changes requested):
Hey @seratch the code looks totally fine but I actually don't get why we need this? Why can't the files just be hand-written, takes like 5 seconds lol

seratch:
@rm-openai indeed, creating a single file takes only a few seconds, but checking which files are missing (like this PR does) could take much longer and automation script can do better job.

waveformer1984 (commented):
yes across all projects

PR #1050, opened 2025-07-10
Add Runtime Validation for `name` Attribute in `Agent` Class

EngineerAbdullahIqbal:
This pull request proposes adding runtime validation to the `Agent` class in the `openai-agents-sdk` to ensure the `name` attribute is a non-empty string. This change addresses the issue where non-string types (e.g., integers or dictionaries) are currently accepted, despite the `str` type annotation, which can lead to unexpected behavior. By implementing validation, we make the SDK more robust and user-friendly.

### Proposed Change
The proposed solution adds a `__post_init__` method to the `Agent` class to validate that `name` is a string and not empty or whitespace-only. This approach is standard for Python dataclasses and does not affect other attributes or methods.

### Why It Matters
This change prevents potential errors, aligns with type annotations, and improves the user experience by providing clear feedback on invalid inputs. It benefits the community by making the SDK more reliable, encouraging adoption, and reducing support overhead.

The `__post_init__` method is the most suitable, as it is lightweight, aligns with Python dataclass conventions, and requires minimal changes to the existing codebase.

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -79,7 +79,11 @@ class Agent(Generic[TContext]):
     """
 
     name: str
-    """The name of the agent. Must be a non-empty string."""
+    """The name of the agent. Must be a non-empty string.
+    Raises:
+         TypeError: if not a string
+         ValueError: if empty or only whitespace
+    """
 
     instructions: (
         str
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this PR! but we're already working on this at https://github.com/openai/openai-agents-python/pull/998; please feel free to add comments if you have any.

PR #1052, opened 2025-07-10
Add streaming function call arguments example

devtalker:
This PR adds a streaming function call arguments example that demonstrates real-time incremental display of function parameters during generation, providing immediate feedback to users. Please merge this if you find this example useful.


examples/basic/stream_function_call_args.py
```diff
@@ -0,0 +1,86 @@
...
+                    }
+                    current_active_call_id = call_id
+                    print(f"\n📞 Function call streaming started: {function_name}()")
+                    print("📝 Arguments building...")
+
+            # Real-time argument streaming
+            elif isinstance(event.data, ResponseFunctionCallArgumentsDeltaEvent):
+                if current_active_call_id and current_active_call_id in function_calls:
+                    function_calls[current_active_call_id]['arguments'] += event.data.delta
+                    print(f"   + {event.data.delta}", end="", flush=True)
```
seratch: I prefer a simpler output here:
```suggestion
                    print(event.data.delta, end="", flush=True)
```
What do you think?

seratch (changes requested):

examples/basic/stream_function_call_args.py
devtalker: Okay, thank you for your review. I accept it

PR #1054, opened 2025-07-10
docs : added missing await in run_streamed example

mutahirshah11:
This PR fixes a small issue in the docs. The example using Runner.run_streamed was missing the await keyword, which would lead to a runtime error. I've added the missing await in both streaming code examples.

### DIFF
--- a/docs/streaming.md
+++ b/docs/streaming.md
@@ -21,7 +21,7 @@ async def main():
         instructions="You are a helpful assistant.",
     )
 
-    result = Runner.run_streamed(agent, input="Please tell me 5 jokes.")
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this, but I don't think this change is necessary.

Muhammad-Arham-MERN:
> This PR fixes a small issue in the docs. The example using Runner.run_streamed was missing the await keyword, which would lead to a runtime error. I've added the missing await in both streaming code examples.

Not needed because streaming is not meant to be awaited, its purpose diminishes

seratch:
Let me close this now, but please feel free to write in whenever you have follow-ups!

PR #1065, opened 2025-07-11
Naming_logic_issue_found

hussain9491:
i found naming issue in session docs about  pop_item (work_flow)

assistant_item replace to user_item
user_item replace to assistant_item


seratch:
Thanks for the suggestion, but I don't think your change is a valid code snippet.

PR #1066, opened 2025-07-11
Replaced dummy example function with meaningful user data handler

mutahirshah11:
**Summary**

This PR updates the `FunctionTool` example by replacing the placeholder `do_some_work()` function with a more practical version, `process_user_data()`. The new function showcases how structured input (`username`, `age`) is validated and returned, making the example clearer and more useful for developers.

**Changes Made**

- Replaced the dummy function `do_some_work()` with `process_user_data()`
- Updated input schema class to use realistic field names (`username`, `age`)
- Preserved the structure and logic of the original FunctionTool example

**Why This Matters**

The current example in the SDK returns a static string and doesn't utilize the input fields, which can be confusing for new developers. This updated version provides a more realistic demonstration of tool invocation and structured input processing — while keeping the example minimal and educational.

### DIFF
--- a/docs/streaming.md
+++ b/docs/streaming.md
@@ -21,7 +21,7 @@ async def main():
         instructions="You are a helpful assistant.",
     )
 
-    result = Runner.run_streamed(agent, input="Please tell me 5 jokes.")
+    result = await Runner.run_streamed(agent, input="Please tell me 5 jokes.")
     async for event in result.stream_events():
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the suggestions, but as I mentioned at https://github.com/openai/openai-agents-python/pull/1054, the change is not necessary. Also, the code snippet changes in the documentation could be fine but we may do something like that in a more consistent way in the future. Thanks again for sending this.

PR #1068, opened 2025-07-11
[1/n] Break Agent into AgentBase+Agent

rm-openai:
This allows base agent stuff to be shared with the realtime agent

---
* #1069
* __->__ #1068

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -67,7 +67,63 @@ class MCPConfig(TypedDict):
 
 
 @dataclass
-class Agent(Generic[TContext]):
+class AgentBase(Generic[TContext]):
+    """Base class for `Agent` and `RealtimeAgent`."""
+
+    name: str
+    """The name of the agent."""
+
+    handoff_description: str | None = None
+    """A description of the agent. This is used when the agent is used as a handoff, so that an
+    LLM knows what it does and when to invoke it.
+    """
+
+    tools: list[Tool] = field(default_factory=list)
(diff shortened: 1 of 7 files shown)


src/agents/lifecycle.py
```diff
@@ -28,16 +30,16 @@ async def on_agent_end(
     async def on_handoff(
         self,
         context: RunContextWrapper[TContext],
-        from_agent: Agent[TContext],
-        to_agent: Agent[TContext],
+        from_agent: TAgent,
```
pakrym-oai, line 33: will this cause type checking errors for customers that relied on the type of context?

pakrym-oai: I guess not if TAgent is well typed

src/agents/agent.py
```diff
@@ -67,7 +67,63 @@ class MCPConfig(TypedDict):
...
+
+    NOTE: You are expected to manage the lifecycle of these servers. Specifically, you must call
+    `server.connect()` before passing it to the agent, and `server.cleanup()` when the server is no
+    longer needed.
+    """
+
+    mcp_config: MCPConfig = field(default_factory=lambda: MCPConfig())
+    """Configuration for MCP servers."""
+
+    async def get_mcp_tools(self, run_context: RunContextWrapper[TContext]) -> list[Tool]:
```
pakrym-oai, line 97: uses context generic type var but the type is not generic itself.

pakrym-oai (approved):

PR #1070, opened 2025-07-11
[3/n] Config and items for realtime

rm-openai:


src/agents/realtime/config.py
```diff
@@ -0,0 +1,91 @@
...
+    create_response: NotRequired[bool]
+    eagerness: NotRequired[Literal["auto", "low", "medium", "high"]]
+    interrupt_response: NotRequired[bool]
+    prefix_padding_ms: NotRequired[int]
+    silence_duration_ms: NotRequired[int]
+    threshold: NotRequired[float]
+
+
+class RealtimeSessionConfig(TypedDict):
+    api_key: NotRequired[APIKeyOrKeyFunc]
```
pakrym-oai, line 59: this doesn't align with RunConfig where API key is property of the model/model provider

rm-openai: it's optional, its also a property of the transport (equialent to model provider) here

src/agents/realtime/config.py
```diff
@@ -0,0 +1,91 @@
...
+    type: NotRequired[Literal["semantic_vad", "server_vad"]]
+    create_response: NotRequired[bool]
+    eagerness: NotRequired[Literal["auto", "low", "medium", "high"]]
+    interrupt_response: NotRequired[bool]
+    prefix_padding_ms: NotRequired[int]
+    silence_duration_ms: NotRequired[int]
+    threshold: NotRequired[float]
+
+
+class RealtimeSessionConfig(TypedDict):
```
pakrym-oai, line 58: Should this be named model config to mirror non realtime?

rm-openai: no its not a model config. But you're right that I should separate the model specific configs and the other stuff i plan to add to this (e.g. tracing)

PR #1072, opened 2025-07-11
[5/n] OpenAI realtime transport impl

rm-openai:
Uses the openai realtime impl over websockets. Unlike the TS version, only supports websockets - no in browser stuff.

---
* #1074
* #1073
* __->__ #1072
* #1071

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -1,0 +1,353 @@
+import asyncio
+import base64
+import json
+import os
+from datetime import datetime
+from typing import Any
+
+import websockets
+from openai.types.beta.realtime.realtime_server_event import (
+    RealtimeServerEvent as OpenAIRealtimeServerEvent,
+)
+from pydantic import TypeAdapter
+from websockets.asyncio.client import ClientConnection
+
+from ..exceptions import UserError
+from ..logger import logger
+from .config import RealtimeClientMessage, RealtimeUserInput, get_api_key
+from .items import RealtimeMessageItem, RealtimeToolCallItem
+from .transport import (
+    RealtimeSessionTransport,
+    RealtimeTransportConnectionOptions,
+    RealtimeTransportListener,
+)
+from .transport_events import (
+    RealtimeTransportAudioDoneEvent,
+    RealtimeTransportAudioEvent,
+    RealtimeTransportAudioInterruptedEvent,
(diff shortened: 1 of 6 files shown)


src/agents/realtime/transport.py
```diff
@@ -0,0 +1,107 @@
...
+
+    api_key: NotRequired[APIKeyOrKeyFunc]
+    """The API key to use for the transport. If unset, the transport will attempt to use the
+    `OPENAI_API_KEY` environment variable.
+    """
+
+    model: NotRequired[str]
+    """The model to use."""
+
+    url: NotRequired[str]
```
pakrym-oai, line 43: This seems a bit too specific to OpenAI to be a part of abstraction. Or should be documented differently.

src/agents/realtime/transport.py
```diff
@@ -0,0 +1,107 @@
...
+        """Add a listener to the transport."""
+        pass
+
+    @abc.abstractmethod
+    async def remove_listener(self, listener: RealtimeTransportListener) -> None:
+        """Remove a listener from the transport."""
+        pass
+
+    @abc.abstractmethod
+    async def send_event(self, event: RealtimeClientMessage) -> None:
```
pakrym-oai, line 70: Why is this method public? I assume the message is very specific to the underlying transport.

src/agents/realtime/openai_realtime.py
```diff
@@ -0,0 +1,353 @@
...
+            await self.send_event({"type": "response.create"})
+
+    async def interrupt(self) -> None:
+        """Interrupt the model."""
+        if not self._current_item_id or not self._audio_start_time:
+            return
+
+        await self._cancel_response()
+
+        elapsed_time_ms = (datetime.now() - self._audio_start_time).total_seconds() * 1000
```
pakrym-oai, line 194: Why round to seconds here?

pakrym-oai (approved):

src/agents/realtime/openai_realtime.py
rm-openai: it's not rounding. python mostly uses seconds for everything (it's stored as a float not an int so doesnt lose precision)

src/agents/realtime/transport.py
rm-openai: so that you can send events that aren't necessarily public. Helps with custom implementations, as well as with our own realtime evolving quickly

PR #1073, opened 2025-07-11
[6/n] RealtimeSession + events

rm-openai:
RealtimeSession is like Runner, except for Realtime. It's also stateful and manages the conversation for you.

---
* #1076
* #1074
* __->__ #1073
* #1072
* #1071

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -1,0 +1,369 @@
+"""Minimal realtime session implementation for voice agents."""
+
+from __future__ import annotations
+
+import abc
+import asyncio
+from collections.abc import Awaitable
+from typing import Any, Callable, Literal
+
+from typing_extensions import TypeAlias, assert_never
+
+from ..handoffs import Handoff
+from ..run_context import RunContextWrapper
+from ..tool import FunctionTool
+from ..tool_context import ToolContext
+from .agent import RealtimeAgent
+from .config import APIKeyOrKeyFunc, RealtimeSessionConfig, RealtimeUserInput
+from .events import (
+    RealtimeAgentEndEvent,
+    RealtimeAgentStartEvent,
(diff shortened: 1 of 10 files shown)


src/agents/realtime/session.py
```diff
@@ -0,0 +1,369 @@
...
+    RealtimeTransportInputAudioTranscriptionCompletedEvent,
+    RealtimeTransportToolCallEvent,
+)
+
+
+class RealtimeSessionListener(abc.ABC):
+    """A listener for realtime session events."""
+
+    @abc.abstractmethod
+    async def on_event(self, event: RealtimeSessionEvent) -> None:
```
pakrym-oai, line 53: why callback and not a streaming interface like in Runner?

src/agents/realtime/session.py
```diff
@@ -0,0 +1,369 @@
...
+    your own custom transport layer by implementing the `RealtimeSessionTransport` interface.
+    """
+
+    def __init__(
+        self,
+        starting_agent: RealtimeAgent,
+        *,
+        context: Any | None = None,
+        transport: Literal["websocket"] | RealtimeSessionTransport = "websocket",
+        api_key: APIKeyOrKeyFunc | None = None,
```
pakrym-oai, line 91: I'm surprised the transport is top level and separated from the model. 

If we'd like to support 3rd party models they'll have to come with whatever transport they support.

pakrym-oai (approved):

src/agents/realtime/session.py
rm-openai: yeah this is bad, i will change

src/agents/realtime/session.py
rm-openai: yes true, combining them into model

PR #1076, opened 2025-07-11
[8/n] Make realtime more like the rest of agents sdk

rm-openai:
Key changes:
1. Transport -> model.
2. Extract any model settings into `RealtimeSessionModelSettings`.
3. RealtimeRunConfig, similar to the RunConfig in `run.py`.
4. RealtimeRunner now exists, similar to Runner. Returns a RealtimeSession when you call run().
5. RealtimeSession now uses streaming events instead of listener.

---
* #1080
* #1079
* __->__ #1076


examples/realtime/demo.py
```diff
@@ -38,23 +40,34 @@ def get_weather(city: str) -> str:
...
     async def run(self) -> None:
-        self.session.add_listener(self.on_event)
-        await self.session.connect()
-        self.ui.set_is_connected(True)
-        await self.ui.run_async()
+        # Start UI in a separate task instead of waiting for it to complete
+        ui_task = asyncio.create_task(self.ui.run_async())
+
+        # Set up session immediately without waiting for UI to finish
+        runner = RealtimeRunner(agent)
```
pakrym-oai, line 56: the normal runner takes the agent as a parameter to run, why use constructor for realtime?

src/agents/realtime/model.py
```diff
@@ -0,0 +1,97 @@
...
+        """Add a listener to the model."""
+        pass
+
+    @abc.abstractmethod
+    def remove_listener(self, listener: RealtimeModelListener) -> None:
+        """Remove a listener from the model."""
+        pass
+
+    @abc.abstractmethod
+    async def send_event(self, event: RealtimeClientMessage) -> None:
```
pakrym-oai, line 62: can we make this private?

src/agents/realtime/__init__.py
```diff
@@ -15,26 +27,33 @@
...
-    RealtimeModelName,
-    RealtimeSessionTransport,
-    RealtimeTransportConnectionOptions,
-    RealtimeTransportListener,
+from .model import (
+    RealtimeModel,
+    RealtimeModelConfig,
+    RealtimeModelListener,
 )
+from .runner import RealtimeRunner
```
pakrym-oai, line 34: from .session import RealtimeRunner

src/agents/realtime/session.py
```diff
@@ -86,16 +42,12 @@ def __init__(
         self,
         starting_agent: RealtimeAgent,
         *,
-        context: Any | None = None,
-        transport: Literal["websocket"] | RealtimeSessionTransport = "websocket",
-        api_key: APIKeyOrKeyFunc | None = None,
+        context: TContext | None = None,
+        transport: Literal["websocket"] | RealtimeModel = "websocket",
```
pakrym-oai: can we nuke the transport as well?

examples/realtime/demo.py
rm-openai: A common use case is that your Runner is the same, but you want to launch many different long-lived connections with the same config. This lets you do that without needing to re-specify the agent each time.

src/agents/realtime/model.py
rm-openai: Specifically wanted to make this public to allow arbitrary events

examples/realtime/demo.py
pakrym-oai: But that's true even with a normal runner/agent. Do you think this use case is worth the inconsistency?

pakrym-oai (approved):

examples/realtime/demo.py
rm-openai: Yeah i think it'll be super common with realtime. regular Runner is static so this wasn't really a consideration there

PR #1080, opened 2025-07-12
Tests for realtime session

rm-openai:


seratch:
once the mypy failure is resolved, this is good to go

PR #1086, opened 2025-07-12
Fix incorrect comment order when popping session items

MuhammadHassaanArain:
The previous comments incorrectly suggested that the user's question was removed before the assistant's response. This fix updates the comment order to reflect the actual sequence of the code execution.


seratch (approved):
LGTM

MuhammadHassaanArain:
Thanks @seratch, Glad it looks good.

PR #1089, opened 2025-07-12
Add input validation to prevent empty user messages in translation example

mshsheikh:
Problem:
The script allows empty or whitespace-only input from the user, which results in unnecessary agent executions and invalid translations.

Changes:
- Added input validation to check if the user input is empty or whitespace-only.
- If invalid, raises `ValueError` with a clear message: `"Input message cannot be empty."`
- Prevents downstream agent execution when input is invalid.

```python
msg = input("Hi! Enter a message, and we'll translate it to Spanish.\n\n")
if not msg.strip():
    raise ValueError("Input message cannot be empty.")
```

Benefits:
- Ensures only valid, non-empty input triggers agent workflows.
- Improves user experience by failing early with a descriptive error.
- Avoids unnecessary API calls or agent runs for invalid inputs.

### DIFF
--- a/examples/agent_patterns/parallelization.py
+++ b/examples/agent_patterns/parallelization.py
@@ -20,6 +20,8 @@ translation_picker = Agent(
 
 async def main():
     msg = input("Hi! Enter a message, and we'll translate it to Spanish.\n\n")
+    if not msg.strip():
+        raise ValueError("Input message cannot be empty.")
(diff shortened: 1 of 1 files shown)


seratch:
This may be a good change, but skipping the empty input and waiting for the user's meaningful input might be even more useful.

PR #1092, opened 2025-07-13
feat(function_schema): add enforce_type_annotations flag for stricter schema validation with type enforcement, clearer errors, and updated docstrings

mshsheikh:
Problem:
Previously, the `function_schema` function silently defaulted unannotated parameters to `Any`, leading to:
- Ambiguous JSON schemas (`"type": "any"`)
- Potential runtime errors due to invalid inputs
- Reduced reliability of LLM tool calls This made it harder to ensure type safety without manual validation.

Solution:

This PR introduces the `enforce_type_annotations` flag to `function_schema`, allowing developers to:
1. Enforce type annotations by setting `enforce_type_annotations=True`, which raises a `ValueError` for unannotated parameters.
2. Maintain backward compatibility by defaulting to `Any` when `enforce_type_annotations=False` (default).

Example error message:
```python
raise ValueError(
    f"Parameter '{name}' must be type-annotated. Example: def func({name}: str)"
)
```

Changes Made

1. New Parameter:
```python
   def function_schema(
       func: Callable[..., Any],
       enforce_type_annotations: bool = False,  # New parameter
       ...
   ):
```

2. Validation Logic:
- If `enforce_type_annotations=True`, unannotated parameters now raise a `ValueError` with a clear example.
- Falls back to `Any` if disabled (default behavior preserved).

3. Docstring Updates:
- Added detailed documentation for `enforce_type_annotations`:
```python
     Args:
         enforce_type_annotations: If True, raises a ValueError for any unannotated parameters.
             If False (default), unannotated parameters are assumed to be of type `Any`.
```

4. Error Message Improvements:
- Clear guidance for developers: `Example: def func(x: str)`

Backward Compatibility:
- Preserved: Existing code continues to work as-is (default `enforce_type_annotations=False`).
- Opt-in: Type annotation enforcement is optional, allowing gradual adoption.

Testing Guidance:
To verify the change:
```python
def test_enforce_type_annotations():
    def func(x): ...
    with pytest.raises(ValueError):
        function_schema(func, enforce_type_annotations=True)
    # Should not raise
    function_schema(func, enforce_type_annotations=False)
```

Impact:
- Type Safety: Prevents ambiguous schemas and improves LLM tool reliability.
- Developer Experience: Clear error messages guide users to fix missing annotations.
- Flexibility: Maintains backward compatibility while enabling stricter validation.

Example Usage:
```python
# Strict mode: raises error for unannotated params
schema = function_schema(my_func, enforce_type_annotations=True)

# Default mode: works as before
schema = function_schema(my_func)  # silently defaults to Any
```

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -186,6 +186,7 @@ def generate_func_documentation(
 
 def function_schema(
     func: Callable[..., Any],
+    enforce_type_annotations: bool = False,
     docstring_style: DocstringStyle | None = None,
     name_override: str | None = None,
     description_override: str | None = None,
@@ -198,6 +199,8 @@ def function_schema(
 
     Args:
         func: The function to extract the schema from.
+        enforce_type_annotations: If True, raises a ValueError for any unannotated parameters.
+            If False (default), unannotated parameters are assumed to be of type `Any`.
         docstring_style: The style of the docstring to use for parsing. If not provided, we will
             attempt to auto-detect the style.
         name_override: If provided, use this name instead of the function's `__name__`.
@@ -266,9 +269,14 @@ def function_schema(
         ann = type_hints.get(name, param.annotation)
         default = param.default
 
-        # If there's no type hint, assume `Any`
+        # Raise an error for unannotated parameters if enforcement is on
         if ann == inspect._empty:
-            ann = Any
+            if enforce_type_annotations:
+                raise ValueError(
+                    f"Parameter '{name}' must be type-annotated. Example: def func({name}: str)"
+                )
+            else:
+                ann = Any  # Fallback only if enforcement is off
 
         # If a docstring param description exists, use it
         field_description = param_descs.get(name, None)
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -186,12 +186,12 @@ def generate_func_documentation(
 
 def function_schema(
     func: Callable[..., Any],
-    enforce_type_annotations: bool = False,
     docstring_style: DocstringStyle | None = None,
     name_override: str | None = None,
     description_override: str | None = None,
     use_docstring_info: bool = True,
     strict_json_schema: bool = True,
+    enforce_type_annotations: bool = False,
 ) -> FuncSchema:
     """
     Given a python function, extracts a `FuncSchema` from it, capturing the name, description,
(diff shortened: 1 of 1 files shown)


src/agents/function_schema.py
```diff
@@ -186,6 +186,7 @@ def generate_func_documentation(
 
 def function_schema(
     func: Callable[..., Any],
+    enforce_type_annotations: bool = False,
```
seratch: adding this option may be a good one, but adding here (=2nd arg) is a breaking change. new arguments must be added as the last one when a method accepts both args and keyword args.

seratch (changes requested):
having unit tests would be appreciated too

src/agents/function_schema.py
mshsheikh: Thanks for the feedback! You're absolutely right, the enforce_type_annotations parameter has been moved to the end of the argument list in the latest update to preserve backward compatibility.

mshsheikh:
> having unit tests would be appreciated too

@seratch
Unit tests have been added to validate the `enforce_type_annotations` behavior, including:
- Enforcement of type hints (`True`/`False` cases)
- Fallback to `Any` when disabled
- Backward compatibility with positional args  

Tests are here:
- Test PR: https://github.com/openai/openai-agents-python/pull/1109
- Test file: https://github.com/openai/openai-agents-python/blob/main/tests/test_function_schema.py
Let me know if you’d like further refinements!

mshsheikh:
> My general stance is to be careful about adding new options/settings bc each one makes it that much harder to understand the code. In this case, my thought is -- not using types is your choice, why would someone not use type annotations but also use enforce_type_annotations?

Thank you for the thoughtful feedback! The enforce_type_annotations flag is designed as an opt-in tool for teams gradually adopting type safety, not a default requirement. This allows:
- Progressive enforcement : New code can opt into stricter validation without breaking legacy functions.
- Explicit error prevention : Catches unannotated parameters early, reducing ambiguous schemas and runtime issues.
The default (False) preserves backward compatibility, minimizing overhead for existing users.
Would this approach address your concerns while balancing flexibility and safety?

mshsheikh:
Thanks for the thorough feedback @seratch and @rm-openai!
The latest changes address the positional argument issue and include comprehensive tests. I’ve also revised the flag’s documentation to clarify its purpose as an opt-in tool for stricter validation. Let me know if further adjustments are needed!

**Key Updates:**
- `enforce_type_annotations` added as the last parameter to avoid breaking existing usage.
- Tests validate enforcement, fallback behavior, and backward compatibility.
- Error messages include actionable examples for developers.

This balances flexibility and safety while respecting existing workflows. Let me know your thoughts!

PR #1100, opened 2025-07-14
Update routing.py

SheikhMuhammadHamza-HS:


seratch:
Thanks for sending this! but I personally don't see the necessity of this change. You can cancel the script by ctrl+c etc.

PR #1101, opened 2025-07-14
Correct error message for invalid output types with strict JSON schema

DanielHashmi:
### Problem
1. **Incorrect parameter reference**: Suggests passing `output_schema_strict=False` to `Agent()`, but this parameter doesn't exist
2. **Vague guidance**: The phrase "make the output type strict" doesn't explain HOW to make types strict

### Solution
Updated the error message to provide accurate, actionable guidance:

**Before:**
```
"Either make the output type strict, or pass output_schema_strict=False to your Agent()"
```

**After:**
```
"Either use a dataclass, Pydantic model, or TypedDict, or wrap your type with AgentOutputSchema(your_type, strict_json_schema=False)"
```

### DIFF
--- a/src/agents/agent_output.py
+++ b/src/agents/agent_output.py
@@ -114,9 +114,9 @@ class AgentOutputSchema(AgentOutputSchemaBase):
                 self._output_schema = ensure_strict_json_schema(self._output_schema)
             except UserError as e:
                 raise UserError(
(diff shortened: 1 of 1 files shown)


DanielHashmi:
@rm-openai Updated to only include AgentOutputSchema stuff 👍

PR #1105, opened 2025-07-14
Back out "enhancement: Add tool_name to ToolContext to support shared tool handlers (#1043)"

rm-openai:
Original commit changeset: befe19db273a

causing test failures, so reverting


rm-openai:
cc @vrtnis

vrtnis:
👍 thanks @rm-openai.. i’ve opened PR #1110 as a corrected follow-up with the fixes included. all tests are passing now.

PR #1109, opened 2025-07-14
test(function_schema): add tests for enforce_type_annotations behavior

mshsheikh:
Adds 4 new tests to validate enforce_type_annotations:

- Enforcement: Checks if unannotated functions raise errors.
- Fallback: Confirms unannotated parameters default to Any when disabled.
- Compatibility: Ensures positional args (e.g., function_schema(my_func, "sphinx")) still work.

See Updated Code:
https://github.com/openai/openai-agents-python/blob/main/src/agents/function_schema.py


seratch:
Could you add these tests to #1092 instead? Thank you!

PR #1117, opened 2025-07-15
 add numpy dev dependency

jhills20:


seratch:
Hi @jhills20, this project already has numpy [here](https://github.com/openai/openai-agents-python/blob/v0.1.0/pyproject.toml#L37), and it is installed when you run [`make sync`](https://github.com/openai/openai-agents-python/blob/v0.1.0/Makefile#L2-L3). Do you see a different necessity to have it in dev dependencies too?

rm-openai (approved):

jhills20:
hey @seratch its because we need it for realtime as well, the numpy dependency is only for the `voice` section but its needed for `realtime` as well. plus its not actually needed for the project dependencies but just for the examples

seratch:
ok, thanks for the reply!

PR #1119, opened 2025-07-15
Realtime: fix item parsing

rm-openai:


pakrym-oai (approved):

sibblegp:
Okay, I looked at the demo. Looks fantastic. However, I only see how to capture audio, send it in, process it, and return it.

What about streaming audio like from Twilio that uses server side VAD? Is that supported yet?

PR #1138, opened 2025-07-15
chore(translate): use Chat Completions API for typing compliance

mshsheikh:
This update replaces all `openai_client.responses.create(...)` invocations with `openai_client.chat.completions.create(...)` calls, leveraging the fully‑typed Chat Completions endpoint introduced in `openai>=1.96.0`.

- No changes were made to chunking, code‑block handling, or translation logic.
- Only the request surface has been updated to satisfy mypy/type‑stub requirements.
- Dependencies remain untouched, and the script continues to function identically while passing CI type checks.

### DIFF
--- a/docs/scripts/translate_docs.py
+++ b/docs/scripts/translate_docs.py
@@ -21,7 +21,6 @@ search:
 ---
 """
 
-
 # Define the source and target directories
 source_dir = "docs"
 languages = {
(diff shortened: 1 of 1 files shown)


seratch:
thanks for sending this but we prefer using responses api over chat completions.

PR #1144, opened 2025-07-16
Fix: Validate Agent name to reject empty string and None

Hafiz-Muhammad-Umar12:
This pull request adds strict validation to the Agent class to ensure the name parameter meets the expected requirements:

If name is set to None, an exception (ValueError) is now raised.

If an empty string ("") or a whitespace-only string is passed, a ValueError is also raised.

Although type hints specify that name should be of type str, previously the constructor allowed invalid values such as None or empty strings at runtime, leading to potential issues during agent registration, logging, and debugging.

With this fix:

Agents must always have a meaningful, non-empty name.

Prevents silent errors and ensures better reliability and consistency in multi-agent workflows.

Improves clarity during logging, debugging, and runtime identification of agents.

This validation strictly enforces correct usage, reducing potential misuse of the SDK.

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -67,7 +67,67 @@ class MCPConfig(TypedDict):
 
 
 @dataclass
+@dataclass
 class AgentBase(Generic[TContext]):
+    """Base class for `Agent` and `RealtimeAgent`."""
+
+    name: str
+    """The name of the agent."""
+
+    handoff_description: str | None = None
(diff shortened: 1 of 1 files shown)


seratch:
Thank you so much for sending this! but we've been already working with a different contributor for this at https://github.com/openai/openai-agents-python/pull/998; sharing your insights on the PR reviews instead would be appreciated. Thanks again for your interest and time!

PR #1149, opened 2025-07-16
Realtime: web demo

rm-openai:
Add a web demo that uses a fastapi backend. Also remove the CLI UI demo.


sibblegp:
Nice. However, when I use it, none of the tool outputs are factored into the LLM response. For instance, when I ask for a stock price, it calls the tool properly, the logs show the output properly, but then it gives me some random number as the price.

My tool outputs are dicts though so maybe that's where the issue is.

PR #1154, opened 2025-07-16
fix(logging): include system prompt in debug logs in case of response…

HafizFasih:
Previously, the debug logs for LLM calls only showed user prompts while using responses api,
which made it harder to trace full context during issues. This
change prepends the system prompt to the logged input messages,
without mutating the original input list.

The system prompt is included conditionally (only if present),
and is used for logging only.

This improves debuggability and traceability of agent interactions.

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -242,6 +242,7 @@ class OpenAIResponsesModel(Model):
             else NOT_GIVEN
         )
 
+        system_instructions = self._non_null_or_not_given(system_instructions)
         tool_choice = Converter.convert_tool_choice(model_settings.tool_choice)
         converted_tools = Converter.convert_tools(tools, handoffs)
         response_format = Converter.get_response_format(output_schema)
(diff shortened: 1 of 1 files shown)


src/agents/models/openai_responses.py
```diff
@@ -253,9 +253,16 @@ async def _fetch_response(
         if _debug.DONT_LOG_MODEL_DATA:
             logger.debug("Calling LLM")
         else:
+            # Combine system and user messages only for logging purposes
+            combined_prompt_for_logging = (
```
seratch: This is not the actual "input" payload data sent to the server, so we hesitate to have this kind of modification even for the debug logging. instead, how about simply adding the "instructions" line to the debug log output? It makes more sense.

seratch (changes requested):

src/agents/models/openai_responses.py
HafizFasih: @seratch 
Thanks! I've updated the log to include system_instructions as a separate line without modifying the input.

src/agents/models/openai_responses.py
```diff
@@ -255,6 +256,7 @@ async def _fetch_response(
         else:
             logger.debug(
                 f"Calling LLM {self.model} with input:\n"
+                f"System instructions: {system_instructions}\n"
                 f"{json.dumps(list_input, indent=2, ensure_ascii=False)}\n"
```
seratch, line 260:
```suggestion
                f"{json.dumps(list_input, indent=2, ensure_ascii=False)}\n"
                f"Instructions: {system_instructions}\n"
```

PR #1161, opened 2025-07-16
Add missing guardrail exception import to quickstart

slubbers-openai:
docs: add missing InputGuardrailTripwireTriggered import to quickstart example

Fixes NameError when handling guardrail exceptions by including the required import in the docs.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -95,7 +95,6 @@ You can define custom guardrails to run on the input or output.
 
 ```python
 from agents import GuardrailFunctionOutput, Agent, Runner
-from agents.exceptions import InputGuardrailTripwireTriggered
 from pydantic import BaseModel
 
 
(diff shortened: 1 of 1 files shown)


seratch:
The full version is covered at https://openai.github.io/openai-agents-python/guardrails/ and this quick start page keeps things as simple as possible. Having the complete code for guardrail use cases in quick start too would be fine, but in the case, try/except clauses should be added too (the current code snippet in the quick start page does not use the exception at all).

slubbers-openai:
Thanks for the feedback @seratch! Good catch. I’ve now added the full try/except block (when I was working through this example having the exception handling made the outputs much easier to read and navigate). 

Understand if you’d prefer to keep the quickstart as simple as possible though, so feel free to accept or discard this addition as you see fit.

seratch (approved):
LGTM; @rm-openai let me know if you have any opinions here!

rm-openai (approved):

PR #598, opened 2025-04-24
Add a new GH Actions job to automatically update translated document pagse

seratch:
This pull request adds a new GitHub Actions job to automate the translation of document pages.
- Before this job can run, **OPENAI_API_KEY must be added to the project secrets.**
- It typically takes 8–10 minutes using the o3 model, so the job is configured to run only when there are changes under docs/ or in mkdocs.yml.
- The job commits and pushes the translated changes, but it does not deploy the documents to GitHub Pages. If we think it’s better to deploy the latest changes automatically as well, I’m happy to update the workflow. (Personally, I don’t think it’s necessary, since the changes will be deployed with the next deployment job execution)

PR #1162, opened 2025-07-17
Adjust #598 to only create a PR rather than pushing changes to main branch

seratch:
This pull request resolves the execution error of #598 CI job. The job pushes the changes directly to the main branch. However, our branch policies do not allow bypassing required checks, so it always fails. This pull request changes its behavior just to create a pull request and then ask humans to review (actually you don't need to check translation results though) and merge it.


seratch:
Let me merge this now to prevent noisy false failures

Issue #745, opened 2025-05-22
Memory (with sessions) Implementation


knowsuchagency:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
  - yes
- **Have you searched for related issues?** Others may have had similar requests
  - yes

### Describe the feature

I've noticed it greatly improves developer experience to have an interface with a default implementation or two to handle conversation memory.

<details>
<summary>Currently, the docs suggest the following:
</summary>

```python
async def main():
    agent = Agent(name="Assistant", instructions="Reply very concisely.")

    with trace(workflow_name="Conversation", group_id=thread_id):
        # First turn
        result = await Runner.run(agent, "What city is the Golden Gate Bridge in?")
        print(result.final_output)
        # San Francisco

        # Second turn
        new_input = result.to_input_list() + [{"role": "user", "content": "What state is it in?"}]
        result = await Runner.run(agent, new_input)
        print(result.final_output)
        # California
```

</details>

It's currently left to the user to explicitly manage session memory using `result.to_input_list()`

## Alternatives

A good example of what this could look like is [Google's ADK](https://google.github.io/adk-docs/sessions/).

Another such example from my own AI abstraction library, [Promptic](https://google.github.io/adk-docs/sessions/)

rm-openai:
Yeah fair feedback. Would love to discuss the interface in this issue and would even welcome a contribution once we are happy with it!

knowsuchagency:
Cool! I went ahead and submitted #752 after some experimentation. I'm more than happy to workshop the API -- I just found coding and documentation to be the best way to shape my thinking.

## Usage

```python
from agents import Agent, Runner, Session, SQLiteSession

agent = Agent(name="Assistant", instructions="Reply concisely.")
session: Session = SQLiteSession("convo_123")

messages = [
    "Hi, I'm planning a trip to Japan",
    "What's the best time to visit?",
    "How about cherry blossom season?"
]

for message in messages:
    response = Runner.run_sync(agent, message, session=session)
    print(response.final_output)

```

## Session Interface

The basic idea is to have an interface (`typing.Protocol`) that describes how to manage conversation history for a given session. A `Session` instance would be passed to one of the `Runner` methods to handle conversation history for the session. I included a `SQLiteSession` implementation since that won't introduce any new dependencies.

```python
class CustomSession:
    session_id: str

    async def get_items(self, limit: int | None = None) -> List[dict]:
        ...
[...]

PR #1163, opened 2025-07-17
Initial pass at PostgreSQL session management

artificial-aidan:
### Summary

* Add postgres_session.py
    * Implements session management using PostgreSQL
    * Accepts a pool as a default argument
    * Additional class method for creation from a connection string
* Add optional-dependency for psycopg

### Test plan

Tests were written mocking the postgres calls. This was tested locally against a docker postgres database, with a variety of input messages.

### Issue number

Related to #745

### Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ voice = ["numpy>=2.2.0, <3; python_version>='3.10'", "websockets>=15.0, <16"]
 viz = ["graphviz>=0.17"]
 litellm = ["litellm>=1.67.4.post1, <2"]
 realtime = ["websockets>=15.0, <16"]
+psycopg = ["psycopg[pool]>=3.2.9,<4"]
 
 [dependency-groups]
 dev = [
--- a/src/agents/extensions/memory/postgres_session.py
+++ b/src/agents/extensions/memory/postgres_session.py
@@ -1,0 +1,290 @@
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+try:
+    import psycopg
+    from psycopg import sql
+    from psycopg.rows import class_row
+    from psycopg_pool import AsyncConnectionPool
+except ImportError as _e:
+    raise ImportError(
+        "`psycopg` is required to use the PostgreSQLSession. You can install it via the optional "
+        "dependency group: `pip install 'openai-agents[psycopg]'`."
+    ) from _e
+
+if TYPE_CHECKING:
+    from agents.items import TResponseInputItem
+
+from agents.memory.session import Session
+
+
+@dataclass
+class MessageRow:
+    """Typed row for message queries."""
+
+    message_data: TResponseInputItem
+
+
+@dataclass
+class MessageWithIdRow:
+    """Typed row for message queries that include ID."""
+
+    id: int
+    message_data: TResponseInputItem
+
+
+class PostgreSQLSession(Session):
+    """PostgreSQL-based implementation of session storage.
+
+    This implementation stores conversation history in a PostgreSQL database.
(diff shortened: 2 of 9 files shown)


aidandj:
Wanted to get the code up here in case there was some discussion on how to do things. Will look into the documentation tomorrow.

aidandj:
Got a little stuck on the `old_version_test` needed to have the pyscopg extras installed, not sure how litellm is working. I updated the make target to include all extras when testing python 3.9. Let me know if this isn't the path you'd prefer.

I took a stab at documentation. Open to feedback on structuring. Overall I wanted the postgres session to be in extensions because psycopg can be somewhat heavy on certain platforms. But I wasn't sure if the docs should go with extensions or the other memory docs.

aidandj:
Should be ready now. Have been running it in our testing for about a week now.

aidandj:
Tests fixed. Missed the test changes I had made locally.

src/agents/extensions/memory/postgres_session.py
```diff
@@ -0,0 +1,293 @@
...
+    This implementation stores conversation history in a PostgreSQL database.
+    Requires psycopg to be installed.
+    """
+
+    pool: AsyncConnectionPool
+
+    def __init__(
+        self,
+        session_id: str,
+        pool: AsyncConnectionPool,
```
DamyanBG, line 51: Why you use a hole connection pool just for one session management? One connection will be enough.

aidandj: In my use case this is running in a service with many concurrent sessions. A shared connection pool is used between them. That's why you can initialize it from a connection string, or using an existing pool

DamyanBG: Okay, that is your way, it is not bad.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ voice = ["numpy>=2.2.0, <3; python_version>='3.10'", "websockets>=15.0, <16"]
 viz = ["graphviz>=0.17"]
 litellm = ["litellm>=1.67.4.post1, <2"]
 realtime = ["websockets>=15.0, <16"]
+psycopg = ["psycopg[pool]>=3.2.9,<4"]
```
knowsuchagency, line 41: I'd recommend `psycopg-binary` so users don't need a compiler installed to use this https://pypi.org/project/psycopg-binary/

src/agents/extensions/memory/postgres_session.py
knowsuchagency, line 51: It would be nice to have this as an optional parameter with the default set to `None`. It can then be created in the `__init__` if the user doesn't pass it explicitly.

aidandj: How could it be created? That was the intention of the `from_connection_string` classmethod

pyproject.toml
aidandj: Does this cover all platform/arch combinations? By not specifying binary, any user of this library should be able to choose binary if they'd like. If a binary doesn't exist for a platform then won't over specifying the dependency cause issues? (I have not experimented with the binary library)

Mukish45:
@seratch @rm-openai - May I know the expected timeline for this release?

seratch:
We've accepted this SQLAlchemy based implementation, which supports not only PostgreSQL but any other RDBs: https://github.com/openai/openai-agents-python/pull/1357 

Thus, to avoid duplicated implementations in the built-in modules, let us close this pull request now. Anyone who prefers a simpler one can use this PostgreSQL implementation by copying the code in their projects.

We truly appreciate your time and efforts here. Thank you so much again!

Mukish45:
Makes sense. Thanks @seratch

PR #1169, opened 2025-07-17
Realtime: only update model settings from session

rm-openai:
### Summary:
Was running into bugs. Because the model settings were being set from both runner and session, and that was causing issues. Among other things, handoffs were broken because the runner wasn't reading them, and the session wasn't setting them in the connect() call.

### Test Plan:
Unit tests.

### DIFF
--- a/src/agents/realtime/runner.py
+++ b/src/agents/realtime/runner.py
@@ -2,13 +2,10 @@
 
 from __future__ import annotations
 
-import asyncio
-
-from ..run_context import RunContextWrapper, TContext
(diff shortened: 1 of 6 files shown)


jhills20 (approved):

sibblegp:
Fantastic! Thank you! I feel so vindicated that I was right.

Now I just need to figure out why my interruptions don't work and yours do.

sibblegp:
@rm-openai Could these same reasons be why your demos work with interruptions but mine with a custom config doesn't?

PR #1180, opened 2025-07-18
Fix image_generator example error on Windows OS

seratch:
Due to the method name typo, it does not work on the OS.


examples/tools/image_generator.py
```diff
@@ -12,7 +12,7 @@ def open_file(path: str) -> None:
     if sys.platform.startswith("darwin"):
         subprocess.run(["open", path], check=False)  # macOS
     elif os.name == "nt":  # Windows
-        os.astartfile(path)  # type: ignore
+        os.startfile(path)
```
seratch:
```suggestion
        os.startfile(path)  # type: ignore
```

rm-openai (approved):

PR #1185, opened 2025-07-18
Realtime: add user agent header

rm-openai:
---
* #1198
* __->__ #1185


src/agents/realtime/openai_realtime.py
```diff
@@ -97,6 +98,8 @@
     RealtimeModelSendUserInput,
 )
 
+_USER_AGENT = f"Agents/Python {__version__}"
```
seratch, line 101: We already have this constant in several source files, so it may be about time to have a single unified one. That said, you can go ahead for now, and we can refactor later on.

seratch (approved):

PR #1188, opened 2025-07-19
docs(examples): use UserLocation dataclass in WebSearchTool example

mshsheikh:
This update refines the `examples/web_search.py` snippet to instantiate `WebSearchTool.user_location` with the proper `UserLocation` dataclass rather than a raw `dict`. This change improves type safety, editor autocompletion, and future‑proofs the example against any additional `UserLocation` fields.

See the official docs for `WebSearchTool` and its `user_location` parameter here: [https://openai.github.io/openai-agents-python/ref/tool/#agents.tool.WebSearchTool](https://openai.github.io/openai-agents-python/ref/tool/#agents.tool.WebSearchTool)

### DIFF
--- a/examples/tools/web_search.py
+++ b/examples/tools/web_search.py
@@ -1,13 +1,21 @@
 import asyncio
 
-from agents import Agent, Runner, WebSearchTool, trace
-
+from agents import Agent, Runner, WebSearchTool, UserLocation, trace
 
 async def main():
     agent = Agent(
         name="Web searcher",
         instructions="You are a helpful agent.",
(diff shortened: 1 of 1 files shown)


examples/tools/web_search.py
```diff
@@ -1,13 +1,23 @@
 import asyncio
 
 from agents import Agent, Runner, WebSearchTool, trace
+from agents.tool import UserLocation  # type: ignore[attr-defined]
```
seratch, line 4: the reason mypy says this is that this type comes from openai package, not openai-agents

seratch (commented):
Thanks for the suggestion. This one could be a good one, but using dict can be handier.

PR #1178, opened 2025-07-18
Fix MCP tool name conflicts by adding server name prefixes

fadeoreo:
## Problem

When two MCP servers have tools with the same internal tool name, running `agent list MCP: list tools from server xxxx` will hang, and after the tool list is (supposedly) returned, the agent cannot work properly. 

During troubleshooting, issues such as network, protocol, registration, or async problems were suspected, but the real cause was that one of the MCP tools had a conflicting tool name, making it impossible for the agent SDK to distinguish between tools from different servers.

## Solution

This PR implements server name prefixing for MCP tool names to ensure global uniqueness and avoid conflicts:

1. **Prefixes tool names with server name**: Tool names are prefixed with `{server_name}_{tool_name}` format (e.g., `serverA_run`, `serverB_run`)
2. **Preserves original names**: The original tool name is stored in the `original_name` attribute for actual tool calls
3. **Maintains backward compatibility**: Tool invocation still uses the original name when calling the MCP server

### Core Implementation
- **`src/agents/mcp/server.py`**: Modified `list_tools()` method to add server name prefixes and preserve original names
- **`src/agents/mcp/util.py`**: Updated `invoke_mcp_tool()` to use original names for tool calls

### Testing
- **`tests/mcp/test_tool_name_conflicts.py`**: Comprehensive test suite with 9 test cases covering:
  - Basic prefixing functionality
  - Multiple server conflict resolution  
  - MCPUtil integration
  - Legacy behavior regression testing
  - Edge cases (empty names, special characters)
  - Tool invocation verification

### Before (Conflict)

Server1: [run, echo]
Server2: [run, list]
Result: ❌ Conflict on 'run' tool name → agent list hangs

### After (Resolved)

Server1: [server1_run, server1_echo]
Server2: [server2_run, server2_list]
Result: ✅ All tool names are unique → agent list works correctly

Fixes #1167

PR #1190, opened 2025-07-19
fix(mcp-server): store original_name via setattr to satisfy mypy

mshsheikh:
Directly assigning to `tool.original_name` was triggering “has no attribute ‘original\_name’” errors under mypy. This change replaces that assignment with:

```python
setattr(cast(Any, tool), "original_name", tool.name)
```

Here, we cast each `MCPTool` to `Any` and use `setattr()` so mypy will not complain, but at runtime the `original_name` attribute is still added exactly as before.

In addition, a `mypy.ini` (or equivalent) entry excluding the `tests/` folder ensures that missing‐type‐arg diagnostics and any references to `original_name` in tests no longer break CI. All other logic and tool filter behavior remains unchanged.

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -275,8 +275,14 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
             # Reset the cache dirty to False
             self._cache_dirty = False
             # Fetch the tools from the server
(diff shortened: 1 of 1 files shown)


seratch:
The changes are copied from https://github.com/openai/openai-agents-python/pull/1178 let me close this one as duplicate

PR #1198, opened 2025-07-21
Realtime: make sure we use the initial model settings

rm-openai:
Made a mistake here - we were ignoring the settings passed into the Runner init., and only usig the override settings passed to run()


pakrym-oai (approved):

sibblegp:
Nice work. Appreciate your listening to all the feedback!

Issue #1174, opened 2025-07-17
Something Issue in Agent Class


subhankaladi:
In the **Agent Class**, the ` name: str `property is required — that's fine.
The property ` handoff_description: str | None = None ` is optional because it accepts both str and None types — this is also fine.

### But But But...

```
def as_tool(
    self,
    tool_name: str | None,
    tool_description: str | None,
    custom_output_extractor: Callable[[RunResult], Awaitable[str]] | None = None,
) -> Tool:
```
This is a method of the Agent class. Both` tool_name ` and ` tool_description ` are annotated as str | None.
So why does it raise this error?
`TypeError: Agent.as_tool() missing 2 required positional arguments: 'tool_name' and 'tool_description'`

**My mind is getting disturbed because of this issue.**
Please help me fix it.
Thank You 
Best Regard
[SubhanKaladi](https://www.linkedin.com/in/subhankaladi/)

mar-yam12:
Using `tool_name: str | None` indicates that the `name` parameter can be either a string or None.
However, no default value is provided, so this argument is required when the function is called.
 This is useful when the value must always be explicitly passed, even if it might be None.

def as_tool(name: str | None):
    ...

seratch:
I mentioned passing both is currently recommended for better performance of the tool calls. That being said, one thing I should mention is with Agents SDK TypeScript, actually these are required parameters when calling `Agent#asTool()`: https://openai.github.io/openai-agents-js/guides/tools/#3-agents-as-tools and it's also reasonable.
 
I personally think removing the possibility of None for the name and description here (again, note that not setting these affects stability of tool calls) rather than setting default `None` for the arguments.

@rm-openai This could be a minor breaking change, but what do you think? If you're down to allowing default `None` + falling back to agent.name & empty description, it's fine for me too. Happy to make the adjustments in future release.

Ai80-8:
Stop

seratch:
For most use cases, having handoff_description in Agent definition should work well. So, let us close this one without making changes to the SDK.

PR #1199, opened 2025-07-21
Enh - Use handoff_description as Default Fallback for Tool Description in as_tool() of agents

Kunmeer-SyedMohamedHyder:
### Summary

Replaced the default fallback value for `description_override` in the `as_tool()` method of the `Agent` class from an empty string (`""`) to `self.handoff_description`. This ensures that when no explicit tool description is provided, the agent's `handoff_description` is used as a meaningful and human-readable default. This improves clarity, reduces the need for redundant arguments, and aligns with the DRY principle.

### Checks

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -253,7 +253,7 @@ class Agent(AgentBase, Generic[TContext]):
 
         @function_tool(
             name_override=tool_name or _transforms.transform_string_function_style(self.name),
-            description_override=tool_description or "",
+            description_override=tool_description or self.handoff_description,
         )
         async def run_agent(context: RunContextWrapper, input: str) -> str:
             from .run import Runner


seratch:
Thanks for sending this. Indeed, using an empty string as the default is not effective, but the handoff_description is meant to be used when the agent itself is the destination of the handoff. So, when this tool is used as a tool, the workflow differs. 

I’m proposing to make the tool name and description required parameters to clarify things, as mentioned at https://github.com/openai/openai-agents-python/issues/1174#issuecomment-3085887159  I believe that’s the best way to configure agents as tools.

Kunmeer-SyedMohamedHyder:
Right! Makes sense. Like you have mentioned in the other PR, making name and description mandatory could be a breaking change. But if we are going ahead with the mandatory fields, this PR can be closed.

PR #1213, opened 2025-07-22
Update README.md

Mohsinraza23:
Added Async Example to Quick Start section


seratch:
Thank you for sending this! However, other sections on the same README already have asyncio examples and we'd like to keep the README as simple as possible. So, let us close this one without merging. Thanks again for your time and interest!

PR #1220, opened 2025-07-22
Realtime: fix interrupt and audio tracking

rm-openai:


seratch:
typecheck is failing:
```
Run make mypy
uv run mypy .
src/agents/realtime/openai_realtime.py:459: error: Incompatible types in assignment (expression has type "openai.types.beta.realtime.session.Session", variable has type "openai.types.beta.realtime.session_update_event.Session | None")  [assignment]
Found 1 error in 1 file (checked 269 source files)
make: *** [Makefile:20: mypy] Error 1
```

Issue #715, opened 2025-05-19
Allow agent to return logprobs


jdmatthews-ai:
Enable agents to return the logprobs for their response. This could be used downstream to determine confidence of the agent's response and use the response appropriately.

tompakeman-oai:
This is now available in the latest version

PR #1226, opened 2025-07-23
Fix ResponseTextDeltaEvent validation error for non-OpenAI model providers

vmalyi:
The `ChatCmplStreamHandler.handle_stream()` method fails with a Pydantic validation error when creating `ResponseTextDeltaEvent` objects during streaming operations with non-OpenAI model providers (e.g., Gemini via Google's OpenAI-compatible API).

**Error:**
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for ResponseTextDeltaEvent
logprobs
  Field required [type=missing, input_value={'content_index': 0, 'del...', 'sequence_number': 3}, input_type=dict]
```

**Root Cause:**
The `ResponseTextDeltaEvent` constructor in `chatcmpl_stream_handler.py:193` is missing the required `logprobs` parameter. The OpenAI SDK type definition for `ResponseTextDeltaEvent` defines `logprobs: List[Logprob]` as a required field, but the Agents SDK doesn't provide it when creating these events.

**Type Inconsistency:**
- `ResponseTextDeltaEvent.logprobs`: `List[Logprob]` (required)
- `ResponseOutputText.logprobs`: `Optional[List[Logprob]] = None` (optional)

This inconsistency causes issues when using model providers that don't return logprobs data.

### Debug information
- Agents SDK version: `v0.2.3`
- Python version: `Python 3.10`
- OpenAI SDK version: `1.97.1`
- Pydantic version: `2.11.7`

### Repro steps

```python
import asyncio
from agents.models.openai_chatcompletions import OpenAIChatCompletionsModel

async def reproduce_bug():
    # Using Gemini via OpenAI-compatible API (doesn't provide logprobs)
    gemini_client = AsyncOpenAI(
        base_url="https://generativelanguage.googleapis.com/v1beta/openai/",
        api_key="your-gemini-api-key",
    )
    
    agent = Agent(
        name="Test Agent",
        model=OpenAIChatCompletionsModel(
            model="gemini-2.5-pro",
            openai_client=gemini_client,
        ),
    )
    
    # This will fail with the logprobs validation error
    stream_result = Runner.run_streamed(agent, "Hello, world!")
    
    async for event in stream_result.stream_events():
        if event.type == "raw_response_event":
            continue  # Error occurs here during internal stream processing

if __name__ == "__main__":
    asyncio.run(reproduce_bug())
```

The error occurs in `agents/models/chatcmpl_stream_handler.py` at line 193 where `ResponseTextDeltaEvent` is constructed without the required `logprobs` field.

### Expected behavior

Streaming should work seamlessly with all OpenAI-compatible model providers, including those that don't provide logprobs data (like Gemini). The `ResponseTextDeltaEvent` should be created with proper logprobs handling that:

1. Uses actual logprobs data when available (OpenAI models)
2. Falls back to empty list when not available (non-OpenAI providers)

### Solution

The fix uses `getattr()` to gracefully handle the logprobs field, following the existing pattern used elsewhere in the codebase (like line 91 with `chunk.usage`):

**Before:**
```python
yield ResponseTextDeltaEvent(
    content_index=state.text_content_index_and_output[0],
    delta=delta.content,
    item_id=FAKE_RESPONSES_ID,
    # logprobs field missing - causes validation error
    output_index=state.reasoning_content_index_and_output
    is not None,
    type="response.output_text.delta",
    sequence_number=sequence_number.get_and_increment(),
)
```

**After:**
```python
yield ResponseTextDeltaEvent(
    content_index=state.text_content_index_and_output[0],
    delta=delta.content,
    item_id=FAKE_RESPONSES_ID,
    logprobs=getattr(delta, "logprobs", []) or [],
    output_index=state.reasoning_content_index_and_output
    is not None,
    type="response.output_text.delta",
    sequence_number=sequence_number.get_and_increment(),
)
```

### Impact

This bug affects:
- All users of non-OpenAI model providers (e.g. Gemini)
- Any streaming operations with models that don't return logprobs
- All official SDK examples when used with non-OpenAI providers
- Integration with LiteLLM and other OpenAI-compatible proxy services

### Testing

I've tested the fix with:
- ✅ Gemini 2.5 Pro via Google's OpenAI-compatible API
- ✅ Streaming operations with `Runner.run_streamed()`
- ✅ Both `raw_response_event` and `run_item_stream_event` handling
- ✅ Backward compatibility with OpenAI models (when logprobs are provided)


vmalyi:
Most likely caused by https://github.com/openai/openai-agents-python/issues/715

parthushah8:
an easy fix for me was to downgrade to OpenAI SDK version: 1.97.0

handrew:
Also getting this issue. Switching to `openai==1.96.1` works but unideal - hope this gets merged soon. Hopefully a small change, 🙏🏼 @rm-openai

vmalyi:
Fixed in https://github.com/openai/openai-agents-python/pull/1246

PR #1242, opened 2025-07-25
Realtime: enable a playback tracker

rm-openai:
So far, we've been assuming that audio is played:
- immediately (i.e. with 0 delay/latency)
- at realtime

This causes issues with our interrupt tracking. The model wants to know how much audio the user has actually heard. For example in a phone call agent, this wouldn't work (bc theres a delay of a few hundred ms between model sending audio and the user hearing it). This PR allows you to pass a playback tracker.

---
* #1252
* #1216
* #1243
* __->__ #1242

### DIFF
--- a/src/agents/realtime/_default_tracker.py
+++ b/src/agents/realtime/_default_tracker.py
@@ -1,0 +1,47 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import datetime
+
+from ._util import calculate_audio_length_ms
+from .config import RealtimeAudioFormat
+
+
+@dataclass
+class ModelAudioState:
+    initial_received_time: datetime
+    audio_length_ms: float
+
+
+class ModelAudioTracker:
+    def __init__(self) -> None:
+        # (item_id, item_content_index) -> ModelAudioState
+        self._states: dict[tuple[str, int], ModelAudioState] = {}
+        self._last_audio_item: tuple[str, int] | None = None
+
+    def set_audio_format(self, format: RealtimeAudioFormat) -> None:
+        """Called when the model wants to set the audio format."""
+        self._format = format
+
+    def on_audio_delta(self, item_id: str, item_content_index: int, audio_bytes: bytes) -> None:
+        """Called when an audio delta is received from the model."""
+        ms = calculate_audio_length_ms(self._format, audio_bytes)
+        new_key = (item_id, item_content_index)
+
+        self._last_audio_item = new_key
+        if new_key not in self._states:
+            self._states[new_key] = ModelAudioState(datetime.now(), ms)
+        else:
+            self._states[new_key].audio_length_ms += ms
+
+    def on_interrupted(self) -> None:
+        """Called when the audio playback has been interrupted."""
+        self._last_audio_item = None
+
+    def get_state(self, item_id: str, item_content_index: int) -> ModelAudioState | None:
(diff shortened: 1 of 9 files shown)


seratch:
it seems the file conflicts with main branch need to be resolved

src/agents/realtime/_default_tracker.py
```diff
@@ -0,0 +1,47 @@
...
+    def on_audio_delta(self, item_id: str, item_content_index: int, bytes: bytes) -> None:
+        """Called when an audio delta is received from the model."""
+        ms = calculate_audio_length_ms(self._format, bytes)
+        new_key = (item_id, item_content_index)
+
+        self._last_audio_item = new_key
+        if new_key not in self._states:
+            self._states[new_key] = ModelAudioState(datetime.now(), ms)
+        else:
+            self._states[new_key].audio_length_ms += ms
```
seratch: nit: In general, using built-in/reserved names like `bytes` for variables should be avoided. If `audio` does not sound great, `data`, `delta`, `audio_data` etc. should be fine too.
```suggestion
    def on_audio_delta(self, item_id: str, item_content_index: int, audio: bytes) -> None:
        """Called when an audio delta is received from the model."""
        ms = calculate_audio_length_ms(self._format, audio)
        new_key = (item_id, item_content_index)

        self._last_audio_item = new_key
        if new_key not in self._states:
            self._states[new_key] = ModelAudioState(datetime.now(), ms)
        else:
            self._states[new_key].audio_length_ms += ms
```

src/agents/realtime/_util.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+from .config import RealtimeAudioFormat
+
+
+def calculate_audio_length_ms(format: RealtimeAudioFormat | None, bytes: bytes) -> float:
+    if format and format.startswith("g711"):
+        return (len(bytes) / 8000) * 1000
+    return (len(bytes) / 24 / 2) * 1000
```
seratch: same as above

```suggestion
def calculate_audio_length_ms(format: RealtimeAudioFormat | None, audio: bytes) -> float:
    if format and format.startswith("g711"):
        return (len(audio) / 8000) * 1000
    return (len(audio) / 24 / 2) * 1000
```

src/agents/realtime/model.py
```diff
@@ -6,13 +6,95 @@
...
+    def on_play_bytes(self, item_id: str, item_content_index: int, bytes: bytes) -> None:
+        """Called by you when you have played some audio.
+
+        Args:
+            item_id: The item ID of the audio being played.
+            item_content_index: The index of the audio content in `item.content`
+            bytes: The audio bytes that have been fully played.
+        """
+        ms = calculate_audio_length_ms(self._format, bytes)
+        self.on_play_ms(item_id, item_content_index, ms)
```
seratch, line 50:
```suggestion
    def on_play_bytes(self, item_id: str, item_content_index: int, audio: bytes) -> None:
        """Called by you when you have played some audio.

        Args:
            item_id: The item ID of the audio being played.
            item_content_index: The index of the audio content in `item.content`
            audio: The audio bytes that have been fully played.
        """
        ms = calculate_audio_length_ms(self._format, audio)
        self.on_play_ms(item_id, item_content_index, ms)
```

seratch (approved):
minor suggestion of naming local variables; they are local variables so it's okay to use the same name though

PR #1243, opened 2025-07-25
Realtime: only cancel response if necessary

rm-openai:


seratch:
Please check Python 3.9 compatibility

src/agents/realtime/openai_realtime.py
```diff
@@ -483,6 +491,9 @@ async def _handle_ws_event(self, event: dict[str, Any]):
             await self._emit_event(RealtimeModelTurnEndedEvent())
         elif parsed.type == "session.created":
             await self._send_tracing_config(self._tracing_config)
+            self._update_created_session(parsed.session)  # type: ignore
```
seratch, line 494: I looked into the ways to make the type compatible here, but found it's way more complicated than I expect and dealing with it does not bring any benefits to the code... 😅  So, having ignore comments for these is okay for now.

seratch (approved):

PR #1252, opened 2025-07-25
Realtime: forward all raw model events

rm-openai:


seratch (approved):
Once the tests pass, we can merge this enhancement

apiltamang:
just out of curiousity, what is this PR about? What does it mean to pass "raw model events to server", and can you give one or two example use cases?

PR #1258, opened 2025-07-25
Add parentheses for clarity in exit handler

mshsheikh:
Added parentheses around the boolean expression in `__exit__` to make the evaluation order explicit. This improves readability while maintaining identical functionality.


seratch:
Thanks for sending this! However, the code has only a single condition, so the change does not look necessary to me.

PR #1259, opened 2025-07-25
Fix type placeholder convention

mshsheikh:
Updated the error message to replace the placeholder `your_type` with `YourType`, removing the underscore and adopting PascalCase, which aligns with standard Python type naming conventions.


seratch (approved):
Thanks. This makes sense.

PR #1260, opened 2025-07-26
Tweak on guardrail docstrings

mshsheikh:
✅ The article usage is now grammatically correct with "an" before class names starting with vowel sounds
✅ The possessive form correctly uses "function's name" instead of "function name"
✅ The redundant word "do" has been removed from "to do check" → "to check"
✅ The line length violations (E501) have been resolved with proper docstring formatting
✅ Overall documentation now maintains professional consistency and grammatical accuracy

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -128,7 +128,7 @@ class InputGuardrail(Generic[TContext]):
 @dataclass
 class OutputGuardrail(Generic[TContext]):
     """Output guardrails are checks that run on the final output of an agent.
-    They can be used to check if the output passes certain validation criteria
+    They can be used to do check if the output passes certain validation criteria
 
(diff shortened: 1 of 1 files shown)


src/agents/guardrail.py
```diff
@@ -127,12 +128,12 @@ async def run(
 @dataclass
 class OutputGuardrail(Generic[TContext]):
     """Output guardrails are checks that run on the final output of an agent.
-    They can be used to do check if the output passes certain validation criteria
+    They can be used to check if the output passes certain validation criteria
```
seratch: to me, either is fine; let me revert this
```suggestion
    They can be used to do check if the output passes certain validation criteria
```

seratch (approved):

PR #1267, opened 2025-07-27
refactor: remove inherited method from Agent class

windsornguyen:
## Brief
This PR removes the `get_mcp_tools` method from the `Agent` class to reduce code duplication.

## Motivation
The `Agent` class inherits from `AgentBase` and therefore already has access to the `get_mcp_tools` method through inheritance.


windsornguyen:
Ditto for the `get_all_tools` method. See commit 23c2865.

seratch (approved):
Thank you!

PR #1270, opened 2025-07-27
fix: enforce min trigger size=1 to prevent immediate exports

mshsheikh:
When max_queue_size is small (e.g., 1), max_queue_size * export_trigger_ratio could be less than 1, resulting in a zero trigger size. This would cause immediate exports after adding any item to the queue (since queue_size >= 0 is always true), defeating the purpose of batching. This change ensures the trigger size is at least 1 by using max(1, int(max_queue_size * export_trigger_ratio)).

**Fix**:  
```python
# Before (broken for small queues)
trigger_size = int(max_queue_size * export_trigger_ratio)

# After (fixed: min 1 item needed)
trigger_size = max(1, int(max_queue_size * export_trigger_ratio))
```

**Key change**:  
Added `max(1, ...)` to guarantee the trigger size is **never 0**, so batching works even for tiny queues.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -183,7 +183,7 @@ class BatchTraceProcessor(TracingProcessor):
         self._shutdown_event = threading.Event()
 
         # The queue size threshold at which we export immediately.
-        self._export_trigger_size = int(max_queue_size * export_trigger_ratio)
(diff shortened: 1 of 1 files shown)


seratch (approved):
I think no one could set the max to 1 in reality, but the change logically makes sense. Thanks!

Issue #1262, opened 2025-07-26
Bug: `context.context.name` used in dynamic instruction in Agent documentation, but `name` not in UserContext


Zubair-Ahmad-Atd:
### Summary
In the Agents SDK Documentation for using dynamic_instructions while discussing the Agent, the following line is used:
```python
def dynamic_instructions(
    context: RunContextWrapper[UserContext], agent: Agent[UserContext]
) -> str:
    return f"The user's name is {context.context.name}. Help them with their questions."
```
The line `{context.context.name}` is used, however, the UserContext dataclass shown does not contain a name attribute:
```python
@dataclass
class UserContext:
    uid: str
    is_pro_user: bool
```
This leads to a runtime error:
AttributeError: 'UserContext' object has no attribute 'name'

### Suggested Fixes
You can resolve this by either:
 Adding  name: str to the UserContext model, OR
 Changing the instruction function to use one of the fields that do exist, like uid or is_pro_user.

### Example fix: 
```python
def dynamic_instructions(context: RunContextWrapper[UserContext], agent: Agent[UserContext]) -> str:
    return f"User ID is {context.context.uid}. Help them accordingly."
```

seratch:
Thanks for taking the time to report this. You're right that the document is inconsistent.

sameer20002:
Hi,  I’m interested in fixing this. Can I take this up?

kankute-sameer:
Hi, I’m interested in fixing this. Can I take this up?

PR #1280, opened 2025-07-28
docs: fix UserContext example

kankute-sameer:
### Problem
The documentation in docs/agents.md & docs/ja/agents.md contained an inconsistency where the dynamic_instructions function referenced context.context.name, but the UserContext dataclass only defined uid and is_pro_user fields. This would cause a runtime AttributeError when users tried to follow the example.

### Solution
- Added the missing name: str field to the UserContext dataclass to match what the dynamic_instructions function expects.

### Changes Made
- Added name: str field to the UserContext dataclass in the Context section
- Maintained consistency between the dataclass definition and the dynamic instructions example
- Fixed formatting issues in related files

### Testing
✅ Code formatting passes (make format-check)
✅ Documentation example is now consistent and functional
✅ No runtime errors when following the example
✅ Verified locally: Ran make serve-docs and confirmed the documentation renders correctly with the fix(make serve-docs)

### Issue number
Fixes https://github.com/openai/openai-agents-python/issues/1262

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -34,6 +34,7 @@ Agents are generic on their `context` type. Context is a dependency-injection to
 ```python
 @dataclass
 class UserContext:
+    name: str
     uid: str
     is_pro_user: bool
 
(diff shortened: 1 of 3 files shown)


seratch (approved):
Thank you!

PR #1284, opened 2025-07-28
feat: add loop safeguard and fix instruction spacing/grammar

mshsheikh:
* Introduced `max_attempts` counter to prevent infinite judging loops (defaults to 5)
* Added trailing spaces between adjacent string literals for clearer agent prompts
* Inserted missing “the” in evaluator instructions for grammatical accuracy

### DIFF
--- a/examples/agent_patterns/llm_as_a_judge.py
+++ b/examples/agent_patterns/llm_as_a_judge.py
@@ -15,7 +15,7 @@ with the outline.
 story_outline_generator = Agent(
     name="story_outline_generator",
     instructions=(
-        "You generate a very short story outline based on the user's input."
+        "You generate a very short story outline based on the user's input. "
         "If there is any feedback provided, use it to improve the outline."
     ),
 )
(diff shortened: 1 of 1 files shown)


examples/agent_patterns/llm_as_a_judge.py
```diff
@@ -30,9 +30,9 @@ class EvaluationFeedback:
 evaluator = Agent[None](
     name="evaluator",
     instructions=(
-        "You evaluate a story outline and decide if it's good enough."
-        "If it's not good enough, you provide feedback on what needs to be improved."
-        "Never give it a pass on the first try. After 5 attempts, you can give it a pass if story outline is good enough - do not go for perfection"
```
seratch, line 35: the instructions allow running 5+ times, so the changes in this PR are inconsistent. if you remove the max_attempts etc., we are happy to merge other changes.

examples/agent_patterns/llm_as_a_judge.py
```diff
@@ -46,6 +46,8 @@ async def main() -> None:
 
     # We'll run the entire workflow in a single trace
     with trace("LLM as a judge"):
+        max_attempts = 5
```
seratch: as i mentioned above, please remove this additional logic, which is not necessary

mshsheikh: Apologies, I clicked the review request by mistake before reverting the changes.

seratch (approved):

examples/agent_patterns/llm_as_a_judge.py
mshsheikh: Thanks for pointing that out. The `max_attempts` logic has been removed, and only the instruction formatting fixes are kept.

PR #1292, opened 2025-07-29
Handle processor exceptions and fix tracing log formatting

mshsheikh:
* Wrap each call in `SynchronousMultiTracingProcessor` (`on_trace_start`, `on_trace_end`, `on_span_start`, `on_span_end`, `shutdown`, `force_flush`) in a `try/except` so one failing processor won’t prevent the others from running.

* Add a missing space in the `logger.error` message when no active trace is found, ensuring it reads “…first Returning NoOpSpan.” cleanly.

### DIFF
--- a/src/agents/tracing/provider.py
+++ b/src/agents/tracing/provider.py
@@ -43,28 +43,40 @@ class SynchronousMultiTracingProcessor(TracingProcessor):
         Called when a trace is started.
         """
         for processor in self._processors:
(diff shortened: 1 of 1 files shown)


seratch (approved):
in general, on_xxx callbacks are responsible to handle runtime exceptions, but having these should be good to go; @rm-openai can you review this just in case?

rm-openai (approved):

PR #1298, opened 2025-07-29
Allow to pass both session and input list

### LINKED ISSUE #1140: multiple inputs with session support
### Describe the bug
when I use multiple inputs with session,  There was a Error:
Error processing message: Cannot provide both a session and a list of input items. When using session memory, provide only a string input to append to the conversation, or use session=None and provide a list to manually manage conversation history.

### Debug information
- Agents SDK version: v0.2.0
- Python version (Python 3.10)

### Expected behavior
I want to use the multiple inputs with session in agent

DanieleMorotti:
### Summary

This PR has been implemented to resolve #1140 . I added a parameter to the `RunConfig` object such that it can be specified if the given input list will replace the session history or it will be appended to it.

The current implementation is a possible approach but I would be happy to change the code if there are better solutions.

@seratch @rm-openai whenever you have a moment, please let me know your thoughts.

### Test plan

I added 2 unit tests in `tests/test_session.py`.

### Checks

### DIFF
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -394,8 +394,8 @@ async def test_session_memory_rejects_both_session_and_list_input(runner_method)
             await run_agent_async(runner_method, agent, list_input, session=session)
 
         # Verify the error message explains the issue
-        assert "You must specify the `session_input_callback` in" in str(exc_info.value)
-        assert "manually manage conversation history" in str(exc_info.value)
+        assert "list inputs require a `RunConfig.session_input_callback" in str(exc_info.value)
+        assert "to manage the history manually" in str(exc_info.value)
 
         session.close()
 


src/agents/run.py
```diff
@@ -1212,8 +1223,17 @@ async def _prepare_input_with_session(
         # Convert input to list format
         new_input_list = ItemHelpers.input_to_new_input_list(input)
 
-        # Combine history with new input
-        combined_input = history + new_input_list
+        if session_input_handling == "append" or session_input_handling is None:
```
seratch: To me, allowing to pass callback function here to combine the item list to save sounds more flexible. @rm-openai what do you think?

DanieleMorotti: Yes, this would offer much more flexibility.
However, would you still preserve the option for the users to specify "replace" or "append" in order to apply some predefined processing?

In that scenario, the `RunConfig` attribute could accept a `Callable`, a `str` or `None`.

Or would you prefer to have a more complex implementation like the one done for MCP filters (`ToolFilterCallable`, etc)?

seratch: Sessions are supposed to be used for maintaining conversation history across multiple agent runs. So, I don't think "replace" is a common use case. If "replace" works well, do you really need a session for the case?

 If a developer would like to make some adjustments to stored history based on some conditions, that could be a valid use case (I still think it's an edge case though). That's why I thought enabling callback functions here would make more sense. 

Let me know if I am missing anything!

DanieleMorotti: I agree. And what should happen if the callback function is left unspecified (`None`) and the input is a list? We may raise an error, as in the current implementation.

Moreover, does the callback function only take the history and the list of new items as input, or should I allow additional parameters?

DanieleMorotti:
I implemented a new version, where the `RunConfig` attribute accepts a callback function that takes as input the history from the session and the new inputs.

Let me know what else should be improved.

rm-openai:
I'm missing some context, what's the use case for this? I read the issue but couldn't quite figure it out

gonzamordecki:
@rm-openai Adding some context, as I just faced this limitation:

An agent currently cannot be run using built-in sessions from the SDK while receiving two new inputs from the user. This is requested, for example, when passing an image and text (i.e., Product Image + “Do you have this product in black?”).

DanieleMorotti:
@rm-openai I implemented this feature because it seemed useful to several people, but if you don't want to include it, just let me know and I'll close the PR.

fdb02983rhy:
Multimodal input is a fundamental feature for agents, so please allow users to handle it through the built-in session management. Currently, I have to call .to_input_list() and manage it manually.

gemue-parndt:
We really need this feature! 
I want to handle images and texts but currently only text is allowed therefore the whole Session feature is not usable...
@rm-openai please approve this PR :)

tauzhang-cn:
I also have the same problem above， please！

src/agents/memory/util.py
```diff
@@ -0,0 +1,29 @@
...
+Args:
+    history_items: The list of items from the session history.
+    new_items: The list of new input items for the current turn.
+
+Returns:
+    A list of combined items to be used as input for the agent. Can be sync or async.
+"""
+
+
+SessionInputHandler = Union[SessionMixerCallable, None]
```
seratch: For consistency with other code in this SDK, can you remove None from here and use `SessionInputHandler | None` in other source files? Also, can you make the type name consistent with the property name session_input_callback?

src/agents/run.py
```diff
@@ -179,6 +179,14 @@ class RunConfig:
     An optional dictionary of additional metadata to include with the trace.
     """
 
+    session_input_callback: SessionInputHandler = None
```
seratch: To avoid confusion for code readers, remove None union type from SessionInputHandler and change the type def here (and others):
```suggestion
    session_input_callback: SessionInputCallback | None = None
```

src/agents/run.py
```diff
@@ -1479,18 +1491,18 @@ async def _prepare_input_with_session(
...
     ) -> str | list[TResponseInputItem]:
         """Prepare input by combining it with session history if enabled."""
         if session is None:
             return input
 
-        # Validate that we don't have both a session and a list input, as this creates
-        # ambiguity about whether the list should append to or replace existing session history
-        if isinstance(input, list):
+        # If the user doesn't explicitly specify a mode, raise an error
+        if isinstance(input, list) and not session_input_callback:
```
seratch, line 1495: Asking users to pass session_input_callback is a breaking change, plus not having the callback should be fine

src/agents/memory/util.py
DanieleMorotti: I renamed `SessionMixerCallable` to `SessionInputCallback` since we don’t need two separate variables referring to the same thing (imo `SessionInputHandler` is not necessary)

src/agents/run.py
DanieleMorotti: This doesn’t seem like a breaking change to me, because it already raises an error if the user provides a list as input.
The difference now is that it should only raise an error if the input is a list and `session_input_callback` is `None`. Otherwise, it should use that function to handle the input. Let me know what do you think about it

seratch: Ah, yes. Thanks for clarifying this. The current error message should be better than the current one

DanieleMorotti:
I resolved the conflicts and added a clearer explanation of the error when a session and input list are provided together.

seratch:
Thanks for the PR and working on it for a while!

DanieleMorotti:
> Thanks for the PR and working on it for a while!

You're welcome! I'm glad for the opportunity to contribute and grow as a programmer.

PR #1300, opened 2025-07-29
Fix: Emit tool_called events immediately in streaming runs

### LINKED ISSUE #1282: Bug: `run_streamed` Buffers `tool_call_item` Events, Delaying Real-Time Feedback
### Describe the bug

The `Runner.run_streamed()` method buffers `tool_call_item` events and only emits them when the tool actually starts executing, rather than streaming them immediately when the tool call is made. This significantly delays real-time feedback in streaming applications and defeats the purpose of using streamed execution for responsive UIs.

### Debug information
- Agents SDK version: 0.2.3
- Python version: 3.10.16

### Repro steps

```python
import asyncio
import datetime
import time
from agents import Agent, ItemHelpers, Runner, function_tool

@function_tool
def get_server_time() -> str:
    time.sleep(10)  # Simulate a slow tool
    return datetime.datetime.now().isoformat()

async def main():
    agent = Agent(
        name="TimeKeeper",
        instructions="First call the `get_server_time` tool to get the current server time, then tell the user what time it is.",
        tools=[get_server_time],
    )

    result = Runner.run_streamed(agent, input="Hello")
    print("=== Run starting ===")

        timestamp = datetime.datetime.now().strftime("%H:%M:%S.%f")[:-3]
        if event.type == "run_item_stream_event":
            if event.item.type == "tool_call_item":
                print(f"[{timestamp}] -- Tool was called")
            elif event.item.type == "tool_call_output_item":
                print(f"[{timestamp}] -- Tool output: {event.item.output}")

    print("=== Run complete ===")

if __name__ == "__main__":
    asyncio.run(main())
```

**Current output:**
```
=== Run starting ===
[15:44:50.553] Agent updated: TimeKeeper
[15:45:01.710] -- Tool was called        # ← This appears AFTER the 10-second delay
[15:45:01.710] -- Tool output: 2025-07-28T15:45:01.708868
=== Run complete ===
```

### Expected behavior

The `tool_call_item` event should be emitted immediately when the agent decides to call a tool, not when the tool starts executing. The expected output should be:

```
=== Run starting ===
[15:44:50.553] Agent updated: TimeKeeper
[15:44:50.XXX] -- Tool was called        # ← Should appear immediately
# ... 10 second delay while tool executes ...
[15:45:01.710] -- Tool output: 2025-07-28T15:45:01.708868
=== Run complete ===
```

habema:
### Summary

Resolves #1282 

### The Solution

The implementation has been updated to emit the semantic `tool_called` event as soon as the first streaming chunk indicating a tool call is received from the model.

1. The `_run_single_turn_streamed` method now inspects raw `ResponseOutputItemAddedEvent` events as they arrive. If an event's item is a tool call, a corresponding `RunItemStreamEvent(name="tool_called")` is immediately constructed and put on the queue.

2. To avoid sending a duplicate event when the full turn results are processed at the end of the stream, a set of `emitted_tool_call_ids` is maintained for the duration of the turn.

3. The check for whether an event item is a tool call has been refactored. Instead of using a long, hardcoded tuple of types, it now dynamically creates this tuple from the `ToolCallItemTypes` alias.


habema:
@seratch  My apologies for the duplicate work.

After looking at both, I think this approach might have a couple of advantages: 
- Instead of waiting for the model's full response to finish, this implementation emits the tool_called event the instant the first tool-call chunk arrives in the stream.
- From a maintainability perspective, it also dynamically checks for any tool-call type, so the code won't need to be updated when more tools are added down the line.

This should make the UI feel more responsive, and the code change itself is a bit smaller and more contained. Happy to chat about it or defer to the other solution if that's preferred.

habema:
Hey @seratch @rm-openai.
Merging #974 resolves this problem, but I still see some advantages/improvements in this implementation. Would it be possible to take a look at it, and either go forward with it or close it.

seratch:
While I am confident #974 works well, I am not sure if your code here works without any issues. Would it be possible to enhance the current code with #974 to produce "tool_called" events as you intend here?

habema:
It doesnt. There are merge conflicts that were caused by merging #974. I'll fix them briefly.

habema:
@seratch 
Rebased onto latest `main` and resolved conflicts. Happy to proceed with this, or close if you prefer sticking with #974 alone.

As previously mentioned, advantages to this implementation are:

> * Instead of waiting for the model's full response to finish, this implementation emits the tool_called event the instant the first tool-call chunk arrives in the stream.
> * From a maintainability perspective, it also dynamically checks for any tool-call type, so the code won't need to be updated when more tools are added down the line.

seratch:
@rm-openai I think this is good to go. You can check the behavior by running something like this:
```python
    result = Runner.run_streamed(agent, input="Please tell me 5 jokes.")
        if event.type == "run_item_stream_event" and event.name == "tool_called":
            print(event.item.raw_item.name + " called with " + event.item.raw_item.arguments)
        if event.type == "raw_response_event" and isinstance(event.data, ResponseTextDeltaEvent):
```

habema:
@rm-openai @seratch any updates on this?

seratch:
@rm-openai let me know in a few business days if you have any comments on this. I'll merge this PR if there is no concern.

rm-openai:
@seratch - the idea makes sense. Didn't have time to closely review the PR so will trust you on that

felixs8696:
thanks for this! is this in the public docs yet?

habema:
@felixs8696 its an underlying change in the implementation of `Runner.run_streamed()`. It is live, but I don't think there is anything to document.

PR #1302, opened 2025-07-30
Fix #1238 by enhancing HandoffInputData and enable passing async functions

### LINKED ISSUE #1238: Items added to sessions during handoff do not show up
In the following test, I expect "name a blue fruit" to be present in the history; but it is not:

```python
import pytest

# Agents SDK imports
from agents.run import Runner
from agents import (
    Agent,
    AgentHooks,
    RunContextWrapper,
    handoff,
    SQLiteSession,
)

# --- Hooks for Foo ----------------------------------------------------------

class FooHooks(AgentHooks):
    """Inject an extra prompt into the session during hand-off."""

    async def on_handoff(
        self,
        ctx: RunContextWrapper,
        agent: Agent,  # Destination agent (Bar)
        source: Agent,  # Source agent (Foo)
    ) -> None:
        # Append a follow-up question to the conversation history so that Bar
        # sees it when it takes over.
        print(await ctx.context.session.get_items())
        await ctx.context.session.add_items(
            [{"role": "assistant", "content": "name a blue fruit"}]
        )
        print(await ctx.context.session.get_items())


# --- Test -------------------------------------------------------------------

@pytest.mark.asyncio
async def test_handoff_session_runs():
    """Run a simple Foo → Bar conversation with a session in the context."""

    # Shared session so that hooks can write to the conversation history.
    session = SQLiteSession(session_id="test_session")
    runtime_context = SimpleNamespace(session=session)

    # Destination agent (Bar)
    bar = Agent(
        name="Bar",
        instructions="Respond politely to follow-up questions.",
    )

    # Starting agent (Foo)
    foo = Agent(
        name="Foo",
        instructions="Answer with the name of a red fruit, then hand off to Bar.",
        hooks=FooHooks(),
    )

    # Foo hands the conversation to Bar once it's answered.
    foo.handoffs = [handoff(bar)]

    # Kick off the run; no assertions needed – success is that it executes.
    await Runner.run(starting_agent=foo, input="name a red fruit", context=runtime_context) 
```

Note that the print statements in the handoff, show that the item is now present:

```bash
tests/test_handoff_session.py []
[{'role': 'assistant', 'content': 'name a blue fruit'}]
.
```

But in the console, it is not there:

Image

seratch:
This pull request resolves #1238

### DIFF
--- a/src/agents/_run_impl.py
+++ b/src/agents/_run_impl.py
@@ -774,6 +774,7 @@ class RunImpl:
                     else original_input,
                     pre_handoff_items=tuple(pre_step_items),
                     new_items=tuple(new_step_items),
+                    run_context=context_wrapper,
                 )
                 if not callable(input_filter):
                     _error_tracing.attach_error_to_span(
@@ -785,6 +786,8 @@ class RunImpl:
                     )
                     raise UserError(f"Invalid input filter: {input_filter}")
                 filtered = input_filter(handoff_input_data)
+                if inspect.isawaitable(filtered):
+                    filtered = await filtered
                 if not isinstance(filtered, HandoffInputData):
                     _error_tracing.attach_error_to_span(
                         span_handoff,
--- a/src/agents/handoffs.py
+++ b/src/agents/handoffs.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import inspect
 import json
 from collections.abc import Awaitable
-from dataclasses import dataclass
+from dataclasses import dataclass, replace as dataclasses_replace
 from typing import TYPE_CHECKING, Any, Callable, Generic, cast, overload
 
 from pydantic import TypeAdapter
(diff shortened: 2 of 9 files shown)


src/agents/_run_impl.py
```diff
@@ -785,6 +786,8 @@ async def execute_handoffs(
                     )
                     raise UserError(f"Invalid input filter: {input_filter}")
                 filtered = input_filter(handoff_input_data)
+                if inspect.isawaitable(filtered):
```
seratch, line 789: to support this use case: https://github.com/openai/openai-agents-python/issues/1238#issuecomment-3130234196

src/agents/handoffs.py
```diff
@@ -49,8 +49,24 @@ class HandoffInputData:
     handoff and the tool output message representing the response from the handoff output.
     """
 
+    run_context: RunContextWrapper[Any] | None = None
+    """
+    The run context at the time the handoff was invoked.
+    Note that, since this property was added later on, it's optional for backwards compatibility.
+    """
+
+    def clone(self, **kwargs: Any) -> HandoffInputData:
```
seratch, line 58: added clone method for convenience but if we don't want to have this, happy to revert this change

tests/test_agent_runner.py
```diff
@@ -262,7 +263,7 @@ async def test_handoff_filters():
 
 
 @pytest.mark.asyncio
-async def test_async_input_filter_fails():
+async def test_async_input_filter_supported():
```
seratch, line 266: if we didn't intentionally block async functions, i believe  changing this is a good addition

tests/test_agent_runner_streamed.py
```diff
@@ -241,6 +241,7 @@ def remove_new_items(handoff_input_data: HandoffInputData) -> HandoffInputData:
         input_history=handoff_input_data.input_history,
         pre_handoff_items=(),
         new_items=(),
+        run_context=handoff_input_data.run_context,
```
seratch, line 244: I added run_context param in tests but the ones under examples do not have the param. this means backward compatibility is kept

src/agents/handoffs.py
```diff
@@ -49,8 +49,24 @@ class HandoffInputData:
     handoff and the tool output message representing the response from the handoff output.
     """
 
+    run_context: RunContextWrapper[Any] | None = None
```
seratch, line 52: if a different name is preferred, happy to rename this

seratch (commented):
comments for reviewers

PR #1321, opened 2025-07-31
Enhance exception guide in docs

MuhammadHamidRaza:
This Pull Request aims to significantly improve the clarity and understanding of `Agents SDK` exceptions within the documentation. I have updated the definitions for all core exceptions (found in `agents.exceptions`) to be more precise, comprehensive, and easier for beginners to grasp.

**Why these changes are important:**

Previously, some exception definitions might have been concise but lacked the detailed explanation necessary for new users to fully understand *when* and *why* a particular exception is raised. The goal here is to make the documentation more beginner-friendly by:

* **Providing Clearer Context:** Each definition now offers more context on the circumstances that lead to the exception.
* **Actionable Understanding:** Users will better understand the root cause of the error, helping them to debug and implement agents more effectively.
* **Improved Onboarding:** A clear understanding of error handling from the outset will greatly enhance the onboarding experience for developers new to the `Agents SDK`.

**Specific improvements for each exception:**

* **`AgentsException`**: Clarified its role as the base class for all SDK exceptions.
* **`MaxTurnsExceeded`**: Explicitly mentions all relevant `Runner` methods (`run`, `run_sync`, `run_streamed`) where `max_turns` can be exceeded.
* **`ModelBehaviorError`**: Detailed the common scenarios such as "Malformed JSON" and "Non-existent Tools", which are crucial for understanding unexpected model output.
* **`UserError`**: Emphasized that this exception is for errors made by the developer using the SDK, due to incorrect implementation or configuration.
* **`InputGuardrailTripwireTriggered` / `OutputGuardrailTripwireTriggered`**: Provided clearer explanations of what guardrails are and their purpose (checking inputs before processing, and checking outputs before delivery, respectively).

These updated definitions will be a valuable asset to the SDK's documentation, making it much easier for anyone, regardless of their prior experience, to understand and handle errors effectively.

I believe this update will directly contribute to a smoother learning curve and a better overall experience for all `Agents SDK` users.


docs/running_agents.md
```diff
@@ -128,8 +128,10 @@ You can use the Agents SDK [Temporal](https://temporal.io/) integration to run d
...
+-   [`AgentsException`][agents.exceptions.AgentsException]: This is the base class for all exceptions raised within the SDK. It serves as a generic type from which all other specific exceptions are derived.
+-   [`MaxTurnsExceeded`][agents.exceptions.MaxTurnsExceeded]: This exception is raised when the agent's run exceeds the `max_turns` limit passed to the `Runner.run`, `Runner.run_sync`, or `Runner.run_streamed` methods. It indicates that the agent could not complete its task within the specified number of interaction turns.
+-   [`ModelBehaviorError`][agents.exceptions.ModelBehaviorError]: This exception occurs when the underlying model (LLM) produces unexpected or invalid outputs. This can include:
+    -   Malformed JSON: When the model provides a malformed JSON structure for tool calls or in its direct output, especially if a specific `output_type` is defined.
+    -   Non-existent Tools: When the model attempts to call a tool that is not present in the agent's defined list of tools.
```
seratch: The missing tools situation is not common, and it can potentially happen only with handoffs. So, instead of highlighting the specific one, we want to mention more general idea. 
```suggestion
    -   Unexpected tool-related failures: When the model fails to use tools in an expected manner
```

seratch (approved):

PR #1331, opened 2025-07-31
docs: clarify usage of `name` and `instructions` fields in Agent constructor Fix agent docs

Hafiz-Muhammad-Umar12:
This pull request improves the documentation by clarifying the difference between the name and instructions parameters in the Agent constructor.
These small additions help new users better understand how to define their agent’s identity and behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,12 @@ print(result.final_output)
 
 (_For Jupyter notebook users, see [hello_world_jupyter.ipynb](examples/basic/hello_world_jupyter.ipynb)_)
 
+### Note on `name` vs `instructions`
+
+- `name` is a **required** parameter used to identify your agent internally (for logs, tracing, handoffs, etc.), **but it is not passed to the LLM**.
+- `instructions` is an **optional** parameter, but it **is sent to the LLM** as a system prompt to define how the agent should behave.
+
+
 ## Handoffs example
 
 ```python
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -168,6 +168,13 @@ print(result.final_output)
 
 (_For Jupyter notebook users, see [hello_world_jupyter.ipynb](examples/basic/hello_world_jupyter.ipynb)_)
 
+### Note on `name` vs `instructions`
```
seratch: I don't think having a new section for this on README is the best way to tell this. How about adding a simple line comment to the above code snippet?

seratch (changes requested):

MuhammadHamidRaza:
it's soo great because instruction is optional and name is required it's soo great idea i also want to do this add

README.md
Hafiz-Muhammad-Umar-dev: fixit

docs/agents.md
```diff
@@ -19,7 +18,9 @@ def get_weather(city: str) -> str:
     return f"The weather in {city} is sunny"
 
 agent = Agent(
+    # The 'name' is the agent's identity, used for context and tool calling.
```
seratch, line 21: What I meant was mentioning "required" or not for these two parameters rather than adding these description

seratch:
Thanks for sending this, but I think the changes are not necessary for now, so let me close this one.

PR #1332, opened 2025-07-31
Realtime: make example customizable:

rm-openai:
Trying to make it easy to customize/use your own agents and reuse the UI.


pakrym-oai (approved):

pakrym-oai:
might be cool to be able to point it to the file `python server.py myagent.py`

PR #1335, opened 2025-08-01
Merge pull request #1 from openai/main

UmmeHabiba1312:
This PR improves the explanation for using the `trace` parameter in the documentation. It helps new users understand how to pass and use trace objects for debugging agent runs.


seratch:
It seems nothing is changed in this PR

PR #1337, opened 2025-08-01
docs: Add Tracing to the core primitives list

MuhammadHamidRaza:
This Pull Request makes a small but important update to the "Core Concepts" section of the documentation.

**The change is:**
* The existing documentation mentions `Tracing` as a built-in feature, but it is not included in the bulleted list of core primitives. This update adds `Tracing` to that list, making the overview of the SDK's fundamental components more accurate and complete.

The update is kept to a single line and maintains the overall conciseness of the page, in line with previous feedback. This change helps new users immediately recognize Tracing as a core part of the SDK's architecture for debugging and observability.


seratch:
Thanks for the suggestion, however the following paragraph mentions tracing. Thus, we don't need this document changes. Thanks again for sending this.

PR #1343, opened 2025-08-01
docs: Improve clarity of LitellmModel and InternalChatCompletionMessage

MuhammadHamidRaza:
This Pull Request improves the clarity and user-friendliness of two key docstrings in the `litellm_model.py` file. The goal is to make these core components more accessible to new developers without altering any functionality.

**The changes are:**

* **InternalChatCompletionMessage**: The docstring has been updated to explain *why* this internal class is important. It now clarifies that it's used to store the agent's "reasoning content," which is vital for debugging and understanding the agent's thought process.
* **LitellmModel**: The docstring has been expanded to better explain the value of using a `LiteLLM` model. It now highlights its key benefit: providing a flexible way to switch between multiple LLM providers (like Gemini, OpenAI, etc.) without changing core agent code.

These changes are purely textual and do not affect the functionality of the SDK. They enhance the documentation by providing a more complete and understandable overview of these core primitives.


seratch:
@MuhammadHamidRaza I think the current comments are not so bad, so I don't see the necessity to change them. Also, when you send a pull request, please include only the relevant diff. This means you need to start a new branch based on the latest main branch instead of reusing a long-lived branch. Also, having only single commit as the starting point would be appreciated.

PR #1347, opened 2025-08-01
Enhanced handoff_prompt.py docstring with example and args for better clarity 

Hafiz-Muhammad-Umar12:
Improved handoff_prompt.py documentation by adding a detailed docstring with args, returns, and an example. The changes aim to make the function clearer for beginners while preserving the original content.


seratch:
Thanks for sending this. However, we don't need this change to the docstring as it's a duplicate of the prefix prompt in the comment.

PR #1349, opened 2025-08-02
Fix HTTPX timeout and include all export items

mshsheikh:
* Correct the `httpx.Timeout` call to use separate `connect` and `read` timeouts
* Prevent client initialization errors by using a valid timeout signature
* Remove the `if item.export()` filter so no exported data is dropped
* Ensure empty or falsy `export()` results are still sent to the backend
* Improve reliability of trace exports without changing existing APIs

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -60,7 +60,8 @@ class BackendSpanExporter(TracingExporter):
         self.max_delay = max_delay
 
         # Keep a client open for connection pooling across multiple export calls
-        self._client = httpx.Client(timeout=httpx.Timeout(timeout=60, connect=5.0))
+        # use separate connect/read timeouts
+        self._client = httpx.Client(timeout=httpx.Timeout(connect=5.0, read=60.0))
 
     def set_api_key(self, api_key: str):
         """Set the OpenAI API key for the exporter.
@@ -96,7 +97,8 @@ class BackendSpanExporter(TracingExporter):
             logger.warning("OPENAI_API_KEY is not set, skipping trace export")
             return
 
-        data = [item.export() for item in items if item.export()]
+        # include all export() results, even if they’re empty/falsy
+        data = [item.export() for item in items]
         payload = {"data": data}
 
         headers = {


src/agents/tracing/processors.py
```diff
@@ -96,7 +97,8 @@ def export(self, items: list[Trace | Span[Any]]) -> None:
             logger.warning("OPENAI_API_KEY is not set, skipping trace export")
             return
 
-        data = [item.export() for item in items if item.export()]
+        # include all export() results, even if they’re empty/falsy
+        data = [item.export() for item in items]
```
seratch, line 100: Can you elaborate more on this? If there is a certain situation to resolve, we're happy to adjust this logic, but we aren't aware of anything.

src/agents/tracing/processors.py
```diff
@@ -60,7 +60,8 @@ def __init__(
         self.max_delay = max_delay
 
         # Keep a client open for connection pooling across multiple export calls
-        self._client = httpx.Client(timeout=httpx.Timeout(timeout=60, connect=5.0))
+        # use separate connect/read timeouts
+        self._client = httpx.Client(timeout=httpx.Timeout(60.0, connect=5.0))
```
seratch: This doesn't change anything on the actual behavior, so we don't need this change.

seratch (changes requested):
If the second change is really relevant, we may be interested in it. That said, we need more information on it.

mshsheikh:
> If the second change is really relevant, we may be interested in it. That said, we need more information on it.

We made `NoOpSpan.export()` and `NoOpTrace.export()` return `{ }` on purpose; that `{ }` tells you “this span/trace ran but there was nothing extra to send.”
If you drop every falsy result, you’ll never see those no-ops and your dashboards end up missing real events.

**Bottom line**: the core tracer should hand over every export result, even if it’s empty. If a backend or custom exporter wants to drop empties, they can do that themselves where it’s obvious and logged, not buried in the heart of the tracing code.

seratch:
Thanks for your reply, but changing the behavior could break existing integrations' behaviors, plus I don't see the necessity to send empty data to tracing data platform as it does not provide meaningful information. Thanks again for sending this but let me close it now.

PR #1355, opened 2025-08-03
Fix type annotation, avoid input shadowing, and clarify tool behavior docs

mshsheikh:
This commit improves type safety by replacing Any with the correct generic type TContext in the get\_all\_tools method.
It also avoids confusion by renaming the input parameter in the run\_agent function to user\_input because input is a built-in function in Python.
Lastly, it updates the documentation of tool\_use\_behavior to clearly describe the use of the StopAtTools object instead of a plain list of tool names.
These changes make the code more accurate, easier to understand, and better aligned with best practices.

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -101,7 +101,7 @@ class AgentBase(Generic[TContext]):
             self.mcp_servers, convert_schemas_to_strict, run_context, self
         )
 
-    async def get_all_tools(self, run_context: RunContextWrapper[Any]) -> list[Tool]:
+    async def get_all_tools(self, run_context: RunContextWrapper[TContext]) -> list[Tool]:
         """All agent tools, including MCP tools and function tools."""
         mcp_tools = await self.get_mcp_tools(run_context)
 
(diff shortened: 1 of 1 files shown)


src/agents/agent.py
```diff
@@ -101,7 +101,7 @@ async def get_mcp_tools(self, run_context: RunContextWrapper[TContext]) -> list[
             self.mcp_servers, convert_schemas_to_strict, run_context, self
         )
 
-    async def get_all_tools(self, run_context: RunContextWrapper[Any]) -> list[Tool]:
+    async def get_all_tools(self, run_context: RunContextWrapper[TContext]) -> list[Tool]:
```
seratch, line 104: this is good catch

src/agents/agent.py
```diff
@@ -262,12 +264,12 @@ def as_tool(
             name_override=tool_name or _transforms.transform_string_function_style(self.name),
             description_override=tool_description or "",
         )
-        async def run_agent(context: RunContextWrapper, input: str) -> str:
+        async def run_agent(context: RunContextWrapper, user_input: str) -> str:
```
seratch: In general, avoiding "input" as name is recommended, but this is used only within this function locally plus the function never uses built-in `input`. For these reasons, it's safe enough. We may revisit the internal code like this in the future but can you revert this change for now?

PR #1357, opened 2025-08-03
Add SQLAlchemy session backend for conversation history management

### LINKED ISSUE #1328: SQLAlchemy Session Store for Production Databases
The current `SQLiteSession` is excellent for development and simple use cases. For production environments, it would be highly beneficial to support more robust databases like PostgreSQL or MySQL for storing conversation history.

Would the project be open to a contribution that adds a new SQLAlchemySession class?

This would implement the existing `Session` protocol, allowing users to connect to any database supported by SQLAlchemy. It would be an optional feature, and sqlalchemy would be an extra dependency.

If this sounds like a welcome addition, I'm happy to contribute a PR.

habema:
This PR introduces `SQLAlchemySession`, providing a production-grade session storage backend that can connect to any database supported by SQLAlchemy (e.g., PostgreSQL, MySQL).

This implementation is based on the discussion in Issue #1328 and follows the agreed-upon architectural pattern:

*   The new session store is located in `src/agents/extensions/memory/`.
*   SQLAlchemy is added as an optional dependency, available via the `[sqlalchemy]` extra.
*   This is a non-breaking, purely additive change.

A test has been added.

Resolves #1328.

### DIFF
--- a/src/agents/extensions/memory/sqlalchemy_session.py
+++ b/src/agents/extensions/memory/sqlalchemy_session.py
@@ -167,11 +167,11 @@ class SQLAlchemySession(SessionABC):
         engine = create_async_engine(url, **engine_kwargs)
         return cls(session_id, engine=engine, **kwargs)
 
-    def _serialize_item(self, item: TResponseInputItem) -> str:
+    async def _serialize_item(self, item: TResponseInputItem) -> str:
         """Serialize an item to JSON string. Can be overridden by subclasses."""
         return json.dumps(item, separators=(",", ":"))
 
-    def _deserialize_item(self, item: str) -> TResponseInputItem:
+    async def _deserialize_item(self, item: str) -> TResponseInputItem:
         """Deserialize a JSON string to an item. Can be overridden by subclasses."""
         return json.loads(item)  # type: ignore[no-any-return]
 
@@ -213,7 +213,7 @@ class SQLAlchemySession(SessionABC):
             items: list[TResponseInputItem] = []
             for raw in rows:
                 try:
-                    items.append(self._deserialize_item(raw))
+                    items.append(await self._deserialize_item(raw))
                 except json.JSONDecodeError:
                     # Skip corrupted rows
                     continue
@@ -227,7 +227,7 @@ class SQLAlchemySession(SessionABC):
         payload = [
             {
                 "session_id": self.session_id,
-                "message_data": self._serialize_item(item),
+                "message_data": await self._serialize_item(item),
             }
             for item in items
         ]
@@ -282,7 +282,7 @@ class SQLAlchemySession(SessionABC):
                 if row is None:
                     return None
                 try:
-                    return self._deserialize_item(row)
+                    return await self._deserialize_item(row)
                 except json.JSONDecodeError:
                     return None
 


habema:
@seratch 

The new `SQLAlchemySession` feature introduces an optional dependency on `sqlalchemy`. This creates a conflict between two goals:

1.  Keeping `sqlalchemy` as an optional dependency that users don't need to install if they are not using `SQLAlchemySession`.
2.  Ensuring the `SQLAlchemySession` feature is tested in all our CI environments, including `old_version_tests`.

The `old_version_tests` CI job does not install optional extras by default. As a result, directly importing `SQLAlchemySession` in the top-level `agents` package causes an `ImportError` and breaks that CI job.

**Temporary Solution**

As a temporary measure to allow the feature to be merged without breaking the build, the following changes have been made:

1.  **No Top-Level Import**: `SQLAlchemySession` is no longer imported in the root `__init__.py` of the `agents` package. Code that uses it must now import it directly from `agents.extensions.memory.sqlalchemy_session`.
2.  **Conditional Tests**: The tests for `SQLAlchemySession` are now skipped in environments where `sqlalchemy` is not installed, using `pytest.importorskip("sqlalchemy")`.

**Next Steps**

This solution is a trade-off. It prevents the CI from breaking, but it means that the `SQLAlchemySession` feature is **not currently tested** in the `old_version_tests` environment.

I think the correct long-term fix is to update `make old_version_tests` to install the `[sqlalchemy]` extra (and any other extra dependencies needed) for all test jobs. This will allow the tests to run everywhere and provide full test coverage for the new feature.

Let me know what you think.

seratch:
@habema could you resolve the conflict?

habema:
@seratch you got it.
On a sidenote, all mypy and linter errors I fixed are unrelated to any files I changed in this PR. Might need some looking into their behavior.

src/agents/extensions/memory/sqlalchemy_session.py
```diff
@@ -0,0 +1,288 @@
...
+            Forwarded to the main constructor (``create_tables``, custom table names, …).
+        """
+        engine_kwargs = engine_kwargs or {}
+        engine = create_async_engine(url, **engine_kwargs)
+        return cls(session_id, engine=engine, **kwargs)
+
+    # ------------------------------------------------------------------
+    # Session protocol implementation
+    # ------------------------------------------------------------------
+    async def _ensure_tables(self) -> None:
```
seratch, line 181: this could be a slight overhead; enabling this only for a kind of dev mode?

seratch (approved):
Overall, this enhancement looks good to me; great work 👍

src/agents/extensions/memory/sqlalchemy_session.py
habema: Just to confirm I'm getting what you mean:
The idea is to default `create_tables` to `False`, where `True` is for development/testing where migrations aren't used, to avoid the connection overhead on each call. Correct?

seratch: Yes, we're on the same page! Thanks for clarifying this!

habema: done

seratch:
@habema Thanks for working on this. Overall, this PR looks good to me, but I don't have the bandwidth to do thorough testing with it before merging. I will take a look this week or early next week, but until then, please feel free to make further adjustments and/or adding unit tests.

src/agents/extensions/memory/sqlalchemy_session.py
```diff
@@ -0,0 +1,291 @@
...
+                server_default=sql_text("CURRENT_TIMESTAMP"),
+                nullable=False,
+            ),
+            sqlite_autoincrement=True,
+        )
+
+        # Index for efficient retrieval of messages per session ordered by time
+        from sqlalchemy import Index
+
+        Index(
```
seratch: Can you move this to the table definition? The following code should work.

```python
self._messages = Table(
    messages_table,
    self._metadata,
    Column("id", Integer, primary_key=True, autoincrement=True),
    Column("session_id", String, ForeignKey(f"{sessions_table}.session_id", ondelete="CASCADE"), nullable=False),
    Column("message_data", Text, nullable=False),
    Column("created_at", TIMESTAMP(timezone=False), server_default=sql_text("CURRENT_TIMESTAMP"), nullable=False),
    Index(f"idx_{messages_table}_session_time", "session_id", "created_at"),
    sqlite_autoincrement=True,
)
```

src/agents/extensions/memory/sqlalchemy_session.py
```diff
@@ -0,0 +1,291 @@
...
+                stmt = (
+                    select(self._messages.c.message_data)
+                    .where(self._messages.c.session_id == self.session_id)
+                    .order_by(self._messages.c.created_at.asc())
+                )
+            else:
+                stmt = (
+                    select(self._messages.c.message_data)
+                    .where(self._messages.c.session_id == self.session_id)
+                    .order_by(self._messages.c.created_at.desc())
```
seratch, line 203: The logic looks good to me, but can you add a quick comment to the code explaining why desc here and we do rows.reverse() later?

src/agents/extensions/memory/sqlalchemy_session.py
```diff
@@ -0,0 +1,291 @@
...
+
+    async def add_items(self, items: list[TResponseInputItem]) -> None:
+        if not items:
+            return
+
+        await self._ensure_tables()
+        payload = [
+            {
+                "session_id": self.session_id,
+                "message_data": json.dumps(item, separators=(",", ":")),
```
seratch: We may want to have a method to do this; with that, users can enhance this part just by inheriting this class.

tests/extensions/memory/test_sqlalchemy_session.py
```diff
@@ -0,0 +1,153 @@
+from __future__ import annotations
```
seratch, line 1: thanks for adding tests!

seratch (changes requested):
Thanks for this PR! Overall, it already looks great, but I have a few minor suggestions. Let me know what you think!

```diff
diff --git a/src/agents/extensions/memory/sqlalchemy_session.py b/src/agents/extensions/memory/sqlalchemy_session.py
index 038f845..20f0555 100644
--- a/src/agents/extensions/memory/sqlalchemy_session.py
+++ b/src/agents/extensions/memory/sqlalchemy_session.py
@@ -31,6 +31,7 @@ from sqlalchemy import (
     TIMESTAMP,
     Column,
     ForeignKey,
+    Index,
     Integer,
     MetaData,
     String,
@@ -122,18 +123,10 @@ class SQLAlchemySession(SessionABC):
                 server_default=sql_text("CURRENT_TIMESTAMP"),
                 nullable=False,
             ),
+            Index(f"idx_{messages_table}_session_time", "session_id", "created_at"),
             sqlite_autoincrement=True,
         )
 
-        # Index for efficient retrieval of messages per session ordered by time
-        from sqlalchemy import Index
-
-        Index(
-            f"idx_{messages_table}_session_time",
-            self._messages.c.session_id,
-            self._messages.c.created_at,
-        )
-
         # Async session factory
         self._session_factory = async_sessionmaker(
             self._engine, expire_on_commit=False
@@ -180,21 +173,28 @@ class SQLAlchemySession(SessionABC):
                 await conn.run_sync(self._metadata.create_all)
             self._create_tables = False  # Only create once
 
+    async def _serialize_message_data(self, item: TResponseInputItem) -> str:
+        return json.dumps(item, separators=(",", ":"))
+
+    async def _deserialize_message_data(self, item: str) -> TResponseInputItem:
+        return json.loads(item)
+
     async def get_items(self, limit: int | None = None) -> list[TResponseInputItem]:
         await self._ensure_tables()
         async with self._session_factory() as sess:
-            if limit is None:
+            if limit is not None:
                 stmt = (
                     select(self._messages.c.message_data)
                     .where(self._messages.c.session_id == self.session_id)
-                    .order_by(self._messages.c.created_at.asc())
+                    # pick up the top-n new messages here, then reverse the order below
+                    .order_by(self._messages.c.created_at.desc())
+                    .limit(limit)
                 )
             else:
                 stmt = (
                     select(self._messages.c.message_data)
                     .where(self._messages.c.session_id == self.session_id)
-                    .order_by(self._messages.c.created_at.desc())
-                    .limit(limit)
+                    .order_by(self._messages.c.created_at.asc())
                 )
 
             result = await sess.execute(stmt)
@@ -206,7 +206,7 @@ class SQLAlchemySession(SessionABC):
             items: list[TResponseInputItem] = []
             for raw in rows:
                 try:
-                    items.append(json.loads(raw))
+                    items.append(await self._deserialize_message_data(raw))
                 except json.JSONDecodeError:
                     # Skip corrupted rows
                     continue
@@ -220,7 +220,7 @@ class SQLAlchemySession(SessionABC):
         payload = [
             {
                 "session_id": self.session_id,
-                "message_data": json.dumps(item, separators=(",", ":")),
+                "message_data": await self._serialize_message_data(item),
             }
             for item in items
         ]
```

tests/extensions/memory/test_sqlalchemy_session.py
habema: of course :)

src/agents/extensions/memory/sqlalchemy_session.py
```diff
@@ -0,0 +1,294 @@
...
+                res_data = await sess.execute(
+                    select(self._messages.c.message_data).where(self._messages.c.id == row_id)
+                )
+                row = res_data.scalar_one_or_none()
+                await sess.execute(delete(self._messages).where(self._messages.c.id == row_id))
+
+                if row is None:
+                    return None
+                try:
+                    return json.loads(row)  # type: ignore[no-any-return]
```
seratch: _deserialize as well? also make the internal methods async not to receive such a breaking change request in the future

docs/sessions.md
```diff
@@ -164,6 +164,64 @@ result2 = await Runner.run(
 )
 ```
 
+### SQLAlchemy-powered sessions
```
seratch: can you move this document change to a separate pull request? with that, we can merge this PR immediately and then we can merge the document changes after the release

seratch (changes requested):
can you check my last two requests?

src/agents/extensions/memory/sqlalchemy_session.py
habema: yep just fixed this. missed it the first time

docs/sessions.md
habema: sure

src/agents/extensions/memory/sqlalchemy_session.py
```diff
@@ -0,0 +1,298 @@
...
+        engine_kwargs
+            Additional kwargs forwarded to :pyfunc:`sqlalchemy.ext.asyncio.create_async_engine`.
+        kwargs
+            Forwarded to the main constructor (``create_tables``, custom table names, …).
+        """
+        engine_kwargs = engine_kwargs or {}
+        engine = create_async_engine(url, **engine_kwargs)
+        return cls(session_id, engine=engine, **kwargs)
+
+    def _serialize_item(self, item: TResponseInputItem) -> str:
```
seratch: i do understand making these methods async could be weird for you, but can you make these async for any potential customization on the user side?

habema: that makes sense, done

seratch (approved):

russ-safeturf:
When we get limited amount of rows, we can get to the point when we get function_call_output row, but truncate function_call which will result in Error getting response: 
`Error code: 400 - {'error': {'message': 'No tool output found for function call call_.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}. (request_id: req_).`



This is situation I am talking about:
image
image

habema:
@russ-safeturf from my understanding what you mean isnt truncation, but rather retrieval of coupled items that might be separated due to the limit. 

If thats what you mean, this problem isnt inherent to the SQLAlchemy session implementation.

salah9003:
جزاك الله، يا أخي الكريم، على نسجك المتواصل لخيوط الإبداع في مشروع مفتوح المصدر، حيث أضفتَ ببراعة نجمةً ساطعةً لتخزين الجلسات، تُنير دروب إدارة التاريخ بكفاءة. أسأل الله أن يجعل عملك زهرةً في ميزان حسناتك، ويروي قلبك بفيض التوفيق.

PR #1358, opened 2025-08-03
docs: Add a type-safe example for custom `tool_use_behavior`

MuhammadHamidRaza:
This PR improves an existing documentation example by making it more type-safe.

The change replaces the generic `Any` type with `TContext` for the `RunContextWrapper` in the `custom_tool_handler` function. This provides a more accurate and robust example for developers, demonstrating a best practice for using the library's context system.

The improved example highlights:

-   Proper use of the generic `TContext` type.
-   Enhanced code clarity and type safety.
-   A strong reference for other developers to follow when implementing custom tool handlers.

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -206,9 +206,9 @@ agent = Agent(
 - `ToolsToFinalOutputFunction`: A custom function that processes tool results and decides whether to stop or continue with the LLM.
 
 ```python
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this, but TContext is an abtract type representation. So, this change is not relevant for the example code.

PR #1360, opened 2025-08-03
Fix docstring formatting and add safe check for empty message content

mshsheikh:
Improved clarity in a multiline docstring in `RunItemBase` by correcting misplaced parentheses.

Also added guards in `ItemHelpers.extract_last_content` and `extract_last_text` to handle empty `message.content` safely and prevent potential `IndexError`.

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -66,7 +66,7 @@ class RunItemBase(Generic[T], abc.ABC):
     """The agent whose run caused this item to be generated."""
 
     raw_item: T
-    """The raw Responses item from the run. This will always be a either an output item (i.e.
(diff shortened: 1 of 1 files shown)


seratch:
> handle empty message.content

Have you observed this pattern in real use case?

mshsheikh:
> > handle empty message.content
> 
> Have you observed this pattern in real use case?

Yes @seratch , sometimes the model returns a control item with no text, so `message.content` can be completely empty and `message.content[-1]` would crash without the guard. This often happens in streaming mode when you’re piping partial results to a UI and no tokens have arrived yet.
It can also occur if the LLM makes a tool-only response or hands off right away without producing a normal message.

PR #1361, opened 2025-08-03
Fix docstring typo and possessive formatting in lifecycle.py

mshsheikh:


seratch:
Thanks for sending this, but these are all correct. So, there is no need to change them.

PR #1366, opened 2025-08-04
Add functionality to update agent during realtime session

marysha-openai:
Adding support to update current agent during a Realtime Session much like the TS/JS method: https://openai.github.io/openai-agents-js/openai/agents-realtime/classes/realtimesession/#updateagent

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -180,6 +180,29 @@ class RealtimeSession(RealtimeModelListener):
         """Interrupt the model."""
         await self._model.send_event(RealtimeModelSendInterrupt())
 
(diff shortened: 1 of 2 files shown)


src/agents/realtime/session.py
```diff
@@ -180,6 +180,29 @@ async def interrupt(self) -> None:
...
+            agent=self._current_agent,
+        )
+
+        await self._put_event(
+            RealtimeHandoffEvent(
+                from_agent=previous_agent,
+                to_agent=self._current_agent,
+                info=self._event_info,
+            )
+        )
```
rm-openai: We shouldn't send a handoff event - since it's not actually a handoff, it's the developer changing the agent.

rm-openai (approved):

PR #1368, opened 2025-08-04
feat(viz): draw MCP servers

aaron-ang:
### Summary

Add nodes and edges for MCP servers in agent visualization. I'm open to suggestions to modify the style of visualizations.

### Checks

Example graph with MCP server:
automobile_agent_graph

### DIFF
--- a/docs/visualization.md
+++ b/docs/visualization.md
@@ -15,13 +15,17 @@ pip install "openai-agents[viz]"
 You can generate an agent visualization using the `draw_graph` function. This function creates a directed graph where:
 
 - **Agents** are represented as yellow boxes.
+- **MCP Servers** are represented as grey boxes.
 - **Tools** are represented as green ellipses.
 - **Handoffs** are directed edges from one agent to another.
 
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
Thanks for sending this. Looks pretty cool! Can you resolve the CI errors?

aaron-ang:
`mcp` library requires python 3.10 but our `old_version_tests` is pinned at version 3.9. Can we skip visualization tests for this?

seratch:
@aaron-ang Ideally, instead of skipping the test, the tests should ensure that the code does not fail when rendering 3.9-compatible agents. Disabling the entire test file isn’t ideal in this case.

aaron-ang:
If the Python version is < 3.10, we won't be able to create an agent with MCP in the first place due to version conflicts. I've added checks to only create and test for MCP server components if the version is >= 3.10.

seratch:
Thanks for quickly adjusting the tests! Sounds great to me!

PR #1370, opened 2025-08-04
Realtime: fix interrupt logic

rm-openai:


src/agents/realtime/openai_realtime.py
```diff
@@ -351,6 +351,13 @@ async def _send_interrupt(self, event: RealtimeModelSendInterrupt) -> None:
                 int(elapsed_ms),
             )
             await self._send_raw_message(converted)
+        else:
+            logger.info(
```
seratch: nit: do you want this logging at info-level in production? if it's mainly for debugging, debug-level may be more suitable; either is fine though!

seratch (approved):

src/agents/realtime/openai_realtime.py
rm-openai: yeah debug makes more sense

PR #1373, opened 2025-08-05
fix(session): safely handle None run_config when reading guardrails settings

mshsheikh:
Previously, if `run_config` was None, calling `.get("guardrails_settings")` would raise an AttributeError.

This patch adds a safe fallback (`self._run_config or {}`) so that `debounce_text_length` always falls back to 100 when no guardrails settings are provided.


seratch:
Thanks, but the _run_config never be None (see L95)

PR #1377, opened 2025-08-05
fix(handoff): enforce strict boolean return in _check_handoff_enabled

mshsheikh:
This patch ensures `_check_handoff_enabled()` always returns a strict `bool`, regardless of what the agent’s `is_enabled` returns.

### Problem:
Previously, `is_enabled` could return a non-boolean truthy/falsy value (e.g., `None`, `1`, `"yes"`), which could lead to unpredictable behavior in downstream logic, especially in conditionals like:
```python
if await self._check_handoff_enabled(...):
````

### Fix:
Wrapped the result in `bool()`:
```python
if inspect.isawaitable(res):
    res = await res
return bool(res)
```

This guarantees boolean output while maintaining backward compatibility.


seratch:
Since the callable for is_enabled is supposed to return a boolean value, I don't think this change is necessary. Thanks for taking the time for this!

PR #1384, opened 2025-08-06
docs: Add a simple lifecycle hooks example for Python

MuhammadHamidRaza:
This pull request adds a simple and easy-to-understand code example for `Agent` lifecycle hooks to the Python documentation.

I noticed that while the `example` folder does contain examples, they are often more complex and might be difficult for a beginner to grasp. My goal with this PR is to provide a very small, direct, and user-friendly example that shows only the core functionality of `AgentHooks`.

This example is intentionally minimal and focuses on demonstrating how to:
-   Define a custom `AgentHooks` class.
-   Use the `hooks` parameter in the `Agent` definition.

This will make the feature much more accessible to new users, helping them quickly understand and implement lifecycle events without being overwhelmed by a more comprehensive example.


seratch:
Thanks for sending this. However, we generally recommend checking examples.

PR #1392, opened 2025-08-07
Updated the default values for two parameters in the ModelSettings class to align with their documented behavior:

### LINKED ISSUE #1234: Agent-level MCP server tool filtering

rudra-attri:
This PR updates the default values for two parameters in the ModelSettings class to align with their documented behavior


src/agents/model_settings.py
```diff
@@ -104,11 +104,11 @@ class ModelSettings:
     metadata: dict[str, str] | None = None
     """Metadata to include with the model response call."""
 
-    store: bool | None = None
+    store: bool | None = True
     """Whether to store the generated model response for later retrieval.
     Defaults to True if not provided."""
```
seratch: indeed, this needs to be accurate; this should be:
- for Responses API, when you don't pass true here, storing data is automatically enabled
- for Chat Completions API, when you don't pass true here, storing dat is disabled

src/agents/model_settings.py
```diff
@@ -104,11 +104,11 @@ class ModelSettings:
     metadata: dict[str, str] | None = None
     """Metadata to include with the model response call."""
 
-    store: bool | None = None
+    store: bool | None = True
```
seratch: please revert this

src/agents/model_settings.py
```diff
@@ -104,11 +104,11 @@ class ModelSettings:
...
     """Whether to store the generated model response for later retrieval.
     Defaults to True if not provided."""
 
-    include_usage: bool | None = None
+    include_usage: bool | None = True
```
seratch: please revert this

src/agents/model_settings.py
```diff
@@ -104,11 +104,11 @@ class ModelSettings:
...
     """Whether to include usage chunk.
     Defaults to True if not provided."""
```
seratch: more accurately, this option can be used only for Chat Completions API.

seratch (changes requested):

src/agents/model_settings.py
rudra-attri: Done

src/agents/model_settings.py
rudra-attri: Done

src/agents/model_settings.py
rudra-attri: Done

src/agents/model_settings.py
rudra-attri: Updated the description to clarify the same

seratch (approved):

PR #1410, opened 2025-08-08
Implement bidirectional handoff system (#1376)

### LINKED ISSUE #1376: Proposal: Return results to the parent Agent
### Describe the feature
So when I think of Agentic systems, I think of a system that can perform multiple actions all in one step/prompt. These can be across a wide variety of agents.

As it stands right now, in this SDK, once a handoff occurs, it will never get handed back off to the parent agent (what I call the orchestrator). This means if I want a sub-agent to hand off to another one, I need to specify which agents it can hand off to.

Take a coding agent for instance. It may need to read a file, write changes, read another file, run a bash command, and then keep track of a todo list.

This all has a single agent at the top of it handing off to multiple agents below it which then hand back off to the top. This SDK can only hand off to agents downstream which means creating that kind of system would basically be impossible. This can be accomplished by having everything as tools, but that kind of defeats the purpose of having multiple agents with their own tools.

I suggest that, at least as an option, handoffs can hand back off to the agent that called them so they can then call other agents/tools. This creates more of a pattern of a parent agent handling many actions on their own.

This is especially important in the Realtime system where each voice request may be totally unrelated to the agent it's already been handed off to. Unlike the text based system, once a realtime handoff occurs, it stays in the sub agent.

Just an idea since it's preventing me from doing things in a way that makes sense. For example:

"Fetch the quarterly financials for AAPL and save them in a Google Doc called Apple Financials"

In my mind, this would handoff to my financial agent, return to the original agent, and then hand off to the google docs agent. But the way it works now is I have to hand off to the financial agent, and then it hands off to google docs.It's very linear and can only go in one direction.

Just a thought. Let me know what you think.

CodeVoyager007:
Resolves #1376: Bidirectional Handoff System

This PR implements a bidirectional handoff system for orchestrator-style workflows, allowing sub-agents to return control to their parent agent. Key features:
- `AgentRegistry` and `AgentRunner` for scalable agent/workflow management
- `return_to_parent_handoff` and `create_bidirectional_handoff_workflow` for flexible workflows
- Parallel task execution via `run_parallel`
- Convenience functions (`create_financial_workflow_registry`, `create_support_workflow_registry`)
- Comprehensive test suites in `tests/test_bidirectional_handoffs.py` and `tests/test_registry_handoffs.py`
- Documentation in `docs/bidirectional_handoffs.md`, with updates to `docs/index.md`, and `README.md`

### Testing
- All tests pass (`pytest tests/test_bidirectional_handoffs.py tests/test_registry_handoffs.py -v`)
- Covers edge cases (missing agents, circular references, parallel execution)

### Checklist
- [x] Code follows SDK conventions
- [x] Tests cover all new functionality
- [x] Documentation updated
- [x] Backward compatible with existing handoffs

Please review and let me know if any changes are needed


seratch:
Thanks for sending this. We're not planning to add this mechanism in the short term, but sharing your work with communities and knowing what people think would be beneficial for us too.

seratch:
Since we don't have plans to merge this, let me close this one now. Thanks again for your interest!

PR #1419, opened 2025-08-11
Fix TypeError when instantiating ChatCompletionMessageToolCallParam due to typing.Union

Mazz-Ather:
This fixes the error "TypeError: Cannot instantiate typing.Union" that happens in recent openai-python versions (e.g., 1.99.3+). The param type is a Union, so it can't be created directly. Changed to use a dict, which works with the API.

Tested with my own agent code that was failing—now it runs fine without downgrading openai.

Related to openai/openai-python issues like #2529 or #2541 (if they exist; I couldn't find exact matches, but similar problems reported).

Thanks!


seratch:
Thanks for sending this, but we'll resolve this by https://github.com/openai/openai-agents-python/pull/1415

Issue #1416, opened 2025-08-11
TypeError: Cannot instantiate typing.Union when calling tools due to misuse of TypeAlias in chatcmpl_converter.py


Ali-Gadit:
Description:
When running an agent with a tool call, the agents SDK crashes with:
TypeError: Cannot instantiate typing.Union

Python: 3.13
openai (version : 1.99.6)
openai-agents  (version: 0.0.16 )
OS ( Windows:11)

```
Full Traceback (shortened):
  File "...\agents\models\chatcmpl_converter.py", line 443, in items_to_messages
    new_tool_call = ChatCompletionMessageToolCallParam(...)
TypeError: Cannot instantiate typing.Union
```
Root cause:
ChatCompletionMessageToolCallParam is defined in openai.types.chat as:

```
ChatCompletionMessageToolCallParam: TypeAlias = Union[
    ChatCompletionMessageFunctionToolCallParam,
    ChatCompletionMessageCustomToolCallParam
]
This is a TypeAlias for a Union, not a class.
The converter calls it as if it were a constructor, which is invalid in Python:

new_tool_call = ChatCompletionMessageToolCallParam(...)  # ❌
This triggers the error because typing.Union is not instantiable.
```

Expected behavior:

The converter should inspect the data, determine whether it matches
ChatCompletionMessageFunctionToolCallParam or ChatCompletionMessageCustomToolCallParam,
and then instantiate the correct concrete type:

Solution:
```python
elif func_call := cls.maybe_function_tool_call(item):
                asst = ensure_assistant_message()
                tool_calls = list(asst.get("tool_calls", []))
                arguments = func_call["arguments"] if func_call["arguments"] else "{}"
                if func_call.get("type") == "function_call":
                    new_tool_call = ChatCompletionMessageFunctionToolCallParam(
                        id=func_call["call_id"],
                        type="function",
                        function={
                            "name": func_call["name"],
                            "arguments": arguments,
                        },
                    )
                else:
                    new_tool_call = ChatCompletionMessageCustomToolCallParam(
                        id=func_call["call_id"],
                        type="function",
                        custom={
                            "name": func_call["name"],
                            "arguments": arguments,
                        },
                    )
                tool_calls.append(new_tool_call)
                asst["tool_calls"] = tool_calls
```

<img width="600" src="https://github.com/user-attachments/assets/b8aa6fe8-326a-4d8c-bf2e-a80ce2c50c77" />

Steps to reproduce:

[...]

PR #1421, opened 2025-08-11
Fix: resolve union type handling for function call tool parameters in chat completions

Ali-Gadit:
This PR addresses the union type error discussed in issue #1416 by properly distinguishing between ChatCompletionMessageFunctionToolCallParam and ChatCompletionMessageCustomToolCallParam when processing function call tool parameters. This fix prevents runtime errors and improves type safety in chat completions.


seratch:
Thanks for sending this, but we'll resolve this by https://github.com/openai/openai-agents-python/pull/1415

PR #1426, opened 2025-08-12
Fix: enforce strict instructions function signature in get_system_prompt

abdullahimran49:
Previously, the `Agent` class allowed dynamic instructions functions with any two parameters,
including meaningless dummy names (e.g., `def f(wrapper, a)`), which caused confusion and
allowed improper usage.

This commit adds strict signature checking in `get_system_prompt` to ensure the instructions
callable accepts exactly two parameters (context and agent), enforcing correct usage.

If the function signature does not have exactly two parameters, a clear TypeError is raised.

This prevents dummy or incorrect parameter usage, improving API correctness and developer
experience.


seratch:
It seems the tests are still failing

seratch (approved):

PR #1433, opened 2025-08-12
Improve SQLiteSession documentation with a runnable async example

MuhammadAbdullah95:
This update improves the SQLiteSession usage example in the documentation by making it a fully runnable script.
Changes include:
- Wrapped async calls inside an `async def main()` function.
- Used `asyncio.run()` for proper async execution.
- Added `if __name__ == "__main__":` guard for clarity.
- Added comments to guide users and make the snippet beginner-friendly.

This makes it easier for developers (especially those new to async in Python) to run and understand the example without additional setup.

### DIFF
--- a/docs/sessions.md
+++ b/docs/sessions.md
@@ -60,26 +60,38 @@ This eliminates the need to manually call `.to_input_list()` and manage conversa
 Sessions supports several operations for managing conversation history:
 
 ```python
+import asyncio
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. However, we'd like to avoid repeatedly having main function for all code snippets in the page.

PR #1438, opened 2025-08-13
Clarify provider-agnostic wording in documentation

ausafulislam:
Rephrased the sentence for better readability:
From: "It is provider-agnostic, supporting the OpenAI Responses and Chat Completions APIs, as well as 100+ other LLMs."
To: "It works with multiple AI providers, supporting the OpenAI Responses and Chat Completions APIs, as well as 100+ other LLMs."


seratch:
Thanks for sending this PR! However, we don't need this change.

PR #1440, opened 2025-08-13
Add a simple gpt-oss example

seratch:
This pull request adds a simple gpt-oss example app

### DIFF
--- a/examples/basic/hello_world_gpt_oss.py
+++ b/examples/basic/hello_world_gpt_oss.py
@@ -1,0 +1,38 @@
+import asyncio
+import logging
+
+from openai import AsyncOpenAI
+
+from agents import Agent, OpenAIChatCompletionsModel, Runner, set_tracing_disabled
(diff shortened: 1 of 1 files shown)


examples/basic/hello_world_gpt_oss.py
```diff
@@ -0,0 +1,38 @@
...
+        instructions="You're a helpful assistant. You provide a concise answer to the user's question.",
+        model=gpt_oss_model,
+    )
+
+    result = await Runner.run(agent, "Tell me about recursion in programming.")
+    print(result.final_output)
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
```
rm-openai, line 38: might be nice to automatically start ollama here as well, but not required

seratch: yeah, but perhaps we can do something in the future 😸

PR #1460, opened 2025-08-13
Add an example using Ollama Turbo

jason-riddle:
I've created a small example for using the agents sdk with with Ollama Turbo.

Docs:
- https://ollama.com/turbo
- https://github.com/ollama/ollama/blob/main/docs/turbo.md

### DIFF
--- a/examples/basic/simple_ollama_turbo.py
+++ b/examples/basic/simple_ollama_turbo.py
@@ -1,0 +1,40 @@
+import asyncio
+
+from agents.extensions.models.litellm_model import LitellmModel
+
+from agents import (
+    Agent,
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. Ideally, we would like LiteLLM integration to handle the API key properly for this provider too: https://openai.github.io/openai-agents-python/models/litellm/ So, instead of having the workaround example suggested here, we'd like to enhance our implementation for this use case and/or contribute to LiteLLM to enable the straight-forward way.

seratch:
Thanks for the reaction. Let me close this PR now but thanks again for taking the time for it!

PR #1467, opened 2025-08-14
Fix a bug where this SDK's internals fail to handle items restored using Pydantic etc.

seratch:
This pull request resolves a known issue with serialized/deserialized items.

The most common one is `TypeError: cannot pickle 'pydantic_core._pydantic_core.ValidatorIterator' object` with the data loaded via Pydantic methods, but we've received similar reports with other objects too.

This pull request minimizes the use of `copy.deepcopy()` inside this SDK to eliminate the potential runtime errors.

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import abc
-import copy
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Generic, Literal, TypeVar, Union
 
@@ -41,6 +40,7 @@ from typing_extensions import TypeAlias
 
 from .exceptions import AgentsException, ModelBehaviorError
 from .usage import Usage
+from .util._safe_copy import safe_copy
 
 if TYPE_CHECKING:
     from .agent import Agent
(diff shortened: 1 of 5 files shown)


seratch:
@rm-openai yeah, we can simplify the implementation. I think I can remove the memo and deepcopy for basic objects and primitives. anything else you want to remove?

rm-openai:
I think we don't need anything except replace all `copy.deepcopy(...)` with `list(...)`

seratch:
@rm-openai Thanks for clarifying. Since we'd been using copy.deepcopy for union type values (str | list) and potentially we may encounter this issue with other types such as set, tuple. So, I think having this simplified common utility should be helpful for us to more easily avoid making the same mistakes in the future. 

That said, if you prefer a different implementation, you can come up with your own changes. I will be happy to be a reviewer instead.

rm-openai:
Yeah I think this PR is simpler: https://github.com/openai/openai-agents-python/pull/1490. For example, it doesn't recursively copy everything in the list, because the only thing we care about is making a new top level list so that mutations dont affect the original list.

PR #1474, opened 2025-08-14
Feat: Add Optional Structured Session Metadata

### LINKED ISSUE #1385: Suggestion: Database schema for conversations could be better
Right now, we're storing all the parts of a conversation (user messages, tool calls, tool outputs, AI responses) as a simple list of events in one table. Each event is basically a JSON blob.

Here's what it looks like:
| id | session_id | message_data                                                                                                                                                                                                                                   | created_at           |
|----|------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------|
| 1  | user_123   | {"content": "What's the time?", "role": "user"}                                                                                                                                                                                                         | 2025-08-06 14:06:39  |
| 2  | user_123   | {"arguments": "{}", "call_id": "call_ErZ8...", "name": "get_server_time", "type": "function_call", "id": "fc_6893...", "status": "completed"}                                                                                                | 2025-08-06 14:06:39  |
| 3  | user_123   | {"call_id": "call_ErZ8...", "output": "2025-08-06T17:06:37.796774", "type": "function_call_output"}                                                                                                                                           | 2025-08-06 14:06:39  |
| 4  | user_123   | {"id": "msg_6893...", "content": [{"annotations": [], "text": "Hello! The current server time is 5:06 PM on August 6, 2025. How can I assist you today?", "type": "output_text", "logprobs": []}], "role": "assistant", "status": "completed", "type": "message"} | 2025-08-06 14:06:39  |


This works, but it's not great for a few reasons:
- Querying for specific things (like all messages and their respective responses) is a pain because you have to parse JSON in the query.
- I am specifically running into this problem as I'm trying to implement an "edit message" functionality, which needs to change the course of a conversation at a specific message in the session history.

We should probably switch to a more structured schema with separate tables, especially considering the ongoing developments on extending the session functionality.

Perhaps something like:

A `runs` table to hold the main conversation thread:
- `id`
- `session_id`
- `user_message`
- `assistant_message`
- `created_at`

And a `tool_calls` table to log all the tool activity for a run:
- `id`
- `run_id` (foreign key to `runs`)
- `tool_name`
- `arguments`
- `output`
- `created_at`

This would make our lives a lot easier when it comes to analyzing or debugging conversations.

habema:
Resolves #1385

This PR introduces an optional structured storage mode to `SQLiteSession` to improve the observability and query-ability of conversation histories.

When a user initializes a session with `structured=True`, two new tables are created and populated alongside the existing raw JSON log:
- `agent_conversation_messages`: Stores distinct user, assistant, and system messages.
- `agent_tool_calls`: Records tool call invocations and their corresponding outputs.

This makes it significantly easier to analyze and debug agent interactions using standard SQL, addressing the limitations of querying JSON blobs.

Key points:
- The feature is **opt-in**. Existing implementations are unaffected.
- Foreign key constraints ensure data integrity when items are popped or sessions are cleared.
- Includes comprehensive unit tests and updated documentation.


habema:
Switching this back to a draft as it needs some reviewing

habema:
In true developer fashion, I might have dived a bit too deep. Yet, this all would be very beneficial to my current project.

The problem with my earlier proposed schema is that it lacks a fundamental feature, the ability to connect raw events to each other (user message to assistant message, tool calls to the invoking user message or resulting assistant message(s), usage, etc.), very similarly to tracing functionality.

To sum up the goal, its to make mimic tracing functionality in a structured and query-able manner.

- Added spans and linkage
  - `agent_conversation_messages`: added `parent_raw_event_id`, `trace_id`, `span_id`. User rows keep the agent span; assistant rows are now attributed to the model’s generation/response span.
  - `agent_tool_calls`: added `trace_id`, `span_id`.

 - Introduced `agent_usage` to record per‑response usage: `model`, `requests`, token counts, details, and `trace_id`/`span_id`. Indexed by `(trace_id, created_at)` and `response_id`.

- Configuration
  - The opt‑in flag is now `structured_metadata=True` (formerly `structured=True`).
  - Existing non‑structured behavior is unchanged.

- Necessary tests, docs, and demo are included.

This keeps the feature opt‑in while making stored conversations, tool calls, and usage easy to analyze with SQL, and ensures accurate span attribution for observability.

seratch:
Thanks for sending this PR. I quickly checked the changes and felt the current changes in this PR make the SDK internals way more complex, so I am a bit hesitant to have this.

habema:
Thank you for the feedback. I understand the concern about complexity. 

To clarify:
- Is the core idea of structured session metadata (making conversations queryable via SQL) welcome, but the current implementation too complex? If so, I'd be happy to look into alternative approaches. Perhaps just the basic structured tables (messages, tool calls, usage) without the tracing ingestion, or a implement this as an extension to separate it from the SDK internals.
- Or is the core concept itself considered too complex for the SDK? If the tracing integration is the main concern, I could remove that entirely and focus just on the basic structured storage tables.

As mentioned earlier.
> To sum up the goal, its to make mimic tracing functionality in a structured and query-able manner.

For my usecase, exact `trace_ids` and `span_ids` are not important, I just want to be able to connect every full interaction and its usage.

seratch:
>Is the core idea of structured session metadata (making conversations queryable via SQL) welcome, but the current implementation too complex? 

Yes, it is. As we discussed at the issue, providing an option to effectively use relational database schema is totally fine but I don't think we need tracing integration this time. We already have sessions feature, so I was assuming that we can have yet another solution for the same purpose.

> a implement this as an extension to separate it from the SDK internals.

I haven't checked how this can be at all, but if you have an idea like this in your mind, this could be more clean, plus enhancing the layer could be easier for users too.

PR #1483, opened 2025-08-14
Allow modifying the input sent to the model

rm-openai:
Some customers have reported that the agent loop can go on for a long time and use up the entire context window. This PR allows modifying the data sent to the model.


seratch (approved):
We may want to have some unit tests for the patterns, but the changes look good to me

rm-openai:
> We may want to have some unit tests for the patterns, but the changes look good to me

Whoops, did have tests but forgot to add them. included now.

PR #1486, opened 2025-08-15
docs: Add `strict_json_schema` and `is_enabled` to `FunctionTool` docs

MuhammadHamidRaza:
This pull request improves the `FunctionTool` documentation by including two important parameters that were previously missing.

I've noticed that the `FunctionTool` class in the source code includes `strict_json_schema` and `is_enabled` parameters, but these were not listed in the "Custom function tools" section of the documentation.

This update adds these parameters to the documentation, providing a more complete and accurate reference for developers. This change will make it easier for users to understand the full capabilities of the `FunctionTool` and enhance the overall quality of the SDK's documentation.


seratch:
Thanks for sending this. However, [this reference document page](https://openai.github.io/openai-agents-python/ref/tool/#agents.tool.FunctionTool), which is linked from the page, clearly explain what these properties that are not used in the code snippet in the page. Thus, I don't see the necessity to have this change for now.

PR #1491, opened 2025-08-15
Docs: Add missing docstring to `get_weather` function (in `forcing_tool_use.py`)

Abbas-Asad:
## Summary

Added a missing docstring to the `get_weather` function (in `agent_patterns/forcing_tool_use.py`) to improve code documentation and tool description.

## Problem

The `get_weather` function (tool) was missing a docstring, which is important for:
- Function documentation and code clarity
- Tool description that the LLM uses to understand when to call the tool

## Changes Made

Just added a clear and concise docstring to the `get_weather` function:
```python
@function_tool
def get_weather(city: str) -> Weather:
    """Get the current weather information for a specified city."""
    return Weather(city=city, temperature_range="14-20C", conditions="Sunny with wind")
```

This docstring provides a clear description of the function's purpose and helps both developers and LLMs understand when and how to use this tool.

### DIFF
--- a/examples/agent_patterns/forcing_tool_use.py
+++ b/examples/agent_patterns/forcing_tool_use.py
@@ -44,6 +44,7 @@ class Weather(BaseModel):
 
 @function_tool
 def get_weather(city: str) -> Weather:
+    """Get the current weather information for a specified city."""
     print("[debug] get_weather called")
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. However, in this case, the method name tells what it is, so actually there is no need to have the docstring to make it functional.

PR #1493, opened 2025-08-15
Docs: Add missing docstring to `get_weather` function (in `README.md`)

Abbas-Asad:
## Summary

Added a missing docstring to the `get_weather` function (in `README.md`) in Functions example to improve code documentation and tool description.

## Problem

The `get_weather` function in `README.md` was missing a proper docstring, which is important for:

- Clear function documentation and readability
- Helping language models understand the tool's purpose and when to use it

## Changes Made

Added a concise, one-line docstring to the `get_weather` function in `README.md`:

```python
@function_tool
    """Get the current weather for a given city."""
    return f"The weather in {city} is sunny."
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ from agents import Agent, Runner, function_tool
 
 @function_tool
 def get_weather(city: str) -> str:
+    """Get the current weather for a given city."""
     return f"The weather in {city} is sunny."
 
 


seratch:
Thanks for sending this. However, in this case, the method name tells what it is, so actually there is no need to have the docstring to make it functional.

PR #1497, opened 2025-08-15
Fix: Added SQLiteSession for Persistent Memory to Avoid Repeated Inputs

AyanAhmedKhan381:
### Summary
Implemented a `SQLiteSession` class for persistent session memory storage to resolve the issue of repeated inputs during agent runs.

### Changes
- Added `sqlite_session.py` in `src/agents/memory/`.
- Integrated `SQLiteSession` initialization in session management.
- Ensured memory persistence across multiple runs.

### Impact
- Fixes repeated prompt/input issue.
- Improves developer experience with consistent state retention.

### DIFF
--- a/src/agents/memory/session.py
+++ b/src/agents/memory/session.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-
+from agents.memory.sqlite_session import SQLiteSession
 import asyncio
 import json
 import sqlite3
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sending this patch. However, I don't think we should have a global object creation when loading module here, plus I don't see the necessity to do so.

PR #1509, opened 2025-08-17
Refactor: Update routing.py to use SQLiteSession for session management

IbadUddinSiddqui:
Refactor: Update routing.py to Use SQLiteSession
This pull request refactors the routing.py example to use the built-in SQLiteSession class for managing conversation history. The existing manual handling of the inputs list has been replaced with a more robust and scalable approach.

Why This Change?
The "Handoffs and routing" section aims to demonstrate a practical agentic pattern. Using the SQLiteSession primitive aligns this example with the SDK's recommended best practices for session management. This change:

Simplifies the code: It removes the need for manual message list management, making the core routing logic clearer and easier to follow.

Adds persistence: The example now correctly reflects how a real-world application would handle persistent conversation state, which is essential for multi-turn interactions.

Enhances robustness: The use of a dedicated session primitive makes the example more reliable and scalable.

This refactoring does not change the fundamental behavior of the example—the triage agent still correctly hands off the conversation to the appropriate language agent. Instead, it improves the underlying implementation, making the example a more accurate and valuable resource for developers.

### DIFF
--- a/examples/agent_patterns/routing.py
+++ b/examples/agent_patterns/routing.py
@@ -2,29 +2,11 @@ import asyncio
 import uuid
 
 from openai.types.responses import ResponseContentPartDoneEvent, ResponseTextDeltaEvent
+from agents import Agent, RawResponsesStreamEvent, Runner, trace, SQLiteSession
 
-from agents import Agent, RawResponsesStreamEvent, Runner, TResponseInputItem, trace
-
-"""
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. However, we already have https://github.com/openai/openai-agents-python/blob/main/examples/basic/session_example.py and would like to keep each example as simple as possible. Thus, let us close this one.

PR #1510, opened 2025-08-17
clarify docs for `tool_use_behavior` and simplify description of `stop_on_first_tool` behavior

Abishai-kashif:


Abishai-kashif:
> Thanks, these are helpful for users

You're most welcome @seratch!

PR #1513, opened 2025-08-17
Fix: filename not provided in chatcmpl_converter.py

Nifury:
When a file is provided in base64 encoding, only the `file_data` part is converted to the chat completion API format. 
However, now without the `filename` field, OpenAI returns
```openai.BadRequestError: Error code: 400 - {'error': {'message': "Missing required parameter: 'messages[1].content[1].file.file_id'.", 'type': 'invalid_request_error', 'param': 'messages[1].content[1].file.file_id', 'code': 'missing_required_parameter'}}```

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -271,11 +271,16 @@ class Converter:
                     raise UserError(
                         f"Only file_data is supported for input_file {casted_file_param}"
                     )
+                if "filename" not in casted_file_param or not casted_file_param["filename"]:
+                    raise UserError(
+                        f"filename must be provided for input_file {casted_file_param}"
+                    )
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ant31** [comment]
The error was already explicit. 
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Missing required parameter: 'messages[1].content[1].file.file_id'.", 'type': 'invalid_request_error', 'param': 'messages[1].content[1].file.file_id', 'code': 'missing_required_parameter'}}
```

Openai-agents is not exclusive to GPT.X/openai models, and many models are not requiring a filename to works with file_input.  
This change enforce the to add filename, breaking existing and working code with models not requiring it. 


if the model used is requiring filename, then error is already there and explicit and the enduser knows what to do.

PR #1515, opened 2025-08-18
Update README.md

Fatima367:
Updating the Installation guide for those who will use it with uv.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,16 @@ source .venv/bin/activate  # On Windows: .venv\Scripts\activate
 
 2. Install Agents SDK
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this PR! Indeed, the README needs improvements but let us go with this change: https://github.com/openai/openai-agents-python/pull/1516 Thanks again for your interest and efforts!

Fatima367:
> Thanks for sending this PR! Indeed, the README needs improvements but let us go with this change: #1516 Thanks again for your interest and efforts!

Happy to help!

PR #1523, opened 2025-08-19
Example: using OpenAI Agents for reasoning and summary extraction

princeaden1:
### Summary
This PR adds an example script demonstrating how to use OpenAI Agents for reasoning tasks.

### Changes
- `examples/basic/reasoning.py`:

### Motivation
Provide developers a ready-to-run example for leveraging OpenAI Agents
with structured reasoning responses.

### DIFF
--- a/examples/basic/reasoning.py
+++ b/examples/basic/reasoning.py
@@ -1,0 +1,33 @@
+import asyncio
+
+from agents import Agent, Runner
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. However, we already have https://github.com/openai/openai-agents-python/tree/main/examples/reasoning_content so we don't need this new file for now.

PR #1527, opened 2025-08-19
Add context parameter to run_demo_loop

SeeYangZhi:
Introduces an optional 'context' argument to the run_demo_loop function and passes it to Runner.run_streamed. This allows additional context information to be provided during agent execution for enhanced flexibility.


src/agents/repl.py
```diff
@@ -11,7 +11,7 @@
 from .stream_events import AgentUpdatedStreamEvent, RawResponsesStreamEvent, RunItemStreamEvent
 
 
-async def run_demo_loop(agent: Agent[Any], *, stream: bool = True) -> None:
+async def run_demo_loop(agent: Agent[Any], *, stream: bool = True, context: Any = None) -> None:
```
seratch: Having this additional argument is fine, but we would like to avoid Any type here

SeeYangZhi: Okay thanks, I have updated it to be `TContext | None`

PR #1530, opened 2025-08-19
Handoffs example enhanced

Fatima367:


seratch:
Thanks for sending this. However, we would like to keep the snippet simple, so we don't need this change.

Issue #1397, opened 2025-08-07
GPT-5 does not work well compared to 4.1


sibblegp:
### Debug information
- Agents SDK version: 0.2.4
- Python version 3.13

### Repro steps

I've been using GPT-5 as a drop in for GPT-4.1 for the last hour. It's a disaster. While 4.1 understood my requests, routed to the right agent, and called tools, GPT-5 keeps asking for more specifics and details before it'll call a tool. It's a much worse experience.

Like, for example, I sent in "What is the weather tomorrow night?"

My system prompt has a location in it for each request as well as the current date time and time zone. There is a tool that takes latitude and longitude and a date and gets the hourly forecast.

ChatGPT 4.1 does this 100% correctly. Gets the hourly forecast and gives a great answer.

GPT-5 asks if I mean tomorrow night (with the date) specifically and doesn't call the tool without more input.

It even sometimes say it'll do something but doesn't actually call the tool.

This is an unmitigated disaster and GPT-5 doesn't work at all for agents.

It's also very slow at handing off while actual responses do come back quickly. 4.1 overall is faster.

### Expected behavior
GPT-5 should outperform 4.1 in terms of tool calling but it understand commands and tools less, fails to take the correct actions, and is slower than 4.1

sibblegp:
Two examples. These are queries that 4.1 got right 100% of the time.

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/3b58c4f0-19f5-4384-8ea2-6db4073ffaf8" />

It should know that I mean tomorrow night, Tuesday. Not sure why it's asking for clarification.

<img width="500" src="https://github.com/user-attachments/assets/65aee49c-53ad-41a9-b7ce-304b7d960803" />

In this one it transferred to the agent, but didn't actually make the tool call. This is literally just broken. Again, 4.1 nails this.

sibblegp:
Compare to 4.1's handling of the same question.

<img width="500" src="https://github.com/user-attachments/assets/9e957803-f954-4a2a-bf89-d0fef97f44d5" />

<img width="500" src="https://github.com/user-attachments/assets/99edda0e-c23e-4b67-bace-3b0e344a8c52" />

Clear, concise and fast.

sibblegp:
And for asking for a flight status:

<img width="500" src="https://github.com/user-attachments/assets/904aff9e-9214-4c51-bfcd-a571c58ec14b" />

[...]

PR #1534, opened 2025-08-20
Add a quick opt-in option to switch to gpt-5 model

seratch:
This pull request enables users to quickly switch between gpt-4 and gpt-5 without lots of code changes. GPT-5 is still an opt-in option, so there is no breaking change to existing apps.

For more details, this pull request brings the following changes:
- Introduce OPENAI_DEFAULT_MODEL env variable
- When the default model is any of gpt-5 ones, this SDK automatically sets relevant settings
- Change the default one from gpt-4o to gpt-4.1 (requires minor version bump)
- Improve some of the existing examples

Using this recommended model settings (reasoning.effort = low) may not resolve all the concerns shared at https://github.com/openai/openai-agents-python/issues/1397 but I believe this configuration is the most suitable for running AI agents at this moment.

I will come up with document changes in a separate pull request.


examples/financial_research_agent/agents/search_agent.py
```diff
@@ -12,6 +12,7 @@
 
 search_agent = Agent(
     name="FinancialSearchAgent",
+    model="gpt-4.1",
     instructions=INSTRUCTIONS,
     tools=[WebSearchTool()],
     model_settings=ModelSettings(tool_choice="required"),
```
seratch, line 18: this option is not compatible with gpt-5, so explicitly set gpt-4.1 for this example

examples/hosted_mcp/approvals.py
```diff
@@ -44,7 +44,10 @@ async def main(verbose: bool, stream: bool):
                 print(f"Got event of type {event.item.__class__.__name__}")
         print(f"Done streaming; final result: {result.final_output}")
     else:
-        res = await Runner.run(agent, "Which language is this repo written in?")
+        res = await Runner.run(
+            agent,
+            "Which language is this repo written in? Your MCP server should know what the repo is.",
```
seratch, line 49: gpt-5 models could be confused with this MCP server's rule (= repo information is available in the MCP server URL), so I updated the instructions to be clearer

examples/reasoning_content/main.py
```diff
@@ -36,10 +37,11 @@ async def stream_with_reasoning_content():
     reasoning_content = ""
     regular_content = ""
 
+    output_text_already_started = False
     async for event in model.stream_response(
         system_instructions="You are a helpful assistant that writes creative content.",
         input="Write a haiku about recursion in programming",
-        model_settings=ModelSettings(),
+        model_settings=ModelSettings(reasoning=Reasoning(effort="medium", summary="detailed")),
```
seratch, line 44: Without these settings, reasoning summary items won't show up with OpenAI models; so I've updated this example.

examples/reasoning_content/main.py
```diff
@@ -48,18 +50,16 @@ async def stream_with_reasoning_content():
...
-            print(
-                f"\033[33m{event.delta}\033[0m", end="", flush=True
-            )  # Yellow for reasoning content
+            # Yellow for reasoning content
+            print(f"\033[33m{event.delta}\033[0m", end="", flush=True)
             reasoning_content += event.delta
         elif event.type == "response.output_text.delta":
-            print(f"\033[32m{event.delta}\033[0m", end="", flush=True)  # Green for regular content
+            if not output_text_already_started:
+                print("\n")
```
seratch, line 58: Improved the readability of the output

examples/reasoning_content/runner_example.py
```diff
@@ -35,53 +37,34 @@ async def main():
         result = await Runner.run(
             agent, "What is the square root of 841? Please explain your reasoning."
         )
-
         # Extract reasoning content from the result items
         reasoning_content = None
-        # RunResult has 'response' attribute which has 'output' attribute
```
seratch, line 41: This does not work with OpenAI models, so I revised this example

examples/reasoning_content/runner_example.py
```diff
@@ -35,53 +37,34 @@ async def main():
...
-                    f"\033[32m{event.text}\033[0m", end="", flush=True
-                )  # Green for regular content
-
-        print("\n\nCollected Reasoning Content:")
-        print(reasoning_buffer)
-
-        print("\nCollected Final Answer:")
-        print(content_buffer)
+        output_text_already_started = False
+        async for event in stream.stream_events():
```
seratch, line 57: same as above

examples/tools/file_search.py
```diff
@@ -1,16 +1,43 @@
...
+
+from openai import OpenAI
 
 from agents import Agent, FileSearchTool, Runner, trace
 
 
 async def main():
+    vector_store_id: str | None = None
+
+    if vector_store_id is None:
```
seratch, line 11: Enabled running this example without any preparation

src/agents/agent.py
```diff
@@ -168,10 +173,10 @@ class Agent(AgentBase, Generic[TContext]):
     """The model implementation to use when invoking the LLM.
 
     By default, if not set, the agent will use the default model configured in
-    `openai_provider.DEFAULT_MODEL` (currently "gpt-4o").
+    `agents.models.get_default_model()` (currently "gpt-4.1").
```
seratch, line 176: We can safely change the default one from gpt-4o to gpt-4.1 in the next minor version (0.3.0); TS SDK has been using gpt-4.1 and there is no performance issue with it, plus I believe gpt-4.1 works even better for agentic workflows.

src/agents/models/default_models.py
```diff
@@ -0,0 +1,57 @@
+import copy
```
seratch, line 1: This file is the essential change in this PR. I've added as many comments as possible, but if there are still unclear points, I am happy to add further.

src/agents/agent.py
```diff
@@ -286,6 +291,26 @@ def __post_init__(self):
...
+                    isinstance(self.model, Model)
+                    or gpt_5_reasoning_settings_required(self.model) is False
+                )
+                # The model settings are not customized for the specified model
+                and self.model_settings == get_default_model_settings()
+            )
+        ):
+            # In this scenario, we should use a generic model settings
+            # because non-gpt-5 models are not compatible with the default gpt-5 model settings.
+            # This is a best-effort attempt to make the agent work with non-gpt-5 models.
```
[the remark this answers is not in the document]
seratch: Since it's not possible to check the model name when a Model instance is passed, that's fine to remove the pre-sett model settings in the scenario.

src/agents/agent.py
```diff
@@ -286,6 +291,26 @@ def __post_init__(self):
...
+            self.model_settings = ModelSettings()
+
```
rm-openai, line 313: this seems complex and fragile. Is there something simpler and less invasive we can do?

src/agents/models/openai_provider.py
```diff
@@ -80,7 +82,7 @@ def _get_client(self) -> AsyncOpenAI:
 
     def get_model(self, model_name: str | None) -> Model:
         if model_name is None:
-            model_name = DEFAULT_MODEL
+            model_name = get_default_model()
```
rm-openai, line 85: can we do only this? feels equally useful

src/agents/agent.py
seratch: Yeah, I do understand your point, but this is because of the complicated reality introduced by gpt-5 design. So, I don't think this could be much simpler.

src/agents/models/openai_provider.py
seratch: Setting model_settings is required for this change. default reasoning effort & verbosity does not work for most of agentic workflows because of its latency.

seratch (commented):
I hear your concerns but these changes should be necessary for better dev experience.

src/agents/models/openai_provider.py
rm-openai: ah good point.

src/agents/agent.py
rm-openai: makes sense. Can you add a test as well? To ensure the logic is identical before/after this change, and in various scenarios. Feel free to merge before or after, no need to block on test.

rm-openai (approved):

PR #1538, opened 2025-08-20
docs: Correct and clarify LocalShellTool documentation

MuhammadHamidRaza:
This pull request corrects an inconsistency and clarifies the documentation for the `LocalShellTool`.

Currently, the `LocalShellTool` is listed under the "Hosted tools" section, which is misleading as it does not run on an LLM server. This PR adds a note that explicitly states the tool is not hosted and runs on the local machine. The change also includes a warning about permissions and security, which is critical for users to understand when using a tool that interacts with their local environment.

This update will improve the documentation's accuracy and help developers use the tool safely and correctly.


seratch:
Indeed, the tool is confusing, but instead of updating the page with more words, we prefer updating the docstring to have the detailed document page link: https://platform.openai.com/docs/guides/tools-local-shell

PR #1548, opened 2025-08-21
Performance: only create the OpenAIRealtimeServerEvent TypeAdapter once

michieldwitte:
For every event a new TypeAdapter is created, which has a significant performance impact.
image
Creating it once and reusing it, makes event handling a lot faster.


rm-openai (approved):
This is a great PR. Thanks. I think you might have to rebase. Btw could also make it a module level constant (lazy loaded) instead of once per-instance (but no pressure),

rm-openai (approved):

PR #1550, opened 2025-08-21
Save session on turn rather than at final response

spokeydokeys:
The current implementation of sessions saves the session on the final output. This works as a "memory" of previous full conversations, but does not work if you need access to the session during the turn. A use case for this is providing access to the full conversation to an agent-as-tool during the turn.

This PR updates the `run` and `_start_streaming` methods in `run.py` to save the session on each turn.

The basic flow is:
- On turn start, save the input into the session.
- After each turn, add only the turn results to the session.

Also included is an example file that shows the use of this pattern. It is an adapted version of the agents-as-tools example where a check agent is called within the same turn as the translation agents and examines their outputs via the conversation history, rather than as a generated input.

[x] Current tests pass and I don't think an update to the tests is needed for this PR.
[x] Documentation about sessions did not cover this case, but I felt it was implied that it would work. I think these changes bring the behavior in line with the documentation.
[x] I have added an example that demonstrates it in action.


spokeydokeys:
Addresses issue #1551

examples/agent_patterns/agents_as_tools_with_history.py
```diff
@@ -0,0 +1,116 @@
+import asyncio
```
seratch: instead of having this example, we need unit tests verifying the pattern

seratch (changes requested):
Thanks for taking the time to send this. At a glance, the changes look good, but we need unit tests verifying the patterns and we don't need the additional example.

spokeydokeys:
I removed the example and added a unit test to `test_agent_runner.py`.

There were a few linting and formatting issues that `make lint` and `make check-format` brought up that weren't in my changes, but I let ruff fix them and included them in the last commit.  Happy to drop those if you'd rather leave them.

src/agents/realtime/model_events.py
```diff
@@ -84,6 +84,7 @@ class RealtimeModelInputAudioTranscriptionCompletedEvent:
 
     type: Literal["input_audio_transcription_completed"] = "input_audio_transcription_completed"
 
+
```
seratch: can you remove these unrelated changes to make the review easier?

spokeydokeys: > can you remove these unrelated changes to make the review easier?

Of course! I removed the final commit that had all the `make lint/check-format` changes that weren't part of the bug fix.

spokeydokeys:
@seratch, I made the changes you were looking for.  Would you like me to rebase or update the branch before you review?

seratch (approved):

PR #1555, opened 2025-08-22
v0.2.9

rm-openai:


seratch (approved):
As mentioned in https://github.com/openai/openai-agents-python/pull/1534 description, I’ve changed the default model from 4o to 4.1. If you think it should bump minor, that’s fine too. I can approve again.

Issue #798, opened 2025-06-02
Feature Request: Enhanced Run Lifecycle Management - Interrupt and Update Active Runs


chiehmin-wei:
**Is your feature request related to a problem? Please describe.**
The current `Runner.run` and `Runner.run_streamed` methods in `src/agents/run.py` execute an agent workflow to completion and return a final `RunResult` or `RunResultStreaming`. This design doesn't allow for real-time interruption of an ongoing run or the injection of new information that could alter the agent's subsequent actions. This is a limitation for dynamic use cases like on-call incident response, where an incident might be resolved mid-investigation or new critical findings might emerge.

**Describe the solution you'd like**
We propose enhancements to the `Runner` and its associated components to support a more dynamic run lifecycle:

1.  **Cancellable/Interruptible Runs:**
    * **Need:** The `Runner.run` and `Runner.run_streamed` methods should return an object (e.g., an `ActiveRun` or `Task`-like handle) that exposes a method to signal cancellation (e.g., `active_run.cancel()`).
    * **How it could work:**
        * The `while True` loop within `Runner.run` (and the equivalent in `_run_streamed_impl`) would need to check an internal flag or an `asyncio.Event` set by the `cancel()` method on the returned handle.
        * Upon detecting a cancellation request, the loop would terminate gracefully, potentially allowing for a final cleanup span and then raising a specific `RunCancelledException` or returning a `RunResult` indicating cancellation.
        * For `run_streamed`, the `_run_impl_task` (an `asyncio.Task`) created in `Runner.run_streamed` seems like a natural candidate to be cancelled. The `RunResultStreaming` object could expose a method that cancels this underlying task.

2.  **Dynamically Add Information to an Active Run:**
    * **Need:** A mechanism to inject new information (e.g., new user messages or tool-like observations) into an active run, so the agent can consider it in subsequent turns.
    * **How it could work:**
        * The object returned by `run`/`run_streamed` could also expose a method like `active_run.add_information(item: TResponseInputItem)` or `active_run.update_history(items: list[TResponseInputItem])`.
        * Internally, this method would need to append to a queue or a shared list that `_run_single_turn` (or its streaming equivalent) checks before preparing the `input` for `model.get_response` or `model.stream_response`.
        * Specifically, the `input` variable within `_run_single_turn` or `_run_single_turn_streamed` (which is currently built from `original_input` and `generated_items` or `streamed_result.input` and `streamed_result.new_items`) would incorporate this externally added information before the next model invocation.
[...]

PR #1557, opened 2025-08-22
feat(run): lifecycle interrupt + live input injection; cancel-aware tools; safer tracing

vrtnis:
#### Included in this PR (This work addresses #798) :

`active_run` handle on `Runner.run` and `Runner.run_streamed`

`cancel(reason)` for a clean stop

`inject(items)` included on the next turn

`state()` for a tiny debug snapshot

Streamed runs emit a cancelled status event and close the queue; non-streamed return a terminal result without final output.

Public APIs stay the same; tracing and hook order are preserved for normal completions.

#### Tests
New tests cover cancel, inject, and status emission in both streamed and non-streamed paths.

#### RunState + run_single_step (next step)

What I'm thinking is that this PR lays the groundwork for a proper step API while solving the immediate problem of cancellation and live input injection. The follow-up I’m proposing could perhaps follow this path (all open to discussion!):

Introduce a light `RunState` that captures current agent, turn, items, and bookkeeping.

Expose `run_single_step(state)` -> `new_state` that does “call model → run tools/handoffs → decide next”.

Keep Runner.run* as a thin loop over run_single_step for backward compatibility.

Add step-level tests and examples so folks can drive agents interactively.

If this direction sounds good, I can open a focused PR right after this one that sketches `RunState` and a minimal `run_single_step` while maintaining today’s surface.

Happy to make changes to better align with roadmap.


seratch:
Thanks for sending this PR. This pull request brings lots of changes, so I am not sure if it really resolves the issue/need. I don't have the bandwidth to check this PR right now, but our team may be able to look into this soon.

vrtnis:
All good, no worries, :) happy to jump in whenever the team has time and it aligns with the roadmap

tauzhang-cn:
this is a very helpful feature, thank you for the PR!

PR #1568, opened 2025-08-23
Docs: Clarify installation instructions for uv tool in README

SeeYangZhi:
This PR updates the README to clarify the installation instructions for using the [uv](https://docs.astral.sh/uv/) tool when setting up the OpenAI Agents SDK.

Changes
- Rephrased the note under the uv section to make the installation process more explicit and clear.
- Updated the wording from “the tool would be even similar” to a clearer explanation that installing the OpenAI Agents SDK with uv is straightforward.

The previous wording was ambiguous (I'm not sure what "even similar" is referring to) and might confuse users unfamiliar with uv. This update improves clarity and makes the instructions easier to follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ For voice support, install with the optional `voice` group: `pip install 'openai
 
 ### uv
 
-If you're familiar with [uv](https://docs.astral.sh/uv/), using the tool would be even similar:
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the suggestion. Being similar there means you can do the same with uv too. If a user is not familiar with uv and want to start without uv anyways, they can just skip the part. So, the current wording should be fine. (I do understand once you start using uv, you never go back to directly using venv tho!)

PR #1569, opened 2025-08-23
docs: clarify ToolsToFinalOutputResult docstrings

MuhammedSuhaib:
- Reworded `is_final_output` docstring for clarity.
- Made rules for `final_output` explicit:
  * Must be non-None and match `output_type` when `is_final_output=True`.
  * Must be None when `is_final_output=False`.
- No functional changes, docs only.


seratch:
Thanks for the suggestion, but I don't think the current one is so bad. Let me skip the change this time.

PR #1570, opened 2025-08-23
docs: clarify ModelSettings.truncation behavior

MuhammedSuhaib:
docs: clarify ModelSettings.truncation behavior

- Expanded the docstring for `ModelSettings.truncation`.
- Documented the behavior of:
  * "auto": drops middle conversation items if input exceeds context window, request still succeeds and streaming continues normally.
  * "disabled": (default) request fails with 400 if input exceeds context window.
  * None: defers to provider default (typically "disabled").
- Aligns SDK docs with the official API reference.

### DIFF
--- a/src/agents/model_settings.py
+++ b/src/agents/model_settings.py
@@ -93,8 +93,16 @@ class ModelSettings:
 
     truncation: Literal["auto", "disabled"] | None = None
     """The truncation strategy to use when calling the model.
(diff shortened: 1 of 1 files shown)


seratch:
The current one has the link to the document, so it should be sufficient. Also, we'd like to avoid duplicated list there for this property.

PR #1571, opened 2025-08-23
fix(agent): stricter validation for Agent.prompt type checking

muhammadsami987123:
This PR improves the validation logic for the Agent.prompt field in agent.py.

🔧 Changes

⚫ Replaced hasattr(self.prompt, "get") check with a strict isinstance(self.prompt, Prompt) check.

⚫ Ensures only valid Prompt objects or DynamicPromptFunction (callable) are accepted.

⚫ Prevents invalid objects (e.g., dict with .get() method) from being mistakenly allowed.

✅ Why this matters

⚫ Increases type safety and prevents runtime errors when an invalid prompt object is passed.

⚫ Makes the behavior more predictable and aligned with the intended usage (Prompt | DynamicPromptFunction | None).

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -28,6 +28,7 @@ from .run_context import RunContextWrapper, TContext
 from .tool import FunctionTool, FunctionToolResult, Tool, function_tool
 from .util import _transforms
 from .util._types import MaybeAwaitable
+from .prompts import Prompt, DynamicPromptFunction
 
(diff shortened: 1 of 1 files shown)


seratch:
Your code's Python indentation looks broken. But more importantly, Prompt class is TypedDict, so I don't think simply checking this way can work out well. Thanks for your interest and sending this!

PR #1582, opened 2025-08-25
Docs: Improve example for custom Session class implementation.

luke-wriglesworth:
### Summary

- Removes the unused import of Session
- Update type annotations of MyCustomSession to match the Session protocol

### Test plan

This is a documentation only change.

### Checks

### DIFF
--- a/docs/sessions.md
+++ b/docs/sessions.md
@@ -227,7 +227,7 @@ if __name__ == "__main__":
 You can implement your own session memory by creating a class that follows the [`Session`][agents.memory.session.Session] protocol:
 
 ```python
-from agents.memory import Session
+from agents.items import TResponseInputItem
 from typing import List
 
 class MyCustomSession:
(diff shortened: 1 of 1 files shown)


docs/sessions.md
```diff
@@ -227,7 +227,7 @@ if __name__ == "__main__":
 You can implement your own session memory by creating a class that follows the [`Session`][agents.memory.session.Session] protocol:
 
 ```python
-from agents.memory import Session
+from agents.items import TResponseInputItem
 from typing import List
 
 class MyCustomSession:
```
seratch: This class should inherit SessionABC

luke-wriglesworth: I made the requested change but noticed that the SessionABC docstring says it should be inherited only for internal implementations. Should we also update this docstring? 

https://github.com/openai/openai-agents-python/blob/86e26e975f57d396f577c7d1915ae2f323ef7888/src/agents/memory/session.py#L58-L66

seratch (approved):

PR #1586, opened 2025-08-26
Add reasoning text delta event support for gpt-oss models

seratch:
Since TS SDK has its own protocol layer, the way to handle data is a bit different but this aligns with https://github.com/openai/openai-agents-js/pull/292


rm-openai (approved):

diwu-sf:
This is awesome, was just trying to debug and get reasoning to work through Cerebras GPT OSS 120b and this fixes the problem

PR #1587, opened 2025-08-26
Fix #1564 Add conversations API support

### LINKED ISSUE #1564: Add conversations API support
I will work on this early next week:

https://platform.openai.com/docs/guides/conversation-state?api-mode=responses#using-the-conversations-api

seratch:
This pull request resolves #1564


rm-openai (commented):
This generally looks good to me. A couple of quick questions in case you've thought about them:
1. It's possible that the user doesn't want every LLM call or every agent to read from or write to the conversation. e.g. imagine the setup is classify the query -> route to an agent. Maybe you only want the 2nd one to be reading/writing from conversation. How would that be enabled?
2. This is somewhat tangential but curious if you think the Sessions concept should have a conversations based wrapped?

seratch:
>It's possible that the user doesn't want every LLM call or every agent to read from or write to the conversation. e.g. imagine the setup is classify the query -> route to an agent. Maybe you only want the 2nd one to be reading/writing from conversation. How would that be enabled?

If these two runs are separate (i.e., deterministic workflow that just sequentially runs two agents), that's simple. If you mean handoffs, indeed it's not feasible with this feature addition. That's the same for session as well. I am not sure if attaching conversation_id when instantiating an `Agent` object is a right solution here. Probably not. Then, I don't have any other ideas. 

I think the immediate suggestion from us would be you can use conversation_id, previous_response_id, and session per an agent run.

>This is somewhat tangential but curious if you think the Sessions concept should have a conversations based wrapped?

Conversations API based sessions implementation is worth considering, and it should be feasible. If a user is fine with frequently calling the API endpoints, it should be an good option. I can explore the implementation tomorrow or later this week.

src/agents/run.py
```diff
@@ -239,6 +243,7 @@ async def run(
             run_config: Global settings for the entire agent run.
             previous_response_id: The ID of the previous response, if using OpenAI models via the
                 Responses API, this allows you to skip passing in input from the previous turn.
+            conversation_id: The ID of the stored conversation, if any.
```
rm-openai: Would recommend expanding on this documentation.
```suggestion
            conversation_id:  The conversation ID (https://platform.openai.com/docs/guides/conversation-state?api-mode=responses). If provided, the conversation will be used to read and write items. Every agent will have access to the conversation history so far, and it's output items will be written to the conversation. We recommend only using this if you are exclusively using OpenAI models; other model providers don't write to the Conversation object, so you'll end up having partial conversations stored.
```

pyproject.toml
```diff
@@ -7,7 +7,7 @@ requires-python = ">=3.9"
 license = "MIT"
 authors = [{ name = "OpenAI", email = "support@openai.com" }]
 dependencies = [
-    "openai>=1.99.6,<2",
+    "openai>=1.101.0,<2",
```
seratch:
```suggestion
    "openai>=1.102.0,<2",
```

src/agents/memory/session.py
```diff
@@ -102,268 +97,3 @@ async def pop_item(self) -> TResponseInputItem | None:
     async def clear_session(self) -> None:
         """Clear all items for this session."""
         ...
-
-
-class SQLiteSession(SessionABC):
```
seratch, line 107: moved this to sqlite_session.py

src/agents/memory/openai_session.py
```diff
@@ -0,0 +1,79 @@
...
+        _openai_client = get_default_openai_client() or AsyncOpenAI()
+
+    response = await _openai_client.conversations.create(items=[])
+    return response.id
+
+
+class OpenAISession(SessionABC):
+    def __init__(
+        self,
+        session_id: str | None = None,
```
seratch: use conversation_id as session_id; you can pass the id if you already have, but if you don't, this class creates a new one under the hood

examples/memory/sqlalchemy_session_example.py
```diff
@@ -20,28 +20,56 @@ async def main():
         create_tables=True,
```
seratch, line 20: moved to memory dir and made the code consistent with others

src/agents/memory/openai_session.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class OpenAISession(SessionABC):
+    def __init__(
+        self,
+        session_id: Optional[str] = None,
+        openai_client: Optional[AsyncOpenAI] = None,
+    ):
+        # this implementation allows to set this value later
+        self.session_id = session_id  # type: ignore
```
seratch: using conversation_id as session_id; you can pass if you already have; otherwise, this class generates a new one under the hood

src/agents/memory/sqlite_session.py
```diff
@@ -0,0 +1,275 @@
...
+import json
+import sqlite3
+import threading
+from pathlib import Path
+
+from ..items import TResponseInputItem
+from .session import SessionABC
+
+
+class SQLiteSession(SessionABC):
```
seratch, line 13: moved from session.py

seratch (commented):
comments for reviwers

src/agents/memory/openai_session.py
```diff
@@ -0,0 +1,84 @@
...
+async def start_openai_session(openai_client: Optional[AsyncOpenAI] = None) -> str:
+    _openai_client = openai_client
+    if openai_client is None:
+        _openai_client = get_default_openai_client() or AsyncOpenAI()
+
+    response = await _openai_client.conversations.create(items=[])  # type: ignore [union-attr]
+    return response.id
+
+
+class OpenAISession(SessionABC):
```
rm-openai:
```suggestion
class OpenAIConversationsSession(SessionABC):
```

src/agents/memory/openai_session.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(
+        self,
+        session_id: Optional[str] = None,
```
rm-openai:
```suggestion
        conversation_id: str | None = None,
```

src/agents/memory/openai_session.py
```diff
@@ -0,0 +1,84 @@
...
+        openai_client: Optional[AsyncOpenAI] = None,
```
rm-openai:
```suggestion
        openai_client: AsyncOpenAI | None = None,
```

Also feels kinda weird to ask for a client here when in most places we don't require this?

rm-openai (commented):
this generally looks good but feels like there are too many `type:ignores` here

src/agents/memory/openai_session.py
seratch: somehow, this fails with python 3.9 tests; I had to use optional in this file

src/agents/memory/openai_session.py
seratch: having to create an instance isn't not great when you're using default openai client

seratch:
> feels like there are too many type:ignores here

yeah, but it was inevitable to call model_dump() to make the returned objects to be serializable

src/agents/memory/openai_session.py
seratch: renamed

src/agents/memory/openai_session.py
seratch: i don't know what's the trigger of the A | B error with python 3.9 but for this one, let me go with Optional[A]. When we drop python 3.9 support, we can make it consistent

src/agents/memory/openai_session.py
seratch: forgot to mention other patterns; if you want to pass a client for a certain reason (say, you manage multiple instances for multi tenant apps?), accepting this should be helpful for those.

seratch:
@rm-openai renamed the class and resolved most of the type ignores. let me know if you see anything else to change; otherwise, you can merge at any time

src/agents/memory/openai_session.py
rm-openai: You can just add `from __future__ import annotations` to the top of the file for these

seratch: Ah I see. will do

seratch: OK, this is done ✅

src/agents/memory/openai_conversations_session.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+_EMPTY_SESSION_ID = ""
+
+
+class OpenAIConversationsSession(SessionABC):
+    def __init__(
+        self,
+        *,
+        session_id: str | None = None,
```
rm-openai: can you rename to `conversation_id` please?

src/agents/memory/openai_conversations_session.py
```diff
@@ -0,0 +1,92 @@
...
+        openai_client: AsyncOpenAI | None = None,
+    ):
+        # this implementation allows to set this value later
+        self.session_id = session_id or _EMPTY_SESSION_ID
```
rm-openai: a better pattern is:
```
self._session_id: str | None
```

with
```
async def _get_session_id(self) -> str:
  if not self._session_id:
    self._session_id = await ...
  return self._session_id
```

src/agents/memory/openai_conversations_session.py
```diff
@@ -0,0 +1,92 @@
...
+        _openai_client = openai_client
+        if _openai_client is None:
+            _openai_client = get_default_openai_client() or AsyncOpenAI()
+        # this never be None here
+        self.openai_client: AsyncOpenAI = _openai_client
```
rm-openai: generally prefix with _

src/agents/models/openai_responses.py
```diff
@@ -149,7 +151,8 @@ async def stream_response(
         output_schema: AgentOutputSchemaBase | None,
         handoffs: list[Handoff],
         tracing: ModelTracing,
-        previous_response_id: str | None,
+        previous_response_id: str | None = None,  # unused
+        conversation_id: str | None = None,  # unused
```
rm-openai: seems used to me? line 170/171

seratch: ah, ai editor did extra...

surfingdev:
This is exactly what I was looking for. I see this is already out but there is no documentation for it right? cc @seratch

Issue #985, opened 2025-07-01
Error "Item ‘rs_ABCD’ of type ‘reasoning’ was provided without its required..." when using CodeInterpreter


molefrog:
### Describe the bug
We have started to experience error of type `"Item ‘rs_ABCD’ of type ‘reasoning’ was provided without its required following item"` when an agent does a handoff to another agent (`o4-mini` with `low` effort to `o4-mini` with `medium` effort). 

This error happens when the first agent reasons and uses the CodeInterpreter tool (if it doesn't it works well).

This might be related to [this issue](https://community.openai.com/t/code-interpreter-reasoning-and-previous-response-id-bug-item-of-type-reasoning-was-provided-without-its-required-following-item/1279124) from the Developer Community.

### Debug information
- Agents SDK version: (e.g. `v0.1.0`)
- Python version (e.g. Python 3.11)

fitzjalen:
Same issue

seratch:
Thank you for taking the time to report this and we're sorry about the disruption. At a glance, this looks like an issue on the API server side. If there is anything to change on this SDK side, we as the maintainer of this SDK will quickly work on it too.

molefrog:
> Thank you for taking the time to report this and we're sorry about the disruption. At a glance, this looks like an issue on the API server side. If there is anything to change on this SDK side, we as the maintainer of this SDK will quickly work on it too.

Thank you for your reply. This does indeed look like a regression in the API, since we were using the same (latest) version of the SDK before the error appeared. 

I did a simple test yesterday and I can confirm that the issue still persists. I don't know if this can be helpful, but this discussion might be related:
https://github.com/openai/openai-agents-python/issues/722

Kalai-off:
At a first glance it seem to sdk issues. kindly check with zero data retention policy. this will resolve this issue

Tadhgus:
Hit this today also when I started using the code interpreter via Azure. As soon as the agent finishes using the code interpreter and calls another tool the error response is generated.

norika1207-lab:
One practical way to make this easier to evaluate is to turn the report into a small contract test around the runtime boundary, not only the user-facing symptom.

For `CodeInterpreter reasoning item error after tool use`, I would check three things: first, the failing input is preserved as a minimal fixture; second, the SDK emits the same observable event/result shape before and after the fix except for the intended change; third, the trace or serialized session contains enough metadata to explain what happened after a resume or retry.

[...]

PR #1593, opened 2025-08-27
fix(responses): normalize reasoning items by inserting placeholder as…

fitzjalen:
Bugfix for https://github.com/openai/openai-agents-python/issues/985

Fix: Ensure every reasoning item in Responses API input is immediately followed by a valid assistant follower. Inserts a minimal placeholder assistant message after any orphan reasoning item (i.e. not followed by message / function_call / code_interpreter_call). Prevents 400 errors: “Item '<id>' of type 'reasoning' was provided without its required following item.” Preserves original reasoning items (for later references) and generates unique placeholder IDs. Includes defensive fail‑open behavior (logs and proceeds if normalization fails). 
All tests pass (679).


seratch:
Thanks for sending this. However, I believe this issue should be resolved in a different way (ideally on the server-side), and we're hesitant to have this kind of workaround on the client side SDK.

fitzjalen:
> Thanks for sending this. However, I believe this issue should be resolved in a different way (ideally on the server-side), and we're hesitant to have this kind of workaround on the client side SDK.

Thanks for your reply! I fully agree that the proper fix should ideally be handled server-side, and I understand the hesitation to introduce this kind of workaround in the client SDK.

That said, the issue is still present right now, and in my case it’s a critical blocker that breaks LLM calls. For anyone else running into the same problem, this patch provides a temporary way forward until an upstream fix is available.

So yes, I’m on the same page with you — but in the meantime, this workaround may help others keep their projects moving.

PR #1613, opened 2025-08-29
Added .run() method to Agent as a shortcut to Runner.run_sync.

Syed-Hamza-Ali-8:
This PR introduces a convenient .run() method for the Agent class, acting as a shortcut for Runner.run_sync(agent, ...).

Changes included:

Added .run() method to Agent for simplified synchronous execution.

Added tests/test_agent_run.py to validate:

.run() alias functionality.

Direct Runner.run_sync execution.

Mocked Runner.run_sync in tests to ensure deterministic results without calling real LLM models.

Benefits:

Developers can now call agent.run("prompt") directly instead of always using Runner.run_sync.

Test coverage ensures both the alias and direct execution work correctly.

Example Usage:

result = agent.run("What is 7 plus 11?")
print(result.final_output)  # Expected output: "18"

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -458,3 +458,14 @@ class Agent(AgentBase, Generic[TContext]):
     ) -> ResponsePromptParam | None:
         """Get the prompt for the agent."""
         return await PromptUtil.to_model_input(self.prompt, run_context, self)
+    
+    def run(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the suggestion. I agree this could be convenient, but for design consistency, we won't have the changes this time.

PR #1620, opened 2025-08-29
Fix trace input validation

Syed-Hamza-Ali-8:
This PR adds strict type validation for trace() parameters:

- workflow_name must be str
- trace_id must be str or None
- group_id must be str or None
- metadata must be dict or None

Previously, passing an integer as workflow_name would still execute tracing and return output, but cause a non-fatal 400 error from the tracing client.

Example :
```
    with trace(workflow_name=1):
        first_result = await Runner.run(agent, "Tell me a joke")
        second_result = await Runner.run(
            agent, f"Rate this joke: {first_result.final_output}"
        )
        print(f"Joke: {first_result.final_output}")
        print(f"Rating: {second_result.final_output}")
```        
And Output is :
     Joke: Okay, here's one:

Why don't scientists trust atoms?   

... Because they make up everything!

Rating: Okay, I've got a few to choose from, so let's see...

**Rating:** I'd give that joke a **6/10**. Here's why:

*   **Pros:** It's a classic pun, and the setup is simple and relatable. The punchline is quick and delivers the pun effectively. It's also generally inoffensive, so good for mixed company.
*   **Cons:** It's a pretty well-known joke, so it might not get a big laugh from everyone. The pun itself is fairly obvious, reducing the surprise factor.

Overall, it's a solid, reliable joke that will likely get a chuckle, but it's not going to bring the house down.

```
[non-fatal] Tracing client error 400: {
  "error": {
    "message": "Invalid type for 'data[0].workflow_name': expected a string, but got an integer instead.",
    "param": "data[0].workflow_name",
    "code": "invalid_type"
  }
}
```

Now, incorrect types will raise TypeError immediately, preventing invalid requests and making the API safer.

Example:
with trace(workflow_name="my_workflow"):

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -458,3 +458,14 @@ class Agent(AgentBase, Generic[TContext]):
     ) -> ResponsePromptParam | None:
         """Get the prompt for the agent."""
         return await PromptUtil.to_model_input(self.prompt, run_context, self)
+    
+    def run(self, *args, **kwargs):
+        """
+        Alias for Runner.run_sync.
+
+        Example:
+            agent.run("Hello")  # Same as Runner.run_sync(agent, "Hello")
+        """
+        from .run import Runner
+        return Runner.run_sync(self, *args, **kwargs)
+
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for sending this, but we don't need this change.

PR #1621, opened 2025-08-29
Update index.md

Abdulbasit110:
added uv add command in the intro page


seratch:
Thanks for sending this, but we may consider enhancing the document section in a different way in the future

Issue #991, opened 2025-07-02
Input guardrail tripwire allows tool execution to continue after exception is raised


adeel0x01:
## Bug Description
When an input guardrail triggers a tripwire and raises `InputGuardrailTripwireTriggered`, the agent execution should stop immediately. However, tools are still being executed after the guardrail exception is raised, as evidenced by OpenAI API traces.

## Expected Behavior
- Input guardrail evaluates the input
- If tripwire is triggered, `InputGuardrailTripwireTriggered` exception is raised
- Agent execution stops immediately
- No tools should be called

## Actual Behavior
- Input guardrail correctly triggers and raises exception
- Tools are still executed after the exception (visible in OpenAI dashboard traces)
- This creates unexpected behavior and potential security/cost implications

## Environment
- openai-agents version: 0.1.0
- Python version: 3.13+
- Operating System: Windows

## Steps to Reproduce
1. Create an agent with an input guardrail that triggers a tripwire
2. Create an agent with tools
3. Run the agent with input that should trigger the guardrail
4. Observe that exception is raised but tools are still called in OpenAI traces

## OpenAI Trace
![Image](https://github.com/user-attachments/assets/06953053-13f9-45b9-b1c8-48b97cbedc45)

## Additional Context
This issue was discovered while testing guardrails functionality. The guardrail logic works correctly (proper exception is raised), but the execution flow doesn't respect the guardrail's decision to stop processing.

seratch:
If I understand this correctly, this seems to be the same issue with https://github.com/openai/openai-agents-python/issues/889; https://github.com/openai/openai-agents-python/pull/931 will resolve the issue.

rjmcmillan:
> If I understand this correctly, this seems to be the same issue with [#889](https://github.com/openai/openai-agents-python/issues/889); [#931](https://github.com/openai/openai-agents-python/pull/931) will resolve the issue.

I'm seeing the same thing, unless I'm understanding the intended behaviour incorrectly. I just followed the homework assignment tutorial. The guardrail tripwire triggers, but it then proceeds to trigger the next agent. I don't think it's specifically to do with tools.

<img width="1671" height="764" alt="Image" src="https://github.com/user-attachments/assets/02813c52-f2f1-4ebe-b8a4-b451dc3f0a35" />

Ai80-8:
Stop

lionelchg:
[...]

PR #1622, opened 2025-08-30
Ensure that input_guardrails can block model/tools from running

### LINKED ISSUE #889: FileSearchTool runs despite InputGuardrailTripwireTriggered

lionelchg:
## Summary

- This PR has been started from #931 but due to too many conflicts with `main` branch it has been redone from scratch. It should close #889 and #991 
- Add optional blocking behavior for input guardrails so they can either gate the first model call and tool execution
or run in parallel.
- Fix a streaming hang by adding a short timeout when reading the event queue so guardrail tripwires and background
task errors surface even if no model events are produced.

## Motivation

- Some guardrails must run before any downstream LLM/tool activity (e.g., safety tripwires), while others can run
concurrently (e.g., telemetry or advisory checks). Previously, all input guardrails ran in parallel with the first
model call, which meant tripwires could trigger after downstream actions had already started.
- In streaming mode, if a blocking guardrail triggers before the model produces any events, the consumer could hang
awaiting events that never arrive.

## Key Changes

- Input Guardrails API:
    - src/agents/guardrail.py
    - Add `block_downstream_calls: bool = True` to `InputGuardrail`.
    - Extend `@input_guardrail` decorator with `block_downstream_calls` (default `True` for safety/back-compat).
- Runner (non-streamed):
    - src/agents/run.py
    - On first turn, separate input guardrails into blocking vs non-blocking.
    - If any blocking guardrails exist:
      - Await them to complete before the first model call and tool execution.
      - Run non-blocking guardrails concurrently with the model call/tools.
    - If none blocking:
      - Run all guardrails concurrently with the model call/tools as before.
    - Helpers:
      - `_separate_blocking_guardrails(...)`: returns `(blocking, non_blocking)`.
      - `_get_model_response_only(...)`: obtains the model response without executing tools.
      - `_execute_tools_from_model_response(...)`: runs tool execution/side effects from a model response.
- Runner (streamed):
    - src/agents/run.py
    - On first turn, run blocking guardrails to completion first and push their results to the guardrail queue. If a
tripwire triggers, the run task exits and the stream consumer detects the error.
    - Start non-blocking guardrails in the background queue (or all guardrails if none are blocking).
    - Append guardrail results to `RunResultStreaming.input_guardrail_results` to avoid overwriting.
- Streaming Hang Fix (important):
    - src/agents/result.py
    - In `RunResultStreaming.stream_events()`, wrap `self._event_queue.get()` in a short timeout:
      - `item = await asyncio.wait_for(self._event_queue.get(), timeout=0.1)`
      - On `TimeoutError`, loop continues and re-checks for stored exceptions and guardrail tripwires.
    - Why: If a blocking input guardrail triggers before any model events are streamed, the event queue may stay empty
forever. The timeout wakes the loop to:
      - Drain `_input_guardrail_queue` and raise `InputGuardrailTripwireTriggered`.
      - Detect exceptions stored from the background run task (e.g., tripwire raised) and surface them.
    - Effect: Prevents indefinite hangs when there are no model events. Normal streaming behavior is unchanged; the
consumer still yields events as they arrive.
- Tests:
    - Add tests covering:
    - Blocking guardrail prevents the first model call.
    - Blocking guardrail delays tool execution until completion.
    - Non-blocking guardrail allows parallel tool execution with tools.
    - Mixed blocking and non-blocking ordering.
    - Streamed: blocking guardrail prevents starting model streaming.
- Small adjustments in existing tests to attach a model when asserting model behavior.

## Behavior

- Default behavior is unchanged for existing users: guardrails default to block_downstream_calls=True, preserving the
safest behavior.
- Opt-in parallelism: set block_downstream_calls=False to allow a guardrail to run concurrently with the first model
call/tools.
- Streaming never hangs waiting for model events that never come due to early guardrail tripwires.
    - When idle (no events), the loop wakes 10 times/second to re-check errors, which is lightweight and ensures
responsiveness.

## Tests

- Unit tests added/updated; run with make tests.
- Specific reproduction fixed: tests/
test_agent_runner_streamed.py::test_input_guardrail_tripwire_triggered_causes_exception_streamed previously hung; with
the timeout it surfaces the tripwire and exits as expected.

## Files Touched

- src/agents/guardrail.py: New flag + decorator param.
- src/agents/run.py: Guardrail separation, gating logic, streamed behavior, helper methods, accumulate results.
- src/agents/result.py: Short timeout in stream_events() to avoid hangs.
- tests/*: New and adjusted tests.

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -97,6 +97,13 @@ class InputGuardrail(Generic[TContext]):
     function's name.
     """
 
+    block_downstream_calls: bool = True
+    """Whether this guardrail should block downstream calls until it completes.
+    If any input guardrail has this set to True, the initial model call and any
+    subsequent tool execution will be delayed until all blocking guardrails finish.
+    Defaults to True for backwards compatibility and safety.
+    """
+
     def get_name(self) -> str:
         if self.name:
             return self.name
@@ -209,6 +216,7 @@ def input_guardrail(
 def input_guardrail(
     *,
     name: str | None = None,
+    block_downstream_calls: bool = True,
 ) -> Callable[
     [_InputGuardrailFuncSync[TContext_co] | _InputGuardrailFuncAsync[TContext_co]],
     InputGuardrail[TContext_co],
@@ -221,6 +229,7 @@ def input_guardrail(
     | None = None,
     *,
     name: str | None = None,
+    block_downstream_calls: bool = True,
 ) -> (
     InputGuardrail[TContext_co]
     | Callable[
@@ -235,7 +244,7 @@ def input_guardrail(
         @input_guardrail
         def my_sync_guardrail(...): ...
 
-        @input_guardrail(name="guardrail_name")
+        @input_guardrail(name="guardrail_name", block_downstream_calls=False)
         async def my_async_guardrail(...): ...
     """
 
@@ -246,6 +255,7 @@ def input_guardrail(
             guardrail_function=f,
             # If not set, guardrail name uses the function’s name by default.
             name=name if name else f.__name__,
+            block_downstream_calls=block_downstream_calls,
         )
 
     if func is not None:
--- a/src/agents/result.py
+++ b/src/agents/result.py
@@ -196,7 +196,11 @@ class RunResultStreaming(RunResultBase):
                 break
 
             try:
-                item = await self._event_queue.get()
+                # Avoid blocking forever if the background task errors before enqueuing.
+                item = await asyncio.wait_for(self._event_queue.get(), timeout=0.1)
+            except asyncio.TimeoutError:
+                # No item yet; re-check for stored exceptions and completion conditions.
+                continue
             except asyncio.CancelledError:
                 break
 
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -478,15 +478,45 @@ class AgentRunner:
                     )
 
                     if current_turn == 1:
-                        input_guardrail_results, turn_result = await asyncio.gather(
-                            self._run_input_guardrails(
+                        # Separate blocking and non-blocking guardrails for first turn
+                        all_guardrails = starting_agent.input_guardrails + (
+                            run_config.input_guardrails or []
+                        )
+                        blocking_guardrails, non_blocking_guardrails = (
+                            self._separate_blocking_guardrails(all_guardrails)
+                        )
+
+                        if blocking_guardrails:
+                            # Gate the initial model call on blocking guardrails completing.
+                            # If a tripwire is triggered, this will raise BEFORE any LLM call.
+                            blocking_results = await self._run_input_guardrails(
                                 starting_agent,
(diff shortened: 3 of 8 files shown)


lionelchg:
The following python script shows the expected behaviour with blocking and non-blocking guardrails:

```python
import asyncio
from agents import Agent, InputGuardrail, GuardrailFunctionOutput, Runner, ItemHelpers
from agents.exceptions import InputGuardrailTripwireTriggered

class HomeworkOutput(BaseModel):
    is_homework: bool
    reasoning: str

guardrail_agent = Agent(
    name="Guardrail check",
    instructions="Check if the user is asking about homework.",
    output_type=HomeworkOutput,
)

math_tutor_agent = Agent(
    name="Math Tutor",
    instructions="You provide help with math problems. "
    "Explain your reasoning at each step and include examples",
)

async def homework_guardrail(ctx, agent, input_data):
    result = await Runner.run(guardrail_agent, input_data, context=ctx.context)
    final_output = result.final_output_as(HomeworkOutput)
        output_info=final_output,
        tripwire_triggered=not final_output.is_homework,
    )

triage_agent_blocking = Agent(
    name="Triage Agent",
    instructions="You determine which agent to use based on the user's homework question",
    handoffs=[math_tutor_agent],
    input_guardrails=[
        InputGuardrail(guardrail_function=homework_guardrail, block_downstream_calls=True),
    ],
)

triage_agent_non_blocking = Agent(
    name="Triage Agent",
    instructions="You determine which agent to use based on the user's homework question",
    handoffs=[math_tutor_agent],
    input_guardrails=[
        InputGuardrail(guardrail_function=homework_guardrail, block_downstream_calls=False),
    ],
)

async def main():
    # Example: General/philosophical question
    try:
        result = await Runner.run(triage_agent_blocking, "What is the meaning of life?")
        print(result.final_output)
        print("Guardrail blocked this input:", e)

    # Example: Non-blocking guardrail
    try:
        result = await Runner.run(triage_agent_non_blocking, "What is the meaning of life?")
        print(result.final_output)
        print("Guardrail blocked this input:", e)

if __name__ == "__main__":
    asyncio.run(main())
``` 

Triggering the following traces in OpenAI platform:
Screenshot 2025-08-30 at 23 03 52
which shows the first call being blocked and the second non-blocking where the subsequent call has been run

lionelchg:
Happy to get your feedback @seratch and @rm-openai here as I started from your original PR :). Hope that helps!

lionelchg:
@seratch @rm-openai would you have some time to review the PR this week :)?

seratch:
Sorry for our slow response here. Overall, this approach looks good to me. This has been a customization option we've been requested several times from users, so I think having this change would be good.

@rm-openai This PR is based on your past PR, so you should be comfortable to review this. Can you take a look soon?

ietheredge:
I see the failed checks but just bumping on this stale PR as it seems like a critical issue.

PR #1631, opened 2025-09-01
Add exit condition and user prompt to routing.py for loop termination

SheikhMuhammadHamza-HS:
This PR introduces an exit mechanism to the routing.py example, addressing the infinite loop issue in the original code. The changes include:

Exit Condition: Added a check to break the loop when the user inputs exit, sortie, or salida, aligning with the supported languages (English, French, Spanish).
User Prompt: Included a prompt to inform users about valid exit commands after each interaction, enhancing clarity and usability.

Benefits:
Improves user experience by allowing users to gracefully end the conversation.
Makes the example more practical for real-world applications by resolving the infinite loop issue.
Maintains compatibility with the existing multi-language handoff pattern.

### DIFF
--- a/examples/agent_patterns/routing.py
+++ b/examples/agent_patterns/routing.py
@@ -1,12 +1,11 @@
 import asyncio
 import uuid
-
 from openai.types.responses import ResponseContentPartDoneEvent, ResponseTextDeltaEvent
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. However, we don't need this change for the example.

PR #1645, opened 2025-09-03
feat(repl): add preload_history support to run_demo_loop

DevHammad0:
### Summary
This PR introduces support for an optional `preload_history` parameter in the `run_demo_loop` function.  
It allows users to initialize a session with example user/assistant exchanges, enabling few-shot prompting.

### Changes
- Extended `run_demo_loop` to accept `preload_history`.
- Updated docstring and `docs/repl.md` with usage example.
- Added test (`test_run_demo_loop_with_preload_history`) to verify functionality.

### Motivation
Few-shot prompting is a common technique for shaping agent behavior.  
This update makes it easier to bootstrap interactive sessions with predefined examples.

### Testing
- Added a new test in `tests/test_repl.py` covering preload history.  
- Ran full test suite with `pytest`, all tests passed.

### DIFF
--- a/docs/repl.md
+++ b/docs/repl.md
@@ -20,10 +20,10 @@ if __name__ == "__main__":
 To end this chat session, simply type `quit` or `exit` (and press Enter) or use the `Ctrl-D` keyboard shortcut.
 
 
-## Few-shot initialization with preload_history
+## Few-shot initialization with `preload_history`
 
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for sending this pull request. The feature is interesting, but we don't want to make the repl demo module complex, so if you need this feature, use your code for it.

DevHammad0:
> Thanks for sending this pull request. The feature is interesting, but we don't want to make the repl demo module complex, so if you need this feature, use your code for it.

Got it thanks for the feedback!

PR #1659, opened 2025-09-04
Pr fix incomplete agent graphs when using handoff() function

LAHARI2003:
**##Problem##**
The draw_graph function in _src/agents/extensions/visualization.py_ exhibits inconsistent behavior when visualizing agent chains. visualization.py:148-165 Currently, the visualization system only recursively traverses direct Agent references but fails to recursively traverse Handoff objects created by the handoff() function. visualization.md:15-21

This limitation causes incomplete visualizations where only the first few agents in a handoff chain are displayed, breaking the user's expectation that the complete agent workflow should be visible regardless of the handoff implementation method used.

**##Current Behavior:##**

Agent chains using direct Agent references: Complete visualization
Agent chains using handoff() function:  Incomplete visualization (stops after first handoff)
Root Cause
The visualization functions get_all_nodes() and get_all_edges() contain conditional logic that only triggers recursive traversal for isinstance(handoff, Agent) but not for isinstance(handoff, Handoff).

**##Solution##**
This PR implements a three-part fix to enable complete agent chain visualization:

Extend Handoff class: Add an optional agent field to store a reference to the target agent object for visualization purposes
Update handoff() function: Populate the agent reference when creating Handoff objects
Enhance visualization functions: Modify get_all_nodes() and get_all_edges() to recursively traverse Handoff objects when the agent reference is available
The solution maintains backward compatibility while enabling consistent visualization behavior across all handoff types.

**##Impact##**
This fix resolves a user experience inconsistency where developers using the recommended handoff() function received incomplete visualizations compared to those using direct agent references. visualization.md:94-105 The change ensures that all agent workflows are fully visualizable regardless of implementation approach.

### DIFF
--- a/src/agents/extensions/visualization.py
+++ b/src/agents/extensions/visualization.py
@@ -84,6 +84,9 @@ def get_all_nodes(
                 f"shape=box, style=filled, style=rounded, "
                 f"fillcolor=lightyellow, width=1.5, height=0.8];"
             )
+            if handoff.agent and handoff.agent.name not in visited:  
+                parts.append(get_all_nodes(handoff.agent, agent, visited))
+            
         if isinstance(handoff, Agent):
             if handoff.name not in visited:
                 parts.append(
@@ -134,6 +137,8 @@ def get_all_edges(
         if isinstance(handoff, Handoff):
             parts.append(f"""
             "{agent.name}" -> "{handoff.agent_name}";""")
+            if handoff.agent:  
+                parts.append(get_all_edges(handoff.agent, agent, visited)) 
         if isinstance(handoff, Agent):
             parts.append(f"""
             "{agent.name}" -> "{handoff.name}";""")
--- a/src/agents/handoffs.py
+++ b/src/agents/handoffs.py
@@ -118,7 +118,8 @@ class Handoff(Generic[TContext, TAgent]):
     """Whether the input JSON schema is in strict mode. We **strongly** recommend setting this to
     True, as it increases the likelihood of correct JSON input.
     """
-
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for sending this PR. However, I think this change could cause circular dependency issues with bi-directional handoffs. So, I don't think we can have this change.

LAHARI2003:
@seratch Can you please explain you major concerns with it? So that i can try to resolve this with a new PR is it because of the memory management issues if the bi directional handoffs happen?  because logically draw_graph doesnt give the full flow if handoff objects are being used instead of passing them directly

seratch:
when you pass `hanfoffs` parameter with an array when initializing an agent, it should work with your code. However, it's also possible to append any agents to `agent.handoffs` array after initializing an agent. In this case, drawing the whole graph with recursively resolved handoff relations may not work well. Even if it works, I don't think it's useful for developers (correct me if this is wrong).

Also, I didn't mention this in the previous comment, but we hesitate to add the agent reference just for this purpose because the SDK's functionalities do not need the changes.

LAHARI2003:
i agree that only for visualization purposes adding the agent reference to a core module is not an acceptable solution but , my point here was that when we are using the **_draw graph functionality_** we expect it to show the full working flow of our set up not just the first handoff right regards to how ever we pass them,

so for example i am passing the agent using the handoff object just to filter out the input received by the next agent , in this case i am not appending anything new , but when i use the draw _graph i would want the whole flow to be shown not just the first handoff


so if i can come with some solution which can track the agents with out modifying the core components it should be fine right?
not sure how but i can think this can be done with weak ref

PR #1662, opened 2025-09-04
Feat: Add StructuredSQLiteSession with conversation branching & usage tracking

habema:
This PR introduces `StructuredSQLiteSession` as a drop-in replacement for `SQLiteSession` with enhanced conversation management capabilities. Addresses #1385 and builds on the foundation from #1474.

### Key Features
- **Structured queries**: Get conversations by turns, tool usage analytics, etc.
- **Usage tracking**: Turn-level token usage with detailed breakdowns
- **Conversation branching**: Soft deletion allows "undoing" parts of conversations and continuing from any point

### New Table Schema
```sql
-- Message structure with soft deletion
message_structure (
    id, session_id, message_id, message_type, sequence_number,
    user_turn_number, tool_name, is_active, created_at, deactivated_at
)

-- Turn-level usage tracking  
turn_usage (
    id, session_id, user_turn_number, requests, input_tokens, 
    output_tokens, total_tokens, input_tokens_details, output_tokens_details
)
```

### Usage Pattern
Currently requires manual usage storage:
```python
result = await Runner.run(agent, "Hello", session=session)
await session.store_run_usage(result)  # Manual step
```
**Question**: Should usage storage be automatic? 

Currently users need to manually call `store_run_usage()` after each run to get the analytics features (turn-level token tracking, usage breakdowns, etc.). It works fine but could be better UX-wise.

However, since this is an extension feature, it shouldn't require changes to the core SDK (`run.py`). Making it automatic could be quite a complex change - as it could need to patch the runner itself, and potentially break the clean separation between core and extensions.

What's the preference here - keep the explicit manual approach or explore automatic storage despite the complexity?

### Testing Notes
Tests skip in `old_versions_test` due to sqlalchemy dependency in the same directory (same issue as #1357). The structured session itself doesn't need sqlalchemy, but the import path requires it.

### DIFF
--- a/examples/memory/advanced_structured_session_example.py
+++ b/examples/memory/advanced_structured_session_example.py
@@ -1,0 +1,191 @@
+"""
+Advanced example demonstrating conversation branching with StructuredSQLiteSession.
+
+This example shows how to use soft deletion for conversation editing/branching,
+allowing you to "undo" parts of a conversation and continue from any point.
+"""
+
+import asyncio
+
+from agents import Agent, Runner, function_tool
+from agents.extensions.memory import StructuredSQLiteSession
+
+
+@function_tool
+async def get_weather(city: str) -> str:
+    if city.strip().lower() == "new york":
+        return f"The weather in {city} is cloudy."
+    return f"The weather in {city} is sunny."
+
+
+async def main():
+    # Create an agent
+    agent = Agent(
+        name="Assistant",
+        instructions="Reply very concisely.",
+        tools=[get_weather],
+    )
+
+    # Create a structured session instance
+    session = StructuredSQLiteSession("conversation_advanced")
+
+    print("=== Advanced Structured Session: Conversation Branching ===")
+    print("This example demonstrates conversation editing and branching.\n")
+
+    # Build initial conversation
+    print("Building initial conversation...")
+
+    # Turn 1
+    print("Turn 1: User asks about Golden Gate Bridge")
+    result = await Runner.run(
+        agent,
+        "What city is the Golden Gate Bridge in?",
+        session=session,
+    )
+    print(f"Assistant: {result.final_output}")
+    await session.store_run_usage(result)
+
+    # Turn 2
+    print("Turn 2: User asks about weather")
+    result = await Runner.run(
+        agent,
+        "What's the weather in that city?",
+        session=session,
+    )
+    print(f"Assistant: {result.final_output}")
+    await session.store_run_usage(result)
+
+    # Turn 3
+    print("Turn 3: User asks about population")
+    result = await Runner.run(
+        agent,
+        "What's the population of that city?",
+        session=session,
+    )
+    print(f"Assistant: {result.final_output}")
+    await session.store_run_usage(result)
+
+    print("\n=== Original Conversation Complete ===")
+
+    # Show current conversation
+    print("Current conversation:")
+    current_items = await session.get_items()
+    for i, item in enumerate(current_items, 1):
+        role = str(item.get("role", item.get("type", "unknown")))
+        if item.get("type") == "function_call":
+            content = f"{item.get('name', 'unknown')}({item.get('arguments', '{}')})"
+        elif item.get("type") == "function_call_output":
+            content = str(item.get("output", ""))
+        else:
+            content = str(item.get("content", item.get("output", "")))
+        print(f"  {i}. {role}: {content}")
+
+    print(f"\nTotal items: {len(current_items)}")
+
+    # Demonstrate conversation branching
+    print("\n=== Conversation Branching Demo ===")
+    print("Let's say we want to edit the conversation from turn 2 onwards...")
+
+    # Soft delete from turn 2 to create a branch point
+    print("\nSoft deleting from turn 2 onwards to create branch point...")
+    deleted = await session.soft_delete_from_turn(2)
+    print(f"Deleted: {deleted}")
+
+    # Show only active items (turn 1 only)
+    active_items = await session.get_items()
+    print(f"Active items after deletion: {len(active_items)}")
+    print("Active conversation (turn 1 only):")
+    for i, item in enumerate(active_items, 1):
+        role = str(item.get("role", item.get("type", "unknown")))
+        if item.get("type") == "function_call":
+            content = f"{item.get('name', 'unknown')}({item.get('arguments', '{}')})"
+        elif item.get("type") == "function_call_output":
+            content = str(item.get("output", ""))
+        else:
+            content = str(item.get("content", item.get("output", "")))
+        print(f"  {i}. {role}: {content}")
+
+    # Create a new branch
+    print("\nCreating new conversation branch...")
+    print("Turn 2 (new branch): User asks about New York instead")
+    result = await Runner.run(
+        agent,
+        "Actually, what's the weather in New York instead?",
+        session=session,
+    )
+    print(f"Assistant: {result.final_output}")
+    await session.store_run_usage(result)
+
+    # Continue the new branch
+    print("Turn 3 (new branch): User asks about NYC attractions")
+    result = await Runner.run(
+        agent,
+        "What are some famous attractions in New York?",
+        session=session,
+    )
+    print(f"Assistant: {result.final_output}")
+    await session.store_run_usage(result)
+
+    # Show the new conversation
+    print("\n=== New Conversation Branch ===")
+    new_conversation = await session.get_items()
+    print("New conversation with branch:")
+    for i, item in enumerate(new_conversation, 1):
+        role = str(item.get("role", item.get("type", "unknown")))
+        if item.get("type") == "function_call":
+            content = f"{item.get('name', 'unknown')}({item.get('arguments', '{}')})"
+        elif item.get("type") == "function_call_output":
+            content = str(item.get("output", ""))
+        else:
+            content = str(item.get("content", item.get("output", "")))
+        print(f"  {i}. {role}: {content}")
+
+    # Show that we can still access the original branch
+    all_items = await session.get_items(include_inactive=True)
+    print(f"\nTotal items including inactive (original + new branch): {len(all_items)}")
+
+    print("\n=== Conversation Structure Analysis ===")
+    # Show conversation turns (active only)
+    conversation_turns = await session.get_conversation_by_turns()
+    print("Active conversation turns:")
+    for turn_num, items in conversation_turns.items():
+        print(f"  Turn {turn_num}: {len(items)} items")
+        for item in items:  # type: ignore
+            if item["tool_name"]:  # type: ignore
+                print(
+                    f"    - {item['type']} (tool: {item['tool_name']}) [active: {item['active']}]"  # type: ignore
+                )
+            else:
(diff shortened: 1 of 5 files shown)


seratch:
@codex Can you review this PR? Share both good things and rooms for improvements.

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,567 @@
...
+    - Message structure metadata and tool usage statistics
+    """
+
+    ACTIVE = 1  # Message is active and visible in conversation
+    INACTIVE = 0  # Message is soft-deleted (hidden but preserved)
+
+    def __init__(self, session_id: str, db_path: str | Path = ":memory:", **kwargs):
+        super().__init__(session_id, db_path, **kwargs)
+        self._current_user_turn = 0
+        self._init_structure_tables()
```
seratch: Confirmed running this method every time a developer instantiates this class, but we may want to have an option to run this operation. Say, `create_tables: bool = False`?

examples/memory/advanced_session_advanced_example.py
```diff
@@ -0,0 +1,191 @@
+"""
```
seratch, line 1: the file name can be `advanced_sqlite_session_example.py`?

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,567 @@
...
+    ) -> list[TResponseInputItem]:
+        """Get items, optionally including soft-deleted ones."""
+        if include_inactive:
+            # Use parent implementation for all items
+            return await super().get_items(limit)
+
+        # Filter to only active items
+        def _get_active_items_sync():
+            conn = self._get_connection()
+            with self._lock if self._is_memory_db else threading.Lock():
```
seratch: since this is an async method, you may want to use Lock from asyncio rather than threading's one

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,567 @@
...
+        Session-level usage can be aggregated from turn data when needed.
+        """
+        try:
+            if result.context_wrapper.usage is not None:
+                # Only update turn-level usage - session usage is aggregated on demand
+                await self._update_turn_usage_internal(
+                    self._current_user_turn, result.context_wrapper.usage
+                )
+        except Exception as e:
+            import logging
```
seratch: you can have this import at top level, plus we may want to enable users to pass their custom `logger` (optional) to the constructor.

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,567 @@
...
+    Features:
+    - Turn-based conversation management with soft delete/reactivate
+    - Detailed usage tracking per turn with token breakdowns
+    - Message structure metadata and tool usage statistics
+    """
+
+    ACTIVE = 1  # Message is active and visible in conversation
+    INACTIVE = 0  # Message is soft-deleted (hidden but preserved)
+
+    def __init__(self, session_id: str, db_path: str | Path = ":memory:", **kwargs):
```
seratch: To avoid relying on the argument orders here (we may add more options in the future), let's ask users to give argument names:
```suggestion
    def __init__(self, *, session_id: str, db_path: str | Path = ":memory:", **kwargs):
```

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,567 @@
...
+        """Extract structure metadata with microsecond precision ordering.
+
+        This method:
+        - Increments user turn counter for each user message encountered
+        - Assigns explicit sequence numbers for precise ordering
+        - Links messages to their database IDs for structure tracking
+        - Handles multiple user messages in a single batch correctly
+        """
+        try:
+            conn = self._get_connection()
```
seratch, line 334: I should do the same for the sqlite session too, but how about doing this instead?

```python
from contextlib import closing

with self._get_connection() as conn: 
    with closing(conn.cursor()) as cur:
```

Using this `with closing` helps us do commit/rollback/close under the hood.

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,567 @@
...
+    async def soft_delete_from_turn(self, user_turn_number: int) -> bool:
+        """Soft delete conversation from a specific user turn onwards."""
+        conn = self._get_connection()
+
+        cursor = conn.execute(
+            """
+            UPDATE message_structure
+            SET is_active = 0, deactivated_at = CURRENT_TIMESTAMP
+            WHERE session_id = ? AND user_turn_number >= ? AND is_active = 1
+        """,
```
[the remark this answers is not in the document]
habema: This is somewhat valid. Both active and inactive messages must be trackable for the sake of cost accounting. My implementation needs to be extended so turn usage and session usage can be tracked for active and inactive records.

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,567 @@
...
+        conn.commit()
+        return affected > 0
+
+    async def _update_turn_usage_internal(self, user_turn_number: int, usage_data: Usage) -> None:
+        """Internal method to update usage for a specific turn with full JSON details."""
+
+        def _update_sync():
+            conn = self._get_connection()
+
+            # Serialize token details as JSON
```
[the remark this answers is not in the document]
habema: This is valid, but SQLiteSession needs to reflect this as well

src/agents/extensions/memory/advanced_sqlite_session.py
habema: I don't mind implementing this, but this breaks consistency with the SQLiteSession, and introduces a feature that might not be very useful for an SQLite implementation that isnt meant for production use.

Either this should be implemented here, and therefore in the SQLiteSession, or left later down the line for when this advanced functionality is extended to SQLAlchemy

examples/memory/advanced_session_advanced_example.py
habema: Yes, must've been an accident

src/agents/extensions/memory/advanced_sqlite_session.py
habema: Once again, the base class also uses threading.lock(). I will change this and make another PR to make the SQLiteSession use the same pattern.

habema:
I added the `active_only` param to usage methods. Instead of modifying turn_usage table, I kept all usage data intact (for cost accounting) but added SQL JOINs to filter by active messages when needed.

For `asyncio.Lock`, I kept `threading.Lock` to match base class. Changing it in `AdvancedSQLiteSession` only will require overriding all base class methods, so the `asyncio` refactor should be a separate PR that touches both classes.

habema:
I ran into a bug in my implementation of `reactivate_from_turn()`, which uncovered deeper issues. I replaced the broken soft deletion approach with actual branch-based conversation management. 

**Key changes:**
- Added `branch_id` and `branch_turn_number` columns to schema, removed `is_active`/`deactivated_at` 
- Replaced `deactivate_from_turn()`/`reactivate_from_turn()` with `create_branch_from_turn()` that properly copies conversation history
- Fixed the original bug where reactivating would affect both original and new branches
- Added proper branch switching with `switch_to_branch()` and `list_branches()`
- Updated unique constraints to include `branch_id` so turn numbers can overlap between branches

Now each branch is truly isolated - no more weird cross-branch contamination when doing operations.

I understand this might require some extensive review, so take your time.

src/agents/extensions/memory/advanced_sqlite_session.py
habema: As mentioned in the comment below, this might be better suited for a later PR that changes both SQLiteSession and AdvancedSQLiteSession

habema:
@seratch awaiting your review

seratch:
@habema Thanks for being patient here. I didn't have enough time for intensively reviewing this one today, but I will do tomorrow!

examples/memory/advanced_sqlite_session_example_1.py
```diff
@@ -0,0 +1,111 @@
+"""
```
seratch: It's fine to have a long example code, so can you merge example_1 and example_2 into one file?

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,1166 @@
...
+                cursor.execute(
+                    """
+                    SELECT
+                        ms.branch_turn_number,
+                        am.message_data,
+                        ms.created_at
+                    FROM message_structure ms
+                    JOIN agent_messages am ON ms.message_id = am.id
+                    WHERE ms.session_id = ? AND ms.branch_id = ?
+                    AND ms.message_type = 'user'                    ORDER BY ms.branch_turn_number
```
seratch: Can you have a line break before ORDER BY here?

src/agents/extensions/memory/advanced_sqlite_session.py
seratch: I got it. Can you add TODO comment on it?

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,1166 @@
...
+            branch_name = f"branch_from_turn_{turn_number}_{timestamp}"
+
+        # Copy messages before the branch point to the new branch
+        await self._copy_messages_to_new_branch(branch_name, turn_number)
+
+        # Switch to new branch
+        old_branch = self._current_branch_id
+        self._current_branch_id = branch_name
+
+        self._logger.info(
```
seratch: Using debug-level logging for these ones is reasonable too (if a dev really wants to have an info-level log for this timing, they want to have more specific details as well). Thoughts?

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -0,0 +1,1166 @@
...
+                cursor.execute(
+                    """
+                    SELECT
+                        ms.branch_turn_number,
+                        am.message_data,
+                        ms.created_at
+                    FROM message_structure ms
+                    JOIN agent_messages am ON ms.message_id = am.id
+                    WHERE ms.session_id = ? AND ms.branch_id = ?
+                    AND ms.message_type = 'user'                    AND am.message_data LIKE ?
```
seratch: Can you improve the format here?

src/agents/extensions/memory/advanced_sqlite_session.py
habema: Sure thing

src/agents/extensions/memory/advanced_sqlite_session.py
habema: I think debug is more suitable yes

examples/memory/advanced_sqlite_session_example_1.py
habema: Sure

seratch (approved):
Thanks for adding this!

PR #1666, opened 2025-09-05
Use raw input instead of prepared input to input guardrail

aaron-ang:
### Summary

This change ensures that input guardrails receive the original user input rather than the input combined with session history. Input guardrails should validate the user's current message, not the entire conversation context, as they are designed to check if the new input is appropriate rather than validating the full conversation history.

In addition, I created some helper functions that abstract duplicated chunks of code across `run` and `_start_streaming` methods.

### Checks

- [x] I've run `make lint` and `make format` (skipped file changes since they are unrelated to the PR)

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -445,22 +445,9 @@ class AgentRunner:
                     # Start an agent span if we don't have one. This span is ended if the current
                     # agent changes, or if the agent loop ends.
                     if current_span is None:
-                        handoff_names = [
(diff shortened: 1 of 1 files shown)


rm-openai (changes requested):
Passing the input history was intentional. Some input guardrails only operate on the newest message, and they can just filter the inputs. But others operate in the context of the overall conversation, and this would be a regression for them.

PR #1669, opened 2025-09-05
Fix a minor bug in #1601

seratch:
This pull request resolves this issue: https://github.com/openai/openai-agents-python/pull/1601#issuecomment-3257117745

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -142,7 +142,7 @@ class WebSearchTool:
 
     @property
     def name(self):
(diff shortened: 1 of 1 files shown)


rm-openai (approved):
just confirming this is safe/backwards compatible/wont cause any regressions?

seratch:
Yeah, this shouldn't cause any issues for the existing apps, plus it must be consistent with the actual type name for future users.

PR #1537, opened 2025-08-20
fix(voice/stt): accept conversation.item.input_audio_transcription.completed (keep legacy alias)

dudududukim:
### Summary

This PR updates event handling in `OpenAISTTTranscriptionSession` to support both legacy `"input_audio_transcription_completed"` and the currently observed `"conversation.item.input_audio_transcription.completed"` events when streaming transcription over the WebSocket endpoint (wss://api.openai.com/v1/realtime?intent=transcription).

The test suite (`tests/voice/test_openai_stt.py`) is also updated to cover both event types, ensuring that the transcription output is properly yielded from either event in real scenarios.

### Motivation

In my environment, events retrieved from 
`event = await asyncio.wait_for(self._event_queue.get(), timeout=EVENT_INACTIVITY_TIMEOUT)` correspond to those emitted by the OpenAI WebSocket session.

The observed event for completed transcription is currently `"conversation.item.input_audio_transcription.completed"`, not the legacy event name.

These changes allow realtime STT workflows to work as expected, without manual hotfixes.

### Testing
<figure>
  before/after
  <figcaption><strong>Top:</strong> Using the legacy event name;<br>
  <strong>Bottom:</strong> After adding support for <code>conversation.item.input_audio_transcription.completed</code>.</figcaption>
</figure>

<br>
<br>
Tests pass successfully in my local environment 

[dudududukim/OpenAI_agent_practice](https://github.com/dudududukim/OpenAI_agent_practice)

The updated test confirms correct behavior for both event types.

PR #1679, opened 2025-09-07
Fix event type in openai_stt.py

### LINKED ISSUE #1678: Outdated event listening in  `openai_stt.py`
### Bug description
when running example `examples/voice/streamed`, the transcription is never processed, because it expects an event type `input_audio_transcription_completed`. After update of the API, the correct event to listen for is `conversation.item.input_audio_transcription.completed`

### Debug information
- Agents SDK version: (e.g. `v0.2.11`)
- Python version (e.g. Python 3.12.2)

### Repro steps

running 
```python
python -m examples.voice.streamed.main
```
no output will be generated. After the suggested fix, example runs normally.

majklost:
Fixes #1678

### DIFF
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -226,7 +226,7 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
                     break
 
                 event_type = event.get("type", "unknown")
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. https://github.com/openai/openai-agents-python/pull/1537 resolves this issue in a better way, so let me close this PR.

Issue #1061, opened 2025-07-11
Agent run with previous_response_id fails - No tool output found for function call call_WdnUUKXKvwy3jk....


markns:
- **Have you searched for related issues?** Others may have faced similar issues.
These issues are related, but don't have a resolution:
https://github.com/openai/openai-agents-python/issues/673
https://github.com/openai/openai-agents-python/issues/632

### Describe the bug
Agent runs using previous_response_id fail unexpectedly and sporadically following a tool call, with an error such as `"Error code: 400 - {'error': {'message': 'No tool output found for function call call_WdnUUKXKvwy3jksn4I9ERdRT.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}"`

Here's how the error looks in the trace console:

<img width="1272" height="318" alt="Image" src="https://github.com/user-attachments/assets/03a9a8d3-b851-4ff7-9f1e-cddb6c1700a9" />

The error seems to originate in the previous trace when the output from tool call is not posted back to the OpenAI API

<img width="1275" height="493" alt="Image" src="https://github.com/user-attachments/assets/60009467-6e82-4e6c-8f23-81a029e5c0c6" />

This is also shown in logging, where there is no post to the `/v1/responses` endpoint after `3: Tool data_collection_consent ended`:
```
2025-07-11 12:16:39.717 | INFO     | farmwise.hooks:on_agent_start:18 - ### 1: Agent Onboarding Agent started. Usage: 0 requests, 0 input tokens, 0 output tokens, 0 total tokens
2025-07-11 12:16:40.318 | INFO     | httpx._client:_send_handling_redirects:1695 - HTTP Request: POST https://api.openai.com/v1/responses "HTTP/1.1 200 OK"
2025-07-11 12:16:42.414 | INFO     | httpx._client:_send_handling_redirects:980 - HTTP Request: POST https://api.openai.com/v1/traces/ingest "HTTP/1.1 204 No Content"
2025-07-11 12:16:46.042 | INFO     | farmwise.hooks:on_tool_start:28 - ### 2: Tool data_collection_consent started. Usage: 1 requests, 2669 input tokens, 97 output tokens, 2766 total tokens
2025-07-11 12:16:46.079 | INFO     | httpx._client:_send_handling_redirects:1695 - HTTP Request: POST http://127.0.0.1:8000/api/v1/default/contacts/11/consent "HTTP/1.1 200 OK"
[...]

PR #1689, opened 2025-09-08
Prevent preamble messages from being treated as final output when tool calls are pending

ihower:
# Issue

When using structured outputs, if the model produces both a message and a function call in the same turn,
the agent would terminate prematurely.

This behavior has become much easier to reproduce with GPT-5 tool calling + preambles.
Earlier models could also occasionally produce such outputs. For example, in GPT-4.1, if the prompt instructed the model to think before calling a tool—making it possible to encounter random failures whenever output_type was specified.

I think this fix should also resolve [openai/openai-agents-python#1061](https://github.com/openai/openai-agents-python/issues/1061).

# Reproducible Example

```python

@dataclass
class QueryResult:
    content: str    

@function_tool
    return "hot"

agent = Agent(
    name="Test preamble issue",
    instructions="""You are a helpful agent. <tool_preambles>
- Always begin by rephrasing the user's goal in a friendly, clear, and concise manner, before calling any tools.
- Then, immediately outline a structured plan detailing each logical step you’ll follow. 
- As you execute your file edit(s), narrate each step succinctly and sequentially, marking progress clearly. 
- Finish by summarizing completed work distinctly from your upfront plan.
</tool_preambles>
For the final answer, no plan is needed - just return the result.
""",
    model="gpt-5",
    tools=[get_weather],
    output_type=QueryResult,
)

result = Runner.run_sync(agent, input="What's the weather in Hsinchu?")

print(result.to_input_list())
print("--------------------------------")
print(result.final_output)
```

## Example Output (before this fix)

The to_input_list() looks like:
```
[
  {"content": "What's the weather in Hsinchu?", "role": "user"},
  {"id": "rs_...", "summary": [], "type": "reasoning"},
  {"id": "msg_...", "content": [...{"text": "Your goal: Get the current weather for Hsinchu.\nPlan:\n1) Query the weather service for Hsinchu.\n2) Return the current conditions and temperature briefly.\nProgress:\n- Step 1: Querying the weather service for Hsinchu now."}], "role": "assistant", "type": "message"},
  {"arguments": '{"city":"Hsinchu"}', "call_id": "call_...", "name": "get_weather", "type": "function_call"},
  {"call_id": "call_...", "output": "hot", "type": "function_call_output"}
]

```

At this point, the list ends with the tool call output. The run loop does not pass this tool output back to the LLM for another turn. As a result, the agent finalizes prematurely, treating the preamble message as the structured output instead of producing a final answer that incorporates the tool result:

```
# final_output
QueryResult(content='Your goal: Get the current weather for Hsinchu.\nPlan:\n1) Query the weather service for Hsinchu. \n2) Return the current conditions and temperature briefly. \nProgress:\n- Step 1: Querying the weather service for Hsinchu now.')
```

# Behavior Change

**Before:**
When a structured output schema was used, if the model produced both a  preamble message and a function call in the same turn, the agent would treat the preamble message as the final output, even though tool calls were still pending.  
This caused the run loop to terminate early.

**After:**
Structured output only triggers a final output when there are no pending tool calls or approvals.
If the model emits a tool call, we process it, append the result, and re-run the loop.

I think this aligns with the default tool_use_behavior=run_llm_again, which is more intuitive.
If developers prefer to treat function call output as the final output, they can explicitly configure tool_use_behavior to stop_on_first_tool or StopAtTools.

✅ All tests are passing

## Example Output (after this fix)

The list now continues after the tool call output, feeding the tool result back to the LLM and producing the proper final answer:

```
[
  {"content": "What's the weather in Hsinchu?", "role": "user"},
  {"id": "rs_...", "summary": [], "type": "reasoning"},
  {"id": "msg_...", "content": [...{"text": "Goal: You want to know the current weather in Hsinchu.\n\nPlan:\n1) Query the weather service for Hsinchu. 2) Wait for the response. 3) Present the current weather details succinctly.\n\nProgress:\n- Step 1: Sending request to the weather service for Hsinchu now."}], "role": "assistant", "type": "message"},
  {"arguments": '{"city":"Hsinchu"}', "name": "get_weather", "type": "function_call"},
  {"call_id": "...", "output": "hot", "type": "function_call_output"},
  {"id": "msg_...", "content": [...{"text": "Progress:\n- Step 2: Received the weather data.\n- Step 3: Preparing the concise result for you now."}], "role": "assistant", "type": "message"},
  {"id": "msg_...", "content": [...{"text": "It’s currently hot in Hsinchu."}], "role": "assistant", "type": "message"}
]

```

Note: after a function call output, the model may possibly produce a **post-message** (e.g. explaining that it received the weather data) before generating the final output message in the same turn.

And the final output now correctly incorporates the tool result:

```
# final_output
QueryResult(content='It’s currently hot in Hsinchu.')
```


seratch:
@codex Can you check if this PR should have the same change with https://github.com/openai/openai-agents-js/pull/452 ?

seratch:
I've checked the above possibility, but for Python SDK, we don't need any further changes here.

javi-horizon:
when will this be available via the python open ai agents sdk, i have encountered many errors

KthProg:
@javi-horizon I second that

PR #1692, opened 2025-09-08
adding is_enabled in handoffs example code snippet

Umair4444:
### Summary
This PR updates the handoffs example code snippet by adding `is_enabled`.

### Changes
- Added `is_enabled` parameter to demonstrate how developers can toggle a handoff.
- Updated example instructions for clarity.

### Why
This improves the documentation by showing how `is_enabled` can be applied in practical usage.


seratch:
Thanks for sending this. However, the above part already mentions is_enabled and we'd like to avoid having all the possible properties in the code snippet too. So, we won't have the change this time.

PR #1694, opened 2025-09-08
clarify Guardrail agent instructions and fix tripwire logic in homework_guardrail

Umair4444:
### Description
This update expands and clarifies the definition of a “homework” question by changing the agent instructions from *“Check if the user is asking about homework”* to *“Determine if the user’s input looks like a school-style homework question (e.g., math, history, science, academic subjects).”*  

Additionally, the logic in `homework_guardrail` was fixed, updating  
`tripwire_triggered = not final_output.is_homework` → `tripwire_triggered = final_output.is_homework`.  

### Impact
- Ensures the guardrail now correctly triggers on homework-style inputs rather than the opposite.  
- Makes examples easier to understand.  
- Guarantees the guardrail behaves as intended, effectively preventing homework-style queries.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -104,7 +104,12 @@ class HomeworkOutput(BaseModel):
 
 guardrail_agent = Agent(
     name="Guardrail check",
-    instructions="Check if the user is asking about homework.",
+    instructions=
+        """
+        - Determine if the user's input looks like a school-style homework question.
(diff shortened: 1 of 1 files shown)


seratch:
It might be confusing but the guardrail example there checks if a user is asking homework-related questions. So, the current implementation is correct.

PR #1706, opened 2025-09-10
fix: #1704 Preserve thinking blocks in Anthropic conversations with tool calls

habema:
Fixes #1704

**Problem:**
The SDK was dropping reasoning items when converting conversation history to API messages, causing Anthropic to reject requests with:
```
messages.1.content.0.type: Expected `thinking` or `redacted_thinking`, but found `tool_use`
```

**Root cause:**
`Converter.items_to_messages()` was skipping reasoning items entirely, violating Anthropic's requirement that assistant messages with tool calls must start with thinking blocks.

**Technical changes:**

1. **Enhanced `InternalChatCompletionMessage`** to carry `thinking_blocks` alongside `reasoning_content`

2. **Updated `LitellmConverter.convert_message()`** to extract full thinking blocks from LiteLLM responses:
   - Handles both dict and object formats of thinking blocks
   - Preserves the cryptographic `signature` field from Anthropic
   - Safely converts to dict format for storage

3. **Modified `Converter.message_to_output_items()`** to store complete thinking block structure:
   - Stores thinking blocks in `ResponseReasoningItem.content` as `Content` objects
   - Maintains backward compatibility with existing reasoning content

Repro script available in original issue.

PR #1710, opened 2025-09-10
Fix CI failure in `make old_version_tests`

habema:
The `old_version_tests` target was failing in CI because it creates a separate Python 3.9 environment (`.venv_39`) that doesn't inherit dependencies from the main environment. Some tests import from modules that require optional dependencies.

I just added a step to explicitly install all dependencies for the Python 3.9 environment before running the tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -84,5 +84,7 @@ jobs:
           enable-cache: true
       - name: Install dependencies
         run: make sync
+      - name: Install Python 3.9 dependencies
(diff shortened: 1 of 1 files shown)


seratch (approved):

habema:
@seratch i see you merged this into #1706 but it still didnt solve the problem. 

Perhaps the correct command is this one?
```yaml
run: UV_PROJECT_ENVIRONMENT=.venv_39 uv sync --python 3.9 --all-extras --all-packages --group dev
```

Apologies for the confusion, debugging this locally isn't straightforward.

PR #1713, opened 2025-09-11
Allowing change of max turns in agent as tool

frodulec:
- Allowed passing max_turns through Agent.as_tool(...) to nested agent runs.
- Updated Runner.run to accept max_turns: int | None; when None, default to DEFAULT_MAX_TURNS.
- Added targeted tests to cover both behaviors

Run tests, formatting and linting.


seratch:
Thanks for sending this PR. Like I mentioned at https://github.com/openai/openai-agents-js/issues/460, we may want to unlock other options too and we'd like to align with TypeScript SDK for its design. So, perhaps our ideal solution could look a bit different.

Fatima367:
If you want to set max_turns or tweak run config, create a function tool whose body calls Runner.run(...) for the sub-agent (that you want to use as tool) and returns a string.

frodulec:
> Thanks for sending this PR. Like I mentioned at [openai/openai-agents-js#460](https://github.com/openai/openai-agents-js/issues/460), we may want to unlock other options too and we'd like to align with TypeScript SDK for its design. So, perhaps our ideal solution could look a bit different.

I see, than you for the reply, looking forward for this feature then

seratch:
https://github.com/openai/openai-agents-python/pull/1751 resolved this. thanks again for sending this!

PR #1714, opened 2025-09-11
Fix/consistent generic type handling

HezziCode:
# Fix: Preserve Generic Type Information in ToolContext

## Problem
There was an inconsistency in how `RunContextWrapper[T]` generics were handled between dynamic instructions and tool validation functions:

- **Dynamic Instructions**: `ctx.context.age` worked correctly ✅
- **Tool Validation**: `ctx.context.age` threw `AttributeError: 'dict' object has no attribute 'age'` ❌

This forced developers to manually convert context using `UserDataType.model_validate(ctx.context)` in tool validation functions.

## Root Cause
The issue was in `ToolContext.from_agent_context()` method which used field-by-field copying:
```python
# Old problematic code
for f in fields(RunContextWrapper):
    setattr(tool_context, f.name, getattr(context, f.name))



# New fixed code
return cls(
    context=context.context,  # Direct assignment preserves type
    usage=context.usage,
    tool_name=tool_name,
    tool_call_id=tool_call_id,
)



# Before (tool validation)
async def tool_validation(ctx: RunContextWrapper[UserDataType], agent):
    user_data = UserDataType.model_validate(ctx.context)  # Manual conversion needed
    if user_data.age >= 18:
        return True

# After (tool validation) 
async def tool_validation(ctx: RunContextWrapper[UserDataType], agent):
    if ctx.context.age >= 18:  # Direct access works now!
        return True

### DIFF
--- a/src/agents/tool_context.py
+++ b/src/agents/tool_context.py
@@ -33,10 +33,15 @@ class ToolContext(RunContextWrapper[TContext]):
     ) -> "ToolContext":
         """
         Create a ToolContext from a RunContextWrapper.
+
+        This method preserves the generic type information of the context object,
+        ensuring that ctx.context maintains its proper type in tool validation functions.
         """
-        # Grab the names of the RunContextWrapper's init=True fields
-        base_values: dict[str, Any] = {
-            f.name: getattr(context, f.name) for f in fields(RunContextWrapper) if f.init
-        }
+        # Directly pass the context and usage to preserve generic type information
         tool_name = tool_call.name if tool_call is not None else _assert_must_pass_tool_name()
-        return cls(tool_name=tool_name, tool_call_id=tool_call_id, **base_values)
+        return cls(
+            context=context.context,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by seratch** [comment]
I don't think this change is necessary. The following code works without any issues:
```python
import asyncio
from agents import Agent, RunContextWrapper, Runner, function_tool

class MyContext(BaseModel):
    foo: str

@function_tool
def get_weather(cxt: RunContextWrapper[MyContext], city: str) -> str:
    print(cxt.context.foo)
    return "Sunny with wind."

agent = Agent(
    name="Hello world",
    instructions="You are a helpful agent. You must call the tools.",
    tools=[get_weather],
)

async def main():
    result = await Runner.run(
        agent, input="What's the weather in Tokyo?", context=MyContext(foo="FOO")
    )
    print(result.final_output)

if __name__ == "__main__":
    asyncio.run(main())
```

PR #1719, opened 2025-09-11
bound_context for deep sub-agents

Yankzy:
This change introduces a new optional `bound_context` parameter to `Agent.as_tool` so callers can pass a Pydantic `BaseModel` instance that will be used as the run context for that tool invocation. When provided, the bound context overrides the parent agent's context for the duration of the tool call; when omitted, existing behaviour (inheriting the caller's context) is preserved.

### Why
- Enables safe, explicit context binding for sub-agent tools in orchestrations (e.g., orchestrator → multiple specialized agents).
- Prevents context leakage between parallel or sequential sub-agent invocations and makes it easy to pass typed metadata (email, agent_name, profile data) to sub-agents.
- Solves real-world need in the trading orchestrator example where each agent must receive its own `TradeAgentContext` (email/name) to correctly save trade signals.

### What changed
- `src/agents/agent.py`
  - Added parameter: `bound_context: BaseModel | None = None` to `Agent.as_tool(...)`.
  - `run_agent` now calls `Runner.run(..., context=bound_context or context.context)` so the provided `bound_context` is used when present; otherwise the original behavior is retained.
  - Docstring for `bound_context` describing its behavior and that it should be a Pydantic `BaseModel`.

This is a backward-compatible change: default is `None`, so existing uses of `as_tool` are unaffected.

### Example usage
```python
# create a typed context for the sub-agent

async def generate_trade_signal_for_ticker(ticker: str) -> TradeSignal:
    """
    Runs the sentimental strategist agent for a given stock ticker and saves the result.
    """
    
    try:
        agent_adapter = await AgentAdapter.get_instance()
        agent_tools = []
        for agent in agent_adapter.get_all_agents:
            # 1Create the context for this specific agent
            sub_agent_context = TradeAgentContext(
                agent_email=agent.email,
                agent_name=agent.name,
            )

            agent_tool = agent.as_tool(
                tool_name=sanitize_tool_name(agent.name),
                tool_description=f"Run {agent.name} to generate a trade signal for {ticker}.",
                bound_context=sub_agent_context, 
            )
            agent_tools.append(agent_tool)

        judge_agent = Agent(
            name="Trading Signal Orchestrator Agent",
            instructions="""
            - You are the main agent in a Collaborative Agentic workflow to generate trading Signals.  """,
            tools=[
                # other tools,
                *agent_tools,
            ],
            output_type=JudgeAgentOutput,
            hooks=Hooks(display_name="trading_signal_orchestrator_agent"),
            model='gpt-4.1',
        )


```

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -28,6 +28,7 @@ from .run_context import RunContextWrapper, TContext
 from .tool import FunctionTool, FunctionToolResult, Tool, function_tool
 from .util import _transforms
 from .util._types import MaybeAwaitable
+from pydantic import BaseModel
 
 if TYPE_CHECKING:
     from .lifecycle import AgentHooks
@@ -384,6 +385,8 @@ class Agent(AgentBase, Generic[TContext]):
         custom_output_extractor: Callable[[RunResult], Awaitable[str]] | None = None,
         is_enabled: bool
         | Callable[[RunContextWrapper[Any], AgentBase[Any]], MaybeAwaitable[bool]] = True,
+        # This is our new parameter for context binding
+        bound_context: BaseModel | None = None,
     ) -> Tool:
         """Transform this agent into a tool, callable by other agents.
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this PR. Like I mentioned at https://github.com/openai/openai-agents-js/issues/460, we may want to unlock other options too and we'd like to align with TypeScript SDK for its design. So, perhaps our ideal solution could look a bit different.

Issue #1668, opened 2025-09-05
Handoffs with gpt-5* model + store=False + remove_all_tools  fails due to 404 error response


jihun-im:
### Describe the bug
Reasoning model like gpt-5 or gpt-5-nano cannot use the handoff feature when the `store` option set to `False`.
At the handoff-ed agent I got the following error:
>"Items are not persisted when `store` is set to false. Try again with `store` set to true, or remove this item from your input."

Agent definition:
```python
    triage_agent = Agent[RunContext](
        name="triage_agent",
        instructions=instructions,
        model="gpt-5",
        model_settings=ModelSettings(
            tool_choice="auto",
            parallel_tool_calls=False,
            store=False,
        ),
        handoffs=[
            handoff(image_agent, input_filter=handoff_filters.remove_all_tools),
        ],
        tools=[
            WebSearchTool(**web_search_tool_kwargs),
        ],
    )

    image_agent = Agent[RunContext](
        name="image_agent",
        instructions=image_gen_instructions,
        model=openai_model,
        model_settings=ModelSettings(
            tool_choice="auto",
            parallel_tool_calls=False,
            store=False,
        ),
        handoff_description=image_agent_handoff_prompt,
        tools=image_agent_tools,
    )
```

Error log:
```
Traceback (most recent call last):
  File "/xxx/src/common/openai/openai_stream_handler.py", line 105, in stream_events
    async for event in self.result.stream_events():
  File "/xxx/.venv/lib/python3.11/site-packages/agents/result.py", line 215, in stream_events
    raise self._stored_exception
  File "/xxx/.venv/lib/python3.11/site-packages/agents/run.py", line 840, in _start_streaming
    turn_result = await cls._run_single_turn_streamed(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/xxx/.venv/lib/python3.11/site-packages/agents/run.py", line 1009, in _run_single_turn_streamed
    async for event in model.stream_response(
  File "/xxx/.venv/lib/python3.11/site-packages/agents/models/openai_responses.py", line 163, in stream_response
    stream = await self._fetch_response(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/xxx/.venv/lib/python3.11/site-packages/agents/models/openai_responses.py", line 286, in _fetch_response
    return await self._client.responses.create(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1756, opened 2025-09-16
Add default include of reasoning.encrypted_content for store=False to prevent reasoning errors

ihower:
## Issue

When using store=False without including reasoning.encrypted_content, any function calls (including handoff) that generate reasoning items will result in an error because the server does not persist reasoning items. This makes the Responses API difficult to use in practice.

I believe it would be better if `reasoning.encrypted_content` were included by default in this case, since in almost all scenarios the agent needs it when working with reasoning models and `store=False`.

Related issue: [openai/openai-agents-python#1668](https://github.com/openai/openai-agents-python/issues/1668)

## Reproducible Example

```python
from agents import Agent, Runner, function_tool, ModelSettings

@function_tool
    return "hot"

agent = Agent(
    name="Test store false",
    model="gpt-5-mini",
    tools=[get_weather],
    model_settings=ModelSettings(store=False)
)

result = Runner.run_sync(agent, input="What's the weather in Hsinchu?")

print(result.final_output)
```

This results in:

```
Error getting response: Error code: 404 - {'error': {'message': "Item with id 'rs_xxx' not found. Items are not persisted when `store` is set to false. Try again with `store` set to true, or remove this item from your input.", 'type': 'invalid_request_error', 'param': 'input', 'code': None}}. 
```

## Fix

- When `model_settings.response_include is None` **and** `model_settings.store is False`, automatically add `"reasoning.encrypted_content"` to the include set.
- This change makes it easier for developers using `gpt-5`, since they no longer need to worry about missing reasoning data when `store=False`.

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -259,6 +259,10 @@ class OpenAIResponsesModel(Model):
         include_set: set[str] = set(converted_tools.includes)
         if model_settings.response_include is not None:
             include_set.update(model_settings.response_include)
+
+        if model_settings.store is False and model_settings.response_include is None:
+            include_set.add("reasoning.encrypted_content")
+
         if model_settings.top_logprobs is not None:
             include_set.add("message.output_text.logprobs")
         include = cast(list[ResponseIncludable], list(include_set))
--- a/tests/test_store_response_include.py
+++ b/tests/test_store_response_include.py
@@ -1,0 +1,50 @@
+import pytest
+from openai.types.responses.response_usage import InputTokensDetails, OutputTokensDetails
+
+from agents import ModelSettings, ModelTracing, OpenAIResponsesModel
+
+
+class DummyResponses:
+    async def create(self, **kwargs):
+        self.kwargs = kwargs
+
+        class DummyResponse:
+            id = "dummy"
+            output = []
+            usage = type(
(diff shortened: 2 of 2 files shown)


seratch:
@codex review this

src/agents/models/openai_responses.py
```diff
@@ -259,6 +259,10 @@ async def _fetch_response(
         include_set: set[str] = set(converted_tools.includes)
         if model_settings.response_include is not None:
             include_set.update(model_settings.response_include)
+
+        if model_settings.store is False and model_settings.response_include is None:
+            include_set.add("reasoning.encrypted_content")
```
seratch, line 264: non-reasoning models such as gpt-4.1 do not require this (actually it does not work), so always having this does not work. why don't you manually add this option to include?

seratch (commented):
Thanks for the suggestion but I don't think this is robust enough

src/agents/models/openai_responses.py
ihower: You are right, I forgot to consider that GPT-4.1 would throw errors, and this solution will cause breaking changes, which is not good. However, I do hope there’s a better way to help developers realize that when Store=False, they need to include reasoning.encrypted_content, otherwise GitHub issues will keep showing up.

Issue #1755, opened 2025-09-16
Streamed voice demo not worning


kun432:
### Describe the bug

Streamed voice demo not worning.

### Debug information

- Python: 3.12.10
- uv: 0.8.17
- Agents SDK: 0.0.3
- OpenAI SDK: 1.107.1
- OS: macOS 14.7.6

### Repro steps

```bash
git clone https://github.com/openai/openai-agents-python && cd openai-agents-python
make sync
export OPENAI_API_KEY=XXXXXXXXXX
uv run python -m examples.voice.streamed.main
```

Just after it starts, a parse error happens without any key strokes. Also, even if I press `K`, recording does not start.

<img width="1105" height="663" alt="Image" src="https://github.com/user-attachments/assets/63b03cc6-daef-43f5-8e2b-a8d8cb73a4e4" />

I added some codes to get traceback.

```
examples/voice/streamed/main.py --- 1/2 --- Python
14 14
15 15 from agents.voice import StreamedAudioInput, VoicePipeline
.. 16
.. 17 import traceback
16 18
17 19 # Import MyWorkflow class - handle both module and package use cases
18 20 if TYPE_CHECKING:

examples/voice/streamed/main.py --- 2/2 --- Python
171 173         except Exception as e:
172 174             bottom_pane = self.query_one("#bottom-pane", RichLog)
173 175             bottom_pane.write(f"Error: {e}")
... 176             tb = traceback.format_exc()
... 177             bottom_pane.write(f"[dim]{tb}[/dim]")
174 178         finally:
175 179             self.audio_player.close()
176 180
```

run again.

<img width="1105" height="663" alt="Image" src="https://github.com/user-attachments/assets/fa5ba9ae-381b-4ca0-987b-4777dc231e8d" />

### Expected behavior

Streamed voice "demo" should work out of the box

PR #1759, opened 2025-09-16
feat(voice)!: migrate STT streaming to match GA Realtime API

notV3NOM:
Fixes https://github.com/openai/openai-agents-python/issues/1755

#### Description

The streamed voice demo fails to run because the STT events have changed as part of Realtime GA.  
This PR migrates the STT voice model to be consistent with GA.
As such, this is a BREAKING change for those who want to use beta servers. 
 
#### Changes

1. Use GA event name `session.update` (instead of `transcription_session.update`) and GA session payload shape (like nested `audio.input` with PCM format object).
2. Update STT tests for the same.
3. Misc: Lint for `src/agents/tracing/spans.py`

### Tests

The demo works as expected and all tests/lint/type-checks pass.

### DIFF
--- a/src/agents/tracing/spans.py
+++ b/src/agents/tracing/spans.py
@@ -22,6 +22,7 @@ class SpanError(TypedDict):
         message: A human-readable error description
         data: Optional dictionary containing additional error context
     """
+
     message: str
     data: dict[str, Any] | None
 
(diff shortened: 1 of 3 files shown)


seratch (approved):
Thanks for sending this PR. I've confirmed this resolves the error

Issue #939, opened 2025-06-25
Expose tool call arguments in RunHooks on_tool_start and on_tool_end


skyfallsin:
### Background
Currently, the `RunHooks` API only exposes the `Tool` instance (e.g. `FunctionTool` or `ComputerTool`) to `on_tool_start` and `on_tool_end`, but does not include the raw arguments passed in the tool call. The `ToolContext` object does not carry these call arguments, so hook implementers cannot inspect or log the payload that the model requested.

### Proposal
1. **Include arguments on `ToolContext`**
   - Add an `arguments: Any` field to `ToolContext`, populated with the JSON-decoded payload of the tool call.
2. **Populate `ToolContext` with arguments**
   - Update runner code to call `ToolContext.from_agent_context(context_wrapper, call_id, tool_call.arguments)` when building the context.
3. **Expose `context.arguments` in hooks**
   - Users can then read `context.arguments` in `on_tool_start` and `on_tool_end`.

### Benefits
- Hooks can log or audit exactly what inputs were provided to each tool.
- Enables richer lifecycle monitoring and debugging.

### Example Hook Usage
```python
class LoggingHooks(RunHooks):
    async def on_tool_start(self, context: ToolContext, agent: Agent, tool: Tool):
        print(f"Tool {tool.name} started with args: {context.arguments!r}")

PR #1598, opened 2025-08-27
feat: Expose tool call arguments in ToolContext and lifecycle hooksFeat/tool  context arguments

uzair330:
### **Description**

This PR addresses [Issue #939](https://github.com/openai/openai-agents-python/issues/939) by enhancing the observability of tool calls within the Agents SDK. It exposes the raw JSON arguments of a tool call in the `ToolContext`, making them accessible in lifecycle hooks like `on_tool_start` and `on_tool_end`.

#### **Why These Changes Are Required and Their Importance**

Currently, developers using lifecycle hooks have visibility into *which* tool is being called (`tool_name`) but lack access to the *arguments* being passed to it. This limits the ability to effectively log, debug, or implement custom logic based on the tool's inputs.

This change introduces an `arguments` field to the `ToolContext`, providing the raw JSON string of the arguments sent by the model. The importance of this enhancement is threefold:

1.  **Enhanced Observability:** Developers can now create sophisticated logging and monitoring hooks that record the exact inputs for every tool call, which is invaluable for debugging agent behavior.
2.  **Improved Customization:** It enables advanced use cases, such as input validation or dynamic logic within hooks, before a tool is even executed.
3.  **Backward Compatibility:** The change is implemented in a non-breaking way. Existing agents and hooks will continue to function without any modification, while new capabilities are available for those who need them.

This aligns with the SDK's core goals of extensibility and traceability, giving developers deeper insights into the agent's execution flow.

### **Examples**

The following examples demonstrate how to access `context.arguments` within a custom `RunHooks` implementation across synchronous, asynchronous, and streaming executions.

#### 1. Hello World (Sync)
This example shows a simple agent with a custom hook that prints the arguments passed to the `say_hello` tool during a synchronous run.

```python
import os

from agents import Agent, RunHooks, Runner, function_tool
from agents.tool_context import ToolContext
@function_tool
def say_hello(name: str) -> str:
    return f"Hello, {name}!"

class LoggingArgsHooks(RunHooks):
    async def on_start(self, context, agent):
        pass

    async def on_end(self, context, agent, result):
        pass

    async def on_tool_start(self, context: ToolContext, agent, tool):
        print(f"[HOOK] Tool '{tool.name}' started with args: {context.arguments}")

    async def on_tool_end(self, context: ToolContext, agent, tool, result):
        print(f"[HOOK] Tool '{tool.name}' ended with args: {context.arguments}, result: {result}")


agent = Agent(
    name="HelloSyncAgent",
[...]

PR #1765, opened 2025-09-17
Add tool call arguments in ToolContext for RunHooks

ihower:
## Background 

Currently, the `RunHooks` lifecycle (`on_tool_start`, `on_tool_end`) exposes the `Tool` and `ToolContext`, but does not include the actual arguments passed to the tool call. 

resolves https://github.com/openai/openai-agents-python/issues/939

## Solution

This implementation is inspired by [PR #1598](https://github.com/openai/openai-agents-python/pull/1598).

* Add a new `tool_arguments` field to `ToolContext` and populate it via from_agent_context with tool_call.arguments.
* Update `lifecycle_example.py` to demonstrate tool_arguments in hooks
* Unlike the proposal in [PR #253](https://github.com/openai/openai-agents-python/issues/253), this solution is not expected to introduce breaking changes, making it easier to adopt.

### DIFF
--- a/src/agents/tool_context.py
+++ b/src/agents/tool_context.py
@@ -14,6 +14,10 @@ def _assert_must_pass_tool_name() -> str:
     raise ValueError("tool_name must be passed to ToolContext")
 
 
+def _assert_must_pass_tool_arguments() -> str:
+    raise ValueError("tool_arguments must be passed to ToolContext")
+
+
 @dataclass
 class ToolContext(RunContextWrapper[TContext]):
     """The context of a tool call."""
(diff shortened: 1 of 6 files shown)


seratch:
Thanks for the suggestion. I agree this is safer for us to accept than #253; @rm-openai let me know if you have any comments before I merge this within a few business days.

src/agents/tool_context.py
```diff
@@ -24,6 +24,9 @@ class ToolContext(RunContextWrapper[TContext]):
     tool_call_id: str = field(default_factory=_assert_must_pass_tool_call_id)
     """The ID of the tool call."""
 
+    tool_arguments: Optional[str] = None
```
rm-openai: can you make this non-optional, similar to `tool_call_id` above?

ihower: No problem, but this will require changes in quite a few test cases, since ToolContext will now always need to be passed a tool_arguments.

ihower: Done. I also updated realtime/session.py where ToolContext needs tool_arguments too.

PR #1768, opened 2025-09-18
Export `user_agent_override` contextmanager

jiwon-oai:
This allows us to override the user agent header when using agents with other OA python sdks.


examples/basic/tools.py
```diff
@@ -18,6 +18,7 @@ def get_weather(city: Annotated[str, "The city to get the weather for"]) -> Weat
     print("[debug] get_weather called")
     return Weather(city=city, temperature_range="14-20C", conditions="Sunny with wind.")
 
+
```
rm-openai: undo plz?

rm-openai (approved):

examples/basic/tools.py
jiwon-oai: Ack! fyi this got added when I ran `make format` - I'll undo it but it might reappear on other unrelated PRs!

PR #1772, opened 2025-09-18
Explain what is the displayed UI

inez:


seratch:
Thanks for sending this PR. README focuses on what this client side Agents SDK is, so we don't need the change this time.

PR #1781, opened 2025-09-22
Add a reusable type alias for agent instructions callbacks

Abbas-Asad:
### Summary
Add a reusable type alias for agent instructions callbacks and use it in `Agent.instructions` to improve readability and consistency.

### Changes Made
- Added `AgentInstructionsFunction` in `src/agents/agent.py`:
  - `AgentInstructionsFunction = Callable[[RunContextWrapper[TContext], "Agent[TContext]"], MaybeAwaitable[str]]`
- Updated `Agent.instructions` type to:
  - `instructions: str | AgentInstructionsFunction | None = None`
- No behavioral changes; typing/clarity only.

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -59,6 +59,12 @@ ToolsToFinalOutputFunction: TypeAlias = Callable[
 """
 
 
+AgentInstructionsFunction: TypeAlias = Callable[
+    [RunContextWrapper[TContext], "Agent[TContext]"],
+    MaybeAwaitable[str],
+]
+
+
 class StopAtTools(TypedDict):
     stop_at_tool_names: list[str]
     """A list of tool names, any of which will stop the agent from running further."""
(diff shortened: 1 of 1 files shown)


src/agents/agent.py
```diff
@@ -59,6 +59,12 @@ class ToolsToFinalOutputResult:
 """
 
 
+AgentInstructionsFunction: TypeAlias = Callable[
+    [RunContextWrapper[TContext], "Agent[TContext]"],
```
seratch, line 63: We don't want to make this `Agent[TContext]` type be a string just for extracting the type here

seratch:
Thanks for sending this patch but:
- we don't want to make the `Agent[TContext]` type to be string just for extracting the type
- the type is not reused at this moment, so I don't see the benefit of reusablity

For these reasons, let us close this.

PR #1783, opened 2025-09-22
docs: added LLM context usage example

AbdulSamad94:
Added **Agent/LLM context** code example in the Context Management docs.
This shows how to pass static user info through the `instructions` parameter so the LLM can access it during response generation.

### Details

* Added a minimal code snippet demonstrating `Agent` with an `instructions` system prompt.
* Complements the existing **Local context** example for better understanding of the two context types.


seratch:
Thanks for sending this. However, we don't need the code snippet, which just demonstrates one of the use cases, in the section.

PR #1785, opened 2025-09-22
feat: add Redis session support for scalable distributed memory

damianoneill:
- Add RedisSession class with full Session protocol implementation
- Support Redis URL connection strings and direct client injection
- Include TTL (time-to-live) support for automatic session expiration
- Add key prefixes for multi-tenancy and namespace isolation
- Implement atomic operations using Redis pipelines for data integrity
- Add comprehensive test suite with in-memory fakeredis support
- Include detailed example demonstrating Redis session features
- Update documentation with Redis session reference
- Add redis extra dependency group for easy installation

The RedisSession enables production-grade, distributed session memory that scales across multiple application instances while maintaining full compatibility with the existing Session interface.

### DIFF
--- a/src/agents/extensions/models/litellm_model.py
+++ b/src/agents/extensions/models/litellm_model.py
@@ -23,6 +23,7 @@ from openai.types.chat import (
     ChatCompletionChunk,
     ChatCompletionMessageCustomToolCall,
     ChatCompletionMessageFunctionToolCall,
+    ChatCompletionMessageParam,
 )
 from openai.types.chat.chat_completion_message import (
     Annotation,
@@ -267,6 +268,10 @@ class LitellmModel(Model):
             input, preserve_thinking_blocks=preserve_thinking_blocks
         )
 
+        # Fix for interleaved thinking bug: reorder messages to ensure tool_use comes before tool_result  # noqa: E501
+        if preserve_thinking_blocks:
+            converted_messages = self._fix_tool_message_ordering(converted_messages)
+
         if system_instructions:
             converted_messages.insert(
                 0,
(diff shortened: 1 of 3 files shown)


damianoneill:
@codex review

damianoneill:
@seratch I've raised the coverage to 87%, but this is still failing the coverage check. 

It will be difficult to add any more coverage for this code. 

There is other memory code, not related to this PR, with much less coverage, e.g. src/agents/memory/openai_conversations_session.py has 27%, this might be better targets. 

Also, I wasnt aware there was a coverage issue when developing, the make check target runs make tests, not make coverage, might be good to update this?

damianoneill:
@seratch I've added a test file for openai_conversations_session.py even though this is not directly related to this PR. This has increased the coverage for this code from 27% to 79%, this bumps the overall coverage to 95%. 

Someone who understand this capability better should review this test file, I've just asserted logic based on the implementation, I dont understand this feature and therefore do not know if the assertions are sensible.

examples/basic/dynamic_system_prompt.py
```diff
@@ -28,6 +28,7 @@ def custom_instructions(
     instructions=custom_instructions,
 )
 
+
```
seratch:
```suggestion
```

examples/basic/tools.py
```diff
@@ -18,6 +18,7 @@ def get_weather(city: Annotated[str, "The city to get the weather for"]) -> Weat
     print("[debug] get_weather called")
     return Weather(city=city, temperature_range="14-20C", conditions="Sunny with wind.")
 
+
```
seratch:
```suggestion
```

tests/test_openai_conversations_session.py
```diff
@@ -0,0 +1,445 @@
+"""Tests for OpenAI Conversations Session functionality."""
```
seratch, line 1: thanks for adding this one

PR #1792, opened 2025-09-23
Adding tool input and output guardrails

steven10a:
- This PR was started from [PR 1606: Tool Guardrails](https://github.com/openai/openai-agents-python/pull/1606)
- It adds input and output guardrails at the tool level which can trigger `ToolInputGuardrailTripwireTriggered` and `ToolOutputGuardrailTripwireTriggered` exceptions
- It includes updated documentation, a runnable example, and unit tests
- `make check` and unit tests all pass

## Edits since last review:
- Extracted nested tool running logic in `_run_impl.py`
- Added rejecting tool call or tool call output and returning a message to the model (rather than only raising an exception)
- Added the tool guardrail results to the `RunResult`
- Removed docs


src/agents/tool_guardrails.py
```diff
@@ -0,0 +1,204 @@
...
+
+    agent: Agent[Any]
+    """
+    The agent that is executing the tool.
+    """
+
+    tool_call: ResponseFunctionToolCall
+    """
+    The tool call data including the function name and arguments.
+    """
```
rm-openai: for now can we get rid of this? the ToolContext contains the name, args, call_id already.

src/agents/_run_impl.py
```diff
@@ -569,24 +579,66 @@ async def run_single_tool(
...
-                    _, _, result = await asyncio.gather(
-                        hooks.on_tool_start(tool_context, agent, func_tool),
-                        (
-                            agent.hooks.on_tool_start(tool_context, agent, func_tool)
-                            if agent.hooks
-                            else _coro.noop_coroutine()
-                        ),
-                        func_tool.on_invoke_tool(tool_context, tool_call.arguments),
-                    )
+                    # 1) Run input tool guardrails, if any
```
rm-openai, line 756: this is getting nested enough that it would be nice to extract out. Feel free to do that if you want - but not required, I can follow up with that later too

src/agents/tool_guardrails.py
```diff
@@ -0,0 +1,204 @@
...
+
+    tripwire_triggered: bool
+    """
+    Whether the tripwire was triggered. If triggered, the tool execution will be halted.
+    """
+
+    model_message: str | None = None
+    """
+    Message to send back to the model as the tool output if tripped.
+    """
```
rm-openai: I would expect this to actually be sent to the model automatically? Whereas looks like the current implementation just raises an error. Ideally we can do something like:
```
class InformModelBehavior(TypedDict):
  type: Literal["inform_model"]
  message: str

class RaiseExceptionBehavior(TypedDict):
  type: Literal["raise_exception"]

class ToolGuardrailFunctionOutput:
   ...
  
   behavior: InformModelBehavior | RaiseExceptionBehavior 
 ```
 
 Just a proposal, feel free to do it some other way. Goal is to let the developer decide if the guardrail tripping should:
 1. Continue the run, sending the output to the model
 2. End the run by raising an exception

File: docs/guardrails.md [hunk 4]
```diff
@@ -152,3 +152,83 @@ async def main():
 2. This is the guardrail's output type.
 3. This is the guardrail function that receives the agent's output, and returns the result.
 4. This is the actual agent that defines the workflow.
+
+## Tool guardrails
```
seratch: The document updates need to be in a separate PR to better control the timing of merging code and releasing a new version.

examples/basic/tools.py
```diff
@@ -18,6 +18,7 @@ def get_weather(city: Annotated[str, "The city to get the weather for"]) -> Weat
     print("[debug] get_weather called")
     return Weather(city=city, temperature_range="14-20C", conditions="Sunny with wind.")
 
+
```
seratch: Can you avoid unrelated changes like this one?

src/agents/tool_guardrails.py
steven10a: Removed ✅

src/agents/_run_impl.py
steven10a: Extracted this into separate tool guardrail execution functions

src/agents/tool_guardrails.py
steven10a: There are now three behaviors which the developer sets when defining the tool guardrail:
- `allow`: allow normal execution
- `reject`: reject a function call or reject the function call output and return a message to the agent telling it why
- `raise exception`: Trigger tripwire and end the run execution

examples/basic/tools.py
steven10a: Removed, this was added during `make check`

docs/guardrails.md
steven10a: Removed docs for now

PR #1794, opened 2025-09-23
examples: make function-call-streaming example safer (dataclass + safer attribute access + instruction fix)

AbdulSamad94:
This PR improves the "function call arguments streaming" example by fixing small bugs and making the example more robust and type-safe.

Key changes
-----------
- Use the `FunctionCallInfo` dataclass consistently (replace dict-style access with attribute access).
- Fix incorrect annotation/assignment for `current_active_call_id` (`Optional[str]` -> `current_active_call_id: Optional[str] = None`).
- Guard access to `result.final_output` with `getattr(..., None)` to avoid AttributeError if the field is absent.
- Add JSON parsing attempt for streamed function arguments and print parsed/raw args on completion.
- Minor code tidy and safer attribute access using `getattr`.

Important note about models / why this change was needed
--------------------------------------------------------
- The original example sometimes failed to run with Gemini (and other models) because the example relied on the agent following a strict instruction to make function calls one-at-a-time. In practice some models produced overlapping/parallel `function_call` items which caused the example to error or behave unexpectedly.
- This PR improves the example (clear instruction + safer event handling and parsing) so it's more resilient to variations in model behavior. Still: for real-world usage, consider adding explicit multi-`call_id` handling in a follow-up.

Why
---
These changes prevent runtime AttributeError, make the example consistent with the dataclass usage, and make it safer to run against different SDK event shapes and model behaviors.

Notes / Future suggestions
--------------------------
- The example still documents the sequential-call expectation; recommend adding a second example showing multi-`call_id` handling (concurrent function calls).
- Consider replacing `isinstance(...ResponseFunctionCallArgumentsDeltaEvent)` with a defensive `getattr(data, "delta", None)` check for even greater resilience.
- Add a small unit/integration test that mocks `result.stream_events()` to prevent regressions.

Testing
-------
1. Run the example locally with your configured SDK and model (e.g., Gemini / other models).
2. Verify streaming shows argument chunks, prints "Parsed args" (or raw args if parsing fails), and prints a final summary without exceptions.
3. If the example still fails for some model, add a recorded sample event stream to the tests to help reviewers reproduce.

Checklist
---------
- [x] Apply code fixes (dataclass usage, current_active_call_id annotation, safe final_output access)
- [x] Manual run against configured SDK/model
- [x] Add 1-line note in the example header about "assumes sequential function calls"
- [ ] (Optional) Add unit/integration test that mocks streaming events

### DIFF
--- a/examples/basic/stream_function_call_args.py
+++ b/examples/basic/stream_function_call_args.py
@@ -1,9 +1,17 @@
 import asyncio
-from typing import Annotated, Any, Optional
+import json
+from dataclasses import dataclass
+from typing import Annotated, Optional
 
 from openai.types.responses import ResponseFunctionCallArgumentsDeltaEvent
 
 from agents import Agent, Runner, function_tool
+
+
+@dataclass
+class FunctionCallInfo:
+    name: str
+    arguments: str = ""
 
 
 @function_tool
@@ -31,7 +39,8 @@ async def main():
     """
     agent = Agent(
         name="CodeGenerator",
-        instructions="You are a helpful coding assistant. Use the provided tools to create files and configurations.",
+        instructions="""You are a helpful coding assistant. Use the provided tools to create files and configurations. 
+        **Important: You must use the tools one at a time. Complete one function call before starting another.**""",
         tools=[write_file, create_config],
     )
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. However, I don't see so much difference with your change and to me, the current one looks simpler. Thanks again for your interest.

AbdulSamad94:
@seratch thanks for the review!
Just to share some testing context: with the current instructions of CodeGenerator Agent the example **fails on Gemini (and a few other models)** because the model sometimes triggers **two function\_call items at once**, which leads to a 400 error like this:

```
openai.BadRequestError: Error code: 400 - [{'error': {'code': 400, 'message': 'Request contains an invalid argument.', 'status': 'INVALID_ARGUMENT'}}]
```

Even a minimal run like:

```
uv run stream_function_call_args.py
```

shows overlapping calls:

```
{"dependencies":["fastapi","uvicorn"],"project_name":"my-app","version":"1.0.0"}
{"content":"...","filename":"main.py"}
```

The main issue is the instruction text, if we **tighten it slightly** (e.g. “Strictly call one tool at a time; wait for the previous call to finish before starting the next”) it runs fine on Gemini and OpenAI models.

Would you be open to a very small follow-up PR that *only* updates the instructions in the example so it’s model-agnostic, without the extra type-safety changes?

seratch:
If `model_settings.parallel_tool_calls: False` works for your use case, we're open to adjust the example just by setting the option.

AbdulSamad94:
> If `model_settings.parallel_tool_calls: False` works for your use case, we're open to adjust the example just by setting the option.

I tried with

```python
Agent(..., model_settings=ModelSettings(parallel_tool_calls=False),
```

but Gemini still fails:

* With `False`: `BadRequestError: Request contains an invalid argument`
* With `True`: `BadRequestError: Parallel tool calls are not supported`

So just setting the flag doesn’t fix the issue.
The root cause is that the example allows overlapping function calls when the model ignores the “one-at-a-time” instruction.
Maybe we should tighten the instructions (or add explicit multi-call handling) so it’s more robust across models like Gemini.

ihower:
@AbdulSamad94 

The root cause is not that this example "allows" parallel function calls. OpenAI models natively support parallel function calling, so I think there’s no need to artificially disable it.

Based on your error, I assume you’re directly using Gemini’s OpenAI-compatible endpoint.

The real issue is with Gemini’s OpenAI-compatible endpoint implementation. See related discussions:

* https://discuss.ai.google.dev/t/gemini-openai-compatibility-issue-with-tool-call-streaming/59886
* https://discuss.ai.google.dev/t/tool-calling-with-openai-api-not-working/60140

These are incompatibilities on the Gemini side, not something the SDK example itself needs to workaround. 
I’d suggest reporting them to Google so they can fix their OpenAI endpoint behavior.

I also tested LiteLLM support. Unfortunately, it runs into other errors — in this case a `json.decoder.JSONDecodeError`.

I think this is a LiteLLM issue: when handling multiple function calls, it fails to handle the index properly and ends up concatenating arguments incorrectly. For reference, Google ADK added a workaround to defensively handle this issue:

* https://github.com/google/adk-python/pull/759

I think if OpenAI Agents SDK needs to fix this, it may also have to patch the LiteLLM integration layer in a similar way to Google ADK.

That’s what I’ve found so far.
/cc @seratch

AbdulSamad94:
Thanks for clarifying.
I retested with multiple providers to be sure:

* **gemini-2.5-flash** ✅ works as-is
* **gemini-2.5-pro / gemini-2.0-flash** ❌ 400 *INVALID\_ARGUMENT* unless I explicitly add “step-by-step / one-tool-at-a-time” instructions → then works ✅
* **x-ai/grok-4-fast**, **deepseek-chat-v3.1** ✅
* **nvidia/nemotron-nano-9b-v2** ❌ provider JSON parse error (seems unrelated to the SDK)

So I agree: the example itself is fine and disabling `parallel_tool_calls` isn’t needed.
The failures come from Gemini’s OpenAI-compatible endpoint being stricter with streamed function-call JSON.

Adding a small note in the example (like. “Some providers like Gemini 2.x may require explicit sequential instructions”) might save other users time, but no SDK change is required.

PR #1795, opened 2025-09-24
Feat: Expose MCP message handler configuration

freeznet:
Exposed the optional MCP message_handler, letting each transport surface the hook and ensuring the base session forwards it on.


seratch:
Thanks for sending this PR. Since this SDK doesn’t support MCP sessions, I’m not sure how valuable this exposure actually is.

freeznet:
@seratch Appreciate the look! Even though we don’t surface full MCP sessions to SDK consumers yet, these server helpers already spin up `ClientSession` instances under the hood. Several integrations are relying on notifications (tools list changes, etc.), and today there’s no way to intercept those without forking the server classes. By threading the optional `message_handler` through we let advanced users react to those notifications to bridging server events into agent logic, without changing default behaviour. Since the hook is a straight pass-through, it stays no-op unless someone opts in, so the surface risk is very low while unblocking that customization point.

freeznet:
@seratch could you please help to review the PR when you have time, thanks

PR #1804, opened 2025-09-25
provide env var to enable tool error logging

danilbogdan:
Currently there is no way to log errors other than providing failure_error_function in each function_tool decorator use. Current PR provides env variable OPENAI_AGENTS_LOG_TOOL_ERRORS that enables such logs.


danilbogdan:
Done. However, I’m now uncertain about the purpose of this flag.

The reason of PR is that I couldn’t detect tool failures in production, not just in debug mode. I assumed that not everyone wanted to log tool errors, which is why I included the flag that, by default, maintains the current behavior but allows enabling error logs.

Now, by default, we log everything, and in debug mode, we can hide some data which is useless imho.

@rm-openai please take a look at this.

seratch (approved):

PR #1807, opened 2025-09-26
Propagate OpenAI request id into streaming traces

marysha-openai:
## Summary
- capture the x-request-id from streamed OpenAI Responses API calls and attach it to the final response before tracing
- ensure traced streaming responses persist the request id when tracing data is recorded
- add a regression test that stubs the streamed client and verifies the ResponseSpanData holds the request id after Runner.run_streamed

## Testing
- make format
- make lint
- make mypy
- make tests
- uv run pytest tests/test_agent_tracing.py::test_streamed_response_request_id_recorded

------
https://chatgpt.com/codex/tasks/task_i_68d5e042271c832e806c32939e01610d

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal, cast, overload
 
 from openai import NOT_GIVEN, APIStatusError, AsyncOpenAI, AsyncStream, NotGiven
+from openai._models import add_request_id
 from openai.types import ChatModel
 from openai.types.responses import (
     Response,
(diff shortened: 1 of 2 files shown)


src/agents/models/openai_responses.py
```diff
@@ -7,6 +7,7 @@
 from typing import TYPE_CHECKING, Any, Literal, cast, overload
 
 from openai import NOT_GIVEN, APIStatusError, AsyncOpenAI, AsyncStream, NotGiven
+from openai._models import add_request_id
```
seratch, line 10: Accessing `_models` module may be broken when the underlying "openai" package makes breaking changes in the future. We've been dealing with incompatible changes on the "openai" package side several time. Accessing underscore prefixed (kind of private) module could be even more risky in the long run.

seratch (commented):
The change is fine, but accessing `openai._models` may be a bit fragile in the long run. We may want to move the function from `_models` module for safety (meaning making changes to openai package first).

src/agents/models/openai_responses.py
marysha-openai: yea makes sense - i'll look into doing this differently later today

Issue #1789, opened 2025-09-23
Duplicate item found with id fc_xxxx when using conversation_id with function calling


kinnari-patwa:
### Describe the bug
When using conversation_id in Runner.run_streamed method for an agent with function calling (custom function), I get the the following error:
{'error': {'message': 'Duplicate item found with id fc_68d25eda79d481959bd7e424738091a006b9ae17d036a0e7. Remove duplicate items from your input and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}

### Debug information
- Agents SDK version: 0.3.0
- Python version 3.12

### Repro steps
       @function_tool
        async def query_database(wrapper: RunContextWrapper[AgentContext], query: str) -> str:
            """Query the database with the provided SQL query. Only SELECT queries are allowed."""
            return await self.agent_tools.query_database(query, wrapper.context.websocket)
        
        agent = Agent(
            name="Database Assistant",
            instructions="You are a helpful assistant that can query databases. Use the query_database function to execute SQL queries when users ask for data.",
            model="gpt-4o",
            model_settings=ModelSettings(temperature=0.2),
            tools=[query_database]
        )

	run_stream = Runner.run_streamed(
                    agent,
                    [user_message],
                )

Try any user prompt that results in custom function call - query_database. You will get the following error:
Error - {'error': {'message': 'Duplicate item found with id fc_68d25eda79d481959bd7e424738091a006b9ae17d036a0e7. Remove duplicate items from your input and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}

### Expected behavior
run_streamed should run without any errors and allow function calling (custom function) when using conversation_id

ihower:
According to the docs: https://openai.github.io/openai-agents-python/sessions/#openai-conversations-api-memory
you should be using `OpenAIConversationsSession` if you want to integrate with the OpenAI Conversations API. That is the officially supported way right now.

But I also confirmed that using `conversation_id` directly does not work: the same error happens with both `run_streamed` and `run_sync`. Here is a minimal reproducible example:

```python
from agents import Agent, function_tool, Runner
from openai import OpenAI
client = OpenAI()

conversation = client.conversations.create()
conv_id = conversation.id

@function_tool
async def foobar() -> str:
    return "41"

agent = Agent(
    name="Assistant",
    instructions="just call foobar and answer me only the number",
    model="gpt-4.1-mini",
    tools=[foobar]
)

[...]

PR #1827, opened 2025-09-28
fix: #1789 Fix multi-turn handling for conversation_id and previous_response_id: only send new items

ihower:
This PR fixes incorrect behavior when using `conversation_id` or `previous_response_id` during internal multi-turn calls, i.e. when function calling is involved.

It provides better support for the OpenAI Conversation state feature, as described in the https://platform.openai.com/docs/guides/conversation-state?api-mode=responses, aligning with the intended API design.

Resolved: https://github.com/openai/openai-agents-python/issues/1789

### Problem

- When using **conversation_id**, the runner resent all items on every turn, which caused server-side duplicate item error. 
- When using **previous_response_id**, the SDK also resent all items and the `previous_response_id` was not updated across internal loops. This led to duplicated items and wasted bandwidth, even though it did not trigger errors currently. On the OpenAI platform dashboard logs, this can be observed as the responses repeatedly forking (based on the initially provided `previous_response_id`) instead of properly chaining.
    

### Minimal repro code:

```
from agents import Agent, function_tool, Runner
from openai import OpenAI
client = OpenAI()

conversation = client.conversations.create()
conv_id = conversation.id

@function_tool
async def foobar() -> str:
    return "41"

agent = Agent(
    name="Assistant",
    instructions="just call foobar and answer me only the number",
    model="gpt-4.1-mini",
    tools=[foobar]
)

Runner.run_sync(agent, "hi", conversation_id=conv_id)
```

This produces the following error:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'Duplicate item found with id fc_68d281ab76108195ac719f9b7c48eac103032f3a88a761f2. Remove duplicate items from your input and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}
```

### Solution

- Added a _ServerConversationTracker to track items that should not be sent:
    - **conversation_id**: only send  new items on each turn.
    - **previous_response_id**: only send new items and update `previous_response_id` to the latest server response.
- Added tests to cover six scenarios:
  * **conversation_id**
    * non-streaming: `test_conversation_id_only_sends_new_items_multi_turn`
    * streaming: `test_conversation_id_only_sends_new_items_multi_turn_streamed`
  * **previous_response_id**
    * non-streaming: `test_previous_response_id_only_sends_new_items_multi_turn`
    * streaming: `test_previous_response_id_only_sends_new_items_multi_turn_streamed`
  * **default (no conversation_id or previous_response_id)**
    * non-streaming: `test_default_send_all_items`
    * streaming: `test_default_send_all_items_streamed`
- Updated an existing test that checks the `previous_response_id` of the last model response.
  * Since behavior has changed to update `previous_response_id` on each turn, the assertion was modified accordingly.
- Updated documentation to clarify how to use conversation_id and previous_response_id with server-managed conversations.
- Unlike session memory e.g. `OpenAIConversationsSession`, which does not support passing lists of items containing images or files by default, this approach works out of the box and fully supports such items between turns.

### DIFF
--- a/docs/running_agents.md
+++ b/docs/running_agents.md
@@ -121,6 +121,71 @@ Sessions automatically:
 
 See the [Sessions documentation](sessions.md) for more details.
 
+
+### Server-managed conversations
+
+You can also let the OpenAI conversation state feature manage conversation state on the server side, instead of handling it locally with `to_input_list()` or `Sessions`. This allows you to preserve conversation history without manually resending all past messages. See the [OpenAI Conversation state guide](https://platform.openai.com/docs/guides/conversation-state?api-mode=responses) for more details.
+
+OpenAI provides two ways to track state across turns:
+
+#### 1. Using `conversation_id`
+
+You first create a conversation using the OpenAI Conversations API and then reuse its ID for every subsequent call:
+
+```python
+from agents import Agent, Runner
+from openai import AsyncOpenAI
+
+client = AsyncOpenAI()
+
+async def main():
+    # Create a server-managed conversation
+    conversation = await client.conversations.create()
+    conv_id = conversation.id    
+
+    agent = Agent(name="Assistant", instructions="Reply very concisely.")
+
+    # First turn
+    result1 = await Runner.run(agent, "What city is the Golden Gate Bridge in?", conversation_id=conv_id)
+    print(result1.final_output)
+    # San Francisco
+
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for sending this. Overall, this looks great! I have a few other things on my plate right now, but I will look into details shortly.

PR #1832, opened 2025-09-30
feat: #1831 Add opt-in cost tracking for LiteLLM models

### LINKED ISSUE #1831: Add cost tracking for LiteLLM models
**Problem:**
When using LiteLLM models, there's no way to track how much each request costs. LiteLLM provides this info via `_hidden_params["response_cost"]` and their `completion_cost()` function, but the SDK doesn't expose it.

**Proposed Solution:**
Add opt-in cost tracking via `ModelSettings` that populates `usage.cost` with [cost estimates from LiteLLM](https://docs.litellm.ai/docs/completion/token_usage).

**Usage:**
```python
agent = Agent(
    model=LitellmModel(model="...", api_key="..."),
    model_settings=ModelSettings(
        include_usage=True,
        track_cost=True,  # opt-in
    ),
)

result = await Runner.run(agent, "Hello")
print(f"Cost: ${result.context_wrapper.usage.cost}")
```

habema:
Resolves #1831

Adds cost tracking for LiteLLM models when users enable it via `ModelSettings(track_cost=True)`.

**Changes**
- Added `cost` field to `Usage` class
- Added `track_cost` setting to `ModelSettings`. Opt-in (defaults to `False`)
- LiteLLM cost extraction:
  - Non-streaming: reads from `_hidden_params["response_cost"]`
  - Streaming: calculates via `litellm.completion_cost()`
  - Includes cached & reasoning tokens for accurate pricing


src/agents/extensions/models/litellm_model.py
```diff
@@ -124,6 +124,15 @@ async def get_response(
 
             if hasattr(response, "usage"):
                 response_usage = response.usage
+
+                # Extract cost from LiteLLM's hidden params if cost tracking is enabled.
+                cost = None
+                if model_settings.track_cost:
+                    if hasattr(response, "_hidden_params") and isinstance(
```
seratch: We'd like to avoid accessing the internal modules

habema: Make sense, this has now been refactored

habema:
`litellm.completion_cost()` is not used for both streaming and non-streaming responses.

src/agents/model_settings.py
```diff
@@ -120,6 +120,12 @@ class ModelSettings:
     """Whether to include usage chunk.
     Only available for Chat Completions API."""
 
+    track_cost: bool | None = None
```
seratch, line 123: This is only related to LiteLLM use cases. Even if we decided to have this feature, I don't think this property should be here. If we add something like this, we may want to introduce LiteLLMSettings, which has only track_cost so far, and pass it to LiteLLMModel constructor as a new arg.

src/agents/run.py
```diff
@@ -1179,6 +1179,9 @@ async def _run_single_turn_streamed(
             prompt=prompt_config,
         ):
             if isinstance(event, ResponseCompletedEvent):
+                # Extract cost if it was attached by LiteLLM model.
+                cost = getattr(event.response, "_litellm_cost", None)
```
seratch, line 1183: Sorry, but I still hesitate to rely on LiteLLM's underscored property

src/agents/model_settings.py
habema: I actually like this idea. I thought it could be too big of a change, but if this is welcome I will go ahead with this.

src/agents/run.py
habema: How about storing the cost in a `_last_stream_cost`  attribute on the LitellmModel instance instead, and have run.py extract it from there?

PR #1839, opened 2025-10-01
fix: making input_file's filename optional in code for supporting non-OpenAI models

ant31:
For input_file the  `filename` parameter was recently made required in https://github.com/openai/openai-agents-python/pull/1513. 
This conflicts with many models outside of OpenAI, which do not require it.

This PR makes `filename` optional again to improve compatibility across the ecosystem. 
For gpt* models, the server already handles missing filename errors, making the current client-side validation redundant.

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -293,17 +293,12 @@ class Converter:
                     raise UserError(
                         f"Only file_data is supported for input_file {casted_file_param}"
                     )
-                if "filename" not in casted_file_param or not casted_file_param["filename"]:
-                    raise UserError(f"filename must be provided for input_file {casted_file_param}")
-                out.append(
-                    File(
-                        type="file",
-                        file=FileFile(
-                            file_data=casted_file_param["file_data"],
-                            filename=casted_file_param["filename"],
-                        ),
-                    )
-                )
+                filedata = FileFile(file_data=casted_file_param["file_data"])
+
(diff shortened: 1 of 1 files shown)


seratch (approved):
>For gpt* models, the server already handles missing filename errors, making the current client-side validation redundant.

Thanks, this makes sense. This code is used by LiteLLM integration too, so this client-side validation would be a blocker for some models.

Sourav-Nandy-ai:
Hi Antoine,

I maintain agent-eval (agent_regress), an open-source harness for before-and-after behavior comparison of agent code. I pulled openai-agents==0.3.3 (the last release before this merge) and 0.4.0 (the first release after) into separate virtual environments. I then ran the same input_file item, where file_data was present but no filename was provided, through Converter.extract_all_content on both sides.

The Results
- Pre-fix: It results in a real UserError: filename must be provided for input_file {...} error.
- Post-fix: The execution succeeds, and the returned content part genuinely omits the filename rather than defaulting it, which matches the diff exactly.

This is a good, clean walk-back of the regression from #1513. Testing it this way confirms the fix against the actual installed package rather than just reading the diff.

If it would be useful, agent-eval can also run this as an N-run before-and-after distributional comparison (using a Mann-Whitney U test and bootstrap CI) across a fuller LiteLLM-routed test suite rather than just a single validation check. I am happy to share the script if you are interested.

You can find the repository here: [https://github.com/RudrenduPaul/agent-eval](https://www.google.com/search?q=https://github.com/RudrenduPaul/agent-eval)

Best, 
Sourav

PR #1851, opened 2025-10-03
Update multi_agent.md

Pritiks23:
Added a diagram to takeaway from dense paragraphs


seratch:
Thanks for sending this, but this prevents translating those. Thus, we're unable to accept it.

PR #1852, opened 2025-10-03
fix: invalid model setting when passing prompt to Agent

seratch:
This pull request fixes the bug reported for TS SDK, which exists for Python too.

- https://github.com/openai/openai-agents-js/pull/563
- https://github.com/openai/openai-agents-js/issues/562

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -6,7 +6,7 @@ from contextvars import ContextVar
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal, Union, cast, overload
 
-from openai import APIStatusError, AsyncOpenAI, AsyncStream, NotGiven, Omit, NOT_GIVEN, omit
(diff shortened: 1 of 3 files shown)


src/agents/models/openai_responses.py
```diff
@@ -306,7 +306,7 @@ async def _fetch_response(
             previous_response_id=self._non_null_or_not_given(previous_response_id),
             conversation=self._non_null_or_not_given(conversation_id),
             instructions=self._non_null_or_not_given(system_instructions),
-            model=self.model,
+            model=self.model if prompt is None else NOT_GIVEN,
```
rm-openai: can you add a comment here explaining why?

seratch:
@rm-openai good point; actually it's feasible. so, when a developer explicitly pass a model, it should be allowed. the default model still needs to be ignored.

seratch:
@codex Can you review the changes with fresh eyes?

PR #1855, opened 2025-10-04
Fix Local shell tool: return tool output to the LLM

ihower:
Based on prior work in https://github.com/openai/openai-agents-python/pull/732, the Local Shell Tool was nearly complete but still didn’t work in practice. This PR finishes the logic, adds tests, and provides a working example.
    
## Changes 
    
1. Return tool output to the LLM

    The local command ran, but its **output was not returned to the LLM**. This PR sends the output back, as described in the Local Shell guide: [https://platform.openai.com/docs/guides/tools-local-shell](https://platform.openai.com/docs/guides/tools-local-shell)
    
2. Use `call_id` 

    Upstream (openai-python) `LocalShellCallOutput` uses **`id`**, but the server actually expects **`call_id`**. This PR sets `call_id` so the server accepts the output. A small type-check ignore workaround is included until the upstream type is fixed.
    
3. Adds `examples/tools/local_shell.py` a working example.
    
4. Add Tests

## Before fix

You would get:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'No tool output found for local shell call call_MEmrSaOboO5WG3MXGazSubb7.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}
```

## After fix

The example runs successfully and returns the expected result.

### DIFF
--- a/examples/tools/local_shell.py
+++ b/examples/tools/local_shell.py
@@ -1,0 +1,45 @@
+import asyncio
+import os
+import subprocess
+
+from agents import Agent, LocalShellCommandRequest, LocalShellTool, Runner, trace
+
+
+def shell_executor(request: LocalShellCommandRequest) -> str:
+    args = request.data.action
+
+    try:
+        completed = subprocess.run(
+            args.command,
+            cwd=args.working_directory or os.getcwd(),
+            env={**os.environ, **args.env} if args.env else os.environ,
+            capture_output=True,
+            text=True,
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for sending this. It looks good. I will check after [DevDay](https://devday.openai.com/)!

seratch:
Thank you so much for being patient here. I will check this PR this week.

seratch (approved):
Thanks for sending this patch. I've confirmed this fix works well.

Issue #1767, opened 2025-09-18
ReasoningItem of RunStreamEvents are getting emitted out of order


waziers:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) - Yes
- **Have you searched for related issues?** Others may have faced similar issues. - Yes

### Describe the bug

ReasoningItems events are getting emitted out of order.  For example, we get `Tool Call Initiated`, `Reasoning Item`, `Tool Call Output` which is not in line with the spec.

### Debug information
- Agents SDK version: (e.g. `v0.0.3`)
- Python version (e.g. Python 3.10)

### Repro steps

Take the stock streaming example script from the library and add some additional print statements. 

```
import asyncio
import random

from agents import Agent, ItemHelpers, ModelSettings, Runner, function_tool
from openai.types.shared import Reasoning


@function_tool
def how_many_jokes() -> int:
    """Return a random integer of jokes to tell between 1 and 10 (inclusive)."""
    return random.randint(1, 10)


async def main():
    agent = Agent(
        name="Joker",
        model="gpt-5",
        model_settings=ModelSettings(
            reasoning=Reasoning(
                effort="high",
                summary="auto"
            )
        ),
        instructions="First call the `how_many_jokes` tool, then tell that many jokes.",
        tools=[how_many_jokes],
    )

    result = Runner.run_streamed(
        agent,
        input="Hello"

    )

    print("\n" + "="*60)
    print("🚀 AGENT STREAM STARTING")
    print("="*60 + "\n")

    previous_data_type = None
    reasoning_buffer = []

    async for event in result.stream_events():
        # Handle raw response events with progress indicators
        if event.type == "raw_response_event":
            data_type = event.data.type

            # Special handling for different event types
            if data_type == "reasoning":
                if data_type != previous_data_type:
                    print("🧠 [REASONING] ", end='', flush=True)
                print('▸', end='', flush=True)
            elif data_type == "tool_calls":
                if data_type != previous_data_type:
                    print("\n🔧 [TOOL CALLS] ", end='', flush=True)
                print('▸', end='', flush=True)
            elif data_type == "content":
                if data_type != previous_data_type:
                    print("\n💬 [CONTENT] ", end='', flush=True)
                print('▸', end='', flush=True)
            elif "response.output_item.done" in data_type.lower():
                if data_type != previous_data_type:
[...]

Issue #583, opened 2025-04-24
Ordering of events in Runner.run_streamed is incorrect


PrinceGupta1999:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug
When using `Runner.run_streamed`, the events for `run_item_stream_event` are not received until the agent message and tool execution completes. If `on_tool_start` and `on_tool_end` hooks are used, the events are streamed even after these methods return.

### Debug information
- Agents SDK version: `v0.0.12`
- Python version: `3.10.14`

### Repro steps
Adapted from 
```python
import asyncio
import random
from typing import Any

from agents import Agent, AgentHooks, ItemHelpers, set_default_openai_key, RunContextWrapper, Runner, Tool, function_tool

set_default_openai_key('sk-your-openai-key')

class JokeAgentHooks(AgentHooks):
    async def on_tool_start(self, wrapper: RunContextWrapper[Any], agent: Agent, tool: Tool):
        print(f"-- Hook: On Tool Start --")

    async def on_tool_end(
        self, wrapper: RunContextWrapper[Any], agent: Agent, tool: Tool, result: Any
    ):
        print(f"-- Hook: On Tool End --")

@function_tool
def how_many_jokes() -> int:
    print(f"-- Tool Execution --")
    # await asyncio.sleep(1)
    return random.randint(1, 10)


async def main():
    agent = Agent(
        name="Joker",
        model="gpt-4.1",
        instructions='Tell the user you are searching jokes for them and call the `how_many_jokes` tool simultaneously. Then tell that many jokes',
        tools=[how_many_jokes],
        hooks=JokeAgentHooks()
    )

    result = Runner.run_streamed(
        agent,
        input="Hello",
    )
    print("=== Run starting ===")
    async for event in result.stream_events():
        # We'll ignore the raw responses event deltas
        if event.type == "raw_response_event":
            if event.data.type == "response.output_item.added":
                print(f"-- Raw Output: Item done --")
        elif event.type == "agent_updated_stream_event":
            print(f"-- Agent updated --")
            continue
        elif event.type == "run_item_stream_event":
            if event.item.type == "tool_call_item":
                print("-- Run Item: Tool Called --")
            elif event.item.type == "tool_call_output_item":
                print(f"-- Run Item: Tool Output --")
            elif event.item.type == "message_output_item":
[...]

PR #1869, opened 2025-10-07
Fix: Correct streaming order for ReasoningItem and RawResponsesStreamEvent events

ihower:
## Issues

This PR fixes incorrect streaming order for `ReasoningItem` and `RawResponsesStreamEvent`:

1. `RawResponsesStreamEvent` should be emitted ASAP.  

    This behavior was accidentally broken in https://github.com/openai/openai-agents-python/pull/1300, which changed the original correct ordering.
    
2. `ReasoningItem` should be emitted right after its  ```response.output_item.done``` event, rather than being delayed until after `response.completed`.  This issue was reported in https://github.com/openai/openai-agents-python/issues/1767#issuecomment-3319824140

## Solution

1. **Move raw event emission earlier**  

    Restored the correct ordering by moving `RawResponsesStreamEvent` to the top of the handler.

2. **Emit `ReasoningItem` earlier**  
   
    Applied the exact same approach as in PR #1300 for early emission of `ToolCallItem`, also applying it to ensure that `ReasoningItem` is emitted right after its `ResponseOutputItemDoneEvent`.
    
   Code intentionally not refactored for DRYness to keep the diff explicit and review-friendly. Refactoring can be done later after test code are merged.
        
3. **Add comprehensive unit tests**
    
   Extended `FakeModel` to fully simulate detailed streaming sequence and added tests verifying the exact event order, ensuring future changes won’t reintroduce regressions. 
   
   For reference, here is a real example script that calls the OpenAI API, along with the corresponding 27 test events:  https://gist.github.com/ihower/b6d9fe9b72b02c5131e210e2202b31b7

## Related

- Fixes: https://github.com/openai/openai-agents-python/issues/1767#issuecomment-3319824140
- Fixes: https://github.com/openai/openai-agents-python/issues/583
- Related: https://github.com/openai/openai-agents-python/pull/1300

### DIFF
--- a/tests/fake_model.py
+++ b/tests/fake_model.py
@@ -3,7 +3,33 @@ from __future__ import annotations
 from collections.abc import AsyncIterator
 from typing import Any
 
-from openai.types.responses import Response, ResponseCompletedEvent, ResponseUsage
+from openai.types.responses import (
+    Response,
+    ResponseCompletedEvent,
+    ResponseContentPartAddedEvent,
+    ResponseContentPartDoneEvent,
+    ResponseCreatedEvent,
+    ResponseFunctionCallArgumentsDeltaEvent,
+    ResponseFunctionCallArgumentsDoneEvent,
+    ResponseFunctionToolCall,
+    ResponseInProgressEvent,
+    ResponseOutputItemAddedEvent,
+    ResponseOutputItemDoneEvent,
+    ResponseOutputMessage,
+    ResponseOutputText,
+    ResponseReasoningSummaryPartAddedEvent,
+    ResponseReasoningSummaryPartDoneEvent,
+    ResponseReasoningSummaryTextDeltaEvent,
+    ResponseReasoningSummaryTextDoneEvent,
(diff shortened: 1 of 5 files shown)


waziers:
Thanks for fixing @seratch!!

tests/fake_model.py
```diff
@@ -143,10 +169,151 @@ async def stream_response(
...
+                        output_index=output_index,
+                        delta=output_item.arguments,
+                        sequence_number=sequence_number,
+                    )
+                    sequence_number += 1
+
+                    yield ResponseFunctionCallArgumentsDoneEvent(
+                        type="response.function_call_arguments.done",
+                        item_id=output_item.call_id,
+                        output_index=output_index,
```
ihower, line 254: @seratch Insert `name=output_item.name` here to fix the mypy error that appeared after upgrading to openai-python v2.

praveenr1793:
When will this be part of the release? Will this fix the issue that we face with Runner.run() method as well?

Issue #1867, opened 2025-10-07
Dependency constraint blocks upgrade to openai>=2.0


aprendendo-codex:
### Describe the bug
Currently, installing or upgrading `openai-agents` (v0.3.3) forces pip to *downgrade* the `openai` SDK from `2.x` to the latest `1.x` release (`1.109.1`).

### Repro steps

`$ pip install --upgrade openai openai-agents`

### Expected behavior
The latest versions of openai and openai-agents are expected to be mutually compatible.

As new features were launched on devday like chatkit, we require the newer version of the openai sdk in order to have access to them. 

But the agents sdk is currently incompatible with them.

Example of why this is necessary:
```python
# this is not compatible with openai 1.109.1
ck = client.beta.chatkit.sessions.create(workflow={'id':'xxxxxxx'}, user='user')
```

bhanuc:
Facing this issue. Please fix - I was trying out chatkit.

joansegura:
yes! please fix this! we need to be on v2 for sora, and chatkit

thanks guys

MichaelPeterJoyce:
Please fix. Cant try chatkit

PR #1874, opened 2025-10-08
Migrate openai from 1.x to 2.2.0

seratch:
This pull request resolves https://github.com/openai/openai-agents-python/issues/1867 This is a blocker for developers trying ChatKit Python server SDK, so we should release a new version including this bump as early as possible.

I've confirmed migrating to 2.x major version does not bring any incompatibility issues with examples in this repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,7 @@ celerybeat.pid
 .python-version
 .env*
 .venv
+.venv*
 env/
 venv/
 ENV/
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 from collections.abc import Iterable
-from typing import Any, Literal, cast
+from typing import Any, Literal, Union, cast
 
 from openai import Omit, omit
 from openai.types.chat import (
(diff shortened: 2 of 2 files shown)


seratch:
need to resolve mypy errors

.gitignore
```diff
@@ -103,6 +103,7 @@ celerybeat.pid
 .python-version
 .env*
 .venv
+.venv*
```
seratch, line 106: just to do python 3.9 more easily

src/agents/extensions/models/litellm_model.py
```diff
@@ -18,7 +18,7 @@
         "dependency group: `pip install 'openai-agents[litellm]'`."
     ) from _e
 
-from openai import NOT_GIVEN, AsyncStream, NotGiven
+from openai import AsyncStream, NotGiven, omit
```
seratch, line 21: not_given still exists for backward-compatibility but migrating to omit lets this code base fully align with any patterns

seratch:
@rm-openai now this PR is ready. all the things are internal changes; no public interface level breaking changes. that said, given the code changes and incompatibility with openai 1.x series, releasing a new minor version should be necessary

.gitignore
rm-openai: we can actually remove 3.9 support soon, it is EOL this month

PR #1878, opened 2025-10-10
docs: Add Chinese translation for documents

SepineTam:


docs/scripts/translate_docs.py
```diff
@@ -27,6 +27,7 @@
 languages = {
     "ja": "Japanese",
     "ko": "Korean",
+    "zh": "Chinese",
     # Add more languages here, e.g., "fr": "French"
 }
```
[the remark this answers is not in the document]
seratch: Yeah, we'd like to consistently use the lang code "zh" like we already do for TS project: https://openai.github.io/openai-agents-js/zh/

docs/scripts/translate_docs.py
```diff
@@ -114,6 +115,30 @@
         "Build your first agent in minutes.": "단 몇 분 만에 첫 에이전트를 만들 수 있습니다",
         "Let's build": "시작하기",
     },
+    "cn": {
```
seratch: can you use lang code "zh" instead?

docs/cn/agents.md
```diff
@@ -0,0 +1,285 @@
...
+
+agent = Agent(
+    name="Weather Agent",
+    instructions="Retrieve weather details.",
+    tools=[get_weather],
+    tool_use_behavior="stop_on_first_tool"
+)
+```
+
+- `StopAtTools(stop_at_tool_names=[...])`: 指定したツールのいずれかが呼び出された場合に停止し、その出力を最終応答として使用します。
```
seratch: I haven't checked the cause but this part is somehow written in Japanese. Tweak in the prompt in generator script may be necessary.

docs/scripts/translate_docs.py
SepineTam: Right, I will change the `docs/cn` to `docs/zh`

docs/cn/agents.md
SepineTam: Oh, this mistake caused I forgot to commit this change to git, as I translate docs based cp from japanese, I will fix it now.

SepineTam:
[build.log](https://github.com/user-attachments/files/22869754/build.log)

Hi, I am not sure does it is ok, here is the build log, and is there any other thing should be done?

docs/scripts/translate_docs.py
```diff
@@ -114,6 +115,30 @@
...
+        "user": "用户",
+        "parameter": "参数",
+        "processor": "进程",
+        "server": "服务",
+        "web search": "网络检索",
+        "file search": "文件检索",
+        "streaming": "流式传输",
+        "system prompt": "系统提示词",
+        "Python first": "Python 优先",
+        # Add more Japanese mappings here
```
seratch:
```suggestion
        # Add more mappings here
```

PR #1885, opened 2025-10-12
Add type validation to prompt_with_handoff_instructions function

AbdulSamad94:
## Summary

This PR adds runtime type validation to the `prompt_with_handoff_instructions` function to ensure it only accepts string inputs. Previously, the function had a type hint but no runtime validation, so passing non-string types did **not raise any error**, which could lead to unexpected behavior or silent failures.

## Changes

* Added `isinstance(prompt, str)` check in `prompt_with_handoff_instructions`
* Raises `TypeError` with descriptive message when non-string types are provided
* Updated function docstring to document the new validation and exception

## Testing

The validation has been tested with various input types:

* Valid strings: Work as expected
* Invalid types (int, list, None, etc.): Previously caused silent failures, now properly raise `TypeError` with a clear message

### DIFF
--- a/src/agents/extensions/handoff_prompt.py
+++ b/src/agents/extensions/handoff_prompt.py
@@ -15,17 +15,8 @@ RECOMMENDED_PROMPT_PREFIX = (
 def prompt_with_handoff_instructions(prompt: str) -> str:
     """
     Add recommended instructions to the prompt for agents that use handoffs.
-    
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this PR. However, we prefer suggesting to use mypy or similar solution to detect this type of coding error.

PR #1898, opened 2025-10-14
Image, file output types for functions

### LINKED ISSUE #1850: Support for Images in Function Call Outputs
### Please read this first

### Describe the bug
The openai responses api supports entities like images in function call outputs however the current agents sdk does not, instead it casts all outputs to a string [here](https://github.com/openai/openai-agents-python/blob/f3cac173055e180752167d0327a241240a2248a2/src/agents/_run_impl.py#L826). The openai python sdk does support the updated `output` type in the `FunctionCallOutput` object, however it would need to be upgraded a major version to 2+. Once that upgrade is done this looks like a two line change.

### Debug information
- Agents SDK version: (e.g. `v0.0.3`) - latest version or any other version
- Python version (e.g. Python 3.10) - any python version

### Repro steps

Have a tool output a list of messages with an image e.g.
```python
[{
  'type': 'input_image',
  'image_url': 'data:image/png;base64,...',
}]
```

### Expected behavior
Do not cast the above into a str, just let it flow through

rm-openai:
To allow the new output types for image/file, you can now return one of the three new types (or lists of those types, or even a typed dict version). If you use those, we'll convert to the correct tool call output type.

Resolves #1850


rm-openai:
@codex review

rm-openai:
@seratch ready for review!

seratch:
If you want to add data URL example too, something like this works, but it's nice-to-have:
```python
FILEPATH = os.path.join(os.path.dirname(__file__), "media/image_bison.jpg")

def image_to_base64(image_path):
    with open(image_path, "rb") as image_file:
        encoded_string = base64.b64encode(image_file.read()).decode("utf-8")
    return encoded_string

@function_tool
def fetch_random_image() -> ToolOutputImage | ToolOutputImageDict:
    """Fetch a random image."""

    print("Image tool called")
    b64_image = image_to_base64(FILEPATH)
    return {
        "type": "image",
        "image_url": f"data:image/jpeg;base64,{b64_image}",
        "detail": "auto",
    }
```

weathon:
I am a bit confused how this will work. I tried return a list of PIL images but the model seems did not get them

PR #1905, opened 2025-10-16
Fixes realtime example app stability issues

thli-openai:
1. Replaced deprecated ScriptProcessor with audio worklets for both recording and playback
2. Sanitized `history_updated` events to remove audio bytes from bloating the websocket events to the client
3. Swapped default realtime config to use server vad for snappier interruptions


seratch:
Can you fix this mypy error?
```
Run make mypy
uv run mypy . --exclude site
examples/realtime/app/server.py:136: error: Returning Any from function declared to return "dict[str, Any]"  [no-any-return]
Found 1 error in 1 file (checked 342 source files)
make: *** [Makefile:20: mypy] Error 1
```

seratch (approved):
I did a few manual tests and didn't see any issues. This improvement is fantastic 👍

PR #1914, opened 2025-10-17
Fix: Input guardrails now properly block tool execution

### LINKED ISSUE #889: FileSearchTool runs despite InputGuardrailTripwireTriggered

pxkundu:
## Summary

Fixes #889 - Input guardrails were running in parallel with agent execution, allowing tools to execute even when guardrails should block them. This caused unnecessary token consumption and potential security issues.

### Non-streaming path (Runner.run)
- Modified to run input guardrails **sequentially before** `_run_single_turn`
- Guardrails now fully complete and check for tripwires before the agent starts

### Streaming path (Runner.run_streamed)
- Changed `_run_input_guardrails_with_queue` from background task to awaited call
- Added immediate exception raising when tripwire is triggered in `_run_input_guardrails_with_queue`
- Ensures guardrails complete **before** streaming starts

## Impact

**Before this fix:**
- Input guardrails ran in parallel with agent execution (using `asyncio.gather`)
- Tools could execute even when guardrails triggered
- LLM calls were made despite guardrail blocking the input
- Tokens were consumed unnecessarily

**After this fix:**
- Input guardrails run and complete first
- If a guardrail triggers, agent execution never starts
- No LLM calls or tool executions when input is blocked
- Proper security boundary enforcement

## Testing

- ✅ All existing agent runner tests pass (50/50)
- ✅ All guardrail tests pass (8/8)
- ✅ Maintains backward compatibility for passing guardrails
- ✅ Properly blocks execution when guardrails trigger in both streaming and non-streaming modes

## Reviewers

This addresses the concern raised by multiple users in #889 where tools were executing despite guardrail tripwires being triggered. The fix ensures guardrails work as documented: blocking agent execution before any resources are consumed.


seratch:
Thanks for sending this PR. However, getting rid of parallel tool execution by removing `asyncio.gather()` is not the behavioral change we want to have. We had a few attempts by introducing an option to customize the behavior, but we haven't managed to have a viable solution yet.

PR #1919, opened 2025-10-17
v0.4.0

seratch:
We're going to release a new version - 0.4.0


seratch:
>Don't forget to run make sync

good catch; it's done ✅

PR #1921, opened 2025-10-17
Fix streaming trace end before guardrails complete

zbirenbaum:
This addresses an issue that came up while investigating a bug report impacting a wandb/weave user using our agents SDK TracingProcessor

Fixes a race condition where:
1. The streaming response to the query to the agent finishes before the input guardrails finish executing
2. The parent span for both the streaming response and the input guardrails is completed
3. The entire trace gets completed
4. The input guardrails complete and try to trigger on_span_end in the tracing processor for a span that has already completed in a trace that has already completed

It may be the case that this only occurs when there is no next step, since tripwires are awaited there but I'm not sure

Photos to illustrate the issue:
Before Fix:
image

After Fix:
image

### DIFF
--- a/tests/test_stream_input_guardrail_timing.py
+++ b/tests/test_stream_input_guardrail_timing.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import asyncio
 from datetime import datetime
 from typing import Any


src/agents/run.py
```diff
@@ -1138,6 +1138,12 @@ async def _start_streaming(
 
             streamed_result.is_complete = True
         finally:
+            if streamed_result._input_guardrails_task:
```
zbirenbaum, line 1141: I think a better place to put this might be where the output guardrails are awaited (await both the input and output rails with gather or something if they haven't finished by when output ones are running) but this preserves the existing functionality. Would like input if someone with more context has any.

seratch: Currently, input guardrails run in parallel, and there’s no requirement for them to finish before executing subsequent agent code or its output guardrails. We might later add an option to customize this behavior so that subsequent logic waits until the input guardrails complete, but that’s not the case yet.

For now, simply awaiting the input guardrails in the finally clause should be fine.

src/agents/run.py
```diff
@@ -1138,6 +1138,12 @@ async def _start_streaming(
...
+                try:
+                    streamed_result.input_guardrail_results = await streamed_result._input_guardrails_task
+                except Exception:
+                    # Exceptions will be checked in the stream_events loop
+                    output_guardrail_results = []
```
seratch: I don't think this output guardrail related local variable is necessary here. Instead, can you add debug-level logging for debugging what's going on during local development?

seratch (changes requested):
Lint etc. are not successful with your second commit. Can you fix the issue too?

src/agents/run.py
zbirenbaum: Updated it to be just an await with a debug log on exception. I think the tripwire exception might be suppressed so I added a raise.

Regarding the comment above:
That makes sense to me, definitely don't want to wait until the guardrail completes to start getting a response. It should complete at some point before the trace ends and be able to raise an error before completing the span and trace if tripped

seratch: >Updated it to be just an await with a debug log on exception. I think the tripwire exception might be suppressed by this however.

Yeah, after leaving the above comment, I was wondering how things go in the scenario where the tripwire comes up there. The suggested change here should be better than the current behavior in terms of the tracing consistency, but I see the need to take care of the tripwire pattern too.

Would it be possible to ask you to do some manual tests to see how it actually works with this change? The tripwire should show up in the case in any tracing providers. If you have more time to take for this, adding unit tests verifying the pattern would be greatly appreciated.

zbirenbaum: I'm definitely open to taking some time to ensure it works as intended, I am a little limited on time today so may need to followup tomorrow or Monday if that's alright.

I already have a repro file I can use to test the before and after behavior so shouldn't be hard. I'll take a look at the unit tests and see if I can add a few for the relevant scenarios

seratch: Thank you so much!

zbirenbaum: I managed to confirm the branch in its current state addresses the inconsistency in trace completion and the trace structure/code behavior is consistent in the following scenarios:

1. Input guardrail completes first (prior to changes - happy path)
2. Input guardrail completes first (after changes - happy path)
3. Input guardrail completes after main request  (after changes - scenario PR intends to fix)

I also added some unit tests for various delay scenarios and exception behavior:
- On main: `test_parent_span_and_trace_finish_after_slow_input_guardrail` is the only test that fails
- On this branch: `test_parent_span_and_trace_finish_after_slow_input_guardrail` passes

The raise statement in the exception turned out not to be necessary to propagate the tripwire error and could short circuit completion so I took it out and left the debug statement. I think this should be good to go.

src/agents/run.py
```diff
@@ -905,7 +905,10 @@ async def _run_input_guardrails_with_queue(
                 t.cancel()
             raise
 
+        # Store the full set of input guardrail results on the streamed result
+        # and return them so callers awaiting this task can receive the list.
         streamed_result.input_guardrail_results = guardrail_results
+        return guardrail_results
```
seratch: I couldn't understand why you added this line. Is this change necessary this time?

seratch (changes requested):
Overall, this looks great to me. Left one comment. Also, can you resolve the lint error? You can run `make lint`, `make mypy`, and `make tests` before pushing commits.

src/agents/run.py
zbirenbaum: Ran the make commands and all seem to pass now!

Good catch, I was using those lines before but they are not needed anymore, just removed them.

tests/test_stream_input_guardrail_timing.py
```diff
@@ -0,0 +1,228 @@
+import asyncio
```
seratch, line 3: We still support python 3.9, so this CI build fails: https://github.com/openai/openai-agents-python/actions/runs/18673545373/job/53240062121?pr=1921

Can you add this line at the top of this file?
```python
```

seratch (changes requested):

src/agents/run.py
seratch: Thanks!

tests/test_stream_input_guardrail_timing.py
zbirenbaum: Added!

Issue #1938, opened 2025-10-19
Add session-level metadata storage to SQLAlchemySession


Jash271:
### Please read this first

- **Have you read the docs?** Yes - [Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Yes - no existing issues found

### Describe the feature

**Add session-level metadata storage to `SQLAlchemySession`**

#### Problem

`SQLAlchemySession` currently stores conversation messages but has no built-in way to store session-level metadata like:
- User ownership (`owner_id`)
- Session titles and tags
- Configuration (incognito mode, preferences)
- Custom application data

Developers must maintain separate mapping tables or external systems, causing data fragmentation and complexity.

#### Proposed Solution

Add a `session_metadata` table using Entity-Attribute-Value (EAV) pattern for flexible key-value storage.

**API:**
```python
# Set metadata
await session.set_metadata({"owner_id": "user_123", "title": "My Chat", "tags": ["work"]})

# Get metadata (returns None for missing keys)
await session.get_metadata(keys=["owner_id", "title"])
# Returns: {"owner_id": "user_123", "title": "My Chat"}

# Get all metadata
await session.get_metadata()

# Delete metadata
await session.delete_metadata(keys=["tags"])  # Delete specific keys
await session.delete_metadata()  # Delete all

# Find sessions by metadata (cross-session query)
session_ids = await session.find_sessions_by_metadata("owner_id", "user_123")
```

**Schema:**
```sql
CREATE TABLE agent_sessions_metadata (
    session_id VARCHAR REFERENCES agent_sessions(session_id) ON DELETE CASCADE,
    key VARCHAR(255) NOT NULL,
    value TEXT,
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    PRIMARY KEY (session_id, key)
);
CREATE INDEX idx_session_metadata_key_value ON session_metadata(key, value);
```

#### Benefits

- **Unified storage**: Messages + metadata in one system
- **Efficient queries**: Index enables fast "find all sessions for user X"
- **Flexible**: Add any metadata keys without schema changes
- **Cross-database**: PostgreSQL, MySQL, SQLite compatible
- **Backward compatible**: No breaking changes

#### Use Cases

- Multi-tenant apps: Associate sessions with users
- Organization: Titles, tags, descriptions
[...]

PR #1939, opened 2025-10-20
Add session metadata support to SQLAlchemySession

Jash271:
Adds the ability to store and query metadata for sessions using a new agent_sessions_metadata table. This allows applications to associate metadata like owner_id, title, or tags with sessions.

New methods:
- set_metadata() - store metadata key-value pairs
- get_metadata() - retrieve metadata for a session
- delete_metadata() - remove metadata
- find_sessions_by_metadata() - find sessions by metadata values

Also updates clear_session() to clean up metadata when clearing a session.

Feat:  #1938

### DIFF
--- a/src/agents/extensions/models/litellm_model.py
+++ b/src/agents/extensions/models/litellm_model.py
@@ -269,7 +269,7 @@ class LitellmModel(Model):
         )
 
         # Fix for interleaved thinking bug: reorder messages to ensure tool_use comes before tool_result  # noqa: E501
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this code change. It may work but we still hesitate having this change as it could be a specific change for your use case (see https://github.com/openai/openai-agents-python/issues/1938#issuecomment-3420279265 for more details on what I think). So, for now, please inherit the built-in class and add these functionalities on your own sub class.

PR #1945, opened 2025-10-20
fix: MCP tools return error messages instead of crashing on HTTP errors (fixes #879)

### LINKED ISSUE #879: MCP tools turn upstream HTTP errors into AgentsException, crashing the run
### Description

If an MCP tool call receives a non-2xx response from its upstream service, the wrapper raises an `AgentsException`. Because the tool never returns a result object, the agent cannot handle the failure and the entire turn aborts.

---

### Steps to Reproduce

1. Invoke any MCP tool with parameters that make the upstream API reply with an error (e.g., an invalid search query).
2. Observe that the run terminates with an uncaught `AgentsException`.

---

### Expected Result

The tool returns a structured error payload so the agent can decide how to respond.

### Actual Result

An `AgentsException` is raised and the run stops.

---

### Sanitized Traceback

```text
Error invoking MCP tool <tool_name>: GET <service-url>?q=<masked_query>: 422 <service-error>

  File "<site-packages>/agents/mcp/util.py", line 104, in invoke_mcp_tool
    result = await server.call_tool(tool.name, json_data)
  ...
mcp.shared.exceptions.McpError: GET <service-url>: 422 <service-error>
agents.exceptions.AgentsException: Error invoking MCP tool <tool_name>: ...
```

gn00295120:
## Summary

Fixes #879 - MCP tools no longer crash the entire agent run when they encounter HTTP errors. Instead, they return a structured error message that the agent can handle gracefully.

## Problem

When an MCP tool receives a non-2xx response from its upstream service, the wrapper raises an `AgentsException`. Because the tool never returns a result object, the agent cannot handle the failure and the entire turn aborts.

## Solution

Modified `src/agents/mcp/util.py` to return a structured JSON error message instead of raising an exception when a tool call fails.

### Error Message Format

```json
{
  "error": "error description", 
  "tool": "tool_name",
  "type": "tool_error"
}
```

## Changes

- **File modified**: `src/agents/mcp/util.py`
- **Lines changed**: 202-213
- **Behavior change**: `invoke_mcp_tool()` now returns error JSON instead of raising `AgentsException`

## Testing Notes

The existing test `test_mcp_invocation_crash_causes_error` in `tests/mcp/test_mcp_util.py` currently verifies the crashing behavior. This test will need to be updated to verify the new graceful error handling behavior.

### How to Test

1. Set up an MCP server that returns HTTP errors
2. Invoke the tool with parameters that trigger an error
3. Previously: Run terminates with uncaught `AgentsException`
4. Now: Tool returns structured error that agent can process

## Impact Analysis

- **Risk Level**: Low - only affects error handling path
- **Backward Compatibility**: ✅ Error messages are strings (JSON), same type as successful calls
- **User Benefit**: Agents can now handle MCP tool failures gracefully

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -203,7 +203,14 @@ class MCPUtil:
             result = await server.call_tool(tool.name, json_data)
         except Exception as e:
             logger.error(f"Error invoking MCP tool {tool.name}: {e}")
-            raise AgentsException(f"Error invoking MCP tool {tool.name}: {e}") from e
+            # Return error as a structured message instead of raising
+            # This allows the agent to handle the error gracefully
+            error_message = {
+                "error": str(e),
+                "tool": tool.name,
+                "type": "tool_error"
+            }
+            return json.dumps(error_message)
 
         if _debug.DONT_LOG_TOOL_DATA:
             logger.debug(f"MCP tool {tool.name} completed.")


gn00295120:
Closing this PR after further analysis. The proposed fix has a critical issue: converting exceptions to JSON strings creates silent failures and breaks existing tests that expect `AgentsException`.

A proper solution requires changes at the agent loop level to catch and handle `AgentsException` gracefully, rather than converting exceptions to JSON at the tool level.

I'll comment on #879 with the analysis and suggested approach.

PR #1646, opened 2025-09-03
feat: #1614 gpt-realtime migration (Realtime API GA)

seratch:
this is still in progress but will resolve #1614

PR #1952, opened 2025-10-21
fix: Remove trailing commas causing tuple assignment in response cancellation

### LINKED ISSUE #1951: Response cancellation broken due to tuple assignment bug in Realtime API
## Summary

Response cancellation is completely broken in the Realtime API due to trailing commas creating tuples instead of booleans.

## Bug Location

`src/agents/realtime/openai_realtime.py` lines 434 and 619

## Description

The trailing commas in the `automatic_response_cancellation_enabled` assignment create single-element tuples instead of booleans:

```python
# Current (buggy) code:
automatic_response_cancellation_enabled = (
    session
    and session.audio is not None
    and session.audio.input is not None
    and session.audio.input.turn_detection is not None
    and session.audio.input.turn_detection.interrupt_response is True,  # ← Trailing comma
)
```

## Impact

The condition `if not automatic_response_cancellation_enabled:` always evaluates to `False` because:
- With trailing comma: `type = tuple`, value = `(False,)`, `not (False,) = False`
- Non-empty tuples are always truthy in Python

**Result**: `_cancel_response()` is **never called**, breaking response cancellation when `interrupt_response=False`.

## Reproduction

```python
# Demonstrates the bug:
enabled = (False,)  # tuple
print(not enabled)  # False - _cancel_response() NOT called!

# Expected behavior:
enabled = False  # bool  
print(not enabled)  # True - _cancel_response() IS called
```

## Environment

- Introduced in: PR #1646 (commit 91683482, 2025-09-12)
- Affects: All Realtime API users with `interrupt_response=False`

## Fix

Remove the trailing commas on lines 434 and 619.

I'll submit a PR with the fix.

gn00295120:
Fixes #1951

## Problem
Trailing commas on lines 434 and 619 in `src/agents/realtime/openai_realtime.py` create single-element tuples instead of boolean values, breaking the response cancellation logic when `interrupt_response=False`.

## Root Cause
```python
# Buggy (creates tuple):
automatic_response_cancellation_enabled = (..., )  # type: tuple

# Python semantics:
(False,) → tuple, not (False,) → False (truthy)
```

This causes `_cancel_response()` to never be called when automatic response cancellation is disabled.

## Changes
- Removed trailing commas on lines 434 and 619
- Changes convert tuple assignments to proper boolean expressions

## Verification
- Created test demonstrating bug behavior
- Verified bug exists in code (2 occurrences)
- Applied fix and confirmed trailing commas removed
- All 23 realtime tests pass

## Impact
- Introduced in: PR #1646 (commit 9168348, 2025-09-12)
- Severity: CRITICAL - response cancellation completely broken for `interrupt_response=False`
- Test coverage gap: No existing tests for this scenario

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -431,7 +431,7 @@ class OpenAIRealtimeWebSocketModel(RealtimeModel):
             and session.audio is not None
             and session.audio.input is not None
             and session.audio.input.turn_detection is not None
-            and session.audio.input.turn_detection.interrupt_response is True,
+            and session.audio.input.turn_detection.interrupt_response is True
         )
         if not automatic_response_cancellation_enabled:
             await self._cancel_response()
@@ -616,7 +616,7 @@ class OpenAIRealtimeWebSocketModel(RealtimeModel):
                     and session.audio is not None
                     and session.audio.input is not None
                     and session.audio.input.turn_detection is not None
-                    and session.audio.input.turn_detection.interrupt_response is True,
+                    and session.audio.input.turn_detection.interrupt_response is True
                 )
                 if not automatic_response_cancellation_enabled:
                     await self._cancel_response()


seratch (approved):
Aha, this is a great catch

PR #1953, opened 2025-10-21
fix: Replace per-call Lock creation with shared lock in SQLiteSession

gn00295120:
## Problem

The lock pattern in `SQLiteSession` creates a new `threading.Lock()` object for each operation when using file-based databases:

```python
with self._lock if self._is_memory_db else threading.Lock():
    # ...
```

This provides no thread safety for concurrent access to file databases, as each call uses a different lock instance.

## Solution

Use the shared `self._lock` for both memory and file databases:

```python
with self._lock:
    # ...
```

## Changes

- `get_items()` (line 123)
- `add_items()` (line 177)
- `pop_item()` (line 218)
- `clear_session()` (line 255)

## Testing

- Existing test `test_sqlite_session_concurrent_access` passes
- Manual verification confirms proper lock reuse
- No breaking changes to API or behavior

## Impact

- Fixes potential race conditions in file-based SQLite sessions
- Memory-based sessions behavior unchanged
- Adds proper synchronization for concurrent operations

### DIFF
--- a/src/agents/memory/sqlite_session.py
+++ b/src/agents/memory/sqlite_session.py
@@ -120,7 +120,7 @@ class SQLiteSession(SessionABC):
 
         def _get_items_sync():
             conn = self._get_connection()
-            with self._lock if self._is_memory_db else threading.Lock():
+            with self._lock:
                 if limit is None:
                     # Fetch all items in chronological order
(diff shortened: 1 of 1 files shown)


gn00295120:
Closing to recreate from clean branch

PR #1957, opened 2025-10-21
fix: add closed state check to prevent operations on closed session

gn00295120:
## Summary
Added `UserError` guards to `send_message()`, `send_audio()`, and `interrupt()` methods to prevent operations on closed `RealtimeSession` instances, replacing implicit AssertionError with explicit state validation following codebase conventions.

### Current Behavior (Without Fix)

When a session is closed but background tasks continue running:

```python
# Session closed by remote disconnect or error
# But background task (e.g., buffer flush) still running

await session.send_audio(buffer_data)
  → session._model.send_event(...)
    → model._send_raw_message(...)
      → assert self._websocket is not None, "Not connected"  # ← AssertionError
```

**Issues with AssertionError:**
1. **Unpredictable**: Assert statements can be disabled with `python -O` (optimization mode)
2. **Unclear semantics**: AssertionError indicates programming logic errors, not runtime state errors
3. **Poor error handling**: Developers don't expect to catch AssertionError for normal operation failures
4. **Hidden failure mode**: No clear indication that the issue is a closed session vs other problems

### Real-World Scenario (Twilio Example)

From `examples/realtime/twilio/twilio_handler.py`:

```python
class TwilioHandler:
    async def start(self):
        # Three concurrent tasks:
        self._realtime_session_task = asyncio.create_task(
            self._realtime_session_loop()    # Monitors OpenAI events
        )
        self._buffer_flush_task = asyncio.create_task(
            self._buffer_flush_loop()        # Flushes audio every 50ms
        )
```

**Timeline of the bug:**
1. OpenAI WebSocket disconnects (network error, server restart, etc.)
2. `_realtime_session_loop()` exits, session calls `_cleanup()`, sets `_closed = True`
3. `_buffer_flush_task` still running (not automatically cancelled)
4. 50ms later: `await session.send_audio(...)` called on closed session
5. **AssertionError**: "Not connected" (confusing and hard to handle)

### Who Controls Session Closure?

**SDK cannot control all closure scenarios:**
- ❌ Remote disconnect (OpenAI server initiated)
- ❌ Network errors (timeout, connection loss)
- ❌ Internal exceptions (guardrail failures, model errors)
- ❌ Resource exhaustion (token limits, etc.)

Background tasks have different lifecycles than the session and may outlive it.

### Why UserError?

This codebase has a well-defined exception hierarchy:

```python
# src/agents/exceptions.py

class UserError(AgentsException):
    """Exception raised when the user makes an error using the SDK."""
```

**Existing usage in realtime module:**
- `UserError("API key is required but was not provided.")` - Configuration error
- `UserError("on_handoff must take two arguments: context and input")` - API usage error  
- `UserError(f"Tool {tool.name} is unsupported. Must be a function tool.")` - Unsupported operation

**Our case**: Calling methods on a closed session = SDK usage error → **UserError** is appropriate

### Changes Made

Added explicit state checks using `UserError`:

```python
async def send_message(self, message: RealtimeUserInput) -> None:
    if self._closed:
        raise UserError("Cannot send message: session is closed")
    await self._model.send_event(...)

async def send_audio(self, audio: bytes, *, commit: bool = False) -> None:
    if self._closed:
        raise UserError("Cannot send audio: session is closed")
    await self._model.send_event(...)

async def interrupt(self) -> None:
    if self._closed:
        raise UserError("Cannot interrupt: session is closed")
    await self._model.send_event(...)
```

### Benefits

1. **Follows codebase conventions**: Uses `UserError` like other SDK usage errors
2. **Correct exception type**: Semantically appropriate for API misuse
3. **Always enforced**: Cannot be disabled with optimization flags
4. **Clear error message**: "session is closed" immediately indicates the problem
5. **Better error handling**: Developers can catch and handle appropriately

**Before (unpredictable):**
```python
try:
    await session.send_audio(buffer_data)
except AssertionError:  # Weird, indicates logic error
    # What went wrong? Hard to tell
    print(f"Unexpected error: {e}")
```

**After (clear):**
```python
try:
    await session.send_audio(buffer_data)
except UserError as e:
    if "session is closed" in str(e):
        # Clear: session closed, can handle gracefully
        logger.info("Session closed, stopping buffer flush")
        break
    raise
```

6. **Fail-fast at correct layer**: Check at session level (where `_closed` flag lives) instead of propagating to model layer

## Performance Impact

- **Cost per check**: ~5 nanoseconds (boolean read + branch)
- **Relative to network I/O**: < 0.01% overhead
- **Memory**: No allocations, cache-friendly
- **Negligible in practice**: Even at 1000 calls/sec = 5 microseconds/sec total

## Testing

- All 53 tests in `tests/realtime/test_session.py` passed
- No behavioral changes to existing functionality
- Only adds defensive state validation

## Migration Notes

No breaking changes. This only adds earlier failure detection with clearer error messages. Code that previously raised AssertionError will now raise UserError with more context, following the codebase's existing exception conventions.

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -187,14 +187,20 @@ class RealtimeSession(RealtimeModelListener):
 
     async def send_message(self, message: RealtimeUserInput) -> None:
         """Send a message to the model."""
+        if self._closed:
+            raise UserError("Cannot send message: session is closed")
         await self._model.send_event(RealtimeModelSendUserInput(user_input=message))
 
     async def send_audio(self, audio: bytes, *, commit: bool = False) -> None:
         """Send a raw audio chunk to the model."""
+        if self._closed:
+            raise UserError("Cannot send audio: session is closed")
         await self._model.send_event(RealtimeModelSendAudio(audio=audio, commit=commit))
 
     async def interrupt(self) -> None:
         """Interrupt the model."""
+        if self._closed:
+            raise UserError("Cannot interrupt: session is closed")
         await self._model.send_event(RealtimeModelSendInterrupt())
 
     async def update_agent(self, agent: RealtimeAgent) -> None:


seratch:
Thanks for sending this. However, I don't see a meaningful difference with this change. The current error also tells what's wrong.

gn00295120:
Thanks for the feedback! You're absolutely right that the current error message is clear. If you prefer to keep the existing behavior, that's totally fine.

The main reasons I suggested this change:

1. **Exception semantics**: `AssertionError` typically indicates programming logic errors (should never happen), while `UserError` indicates API usage errors (can happen in normal use). Operations on a closed session fall into the latter category.

2. **Consistency with codebase**: Other API misuse errors in the realtime module use `UserError` (e.g., missing API key, unsupported tools, invalid handoff results).

3. **Reliability**: Assertions can be disabled with `python -O`, though this is rare in production.

4. **Earlier failure**: Fails at the session layer (where `_closed` flag lives) rather than propagating to the model layer.

However, these are minor improvements. If you feel the current behavior is sufficient, I'm happy to close this PR.

seratch:
Let me close this one

PR #1958, opened 2025-10-21
Improved docstrings for model settings parameters to provide clearer explanations

EngineerAbdullahIqbal:
Improved docstrings for model settings parameters to provide clearer explanations of their behavior and valid ranges.

## Changes
- Enhanced `temperature` docstring with range and behavior details
- Enhanced `top_p` docstring with nucleus sampling explanation
- Enhanced `frequency_penalty` docstring with use case
- Enhanced `presence_penalty` docstring with distinction from frequency_penalty
- Enhanced `tool_choice` docstring with available options

## Why this need
The current docstrings merely repeat the parameter names without explaining what they do or how they affect model behavior, which can confuse developers.


seratch:
Thanks for suggesting these changes. However, we don't need them this time.

PR #1960, opened 2025-10-21
Add engine accessor to SQLAlchemySession for closing it when it's created from a URL

gn00295120:
### Interface Inconsistency
Currently, `SQLAlchemySession` lacks a `close()` method, making it inconsistent with other session implementations:

- ✅ `RedisSession` has `async def close()` 
- ✅ `SQLiteSession` has `def close()`
- ❌ `SQLAlchemySession` has **no close() method**

This inconsistency means users cannot properly clean up resources when using `from_url()`.

### Resource Leak in Long-Running Applications

When using `from_url()`, each call creates a new `AsyncEngine` with its own connection pool:

```python
# This pattern leaks engines in long-running apps
@app.post("/chat/{user_id}")
async def chat(user_id: str):
    session = SQLAlchemySession.from_url(
        session_id=f"user_{user_id}",
        url=DATABASE_URL
    )
    # ... use session ...
    # ❌ No way to clean up the engine!
```

### SQLAlchemy Official Requirement

From [SQLAlchemy Discussion #7531](https://github.com/sqlalchemy/sqlalchemy/discussions/7531):

> "invoke the AsyncEngine.dispose() method using await when the AsyncEngine object will go out of scope... Failing to explicitly dispose may result in warnings like 'RuntimeError: Event loop is closed' within garbage collection."

---

### 1. **API Consistency**
Users expect all session types to have consistent lifecycle management. The `close()` method should be available across all session implementations.

### 2. **Follows Best Practices**
SQLAlchemy's async engine requires explicit disposal. Without it, applications may see:
- Connection pool exhaustion
- "Event loop is closed" warnings during GC
- Resource leaks in long-running services

### 3. **Prevents Resource Leaks**
In production web services handling multiple users, each `from_url()` call accumulates engines that are never cleaned up.

### 4. **Backward Compatible**
This is a non-breaking change:
- Adds optional `close()` method
- Existing code continues to work
- Only affects users who explicitly call `close()`

---

## Solution

Following the **same pattern as `RedisSession`**:

### 1. Track Engine Ownership
```python
self._owns_engine = False  # Default: external engine
```

### 2. Mark Ownership in `from_url()`
```python
session = cls(session_id, engine=engine, **kwargs)
session._owns_engine = True  # SDK created this engine
return session
```

### 3. Add `close()` Method
```python
async def close(self) -> None:
    """Dispose engine only if we own it."""
    if self._owns_engine:
        await self._engine.dispose()
```

---

### Example 1: Using `from_url()` (SDK owns engine)
```python
session = SQLAlchemySession.from_url(
    "user_123",
    url="postgresql+asyncpg://localhost/db"
)
try:
finally:
    await session.close()  # ✅ Engine disposed
```

### Example 2: External engine (user owns engine)
```python
engine = create_async_engine("postgresql+asyncpg://localhost/db")
session = SQLAlchemySession("user_123", engine=engine)

await session.close()  # ✅ Engine NOT disposed (user manages it)

# User responsible for cleanup
await engine.dispose()
```

---

## Testing

Tested manually with both scenarios:
- ✅ `from_url()` disposes engine on `close()`
- ✅ External engine is NOT disposed on `close()`
- ✅ Multiple sessions clean up properly

---

## Changes

- Add `_owns_engine` flag to track engine ownership
- `from_url()` sets `_owns_engine=True`
- Direct engine injection keeps `_owns_engine=False` 
- New `async def close()` method
- Matches `RedisSession` resource management pattern

---

## Impact

**Scope**: Long-running applications using `from_url()` in multi-user scenarios

**Severity**: Medium - Not critical for short-lived scripts, but important for production web services

**Compatibility**: Fully backward compatible - existing code unaffected


gn00295120:
Thanks for the review\! I've updated the docstring to include the `Returns: None` section as suggested, ensuring consistency with Python documentation conventions.

seratch:
I think simply adding the accessor to _engine would be simpler and more flexible

gn00295120:
Thank you for the feedback! I've updated the PR to add the `engine` property accessor while keeping the existing `close()` method.

Now users have both options:
- **Convenience**: `await session.close()` - automatically handles engine disposal based on ownership
- **Flexibility**: `session.engine` - direct access for advanced use cases (checking pool status, manual disposal, etc.)

This approach:
- Provides the simplicity and flexibility you suggested via the `engine` accessor
- Maintains consistency with `RedisSession` and `SQLiteSession` which both have `close()` methods
- Keeps the ownership tracking to prevent accidentally disposing shared engines

Does this align with what you had in mind? Happy to adjust further if you prefer a different approach.

src/agents/extensions/memory/sqlalchemy_session.py
```diff
@@ -319,3 +322,29 @@ async def clear_session(self) -> None:
...
+        This property provides direct access to the engine for advanced use cases,
+        such as checking connection pool status, configuring engine settings,
+        or manually disposing the engine when needed.
+
+        Returns:
+            AsyncEngine: The SQLAlchemy async engine instance.
+        """
+        return self._engine
+
+    async def close(self) -> None:
```
seratch: I'd like to remove this general method because the sessions store interface does not declare this method

seratch (changes requested):

seratch:
No, I don't think we should add close() method

gn00295120:
Thank you for the clarification! I completely understand now.

I've updated the PR to follow your suggestion:
- ✅ **Added** `engine` property accessor for direct access to AsyncEngine
- ❌ **Removed** `close()` method (not part of SessionABC interface)
- ❌ **Removed** ownership tracking (`_owns_engine`)

Now the implementation is simple and clean - just a property accessor. Users are responsible for managing the engine lifecycle themselves using `await session.engine.dispose()` when needed.

This aligns with the SessionABC interface and keeps the implementation minimal.

src/agents/extensions/memory/sqlalchemy_session.py
```diff
@@ -319,3 +319,16 @@ async def clear_session(self) -> None:
                 await sess.execute(
                     delete(self._sessions).where(self._sessions.c.session_id == self.session_id)
                 )
+
+    @property
+    def engine(self) -> AsyncEngine:
```
seratch, line 324: Can you add unit tests verifying the behavior of this method?

seratch (changes requested):

gn00295120:
I've added comprehensive unit tests to verify the `engine` property behavior:

## Test Coverage

1. **test_engine_property_from_url()** 
   - Verifies engine property returns AsyncEngine instance
   - Tests that users can access pool status
   - Confirms manual disposal works: `await session.engine.dispose()`

2. **test_engine_property_from_external_engine()**
   - Verifies engine property returns the same external engine instance
   - Ensures users retain control when injecting their own engine

3. **test_engine_property_is_read_only()**
   - Confirms property is immutable (raises AttributeError on assignment)
   - Ensures engine reference cannot be accidentally modified

## ✅ Local Test Results

All tests pass successfully:
```
tests/extensions/memory/test_sqlalchemy_session.py::test_engine_property_from_url PASSED
tests/extensions/memory/test_sqlalchemy_session.py::test_engine_property_from_external_engine PASSED
tests/extensions/memory/test_sqlalchemy_session.py::test_engine_property_is_read_only PASSED

============================== 3 passed in 0.14s ===============================
```

The tests follow the existing test patterns in the file and use the same in-memory SQLite setup.

gn00295120:
Fixed CI failures:

1. **mypy error** - Added `# type: ignore[misc]` to the intentional read-only property assignment test
2. **lint errors** - Shortened docstrings to stay within 100 character limit

All tests still pass locally:
```
tests/extensions/memory/test_sqlalchemy_session.py::test_engine_property_from_url PASSED
tests/extensions/memory/test_sqlalchemy_session.py::test_engine_property_from_external_engine PASSED  
tests/extensions/memory/test_sqlalchemy_session.py::test_engine_property_is_read_only PASSED

============================== 3 passed in 0.14s ===============================
```

CI should pass now.

ProCityHub:
(0.0)= Artifact

(0.6) = Artifact

(0.1) = Energy

(1.6)= 7 = Cube with (6 walls) + (1 Energy) =7=1
walls 1 to 6, all have a lattice structure, like wall 1, 1a, 1b, 1c, etc, all walls structured the same,  whereas, 0.0 was the neutral point (•)  
XYZ Axis  =(mind, body, spirit), (father, son, holy spirit), (trinity)
0.0=(.)

cube, 6 walls, can and may curve +/- for light defraction, , 6 dimensions, wall (like a lattice),  governed by xyz axis.  the box has 8 corners, amplification, latency and capacitors as a metaphor of data.  solve

Observation... .. import numpy as np
import matplotlib.pyplot as plt

def solve_lattice_system(size=15, iterations=100, tolerance=1e-5):
    """
    Solves the mathematical core by iteratively relaxing a 3D grid.

    Args:
        size (int): The dimension of the cubic grid (N x N x N).
        iterations (int): The maximum number of iterations to perform.
        tolerance (float): The stability threshold to stop early.

    Returns:
        numpy.ndarray: The final, stable 3D grid.
    """
    # 1. Initialize the 3D grid G with zeros.
    grid = np.zeros((size, size, size))

    # 2. Apply input as initial/boundary conditions.
    # We set a "hot" plane (value = 1.0) at one face to create a gradient.
    grid[0, :, :] = 1.0

    print("Solving... Iterating until the system reaches equilibrium.")
    # 3. Repeat until stable.
    for i in range(iterations):
        # Keep a copy of the old grid to calculate changes.
        grid_old = grid.copy()

        # For each internal point P in G...
        # Update P.value based on the average of its 6 neighbors.
        # This is a vectorized operation for efficiency.
        grid[1:-1, 1:-1, 1:-1] = (
            grid_old[0:-2, 1:-1, 1:-1] + grid_old[2:, 1:-1, 1:-1] +
            grid_old[1:-1, 0:-2, 1:-1] + grid_old[1:-1, 2:, 1:-1] +
            grid_old[1:-1, 1:-1, 0:-2] + grid_old[1:-1, 1:-1, 2:]
        ) / 6.0

        # Check for convergence by measuring the maximum change.
        change = np.max(np.abs(grid - grid_old))
        if change < tolerance:
            print(f"System stabilized after {i+1} iterations.")
            break
    else: # This else belongs to the for loop
        print(f"Reached maximum iterations ({iterations}) without full convergence.")


    # 4. Output = read(final_state).
    return grid

# --- Execute the Solver ---
final_grid = solve_lattice_system()

# --- Visualize the Result ---
# We'll show a 2D slice from the middle of the 3D cube.
plt.figure(figsize=(8, 6))
plt.imshow(final_grid[:, :, final_grid.shape[2] // 2], cmap='inferno')
plt.colorbar(label="Potential / Value")
plt.title("Equilibrium State (Central 2D Slice)")
plt.xlabel("X-axis")
plt.ylabel("Y-axis")
plt.show()

The two way mirror structure in the lattice is amplified by eight corners, positive and negative.. latency is evident based on the data transfer by light and molecules.. 0.0 is the neutral position. All governed by x y z axes, if you go afford to the next room, it's the exact same room, except it is wall 1a, 0.0.  Etc..

ProCityHub:
(0.0)= Artifact

(0.6) = Artifact

(0.1) = Energy

(1.6)= 7 = Cube with (6 walls) + (1 Energy) =7=1
walls 1 to 6, all have a lattice structure, like wall 1, 1a, 1b, 1c, etc, all walls structured the same, whereas, 0.0 was the neutral point (•)
XYZ Axis =(mind, body, spirit), (father, son, holy spirit), (trinity)
0.0=(.)

cube, 6 walls, can and may curve +/- for light defraction, , 6 dimensions, wall (like a lattice), governed by xyz axis. the box has 8 corners, amplification, latency and capacitors as a metaphor of data. solve

Observation... .. import numpy as np
import matplotlib.pyplot as plt

def solve_lattice_system(size=15, iterations=100, tolerance=1e-5):
"""
Solves the mathematical core by iteratively relaxing a 3D grid.

Args:
    size (int): The dimension of the cubic grid (N x N x N).
    iterations (int): The maximum number of iterations to perform.
    tolerance (float): The stability threshold to stop early.

Returns:
    numpy.ndarray: The final, stable 3D grid.
"""
# 1. Initialize the 3D grid G with zeros.
grid = np.zeros((size, size, size))

# 2. Apply input as initial/boundary conditions.
# We set a "hot" plane (value = 1.0) at one face to create a gradient.
grid[0, :, :] = 1.0

print("Solving... Iterating until the system reaches equilibrium.")
# 3. Repeat until stable.
for i in range(iterations):
    # Keep a copy of the old grid to calculate changes.
    grid_old = grid.copy()

    # For each internal point P in G...
    # Update P.value based on the average of its 6 neighbors.
    # This is a vectorized operation for efficiency.
    grid[1:-1, 1:-1, 1:-1] = (
        grid_old[0:-2, 1:-1, 1:-1] + grid_old[2:, 1:-1, 1:-1] +
        grid_old[1:-1, 0:-2, 1:-1] + grid_old[1:-1, 2:, 1:-1] +
        grid_old[1:-1, 1:-1, 0:-2] + grid_old[1:-1, 1:-1, 2:]
    ) / 6.0

    # Check for convergence by measuring the maximum change.
    change = np.max(np.abs(grid - grid_old))
    if change < tolerance:
        print(f"System stabilized after {i+1} iterations.")
        break
else: # This else belongs to the for loop
    print(f"Reached maximum iterations ({iterations}) without full convergence.")


# 4. Output = read(final_state).
return grid
--- Execute the Solver ---
final_grid = solve_lattice_system()

--- Visualize the Result ---
We'll show a 2D slice from the middle of the 3D cube.
plt.figure(figsize=(8, 6))
plt.imshow(final_grid[:, :, final_grid.shape[2] // 2], cmap='inferno')
plt.colorbar(label="Potential / Value")
plt.title("Equilibrium State (Central 2D Slice)")
plt.xlabel("X-axis")
plt.ylabel("Y-axis")
plt.show()

The two way mirror structure in the lattice is amplified by eight corners, positive and negative.. latency is evident based on the data transfer by light and molecules.. 0.0 is the neutral position. All governed by x y z axes, if you go forward to the next room, it's the exact same room, except it is wall 1a, 0.0. Etc.

PR #1206, opened 2025-07-22
Fix #980 Chat Completions: fails with function name for tool_choice parameter w/ streaming enabled

seratch:
This pull request resolves #980

PR #1971, opened 2025-10-21
Fix #1846 Litellm: fails with function name for tool_choice parameter w/ streaming enabled

### LINKED ISSUE #1846: litellm + streaming + tool_choice = 'mytool' fails with pydantic error
### Describe the bug
When using a custom llm endpoint (via litellm), in a streaming context, the sdk throws an error if you try to specify `tool_choice = 'my_required_tool_name'`, when this should indeed be allowed.

### Debug information
- Agents SDK version: 0.3.2 and 0.3.3 confirmed to fail
- Python 3.13

### Repro steps

With streaming enabled, and using litellm targeting i.e. `openrouter/qwen/qwen3-30b-a3b-instruct-2507`, and specifying a name for `tool_choice` on the request:

```python
    agent = Agent(
        name="RespondingAgent",
        tools=tools,
        handoffs=handoffs,  # type: ignore
        instructions=_create_prompt(cur_date),
        model=get_model(),
        model_settings=ModelSettings(
            temperature=temperature,
            top_p=top_p,
            parallel_tool_calls=config.PARALLEL_TOOL_CALLS,
            
            # BUG: uncommenting the below line will throw a pydantic model error, saying the value must be `auto`, `required`, or `none`.
            # tool_choice="reason"
        ),
    )
```

The pydantic model error is thrown from `openai-agents-python / src / agents / extensions / models / litellm_model.py L377`.

Looking there, it validates like so:

```python
tool_choice=cast(Literal["auto", "required", "none"], tool_choice)
```

Note: I pulled the repo and tried naively changing that line to

```python
tool_choice=cast(str, tool_choice)
```

But that resulted in yet another validation error elsewhere in the sdk.

### Expected behavior
It should allow specifying an arbitrary string, since that represents the function name that the caller expects to be invoked.

Note: it seems to work when streaming is disabled. I'm not 100% sure if it's streaming that matters, though, since I'm using it in a slightly different scenario where the error doesn't appear, and the only difference I can think of is that it doesn't use streaming.

ihower:
Resolves https://github.com/openai/openai-agents-python/issues/1846

This PR provides a simpler fix than https://github.com/openai/openai-agents-python/pull/1929.

The root cause is the same as in https://github.com/openai/openai-agents-python/pull/1206 
When returning the final `Response`, it should be converted to the Responses API format.
Applying the exactly same fix from PR #1206 (by @seratch) resolves this issue.

No new tests were added, since `Converter.convert_tool_choice` is already tested.

### DIFF
--- a/src/agents/extensions/models/litellm_model.py
+++ b/src/agents/extensions/models/litellm_model.py
@@ -44,6 +44,7 @@ from ...models.chatcmpl_helpers import HEADERS, HEADERS_OVERRIDE
 from ...models.chatcmpl_stream_handler import ChatCmplStreamHandler
 from ...models.fake_id import FAKE_RESPONSES_ID
 from ...models.interface import Model, ModelTracing
+from ...models.openai_responses import Converter as OpenAIResponsesConverter
 from ...tool import Tool
 from ...tracing import generation_span
 from ...tracing.span_data import GenerationSpanData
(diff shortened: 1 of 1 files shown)


seratch (approved):
Thank you!

PR #1927, opened 2025-10-18
fix: raise UserError instead of TypeError for Agent validation (fixes #1443)

gn00295120:
## Summary

Fixes #1443

This PR changes Agent validation to raise `UserError` instead of `TypeError` when users pass invalid types to Agent parameters, providing better error messages and aligning with the SDK's documented behavior.

### Step 1: Review the Issue Report

From issue #1443, when users pass invalid types to Agent, they get `TypeError` instead of `UserError`:

```python
from agents import Agent

def predict_weather():
    return "Sunny"

# Passing a string instead of a list of tools
agent = Agent(
    name="TestAgent",
    instructions="Test agent",
    tools="predict_weather"  # ❌ should be a list
)
```

**Expected**: `UserError: 'tools' must be a list of Tool objects, not a string.`
**Actual**: `TypeError: Agent tools must be a list, got str`

### Step 2: Check the SDK Documentation

According to the SDK documentation:
> `UserError` should be raised when the user passes an invalid value to a parameter (e.g., wrong type).

But the code was raising `TypeError` instead!

### Step 3: Create Reproduction Test

Create `test_reproduce_issue_1443.py`:

```python
from agents import Agent
from agents.exceptions import UserError

print("[Test 1] Invalid tools type (string instead of list)")
try:
    agent = Agent(
        name="TestAgent",
        instructions="Test agent",
        tools="predict_weather",  # Wrong type!
    )
    print("❌ No error raised!")
except UserError as e:
    print(f"✅ Got UserError: {e}")
except TypeError as e:
    print(f"❌ Got TypeError instead of UserError: {e}")

print("\n[Test 2] Invalid name type (list instead of string)")
try:
    agent = Agent(
        name=["Test", "Agent"],  # Wrong type!
        instructions="Test agent",
    )
    print("❌ No error raised!")
except UserError as e:
    print(f"✅ Got UserError: {e}")
except TypeError as e:
    print(f"❌ Got TypeError instead of UserError: {e}")

print("\n[Test 3] Invalid handoffs type (string instead of list)")
try:
    agent = Agent(
        name="TestAgent",
        instructions="Test agent",
        handoffs="some_agent",  # Wrong type!
    )
    print("❌ No error raised!")
except UserError as e:
    print(f"✅ Got UserError: {e}")
except TypeError as e:
    print(f"❌ Got TypeError instead of UserError: {e}")
```

Run it:

```bash
python test_reproduce_issue_1443.py
```

**Output (Before Fix)**:
```
[...]

Issue #1897, opened 2025-10-14
AttributeError: 'Handoff' object has no attribute 'name'


tnduyh5:
Hi team -> can you help me verify this bugs. I'm using this sdk have see the error Handoff' object has no attribute 'name' -> Then I use AI to help debug, please review below
# Handoff.name Bug Fix

## Issue

The application was experiencing an `AttributeError` when using agent handoffs:

```
AttributeError: 'Handoff' object has no attribute 'name'
```

This error occurred at runtime when the OpenAI Agents SDK tried to list tool names in `agents/run.py` line 903:

```python
tool_names = [t.name for t in all_tools]
```

## Root Cause

The issue is in the `openai-agents` SDK v0.3.3:

1. The `Handoff` class has a `tool_name` attribute
2. The `FunctionTool` class has a `name` attribute
3. The SDK's `run.py` tries to access `.name` on all tools (including handoffs)
4. Since `Handoff` doesn't have `.name`, it raises an `AttributeError`

### Debug information
- Agents SDK version: v0.3.3
- Python version:3.11

tnduyh5:
# Please review the document below also:

```

# OpenAI Agents SDK v0.3.3 - Handoff Bugs Fix

## Overview

Fixed **TWO CRITICAL BUGS** in `openai-agents` SDK v0.3.3 that prevented Handoff from working:

1. ❌ **Bug #1**: `AttributeError: 'Handoff' object has no attribute 'name'`
2. ❌ **Bug #2**: `UserError: Hosted tools are not supported with the ChatCompletions API`

Both bugs are in the **upstream SDK**, not our code!

---

## Bug #1: Missing `name` Attribute

### Issue

```
AttributeError: 'Handoff' object has no attribute 'name'
```

Occurred in `agents/run.py` line 903:

```python
tool_names = [t.name for t in all_tools]
```

### Root Cause

Inconsistency in the SDK:

1. `Handoff` class has `tool_name` attribute ✅
2. `FunctionTool` class has `name` attribute ✅
3. SDK's `run.py` assumes ALL tools have `.name` ❌
4. Result: `AttributeError` when accessing `Handoff.name`

### Solution for Bug #1

Add a `name` property to `Handoff` class:

```python
@property
def name(self):
    return self.tool_name

Handoff.name = name
```

---

## Bug #2: Handoff Not Supported in ChatCompletions API

### Issue

```
UserError: Hosted tools are not supported with the ChatCompletions API. 
[...]

PR #1974, opened 2025-10-22
fix: validate tool types in Agent.tools list to prevent runtime AttributeError (fixes #1443)

### LINKED ISSUE #1443: `UserError` not raised when invalid tool type is passed to `tools` parameter
**Description:**
According to the SDK documentation, a `UserError` should be raised when the user passes an invalid value to the `tools` parameter (e.g., wrong type).

However, when I pass a string instead of a list of tools, an `AttributeError` is thrown instead of the documented `UserError`.

This seems to be due to missing type validation before the SDK iterates over the `tools` parameter.

**Steps to Reproduce:**

```python
from openai import OpenAI
from openai.agents import Agent

def predict_weather():
    return "Sunny"

client = OpenAI()

# Passing a string instead of a list of tools
agent = Agent(
    name="TestAgent",
    instructions="Test agent",
    tools="predict_weather"  #  should be a list
)
```

**Expected Behavior:**
Raise:

```plaintext
UserError: 'tools' must be a list of Tool objects, not a string.
```

**Actual Behavior:**

```plaintext
AttributeError: 'str' object has no attribute 'name'
```

**Possible Fix:**
Add an explicit type check in the constructor before iterating over the tools:

```python
if not isinstance(tools, list):
    raise UserError("'tools' must be a list of Tool objects, not a string.")
```

**Note:**
The `UserError` class already exists in the SDK (`class UserError(AgentsException)`) and is designed to handle such invalid usage scenarios. However, in this case, it’s not triggered because there’s no type check before the code tries to access `tool.name`.

**Environment:**

* SDK version: `openai-agents[viz]>=0.2.4`
* Python version: `Python 3.13.2`
* OS: `Windows-11-10.0.26100-SP0`

---

I can also add a **direct reference to the file and line number** in the SDK’s GitHub repo where `UserError` is defined so they don’t have to search for it.
Do you want me to include that as well?

gn00295120:
## Summary

This PR fixes issue #1443 where passing invalid tool types (e.g., raw functions instead of `FunctionTool` objects) in the `tools` list would pass initialization but fail at runtime.

## Problem

Currently, when users mistakenly pass raw functions or other invalid types in the `tools` parameter, the error only appears during runtime:

```python
agent = Agent(
    name="Test",
    tools=[some_function]  # Raw function, not FunctionTool
)

# Agent creation succeeds ✅

result = await Runner.run(agent, "test")
# ❌ Runtime error: AttributeError: 'function' object has no attribute 'name'
#    (in src/agents/run.py line 584)
```

This is confusing because:
1. The error happens far from where the mistake was made
2. The error message doesn't clearly explain what went wrong
3. Users might not understand they need to use `@function_tool` decorator

### Why AttributeError Occurs at Runtime

The root cause is in `src/agents/run.py:584`:

```python
current_span.span_data.tools = [t.name for t in all_tools]
```

This code assumes all items in `all_tools` have a `.name` attribute. However:

- **Raw Python functions** (type `function`) do NOT have a `.name` attribute (they only have `__name__`)
- **Valid Tool objects** (FunctionTool, FileSearchTool, etc.) all have a `.name` attribute

When a raw function reaches this line, Python raises `AttributeError: 'function' object has no attribute 'name'`.

The existing validation in `Agent.__post_init__` only checks if `tools` is a list, not the type of each element:

```python
if not isinstance(self.tools, list):
    raise TypeError("Agent tools must be a list, got ...")
# ❌ Does not validate individual tool types
```

## Solution

Add validation in `Agent.__post_init__` to check each tool in the list:

```python
agent = Agent(
    name="Test",
    tools=[some_function]
)
# ❌ Immediate error at initialization:
# UserError: tools[0] must be a valid Tool object (FunctionTool, FileSearchTool, ...),
# got function. Did you forget to use @function_tool decorator or pass the function
# itself instead of a tool?
```

This provides better developer experience by:
- **Early detection**: Errors caught at initialization, not runtime
- **Clear error messages**: Explicitly states what went wrong
- **Helpful suggestions**: Mentions `@function_tool` decorator when raw functions detected

## Changes

**src/agents/agent.py**:
- Added `UserError` import at module level
- Added validation loop in `__post_init__` to check each tool type using `get_args(Tool)`
- Uses dynamic type extraction to automatically include all Tool types (FunctionTool, FileSearchTool, WebSearchTool, ComputerTool, HostedMCPTool, LocalShellTool, ImageGenerationTool, CodeInterpreterTool, and Handoff)

**tests/test_agent_config.py**:
- Added `test_tools_content_validation_issue_1443` with 3 test cases:
  - Case 1: Raw function in tools list
  - Case 2: String in tools list  
  - Case 3: Mixed valid and invalid tools (verifies correct index is reported)

## Testing

All tests pass:
```bash
pytest tests/test_agent_config.py::TestAgentValidation -v
# 6 passed
```

## Note on PR #1927

This PR builds on the learnings from PR #1927. While #1927 attempted to change `TypeError` to `UserError` for existing validations, this PR adds **new validation** that was previously missing, preventing runtime `AttributeError` by catching invalid tool types early.

Fixes #1443

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -11,6 +11,7 @@ from openai.types.responses.response_prompt_param import ResponsePromptParam
 from typing_extensions import NotRequired, TypeAlias, TypedDict
 
 from .agent_output import AgentOutputSchemaBase
+from .exceptions import UserError
 from .guardrail import InputGuardrail, OutputGuardrail
 from .handoffs import Handoff
 from .items import ItemHelpers
@@ -245,6 +246,23 @@ class Agent(AgentBase, Generic[TContext]):
 
         if not isinstance(self.tools, list):
             raise TypeError(f"Agent tools must be a list, got {type(self.tools).__name__}")
+
+        # Validate each tool is a valid Tool type
+        # Tool is a Union type, so we need to get the valid types from it
+        from typing import get_args
+
+        valid_tool_types = get_args(Tool) + (Handoff,)
+
+        for i, tool in enumerate(self.tools):
+            if not isinstance(tool, valid_tool_types):
+                # Generate a friendly list of valid types for the error message
+                type_names = ", ".join(t.__name__ for t in valid_tool_types)
+                raise UserError(
+                    f"tools[{i}] must be a valid Tool object ({type_names}), "
+                    f"got {type(tool).__name__}. "
+                    f"Did you forget to use @function_tool decorator or pass the function itself "
+                    f"instead of a tool?"
+                )
 
         if not isinstance(self.mcp_servers, list):
             raise TypeError(
--- a/tests/test_agent_config.py
+++ b/tests/test_agent_config.py
@@ -214,6 +214,49 @@ class TestAgentValidation:
         with pytest.raises(TypeError, match="Agent handoffs must be a list"):
             Agent(name="test", handoffs="not_a_list")  # type: ignore
 
+    def test_tools_content_validation_issue_1443(self):
+        """Test that tools list validates each element is a valid Tool object (Issue #1443)
+
+        This test addresses the issue where passing invalid tool types (e.g., raw functions)
+        in a list would pass __post_init__ validation but fail later at runtime with:
+        AttributeError: 'function' object has no attribute 'name'
+
+        The fix validates each tool in the list during initialization.
+        """
+        from agents.exceptions import UserError
+
+        def raw_function():
+            """A raw function, not decorated with @function_tool"""
+            return "test"
+
+        # Case 1: Raw function in tools list should raise UserError at init
+        with pytest.raises(
+            UserError,
+            match=r"tools\[0\] must be a valid Tool object.*got function.*@function_tool",
+        ):
+            Agent(name="test", tools=[raw_function])  # type: ignore
+
+        # Case 2: String in tools list should raise UserError at init
+        with pytest.raises(
+            UserError,
+            match=r"tools\[0\] must be a valid Tool object.*got str",
+        ):
+            Agent(name="test", tools=["invalid_string"])  # type: ignore
+
+        # Case 3: Mixed valid and invalid tools - should catch invalid at correct index
+        from agents import function_tool
+
+        @function_tool
+        def valid_tool() -> str:
+            """A valid tool"""
+            return "ok"
+
+        with pytest.raises(
+            UserError,
+            match=r"tools\[1\] must be a valid Tool object.*got str",
+        ):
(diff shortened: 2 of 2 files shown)


gn00295120:
Thanks for the review! I've updated the code to use `get_args(Tool)` to dynamically extract valid tool types from the `Tool` Union type. This makes the code more maintainable - any future Tool subclasses added to the `Tool` Union will be automatically validated without code changes.

Changes:
- Removed explicit import of all concrete tool types
- Use `typing.get_args(Tool)` to dynamically get valid types
- Error message now automatically includes all valid types from the Union

This approach is cleaner and future-proof! ✨

gn00295120:
Thanks for catching that! However, I've already updated the code to use `get_args(Tool)` which dynamically extracts all types from the `Tool` Union. This means `HostedMCPTool` is automatically included:

```python
from typing import get_args
valid_tool_types = get_args(Tool) + (Handoff,)
```

When I run this, it returns:
```
FunctionTool, FileSearchTool, WebSearchTool, ComputerTool, 
HostedMCPTool, LocalShellTool, ImageGenerationTool, 
CodeInterpreterTool, Handoff
```

So existing agents using `HostedMCPTool` will continue to work! ✅

The benefit of this approach is that any future tool types added to the `Tool` Union will be automatically validated without code changes.

src/agents/agent.py
```diff
@@ -246,6 +247,23 @@ def __post_init__(self):
         if not isinstance(self.tools, list):
             raise TypeError(f"Agent tools must be a list, got {type(self.tools).__name__}")
 
+        # Validate each tool is a valid Tool type
+        # Tool is a Union type, so we need to get the valid types from it
+        from typing import get_args
+
+        valid_tool_types = get_args(Tool) + (Handoff,)
```
ihower, line 254: I believe Handoff should not be included in valid_tool_types.

See the related discussion here:
https://github.com/openai/openai-agents-python/issues/1897#issuecomment-3404527981

src/agents/agent.py
```diff
@@ -246,6 +247,23 @@ def __post_init__(self):
         if not isinstance(self.tools, list):
             raise TypeError(f"Agent tools must be a list, got {type(self.tools).__name__}")
 
+        # Validate each tool is a valid Tool type
+        # Tool is a Union type, so we need to get the valid types from it
+        from typing import get_args
```
ihower, line 252: I think this import should be placed at the top of the file with the other imports.

seratch:
Thanks for sending this. The latest code already validates if the "tools" is a list object. So, the original issue could be closed now.

PR #1976, opened 2025-10-22
Fix: Properly await cancelled guardrail tasks during cleanup

gn00295120:
## Summary

Fixes a resource leak bug where guardrail tasks were cancelled but not properly awaited during cleanup, which could lead to unhandled task exception warnings and memory leaks.

## Problem

The `_cleanup_guardrail_tasks()` method in `RealtimeSession` (`src/agents/realtime/session.py:749`) was only calling `task.cancel()` on pending guardrail tasks but not awaiting them. According to Python's asyncio best practices, cancelled tasks should be awaited to properly collect exceptions.

## Evidence

1. **Test code shows correct pattern**: `tests/realtime/test_session.py:1199` already demonstrates the proper cleanup:
   ```python
   await asyncio.gather(*session._guardrail_tasks, return_exceptions=True)
   ```

2. **Similar pattern used elsewhere**: The WebSocket task cleanup in `openai_realtime.py:519-523` follows the same pattern of cancel + await.

## Solution

1. Changed `_cleanup_guardrail_tasks()` from sync to async method
2. Added `await asyncio.gather(*self._guardrail_tasks, return_exceptions=True)` after cancelling to properly collect exceptions
3. Updated the caller (`_cleanup()`) to await the cleanup method

## Testing

**New tests** in `tests/realtime/test_guardrail_cleanup.py`:
- ✅ Test 1: Verify cancelled tasks are properly awaited
- ✅ Test 2: Verify exceptions during cleanup are handled  
- ✅ Test 3: Verify multiple concurrent tasks are cleaned up

**Existing tests**: All 8 guardrail tests in `test_session.py` continue to pass.

## Technical Details

**Before:**
```python
def _cleanup_guardrail_tasks(self) -> None:
    for task in self._guardrail_tasks:
        if not task.done():
            task.cancel()
    self._guardrail_tasks.clear()
```

**After:**
```python
async def _cleanup_guardrail_tasks(self) -> None:
    for task in self._guardrail_tasks:
        if not task.done():
            task.cancel()
    
    # Wait for all tasks to complete and collect any exceptions
    if self._guardrail_tasks:
        await asyncio.gather(*self._guardrail_tasks, return_exceptions=True)
    
    self._guardrail_tasks.clear()
```

## Impact

- No breaking changes
- Prevents resource leaks
- Eliminates unhandled task exception warnings
- Follows asyncio best practices

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "openai-agents"
-version = "0.4.1"
+version = "0.4.2"
 description = "OpenAI Agents SDK"
 readme = "README.md"
 requires-python = ">=3.9"
--- a/src/agents/extensions/models/litellm_model.py
+++ b/src/agents/extensions/models/litellm_model.py
@@ -110,18 +110,26 @@ class LitellmModel(Model):
                 prompt=prompt,
             )
 
-            assert isinstance(response.choices[0], litellm.types.utils.Choices)
+            message: litellm.types.utils.Message | None = None
+            first_choice: litellm.types.utils.Choices | None = None
+            if response.choices and len(response.choices) > 0:
+                choice = response.choices[0]
+                if isinstance(choice, litellm.types.utils.Choices):
+                    first_choice = choice
+                    message = first_choice.message
 
             if _debug.DONT_LOG_MODEL_DATA:
                 logger.debug("Received model response")
             else:
-                logger.debug(
-                    f"""LLM resp:\n{
-                        json.dumps(
-                            response.choices[0].message.model_dump(), indent=2, ensure_ascii=False
-                        )
-                    }\n"""
-                )
+                if message is not None:
+                    logger.debug(
+                        f"""LLM resp:\n{
+                            json.dumps(message.model_dump(), indent=2, ensure_ascii=False)
+                        }\n"""
+                    )
+                else:
+                    finish_reason = first_choice.finish_reason if first_choice else "-"
+                    logger.debug(f"LLM resp had no message. finish_reason: {finish_reason}")
 
             if hasattr(response, "usage"):
                 response_usage = response.usage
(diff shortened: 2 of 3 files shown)


src/agents/realtime/session.py
```diff
@@ -746,16 +746,32 @@ def _on_guardrail_task_done(self, task: asyncio.Task[Any]) -> None:
...
+        and prevents warnings about unhandled task exceptions.
+        """
+        # Collect real asyncio.Task objects that need to be awaited
+        real_tasks = []
+
         for task in self._guardrail_tasks:
             if not task.done():
                 task.cancel()
+            # Only await real asyncio.Task objects (not mocks in tests)
+            if isinstance(task, asyncio.Task):
```
seratch: I understand this reason but it looks like a code smell for maintainers. If this could be simpler and naturally work with tests, it'd be proferrable.

src/agents/realtime/session.py
```diff
@@ -746,16 +746,32 @@ def _on_guardrail_task_done(self, task: asyncio.Task[Any]) -> None:
...
+        if real_tasks:
+            await asyncio.gather(*real_tasks, return_exceptions=True)
+
         self._guardrail_tasks.clear()
 
     async def _cleanup(self) -> None:
         """Clean up all resources and mark session as closed."""
         # Cancel and cleanup guardrail tasks
-        self._cleanup_guardrail_tasks()
+        await self._cleanup_guardrail_tasks()
```
seratch, line 828: Enabling await for this operation is indeed a good improvement

src/agents/realtime/session.py
gn00295120: I've removed the `isinstance` check and updated the test to use real `asyncio.Task`
  objects instead. Much cleaner now - all tests passing

src/agents/realtime/session.py
gn00295120: Thanks! 👍

gn00295120:
Resolved the merge conflict by extending the same fix to `_cleanup_tool_call_tasks()`.

Since PR #1984 introduced `_cleanup_tool_call_tasks()` with the same bug pattern (cancel without await), I applied the async cleanup approach to both methods for consistency:

```python
async def _cleanup(self) -> None:
    await self._cleanup_guardrail_tasks()
    await self._cleanup_tool_call_tasks()  # Also made async
```

Added 3 tests for tool call cleanup. All 6 tests pass.

PR #1977, opened 2025-10-22
Fix: Properly await cancelled tasks in voice STT cleanup

gn00295120:
## Summary

Fixes resource leak in OpenAISTTTranscriptionSession where tasks are cancelled but not properly awaited during cleanup.

## Problem

The `_cleanup_tasks()` method in `OpenAISTTTranscriptionSession` was only calling `task.cancel()` on pending tasks (listener, process_events, stream_audio, connection) but not awaiting them. This could lead to:
1. Unhandled task exception warnings
2. Potential resource leaks (websocket connections, file descriptors)
3. Improper cleanup of background tasks

## Evidence

- Similar to recently fixed guardrail tasks cleanup (PR #1976)
- Similar to fixed websocket task cleanup (PR #1955)
- asyncio best practices require awaiting cancelled tasks
- Bug documented in internal analysis: `.claude/bug-analysis/03-resource-leaks.md`

## Solution

1. Made `_cleanup_tasks()` async
2. Collect all real asyncio.Task objects that need to be awaited
3. Added `await asyncio.gather()` with `return_exceptions=True` to properly collect exceptions from cancelled tasks
4. Updated `close()` method to await `_cleanup_tasks()`

## Testing

- All existing voice/STT tests pass (17 passed)
- Uses `isinstance` check to support mock objects in tests
- Follows the same pattern as PR #1976 and PR #1955

## Files Changed

- `src/agents/voice/models/openai_stt.py`: Update task cleanup to properly await cancelled tasks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "openai-agents"
-version = "0.4.1"
+version = "0.4.2"
 description = "OpenAI Agents SDK"
 readme = "README.md"
 requires-python = ">=3.9"
--- a/src/agents/extensions/models/litellm_model.py
+++ b/src/agents/extensions/models/litellm_model.py
@@ -110,18 +110,26 @@ class LitellmModel(Model):
                 prompt=prompt,
             )
 
-            assert isinstance(response.choices[0], litellm.types.utils.Choices)
+            message: litellm.types.utils.Message | None = None
+            first_choice: litellm.types.utils.Choices | None = None
+            if response.choices and len(response.choices) > 0:
+                choice = response.choices[0]
+                if isinstance(choice, litellm.types.utils.Choices):
+                    first_choice = choice
+                    message = first_choice.message
 
             if _debug.DONT_LOG_MODEL_DATA:
                 logger.debug("Received model response")
             else:
-                logger.debug(
-                    f"""LLM resp:\n{
-                        json.dumps(
-                            response.choices[0].message.model_dump(), indent=2, ensure_ascii=False
-                        )
-                    }\n"""
-                )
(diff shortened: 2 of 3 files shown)


gn00295120:
@codex review

src/agents/voice/models/openai_stt.py
```diff
@@ -321,18 +321,37 @@ def _check_errors(self) -> None:
...
+        """Cancel all pending tasks and wait for them to complete.
+
+        This ensures that any exceptions raised by the tasks are properly handled
+        and prevents warnings about unhandled task exceptions.
+        """
+        tasks = []
+
         if self._listener_task and not self._listener_task.done():
             self._listener_task.cancel()
+            if isinstance(self._listener_task, asyncio.Task):
```
seratch: Perhaps, this is due to the same reason with https://github.com/openai/openai-agents-python/pull/1976 but we don't want to have this check, which is unnecessary for main code

seratch (commented):
any chance to add unit tests for this code path?

gn00295120:
> any chance to add unit tests for this code path?

✅ Added 3 unit tests:
test_cleanup_tasks_cancels_and_awaits_all_tasks – tests canceling and awaiting all pending tasks.
test_cleanup_tasks_handles_exceptions – tests handling exceptions with return_exceptions=True.
test_close_calls_cleanup_tasks – tests that close() correctly calls cleanup_tasks().

These tests cover _cleanup_tasks() and verify that:
Tasks are properly canceled and awaited.
No exceptions are raised (thanks to return_exceptions).
The close() method correctly invokes the cleanup logic.

PR #1996, opened 2025-10-27
Nest handoff history by default

jhills20:
## Summary
- add a `nest_handoff_history` flag to `RunConfig` and call a new helper that condenses the prior transcript into a developer-role summary when handing off
- update the default handoff path, docs, and helper library so that the developer summary is produced automatically unless a custom filter overrides it
- expand the handoff-focused tests to cover the new behavior (including helper unit tests) and update existing expectations

## Testing
- `uv run pytest tests/test_extension_filters.py`
- `uv run pytest tests/test_agent_runner.py -k handoff`
- `uv run pytest tests/test_agent_runner_streamed.py -k handoff`

------
https://chatgpt.com/codex/tasks/task_i_68ff73bda0f4832496f3d1fa9103905f


src/agents/_run_impl.py
```diff
@@ -51,6 +51,7 @@
     ToolOutputGuardrailTripwireTriggered,
     UserError,
 )
+from .extensions.handoff_filters import nest_handoff_history
```
seratch: I understand other existing filters are in the extension file. but should this be under extensions? If we enable this by default (RunConfig#nest_handoff_history is true by default), the filter should be part of the main module. Also, importing extensions here looks like a code smell to me (meaning it may not be ideal in terms of pluggable module system design)

src/agents/_run_impl.py
```diff
@@ -1008,6 +1009,9 @@ async def execute_handoffs(
                     new_items=tuple(new_step_items),
                     run_context=context_wrapper,
                 )
+
+            if input_filter and handoff_input_data is not None:
+                logger.debug("Filtering inputs for handoff")
```
seratch: this debug logging may not be so useful. we may want to remove it or include more info

rm-openai (dismissed):
All of kaz's feedback seems right. In addition, I'd say two major things:
1. This setting should be on the `Handoff` object
2. Since you're updating the default, this is a breaking change, so please update the changelog.

jhills20 (commented):
moved to Handoff, added handoff history .py file for non extension parts of handoff filter, added changelog

docs/release.md
```diff
@@ -25,6 +25,8 @@ This version doesn’t introduce any visible breaking changes, but it includes n
 
 - Added support for `RealtimeRunner` to handle [SIP protocol connections](https://platform.openai.com/docs/guides/realtime-sip)
 - Significantly revised the internal logic of `Runner#run_sync` for Python 3.14 compatibility
+- By default handoff history is now packaged into a single assistant message instead of exposing the raw user/assistant turns, giving downstream agents a concise, predictable recap
+- The existing single-message handoff transcript now by default starts with "For context, here is the conversation so far between the user and the previous agent:" before the `<CONVERSATION HISTORY>` block, so downstream agents get a clearly labeled recap
```
rm-openai: we;ve release 0.5.0 so you'll need to move this to a 0.6.0 section

src/agents/_run_impl.py
```diff
@@ -1008,6 +1014,17 @@ async def execute_handoffs(
...
+            if input_filter and handoff_input_data is not None:
+                filter_name = getattr(input_filter, "__qualname__", repr(input_filter))
+                from_agent = getattr(agent, "name", agent.__class__.__name__)
+                to_agent = getattr(new_agent, "name", new_agent.__class__.__name__)
+                logger.debug(
+                    "Filtering handoff inputs with %s for %s -> %s",
+                    filter_name,
+                    from_agent,
+                    to_agent,
+                )
```
rm-openai, line 1241: weird syntax - also can you use f-strings and get rid of getattr?

src/agents/extensions/handoff_filters.py
```diff
@@ -13,6 +17,12 @@
 
 """Contains common handoff input filters, for convenience. """
 
+__all__ = [
+    "remove_all_tools",
+    "nest_handoff_history",
+    "default_handoff_history_mapper",
+]
```
rm-openai, line 24: can remove this, not super useful

docs/handoffs.md
```diff
@@ -82,6 +82,8 @@ handoff_obj = handoff(
...
...nest_handoff_history`][agents.run.RunConfig.nest_handoff_history]). The summary appears inside a `<CONVERSATION HISTORY>` block that keeps appending new turns when multiple handoffs happen during the same run. You can provide your own mapping function via [`RunConfig.handoff_history_mapper`][agents.run.RunConfig.handoff_history_mapper] to replace the generated message without writing a full `input_filter`. That default only applies when neither the handoff nor the run supplies an explicit `input_filter`, so existing code that already customizes the payload (including the examples in this repository) keeps its current behavior without changes. You can override the nesting behaviour for a single handoff by passing `nest_handoff_history=True` or `False` to [`handoff(...)`][agents.handoffs.handoff], which sets [`Handoff.nest_handoff_history`][agents.handoffs.Handoff.nest_handoff_history]. If you just need to change the wrapper text for the generated summary, call [`set_conversation_history_wrappers`][agents.handoffs.set_conversation_history_wrappers] (and optionally [`reset_conversation_history_wrappers`][agents.handoffs.reset_conversation_history_wrappers]) before running your agents.
```
seratch, line 85: to make merging the code changes easier, can you make a separate pull request for document changes? with this, we can merge code first, and then merge document changes after thre release including the changes.

seratch (dismissed):

rm-openai (approved):

alexmojaki:
Are you aware that this is duplicating everything by default? For example this code:

```python
import asyncio

@function_tool
def random_number() -> int:
    return 4


agent2 = Agent(name='agent2', instructions='Return double the number')
agent1 = Agent(name='agent1', tools=[random_number], handoffs=[agent2])


asyncio.run(Runner.run(agent1, input='Generate a random number then, hand off to agent2.'))
```

sends all of this:

```json
[
  {
    "role": "assistant",
    "content": "For context, here is the conversation so far between the user and the previous agent:\n<CONVERSATION HISTORY>\n1. user: Generate a random number then, hand off to agent2.\n2. function_call: {\"arguments\": \"{}\", \"call_id\": \"call_Wsu7AtBXA186HIw3qOLST3KJ\", \"name\": \"random_number\", \"id\": \"fc_0001f0f3cb0a689000691db920bd54819d805de7a682ee3d08\", \"status\": \"completed\"}\n3. function_call_output: {\"call_id\": \"call_Wsu7AtBXA186HIw3qOLST3KJ\", \"output\": \"4\"}\n4. function_call: {\"arguments\": \"{}\", \"call_id\": \"call_9xPh03L6GqirBstlCgFE9Oqn\", \"name\": \"transfer_to_agent2\", \"id\": \"fc_0001f0f3cb0a689000691db922b564819dbf86d6da3014bb36\", \"status\": \"completed\"}\n5. function_call_output: {\"call_id\": \"call_9xPh03L6GqirBstlCgFE9Oqn\", \"output\": \"{\\\"assistant\\\": \\\"agent2\\\"}\"}\n</CONVERSATION HISTORY>"
  },
  {
    "arguments": "{}",
    "call_id": "call_Wsu7AtBXA186HIw3qOLST3KJ",
    "name": "random_number",
    "type": "function_call",
    "id": "fc_0001f0f3cb0a689000691db920bd54819d805de7a682ee3d08",
    "status": "completed"
  },
  {
    "call_id": "call_Wsu7AtBXA186HIw3qOLST3KJ",
    "output": "4",
    "type": "function_call_output"
  },
  {
    "arguments": "{}",
    "call_id": "call_9xPh03L6GqirBstlCgFE9Oqn",
    "name": "transfer_to_agent2",
    "type": "function_call",
    "id": "fc_0001f0f3cb0a689000691db922b564819dbf86d6da3014bb36",
    "status": "completed"
  },
  {
    "call_id": "call_9xPh03L6GqirBstlCgFE9Oqn",
    "output": "{\"assistant\": \"agent2\"}",
    "type": "function_call_output"
  }
]
```

Is it better in other cases?

giuliohome:
I've tested this in our environment and it looks good to me.
I noticed the context size has increased — assuming that's intentional, it should have a positive impact overall.

KthProg:
Ever since this change, I keep getting this error:
`No tool output found for function call call_NAwYOrJ0yPljsKVLSjKC9W8v.`.

When I check traces, the tool calls are still there but with empty responses. Any idea what gives? Does the `remove_all_tools` input filter need to be updated?

For reference this is my setup:

```
      file_search_tool = FileSearchTool(
        max_num_results=5,
        vector_store_ids=[vector_store_id],
      )

  agent_1= Agent(name=f"Agent 1", model="gpt-4o-mini",   instructions=prompt_with_handoff_instructions(agent_1_instructions), tools=[
        file_search_tool
      ], handoff_description="Agent 1...")

      agent_2= Agent(
        name=f"Agent 2", model="gpt-4o-mini", instructions=prompt_with_handoff_instructions(agent_2_instructions), tools=[
          file_search_tool
        ],
        mcp_servers=[mcp_server],
        handoff_description="Agent 2...")

      agent_1_handoff= handoff(
        agent=agent_1,
        on_handoff=on_handoff_agent_1,
        nest_handoff_history=False,
        input_filter=remove_all_tools
      )

      agent_2_handoff = handoff(
        agent=agent_2,
        on_handoff=agent_2,
        nest_handoff_history=False,
        input_filter=remove_all_tools
      )

      triage_agent = Agent(
        name=f"Triage Agent",
        model="gpt-4o-mini",
        instructions=prompt_with_handoff_instructions(triage_instructions),
        handoffs=[agent_1_handoff, agent_2_handoff],
        handoff_description="An agent that triages customer inquiries and hands off to the appropriate agent",
      )

  result = Runner.run_streamed(
    agent,
    user_message,
    previous_response_id=previous_response_id,
  )
```

The error occurs only when a previous response id is provided.

Sourav-Nandy-ai:
Hi James,

This PR is a good example of a "threshold evals still pass, behavior distribution shifts" change. Making handoff-history nesting the default is the right call architecturally, but it changes what every downstream agent sees as context on every handoff.

I actually ran this end-to-end against the real SDK on both sides. The agents.RunConfig() on version 0.5.1 has no nest_handoff_history field at all, whereas on 0.6.0 it defaults to True. I then built a real triage-to-specialist handoff with a manufactured 3-turn prior conversation and called Runner.run() for real (using one gpt-4o-mini call per side).

Pre-fix (0.5.1): The result.to_input_list() keeps the 3 prior turns as 3 separate flat items.

Post-fix (0.6.0): Those same 3 turns collapse into a single role: "assistant" item with a <CONVERSATION HISTORY> summary block. They are genuinely nested rather than just reordered.  
openai.github.io

The final output was identical on both sides, so this is a clean, isolated internal-representation change exactly as advertised.

As a minor aside, your merge commit calls it a "developer-role" summary, but the shipped code and what I actually observed both use role: "assistant". This is probably just stale commit-message wording.

That is the exact class of change agent-eval (agent_regress) is built to catch at scale. Instead of asserting individual outputs pass or fail, it runs an agent N times before and after a change and performs a real statistical comparison for a REGRESSED, STABLE, or IMPROVED verdict on the distribution. This would be useful here since two people (alexmojaki and KthProg) already hit real downstream regressions from this default flip before reaching for nest_handoff_history=False.

My OpenAI Agents SDK integration used to be flatly broken. It called a nonexistent agent.run() and crashed on the first call. That is fixed now, and I verified it against the real installed SDK for this check. I am genuinely curious whether a distributional before-and-after comparison on a handoff-nesting toggle specifically would have caught anything useful pre-merge. Let me know what you think if you get a chance to try it.

You can find the repository here: [https://github.com/RudrenduPaul/agent-eval](https://www.google.com/search?q=https://github.com/RudrenduPaul/agent-eval)

Best, 
Sourav

PR #2001, opened 2025-10-28
Create devcontainer.json

honest-droid:


seratch:
Thanks for sending this. When I see the necessity, we may add something similar. But for now, let us close this.

PR #2013, opened 2025-10-30
fix: use ResponseOutputRefusal for refusal content part added event

mshsheikh:
In the refusal handling branch, the `ResponseContentPartAddedEvent` was incorrectly using `ResponseOutputText` as the `part` payload.
This violates the OpenAI Agents protocol, which requires refusal content to be represented by `ResponseOutputRefusal` so that downstream consumers can access the `.refusal` field.

This change updates the `part` field to use `ResponseOutputRefusal(refusal="", type="refusal")`, ensuring type consistency with both the accumulated state and the corresponding `ResponseContentPartDoneEvent`.

Additionally, the `output_index` calculation is slightly improved for clarity by explicitly using `(1 if state.reasoning_content_index_and_output else 0)` instead of a boolean expression, though the primary fix is the payload type correction.


seratch:
@codex review

Issue #2007, opened 2025-10-29
TypeError object of type 'Omit' has no len()


browniebroke:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug

After upgrading from openai-agents from 0.1.0 to 0.4.1, we got this error in production. We didn't observe this error on our staging or development environments and even after the fact we're unable to reproduce.

This happens when calling `await Runner.run(agent, ...)` and the stack trace points to this line:

https://github.com/openai/openai-agents-python/blob/1466ddbc6ca5b2c8809a28fe7d4a3d09851fd3e8/src/agents/models/openai_chatcompletions.py#L293

So my best guess is that it's somewhat related to:

- https://github.com/openai/openai-agents-python/pull/1874

As the `NotGiven` was changed to `Omit` but several arguments were changed, so it's a bit tricky to track which one causes an issue.

### Debug information
- Agents SDK version: 0.4.1
- OpenAI version: 2.6.0
- Python version 3.13.9 

### Repro steps

I can't reproduce outside of our production environment yet, but I'm still investigating. We also use Datadog for observability (ddtrace 3.16.1), and it's typically not installed in the dev environment, so I'm suspecting that it might patch something in a weird way causing that.

EDIT: We were using sentry-sdk==2.42.0 when we saw this error.

Has anyone else encountered this error? 

### Expected behavior

No error when calling the runner.

josiahcoad:
As a short term fix

adding `default_integrations=False` to `sentry_sdk.init` stopped the error for me

browniebroke:
> As a short term fix
> 
> adding `default_integrations=False` to `sentry_sdk.init` stopped the error for me

Thanks for sharing, we also use Sentry and that seemed to be the culprit. The problme occured with sentry-sdk==2.42.0 and openai-agents==0.4.1. The issue in Sentry: 

- https://github.com/getsentry/sentry-python/issues/5057

When I upgraded to sentry-sdk==2.42.1, the error is now gone.

Going to close this issue as resolved now. Thanks

PR #1995, opened 2025-10-27
feat: #1994 Add a list of per-request usage data to Usage 

habema:
Resolves #1994

**Changes:**
- Added `RequestUsage` class to store per-request details
- Added `request_usage_entries` field to `Usage` class  
- Modified `Usage.add()` to automatically create per-request entries for single requests

Sample `Usage` object after changes:
```py
Usage(
    requests=2,
    input_tokens=158,
    input_tokens_details=InputTokensDetails(cached_tokens=0),
    output_tokens=98,
    output_tokens_details=OutputTokensDetails(reasoning_tokens=0),
    total_tokens=256,
    request_usage_entries=[
        RequestUsage(
            input_tokens=66,
            output_tokens=14,
            total_tokens=80,
            input_tokens_details=InputTokensDetails(cached_tokens=0),
            output_tokens_details=OutputTokensDetails(reasoning_tokens=0),
        ),
        RequestUsage(
            input_tokens=92,
            output_tokens=84,
            total_tokens=176,
            input_tokens_details=InputTokensDetails(cached_tokens=0),
            output_tokens_details=OutputTokensDetails(reasoning_tokens=0),
        ),
    ],
)

```

PR #1986, opened 2025-10-23
feat: add run_in_parallel parameter to input guardrails

rvv-karma:
### Summary

Adds a `run_in_parallel` boolean parameter to `InputGuardrail` that allows developers to choose between parallel execution (default, better latency) and blocking execution (better cost efficiency and prevents race conditions).

**Motivation:**

Currently, input guardrails run in parallel with the agent's execution. While this provides low latency, it can lead to:
- Unnecessary token consumption when a guardrail triggers after the agent has already started processing
- Tools executing before guardrails can halt execution (race condition)
- Security concerns when validation needs to complete before any LLM interaction

**Changes:**

- Added `run_in_parallel: bool = True` field to `InputGuardrail` dataclass (maintains backward compatibility)
- Updated `@input_guardrail` decorator to accept `run_in_parallel` parameter
- Implemented conditional execution logic in `AgentRunner`:
  - **Non-streaming mode**: Sequential guardrails run first (blocking), then parallel guardrails run concurrently with agent via `asyncio.gather`
  - **Streaming mode**: Sequential guardrails run first, then parallel guardrails run as background task
- Updated documentation in `docs/guardrails.md` with "Execution modes" section explaining use cases
- Updated example in `examples/agent_patterns/input_guardrails.py`

**Execution Modes:**

1. **Parallel execution** (`run_in_parallel=True`, default): Guardrail runs concurrently with the agent for minimal latency
2. **Blocking execution** (`run_in_parallel=False`): Guardrail runs and completes before the agent starts, preventing any token consumption or tool execution if the guardrail triggers

**Benefits:**

- ✅ **Backward compatible**: Default is `True`, maintaining current behavior
- ✅ **Solves race conditions**: Completely prevents tool execution when blocking guardrails trigger (fixes #889, #991)
- ✅ **Cost efficient**: Prevents unnecessary agent token usage when guardrails trigger
- ✅ **Flexible**: Developers can choose the right tradeoff for their use case
- ✅ **Explicit**: Clear intent in code about execution order

### Test plan

**Tests Added:**
- `test_input_guardrail_run_in_parallel_default` - Verifies default is `True`
- `test_input_guardrail_run_in_parallel_false` - Verifies can be set to `False`
- `test_input_guardrail_decorator_with_run_in_parallel` - Tests decorator parameter
- `test_input_guardrail_decorator_with_name_and_run_in_parallel` - Tests both `name` and `run_in_parallel` parameters

**Tests Run:**
[...]

PR #1937, opened 2025-10-19
Add Dapr session storage option

filintod:
Implement DaprSession for distributed session management and update makefile for test commands.

Dapr makes it easy to integrate with many different DB as shown in https://docs.dapr.io/reference/components-reference/supported-state-stores

- Added DaprSession class for managing sessions with Dapr state stores, supporting various backends like Redis and PostgreSQL.
- Updated Makefile to include separate targets for unit tests, integration tests, and all tests.

PR #2015, opened 2025-10-30
v0.5.0

seratch:
- add python 3.14 support (changed Runner#run_sync internals a lot for it)
- realtime agent runner's SIP support

Also:
- we should have the following ones in this release too
  - #1995
  - #1937
  - #2039

- optional:
  - #2028
  - #2007
  - #1986 
  - #1996 
  - #2014

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "openai-agents"
-version = "0.4.2"
+version = "0.5.0"
 description = "OpenAI Agents SDK"
 readme = "README.md"
 requires-python = ">=3.9"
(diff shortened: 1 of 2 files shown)


rm-openai (approved):
what was the breaking change in this version?

seratch:
@rm-openai There is actually no hard breaking change but the following could potentially change things for some apps:
- Runner#run_sync internal changes for python 3.14 compatibility
- #2014's Agent and other objects reference changes (I don't think this could affect user side code though)

So, I've bumped the minor for safety. I will update the change section at https://openai.github.io/openai-agents-python/release/ in a separate PR.

Issue #1895, opened 2025-10-14
Streaming runs can't be canceled gracefully


habema:
### Problem
When canceling a streaming run with `result.cancel()`, it stops immediately and discards the current turn. This means:

- The incomplete turn isn't saved to the session
- Usage tracking is cut off mid-turn
- Can't resume the session cleanly

This is painful if the user interrupts a long task - all progress from current turn is lost, and its usage cant be tracked.

### Current behavior
```python
result = Runner.run_streamed(agent, "Task", session=session)
async for event in result.stream_events():
    if timeout:
        result.cancel()  # Stops immediately, turn is discarded
```

Would be nice to have a way to let the current turn complete before stopping.

idzikovski:
I have run into another painful side-effect of not canceling gracefully which is unpredictable/incorrect resource management and release.  In my case I pass an httpx client to the sdk to perform the calls to my LLM and I noticed that after some use and cancelled requests, the httpx client  exhausts every available connection in the pool leading to every request failing with a PoolTimeout.

I would propose that `after_turn` is the default value of `result.cancel()`. If needed I can try to create a minimal reproduction code of the issue I mentioned.

PR #2029, opened 2025-11-02
Faster cancellation for `after_turn` mode

habema:
Related to #1895

Adds two cancellation checkpoints to make `result.cancel('after_turn')` more responsive:

1. After LLM streaming completes, before tool execution — if there are no tools/handoffs, cancel immediately instead of waiting for the tool execution path.
2. After tool execution completes, before next step — ensures tools finish before canceling, then saves session and exits cleanly.

Removed the old checkpoints (after handoff and after turn completion) because:
- The new post-tool-execution checkpoint covers both handoff and final output cases, since it checks before the next step decision.

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1085,24 +1085,6 @@ class AgentRunner:
                     if server_conversation_tracker is not None:
                         server_conversation_tracker.track_server_items(turn_result.model_response)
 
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sending this. I quickly checked the code changes, but am not yet sure if adding these lines of code is reasonable.

Issue #2030, opened 2025-11-03
Expose `RunItems` in `RunContextWrapper`


varunursekar:
Currently `RunContextWrapper` is meant to store context that is not exposed to LLMs and can be used to control the behaviour of tools. The default `AgentRunner` builds up `RunItems` iteratively and only exposes them to the user via `RunResult` once the tool loop is complete. Requesting for intermediate `RunItems` to be exposed via the `RunContextWrapper` to allow for alteration of tool behaviour based on the trace generated so far (e.g. number of turns, number of tool calls). The immediate value unlock is to pass any part of the generated context to sub-agents. For example, at step `i`, the main agent can fork its trajectory `N` ways and e.g. take the best output; importantly this is done without the main agent having to copy its context into the tool kwargs. 

Proposed structure:

```python
@dataclass
class RunContextWrapper(Generic[TContext]):
    context: TContext
    usage: Usage = field(default_factory=Usage)
    run_items: list[RunItem]
```

This will allow for usage such as:

```python
def my_sub_agent(cts: RunContextWrapper, prompt: str):
  """ My sub agent only needs the last 2 messages from my main trace"""
  input = [x.to_input_item() for x in ctx.run_items[-2:]] + [{"role" : "user", "content" : prompt}]
  return await Runner.run(Agent(...), input=input)
```

seratch:
Thanks for writing in! Sharing the entire conversation with a function tool (or an agent as a tool) can indeed be very flexible for certain use cases. However, the tool design in this SDK is intended more as an independent concept rather than for sharing conversations across tools.  

If you share conversation history with a tool, continue it within the tool, and then have the main agent pick up from the same point, it could lead to confusing results—like context mismatches between agents and tools. Also, when using the Responses API, the conversation state can branch into multiple paths, which isn’t really the intended usage.  

If you need a different agent to continue the same conversation, handoffs could be a good design pattern. For running an agent within a function tool, I believe it’s cleaner to define sufficient parameters on the function side and let the LLM decide what to pass.

PR #2031, opened 2025-11-03
Expose intermediate generated items in `RunContextWrapper`

varunursekar:
Minimal PR for [this](https://github.com/openai/openai-agents-python/issues/2030) issue I created


seratch:
Thanks for sharing this patch, but we don't plan to have this for this reason: https://github.com/openai/openai-agents-python/issues/2030#issuecomment-3483553197

PR #2034, opened 2025-11-03
fix(usage): Normalize None token details on Usage initialization

codefromthecrypt:
Some openai API clones (such as ollama) don't populate new token detail fields, resulting in None values that bypass Pydantic validation. This results in agent crashes like this:

```
    ^
  File "/Users/codefromthecrypt/.cache/uv/environments-v2/agent-ce802ac7ec8719af/lib/python3.14/site-packages/agents/run.py", line 1740, in _get_new_response
    context_wrapper.usage.add(new_response.usage)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/Users/codefromthecrypt/.cache/uv/environments-v2/agent-ce802ac7ec8719af/lib/python3.14/site-packages/agents/usage.py", line 76, in add
    cached_tokens=self.input_tokens_details.cached_tokens
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + other.input_tokens_details.cached_tokens
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

This fix adds a `__post_init__` method to normalize None to 0 for "cached_tokens" and "reasoning_tokens" on any Usage object creation.  This defensive approach handles the issue at the boundary we control (just after the stainless generated code), allowing absence of these newer fields to not break agents.

### DIFF
--- a/src/agents/usage.py
+++ b/src/agents/usage.py
@@ -60,6 +60,16 @@ class Usage:
         cost calculation or context window management.
     """
 
+    def __post_init__(self) -> None:
+        # Some providers don't populate optional token detail fields
+        # (cached_tokens, reasoning_tokens), and the OpenAI SDK's generated
+        # code can bypass Pydantic validation (e.g., via model_construct),
+        # allowing None values. We normalize these to 0 to prevent TypeErrors.
+        if self.input_tokens_details.cached_tokens is None:
+            self.input_tokens_details = InputTokensDetails(cached_tokens=0)
+        if self.output_tokens_details.reasoning_tokens is None:
+            self.output_tokens_details = OutputTokensDetails(reasoning_tokens=0)
+
     def add(self, other: "Usage") -> None:
         """Add another Usage object to this one, aggregating all fields.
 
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
FYI I'm aware it could be a bikeshed discussion if these fields are really optional or not. for example, they are more or less implied or explicitly required in the latest [openai openapi yaml](https://github.com/openai/openai-openapi/blob/manual_spec/openapi.yaml#L34779). However, there is defensive code around other token fields, that coerce to zero. Plus the impact is tricky to workaround otherwise even if PRs are raised everywhere. So, hoping we can add similar defense here

src/agents/usage.py
```diff
@@ -60,6 +60,13 @@ class Usage:
         cost calculation or context window management.
     """
 
+    def __post_init__(self) -> None:
+        # Generated code allows None to pass through validation
```
seratch: Can you adda bit more detailed comment like you described in this PR description?

codefromthecrypt: absolutely

codefromthecrypt: should be good now.. lemme know

codefromthecrypt:
ps looks like there was another case like this that popped up. Hopefully #2141 is a more sustainable fix, lemme know if you agree

Issue #2036, opened 2025-11-03
Feature Request: Support all AgentHooks when Agent.as_tool()


hayescode:
When running the [Manager pattern](https://openai.github.io/openai-agents-python/agents/#manager-agents-as-tools) the sub-agents I'm running as tools do not trigger the `on_tool_start` or `on_tool_end` methods of the AgentHooks. [The documentation](https://openai.github.io/openai-agents-python/agents/#lifecycle-events-hooks) does not mention this limitation so i'm not sure if this is a bug or feature request. Maybe related to #1057 .

<details>
<summary> <b>TestAgentHooks</b></summary>

```py
####### AgentHooks ########

from typing import Any, Generic, Optional
from agents import AgentHooks
from typing_extensions import TypeVar

from agents.agent import Agent, AgentBase
from agents.items import ModelResponse, TResponseInputItem
from agents.run_context import RunContextWrapper, TContext
from agents.tool import Tool

class TestAgentHooks(AgentHooks):

    async def on_start(self, context: RunContextWrapper, agent: Agent) -> None:
        """Called before the agent is invoked. Called each time the running agent is changed to this
        agent."""
        print(f"### AgentHook: on_start ({agent.name}) Agent {agent.name} starting...")

    async def on_end(
        self,
        context: RunContextWrapper,
        agent: Agent,
        output: Any,
    ) -> None:
        """Called when the agent produces a final output."""
        print(f"### AgentHook: on_end Agent ({agent.name}) finished with output: {output}")

    async def on_handoff(
        self,
        context: RunContextWrapper,
        agent: Agent,
        source: Agent,
    ) -> None:
        """Called when the agent is being handed off to. The `source` is the agent that is handing
        off to this agent."""
        print(f"### AgentHook: on_handoff Agent({agent.name}) is being handed off to from {source.name}...")

    async def on_tool_start(
        self,
        context: RunContextWrapper,
        agent: Agent,
        tool: Tool,
    ) -> None:
        """Called concurrently with tool invocation."""
        print(f"### AgentHook: on_tool_start Agent {agent.name} is invoking tool {tool.name}... Tool Context: {context.context}")

    async def on_tool_end(
        self,
        context: RunContextWrapper,
        agent: Agent,
        tool: Tool,
        result: str,
    ) -> None:
        """Called after a tool is invoked."""
        print(f"### AgentHook: on_tool_end Agent ({agent.name})  finished invoking tool {tool.name} with result: {result}")

[...]

Issue #1889, opened 2025-10-13
WebSearchTool no LifeCycle Events on Agent via AgentHook subclass


yvanaquino-alchemy:
### Please read this first
Have you read the docs? Yes - Agents SDK docs
Have you searched for related issues? Yes - found issue #778 which mentions similar problems in earlier versions, but this persists in v0.3.3

### Describe the bug

The on_tool_start and on_tool_end lifecycle hooks in AgentHooks are not being invoked when tools (specifically WebSearchTool) are executed by an agent. LLM lifecycle hooks (on_llm_start and on_llm_end) work correctly, but tool lifecycle hooks are completely silent with no output or errors.
The tools themselves are being executed successfully (web search results appear in agent responses), but the lifecycle events are never triggered. This makes it impossible to track, log, or intercept tool execution in custom hooks.

### Debug information

Agents SDK version: v0.3.3
Python version: Python 3.12
OpenAI version: 1.109.1

### Repro steps

```python
import asyncio
from agents import Agent, AgentHooks, Runner, WebSearchTool, RunContextWrapper, TContext, Tool
from agents.lifecycle import TAgent
from agents.models.openai_responses import OpenAIResponsesModel
from openai import AsyncOpenAI
from typing import Optional

class TestHooks(AgentHooks):
    async def on_llm_start(self, context: RunContextWrapper[TContext], 
                          agent: Agent[TContext], system_prompt: Optional[str], 
                          input_items: list) -> None:
        print("✅ LLM START - This prints correctly")
        await super().on_llm_start(context, agent, system_prompt, input_items)

    async def on_llm_end(self, context: RunContextWrapper[TContext], 
                        agent: Agent[TContext], response) -> None:
        print("✅ LLM END - This prints correctly")
        await super().on_llm_end(context, agent, response)

    async def on_tool_start(self, context: RunContextWrapper[TContext], 
                           agent: TAgent, tool: Tool) -> None:
        print(f"❌ TOOL START - This never prints: {tool.name}")
        await super().on_tool_start(context, agent, tool)

    async def on_tool_end(self, context: RunContextWrapper[TContext], 
                         agent: TAgent, tool: Tool, result: str) -> None:
        print(f"❌ TOOL END - This never prints: {tool.name}")
        await super().on_tool_end(context, agent, tool, result)

async def main():
    client = AsyncOpenAI(api_key='your-api-key')
    model = OpenAIResponsesModel(model='gpt-4o-mini', openai_client=client)
    
[...]

Issue #778, opened 2025-05-28
web_search_preview never triggers agents.lifecycle.AgentHooks.on_tool_start


jihun-im-open:
### Please read this first
Yes

### Describe the bug
These functions are not called the tool is web_search_preview
- agents.lifecycle.AgentHooks.on_tool_start
- agents.lifecycle.AgentHooks.on_tool_end
- agents.agent.Agent.tool_use_behavior (A function)

### Debug information
- Agents SDK version: v0.0.15
- Python version: Python 3.11

### Repro steps
Ideally provide a minimal python script that can be run to reproduce the bug.
https://gist.github.com/jihun-im-open/a3892127a0e406500df7e05ff708f090

### Expected behavior
The hooks should be triggered.
As it's tool input is a union of all tools
https://github.com/openai/openai-agents-python/blob/6e078bf7a99badc5a36be961e4e6dd6dde4c3674/src/agents/tool.py#L234

rm-openai:
I should document this better (and we should have named it better), but the on_tool_start is for _local_ tools

PR #2037, opened 2025-11-04
docs: clarify on_tool_start and on_tool_end hooks for local tools only

ihower:
This PR improves the doc comments for the `on_tool_start` and `on_tool_end` hooks.

Clarified that these hooks are for local tools only, and do not include hosted tools running on the OpenAI server side (e.g., `WebSearchTool`, `FileSearchTool`, `CodeInterpreterTool`, `HostedMCPTool`, etc.).

### Related Issues

- [https://github.com/openai/openai-agents-python/issues/778](https://github.com/openai/openai-agents-python/issues/778) rm-openai said: “I should document this better”, this PR addresses that note.
- [https://github.com/openai/openai-agents-python/issues/1889](https://github.com/openai/openai-agents-python/issues/1889)
- [https://github.com/openai/openai-agents-python/issues/2036](https://github.com/openai/openai-agents-python/issues/2036)

### DIFF
--- a/examples/basic/agent_lifecycle_example.py
+++ b/examples/basic/agent_lifecycle_example.py
@@ -28,6 +28,10 @@ class CustomAgentHooks(AgentHooks):
             f"### ({self.display_name}) {self.event_counter}: Agent {source.name} handed off to {agent.name}"
         )
 
+    # Note: The on_tool_start and on_tool_end hooks apply only to local tools.
+    # They do not include hosted tools that run on the OpenAI server side,
+    # such as WebSearchTool, FileSearchTool, CodeInterpreterTool, HostedMCPTool,
+    # or other built-in hosted tools.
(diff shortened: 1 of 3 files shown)


hayescode:
Thank you @ihower ! @seratch any plans to support these hooks for hosted tools? Responses API seems to have create/delta/completed events for all tools, hosted and local, and particularly for MCP servers with lots of tools we don't have a way to see what MCP tools were called, their arguments, outputs, etc. at least for Agent.as_tool().

seratch:
@hayescode when you use streaming, it may be possible to relay those events to the agent lifecycle hooks. But for developers who use non-streaming mode, it's not yet feasible to support the same. in the short term, enabling those hooks only for streaming mode could be indeed a good feature addition on this SDK side (although I myself haven't verified if it's really feasible without any restrictions).

PR #2052, opened 2025-11-06
fix #844: add resume helpers for MaxTurnsExceeded runs

### LINKED ISSUE #844: Max turns exceeded
Hello, 

I am wondering how we could control the agent not to exceed the maximum number of turns. Or, if we encounter this situation, can we give it a strike instruction to provide a FINAL answer based on previous thoughts/tool results?

I have added a Hook:

```python
def make_dynamic_prompt(self, base_prompt: str, max_turns: int):
    """
    Create a dynamic prompt for the agent.
    """
    async def _prompt(ctx: RunContextWrapper[None], agent: Agent) -> str:
        done = ctx.usage.requests
        left = max(max_turns - 1 - done, 0)

        suffix = f"Current turn: {self.current_turn}\nMaximum turns: {max_turns} | Turns left: {left}\n\n"
        if left == 0:
            suffix += "\n\nYou've reached the search limit. Please answer now."

            agent.tools.clear()
            agent.model_settings = replace(
                agent.model_settings,
                tool_choice="none"
            )

        return f"{base_prompt}\n\n{suffix}"
    return _prompt
```

However, the agent mostly ignores it and encounters an error.

thomas-lab-openai:
This pull request resolves https://github.com/openai/openai-agents-python/issues/844:

- add async `resume()` and sync `resume_sync()` helpers on `MaxTurnsExceeded` that replay one final turn with an optional user prompt, reusing the prior agent, context, and a copy of the original `RunConfig`
- extend `RunResult` and `RunErrorDetails` to retain the effective RunConfig, so guardrail/tool settings, tracing metadata, and other knobs survive into a resume attempt
- ensure resume turns disable further tool calls by forcing `tool_choice="none"` while still respecting any other model settings (e.g., temperature) that the caller supplied
- cover the new helpers with unit tests for async/sync flows, verify model-settings preservation, and document the flow via a `examples/basic/max_turns_resume.py` sample that demonstrates recovering after a hard turn limit

### DIFF
--- a/tests/test_max_turns.py
+++ b/tests/test_max_turns.py
@@ -170,7 +170,9 @@ async def test_resume_preserves_run_config_settings():
     assert final_settings.temperature == 0.25
     assert final_settings.tool_choice == "none"
 
-    stored_settings = exc_info.value.run_data.run_config.model_settings
+    run_data = exc_info.value.run_data
+    assert run_data is not None
+    stored_settings = run_data.run_config.model_settings
     assert stored_settings is not None
     assert stored_settings.temperature == 0.25
     assert stored_settings.tool_choice == "auto"


seratch:
You’ve assigned @rm-openai  as a reviewer, so he may come up with other suggestions, but personally, I feel adding the methods to the exception class this way may not be a clean design. I would add some kind of helper that handles the max exceed error and resumes.  

Aside from the implementation details, I’m not yet sure whether it helps when the max turns are exceeded. Even if you resume it, it might hit the max turns again.

PR #2053, opened 2025-11-06
multi_provider: validation and error handling improvements

mshsheikh:
This PR adds defensive checks and clearer error messages to the multi provider implementation.

### Summary of changes

1. Validate model name in get_model and reject None or empty values.
2. Normalize empty model parts in _get_prefix_and_model_name so inputs like "openai/" are handled as missing model name.
3. Replace raw delete on provider mapping with a check and UserError to avoid KeyError.
4. Catch ImportError for optional litellm provider import and raise a helpful UserError that explains how to resolve the issue.
5. Validate set_mapping inputs and shallow copy the provided mapping to prevent external mutation.
6. Small docstring and annotation fixes.

### DIFF
--- a/src/agents/models/multi_provider.py
+++ b/src/agents/models/multi_provider.py
@@ -82,8 +82,8 @@ class MultiProvider(ModelProvider):
             provider_map: A MultiProviderMap that maps prefixes to ModelProviders. If not provided,
                 we will use a default mapping. See the documentation for this class to see the
                 default mapping.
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. However, the type hints for the arguments are clear, so I don't see the necessity to have these additional validation for obviously wrong data. If we do this for many parts of this SDK, it will make the internals unnecessarily complex.

mshsheikh:
Got it, thanks for the clarification, @seratch.

PR #2060, opened 2025-11-09
feat: add MCP Resources support

MuhammadHassaanArain:
## Summary
Adds MCP Resources support (list, read, subscribe) to complete protocol implementation.

## Changes
- `list_resources()` - discover available resources
- `read_resource(uri)` - read resource content
- `subscribe_resource(uri)` - stream resource updates

Resources enable agents to access files, databases, and data streams per the MCP spec.

## Details
- Non-breaking change
- Follows existing SDK patterns
- Proper error handling and async/await
- Delegates to ClientSession

Happy to add tests or docs if needed!

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -92,12 +92,12 @@ class MCPServer(abc.ABC):
         pass
 
     @abc.abstractmethod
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this patch. However, we don't plan to add these methods for now because the core part of this SDK does not use MCP resources at all.

Issue #1930, opened 2025-10-18
Responses API Error 400 Bad request : Missing mutually exclusive parameters


sridharvenkat:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) - Yes
- **Have you searched for related issues?** Others may have faced similar issues. - Yes

### Describe the bug
The recent upgrade to 0.40. causing errors after function_tool_call processing.

`Error code: 400 - {'error': {'message': "Missing mutually exclusive parameters: 'input[17].output[0]'. Ensure you are providing exactly one of: 'file_id' or 'image_url'.", 'type': 'invalid_request_error', 'param': 'input[17].output[0]', 'code': 'missing_mutually_exclusive_parameters'}} `

### Debug information
- Agents SDK version: 0.4.0
- Python version 3.12

### Repro steps
We are noticing after the latest upgrade of SDK response api interactions are failing after some function tool calls.
Unsure if the below item in the session is causing issue
````json
{
  "call_id": "call_z1rcE3rHla5uHQmpMwI2im2n",
  "output": [
    {
      "type": "input_image"
    }
  ],
  "type": "function_call_output"
}
````

### Expected behavior
The SDK should not error after function call

amirna2:
I ran into the same exact issue today after upgrading from 0.3.0 to 0.4.0.  There appears to be a regression with tools structure output when they return a pydantic object. I got around the issue by remove the Session SQLite DB and starting fresh with 0.4.0..Obviously not something you want to do in production agents.

Tools in general have been a pain to deal with.

ihower:
@sridharvenkat @amirna2 I looked into this issue and my best guess is that:

1. In one of your previous function calls, the output didn’t include `image_url`. You might have accidentally returned `None`, for example:

   ```python
   @function_tool
   def fetch_image() -> ToolOutputImage | ToolOutputImageDict:
       return {
           "type": "image",
           "image_url": None,
       }
   ```

   When this gets sent to the OpenAI API, it triggers exactly the error you mentioned:

   ```
   openai.BadRequestError: Error code: 400 - {'error': {'message': "Missing mutually exclusive parameters: 'input[2].output[0]'. Ensure you are providing exactly one of: 'file_id' or 'image_url'.", 'type': 'invalid_request_error', 'param': 'input[2].output[0]', 'code': 'missing_mutually_exclusive_parameters'}}
   ```

[...]

PR #2062, opened 2025-11-09
fix: #2061 Sanitize invalid JSON in tool call arguments

### LINKED ISSUE #2061: Invalid JSON in tool call arguments breaks session with Anthropic models
### Problem
When a tool call contains invalid JSON arguments (e.g., missing closing brace), the session stores the invalid JSON string. This breaks the session with Anthropic models, as LiteLLM attempts to parse the arguments from JSON strings into JSONs (as per Anthropic API specs), and fails rightfully.

### Root Cause
1. Invalid JSON gets stored in session history in `tool_call.arguments`
2. On reload, `chatcmpl_converter.py:527` passes the invalid JSON string directly to litellm
3. litellm converts OpenAI format (JSON string) → Anthropic format (dict) by calling `json.loads()`
4. JSON parsing fails before tool invocation: `"Expecting ',' delimiter: line 1 column 30"`

#### Note: 
Custom `FunctionTool` implementations can bypass the JSON validation that only happens in `_on_invoke_tool_impl` within `function_tool`, so this could cause unexpected behavior in those tools.

### Repro Script:
```py
@function_tool
def get_server_time(format: str = "%Y-%m-%d %H:%M:%S") -> str:
    """
    Returns the current server time, optionally in a specified format.
    """
    return datetime.now().strftime(format)


async def main():
    # Mimic session memory functionality.
    input_list = [
        {"content": "What is the current server time?", "role": "user"},
        {
            "arguments": '{"format":"%Y-%m-%d %H:%M:%S"',  # Invalid JSON on purpose
            "call_id": "call_9YhapDWwiCFcoGZilhyUo8ti",
            "name": "get_server_time",
            "type": "function_call",
            "id": "fc_0e0babe7f7b9bf95006910adb438588193a96b6a1d917c77e7",
            "status": "completed",
        },
        {
            "call_id": "call_9YhapDWwiCFcoGZilhyUo8ti",
            "output": "An error occurred while running the tool. Please try again. Error: Not implemented",
        },
        {
            "id": "msg_0e0babe7f7b9bf95006910adb5a4cc8193b5e3a1a171b76db1",
            "content": [
                {
                    "annotations": [],
                    "text": "I encountered an error trying to get the time now. Would you like me to try again?",
                    "type": "output_text",
                    "logprobs": [],
                }
            ],
            "role": "assistant",
            "status": "completed",
            "type": "message",
        },
    ]

    agent = Agent(
        name="Assistant",
        model="litellm/anthropic/claude-haiku-4-5-20251001",
        tools=[get_server_time],
    )

    print("User: How about now?")
    result = await Runner.run(
        agent,
        input_list + [{"content": "How about now?", "role": "user"}],
    )

    print(result.final_output)
```

habema:
Resolves #2061

Fixes issue where invalid JSON in tool call arguments (stored in session history) would break API calls when using Anthropic models via LiteLLM.

**Solution:**
- Added `_sanitize_tool_call_arguments()` method that validates JSON before sending to APIs
- If invalid, sanitizes to `"{}"` as safe default (prevents API errors while maintaining compatibility)
- Tests

Not too sure if the logging should be more verbose, I'll leave that to you.

### DIFF
--- a/tests/test_openai_chatcompletions_converter.py
+++ b/tests/test_openai_chatcompletions_converter.py
@@ -393,7 +393,7 @@ def test_tool_call_with_valid_json_arguments():
 
     messages = Converter.items_to_messages([function_call])
     tool_call = messages[0]["tool_calls"][0]  # type: ignore
-    assert tool_call["function"]["arguments"] == '{"key": "value", "number": 42}'  # type: ignore
+    assert tool_call["function"]["arguments"] == '{"key": "value", "number": 42}'
 
 
 def test_tool_call_with_invalid_json_arguments():
@@ -414,7 +414,7 @@ def test_tool_call_with_invalid_json_arguments():
     messages = Converter.items_to_messages([function_call])
     tool_call = messages[0]["tool_calls"][0]  # type: ignore
     # Invalid JSON should be sanitized to "{}"
-    assert tool_call["function"]["arguments"] == "{}"  # type: ignore
+    assert tool_call["function"]["arguments"] == "{}"
 
     # Test with None
     function_call_none = ResponseFunctionToolCallParam(
@@ -427,7 +427,7 @@ def test_tool_call_with_invalid_json_arguments():
 
     messages_none = Converter.items_to_messages([function_call_none])
     tool_call_none = messages_none[0]["tool_calls"][0]  # type: ignore
-    assert tool_call_none["function"]["arguments"] == "{}"  # type: ignore
+    assert tool_call_none["function"]["arguments"] == "{}"
 
     # Test with empty string
     function_call_empty = ResponseFunctionToolCallParam(
@@ -440,7 +440,7 @@ def test_tool_call_with_invalid_json_arguments():
 
     messages_empty = Converter.items_to_messages([function_call_empty])
     tool_call_empty = messages_empty[0]["tool_calls"][0]  # type: ignore
-    assert tool_call_empty["function"]["arguments"] == "{}"  # type: ignore
+    assert tool_call_empty["function"]["arguments"] == "{}"
 
 
 def test_tool_call_with_malformed_json_variants():
@@ -469,7 +469,7 @@ def test_tool_call_with_malformed_json_variants():
         messages = Converter.items_to_messages([function_call])
         tool_call = messages[0]["tool_calls"][0]  # type: ignore
         # All malformed JSON should be sanitized to "{}"
-        assert tool_call["function"]["arguments"] == "{}", (  # type: ignore
+        assert tool_call["function"]["arguments"] == "{}", (
             f"Malformed JSON '{malformed_json}' should be sanitized to '{{}}'"
         )
 


seratch:
Thanks for sending this PR. Validating the JSON data format itself is fine (OpenAI SDK's code comment recommends doing so too), but I am unsure if falling back to an empty JSON data works well in the scenario, and it may be even more confusing for developers to debug. Sending an invalid JSON data to the model and receiving the error clearly telling the arguments are broken may be okay.

habema:
@seratch Your concern is valid, but this issue causes sessions to break entirely. Also, when this issue occurs, the currently raised errors are just as confusing for debugging; it is not clear that the issue is in a certain tool call's arguments.

Examples:
```
litellm.BadRequestError: AnthropicException - Expecting ',' delimiter: line 1 column 42 (char 41)
Received Messages=[...] # Entire input list
```
```
litellm.BadRequestError: AnthropicException - Unterminated string starting at: line 1 column 47 (char 46)`
Received Messages=[...] # Entire input list
```

Perhaps this SDK could raise more informative error, or proceed with the sanitization but with a warning level instead of debug. I'm also open to hearing your suggestions.

ihower:
@habema The `json.loads` error actually comes from LiteLLM. So if we want to improve the error message, the right place to fix it may be in LiteLLM, not here.

From a broader perspective, I think the root issue is related to the current session design.
As I mentioned in [#1930 (comment)](https://github.com/openai/openai-agents-python/issues/1930#issuecomment-3419327603), the current Session implementation saves those problematic items before the LLM request is sent. Once this happens, the conversation thread becomes permanently broken, because every later call will hit the same API error since the corrupted items are always included in the following request. The only workaround is to start a new thread.

Maybe we could provide a small helper function to clear out problematic session items, but deciding when to use it should probably be left to the developer, since we cannot know exactly which server-side API errors require clearing. It’s more like a retry or recovery feature.

Just my two cents.

habema:
@ihower Thank you.

Yes, the `litellm.BadRequestError` is raised by LiteLLM. As mentioned in the original issue, this behavior is right, as the tool call's arguments violate Anthropic's API specs. That being said, the right place to fix such an issue is here, not LiteLLM, as this SDK is the one responsible for using LiteLLM's (and by extension Anthropic's) public API correctly.

As for the session's design, I think this can be very opinionated. In my opinion, I think the session should reflect the exact interactions that happen without alteration. If any sanitization is needed, this can be done inside the `get_items()` within a custom session implementation, which is how I managed to circumvent this issue. 

On one side, this takes away the responsibility of handling such problems from this SDK's core internals, and delegates to the session, which can be implemented in a custom manner (and at this point I believe any production use-case will need its own custom session implementation). On the other side, and as @seratch mentioned earlier: 

> Sending an invalid JSON data to the model and receiving the error clearly telling the arguments are broken may be okay.

After all the broken JSON is what the agent sent.

Since I managed to solve this issue by implementing a function very similar `_sanitize_tool_call_arguments()` with my custom session, my only concern moving forward is making this easier to debug within the SDK. I am open to hearing suggestions regarding that, and if they are unrelated to the changes in this PR, this can be closed.

seratch:
Still thinking of a viable solution for this use case; sharing ideas and/or continuing discussions would be appreciated

PR #2064, opened 2025-11-10
Add GitHub Actions workflow for Python application

wenchyuan:
This workflow installs Python dependencies, runs linting with flake8, and executes tests with pytest.


seratch:
Thanks for sending this, but we already have the workflows for running lint and tests: https://github.com/openai/openai-agents-python/tree/main/.github/workflows

PR #2077, opened 2025-11-12
fix: #2073 Improve type hinting in memory extension modules

### LINKED ISSUE #2073: Incorrect import path in SQLAlchemySession documentation
### Description
The documentation currently shows an incorrect import path for `SQLAlchemySession`.

### Current (incorrect) code
```python
from agents.extensions.memory import SQLAlchemySession
```

### Correct code
```python
from agents.extensions.memory.sqlalchemy_session import SQLAlchemySession
```

Found on the official docs page:
(https://openai.github.io/openai-agents-python/sessions/sqlalchemy_session)

This small fix will help avoid ImportError for users following the documentation.

habema:
Resolves #2073 

Fixed linter not picking up imports from `agents.extensions.memory`. 

The module was using lazy imports via `__getattr__` which worked at runtime but broke static analysis. Added `TYPE_CHECKING` imports so linters/IDEs can provide autocomplete and type checking.

Also cleaned up mypy config by excluding `examples/` and moved `site/` exclusion from the Makefile to `pyproject.toml` for consistency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,7 +117,6 @@ strict = true
 disallow_incomplete_defs = false
 disallow_untyped_defs = false
 disallow_untyped_calls = false
-exclude = ["examples/", "site/"]
 
 [[tool.mypy.overrides]]
 module = "sounddevice.*"
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ lint:
 
 .PHONY: mypy
 mypy: 
-	uv run mypy .
+	uv run mypy . --exclude site
 
 .PHONY: tests
 tests: 
--- a/examples/memory/advanced_sqlite_session_example.py
+++ b/examples/memory/advanced_sqlite_session_example.py
@@ -132,7 +132,7 @@ async def main():
     # Show current conversation
     print("Current conversation:")
     current_items = await session.get_items()
-    for i, item in enumerate(current_items, 1):
+    for i, item in enumerate(current_items, 1):  # type: ignore[assignment]
         role = str(item.get("role", item.get("type", "unknown")))
         if item.get("type") == "function_call":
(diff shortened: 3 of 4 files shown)


pyproject.toml
```diff
@@ -117,6 +117,7 @@ strict = true
 disallow_incomplete_defs = false
 disallow_untyped_defs = false
 disallow_untyped_calls = false
+exclude = ["examples/", "site/"]
```
seratch: >Also cleaned up mypy config by excluding examples/

I don't get the point on why you excluded examples this time. can you elaborate more on that?

habema: Sure. Fixing the type hint caused 106 mypy errors, 41 were in the `examples/`. Fixing them makes the examples a bit unreadable, my assumption is that they are meant for demos and not production code that should be type-checked strictly. What do you think?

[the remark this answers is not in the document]
habema: Nope. List works fine.

habema: @seratch I've tried resolving the mypy issues in the examples and they're not as bad as a I thought. I remember trying to tackle this a while back and it got messy, but this time its working fine. I can proceed however you like.

seratch: If we get a large number of mypy errors as the result of the change, I don't think we should make it.

habema: Just to clarify, fixing the typing hints is what caused the mypy errors. I have made a new commit in which they are fixed. Either we proceed with fixed type hinting and fixed mypy errors in example scripts (merge as is), or with fixed type hinting and exclude examples from the mypy check (remove last commit's changes).

seratch: Thanks, I wanted to avoid the exclusition. The current number of skip comments is fine

habema: Thank you @seratch, sorry for the confusion.

PR #2082, opened 2025-11-14
fix: #2081 Redis SSL Connection Error with redis-py v7

### LINKED ISSUE #2081: Redis SSL Connection Error with redis-py
# Redis SSL Connection Error with redis-py

## Description

When using `RedisSession.from_url()` with SSL-enabled Redis URLs (`rediss://`), the application crashes with a `TypeError` when redis-py 7.x is installed. This issue occurs because `RedisSession.from_url()` passes `ssl=True` as a keyword argument, which is incompatible with redis-py 7.x's connection API.

## Error Message

```
TypeError: AbstractConnection.__init__() got an unexpected keyword argument 'ssl'
```

## Environment

- **openai-agents**: 0.5.0
- **redis-py**: 7.0.1 (installed via `openai-agents[redis]`)
- **Python**: 3.12
- **Redis Server**: Upstash Redis (SSL required, `rediss://` URL scheme)

## Root Cause

In redis-py 7.x, `AbstractConnection.__init__()` does not accept an `ssl` boolean parameter. The `rediss://` URL scheme automatically triggers SSL connection handling.

However, `RedisSession.from_url()` in `agents/extensions/memory/redis_session.py` adds `ssl=True` to kwargs when it detects a `rediss://` URL:

```python
@classmethod
def from_url(cls, session_id: str, *, url: str, redis_kwargs: dict[str, Any] | None = None, **kwargs: Any) -> RedisSession:
    redis_kwargs = redis_kwargs or {}
    
    # Parse URL to determine if we need SSL
    parsed = urlparse(url)
    if parsed.scheme == "rediss":
        redis_kwargs.setdefault("ssl", True)  # ❌ Causes TypeError in redis 7.x
    
    redis_client = redis.from_url(url, **redis_kwargs)
    session = cls(session_id, redis_client=redis_client, **kwargs)
    session._owns_client = True
    return session
```

This `ssl=True` kwarg gets passed through the call chain:
```
redis.from_url() → ConnectionPool.from_url() → ConnectionPool.__init__() → AbstractConnection.__init__()
```

 `AbstractConnection.__init__()` does not accepts the `ssl` parameter, resulting in the TypeError.

## Reproduction

```python
import asyncio
from agents.extensions.memory import RedisSession

async def test():
    # This fails with redis 7.x
    session = RedisSession.from_url(
        "test_session",
        url="rediss://default:password@host:6379"
    )
    await session.add_items({"key1": {"data": "value1"}})

asyncio.run(test())
```

## Impact

This affects any user of `openai-agents[redis]` connecting to SSL-enabled Redis servers, as the current dependency specification allows redis-py 7.x:

```
redis = ["redis>=6.4.0"]
```

In earlier versions of redis-py, it’s unclear whether this issue occurred, but in version 7.x it definitely does.

yju0808:
# Fix Redis SSL Connection Error with redis-py

## Summary

This PR fixes a `TypeError` that occurs when using `RedisSession.from_url()` with SSL-enabled Redis URLs (`rediss://`) and redis-py 7.x.

## Problem

`RedisSession.from_url()` currently adds `ssl=True` to kwargs when detecting a `rediss://` URL. This causes a `TypeError` because `AbstractConnection.__init__()` does not accept the `ssl` parameter.

**Error:**
```
TypeError: AbstractConnection.__init__() got an unexpected keyword argument 'ssl'
```

## Changes

- **Removed**: The line that adds `ssl=True` to `redis_kwargs` when `rediss://` is detected
- **Removed**: import of `urlparse` (no longer used)
- **Reason**: In redis-py 7.x, the `rediss://` URL scheme automatically triggers SSL handling via `parse_url()`, making the explicit `ssl=True` parameter unnecessary and incompatible

## Related Issue

Fixes #2081

### DIFF
--- a/src/agents/extensions/memory/redis_session.py
+++ b/src/agents/extensions/memory/redis_session.py
@@ -255,7 +255,7 @@ class RedisSession(SessionABC):
             True if Redis is reachable, False otherwise.
         """
         try:
-            await self._redis.ping()
+            await self._redis.ping()  # type: ignore[misc]  # Redis library returns Union[Awaitable[T], T] in async context
             return True
         except Exception:
             return False
--- a/tests/extensions/memory/test_redis_session.py
+++ b/tests/extensions/memory/test_redis_session.py
@@ -492,14 +492,14 @@ async def test_external_client_not_closed():
         assert len(items) == 1
 
         # Verify client is working before close
-        assert await shared_client.ping() is True
+        assert await shared_client.ping() is True  # type: ignore[misc]  # Redis library returns Union[Awaitable[T], T] in async context
 
         # Close the session
         await session.close()
 
         # Verify the shared client is still usable after session.close()
         # This would fail if we incorrectly closed the external client
-        assert await shared_client.ping() is True
+        assert await shared_client.ping() is True  # type: ignore[misc]  # Redis library returns Union[Awaitable[T], T] in async context
 
         # Should still be able to use the client for other operations
         await shared_client.set("test_key", "test_value")
@@ -781,7 +781,7 @@ async def test_close_method_coverage():
     await session1.close()
 
     # Verify external client is still usable
-    assert await external_client.ping() is True
+    assert await external_client.ping() is True  # type: ignore[misc]  # Redis library returns Union[Awaitable[T], T] in async context
 
     # Test 2: Internal client (should be closed)
     # Create a session that owns its client


yju0808:
> To make things simpler, I'd like to change the minimal requreid version of redis client to ">=7". Can you apply the change as well?

@seratch Updated!
I’ve updated the minimum required Redis version to >=7 as requested.

Changes included:
-Updated pyproject.toml: redis>=6.4.0 → redis>=7
-Regenerated uv.lock: resolved to redis==7.0.1

Let me know if anything else is needed!

seratch (approved):

yju0808:
@seratch Fixed!

I've resolved all mypy errors by adding # type: ignore[misc] comments to the Redis async ping() calls.

Changes included:
-Added type: ignore[misc] comments for ping() calls in redis_session.py
-Added the same ignores for corresponding assertions in test_redis_session.py

Root cause:
The redis-py library defines return types as Union[Awaitable[T], T] for sync/async compatibility, but mypy cannot infer that redis.asyncio.Redis always returns Awaitable[T] in async context.

Note:
The codebase already uses the same type: ignore[misc] pattern for other Redis async methods (e.g., lrange() and rpop()), so I followed the existing convention for consistency.

PR #2059, opened 2025-11-07
fix: #2009 #2010 Improve PCM duration calculation and handle VAD truncation

ruskaruma:
### Summary
In #2010 we noticed realtime PCM16 audio was being reported in microseconds. The root cause was our fallback branch in `calculate_audio_length_ms`, which divided by `24 * 2` and then by 1000, effectively treating every byte as a millisecond. I normalized the format first, split g711 from PCM handling, and introduced explicit constants for the sample rate and sample width so PCM16 now returns the true millisecond duration. This also covers empty buffers and any future lowercase/uppercase format variants.

To keep coverage honest, I updated the realtime playback tracker tests to assert against the corrected math using `pytest.approx`, and refreshed the manual tracker test to reflect the new millisecond totals.

### Test plan
- uv run pytest tests/realtime/test_playback_tracker_manual_unit.py tests/realtime/test_playback_tracker.py tests/realtime/test_openai_realtime.py
- uv run pytest

Fixes #2009
Fixes #2010

PR #2083, opened 2025-11-15
[bug] Fix sample rate calculation

Daniel-Chin:
Originally off by 1000


seratch:
Thanks for sending this patch! However, #2059 already suggested this change!

Daniel-Chin:
> Thanks for sending this patch! However, #2059 already suggested this change!

Thanks, that's good. This is so small I didn't check for duplication. ;-)

PR #2105, opened 2025-11-19
fix: #2097 fix the default settings for SIP realtime runner

### LINKED ISSUE #2097: Wrong audio format override - SIP
### Describe the bug

In case of SIP integration, input_audio_format and output_audio_format should not be overridden with the defaults specified here: https://github.com/openai/openai-agents-python/blob/9bcf0817efda3773b74cffeec8732607bb0d4613/src/agents/realtime/openai_realtime.py#L145-L146

### Debug information
- Agents SDK version: `v0.5.1`
- Python version 3.13

### Repro steps

1. Initiate a SIP call from a client that is using g711-ulaw or g711-alaw
2. Realize that audio is going to be distorted because of the defaults

### Expected behavior
When call_id is present (indicating a SIP session), the client should not override the audio format imposed by SIP, when doing the connect on the websocket.

seratch:
This pull request resolves #2097 by adjusting the default configuration for SIP realtime use cases.

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -198,6 +198,7 @@ class OpenAIRealtimeWebSocketModel(RealtimeModel):
         self._playback_tracker: RealtimePlaybackTracker | None = None
         self._created_session: OpenAISessionCreateRequest | None = None
         self._server_event_type_adapter = get_server_event_type_adapter()
+        self._call_id: str | None = None
 
     async def connect(self, options: RealtimeModelConfig) -> None:
         """Establish a connection to the model and keep it alive."""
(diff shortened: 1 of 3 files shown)


src/agents/realtime/openai_realtime.py
```diff
@@ -845,10 +850,13 @@ def _get_session_config(
             "turn_detection",
             DEFAULT_MODEL_SETTINGS.get("turn_detection"),
         )
-        output_audio_format = model_settings.get(
-            "output_audio_format",
-            DEFAULT_MODEL_SETTINGS.get("output_audio_format"),
-        )
+        if self._call_id:
+            output_audio_format = model_settings.get("output_audio_format")
```
seratch, line 854: If a developer does not explicitly pass format settings, OpenAIRealtimeSIPModel relies on the default one.

PR #2108, opened 2025-11-19
fix: allow static tool filtering without agent and run_context in MCPServer

sevakva:
This PR fixes an issue in `MCPServer` where `run_context` and `agent` parameters were required for all tool filtering scenarios, even when using static tool filtering that doesn't need them.

The fix moves the validation check for `run_context` and `agent` from the caller level into the dynamic filter branch of `_apply_tool_filter`.

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -150,8 +150,8 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
     async def _apply_tool_filter(
         self,
         tools: list[MCPTool],
-        run_context: RunContextWrapper[Any],
-        agent: AgentBase,
+        run_context: RunContextWrapper[Any] | None = None,
+        agent: AgentBase | None = None,
     ) -> list[MCPTool]:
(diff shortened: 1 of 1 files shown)


seratch:
The _apply_tool_filter method always receives those parameters, so I don't see the necessity of this change. Can you elaborate more on the context?

sevakva:
> The _apply_tool_filter method always receives those parameters, so I don't see the necessity of this change. Can you elaborate more on the context?

Hey @seratch 
We have a need to create MCPServer before agent run and have it globally. Then we call `.list_tools()` method to construct a global list of tools on startup, so frontend can fetch it.
Calling this method outside of the agent run with static filter  causing an error

seratch:
Ah, I see. This makes sense.

seratch (approved):

mukul-atomicwork:
Temporal Implementation of this also calls without context and an agent. Because of this, it wasn't possible to make tool filtering through temporal impl. Thanks for fixing this @sevakva .

Issue #2110, opened 2025-11-20
ToolsToFinalOutputResult does not work properly when the Agent is used as a tool function


xJkit:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug
Using _custom_tool_behavior_handler_ with `ToolsToFinalOutputResult` in `tool_use_behavior` allows an agent to directly return a tool’s result when certain conditions are met, effectively bypassing the agent’s reasoning time. However, this mechanism fails when the agent is used as a tool by another agent.

In such cases, the tool-agent correctly returns the tool’s function output when the conditions are met, but the main agent receives only an empty string. This behavior is unexpected.

Please check the minimal code example below.

### Debug information
- Agents SDK version: (e.g. `v0.4.2`)
- Python version (e.g. Python 3.10)

### Repro steps

```python
from agents import Agent, function_tool
from agents.agent import ToolsToFinalOutputResult

@function_tool
def add_numbers_tool(a: int, b: int) -> int:
    return a + b


def custom_tool_behavior_handler(
    context, tool_results,
) -> ToolsToFinalOutputResult:
    """Processes tool results to decide final output."""
    for result in tool_results:
        if result.output > 0:
            return ToolsToFinalOutputResult(
                is_final_output=True, final_output=result.output
            )
    return ToolsToFinalOutputResult(is_final_output=False, final_output=None)

adding_agent = Agent(
    name="Sub Agent",
    instructions="You are a helpful assistant that adds two numbers.",
    model="gpt-5-mini",
    tools=[add_numbers_tool],
    tool_use_behavior=custom_tool_behavior_handler,
)

main_agent = Agent(
    name="Main Agent",
    instructions="You are a helpful assistant.",
    model="gpt-5-mini",
    tools=[adding_agent.as_tool(name="sub_agent_tool")],
)

```

### Expected behavior
The main agent will get the proper value rather than an empty string returned by the tool agent (sub tool agent, use `Agent.as_tool()`).

PR #2112, opened 2025-11-21
Fix as_tool returning blank string on early tool termination

ihower:
Resolved: https://github.com/openai/openai-agents-python/issues/2110

This PR fixes a bug where using `Agent.as_tool` together with `tool_use_behavior` causes the lead agent to receive blank string whenever the tool terminates early.

This issue affects not only the case described in https://github.com/openai/openai-agents-python/issues/2110 (ToolsToFinalOutputResult), but also `stop_on_first_tool` and `StopAtTools`.

## Root Cause

`as_tool` currently returns:

    ItemHelpers.text_message_outputs(output.new_items)

This function only extracts text from `output.new_items`. But when a tool stops early, `new_items` contains no text messages, so the return value becomes an blank string `""`.

## Solution

Return `final_output` instead.  It already handles early tool-return scenarios correctly.

## Behavior Changes

1. `text_message_outputs(output.new_items)` includes the final assistant message and any intermediate preamble assistant messages.

   * Switching to `final_output` removes these preambles. Only final assistant message.
   * I think this is fine and reasonable because:

     1. preamble messages are rare unless explicitly prompted
     2. preambles are intended for human readability during tool-calling explanations, and they unnecessarily consume the lead agent’s context window
     3. when preambles exist and the tool terminates early, `text_message_outputs(output.new_items)` would return the preambles, not actual tool result, which is incorrect anyway.

2. `final_output` may be structured output instead of a plain string.

   * This aligns with the new tool-call behavior, where tool outputs are no longer restricted to text value and can be structured data (i.e. image or files)

## Repro Code

```
from agents import Agent, function_tool, Runner
from agents.agent import ToolsToFinalOutputResult
import random

@function_tool
def generate_random_number_tool() -> int:
    return random.randint(0, 100)


random_number_agent = Agent(
    name="Sub Agent",
    instructions="You are a helpful assistant that generate_random_number_tool",
    model="gpt-5-mini",
    tools=[generate_random_number_tool],
    tool_use_behavior="stop_on_first_tool"
)

main_agent = Agent(
    name="Main Agent",
    instructions="You are a helpful assistant. Must use the random_number_tool to generate a random number. If tool ouput is empty, just say N/A",
    model="gpt-5-mini",
    tools=[random_number_agent.as_tool(tool_name="random_number_tool", tool_description="A tool that generates a random number.")],
)


result = Runner.run_sync(main_agent, "What is the random number?")
print(result.final_output)

```

Before fix: You will get "N/A" instead of a random number

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -13,7 +13,6 @@ from typing_extensions import NotRequired, TypeAlias, TypedDict
 from .agent_output import AgentOutputSchemaBase
 from .guardrail import InputGuardrail, OutputGuardrail
 from .handoffs import Handoff
-from .items import ItemHelpers
 from .logger import logger
 from .mcp import MCPUtil
 from .model_settings import ModelSettings
@@ -417,7 +416,7 @@ class Agent(AgentBase, Generic[TContext]):
             description_override=tool_description or "",
             is_enabled=is_enabled,
         )
-        async def run_agent(context: RunContextWrapper, input: str) -> str:
+        async def run_agent(context: RunContextWrapper, input: str) -> Any:
             from .run import DEFAULT_MAX_TURNS, Runner
 
             resolved_max_turns = max_turns if max_turns is not None else DEFAULT_MAX_TURNS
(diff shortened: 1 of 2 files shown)


seratch (approved):
Thank you!

xJkit:
Thank you!

Issue #2100, opened 2025-11-18
Track Agent and Model Identifiers in Usage Statistics


rvv-karma:
### Have you read the docs?
Yes — [Agents SDK documentation](https://openai.github.io/openai-agents-python/).

### Have you searched for related issues?
Yes; no existing issue addresses per-agent or per-model usage tracking.

***

### Motivation

In complex multi-agent systems, it is common for each agent to serve a distinct purpose and often use different model endpoints. For effective monitoring, accountability, and optimization, developers need granular visibility into which agent made which API request, and with which model. Currently, the usage reporting aggregates statistics without linking them to specific agents or models, making it hard to analyze efficiency or cost per agent.

***

### Current Implementation

Currently, usage statistics returned by the SDK do not include agent or model identifiers. Here’s a typical implementation and output:

#### Example Code

```python
from agents import Agent
from agents.extensions.models.litellm_model import LitellmModel

math_tutor_agent = Agent(
    name="Math Tutor",
    handoff_description="Specialist agent for math questions",
    instructions="Return just the answer",
    model=LitellmModel(model="openai/gpt-3.5-turbo", base_url="https://openrouter.ai/api/v1"),
)

science_tutor_agent = Agent(
    name="Science Tutor",
    handoff_description="Specialist agent for science questions",
    instructions="Return just the answer",
    model=LitellmModel(model="openai/gpt-3.5-turbo", base_url="https://openrouter.ai/api/v1"),
)

chat_agent = Agent(
    name="Chat Agent",
    instructions="Handoff the task to correct Agent",
    model=LitellmModel(model="openai/gpt-3.5-turbo", base_url="https://openrouter.ai/api/v1"),
    handoffs=[math_tutor_agent, science_tutor_agent]
)

from agents import Runner

result = await Runner.run(chat_agent, "2+2?")
print(result.final_output)
print(result.context_wrapper.usage)
```

#### Example Output

```
4
Usage(
    requests=2,
    input_tokens=115,
    output_tokens=14,
    total_tokens=129,
    request_usage_entries=[
        RequestUsage(
            input_tokens=65,
            output_tokens=13,
            total_tokens=78,
            input_tokens_details=InputTokensDetails(cached_tokens=0),
            output_tokens_details=OutputTokensDetails(reasoning_tokens=0)
        ),
        RequestUsage(
            input_tokens=50,
            output_tokens=1,
            total_tokens=51,
            input_tokens_details=InputTokensDetails(cached_tokens=0),
[...]

PR #2114, opened 2025-11-22
Add model_name, agent_name, and response_id to RequestUsage for better tracking

ihower:
Resolved: https://github.com/openai/openai-agents-python/issues/2100 

This PR adds three useful fields to `RequestUsage`: `model_name`, `agent_name`, and `response_id`.

These fields make it easier to:
- analyze usage across different models
- track usage by agent
- correlate each usage record with the exact response ID for inspection

## Note

`model_name` is taken from the `Model#model` attribute. All real subclasses of `Model` (such as `OpenAIResponsesModel`, `OpenAIChatCompletionsModel`, and `LitellmModel`) do initialize a `model` attribute.

However, the parent `Model` class does not define this field, which leads to type-checking errors. So this PR adds the attribute to the base class. To avoid breaking custom `Model` implementations (for example, `FakeModel` in tests that do not set a model name), the default value is an blank string.

### DIFF
--- a/src/agents/models/interface.py
+++ b/src/agents/models/interface.py
@@ -36,6 +36,9 @@ class ModelTracing(enum.Enum):
 class Model(abc.ABC):
     """The base interface for calling an LLM."""
 
+    # The model name. Subclasses can set this in __init__.
+    model: str = ""
+
     @abc.abstractmethod
     async def get_response(
         self,
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1473,7 +1473,12 @@ class AgentRunner:
                     usage=usage,
                     response_id=event.response.id,
                 )
-                context_wrapper.usage.add(usage)
+                context_wrapper.usage.add(
+                    usage,
+                    model_name=model.model,
+                    agent_name=agent.name,
+                    response_id=event.response.id,
+                )
 
             if isinstance(event, ResponseOutputItemDoneEvent):
                 output_item = event.item
@@ -1896,7 +1901,12 @@ class AgentRunner:
             prompt=prompt_config,
         )
 
-        context_wrapper.usage.add(new_response.usage)
+        context_wrapper.usage.add(
+            new_response.usage,
+            model_name=model.model,
+            agent_name=agent.name,
+            response_id=new_response.response_id,
+        )
 
         # If we have run hooks, or if the agent has hooks, we need to call them after the LLM call
         await asyncio.gather(
--- a/src/agents/usage.py
+++ b/src/agents/usage.py
@@ -50,6 +50,15 @@ class RequestUsage:
     output_tokens_details: OutputTokensDetails
     """Details about the output tokens for this individual request."""
 
+    model_name: str = ""
+    """The model name used for this request."""
+
+    agent_name: str = ""
+    """The agent name that made this request."""
+
+    response_id: str | None = None
+    """The response ID from the API for this request."""
+
 
 @dataclass
 class Usage:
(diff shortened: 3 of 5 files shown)


src/agents/models/interface.py
```diff
@@ -36,6 +36,9 @@ def include_data(self) -> bool:
 class Model(abc.ABC):
     """The base interface for calling an LLM."""
 
+    # The model name. Subclasses can set this in __init__.
+    model: str = ""
```
seratch, line 40: in fact, it seems available models have self.model but this could be potentially a breaking change for some custom models. it will require a minor version upgrade and release note annoucements

src/agents/usage.py
```diff
@@ -70,13 +81,22 @@ def __post_init__(self) -> None:
         if self.output_tokens_details.reasoning_tokens is None:
             self.output_tokens_details = OutputTokensDetails(reasoning_tokens=0)
 
-    def add(self, other: "Usage") -> None:
+    def add(
+        self,
+        other: Usage,
+        model_name: str,
```
seratch: these properties are in the request usage data, so i don't think these data should be passed this way. at least, we should avoid passing these in order to keep flexibility of future enhancement (meaning having additional-properties argument, which can accept more properties in the future)

ihower:
@seratch Thanks for the code review.

I’ve updated the PR so that all of these fields go into a single `metadata` dict. This should keep future enhancements flexible.

Regarding `model.model`, I’m not fully sure about the breaking-change risk. Since the base `Model` class now provides a default value `""`, any custom model that inherits from `Model` will have the attribute and won’t raise errors.

Are you concerned about custom models that don’t inherit from `Model` at all (basically duck typing)?

If we want to avoid touching the Model class to fully eliminate this risk, I can switch to a runtime check when calling add(), something like:

```python
if hasattr(model, "model") and model.model:
    metadata["model_name"] = model.model
```

I can update the PR if you prefer this approach. Just let me know. Thanks!

seratch:
There is no rush but once this project starts working on 0.7 series changes, resolving the conflicts would be appreciated

ihower:
@seratch Thanks! I’ve resolved the conflicts and rebased onto the latest main.

ihower:
Rebased again on latest main to resolve new conflicts.

src/agents/models/interface.py
seratch: Can we get rid of this property this time? When a developer uses a ModelProvider, this property could be absent, and it sounds surprisingly inconsistent.

src/agents/run.py
```diff
@@ -1886,7 +1893,14 @@ async def _get_new_response(
             prompt=prompt_config,
         )
 
-        context_wrapper.usage.add(new_response.usage)
+        context_wrapper.usage.add(
+            new_response.usage,
+            metadata={
+                "model_name": model.model,
```
seratch: As mentioned above, I'd like to avoid having model_name this time. Also, I've added `endpoint: str | None = None` (responses.create, responses.compact, or None) to TS SDK for supporting responses.compact API: https://github.com/openai/openai-agents-js/pull/785 We'll add the same to the Usage data class in Python SDK. So, for these SDKs' consistency plus better developer experience, I'd like to avoid dict[str, Any] here and simply having agent_name: str | None and response_id: str: None to request usage data.

ihower: I’ve removed the metadata dict and now put the fields directly on Usage.

I think `model_name` is the most important part here. The original issue was about tracking model cost. 
Since the base Model provides a default value of "", this should not be a breaking change. 
If a developer uses a custom provider and doesn’t set a model name, it should be easy to add once when they notice.  This is also consistent with the other built-in Model implementations in the SDK, which all provide a model name.

Thank you for reconsidering.

seratch:
I revisited this, and I still would like to avoid having `model` proerty in the `Model` class as it's quite confusing for other use cases. Let me close this one but please feel free to come up with a different approach for resolving the same issue.

Issue #2115, opened 2025-11-23
handoff()  helper crashes when passed a callable (lazy loader)


ohcedar:
helper crashes when passed a callable (lazy loader)
SDK Version: openai-agents==0.6.1

Summary
The handoff() helper function documentation states that the agent argument can be "The agent to handoff to, or a function that returns an agent." However, passing a function causes an AttributeError because the code attempts to access .name on the function object itself, rather than on the agent returned by the function.

Steps to Reproduce
```
from agents import Agent, handoff
def get_target_agent():
    return Agent(name="Target", instructions="You are a target.")
# This crashes
h = handoff(get_target_agent)
```
Traceback:
```
File ".../agents/handoffs/__init__.py", line 302, in handoff
    agent_name=agent.name,
               ^^^^^^^^^^
AttributeError: 'function' object has no attribute 'name'
```

Cause In  agents/handoffs/init.py, the handoff function implementation accesses agent.name at line 302 (in version 0.6.1) to set the agent_name field of the Handoff object.

```
return Handoff(
    # ...
    agent_name=agent.name,  # <--- BUG: Assumes 'agent' is an Agent instance
    # ...
)
```
If agent is a callable (as permitted by the type hint and docstring), this line fails.

Suggested Fix
The helper should check if agent is callable. If it is, it cannot determine the agent's name immediately. It should either:

Require tool_name_override and a new agent_name_override argument when a callable is used.
Or, accept that agent_name might be deferred or require the user to supply it explicitly.
Currently, the only workaround is to instantiate the Handoff class directly.

seratch:
Sorry for the confusing documentation. While the docs say callable is also supported, but it's not available at this moment.

ohcedar:
Understand it was an error, but I would like to see it as a valid pattern. Thanks.

PR #2116, opened 2025-11-23
docs: clarify handoff() docs, callable not supported

ihower:
Related issue: https://github.com/openai/openai-agents-python/issues/2115

The docs said that `handoff()` can accept a function that returns an agent, but this is not actually supported in the current code.

* The agent type hint does not include Callable.
* The runtime never calls a function to get an agent.
* There is no test that uses a function; all tests use Agent instances only.

This PR updates the docs to match the real behavior.

If callable support is needed, it will require more work and can be added in a separate PR in the future.

### DIFF
--- a/src/agents/handoffs/__init__.py
+++ b/src/agents/handoffs/__init__.py
@@ -208,7 +208,7 @@ def handoff(
     """Create a handoff from an agent.
 
     Args:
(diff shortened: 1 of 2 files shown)


seratch (approved):

seratch:
Thanks for sending this document update!

Issue #2020, opened 2025-10-31
Item with id `rs_` not found


shankar-dh:
### Please read this first
- [x] **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-python/)
- [x] **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug
In certain scenarios, the Agents SDK returns the following error:

**Example:**
```
openai.NotFoundError: Error code: 404 - {'error': {'message': "Item with id 'rs_3c4caff36e34c099006900f0e3c66481909c15ac262a15b060' not found.", 'type': 'invalid_request_error', 'param': 'input', 'code': None}}
```

**Configuration:**
- Store is not set to `False`
- Using default session management: `SQLiteSession`
- All items stored in the database are sent back to the agent for each turn
- Multi-agent system architecture with handoffs:
    - Agent 1: `gpt-5` with minimal reasoning (acts as the triage agent)
    - Agent 2: `gpt-5` with medium reasoning
    - Agent 3: `gpt-5` with low reasoning

**Issue:**
Agent 1 generates an empty reasoning item as shown below:
```
8|f00b2f3b-770b-4f40-9ca3-ae7700940a19_6555bacd-f286-42b3-9f25-e7a9ae94e3de|{"id":"rs_3c4caff36e34c099006900f0e3c66481909c15ac262a15b060","summary":[],"type":"reasoning"}|2025-10-28 16:36:16
```

After several steps (including outputs produced by Agent 2 and Agent 3), when Agent 1 regains control and any new message is passed, the error above is triggered.

### Debug information
- Agents SDK version: `0.2.8`
- Python version: Python 3.11.6

### Repro steps
While the code that triggers this error in our system involves multiple agents and handoffs, the following simplified code reproduces the same error:
```Python
import asyncio
import random

from agents import (
    Agent,
    ItemHelpers,
    Runner,
    enable_verbose_stdout_logging,
    function_tool,
)

enable_verbose_stdout_logging()


@function_tool
def how_many_jokes() -> int:
    """Return a random integer of jokes to tell between 1 and 10 (inclusive)."""
    return random.randint(1, 10)

ORIGINAL_RS_ID = "rs_07a11282cddc7074006905172505c88198bd41b4889a8d5649"

def mangle_rs_id(original_rs_id: str) -> str:
    return original_rs_id[:-2] + "ff"

async def main():
    agent = Agent(
        name="Joker",
[...]

PR #2117, opened 2025-11-23
 Enable previous_response_id chaining for internal calls on the first turn

ihower:
Resolved issue: https://github.com/openai/openai-agents-python/issues/2124

Relative Issue: [https://github.com/openai/openai-agents-python/issues/2020#issuecomment-3567846339](https://github.com/openai/openai-agents-python/issues/2020#issuecomment-3567846339) (reported in a comment)
Relative PR: [https://github.com/openai/openai-agents-python/pull/1827](https://github.com/openai/openai-agents-python/pull/1827)

This PR enables `previous_response_id` chaining during internal function calls inside the first turn, even though no real `previous_response_id` exists yet.

Only the first turn is special-cased; once the first real `response_id` is obtained, all subsequent turns behave normally.

## Background

Currently, `previous_response_id` chaining only enables when a real previous response ID is explicitly provided. This works fine for subsequent turns, but it has a limitation:

For the first turn of a brand-new conversation, there is no existing `previous_response_id`. As a result, there is no way to tell the SDK that internal function calls can use `previous_response_id` chaining.

## Solution

Introduce a special flag value:

```
previous_response_id = "" # or name it as any other string
```

Setting "" makes the internal function calls inside the first turn use proper response chaining.

## Example usage:

```python
previous_response_id = ""

while True:
    user_input = input()
    result = await Runner.run(
        agent,
        user_input, 
        previous_response_id=previous_response_id
    )
    previous_response_id = result.last_response_id
```

## Alternative solution

"" is just one idea. We can rename it to any other string.

If the empty string is undesirable, an explicit parameter could be added:

```python
previous_response_id = None

while True:
    user_input = input()
    result = await Runner.run(
        agent,
        user_input,
        previous_response_id = previous_response_id,
        enable_internal_previous_response_chaining=True,
    )    
    previous_response_id = result.last_response_id
```

However, this introduces an extra API parameter solely for this feature and feels a little verbose.  
I am open to switching to this approach if preferred.

### DIFF
--- a/docs/running_agents.md
+++ b/docs/running_agents.md
@@ -143,25 +143,16 @@ from openai import AsyncOpenAI
 client = AsyncOpenAI()
 
 async def main():
-    # Create a server-managed conversation
-    conversation = await client.conversations.create()
-    conv_id = conversation.id    
-
     agent = Agent(name="Assistant", instructions="Reply very concisely.")
 
-    # First turn
-    result1 = await Runner.run(agent, "What city is the Golden Gate Bridge in?", conversation_id=conv_id)
-    print(result1.final_output)
-    # San Francisco
+    # Create a server-managed conversation
+    conversation = await client.conversations.create()
+    conv_id = conversation.id
 
-    # Second turn reuses the same conversation_id
-    result2 = await Runner.run(
-        agent,
-        "What state is it in?",
-        conversation_id=conv_id,
-    )
-    print(result2.final_output)
-    # California
+    while True:
+        user_input = input("You: ")
+        result = await Runner.run(agent, user_input, conversation_id=conv_id)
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for sending this. Since I'm on PTOs this week, let me think about solutions for this next week.

ihower:
On second thought, "bootstrap" feels a bit odd, so I changed it to use an empty string "" instead.

seratch:
Neither "bootstrap" nor "" sounds intuitive to me. What do you think about `auto_previous_response_id=True`?

ihower:
@seratch Sure, `auto_previous_response_id` sounds good to me too. I have updated the code accordingly, and also revised the example to use a while loop so it better reflects a real multi-turn conversation.

Issue #1333, opened 2025-08-01
Support for Streaming Events from Long-Running Tools


noe-starlight:
Add support in the OpenAI Agents SDK for tools that:
	•	Run long-running operations (e.g. background tasks, indexing, data processing)
	•	Connect to WebSockets or other asynchronous event sources
And allow those tools to stream intermediate updates or status events back to the agent runtime via custom event channels.

**Motivation:**

In real-world agent applications, tools often perform actions that span longer than a few seconds or need to respond to external events (e.g., via WebSocket, SSE, or polling). The current SDK does not allow for easily streaming intermediate updates from tools. Developers are forced to block until a final result or create workarounds outside the agent runtime.

This limits the user experience and utility for:
	•	Background processing tools (e.g. book indexing, media analysis)
	•	Real-time systems (e.g. stock price streams, WebSocket bots)
	•	Long API workflows (e.g. transcription, document conversion)

Proposed API Behavior:

Allow tools to yield structured event objects (e.g., tool_stream_event, tool_background_task) that the agent runtime can surface to the user (through a developer-defined channel such as SSE or a UI event handler).

Example tool yield behavior:
```python
yield {"type": "tool_background_task", "tool_name": "book_indexer", "message": "Indexing book contents started"}

for i in range(10):
    yield {"type": "tool_stream_event", "tool_name": "book_indexer", "data": f"Indexing page {i+1} of 10"}
```

Developer stream output handler (e.g. in Flask):
```python
if tool_event.get("type") == "tool_stream_event":
    yield f'data: {json.dumps({"event": "tool_update", "tool_name": tool_event["tool_name"], "detail": tool_event["data"]})}\n\n'
elif tool_event.get("type") == "tool_background_task":
    yield f'data: {json.dumps({"event": "tool_background_task_started", "tool_name": tool_event["tool_name"], "detail": tool_event["message"]})}\n\n'
```

**Expected output**

```
event: tool_background_task_started
tool_name: book_indexer
detail: "Indexing book contents"

event: tool_update
tool_name: book_indexer
detail: "Indexing page 1 of 10"

event: tool_update
tool_name: book_indexer
detail: "Indexing page 2 of 10"
```

Benefits:
	•	Enhances interactivity of tools without needing to return blocking responses
	•	Improves UX in agent-hosted environments with real-time feedback
	•	Enables developers to build more robust and scalable agents with observable internal progress

seratch:
[...]

Issue #692, opened 2025-05-14
Feature Request: Support streaming tool call outputs


mikuh:
## Background

Currently, when an agent calls a tool, the tool must fully complete before any output is returned.
However, some tools internally call LLMs or perform long-running tasks, and users have to wait without seeing any intermediate results.

## Request

Please support **streaming tool outputs**, similar to model response streaming.
Allow tools to send incremental outputs while they are still running, so users can see progress in real time.

## Benefits

* Improve user experience by reducing perceived waiting time.
* Enable better interaction for tools involving LLM calls or large content generation.
* Make tool behavior consistent with LLM output streaming.

Feiyangdu12345:
Right now, users have no feedback during long tool runs, which leads to confusion and poor engagement. +1 from me!

lenage:
@seratch any update on this?

RussellLuo:
I have managed to write an extension library [openai-agents-ext](https://github.com/RussellLuo/openai-agents-ext) that supports stream events for tools, see https://github.com/openai/openai-agents-python/issues/1333#issuecomment-3409973617 for details.

seratch:
I've added on_stream handler option for agents as tools. If you need streaming updates from a tool, please consider running a sub-agent as a tool.

Issue #661, opened 2025-05-07
First-class streaming tool output


josephbharrison:
I’m working on a project experimenting with long-running tools and realtime updates in OpenAI agents and MCP.

While MCP supports notifications (for events like progress updates or changes to available tools/resources) and pagination (for fetching complete results in batches), these are not designed for streamed tool outputs — where a tool emits chunks incrementally while running.
What’s missing is a mechanism to surface these incremental outputs in real time, both to the agent for reasoning and to user interfaces for live display.

In my experimental extension (streamable-mcp-client), I’ve been working around this by emitting server notifications during tool execution. This allows updates to be:
- surfaced immediately as ResponseTextDeltaEvent so UIs (CLI, web) can show realtime progress
- appended to RunResultStreaming.new_items for potential agent awareness (in theory)
- stepped forward manually via Runner.continue_run to allow agent event handling

However, these streamed updates do not become part of agent reasoning or tool evaluation in openai-agents today.
The agent only “sees” a completed tool result at the end of the call.
Even if streamed as notifications/messages, incremental results are not surfaced into the agent’s context.

⸻

Why this matters

Many tools naturally produce progressive results:
- AI Agents (chaining of streaming agent responses as tools)
- Recursive planning and tool execution
- Large queries returning partial rows
- Multi-step pipelines
- User-interactive or streaming input/output

These use cases require not just pagination, but true streaming semantics where incremental chunks are emitted and can be ingested before the tool call completes.

⸻

Proposal

It would be ideal for MCP and its SDKs (such as openai-agents) to support this pattern natively:
- Allow tool implementations to yield streamed results incrementally (not just return final output)
- Define protocol messages for streamed tool result chunks (beyond notifications and pagination)
- Ensure agent frameworks can ingest streamed tool chunks as they arrive and optionally surface them into LLM context at agent reasoning steps

This would unlock realtime, incremental processing patterns and much more powerful recursive/planning use cases for agents.

⸻

Project / demo (experimental approach)

https://github.com/josephbharrison/streamable-mcp-client

josephbharrison:
Related: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/484

RussellLuo:
I have managed to write an extension library [openai-agents-ext](https://github.com/RussellLuo/openai-agents-ext) that supports stream events for tools, see https://github.com/openai/openai-agents-python/issues/1333#issuecomment-3409973617 for details.

dhpancor:
[...]

PR #2121, opened 2025-11-24
feat: add custom stream events support for tool functions

vincenzodomina:
This adds support to add custom streaming events from within tool functions to e.g. send progress update messages for long running tools without breaking changes.

- No behavior change for existing tools; the new emit_event() method is opt-in.
- Non-streamed runs simply no-op emit_event (no emitter is set).
- Minimal surface change; no new stream types introduced.

This addition solves issue #1333 and could be used for #692 and  #661 

### Changes:

- src/agents/stream_events.py
  - Extended RunItemStreamEvent.name to accept "tool_event" so we can reuse RunItemStreamEvent without new types.
- src/agents/run_context.py
  - Added an internal emitter and current agent fields to RunContextWrapper:
  - Added async def emit_event(self, event: Mapping[str, Any]) -> None:
- src/agents/run.py
    - After creating RunResultStreaming and context_wrapper, wired an async _emit that enqueues to
    - Kept context_wrapper._current_agent updated each turn.

### How to use (inside tools):

1. In any tool function (receives ToolContext, which inherits RunContextWrapper), you can now do:

```python
await ctx.emit_event({"type": "data-my_tool.step", "data": {"message": "Working..."}})
```

2. The event flows through the SDK’s stream as a RunItemStreamEvent with name "tool_event" and item.raw_item. You can than forward that event as is to your stream consumer.

3. From your stream consumer re-yield these events:

```python
result = Runner.run_streamed(...)
    if event.type == "run_item_stream_event" and event.name == "tool_event":
        yield event.item.raw_item
```

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -887,6 +887,13 @@ class AgentRunner:
             context_wrapper=context_wrapper,
         )
 
+        # Provide an emitter to the context so tools can stream custom events.
+        async def _emit(evt: StreamEvent) -> None:
+            streamed_result._event_queue.put_nowait(evt)
+
+        context_wrapper._emit_fn = _emit
+        context_wrapper._current_agent = starting_agent
+
         # Kick off the actual agent loop in the background and return the streamed result object.
         streamed_result._run_impl_task = asyncio.create_task(
             self._start_streaming(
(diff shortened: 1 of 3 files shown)


seratch:
You can do the same by starting a task (or thread) for the same operation using on_tool_start/on_tool_end: https://github.com/openai/openai-agents-python/blob/main/examples/basic/lifecycle_example.py We'd like to hold off adding this custom event for now.

dfliess:
hey @seratch  on_tool_start/on_tool_end are useful, but they don’t cover cases where a tool needs to emit multiple intermediate updates from inside the tool "body". External shared state or event bus may work , but make tools much more complex. Having an official way for tools to emit mid-execution events (like emit_event) would make DevX better and responsive UIs and long-running workflows much easier to build.

vincenzodomina:
Here is a good explanation and use case of this concept:

https://useworkflow.dev/docs/ai/streaming-updates-from-tools

This is how the Vercel AI sdk handles this

PR #2123, opened 2025-11-25
Updating tracing section to include Arize AX

PubliusAu:


seratch:
Thanks for sending this. We no longer add all the providers on README, so could you send a patch to https://github.com/openai/openai-agents-python/blob/main/docs/tracing.md instead?

PR #2131, opened 2025-11-27
Fix #2130 streaming hang when session.add_items raises an exception

### LINKED ISSUE #2130: Streaming hangs indefinitely when session.add_items() raises an exception
### Describe the bug
When using `Runner.run_streamed()` with a session, if `session.add_items()` raises an exception, the `stream_events()` async generator hangs forever instead of propagating the error or terminating.

### Debug information
```
openai                                      2.8.1
openai-agents                               0.6.1
```

```
Python 3.13.0
```

### Repro steps

MRE:

Run the following and the stream will hang...

```python
from agents import Agent, Runner, SQLiteSession, set_default_openai_client
import asyncio
import os

set_default_openai_client(
    AsyncOpenAI(
        base_url="https://eu.api.openai.com/v1", api_key=os.environ["OPENAI_API_KEY"]
    )
)

agent = Agent(
    name="Assistant",
)

session = SQLiteSession("conversation_123")


async def _fake_add_items(**kwargs):
    raise Exception("Simulated Exception in add_items")


session.add_items = _fake_add_items


async def main():
    message = "What is the meaning of life?"
    result = Runner.run_streamed(
        agent,
        input=message,
        session=session,
    )

        print(type(event), event)


if __name__ == "__main__":
    asyncio.run(main())
```

### Expected behavior
The exception should either:
 • Be raised to the consumer iterating over stream_events()
 • Be surfaced as an error event in the stream
 • Cause the stream to terminate cleanly

ihower:
Resolved: [2130](https://github.com/openai/openai-agents-python/issues/2130)

This PR fixes a bug where `Runner.run_streamed()` could hang forever if `session.add_items()` raises an exception during streaming.

## Root cause

When an exception happens in the outer error handling of `_start_streaming`, the completion sentinel is not sent. Because of that, the event queue never receives `QueueCompleteSentinel`, and `stream_events()` waits forever.

## Solution

This PR marks the stream as complete in the outer error handling when `is_complete` is still `False`, so the stream will not hang and the exception can be raised to the caller.

A regression test is added to make sure exceptions are properly raised instead of causing the stream to hang.

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1285,6 +1285,14 @@ class AgentRunner:
             if streamed_result.trace:
                 streamed_result.trace.finish(reset_current=True)
 
+            # Ensure QueueCompleteSentinel is always put in the queue when the stream ends,
+            # even if an exception occurs before the inner try/except block (e.g., in
+            # _save_result_to_session at the beginning). Without this, stream_events()
+            # would hang forever waiting for more items.
+            if not streamed_result.is_complete:
+                streamed_result.is_complete = True
+                streamed_result._event_queue.put_nowait(QueueCompleteSentinel())
+
     @classmethod
     async def _run_single_turn_streamed(
         cls,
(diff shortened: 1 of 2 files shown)


seratch (approved):
Looks good to me

PR #2134, opened 2025-11-28
feat: preserve logprobs from chat completions API in ModelResponse

JRMeyer:
The SDK already accepts `top_logprobs` in ModelSettings and passes it to the API, but the logprobs returned in the response were discarded during conversion. This change:

1. Adds an optional `logprobs` field to ModelResponse dataclass
2. Extracts logprobs from `choice.logprobs.content` in the chat completions model and includes them in the ModelResponse

This enables use cases like RLHF training, confidence scoring, and uncertainty estimation that require access to token-level log probabilities.

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -268,9 +268,12 @@ class ChatCmplStreamHandler:
                         type="response.content_part.added",
                         sequence_number=sequence_number.get_and_increment(),
                     )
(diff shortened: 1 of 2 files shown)


seratch:
Ah, my suggestion seems missing code format; can you run `make format` command?

JRMeyer:
> Ah, my suggestion seems missing code format; can you run `make format` command?

@seratch -- ci/cd now passing

PR #2136, opened 2025-11-30
Create SECURITY.md for security policy

hwq74295dz-byte:
Add a security policy document outlining supported versions and vulnerability reporting.


seratch:
We'll add this file when it's necessary

Issue #1179, opened 2025-07-18
Streaming: ResponseUsage validation error when token fields are None


kobol:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug
When using streaming responses, some models  may return usage fields (prompt_tokens, completion_tokens, total_tokens) as None  with chat/completion api endpoint. The current implementation passes these values directly to the ResponseUsage Pydantic model, which expects integers. This causes a validation error like:
`Traceback (most recent call last):
  File "/Users/xxx/PycharmProjects/mai-dxo/src/api/routes/chat.py", line 38, in generate_stream
    async for event in result.stream_events():
  File "/Users/xxx/PycharmProjects/mai-dxo/.venv/lib/python3.12/site-packages/agents/result.py", line 215, in stream_events
    raise self._stored_exception
  File "/Users/xxx/PycharmProjects/mai-dxo/.venv/lib/python3.12/site-packages/agents/run.py", line 723, in _start_streaming
    turn_result = await cls._run_single_turn_streamed(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/xxx/PycharmProjects/mai-dxo/.venv/lib/python3.12/site-packages/agents/run.py", line 867, in _run_single_turn_streamed
    async for event in model.stream_response(
  File "/Users/xxx/PycharmProjects/mai-dxo/.venv/lib/python3.12/site-packages/agents/models/openai_chatcompletions.py", line 169, in stream_response
    async for chunk in ChatCmplStreamHandler.handle_stream(response, stream):
  File "/Users/xxx/PycharmProjects/mai-dxo/.venv/lib/python3.12/site-packages/agents/models/chatcmpl_stream_handler.py", line 495, in handle_stream
    ResponseUsage(
  File "/Users/xxx/PycharmProjects/mai-dxo/.venv/lib/python3.12/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 3 validation errors for ResponseUsage
input_tokens
[...]

PR #2141, opened 2025-12-02
fix(usage): Normalize None token detail objects on Usage initialization

codefromthecrypt:
Extends #2034 to handle providers that return None for entire input_tokens_details and output_tokens_details objects (not just the fields within them). This affects non-streaming responses.

Related to #1179 (which fixed the streaming case).

Some providers like llama-stack return null for these optional fields in their JSON responses. The OpenAI SDK maps these to None in Python. Previously, passing None to the Usage constructor would fail Pydantic validation before `__post_init__` could normalize them.

This PR uses Pydantic's BeforeValidator to normalize None values at the field level, before Pydantic's type validation runs.


seratch:
It seems the tests are failing

codefromthecrypt:
sorry about that @seratch

seratch (approved):
LGTM

PR #2146, opened 2025-12-04
Store reference to original function when creating `FunctionTool`

varunursekar:
When `@function_tool`  is used, no reference to the original function is retained, which makes testing the function directly cumbersome. The original function is now just stored as an attribute of `FunctionTool`:

```python
@function_tool
def my_func(x: str, y: str) -> str:
  return x + y

# works like the original function
my_func.func('hello ', 'world!')
```

The alternative is to require users to define helper functions duplicating signatures and doc strings:
```python

# using double definitions instead
def _my_func(x: str, y: str) -> str:
  return x + y

@function_tool
def my_func(x: str, y: str) -> str:
  return _my_func(x, y)

```

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import functools
 import inspect
 import json
 import weakref
@@ -244,8 +243,10 @@ class FunctionTool:
         if self.strict_json_schema:
             self.params_json_schema = ensure_strict_json_schema(self.params_json_schema)
 
+        # Dress the FunctionTool object with the name and docstring of the wrapped function
         if self._func:
-            functools.update_wrapper(self, self._func)
+            self.__name__ = self._func.__name__
+            self.__doc__ = self._func.__doc__
 
     def __call__(self, *args, **kwargs):
         if not self._func:
--- a/tests/extensions/memory/test_advanced_sqlite_session.py
+++ b/tests/extensions/memory/test_advanced_sqlite_session.py
@@ -20,7 +20,7 @@ pytestmark = pytest.mark.asyncio
 
 
 @function_tool
-async def _test_tool(query: str) -> str:
+async def test_tool(query: str) -> str:
     """A test tool for testing tool call tracking."""
     return f"Tool result for: {query}"
 
@@ -28,7 +28,7 @@ async def _test_tool(query: str) -> str:
 @pytest.fixture
 def agent() -> Agent:
     """Fixture for a basic agent with a fake model."""
-    return Agent(name="test", model=FakeModel(), tools=[_test_tool])
+    return Agent(name="test", model=FakeModel(), tools=[test_tool])
 
 
 @pytest.fixture
@@ -961,7 +961,7 @@ async def test_tool_execution_integration(agent: Agent):
         [
             {  # type: ignore
                 "type": "function_call",
-                "name": "_test_tool",
+                "name": "test_tool",
                 "arguments": '{"query": "test query"}',
                 "call_id": "call_123",
             }
--- a/tests/test_function_tool.py
+++ b/tests/test_function_tool.py
@@ -188,6 +188,12 @@ async def test_complex_args_function():
         )
 
 
+def test_func_tool_name_doc_inheritance():
+    tool = function_tool(simple_function)
+    assert tool.__name__ == simple_function.__name__
+    assert tool.__doc__ == simple_function.__doc__
+
+
 def test_absent_func_tool():
     tool = function_tool(simple_function)
     kwargs = asdict(tool)


seratch (changes requested):
Thanks for sending this pull request. Ideally, enabling `my_func('hello ', 'world!')` is the best. I will also explore solutions for it. As for your changes, `make lint` seems to be failing. Can you resolve it?

varunursekar:
@seratch Thanks for the quick response.

To achieve the functionality you are describing we could adopt the following approach:

```python
import inspect
import functools


@dataclass
class FunctionTool:
    ....
    func: ToolFunction[...] | None = None

    def __post_init__(self):
        if self.func:
          functools.update_wrapper(self, self.func)

        if not self.func:
          raise FunctionNotSetError()
        return self.func(*args, **kwargs)

def func(x: str) -> str:
  """this is a test"""
    return x

func = FunctionTool(...,func=func)

sig = inspect.signature(func)
sig_func = inspect.signature(func.func)

assert sig == sig_func
assert func.__doc__ == func.func.__doc__
```

This will ensure that signatures and docs of the original function are retained. What do you think?

seratch:
Yeah you're right. I actually [did the same](https://github.com/slackapi/bolt-python/blob/v1.27.0/slack_bolt/app/app.py#L809-L850) for Slack's official SDKs before joining OpenAI :) 

If you could apply the approach rather than holding func object in FunctionTool instance, it should be better for dev experience. I don't think this approach could cause any side effects but we can verify if nothing is broken due to that.

varunursekar:
@seratch I implemented the solution I proposed. A few notes:

1. I explored the idea of passing in `func` as an init-only variable that is not stored as an attribute of the class. One way would be to override the `__init__` method of the dataclass - but I decided against this because it requires copying all arguments and their defaults in order, which seemed not great for maintainability. Another option would be to use dataclass `InitVars`. However, storing the function as a private attribute `_func` seemed simpler to me. 
2. While the `__call__` method of `FunctionTool` is sync, it technically also handles the `async` case since it will simply return a coroutine, which the user can await. 
3. Caveat to 2: The downside is that the user doesn't have a way of inspecting that the `FunctionTool` returns a coroutine, since `inspect.iscoroutinefunction` doesn't return `True` for async callable classes (only async functions) and `FunctionTool.__call__` is sync. One way around this would be to create a subclass `AsyncFunctionTool` that has an async `__call__` method instead. However, I don't think this should be too much of an issue since the behaviour should be apparent from the original function definition itself.

varunursekar:
Seems like [this test](https://github.com/openai/openai-agents-python/pull/2146/files#diff-6245fc850b276faf7fc65b7ae186419e1395ce40da90134ab2c6473dfb56679cR23) is unrelated but was still failing? Can you see any reason why my changes would cause this test to error? For some reason pytest interpreted `test_tool` as a fixture because the name began with `test` - I changed the name to `_test_tool` and this fixed the issue.

seratch:
```
file /home/runner/work/openai-agents-python/openai-agents-python/tests/extensions/memory/test_advanced_sqlite_session.py, line 22
  @function_tool
  async def test_tool(query: str) -> str:
      """A test tool for testing tool call tracking."""
      return f"Tool result for: {query}"
E       fixture 'query' not found
>       available fixtures: _class_scoped_runner, _function_scoped_runner, _module_scoped_runner, _package_scoped_runner, _session_scoped_runner, agent, anyio_backend, anyio_backend_name, anyio_backend_options, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, capteesys, class_mocker, clear_default_runner, clear_openai_settings, clear_span_processor, disable_real_model_clients, doctest_namespace, ensure_openai_api_key, event_loop_policy, free_tcp_port, free_tcp_port_factory, free_udp_port, free_udp_port_factory, mocker, module_mocker, monkeypatch, package_mocker, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, session_mocker, setup_span_processor, shutdown_trace_provider, snapshot_check, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, unused_tcp_port, unused_tcp_port_factory, unused_udp_port, unused_udp_port_factory, usage_data
>       use 'pytest --fixtures [testpath]' for help on them.
```

This looks like an issue with the combination of the changes here and pytest.fixture mechanism. While the failure affects only tests so far, I'd like to avoid ignoring the error because similar issues could arise with some other libraries.

varunursekar:
Yeah it seems like it's an issue with `functools.update_wrapper` - `pytest.fixture` also uses this and I suspect there is some bad interaction between the two. Will investigate.

seratch:
@varunursekar If you're still interested in working on this, we're happy to have your enhancement here!

varunursekar:
@seratch It seems like `wraps` was indeed having some side-effect on pytest collection. If we keep `__name__` and `__doc__`, pytest will still raise warnings because many of the tools used in the test have the prefix 'test' that is picked up by pytest. Arguably it may be good practice to change the names of those tools to something like `dummy_tool` to avoid this. If you prefer, I can just remove the `__name__` and `__doc__` dressing entirely.

PR #2153, opened 2025-12-04
v0.6.2

rm-openai:


seratch:
we already have this https://github.com/openai/openai-agents-python/pull/2113 but either is fine

PR #2156, opened 2025-12-05
Fix #2155 DeepSeek reasoning content handling for LiteLLM

### LINKED ISSUE #2155: LiteLLM DeepSeek thinking mode fails with missing `reasoning_conten` on tool calls
### Please read this first

- **Have you searched for related issues?** Yes

### Describe the bug
When running the LiteLLM example against DeepSeek’s thinking mode with tools([DeepSeek V3.2 Thinking mode](https://api-docs.deepseek.com/guides/thinking_mode)), the assistant tool-call message lacks the required `reasoning_content` field. DeepSeek responds with a 400 error: “Missing `reasoning_content` field in the assistant message …”.

### Debug information
- Agents SDK version: current main (after latest changes)
- Python version: Python 3.13

### Repro steps
```bash
uv run examples/model_providers/litellm_provider.py \
  --model deepseek/deepseek-reasoner \
  --api-key <deepseek_api_key>
```

The run fails after the tool call with a 400 complaining about missing reasoning_content.

### Expected behavior

Assistant tool-call turns should include reasoning_content so DeepSeek can continue thinking + tool calling and return a final answer without error.

Harrrryz:
Propagate `reasoning_content` onto assistant tool-call messages when using DeepSeek/thinking mode via LiteLLM. Extend the chat completion converter to optionally include reasoning content and enable it for DeepSeek. Added regression test covering reasoning content attachment.

- Fix #2155 

  - **Testing:**
  - `make format`
  - `make lint`
  - `make mypy`
  - `make tests`

### DIFF
--- a/examples/basic/stream_items.py
+++ b/examples/basic/stream_items.py
@@ -31,7 +31,7 @@ async def main():
             continue
         elif event.type == "run_item_stream_event":
             if event.item.type == "tool_call_item":
-                print("-- Tool was called")
+                print(f"-- Tool was called: {getattr(event.item.raw_item, 'name', 'Unknown Tool')}")
             elif event.item.type == "tool_call_output_item":
                 print(f"-- Tool output: {event.item.output}")
             elif event.item.type == "message_output_item":
@@ -47,7 +47,7 @@ if __name__ == "__main__":
 
     # === Run starting ===
     # Agent updated: Joker
-    # -- Tool was called
+    # -- Tool was called: how_many_jokes
     # -- Tool output: 4
     # -- Message output:
     #  Sure, here are four jokes for you:
--- a/examples/realtime/cli/demo.py
+++ b/examples/realtime/cli/demo.py
@@ -25,6 +25,7 @@ CHANNELS = 1
 ENERGY_THRESHOLD = 0.015  # RMS threshold for barge‑in while assistant is speaking
 PREBUFFER_CHUNKS = 3  # initial jitter buffer (~120ms with 40ms chunks)
 FADE_OUT_MS = 12  # short fade to avoid clicks when interrupting
+PLAYBACK_ECHO_MARGIN = 0.002  # extra energy above playback echo required to count as speech
 
 # Set up logging for OpenAI agents SDK
 # logging.basicConfig(
@@ -78,6 +79,7 @@ class NoUIDemo:
         self.fade_total_samples = 0
         self.fade_done_samples = 0
         self.fade_samples = int(SAMPLE_RATE * (FADE_OUT_MS / 1000.0))
+        self.playback_rms = 0.0  # smoothed playback energy to filter out echo
 
     def _output_callback(self, outdata, frames: int, time, status) -> None:
         """Callback for audio output - handles continuous audio stream from server."""
@@ -123,6 +125,7 @@ class NoUIDemo:
                 gain = 1.0 - (idx / float(self.fade_total_samples))
                 ramped = np.clip(src * gain, -32768.0, 32767.0).astype(np.int16)
                 outdata[samples_filled : samples_filled + n, 0] = ramped
+                self._update_playback_rms(ramped)
 
                 # Optionally report played bytes (ramped) to playback tracker
                 try:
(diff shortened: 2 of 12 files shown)


seratch:
Thanks for sending this PR! One thing I want to check is that if this customization could be done on the LiteLLM side, that'd be the most appropriate situation for us. Have you considered contributing to the LiteLLM project?

ihower:
This PR doesn’t cover the case where DeepSeek is accessed directly through its OpenAI-compatible endpoint.

For example:

```python
custom_client = AsyncOpenAI(
    base_url="https://api.deepseek.com",
    api_key=os.getenv("DEEPSEEK_API_KEY")
)

        model="deepseek-reasoner",
        openai_client=custom_client,
    )

agent = Agent(
    name="Assistant",
    model= model,
    ...
)
```

Since the SDK’s internal conversion flow here through `litellm -> chatcmpl -> raw item`, they actually share part of the same code path. It would be ideal to fix both cases together for consistency.

Also, I’m currently updating Gemini 3 Pro support in PR 2158, so there might be some overlapping changes or conflicts to be aware of.

Harrrryz:
> Thanks for sending this PR! One thing I want to check is that if this customization could be done on the LiteLLM side, that'd be the most appropriate situation for us. Have you considered contributing to the LiteLLM project?

Yes—looked into it. LiteLLM already returns reasoning_content on DeepSeek responses, but it doesn’t (and can’t, without keeping state) auto-inject that into subsequent assistant tool-call messages. To make it automatic upstream, they’d need an opt-in flag like “auto_include_reasoning_content” that replays the last reasoning text, or at least a warning/helper for DeepSeek + thinking + tools. I’ve drafted a LiteLLM docs update showing how to echo reasoning_content and linking to DeepSeek’s official guide; and already turned that into a PR.

Harrrryz:
> This PR doesn’t cover the case where DeepSeek is accessed directly through its OpenAI-compatible endpoint.
> 
> For example:
> 
> ```python
> custom_client = AsyncOpenAI(
>     base_url="https://api.deepseek.com",
>     api_key=os.getenv("DEEPSEEK_API_KEY")
> )
> 
> model = OpenAIChatCompletionsModel(
>         model="deepseek-reasoner",
>         openai_client=custom_client,
>     )
> 
> agent = Agent(
>     name="Assistant",
>     model= model,
>     ...
> )
> ```
> 
> Since the SDK’s internal conversion flow here through `litellm -> chatcmpl -> raw item`, they actually share part of the same code path. It would be ideal to fix both cases together for consistency.
> 
> Also, I’m currently updating Gemini 3 Pro support in PR 2158, so there might be some overlapping changes or conflicts to be aware of.

I picked this up and extended the fix to cover DeepSeek when it’s accessed via an OpenAI-compatible client (not just LiteLLM). Changes:

Add reasoning_content forwarding for DeepSeek in OpenAIChatCompletionsModel so tool-call messages include reasoning_content when model name/base_url indicates DeepSeek or thinking is enabled.
Regression tests to ensure reasoning_content is included for DeepSeek and not for non-DeepSeek models (tests/test_reasoning_content.py).

This should cover both LiteLLM and direct DeepSeek/OpenAI-compatible paths.

src/agents/extensions/models/litellm_model.py
```diff
@@ -436,6 +438,25 @@ async def _fetch_response(
         )
         return response, ret
 
+    def _should_include_reasoning_content(self, model_settings: ModelSettings) -> bool:
```
ihower, line 441: I think the logic here should be strictly restricted to DeepSeek models only to avoid potential regressions for other providers.

According to the [DeepSeek Thinking Mode docs](https://api-docs.deepseek.com/guides/thinking_mode), we should only return True in these two cases:

1. The model_name contains `deepseek-reasoner`.
or 
2. The model_name contains `deepseek` and the `thinking` parameter is explicitly set.

Right now your logic will enable this whenever a thinking parameter is passed for non-DeepSeek models, and also for `deepseek-chat` even when thinking is not set.

Harrrryz: > I think the logic here should be strictly restricted to DeepSeek models only to avoid potential regressions for other providers.
> 
> According to the [DeepSeek Thinking Mode docs](https://api-docs.deepseek.com/guides/thinking_mode), we should only return True in these two cases:
> 
>     1. The model_name contains `deepseek-reasoner`.
>        or
> 
>     2. The model_name contains `deepseek` and the `thinking` parameter is explicitly set.
> 
> 
> Right now your logic will enable this whenever a thinking parameter is passed for non-DeepSeek models, and also for `deepseek-chat` even when thinking is not set.

Adjusted reasoning_content forwarding per DeepSeek guidance: only for deepseek-reasoner, or any deepseek-*model when thinking is explicitly set. This applies to both LiteLLM and the OpenAI client paths. Added regression coverage (including deepseek-chat with/without thinking) and reran uv run pytest tests/test_reasoning_content.py tests/test_anthropic_thinking_blocks.py.

Issue #2151, opened 2025-12-04
nest handoff: Handoff is incompatible with server-managed conversation


ihower:
This issue originates from https://github.com/openai/openai-agents-python/pull/1996#issuecomment-3590458080

There are two concrete issues:

1. `remove_all_tools` conflicts with server-managed conversation (using `previous_response_id` or `conversation_id`).
2. “Nest handoff history by default” ([https://github.com/openai/openai-agents-python/pull/1996](https://github.com/openai/openai-agents-python/pull/1996)) also conflicts with server-managed conversation.

The root cause is that any internal modification of chat history makes it inconsistent with the server-managed conversation, leading to unexpected behavior or errors.

### Debug information

* Agents SDK version: v0.6.1
* Python version: 3.10

---

### Repro steps for `remove_all_tools`

```python
from agents import Agent, handoff, Runner, function_tool
from agents.extensions import handoff_filters
from openai import AsyncOpenAI

async def main():

    client = AsyncOpenAI()
    conversation = await client.conversations.create()
    conv_id = conversation.id   

    refund_agent = Agent(name="Refund agent")
    triage_agent = Agent(
        name="Triage agent",
        handoffs=[handoff(refund_agent, input_filter=handoff_filters.remove_all_tools)]
    )

    result = await Runner.run(
        triage_agent,
        "please transfer to refund.",
        conversation_id=conv_id
    )

    print(f"{result.last_agent.name}: {result.final_output}")


import asyncio
if __name__ == "__main__":
    asyncio.run(main())
```

This results in the following error:

```
openai.BadRequestError: Error code: 400 - {
  'error': {
    'message': 'No tool output found for function call call_Vqk7s2LCkloV0vViUrIKnV4V.',
    'type': 'invalid_request_error',
    'param': 'input',
    'code': None
  }
}
```

The cause is that `remove_all_tools` removes the tool-output message, but the server-side conversation history is not modified. This mismatch causes the API error.

Using `previous_response_id` produces the same error.

### Repro steps for “Nest handoff history”

```python
from agents import Agent, handoff, Runner, function_tool
from openai import AsyncOpenAI

async def main():

    client = AsyncOpenAI()
    conversation = await client.conversations.create()
    conv_id = conversation.id   

    refund_agent = Agent(name="Refund agent")
    triage_agent = Agent(
        name="Triage agent",
        handoffs=[handoff(refund_agent)]
    )

    result = await Runner.run(
        triage_agent,
[...]

PR #2157, opened 2025-12-05
fixing handoff history conflicts

jhills20:
- Prevented handoff input filters from running when conversation_id/previous_response_id are used
    and automatically disable history nesting in that mode to keep the SDK in sync with server-managed
    conversations (with a clear error/warning instead of silent API failures).
  - When nesting is enabled without server-managed conversations, the next agent now receives only the
    synthesized <CONVERSATION HISTORY> message (no duplicated raw items) while the full run history is
    still retained for results.
  - Added coverage for the new server-conversation rules and documented the limitation in docs/
    handoffs.md.


ihower:
Link to the issue for context: https://github.com/openai/openai-agents-python/issues/2151

Issue #2137, opened 2025-12-01
Missing thought_signature when using Gemini 3 pro preview


silvasw:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug
Gemini 3 Pro Preview (`google/gemini-3-pro-preview`) model requires` thought_signature` similar to response_id for subsequent calls. Using the Agent and Runner the `thought_signature` is not inject for next loop during call. I have tried to use RunHooks to receive full response from the model but unsuccessfully.

```
File "C:\Users\dom\anaconda3\envs\genai\Lib\site-packages\openai\_base_client.py", line 1594, in request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - [{'error': {'code': 400, 'message': 'Unable to submit request because function call `default_api:get_current_temperature` in the 2. content block is missing a `thought_signature`. Learn more: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures', 'status': 'INVALID_ARGUMENT'}}]
```

### Debug information
- Agents SDK version: 0.6.1
- Python version: 3.11.13
- OpenAI version: 2.8.1

### Repro steps
This snippet using plain OpenAI client returns the thouch_signarute but fails when use Agent and Runner

- OpenAI library:
```python
# Create an async-compatible OpenAI client pointing to GOOGLE's endpoint
# Set up the OpenAI client pointing to GOOGLE endpoint
client = OpenAI(
    api_key=creds,
    base_url=f"https://aiplatform.googleapis.com/v1/projects/{PROJECT_ID}/locations/global/endpoints/openapi",
)

prompt = """
If it's too hot or too cold in London, set the temperature to a comfortable level.
Make your own reasonable assumption for what 'comfortable' means and do not ask for clarification.
"""

messages = [
    {"role": "user", "content": prompt},
]

response1 = client.chat.completions.create(
    model=MODEL_NAME,
    messages=messages,
    tools=tools,
)

print(response1)
```

OUTPUT:
```
[...]

Issue #2109, opened 2025-11-20
Gemini 3 is not supported via LiteLLM


Nifury:
### Please read this first

- **Have you read the custom model provider docs, including the 'Common issues' section?** [Model provider docs](https://openai.github.io/openai-agents-python/models/#using-other-llm-providers)
- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the question
Can not use Gemini 3 model (`gemini/gemini-3-pro-preview`) via LiteLLM when there is a tool call, with or without `model_settings=ModelSettings(reasoning=Reasoning(effort="low"))`. 

### Debug information
- Agents SDK version: v0.6.1
- Python version: 3.13

### Repro steps
```python
from agents import Agent, ModelSettings, Runner, function_tool
from agents.extensions.models.litellm_model import LitellmModel
from openai.types.shared import Reasoning


@function_tool
def echo_tool(input: str) -> str:
    return input


model = LitellmModel(
    model="gemini/gemini-3-pro-preview",
    api_key="[API_KEY]",
)
agent = Agent(
    name="Assistant",
    instructions="You are a helpful assistant",
    model=model,
    tools=[echo_tool],
    # model_settings=ModelSettings(reasoning=Reasoning(effort="low")),
)
result = Runner.run_sync(
    agent, "Can you use the echo_tool to echo the message 'Hello, world!'?"
)
print(result.final_output)
```

It prints
```
Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:echo_tool` , position 2. please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.
```

ihower:
This will need a proper LiteLLM release that supports Gemini 3 Pro first.
LiteLLM has an early implementation here: https://github.com/BerriAI/litellm/pull/16719 but there is no PyPI release for it yet.

Once LiteLLM ships a release with this support, I can provide a hotfix PR as soon as possible.

kkwhale7:
https://github.com/BerriAI/litellm/issues/16780
I noticed that LiteLLM already appears to support Gemini 3

lawalkeyd:
Released in [1.80.5](https://github.com/BerriAI/litellm/releases/tag/v1.80.5-nightly)

ihower:
[...]

Issue #1485, opened 2025-08-15
Title: Handoff fails with Invalid 'input[1].id': '__fake_id__' when using non-OpenAI models and OpenAI tracing


123Mismail:
**Description:**
When using the OpenAI Agents SDK with **tracing enabled** (`set_tracing_export_api_key(...)`) and a **non-OpenAI model** (e.g., Gemini) for generation, `handoff` between agents fails with the following error:

```
openai.BadRequestError: Error code: 400 - {
    'error': {
        'message': "Invalid 'input[1].id': '__fake_id__'. Expected an ID that begins with 'fc'.",
        'type': 'invalid_request_error',
        'param': 'input[1].id',
        'code': 'invalid_value'
    }
}
```

---

**Expected Behavior:**

* Handoffs should work the same way as when tracing is **disabled**, even with non-OpenAI models.
* The tracing dashboard should either:

  * Gracefully handle the absence of an OpenAI `fc_` ID, **or**
  * Generate a synthetic trace ID to maintain the event graph.

---

**Actual Behavior:**

* When tracing is **disabled**, the handoff works fine with Gemini.
* When tracing is **enabled**, the SDK sends a placeholder `"__fake_id__"` to the OpenAI tracing API.
* The OpenAI API rejects it because it is not a valid `fc_` ID.

---

**Why This Matters:**

* Mixed-model workflows (e.g., using Gemini for generation + OpenAI tracing for workflow visibility) are currently **blocked**.
* Tools work fine in this setup — only handoffs break due to ID validation.
* This is inconsistent and limits multi-agent, multi-model orchestration.

---

**Minimal Reproducible Example:**

```python
from agents import Agent, OpenAIChatCompletionsModel, function_tool, Runner, AsyncOpenAI, set_tracing_export_api_key
import os, asyncio
from dotenv import load_dotenv

load_dotenv()

# OpenAI key only for tracing
OPENAI_API_KEY = os.getenv("OPENAI_API_KEY")
set_tracing_export_api_key(OPENAI_API_KEY)

# Non-OpenAI model client
external_client = AsyncOpenAI(
    api_key=os.getenv("GEMINI_API_KEY"),
    base_url=os.getenv("BASE_URL")
)

model = OpenAIChatCompletionsModel(
    model=os.getenv("MODEL_NAME"),
    openai_client=external_client
)

@function_tool
async def city_name_tool():
    return "Karachi"

handof_agent = Agent(
    "handof_agent",
[...]

Issue #270, opened 2025-03-20
Handoff does not work with Claude 3.7 Sonnet


nishant-satpathy:
I am attempting use `claude-3.7-sonnet` with the Agents SDK. I have it set up like this:
```
claude_client = AsyncOpenAI(
    base_url="https://api.anthropic.com/v1/", 
    api_key=os.getenv("ANTHROPIC_API_KEY")
)

manager_agent = Agent(
    name="Manager Agent",
    instructions="""You are a manager who will decide which sub agent to handoff the input to based on the user's input. 
    """,
    model=OpenAIChatCompletionsModel(model="claude-3-7-sonnet-20250219", openai_client=claude_client),
    handoffs=[agent_1, agent_2, agent_3, agent_4]
)
```
When I execute this, I get the following error:
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'input[1].id': '__fake_id__'. Expected an ID that begins with 'msg'.", 'type': 'invalid_request_error', 'param': 'input[1].id', 'code': 'invalid_value'}}
```

I noticed that there was some documentation around `__fake_id__` here:
```
FAKE_RESPONSES_ID = "__fake_id__"
"""This is a placeholder ID used to fill in the `id` field in Responses API related objects. It's
useful when you're creating Responses objects from non-Responses APIs, e.g. the OpenAI Chat
Completions API or other LLM providers.
"""
```

From the examples, it looks like we have to use `OpenAIChatCompletionsModel` for using an LLM not provided by OpenAI, so does this mean that handoffs are not possible with any non OpenAI LLM? 

Side Note: I noticed that there was a comment on issue [#120](https://github.com/openai/openai-agents-python/issues/120#issuecomment-2720548948) that mentions that if the model does not support tool calling, handoff calling is not supported, but claude-3.7 does support tool calling IIRC.

Would appreciate any insights on this, thanks!

rm-openai:
I tried the attached example on Anthropic, and it worked ok. I'm guessing yours is more complex - could you share a script please?
```
import asyncio
import os

from openai import AsyncOpenAI

from agents import (
    Agent,
    OpenAIChatCompletionsModel,
    Runner,
    enable_verbose_stdout_logging,
    function_tool,
    set_tracing_disabled,
)

enable_verbose_stdout_logging()

BASE_URL = os.getenv("EXAMPLE_BASE_URL") or ""
API_KEY = os.getenv("EXAMPLE_API_KEY") or ""
[...]

PR #2158, opened 2025-12-07
Gemini 3 Pro support and cross-model conversation compatibility

ihower:
Resolves: 

* https://github.com/openai/openai-agents-python/issues/2109 (LiteLLM integration)
* https://github.com/openai/openai-agents-python/issues/2137 (ChatCompletions endpoint)

## Summary

This PR does two main things:

1. Adds support for Gemini 3 Pro `thought_signatures` in function calling.
2. Enables cross-model conversations (OpenAI ↔ Gemini, Gemini ↔ Claude, etc.) by storing provider-specific metadata on raw items and passing the target model name into the conversion helpers.

The goal is to make different providers interoperable: allowing them to safely share the same `to_input_list()` items, while each provider only receives the metadata it understands.

## Examples

Besides unit tests, I performed live tests for all the following scenarios:

* LiteLLM + Gemini
  * [LiteLLM Gemini (non-streaming)](https://github.com/ihower/openai-agents-python-live-tests/blob/main/litellm-gemini3.py)
  * [LiteLLM Gemini (streaming)](https://github.com/ihower/openai-agents-python-live-tests/blob/main/litellm-gemini3-streaming.py)
  
* Gemini ChatCompletions (OpenAI-compatible endpoint)
  * [ChatCompletions Gemini (non-streaming)](https://github.com/ihower/openai-agents-python-live-tests/blob/main/chatcmpl-gemini3.py)
  * [ChatCompletions Gemini (streaming)](https://github.com/ihower/openai-agents-python-live-tests/blob/main/chatcmpl-gemini3-streaming.py)
  
* Cross-model conversations (same raw items handled by different models)
  * [Gemini 3 → GPT-5](https://github.com/ihower/openai-agents-python-live-tests/blob/main/crossmodel-gemini-to-openai.py)
  * [Gemini 3 → GPT-5 (streaming)](https://github.com/ihower/openai-agents-python-live-tests/blob/main/crossmodel-gemini-to-openai-streaming.py)
  * [Gemini 3 → Claude](https://github.com/ihower/openai-agents-python-live-tests/blob/main/crossmodel-gemini-to-claude.py)
  * [Claude → GPT-5](https://github.com/ihower/openai-agents-python-live-tests/blob/main/crossmodel-claude-to-openai.py)
  * [Claude → GPT-5 (streaming)](https://github.com/ihower/openai-agents-python-live-tests/blob/main/crossmodel-claude-to-openai-streaming.py)

* Handoffs (disabled `nest_handoff_history`)
  * [Handoff GPT-5 → Gemini](https://github.com/ihower/openai-agents-python-live-tests/blob/main/handoff-gpt5-to-gemini.py)
  * [Handoff Gemini → GPT-5](https://github.com/ihower/openai-agents-python-live-tests/blob/main/handoff-gemini-to-gpt5.py)

## 1. Gemini 3 Pro function calling (`thought_signatures`)

Gemini 3 Pro now requires a `thought_signature` attached to function call in the same turn. 
Docs: [https://ai.google.dev/gemini-api/docs/thought-signatures](https://ai.google.dev/gemini-api/docs/thought-signatures)

This PR supports both integration paths with non-streaming and streaming modes:

1. LiteLLM integration (requires upgrading LiteLLM to version 1.80.5 or later)
2. Google’s Gemini OpenAI-compatible API endpoint

The conversation flow is: LiteLLM ↔ ChatCompletions ↔ our raw items.

### LiteLLM layer

LiteLLM places Gemini’s `thought_signature` inside `provider_specific_fields`.

This PR handles the conversion between: 

LiteLLM’s `provider_specific_fields["thought_signature"]`
↔
Google ChatCompletions format `extra_content={"google": {"thought_signature": ...}}`
    
### ChatCompletions layer

This PR handles the conversion between:

Google ChatCompletions format `extra_content={"google": {"thought_signature": ...}}`
↔
our raw item’s internal new field `provider_data["thought_signature"]`

### Cleaning up LiteLLM’s `__thought__` suffix

LiteLLM adds a `__thought__` suffix to Gemini tool call ids (see:
https://github.com/BerriAI/litellm/pull/16895). This suffix is not needed since we
have `thought_signature` and it causes call_id validation problems when the items are passed to other models.

Therefore, this PR removes it.

## 2. Enables cross-model conversations

To support cross-model conversations, this PR introduces a new `provider_data` field
on raw response items. This field holds metadata not compatible with the OpenAI Responses API, allowing us to:

* Identify which provider produced each item.
* Decide which fields are safe to send to other providers.
* Keep OpenAI Responses API payloads clean and compatible.

For non–OpenAI Responses API models, we now store this into raw item:

```python
provider_data = {
    "model": ...,
    "response_id": ...,  # Previously discarded when using non-Resposnes API; now preserved for inspection and debugging.
    # other provider-specific metadata, e.g. Gemini's "thought_signature": ...
}
```

This design is like PydanticAI, which uses a similar structure. The difference: PydanticAI stores metadata for all models,
whereas this PR stores `provider_data` only for non-OpenAI providers.

With `provider_data` and the model name passed into the converters, agents can now safely switch models while reusing the same raw items from `to_input_list()`. This includes:

* Gemini ↔ OpenAI
* Claude ↔ OpenAI
* Gemini ↔ Claude

It also works with handoffs when `nest_handoff_history=False`.

### Implementation  Details

Because items in a conversation can come from different providers, and each provider has different requirements, this PR passes the target model name into several conversion helpers:

* `Converter.items_to_messages(..., model=...)`
* `LitellmConverter.convert_message_to_openai(..., model=...)`
* `ChatCmplStreamHandler.handle_stream(..., model=...)`
* `Converter.message_to_output_items(..., provider_data=...)`

This lets us branch on behavior for different providers in a controlled way and avoid regressions by handling provider-specific cases. This is especially important for reasoning models, where each provider handles encrypted tokens differently.

There are libraries like PydanticAI and LangChain define their own internal standard formats to enable cross-model conversations:

* https://ai.pydantic.dev/message-history/#accessing-messages-from-results
* https://blog.langchain.com/standard-message-content/

By contrast, LiteLLM has not fully abstracted away these differences. It focuses on making each model call work with provider-specific workarounds, without defining a normalized history format for cross-model conversations. Therefore, we need explicit model-aware at this layer to make cross-model possible. 

For example, when we store Claude's `thinking_blocks` signature inside our reasoning item's `encrypted_content` field, we also need to know that it came from a Claude model. Otherwise, we will send this Claude-only encrypted content to another provider, which cannot safely interpret it.

The guiding principle in this PR is to treat OpenAI Responses API items as the baseline format, and use `provider_data` to extend them with provider-specific metadata when needed.

For OpenAI Responses API:

When sending items to the OpenAI Responses API, we must not send provider-specific metadata or fake ids. 
This PR adds: `OpenAIResponsesModel._remove_openai_responses_api_incompatible_fields(...)` 

* Quickly returns the input unchanged if no item has `provider_data`.
* Otherwise, processes items:
  * Removes `id` when it equals `FAKE_RESPONSES_ID`.
  * Drops reasoning items that have `provider_data` (these are provider-specific).
  * Removes the `provider_data` field from all items.

This keeps the payload clean and compatible with the Responses API, even if the items previously flowed through non-OpenAI providers.

## Design notes: reasoning items vs provider_data

This PR does not introduce a separate reasoning item (e.g. Claude thinking_blocks does) for Gemini function call's `thought_signatures`. Instead it stores the signatures in `provider_data` on the function call item.

The main reasons:

* Gemini thought signatures are function-call-bound and have no summary text. They belong strictly to the function call. Turning them into a separate reasoning   item would add complexity without any benefit.
* Keeping the signature directly on the function call: Matches how Gemini emits its metadata and Keeps the data model simple.

This design is again similar to PydanticAI’s approach and also mirrors the underlying Gemini parts structure: signatures are attached to the parts they describe instead of creating an extra reasoning item with no text.

I also study at the Gemini API raw format, there are four raw part structure with thought_signature:

1. `functionCall: {...}` with `thought_signature: "xxx"` → handled in this PR: keep the thought_signature with the function call.
2. `text: "...."` with `thought_signature: "xxx"` → could attach to the output item (no extra reasoning item needed).
3. `text: ""` with `thought_signature: "xxx"`  → (empty text) this is the case where a standalone reasoning item makes sense.
4. `text: "summary..."` with `thought: true` → (this is thinking summary) this is another case where a standalone reasoning item make sense.

This PR implements case (1), which is sufficient for Gemini’s current function calling requirement. 
Other cases can be added later if needed.

----

This PR should have no side effects on projects that only use the OpenAI Responses API, and I believe it establishes a better groundwork for handling various provider-specific cases.

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -349,6 +349,9 @@ class OpenAIResponsesModel(Model):
         """
         Remove or transform input items that are incompatible with the OpenAI Responses API.
 
+        This data transformation does not always guarantee that items from other provider
+        interactions are accepted by the OpenAI Responses API.
+
         Only items with truthy provider_data are processed.
         This function handles the following incompatibilities:
         - provider_data: Removes fields specific to other providers (e.g., Gemini, Claude).


seratch:
Thanks for sending this PR! Overall, the design is clean and the code looks good to go. If anyone could try this branch out and share early feedback before releasing it, it would be greatly appreciated.

mbegur:
@ihower @seratch is this PR good to merge? Would love to get this in the next release of the package 🙏

seratch:
I am currently working on 0.6.4 release. This one can be included in 0.7.0 or later, so please wait a moment!

src/agents/extensions/models/litellm_model.py
```diff
@@ -436,6 +459,65 @@ async def _fetch_response(
...
+            # Check if this is an assistant message with tool calls
+            if message.get("role") == "assistant" and message.get("tool_calls"):
+                tool_calls = message.get("tool_calls", [])
+
+                for tool_call in tool_calls:  # type: ignore[attr-defined]
+                    if not isinstance(tool_call, dict):
+                        continue
+
+                    # Default to skip validator, overridden if valid thought signature exists
+                    tool_call["provider_specific_fields"] = {
```
markmcd, line 505: FYI - in the context of parallel tool-calls, this adds the dummy signature to *every* tool call returned. In the docs, we specify that a dummy signature is to be provided on the *first* tool call, however it is safe to apply on all of them so no need to change anything.

src/agents/models/openai_responses.py
```diff
@@ -342,6 +344,49 @@ async def _fetch_response(
         )
         return cast(Union[Response, AsyncStream[ResponseStreamEvent]], response)
 
+    def _remove_openai_responses_api_incompatible_fields(self, list_input: list[Any]) -> list[Any]:
```
seratch, line 348: @ihower Mixing OpenAI Responses API items and non-OpenAI Chat Completions items in the same conversation history is not currently supported. We recommend using only Chat Completions for this use case, and in practice, most users do not mix providers.

My question is: does this workaround actually work in real-world usage? In particular, the lack of valid item IDs seems problematic. Beyond that, there are likely many unexpected patterns, so I am not convinced we should simply accept items after removing id and provider_data. Given that, I think it may make more sense to reject these items (either via Responses API rejection or input validation on the SDK side) rather than trying to support this kind of workaround. Thoughts?

pyproject.toml
```diff
@@ -37,7 +37,7 @@ Repository = "https://github.com/openai/openai-agents-python"
 [project.optional-dependencies]
 voice = ["numpy>=2.2.0, <3; python_version>='3.10'", "websockets>=15.0, <16"]
 viz = ["graphviz>=0.17"]
-litellm = ["litellm>=1.67.4.post1, <2"]
+litellm = ["litellm>=1.80.8, <2"]
```
seratch, line 40: Is this upgrade required for this PR? If not, can you move this change and uv.lock changes to a different PR? Having uv.lock changes could block merging this PR due to non-essential conflicts

src/agents/models/openai_responses.py
ihower: I think supporting this is very useful in practice.

1. **Why this matters?**  
    Some real apps let users switch models inside one conversation. For example, Perplexity has a “pick a model” UX. In this SDK, handoffs also often mean switching to a different model. There are several issues requesting switching to non-OpenAI models before(when `nest_handoff_history=False`):
    
- [https://github.com/openai/openai-agents-python/issues/226](https://github.com/openai/openai-agents-python/issues/226)
- [https://github.com/openai/openai-agents-python/issues/237](https://github.com/openai/openai-agents-python/issues/237)
- [https://github.com/openai/openai-agents-python/issues/270](https://github.com/openai/openai-agents-python/issues/270)    
- [https://github.com/openai/openai-agents-python/issues/1485](https://github.com/openai/openai-agents-python/issues/1485)
    
This PR can address these needs together.

Also, frameworks like LangChain and PydanticAI treat cross-model history as a standard feature by defining an internal normalized message format.

2. **Does it work in real usage?**  

    Yes, it works for the common case. I included live test code in the PR description, and I tested mixed histories across Gemini, Claude, and OpenAI. Both streaming and non-streaming paths work.
    
I agree this may not support all patterns, especially when mixing provider-specific hosted tools. But for the typical case with local function calls, it should behave well.

3. **Why removing `__fake_id__` and `provider_data` is enough**  
    Because we do not define an internal standard format ahead of time (unlike LangChain or PydanticAI), the most compatible approach is: treat OpenAI Responses API items as the baseline, and only add extra fields when we pass through non-OpenAI providers.
    
- **About `id`**: `ResponseInputItemParam` does not require `id` as an input field. The only time we have a real, valid item id is when the item originally came from the Responses API. The SDK injected `__fake_id__` is just a placeholder, so removing it before calling the Responses API is safe and makes the payload valid.
    
- **About `provider_data`**: `provider_data` exists only to store provider specific data (i.e. non-OpenAI Responses API). The Responses API does not understand it, so stripping it restores compatibility. If a user never uses non-Responses providers, items never get `provider_data` in the first place, so nothing changes.
    
So I’m aiming to support cross-model conversation compatibility while keeping changes to the raw item format minimal, to avoid impacting developers who only use the Responses API: keep the baseline format unchanged for Responses-API only users, and only do compatibility cleanup for histories that passed through non-OpenAI providers.

pyproject.toml
ihower: Yes, this upgrade is required.

Gemini 3’s `thought_signature` is only available in LiteLLM ≥ 1.80.5 via LiteLLM's  `provider_specific_fields["thought_signature"]`. Older versions do not expose this data.

seratch: okay, thanks

src/agents/models/openai_responses.py
seratch: Thanks for the response. Your changes here should be okay for both many of relatively simple cross-provider use cases and OpenAI Responses API only use cases. That said, this project is not committed to fully support the cross-provider use cases. Having this extra logic is okay, but I'd like to have additional code comment clarifying this logic does not guarantee all the use cases across various providers work without issues. Also, as a maintainer, I will continue encouraging users not to mix those. I understand other frameworks may support such nicely, but our Responses API support won't pursue the same. For those use cases, we generally recommend using chat completions + litellm.

seratch: There is no rush at all but i would like to have some clarification on the expectation of cross-provider support at the code level. Once it's added, we can merge the changes here.

shreyansh26:
Any update on when this can be merged?

src/agents/models/openai_responses.py
```diff
@@ -342,6 +344,49 @@ async def _fetch_response(
...
+        """
+        Remove or transform input items that are incompatible with the OpenAI Responses API.
```
seratch, line 350: @ihower Can you add extra comment like this?
```suggestion
        Remove or transform input items that are incompatible with the OpenAI Responses API.
        This data transformation does not always guarantee that items from other provider interactions 
        are accepted by the OpenAI Responses API.
```

ihower: @seratch Sure, I’ve rebased this branch on main and added the clarifying comment.

edenreich:
Thanks for raising this point, what about updating the OpenAPI spec?

Do we have now this extra_content.google.though_signature?

Or this only fixes the Python SDK?

I have a Golang application that has a similar issue. I think we should adjust the spec too?

https://github.com/inference-gateway/inference-gateway/issues/229

ihower:
@edenreich This extra_content.google.thought_signature field is an extension provided by the Gemini API’s OpenAI-compatible Chat Completions endpoint, not part of the OpenAI spec.

See Google’s documentation here: https://ai.google.dev/gemini-api/docs/thought-signatures#openai

As the PR author, this PR does add support in the Python SDK for Google’s OpenAI-compatible Chat Completions extension. It does not modify the OpenAI OpenAPI specification, nor does it propose adding this field to the OpenAI spec.

edenreich:
> @edenreich This extra_content.google.thought_signature field is an extension provided by the Gemini API’s OpenAI-compatible Chat Completions endpoint, not part of the OpenAI spec.
> 
> See Google’s documentation here: https://ai.google.dev/gemini-api/docs/thought-signatures#openai
> 
> As the PR author, this PR does add support in the Python SDK for Google’s OpenAI-compatible Chat Completions extension. It does not modify the OpenAI OpenAPI specification, nor does it propose adding this field to the OpenAI spec.

Got it, thanks for clarifying.

My question isn’t really about this PR itself. The SDK change makes sense. I’m more thinking about the bigger picture. When providers add OpenAI compatible extensions that aren’t reflected anywhere in the OpenAPI spec, even as optional or vendor extensions, other SDKs and tools end up drifting or re implementing things differently.

It might be worth having some way to reflect this in the spec, even as an extension, so other clients can support it properly without guessing.

I opened a follow up issue here to discuss that more broadly and linked it to your change.

https://github.com/openai/openai-openapi/issues/517

Thanks!

ihower:
@edenreich My personal view is that OpenAI is no longer interested in continuing the Chat Completions API as a standard. The focus is now on the Responses API, and they has recently been involved in a standardization effort through https://www.openresponses.org/

Because of this, I see provider-specific Chat Completions extensions mainly as practical compatibility work, not something the OpenAI OpenAPI spec will integrate across different providers and standardize.

Just my two cents.

edenreich:
> @edenreich This extra_content.google.thought_signature field is an extension provided by the Gemini API’s OpenAI-compatible Chat Completions endpoint, not part of the OpenAI spec.
> 
> See Google’s documentation here: https://ai.google.dev/gemini-api/docs/thought-signatures#openai
> 
> As the PR author, this PR does add support in the Python SDK for Google’s OpenAI-compatible Chat Completions extension. It does not modify the OpenAI OpenAPI specification, nor does it propose adding this field to the OpenAI spec.

Cool, I didn't know about this standard, I'll definitely give it a look, thanks for sharing!

PR #2196, opened 2025-12-16
Add session customization params

danielquintas8:
feat: Add `session_limit` parameter to control conversation history retrieval from sessions and include new test.

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -2070,7 +2070,7 @@ class AgentRunner:
 
         # Get previous conversation history
         # Resolve session settings: session defaults + run config overrides
-        session_settings = session.session_settings or SessionSettings()
+        session_settings = getattr(session, "session_settings", None) or SessionSettings()
         if run_config.session_settings is not None:
             session_settings = session_settings.resolve(run_config.session_settings)
 


danielquintas8:
> Thanks for sending this pull request! Adding the limit option is reasonable, but we'd like to make the new option to be open to future enhancement. So can you create new `SessionSettings` class (in the same way with `ModelSettings`) and move the `limit` parameter as the first and only property in it?

Thanks for the feedback! I've refactored the implementation, following the same pattern as `ModelSettings`.

Changes made:
- Created `SessionSettings` dataclass with `limit` as the first property
- Added `session_settings` parameter to all session constructors and `from_url`/ `from_address` method
- Integrated `SessionSettings` into `RunConfig` for runtime overrides
- Added `resolve()` method for merging settings (base + override)
- Updated all 7 session implementations consistently

I believe this is extensible for future properties. For example, I'd like to add `branch_id` support for conversation branching in `AdvancedSQLiteSession` (and a potential `AdvancedSQLAlchemySession`). If you're open to it, I can also implement it since I need it for my use case anyway.

Let me know if it needs any adjustments

src/agents/memory/session.py
```diff
@@ -16,6 +17,7 @@ class Session(Protocol):
     """
 
     session_id: str
+    session_settings: SessionSettings
```
seratch: For backward compatibility, this property must be optional (plus the default value is None).

src/agents/memory/session.py
```diff
@@ -61,6 +63,7 @@ class SessionABC(ABC):
...
```
seratch: same as above

src/agents/run.py
```diff
@@ -1958,7 +1969,14 @@ async def _prepare_input_with_session(
...
 
         # Get previous conversation history
-        history = await session.get_items()
+        # Resolve session settings: session defaults + run config overrides
+        session_settings = session.session_settings
+        if run_config.session_settings is not None:
+            session_settings = session_settings.resolve(run_config.session_settings)
+
+        history = await session.get_items(
+            limit=session_settings.limit,
```
seratch: the limit could be absent (both intentionally and for backward-compatibility of custom session implementations); it's okay to create an empty SessionSettings above but checking if limit is None and call `get_items()` when limit is None.

danielquintas8: Good point! I've added the empty `SessionSettings()`fallback with a conditional check (I believe the conditional is only needed for backward compatibility with custom session implementations that may not have the `limit` argument on `get_items(...)`).

seratch:
the mypy errors look valid. you may want to add a simple utility function to safely resolve "session_limit" within the built-in sessions stores.

src/agents/extensions/memory/dapr_session.py
```diff
@@ -222,12 +234,14 @@ async def get_items(self, limit: int | None = None) -> list[TResponseInputItem]:
...
 
         Args:
-            limit: Maximum number of items to retrieve. If None, retrieves all items.
+            limit: Maximum number of items to retrieve. If None, uses session_settings.limit.
                    When specified, returns the latest N items in chronological order.
 
         Returns:
             List of input items representing the conversation history
         """
+        session_limit = self._get_session_limit(limit)
```
seratch: can you have a static function that takes limit: int and settings: SessionSettings | None instead?

danielquintas8: internally that method uses the static method on `SessionSettings`:

```python
@staticmethod
def get_limit(
    explicit_limit: int | None,
    settings: SessionSettings | None,
) -> int | None:
    """Safely resolve the effective limit for session operations."""
    if explicit_limit is not None:
        return explicit_limit
    if settings is not None:
        return settings.limit
    return None
```

You think it would make more sense to have that explicit?
Basically call that method directly instead of having the `SessionABC` `_get_session_limit` method?

seratch: i prefer having only static resolve_session_lit(...) method and removing the class method

src/agents/memory/session_settings.py
```diff
@@ -0,0 +1,62 @@
...
+
+    This class holds optional session configuration parameters that can be used
+    when interacting with session methods.
+    """
+
+    limit: int | None = None
+    """Maximum number of items to retrieve. If None, retrieves all items."""
+
+    @staticmethod
+    def get_limit(
```
seratch: no need to be part of this class; move this to top-level static function and name it resolve_session_limit

danielquintas8: I've moved it to be a top-level function in `session_settings.py` since it operates on `SessionSettings`. That said, it could also live in `session.py` if you prefer it closer to the session implementations...
Let me know if you'd like me to move it

src/agents/memory/session_settings.py
```diff
@@ -0,0 +1,62 @@
...
+        return replace(self, **changes)
+
+    def to_json_dict(self) -> dict[str, Any]:
+        """Convert settings to a JSON-serializable dictionary."""
+        dataclass_dict = dataclasses.asdict(self)
+
+        json_dict: dict[str, Any] = {}
+
+        for field_name, value in dataclass_dict.items():
+            if isinstance(value, BaseModel):
```
seratch: it seems this consideration is not necessary for this class

src/agents/memory/session_settings.py
```diff
@@ -0,0 +1,62 @@
...
+
+        changes = {
+            field.name: getattr(override, field.name)
+            for field in fields(self)
+            if getattr(override, field.name) is not None
+        }
+
+        return replace(self, **changes)
+
+    def to_json_dict(self) -> dict[str, Any]:
```
seratch: can you rename this to to_dict() ?

seratch:
Can you also resolve the conflicts?

src/agents/memory/session_settings.py
```diff
@@ -0,0 +1,52 @@
+"""Session configuration settings."""
+
+from __future__ import annotations
+
+import dataclasses
+from dataclasses import fields, replace
+from typing import Any
+
+from pydantic import BaseModel
```
seratch:
```suggestion
```

seratch:
I haven't checked what's wrong here, but the tests are still failing

danielquintas8:
> I haven't checked what's wrong here, but the tests are still failing

I forget to make it into Draft as I was planning to validate everything afterwards. It should work now

PR #2205, opened 2025-12-19
feat: Add extensible data payload to lifecycle hooks (#2180)

### LINKED ISSUE #2180: Add more arguments to agent/run hooks
We've been receiving many feature requests to add more arguments to the agent/run hooks. The examples are:
- turn_input to on_agent_start
- arguments to on_tool_start/on_tool_end

and perhaps other data would be needed in the future. However, the current design of the lifecyle hooks does not allow us to quickly add those because changing the method signature could be a breaking change to existing app that rely on the hooks feature.

To get rid of this restriction, we may want to consider accepting both the current set of the arguments and a new unified dict (or similar) object that contains all possible argument data.

see also: https://github.com/openai/openai-agents-js/pull/765

ruskaruma:
### Summary

This PR adds an optional `data: dict[str, Any] | None = None` parameter to lifecycle hooks (`on_agent_start`, `on_tool_start`, `on_tool_end`, `on_start`) so we can pass extra contextual information (like `turn_input` and tool arguments) without breaking existing hook implementations.

The intent is to make it possible to extend hook inputs going forward without having to keep changing method signatures.

**What I added:**

- An optional `data` parameter as the last argument to hook methods in `RunHooksBase` and `AgentHooksBase`
- Passing `turn_input` via the `data` dict in `on_agent_start` / `on_start`
- Passing tool arguments via the `data` dict in `on_tool_start` / `on_tool_end`
- A small helper using `inspect.signature()` so `data` is only passed to hooks that explicitly accept it (this keeps backward compatibility intact)
- Moved `_call_hook_with_data` into `lifecycle.py` to avoid a circular import between `run.py` and `_run_impl.py`

### Test Plan

- I added a new test `test_hooks_receive_turn_input_and_arguments()` to verify that hooks receive `turn_input` and tool arguments through the `data` payload
- I ran the full test suite locally: **1106 passed**, **3 skipped** (the skipped ones are expected Redis tests that require a real Redis server)
- Verified backward compatibility: existing hooks that don't define a `data` parameter continue to work unchanged
- All existing hook-related tests pass without modification

### Issue

Closes #2180

### Checks

### DIFF
--- a/src/agents/lifecycle.py
+++ b/src/agents/lifecycle.py
@@ -39,7 +39,7 @@ class RunHooksBase(Generic[TContext, TAgent]):
         self,
         context: RunContextWrapper[TContext],
         agent: TAgent,
-        data: dict[str, Any] | None = None,
+        data: Optional[dict[str, Any]] = None,
     ) -> None:
         """Called before the agent is invoked. Called each time the current agent changes."""
         pass
@@ -67,7 +67,7 @@ class RunHooksBase(Generic[TContext, TAgent]):
         context: RunContextWrapper[TContext],
         agent: TAgent,
         tool: Tool,
-        data: dict[str, Any] | None = None,
+        data: Optional[dict[str, Any]] = None,
     ) -> None:
         """Called immediately before a local tool is invoked."""
         pass
@@ -78,7 +78,7 @@ class RunHooksBase(Generic[TContext, TAgent]):
         agent: TAgent,
         tool: Tool,
         result: str,
-        data: dict[str, Any] | None = None,
+        data: Optional[dict[str, Any]] = None,
     ) -> None:
         """Called immediately after a local tool is invoked."""
         pass
@@ -95,7 +95,7 @@ class AgentHooksBase(Generic[TContext, TAgent]):
         self,
         context: RunContextWrapper[TContext],
         agent: TAgent,
-        data: dict[str, Any] | None = None,
+        data: Optional[dict[str, Any]] = None,
     ) -> None:
         """Called before the agent is invoked. Called each time the running agent is changed to this
         agent."""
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing this idea! However, this is not the thing I was thinking of. My plan is to have two patterns - 1. current set of arguments, 2. a single object (perhaps typed dict) containing all the arguments and bring no breaking changes. The second pattern is flexible enough for future enhancement and we don't need to worry about the order of the arguments.

ruskaruma:
Hey @seratch. Thanks for the feedback! I'd like to clarify the exact pattern you're looking for and I have a few questions here:

1. Should I keep the current method signatures completely unchanged and add a new optional parameter (e.g., `hook_data: Optional[TypedDict] = None`)?

2. What should the structure of the typed dict look like? For example:
   - For `on_agent_start`: `{"turn_input": list[TResponseInputItem]}`
   - For `on_tool_start`: `{"arguments": str}`

3. Should hooks be able to accept either:
   - The current signature (backward compatible)
   - The new signature with the typed dict
   - Or both at the same time?

4. Should I look at the JavaScript implementation (#765) as a reference for the pattern?

I want to make sure I implement exactly what you have in mind before proceeding so let me know.

ruskaruma:
and @seratch do let me know if I should either continue working on this or create a new PR with the implementation you've suggested with your plan.

seratch:
Thanks for the questions. I don't have the bandwidth to carefully think about the best design here at this moment, but we may want to introduce new hook type and accepting both legacy and new ones to pass hook functions may be the easiest (but I haven't verified if this idea works well).

It's totally fine if you explore implementation ideas on your side, but when it comes to the expectation of quick merge of the changes, we may prioritize other major enhancements such as human in the loop in 0.7 milestone.

ruskaruma:
thanks for the clarification @seratch.

i went through #2213 in detail, and the context-subclass approach (AgentHookContext) definitely looks cleaner and better aligned than the extensible payload pattern i explored here.

one thing i noticed while reading it is that this pattern currently covers agent hooks, but tool hooks still expose RunContextWrapper in the signature even though function tools already receive richer context at runtime. that feels like a natural follow up area if the same pattern is extended further.

for now, to avoid duplicating approaches, i’m happy to pause and close this pr. if it’s useful later, i’d be glad to help extend the context-based approach consistently to other hooks. thanks again for the guidance.

PR #2213, opened 2025-12-20
Add AgentHookContext with turn_input for agent hooks

ihower:
Based on the idea discussed in https://github.com/openai/openai-agents-python/issues/2180#issuecomment-3677509756, this PR proposes a concrete implementation.

It adds `turn_input` to the agent hook context and resolved:  https://github.com/openai/openai-agents-js/pull/765

The implementation introduces a new `AgentHookContext`.

It covers the following hooks:

* AgentHooks: `on_start` and `on_end`
* RunHooks: `on_agent_start` and `on_agent_end`

This change does not modify the hook parameter list, so it should not introduce breaking changes. 

`AgentHookContext` is simply a subclass extension of `RunContextWrapper`, similar to how `ToolContext` works for tool hooks.

If this extensibility pattern is acceptable, we can further extend other hook contexts via `RunContextWrapper` subclasses when needed, for example `ComputerActionContext` and `ShellActionContext`.

### DIFF
--- a/examples/basic/agent_lifecycle_example.py
+++ b/examples/basic/agent_lifecycle_example.py
@@ -4,7 +4,15 @@ from typing import Any
 
 from pydantic import BaseModel
 
-from agents import Agent, AgentHooks, RunContextWrapper, Runner, Tool, function_tool
+from agents import (
+    Agent,
+    AgentHookContext,
+    AgentHooks,
+    RunContextWrapper,
+    Runner,
+    Tool,
+    function_tool,
+)
 
 
 class CustomAgentHooks(AgentHooks):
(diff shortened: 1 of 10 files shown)


ruskaruma:
@ihower, i have a question here.

while reading through this, i noticed that on_tool_start / on_tool_end still expose RunContextWrapper in the hook signature, even though function tools already receive a ToolContext instance at runtime (there’s a small cast() workaround in lifecycle_example.py).

with the RunContextWrapper subclass pattern introduced here for agent hooks, would it make sense to apply the same approach to tool hooks in a follow up, to keep things consistent and avoid casting?

ihower:
@ruskaruma Good question. The reason `on_tool_start` / `on_tool_end` still expose `RunContextWrapper` in the hook signature, instead of switching to `ToolContext`, is that these hooks are also used by other hosted tools too, such as ComputerAction, LocalShellAction, ShellAction, and ApplyPatchAction. 

For these hosted action tools, they still pass RunContextWrapper rather than ToolContext (which is specifically designed for local function tools).

That’s also why I mentioned the idea of introducing dedicated context types for these action tools as well, for example `ComputerActionContext` or `ShellActionContext`. This is something I have in mind for a follow-up.

ruskaruma:
Totally makes sense. Thank you so much for the explanation!

PR #2215, opened 2025-12-20
Make example prompt consistent with Quick-start context

zhaoyanthu:
The goal of this doc is to provide an example where the triage agent can hand off the question to either a math tutor or a history tutor. However, the example prompt here has nothing to do with history or math; as a result, agent handoff will be skipped, which might confuse Agent SDK beginners. 

Also, in the "Put it all together" section, the prompt used is different from the example prompt here. 

This commit is updating the example prompt to be a history-related question, making it consistent with the section below.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -85,7 +85,7 @@ Let's check that the workflow runs and the triage agent correctly routes between
 from agents import Runner
 
 async def main():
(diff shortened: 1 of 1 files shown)


zhaoyanthu:
This is a minor change. Could a maintainer please approve the pending workflows so CI can run?

PR #2223, opened 2025-12-23
Feat: Add approval workflows for local MCP implementations

habema:
Adds approval workflows to `MCPServerStdio`, `MCPServerSse`, and `MCPServerStreamableHttp`, locally mimicking `HostedMCPTool`'s approval callbacks through the Responses API.

**Changes:**
- Added `on_approval_request` callback and `require_approval` policy to all three server classes
- Approval check happens before tool execution in `MCPUtil.invoke_mcp_tool`
- Reuses `McpApprovalRequest` and `MCPToolApprovalRequest`

**Features:**
- Supports `"always"`, `"never"`, or per-tool dict policies
- Defaults to `require_approval="never"`
- Same callback signature as `HostedMCPTool`, so one callback works for both

As per usual, I will create another PR for the docs. I also plan to tidy up `/examples/mcp`.

### DIFF
--- a/examples/mcp/approval_workflow_example/README.md
+++ b/examples/mcp/approval_workflow_example/README.md
@@ -1,0 +1,13 @@
+# MCP Approval Workflow Example
+
+This example demonstrates how to use the approval workflow with non-hosted MCP server implementations, with approval callbacks.
+
+Run the example via:
+
+```
+uv run python examples/mcp/approval_workflow_example/main.py
+```
+
+## Details
+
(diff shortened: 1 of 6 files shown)


seratch:
Thanks for sharing this. However, I am currently working on finalizing the human-in-the-loop enhancement, and it already includes this enhancement. So, we won't accept this PR.

habema:
@seratch I've checked out the HITL PR and have been waiting for it excitedly. My goal was to bridge this gap till that was ready, but I take it from your response that we should expect it soon! 
I also read that you are creating a new PR to polish the implementation and support additional patterns (and this is one of them). I'd be happy to contribute on the remaining work there.

PR #2239, opened 2025-12-26
Image response from an MCP server

### LINKED ISSUE #2148: Support returning multimodal content such as images from MCP tool calls.
I noticed that `FunctionTool` already supports returning multimodal results like `ToolOutputImage`, but currently `invoke_mcp_tool` in file [util.py](https://github.com/openai/openai-agents-python/blob/main/src/agents/mcp/util.py) converts all content into strings.

stneng:
#1898 adds support for image outputs in functions; the existing fallback logic in MCP tools below is outdated.
https://github.com/openai/openai-agents-python/blob/ba55bbd5961cd6e3f1003b188fa3aee74a73f097/src/agents/mcp/util.py#L217-L227

Resolves #2148 and https://community.openai.com/t/image-response-from-an-mcp-server-with-agents-sdk/1269441

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -248,7 +248,7 @@ class MCPUtil:
         current_span = get_current_span()
         if current_span:
             if isinstance(current_span.span_data, FunctionSpanData):
-                current_span.span_data.output = json.dumps(tool_output)
+                current_span.span_data.output = tool_output
                 current_span.span_data.mcp_data = {
                     "server": server.name,
                 }


seratch:
Thanks for sending this. Indeed, this could work well now. I will do thorough testing early next year.

seratch:
I ran a few local codex reviews and this feedback seems to be valid:

- [P1] Tracing output double-encodes MCP tool results — src/agents/mcp/util.py:250-252
    When recording MCP tool invocations, current_span.span_data.output is now set via json.dumps(tool_output) even though tool_output is already the final return value (string for structured
    content, or dict/list for text/image content). This extra dump double-encodes structured content (producing strings like "{\"foo\":1}") and converts dict/list outputs to JSON strings with
    different quoting than the returned value, causing trace snapshots to be incorrect for any MCP tool call. Span data should store the raw tool output instead of re-encoding it.

stneng:
Thanks for the feedback, just fixed it.

seratch:
will merge your change in https://github.com/openai/openai-agents-python/pull/2369

PR #2243, opened 2025-12-28
fix: #2204 Convert tool execution exception to runtime error

### LINKED ISSUE #2204: Not found tool execution brakes down the session.
### Describe the bug
For the OpenAI Realtime model, when the model hallucinates a tool that is not hooked into the tool list, and platforms SDK fails and raises an exception, causing teardown of the session.

Tool hallucination is very likely to happen in complex prompts, as well as OpenAI's text mode.

Made-up tool errors are also opened with other bugs but closed due to inactivity; #1671 

### Debug information
- Agents SDK version: v0.6.3
- Python version: Python 3.11

### Repro steps
- Add a system instruction for a made-up or non-existent tool. (also a tool name similar but not an exact one of the one added to the system)
- During the execution, when the tool fails exception is raised, the following log is raised, and the whole session is down.

``agents.exceptions.ModelBehaviorError: Tool xxxx not found.

### Expected behavior
As the tool fails, this error can be logged and propagated as a message or handler, allowing the session to continue with the error handled by the application.

aligokalppeker:
This PR fixes #2204 the issue I opened a while ago.

This fix changes "exception raise" to "runtime error message trigger" when an unregistered tool is triggered to be executed by the real-time model.

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -493,7 +493,12 @@ class RealtimeSession(RealtimeModelListener):
                 )
             )
         else:
-            raise ModelBehaviorError(f"Tool {event.name} not found")
+            await self._put_event(
+                RealtimeError(
+                    info=self._event_info,
+                    error={"message": f"Tool {event.name} not found"},
+                )
+            )
 
     @classmethod
     def _get_new_history(
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -8,7 +8,7 @@ from typing import Any, cast
 from typing_extensions import assert_never
 
 from ..agent import Agent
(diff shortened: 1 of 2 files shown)


seratch:
Overall this looks good to me. Can you resolve the CI errors?

aligokalppeker:
As far as i see, there was a linter error for unused import and i fixed it and pushed. The error in ``tests`` seem to be time out for the execution of the pipeline.

aligokalppeker:
Sorry about the unit tests, it is fixed now. First time opening a PR in the repo, and other failed tests in the suite confuse it.

seratch (approved):
Thank you! LGTM

PR #2245, opened 2025-12-28
fix: extract reasoning_content from summary for DeepSeek compatibility

yurekami:
## Summary
- Fixes issue #2155 where LiteLLM with DeepSeek in thinking mode fails with multi-turn conversations that include tool calls
- Extracts `reasoning_content` from the `summary` field of reasoning messages
- Includes the extracted `reasoning_content` in assistant messages that have tool calls

## Problem
When using the OpenAI Agents SDK with LiteLLM configured for DeepSeek's thinking mode (`deepseek/deepseek-reasoner`), multi-turn conversations with tool calls fail because:
1. DeepSeek API returns `reasoning_content` in the reasoning item's `summary` field
2. When the conversation history is sent back for subsequent turns, the `reasoning_content` field is missing from assistant messages
3. DeepSeek API requires this field in assistant messages for proper conversation continuity

## Solution
The fix tracks `pending_reasoning_content` across message conversion:
1. When a reasoning message is encountered, extract the reasoning text from the `summary` field (which contains `summary_text` items)
2. Store it as `pending_reasoning_content`
3. When an assistant message with tool calls is flushed, include the pending `reasoning_content`

## Test plan
- [x] Added test for extracting reasoning_content from summary and including in assistant message with tool calls
- [x] Added test for reasoning message without summary field (should not cause errors)
- [x] Added test to verify reasoning_content is only added to the immediately following assistant message

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -379,13 +379,22 @@ class Converter:
         result: list[ChatCompletionMessageParam] = []
         current_assistant_msg: ChatCompletionAssistantMessageParam | None = None
         pending_thinking_blocks: list[dict[str, str]] | None = None
+        # Track reasoning_content for DeepSeek reasoner models which require this
+        # field in assistant messages for multi-turn conversations with tool calls
+        pending_reasoning_content: str | None = None
 
         def flush_assistant_message() -> None:
-            nonlocal current_assistant_msg
+            nonlocal current_assistant_msg, pending_reasoning_content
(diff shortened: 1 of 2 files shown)


ihower:
There is another PR https://github.com/openai/openai-agents-python/pull/2156 that is also working on this issue. That PR intentionally limits the changes to DeepSeek, whereas this PR does not appear to check for that condition.

seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/2328 ; thanks again for taking the time on this

Issue #2246, opened 2025-12-30
ValueError: ContextVar token "was created in a different Context" when using concurrent Runner.run() calls with asyncio.gather


atpug22:
### Summary

When running multiple `Runner.run()` calls concurrently using `asyncio.gather()` (common in batch processing scenarios), Python raises a `ValueError` related to ContextVar token management. This is similar to the issue fixed in #538 for `run_streamed()`, but affects the regular synchronous `Runner.run()` pattern.

### Environment

- **openai-agents version:** 0.6.x (also affects forked version with Gemini support)
- **Python version:** 3.11+
- **OS:** Linux (production), macOS (development)

### Error Message

```
ValueError: <Token var=<ContextVar name='current_trace' default=None at 0x7f7a7f7279c0> at 0x7f7a73ea0200> was created in a different Context
```

### Stack Trace Location

```
File "agents/tracing/scope.py", line 49, in reset_current_trace
    _current_trace.reset(token)
```

### Reproduction

```python
import asyncio
from agents import Agent, Runner

async def process_batch(inputs: list[str]):
    """Process multiple inputs concurrently - triggers ContextVar error"""

    async def process_single(input_text: str):
        agent = Agent(
            name="processor",
            instructions="Process the input",
            model="gpt-4.1-mini",
        )
        result = await Runner.run(agent, input_text)
        return result.final_output

    # Concurrent execution triggers the error
    tasks = [process_single(inp) for inp in inputs]
    results = await asyncio.gather(*tasks)
    return results

# This will fail with ContextVar error when inputs > 1
asyncio.run(process_batch(["input1", "input2", "input3"]))
```

### Root Cause Analysis

The issue is in `agents/tracing/traces.py` and `agents/tracing/scope.py`:

1. When `Runner.run()` starts, it calls `trace.start(mark_as_current=True)` which does:
   ```python
   self._prev_context_token = Scope.set_current_trace(self)
   ```

2. When `Runner.run()` finishes, it calls `trace.finish(reset_current=True)` which does:
   ```python
   Scope.reset_current_trace(self._prev_context_token)
   ```

3. With concurrent `asyncio.gather()` tasks:
   - Task 1: `set_current_trace()` → gets token A
   - Task 2: `set_current_trace()` → gets token B
   - Task 3: `set_current_trace()` → gets token C
   - Task 1 finishes: `reset_current_trace(token A)` → changes ContextVar
[...]

PR #2255, opened 2026-01-03
fix: #2246 handle ContextVar token mismatch in concurrent execution

Ray0907:
### Summary

  Fix ValueError crash when using concurrent `Runner.run()` calls with `asyncio.gather()`.

  **Problem:**
  When multiple `Runner.run()` calls execute concurrently, the `ContextVar.reset(token)` operation can fail with:
  ValueError: <Token ...> was created in a different Context
  This forces users to disable tracing (`tracing_disabled=True`) as a workaround, losing all observability.

  **Solution:**
  Add defensive fallback mechanism:
  1. Store the previous trace/span value before setting a new one
  2. Catch `ValueError` when `reset()` fails due to context mismatch
  3. Fall back to `set(prev_value)` to restore the previous state
  4. Log a warning to inform users about the concurrent execution impact

  **Changes:**
  - `src/agents/tracing/scope.py`: Add fallback handling in `reset_current_span()` and `reset_current_trace()`
  - `src/agents/tracing/traces.py`: Store `_prev_trace` in `NoOpTrace` and `TraceImpl`
  - `src/agents/tracing/spans.py`: Store `_prev_span` in `NoOpSpan` and `SpanImpl`

  ### Test plan

  - All existing tracing tests pass (30 tests)
  - Verified type checking with mypy
  - The fix is defensive and backward compatible - normal execution path unchanged

  ### Issue number

  #2246

  ### Checks

### DIFF
--- a/src/agents/tracing/scope.py
+++ b/src/agents/tracing/scope.py
@@ -31,8 +31,25 @@ class Scope:
         return _current_span.set(span)
 
     @classmethod
-    def reset_current_span(cls, token: "contextvars.Token[Span[Any] | None]") -> None:
-        _current_span.reset(token)
+    def reset_current_span(
+        cls,
+        token: "contextvars.Token[Span[Any] | None]",
+        prev_span: "Span[Any] | None" = None,
+    ) -> None:
+        try:
+            _current_span.reset(token)
+        except ValueError:
+            # Token was created in a different Context. This can happen when multiple
+            # Runner.run() calls execute concurrently via asyncio.gather().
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for sharing this idea! I think https://github.com/openai/openai-agents-python/pull/2260 can resolve this in a stabler way, so we'll have the per-run config enhancement instead.

seratch:
Closing this in favor of #2260; thanks again for taking the time to send this!

PR #2260, opened 2026-01-05
Add per-run tracing API key support

rm-openai:
Currently tracing uses a fixed global API key for all tracing. This makes it so you can't have different ones with different tracing API keys, despite the fact that LLM calls can actually have different API keys.

This PR (codex written) fixes that. Approach: introduce a `TracingConfig` passed through RunConfig/voice configs into trace creation, storing a per-run `tracing_api_key` on traces/spans. The backend exporter now groups batches by each item’s key (falling back to the global key), so mixed runs use the right header. Docs and tests cover the new per-run key flow.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -38,6 +38,18 @@ from agents import set_tracing_export_api_key
 set_tracing_export_api_key("sk-...")
 ```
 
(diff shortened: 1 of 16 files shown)


rm-openai:
@codex review

PR #2275, opened 2026-01-08
feat(exceptions): add session exception classes for better error hand…

HemanthIITJ:
…ling

Add dedicated exception classes for session-related errors to provide clearer error messages and enable more specific exception handling.

- Add SessionError base class with session_id tracking
- Add SessionNotFoundError for missing session scenarios
- Add SessionSerializationError for serialization failures
- Export all new exceptions from package __init__.py
- Add comprehensive tests for all session exceptions


seratch:
Thanks for sending this PR! However, it seems these exceptions are not used by sessions store implementation. Thus, adding the new ones does not make any sense. If you have a certain use case with granular error types with sessions, please start a discussion issue first.

PR #2277, opened 2026-01-08
feat(items): add utility methods to ItemHelpers class

HemanthIITJ:
feat(items): add reusable utility methods to ItemHelpers


seratch:
Thanks, but we don't need this code change for now.

PR #2278, opened 2026-01-08
feat(run): add __repr__ and copy methods to RunConfig class

HemanthIITJ:


seratch:
We don't have immediate need of this custom code addition, so we won't have this.

PR #2279, opened 2026-01-08
feat(usage): add utility methods to Usage class for better usability

HemanthIITJ:
Add is_default, repr, copy to ModelSettings


seratch:
Thanks for sending this, but we really don't need this change.

PR #2283, opened 2026-01-08
fix(converter): validate JSON arguments to prevent Anthropic session …

HemanthIITJ:
Invalid JSON in tool call arguments can cause session failures with Anthropic models because LiteLLM attempts to parse malformed JSON strings into dictionaries, leading to runtime errors. This update adds JSON validation before parsing, safely replaces invalid inputs with an empty object ({}), logs a warning when malformed JSON is detected, and includes test coverage for both valid and invalid cases, improving overall system robustness and stability.

### DIFF
--- a/src/agents/realtime/_util.py
+++ b/src/agents/realtime/_util.py
@@ -2,8 +2,19 @@ from __future__ import annotations
 
 from .config import RealtimeAudioFormat
 
+PCM16_SAMPLE_RATE_HZ = 24_000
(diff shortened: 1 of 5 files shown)


seratch:
please take a look at the past discussion on this: https://github.com/openai/openai-agents-python/pull/2062 Your approach is exactly the same, and I still hesitate to introduce this additional overhead for all Chat Completions users just for a specific model's behavior.

seratch:
Thanks again for sending this, but as mentioned above, we don't go with this approach for now.

PR #2324, opened 2026-01-16
docs: add minimal Agent instantiation example to docstring

Arnav-Ajay:
### What this PR does
Adds a minimal, runnable example to the `Agent` docstring showing the recommended import path and basic instantiation.

### Why
New users often struggle with where to import `Agent` from and what the minimal setup looks like. This keeps the reference concise while improving clarity.

### Scope
- Docstring only
- No behavior changes


seratch:
Thanks for sending this patch. However, we don't need this small change for now.

Arnav-Ajay:
Thanks for the quick review and clarification!  
Understood — appreciate you taking the time to look at it. Happy to contribute in other areas if a more concrete docs or examples need comes up.

Issue #2285, opened 2026-01-09
will openai agent sdk introduce skills?


MuhammadHamidRazaAi:
hi i hope you are well i just want to ask about skills like claude introduce and claude agent sdk also has skills and this is game changer so will openai agent sdk introduce skills?

seratch:
Thanks for asking the question! This SDK itself does not directly load agent skills, but Codex already supports agent skills. The Codex SDK is currently available only in TypeScript; however, since Codex is a standalone binary, you can also invoke it from a Python script. When used as a tool within an agentic workflow built with this Python SDK, you can still take advantage of agent skills. We may be able to provide more seamless support in the future, but this is the current state.

seratch:
This one could be useful for codex use cases! https://github.com/openai/openai-agents-python/pull/2320

alvarovillalbaa:
I found many issuers asking for a native support of Skills in the OpenAI Agents SDK. I found your response to be a nice workaround, but eludes the question: will OpenAI Agents SDK support Skills natively in the near future? Many of us have to make technical decisions based on the response to that single question. I'm glad #2330 solves this, but we're still wondering whether support will be native to the framework.

seratch:
We've released Skills API and native support of loading skills in a container runtime yesterday. Here is a pull request adding the feature in this SDK: https://github.com/openai/openai-agents-python/pull/2469 I hope you will like it.

mefarion:
@seratch Have you published any documentation for this? It's a bit unclear to me how to use it at the moment.

seratch:
These resources should be useful:
- https://openai.github.io/openai-agents-python/tools/#hosted-container-shell-skills
- https://github.com/openai/openai-agents-python/blob/main/examples/tools/container_shell_inline_skill.py
- https://github.com/openai/openai-agents-python/blob/main/examples/tools/container_shell_skill_reference.py

mefarion:
@seratch Thanks! I explored this, however, although the `SKILL.md` format is great and we're adopting it, the container execution model doesn't fit all use cases. 

In particular, for our use case, the agent need to make authenticated calls using runtime session state that lives outside the container.

Maybe a `skills=...`  field directly on Agent (no in a container) would be a useful addition for these cases.

seratch:
I forgot to mention this, but there is a way to load local skills too: https://github.com/openai/openai-agents-python/blob/main/examples/tools/local_shell_skill.py This is still a tool, not in th main agent though.

[...]

PR #2330, opened 2026-01-18
Add agent skills pattern examples

ihower:
Resolved: https://github.com/openai/openai-agents-python/issues/2285

This PR adds example implementations for the agent skills pattern.

I think agent skills are more of a design pattern than a built-in feature. Only the `skills/` directory structure and the `SKILL.md` format are standardized. The developer prompt and how skill content is loaded are fully customizable.

The main idea is progressive disclosure. The base prompt stays small. Skills are listed first, and full skill content is loaded only when the agent needs it.

## Example details:

- `examples/agent_patterns/agent_skills.py` shows a minimal approach. It discovers available skills, renders a skills section in the developer prompt, and uses a `load_skill` tool to read a specific `SKILL.md` file.
- `examples/agent_patterns/agent_skills_shell.py` shows a shell-based approach. It uses `ShellTool` to read skill files and run skill scripts written in bash when needed.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -9,6 +9,7 @@ Check out a variety of sample implementations of the SDK in the examples section
 
     -   Deterministic workflows
     -   Agents as tools
+    -   Agent skills
     -   Parallel agent execution
     -   Conditional tool usage
     -   Input/output guardrails


ihower:
A related reference is this LangChain post, which also lists skills as a pattern: https://www.blog.langchain.com/choosing-the-right-multi-agent-architecture/

seratch:
Thanks for sharing this implementation idea! Indeed, your code looks functional and it loads only name and description in instructions, and then load the whole skill data including scripts when it can be used. However, I hesitate to add these examples because they don't demonstrate this SDK's built-in functionalities to load skills. 

Also, most of the non-coding models are not optimized for utilizing skills unlike codex models. That's why I added support via codex tool for loading skills.

Thus, your code examples are really helpful for others, but they are not the best suitable ones for this repo's examples.

ihower:
@seratch Thanks for the feedback.

If a skill only involves loading instructions with progressive disclosure and does not execute scripts, it can work without a coding model. In such cases, I think a non-coding model can still handle this pattern reasonably well.

Would it make sense to you to remove `agent_skills_shell.py` and keep only the minimal example `agent_skills.py`, with a note that advanced script execution should rely on the codex tool instead? My intent is to keep the example focused on illustrating the design pattern, rather than advanced capabilities or built-in features, similar to other agentic pattern examples(which do not always rely on special built-in features) and how the LangChain docs present their [skills examples](https://docs.langchain.com/oss/python/langchain/multi-agent/skills).

As a side note from this exploration, I noticed a related limitation in the current SDK. With function tools today, loaded skill content is always returned as tool output, not injected into the user prompt. In systems like Claude Code, function calls are also used to load skills, but the skill content is injected into user message, which may lead to better instruction-following behavior.

At the moment, the OpenAI Agents SDK does not provide a way for tools to dynamically inject user messages. This may be something that could be considered in the future.

seratch:
I may revisit my understanding later, but at the moment, I do not think loading skills via a function tool is an optimal solution, because it is not an instruction to the agent's model but rather tool output. 

If the skill metadata is embedded directly in the instructions, the model is more likely to behave predictably, with clear boundaries between enabled skills and user or tool inputs. That said, if we were to encourage this approach, the SDK should provide a built-in way to load skills when initializing an agent. 

The agent skills concept is still in its early stages, so for now, I would like to see how the ecosystem and usage patterns evolve.

ihower:
Understood, thanks for explaining your perspective.

ihower:
FYI, I've been tracking how Pydantic AI is developing their skills feature in their PR https://github.com/pydantic/pydantic-ai/pull/3780
Their implementation includes four tools:

* list_skills() - List all available skills
* load_skill(skill_name) - Load full instructions for a specific skill
* read_skill_resource(skill_name, resource_name) - Read additional resource files
* run_skill_script(skill_name, script_name, args) - Execute skill scripts

But the core concept remains the same as my example.
The key `load_skill` implementation can be found in file pydantic_ai_skills/toolset.py L363

https://github.com/pydantic/pydantic-ai/pull/3780/changes#diff-d5ab128a4f1335c59e54e2f7c05bb9b188791e1420b9dd98540cef13ee9c70ddR363 It returns the full skill content as a formatted string, and it is also placed in the tool result.

This suggests that tool-based progressive disclosure is becoming a common pattern in the agent ecosystem (e.g. Pydantic AI, LangChain). While may not optimal compared to injecting into user message, it's arguably the most universal and straightforward solution.

Updated: After looking more closely at `Codex`, I found that it also loads skill by reading file via Bash and returning the content as function call output. So in practice, I think there is no fundamental difference from the example implementation above.

seratch:
In a different discussion, I managed to organize my thoughts on skills as follows, so I am posting it here for reference:

My understanding of skills is that they are primarily a packaging mechanism for organizing knowledge and operations under a single name. Also, running a skill often looks like a coding-agent-style workflow: reading files, running commands, iterating, etc. So, using a coding agent (e.g., Codex) as a tool to execute an entire skill-driven process makes a lot of sense. By contrast, running the skill interaction inside the main agent loop can be inefficient (i.e., calling the Responses API with full context just to receive small operation requests like ls, grep, or cat), since it consumes turns in the loop.

If loading skills becomes a common norm among major agent app frameworks in the future, I don't think there is a strong reason to avoid supporting it, even though I may still believe many of the same outcomes can be achieved with well-designed sub-agents-as-tools. However, my current observation is that it still feels a bit early to bet on this idea as a core part of a general-purpose agent framework. It may very well be the right direction long-term, but rushing into it without battle-tested concepts could become technical debt.

Issue #2353, opened 2026-01-22
`StopAtTools` returns tool output as JSON string instead of Pydantic model (Agents SDK v0.4.0)


hedtsx:
## Describe the bug

When using `tool_use_behavior=StopAtTools(...)`, the run correctly stops at the tool call, but the tool result that is returned by `Runner.run(...)` is a **stringified JSON** rather than a Pydantic model instance (or a Python dict). The tool function itself returns a Pydantic `BaseModel`.

I expected the returned value to preserve the tool return type (e.g., `CustomModel(field1="test", field2=2)`), or at least return a Python dict `{ "field1": "test", "field2": 2 }`. Instead, the returned value is a JSON-like string (e.g., `"{"field1":"test","field2":2}"`), which breaks typed downstream handling.

## Debug information

* Agents SDK version: `0.4.0`
* Python version: `3.10`

## Repro steps

Minimal repro (representative; simplified):

```python
from agents import Agent, StopAtTools, Runner
from agents import function_tool
from pydantic import BaseModel

class CustomModel(BaseModel):
    field1: str
    field2: int

@function_tool
def mytool() -> CustomModel:
    return CustomModel(field1="test", field2=2)

myagent = Agent(
    name="myagent",
    instructions="call mytool",
    model="gpt-4o",
    tools=[mytool],
    tool_use_behavior=StopAtTools(stop_at_tool_names=["mytool"]),
)

result = await Runner.run(starting_agent=myagent, input="call the tool")
print(type(result.final_output), result.final_output)
```

## Expected behavior

One of the following (depending on intended SDK contract):

**Typed tool return preserved**: `result.final_output` is a `CustomModel` instance.

## Actual behavior

`result.final_output` is a **string** representing the object, e.g.:

* `"field1="test", field2=2"` (stringified representation)

ihower:
@hedtsx When using `tool_use_behavior` to stop early, you need to set `output_type` if you want to preserve the original return type. Otherwise, the SDK will stringify the tool output.

For example, update your code like this:

```py
myagent = Agent(
    name="myagent",
    instructions="call mytool",
    model="gpt-4o",
    tools=[mytool],
    tool_use_behavior=StopAtTools(stop_at_tool_names=["mytool"]),
    output_type=CustomModel,
)
````

This should give you the typed output behavior you want when the run stops at the tool.

PR #2354, opened 2026-01-22
docs: Clarify output_type requirement for tool_use_behavior

ihower:
Resolved: https://github.com/openai/openai-agents-python/issues/2353

Clarify that `output_type` can be set when using `tool_use_behavior` to stop early, otherwise the SDK converts tool output to a string.

Related code: https://github.com/openai/openai-agents-python/blob/main/src/agents/_run_impl.py#L391-L393

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -282,6 +282,8 @@ agent = Agent(
 )
 ```
 
+When using `tool_use_behavior` to stop early, you can set `output_type` to preserve the tool's return type. Without it, the SDK will convert the output to a string.
+
 !!! note
 
     To prevent infinite loops, the framework automatically resets `tool_choice` to "auto" after a tool call. This behavior is configurable via [`agent.reset_tool_choice`][agents.agent.Agent.reset_tool_choice]. The infinite loop is because tool results are sent to the LLM, which then generates another tool call because of `tool_choice`, ad infinitum.


ihower:
Just sharing an observation. The current behavior feels a bit somewhat odd to me:

1. If you want to preserve the function output type for `result.final_output`, you must provide `output_type`.
2. However, once `output_type` is set, it also enables structured output for the model.

The original intent seems to be ensuring that, regardless of whether the final output comes from normal LLM generation or from an early stop at a tool, the output type is always the specified `output_type`, providing consistency.

That said, the current implementation does not strictly enforce that the structured LLM output type and the function output type are aligned. For example, the following is currently possible:

```py
from agents import Agent, StopAtTools, Runner

@function_tool
def mytool() -> list:
    print("mytool called")
    return [1, 2, 3]

myagent = Agent(
    name="myagent",
    instructions="helpful agent",
    model="gpt-4.1-mini",
    tools=[mytool],
    tool_use_behavior=StopAtTools(stop_at_tool_names=["mytool"]),
    output_type=int,
)

result1 = Runner.run_sync(
    starting_agent=myagent,
    input="generate a random number without calling mytool",
)
print(type(result1.final_output), result1.final_output)  # this is int

result2 = Runner.run_sync(
    starting_agent=myagent,
    input="call mytool",
)
print(type(result2.final_output), result2.final_output)  # this is list
````

Overall, this leads to a use case that the SDK does not currently support well: a function early stop may be triggered after multiple turns where structured output is not really needed, but setting `output_type` forces structured output for all paths.
As a workaround, do not set `output_type` and do not use `final_output` to get function result. When an early stop happens, read the raw function output from `result.new_items[-1].output`.


While writing this up, I noticed that the docs I added in the commit are still not very clear in just a few lines. 
Writing a more detailed docs feels a bit too much. If there is a recommended best practice, I can update the docs later.
I’ll close this PR for now.

PR #2388, opened 2026-01-31
fix: persist MCP tool call outputs to session storage

### LINKED ISSUE #2384: Are local MCP tool calls intentionally excluded from session persistence?
### Description
I’m using the `openai-agents-python` SDK with local MCP servers (MCPServerStdio / MCPServerStreamableHttp) and had a question about how tool calls are persisted to session storage.

While instrumenting session persistence, I noticed a difference between function tools and local MCP tools, and wanted to check whether this behavior is intentional or a potential oversight.

### Environment
SDK: openai-agents-python @ e0bde88e28…
MCP usage: local MCP servers, not Hosted MCP (HostedMCPTool)
Session persistence: custom SessionStorage implementation

### Observed behavior
When logging the items passed from the SDK into session storage, I consistently see:

Persisted:
- User messages
- Assistant messages
- Handoff function calls (transfer_to_*)
- Function tool calls (@function_tool)

Missing:
- Local MCP tool calls (from MCP servers)

This behavior appears consistent across turns.

### Minimal pseudocode illustrating this

```
class CustomSessionStorage(SessionStorage):
    async def add_items(self, items: list[dict]) -> None:
        for item in items:
            log(item.get("type"), item)
```

Observed item types:
- message
- tool_call (function tools)
- handoff

Not observed
- Any item corresponding to local MCP tool calls

This logging reflects exactly what the SDK passes into SessionStorage; no filtering is applied on my side.

### Possible root cause (from code review)
From reviewing the SDK internals, it looks like: 
- Local MCP tools execute out-of-band in `turn_resolution.py` 
- They do not appear to create `ToolCallItem` objects?
As a result, they never reach `session_items_for_turn()` and are not included in `save_result_to_session()`
This may explain why function tools and MCP tools behave differently with respect to session persistence?

### Current workaround
I can observe local MCP tool activity via SSE streaming events (e.g. `tool_call.in_progress`), which provides real-time visibility. However, this doesn’t help with session replay or rehydration, since those events are not persisted.

### Questions
1. Is the exclusion of local MCP tool calls from session persistence intentional?
2. Are MCP tool calls expected to be non-replayable / non-persisted by design?
3. If not intentional, would the correct fix be:
 - creating ToolCallItems for MCP tools during turn resolution, or
 - extending session persistence to capture MCP tool execution metadata?
4. Is there a recommended pattern for persisting MCP tool calls if session replay is needed?

### Documentation context
The sessions documentation mentions that 

>tool calls, etc. are automatically stored

which led me to expect parity between function tools and MCP tools. I didn’t see any explicit exception noted for local MCP tools. 

Docs referenced: https://openai.github.io/openai-agents-python/

### Prior checks
I’ve read the Agents SDK documentation
I’ve searched existing GitHub issues and didn’t find one addressing local MCP tool persistence specifically

Happy to clarify or test changes if this turns out to be unintended behaviour. 

Cheers,
Kieran

saakshigupta2002:
When the model returns McpCall items with output data, the SDK was only creating ToolCallItem for the call but not ToolCallOutputItem for the result. This caused MCP tool call results to be missing from session persistence, making session replay incomplete.

Changes:
- Create ToolCallOutputItem for McpCall items that have output or error
- Add mcp_call -> mcp_call_output mapping to _TOOL_CALL_TO_OUTPUT_TYPE
- Update drop_orphan_function_calls to check 'id' field for mcp_call items (McpCall uses 'id' instead of 'call_id')
- Add comprehensive tests for MCP tool call output persistence

Fixes #2384

### DIFF
--- a/src/agents/run_internal/items.py
+++ b/src/agents/run_internal/items.py
@@ -23,6 +23,7 @@ _TOOL_CALL_TO_OUTPUT_TYPE: dict[str, str] = {
     "apply_patch_call": "apply_patch_call_output",
     "computer_call": "computer_call_output",
     "local_shell_call": "local_shell_call_output",
+    "mcp_call": "mcp_call_output",
 }
 
 __all__ = [
@@ -68,7 +69,8 @@ def drop_orphan_function_calls(items: list[TResponseInputItem]) -> list[TRespons
         if output_type is None:
             filtered.append(entry)
             continue
-        call_id = entry.get("call_id")
+        # Check call_id first, then fall back to id (used by mcp_call)
+        call_id = entry.get("call_id") or entry.get("id")
         if isinstance(call_id, str) and call_id in completed_call_ids.get(output_type, set()):
             filtered.append(entry)
     return filtered
--- a/src/agents/run_internal/turn_resolution.py
+++ b/src/agents/run_internal/turn_resolution.py
@@ -1299,6 +1299,26 @@ def process_model_response(
                     break
             items.append(ToolCallItem(raw_item=output, agent=agent, description=_mcp_description))
             tools_used.append("mcp")
+
+            # Create a ToolCallOutputItem for MCP calls that have completed with output.
+            # This ensures MCP tool call results are persisted to session storage for replay.
+            if output.output is not None or output.error is not None:
+                # Build an MCP call output item for session persistence.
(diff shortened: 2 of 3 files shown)


k1eran:
@saakshigupta2002  @seratch Ignore comment I had here earlier; it’s no longer relevant.

PR #2394, opened 2026-02-02
Add CRLF support for apply_diff

gustavz:
@seratch PR to make apply_diff support both CRLF and LF.
To make use of it it is required to read the input by bytes not text (e.g. Path.read_bytes vs Path.read_text), so we could think about either updating the example WorkspaceEditor to read bytes instead of text, or add another editor editor that does it.


gustavz:
@codex review

gustavz:
@codex review

PR #2399, opened 2026-02-03
fix: remove erroneous isinstance(agent, Tool) check in visualization

### LINKED ISSUE #2397: draw_graph() fails with TypeError: "unhashable type: 'Agent'" and isinstance error on Python 3.12+
## Bug Description
`draw_graph()` from `agents.extensions.visualization` fails on Python 3.12+ due to two issues:

1. Using `isinstance()` with generic type `Tool[...]`
2. Using `Agent` objects in a `set()` (agents are not hashable)

## Environment
- **Python version:** 3.12.8
- **openai-agents version:** 0.7.0
- **OS:** Linux (MX Linux)

## Steps to Reproduce
```python
from agents import Agent, OpenAIChatCompletionsModel
from agents.extensions.visualization import draw_graph

ai_client = AsyncOpenAI(api_key="test", base_url="http://localhost:1234/v1/")
my_model = OpenAIChatCompletionsModel(model="test-model", openai_client=ai_client)

my_agent = Agent(
    name="Test Agent",
    instructions="Test",
    model=my_model,
)

draw_graph(my_agent)  # Fails here
```

## Error Traceback
```
TypeError: Subscripted generics cannot be used with class and instance checks
  File .../agents/extensions/visualization.py:142
    if not agent.handoffs and not isinstance(agent, Tool):
```

## Root Cause
In `agents/extensions/visualization.py` line 142:
- `isinstance(agent, Tool)` fails because `Tool` is a generic type
- `visited = set()` fails because `Agent` objects are not hashable

## Proposed Fix
Replace the `isinstance` check with a class name check and use a list instead of set:
```python
def patched_get_all_edges(agent, parent=None, visited=None):
    if visited is None:
        visited = []  # Use list instead of set
    parts = []
    
    # Check by id instead of object equality
    agent_ids = [id(a) for a in visited]
    if id(agent) in agent_ids:
        return ""
    
    visited.append(agent)
    
    for handoff in agent.handoffs:
        if parent:
            parts.append(f'"{parent.name}" -> "{agent.name}";')
        parts.append(f'"{agent.name}" -> "{handoff.name}";')
        parts.append(patched_get_all_edges(handoff, agent, visited))
    
    # Use class name check instead of isinstance
    is_tool = "Tool" in str(type(agent))
    if not agent.handoffs and not is_tool:
        parts.append(f'"{agent.name}" -> "__end__";')
    
    return "".join(parts)
```

## Workaround
Users can patch the function themselves until this is fixed in the library.

leesta24:
## Summary

Fixes #2397 

This PR fixes a `TypeError` that occurs on Python 3.12+ when calling `draw_graph()`:

```
TypeError: Subscripted generics cannot be used with class and instance checks
```

## Root Cause

The issue is on line 142 of `visualization.py`:

```python
if not agent.handoffs and not isinstance(agent, Tool):  # type: ignore
```

This code has two problems:

1. **Runtime error on Python 3.12+**: `Tool` is a `Union` type alias containing generic types (like `ComputerTool[Any]`). In Python 3.12+, using `isinstance()` with such types raises a `TypeError`.

2. **Logically incorrect**: The `agent` parameter is typed as `Agent`, which can never be an instance of `Tool` (a completely different type). This check always evaluated to `False` (when it didn't error), making it a no-op.

## Changes

- **Removed** the unnecessary `isinstance(agent, Tool)` check
- **Removed** the unused `from agents.tool import Tool` import
- **Added** two new test cases using real `Agent` objects (not mocks) to ensure the fix works and prevent regression

## Testing

Added tests:
- `test_draw_graph_with_real_agent_no_handoffs`: Tests agent without handoffs connects to `__end__`
- `test_draw_graph_with_real_agent_with_handoffs`: Tests parent/child agent graph generation

These tests use real `Agent` objects rather than mocks, which would have caught this bug earlier.

### DIFF
--- a/src/agents/extensions/visualization.py
+++ b/src/agents/extensions/visualization.py
@@ -4,7 +4,6 @@ import graphviz  # type: ignore
 
 from agents import Agent
 from agents.handoffs import Handoff
-from agents.tool import Tool
 
 
 def get_main_graph(agent: Agent) -> str:
@@ -139,7 +138,7 @@ def get_all_edges(
             "{agent.name}" -> "{handoff.name}";""")
             parts.append(get_all_edges(handoff, agent, visited))
 
-    if not agent.handoffs and not isinstance(agent, Tool):  # type: ignore
+    if not agent.handoffs:
         parts.append(f'"{agent.name}" -> "__end__";')
 
     return "".join(parts)
--- a/tests/test_visualization.py
+++ b/tests/test_visualization.py
@@ -179,3 +179,43 @@ def test_cycle_detection():
     assert nodes.count('"B" [label="B"') == 1
     assert '"A" -> "B"' in edges
     assert '"B" -> "A"' in edges
+
+
+def test_draw_graph_with_real_agent_no_handoffs():
+    """Test that draw_graph works with a real Agent object without handoffs.
+
+    This test ensures that the visualization code does not use isinstance()
+    with generic types (like Tool), which would fail on Python 3.12+.
+    See: https://github.com/openai/openai-agents-python/issues/2397
+    """
+    agent = Agent(name="TestAgent", instructions="Test instructions")
+
+    # This should not raise TypeError on Python 3.12+
+    graph = draw_graph(agent)
+
+    assert isinstance(graph, graphviz.Source)
+    assert '"TestAgent"' in graph.source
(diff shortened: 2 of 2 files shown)


seratch (approved):
Thank you!

PR #2428, opened 2026-02-06
Add `Annotated[T, Field(...)]` syntax for tool arguments (`@function_tool`)

georg-wolflein:
### Summary

This PR adds support for **`Annotated[T, Field(...)]`** on tool parameters in addition to the existing **`= Field(...)`** syntax. The original `= Field(...)` support was introduced in #1124.

**Why add this variant?**

1. **Pydantic supports it.** Pydantic allows `Field` either as the default (`x: int = Field(ge=1)`) or inside `Annotated` (`x: Annotated[int, Field(ge=1)]`). Supporting both in `function_schema` keeps the SDK aligned with how users define constrained fields in Pydantic models.

2. **Parameter ordering in Python.** A parameter like `x: int = Field(..., ge=1)` still has a default value as far as Python is concerned (the default is the `Field(...)` object). So you cannot put another parameter without a default after it — that would be a SyntaxError:

   ```python
   # SyntaxError: parameter without default follows parameter with default
   @function_tool
   def bad(
       positive_number: int = Field(..., ge=1),
       another_arg: str,
   ) -> str:
       ...
   ```

   With `Annotated`, the constrained parameter has no default in the signature, so you can have multiple required arguments (some with `Field` constraints) in any order:

   ```python
   @function_tool
   def good(
       positive_number: Annotated[int, Field(..., ge=1)],
       another_arg: str,
   ) -> str:
       ...
   ```

**Examples**

Required argument with constraints:

```python
from typing import Annotated
from pydantic import Field

@function_tool
def score_task(score: Annotated[int, Field(..., ge=0, le=100)]) -> str:
    return f"Score recorded: {score}"
```

Optional argument with default and constraints:

```python
@function_tool
def search(
    query: str,
    limit: Annotated[int, Field(default=10, ge=1, le=100)] = 10,
) -> str:
    return f"Searching for {query} with limit {limit}"
```

Multiple required arguments where one has constraints (only possible without SyntaxError using `Annotated`):

```python
@function_tool
def process(
    positive_number: Annotated[int, Field(..., ge=1)],
    another_arg: str,
) -> str:
    return f"{positive_number}: {another_arg}"
```

Implementation: we preserve per-parameter `Annotated` metadata in `function_schema`, detect `FieldInfo` in that metadata, and use it when building the Pydantic model and JSON schema so that constraints, defaults, and descriptions are applied the same way as for `= Field(...)`.

---

### Test plan

Documentation: the Pydantic `Field` feature for tool arguments was not documented before. As requested in [the PR #1124 review](https://github.com/openai/openai-agents-python/pull/1124#pullrequestreview-3021007243), a new subsection **"Constraining and describing arguments with Pydantic Field"** in `docs/tools.md` now documents both forms (default-based `= Field(...)` and `Annotated[T, Field(...)]`) with example code for each.

The new unit tests mirror the existing tests for the `= Field(...)` format: required constraints (`test_function_with_field_required_constraints` → `test_function_with_annotated_field_required_constraints`), optional with default and constraints, string constraints (min/max length, pattern), and multiple parameters with mixed constraints. Each new test uses the same assertions (JSON schema shape, valid input round-trip, invalid input raises `ValidationError`) as its counterpart, so both syntaxes are covered consistently. 

---

### Checks

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -2,21 +2,21 @@
 
 Tools let agents take actions: things like fetching data, running code, calling external APIs, and even using a computer. The SDK supports five categories:
 
-- Hosted OpenAI tools: run alongside the model on OpenAI servers.
-- Local runtime tools: run in your environment (computer use, shell, apply patch).
-- Function calling: wrap any Python function as a tool.
-- Agents as tools: expose an agent as a callable tool without a full handoff.
-- Experimental: Codex tool: run workspace-scoped Codex tasks from a tool call.
+-   Hosted OpenAI tools: run alongside the model on OpenAI servers.
+-   Local runtime tools: run in your environment (computer use, shell, apply patch).
+-   Function calling: wrap any Python function as a tool.
+-   Agents as tools: expose an agent as a callable tool without a full handoff.
+-   Experimental: Codex tool: run workspace-scoped Codex tasks from a tool call.
 
 ## Hosted tools
 
 OpenAI offers a few built-in tools when using the [`OpenAIResponsesModel`][agents.models.openai_responses.OpenAIResponsesModel]:
 
-- The [`WebSearchTool`][agents.tool.WebSearchTool] lets an agent search the web.
-- The [`FileSearchTool`][agents.tool.FileSearchTool] allows retrieving information from your OpenAI Vector Stores.
-- The [`CodeInterpreterTool`][agents.tool.CodeInterpreterTool] lets the LLM execute code in a sandboxed environment.
-- The [`HostedMCPTool`][agents.tool.HostedMCPTool] exposes a remote MCP server's tools to the model.
-- The [`ImageGenerationTool`][agents.tool.ImageGenerationTool] generates images from a prompt.
+-   The [`WebSearchTool`][agents.tool.WebSearchTool] lets an agent search the web.
+-   The [`FileSearchTool`][agents.tool.FileSearchTool] allows retrieving information from your OpenAI Vector Stores.
+-   The [`CodeInterpreterTool`][agents.tool.CodeInterpreterTool] lets the LLM execute code in a sandboxed environment.
+-   The [`HostedMCPTool`][agents.tool.HostedMCPTool] exposes a remote MCP server's tools to the model.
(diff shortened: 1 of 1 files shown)


docs/tools.md
```diff
@@ -2,21 +2,21 @@
 
 Tools let agents take actions: things like fetching data, running code, calling external APIs, and even using a computer. The SDK supports five categories:
 
--   Hosted OpenAI tools: run alongside the model on OpenAI servers.
```
seratch, line 5: these may be fine, but can you apply these unrelated changes separately?

docs/tools.md
```diff
@@ -278,6 +278,26 @@ As mentioned before, we automatically parse the function signature to extract th
 
 The code for the schema extraction lives in [`agents.function_schema`][].
 
+### Constraining argument values
```
seratch: Having the document change in this PR requires us to immediately release a new version when merging it. Can you create a separate PR for doc chnages?

seratch (changes requested):

haasonsaas:
I can’t push updates to the fork branch from here, so I opened split PRs instead per the review request to keep docs separate:
- #2435 (code-only): support `Annotated[T, Field(...)]` in `function_schema`
- #2436 (docs-only): tools.md subsection

If you prefer updating this PR directly, the minimal change is to revert `docs/tools.md` back to main and keep only the `src/` + `tests/` changes here.

docs/tools.md
georg-wolflein: Absolutely, removed this from the PR

docs/tools.md
georg-wolflein: Makes sense. I removed the documentation changes from this PR. @haasonsaas already opened #2436 so the documentation changes can be merged in a separate PR.

georg-wolflein:
Thanks for the feedback @seratch. I removed the docs changes from this PR, so it should be ready to merge.

@haasonsaas kindly created a separate PR (#2436) for the docs change, which can be merged at the next release.

seratch:
Closing this in favor of https://github.com/openai/openai-agents-python/pull/2435

PR #2429, opened 2026-02-06
Create SECURITY.md

kl-s1:


seratch:
Thanks for sending this, but we'll add this file when necessary

PR #2435, opened 2026-02-07
feat: support Annotated[T, Field(...)] in function schema

haasonsaas:
Derived from #2428, but split docs changes out per review feedback.

- Adds support for `Annotated[T, Field(...)]` on `@function_tool` args.
- Includes tests in `tests/test_function_schema.py`.

Docs follow-up: see the companion PR from `haasonsaas:codex/2428-docs` (kept separate so it can be merged/released independently).

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -210,6 +210,15 @@ def _extract_description_from_metadata(metadata: tuple[Any, ...]) -> str | None:
     return None
 
 
+def _extract_field_info_from_metadata(metadata: tuple[Any, ...]) -> FieldInfo | None:
+    """Returns the first FieldInfo in Annotated metadata, or None."""
+
+    for item in metadata:
+        if isinstance(item, FieldInfo):
+            return item
+    return None
+
+
 def function_schema(
     func: Callable[..., Any],
     docstring_style: DocstringStyle | None = None,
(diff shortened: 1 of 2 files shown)


haasonsaas:
CI is green (lint/typecheck/tests/docs). Looks ready to merge once a maintainer has a moment.

seratch (approved):

Sourav-Nandy-ai:
Hi Jonathan,

I maintain agent-eval (agent_regress), an OSS harness for before-and-after behavior comparisons across agent framework versions. While validating it against real merged PRs, I found my OpenAI Agents SDK integration was completely broken for this kind of test. The openai_agents_runner() was calling agent.run(query), a method that does not exist on the real Agent class. This caused it to crash on the very first call regardless of which PR it was pointed at.

That is fixed now, and I used the corrected integration to verify your PR end-to-end against real openai-agents==0.8.1 (pre-merge) and 0.8.2 (post-merge) installations.

Schema-level: Annotated[int, Field(gt=0)] produces no exclusiveMinimum in the tool's JSON schema pre-merge, but correctly outputs "exclusiveMinimum": 0 post-merge.

Functionally: I ran a real gpt-4o-mini call instructing the model to call the tool with x=-5. Pre-merge, the SDK silently accepted it, and the Python function actually executed with -5. Post-merge, Pydantic rejected it internally. The model then self-corrected to a valid value within the same run, meaning the function never executed on the invalid input.

That self-correction behavior (where the model retries with a fixed value instead of surfacing a raw error) was not something I expected or engineered. It is simply what actually happened. This is exactly the kind of real behavior shift I want a tool like this to catch automatically rather than relying on manual observation.

I am curious whether that retry-on-validation-error pattern matches what you intended for callers who do not handle the tool-error string themselves.

Best,
Sourav

PR #2446, opened 2026-02-09
Include agent in ToolContext tool calls

rm-openai:
Summary
- propagate the current `Agent` into every new `ToolContext`, including realtime session and tool execution paths
- extend `ToolContext` to accept an `agent` keyword, preserve backwards-compatible constructors, and add regression/unit coverage for the new behavior
Testing
- Not run (not requested)


rm-openai:
@codex review, and make sure things are backward compatible

scott-harveyai:
Thanks for adding this! Super useful

PR #2459, opened 2026-02-10
fix(tracing): truncate oversized span input/output for OpenAI ingest

### LINKED ISSUE #260: Truncate span input when input is too large
### Truncate span input when too large
Trace span input has a size limit
~~The whole run fails, not sending the trace should be at most a log warning:
the generation(aka cost) already happened and tracing is secondary.~~

I don't want to disable trace all together. 
I would like to:
1. Skip or truncate input/output automatically when too large on spans
~~2. Don't raise errors but warnings, or at least have the options to ignore them (trace_ignore_errors=True)~~

```
Tracing client error 400: {
  "error": {
    "message": "Invalid 'data[0].span_data.input': data payload too large. Expected a JSON value under 100KB.",
    "param": "data[0].span_data.input",
    "code": null
  }
}
```

weiguangli-io:
## Summary
Fixes #260 by making OpenAI tracing sanitization fail-safe for oversized `span_data.input` / `span_data.output` payloads.

## What changed
- Extended `_sanitize_for_openai_tracing_api` to also sanitize `input` and `output` fields (not just generation usage keys).
- Added byte-size based truncation helpers:
  - Strings are truncated with a stable suffix (`... [truncated]`) while staying under the ingest limit.
  - Non-string oversized payloads are replaced with a compact metadata object (`truncated`, `original_type`, `preview`).
- Kept behavior endpoint-aware: sanitization is only applied for the OpenAI tracing ingest endpoint, existing custom endpoint behavior is unchanged.

## Tests
- `uv run ruff check src/agents/tracing/processors.py tests/test_trace_processor.py`
- `uv run mypy src/agents/tracing/processors.py tests/test_trace_processor.py`
- `uv run pytest tests/test_trace_processor.py -q`
- `make format-check`

Added test coverage for:
- large string truncation
- large non-string replacement
- custom endpoint passthrough
- unchanged small input path
- helper edge-cases (`within limit`, `suffix too large`)

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -261,12 +261,17 @@ class BackendSpanExporter(TracingExporter):
         if isinstance(value, str):
             return self._truncate_string_for_json_limit(value, max_bytes)
 
-        preview = str(value)
-        if len(preview) > 512:
-            preview = preview[:512] + self._OPENAI_TRACING_STRING_TRUNCATION_SUFFIX
+        type_name = type(value).__name__
+        preview = f"<{type_name} truncated>"
+        if isinstance(value, dict):
+            preview = f"<{type_name} len={len(value)} truncated>"
+        elif isinstance(value, (list, tuple, set, frozenset)):
+            preview = f"<{type_name} len={len(value)} truncated>"
+        elif isinstance(value, (bytes, bytearray, memoryview)):
+            preview = f"<{type_name} bytes={len(value)} truncated>"
         return {
             "truncated": True,
-            "original_type": type(value).__name__,
+            "original_type": type_name,
             "preview": preview,
         }
 
--- a/tests/test_trace_processor.py
+++ b/tests/test_trace_processor.py
@@ -461,6 +461,43 @@ def test_backend_span_exporter_truncates_large_non_string_input_for_openai_traci
 
 
 @patch("httpx.Client")
+def test_backend_span_exporter_truncates_large_non_string_input_without_stringifying(mock_client):
+    class NoStringifyDict(dict[str, Any]):
+        def __str__(self) -> str:
+            raise AssertionError("__str__ should not be called for oversized non-string previews")
+
+    class DummyItem:
+        tracing_api_key = None
+
+        def __init__(self):
+            payload_input = NoStringifyDict(
+                blob="x" * (BackendSpanExporter._OPENAI_TRACING_MAX_FIELD_BYTES + 5000)
+            )
+            self.exported_payload: dict[str, Any] = {
+                "object": "trace.span",
+                "span_data": {
+                    "type": "generation",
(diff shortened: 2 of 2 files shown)


src/agents/tracing/processors.py
```diff
@@ -182,32 +185,85 @@ def _should_sanitize_for_openai_tracing_api(self) -> bool:
...
+        if self._value_json_size_bytes(suffix) > max_bytes:
+            return ""
+
+        low = 0
+        high = len(value)
+        best = suffix
+        while low <= high:
+            mid = (low + high) // 2
+            candidate = value[:mid] + suffix
+            if self._value_json_size_bytes(candidate) <= max_bytes:
```
seratch: This line could be called up to 20 times here, and it runs `json.dumps()` each time. Prefer a bit more efficient way to adjust the data size.

weiguangli-io:
Addressed in commit aba2a66.

What changed:
- Reworked `_truncate_string_for_json_limit` to use a proportional size estimate plus overflow-based trimming, instead of a binary search that repeatedly serializes many candidates.
- This reduces repeated `json.dumps()` calls in the hot truncation path while preserving the same safety guarantee (`<= max bytes`).
- Added regression coverage for escape-heavy strings to validate byte-limit behavior with JSON escaping.

Local verification:
- uv run ruff check src/agents/tracing/processors.py tests/test_trace_processor.py
- uv run mypy src/agents/tracing/processors.py
- env -u ALL_PROXY -u all_proxy -u HTTPS_PROXY -u https_proxy -u HTTP_PROXY -u http_proxy -u NO_PROXY -u no_proxy uv run pytest tests/test_trace_processor.py -k "truncate_string_for_json_limit or truncates_large_input_for_openai_tracing" -q

weiguangli-io:
Added one follow-up test commit (3f5a86c) to cover the `max_bytes == suffix_size` branch in `_truncate_string_for_json_limit`, so the updated truncation logic is fully exercised by the targeted trace run.

weiguangli-io:
Thanks for the catch. I pushed a follow-up in `0cddf1c` to avoid materializing `str(value)` for oversized non-string `span_data.input/output` fields.

What changed:
- Replaced `preview = str(value)` with a lightweight type/size summary (`<dict len=... truncated>`, `<list len=... truncated>`, `<bytes bytes=... truncated>`, etc.)
- Added regression test `test_backend_span_exporter_truncates_large_non_string_input_without_stringifying` to assert this path does not call `__str__`

Validation run locally:
- `uv run --with ruff ruff check src/agents/tracing/processors.py tests/test_trace_processor.py`
- `uv run --with pytest pytest -q tests/test_trace_processor.py` (28 passed)

If this looks good, I can resolve the thread.

seratch:
Closing this in favor of #2549; Thank you so much again for working on this!

PR #2460, opened 2026-02-10
fix(tool): resolve default failure handler at invoke time

### LINKED ISSUE #945: Get exception raised in tool invocation
I am the Sentry SDK maintainer, and I implemented instrumentation for OpenAI Agents SDK.

There is one thing I could not instrument. If there is an error in a tool call the exception is swallowed. 

This is the code handling the error in openai-agents:
https://github.com/openai/openai-agents-python/blob/main/src/agents/tool.py#L403-L422

Because `default_tool_error_function` is set as a default value to the `default_tool_error_function` parameter of the `function_tool` decorator I can not monkey patch because this is evaluated during import time and the Sentry SDK is initialized later. 

So right now there is no way for me to capture those errors. 

Are there plans to make this easier to instrument or maybe create a `on_tool_error` hook or something like this?

I am happy to provide a PR with a solution for this!

weiguangli-io:
## Summary
Fixes #945 by resolving the default tool failure handler at invocation time instead of binding it at decorator definition time.

This allows instrumentation libraries (for example Sentry) to patch `default_tool_error_function` after import and still have function tools use the patched handler.

## What changed
- Introduced an internal sentinel for `function_tool(..., failure_error_function=...)` to distinguish:
  - omitted argument (use default handler)
  - explicit `None` (re-raise exceptions)
  - explicit custom handler
- Updated runtime error handling in `function_tool` to resolve `default_tool_error_function` inside the invocation path.
- Added regression test proving that monkey-patching `default_tool_error_function` *after* creating the tool still affects tool failure handling.

## Tests
- `uv run ruff check src/agents/tool.py tests/test_function_tool.py`
- `uv run mypy src/agents/tool.py tests/test_function_tool.py`
- `uv run pytest tests/test_function_tool.py -q`
- `make format-check`
- `uv run python -m trace --count --coverdir /tmp/openai_agents_945_trace.Ig3U8E --module pytest tests/test_function_tool.py -q`

Coverage note: changed branches in `src/agents/tool.py` are all hit by tests (including omitted/default, explicit `None`, and explicit custom handler paths).

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -12,6 +12,7 @@ from typing import (
     Callable,
     Generic,
     Literal,
+    Optional,
     Protocol,
     TypeVar,
     Union,
@@ -929,7 +930,7 @@ def function_tool(
                     resolved_failure_error_function = default_tool_error_function
                 else:
                     resolved_failure_error_function = cast(
-                        ToolErrorFunction | None, failure_error_function
+                        Optional[ToolErrorFunction], failure_error_function
                     )
 
                 if resolved_failure_error_function is None:


weiguangli-io:
Follow-up fix pushed in 6e00ced for CI `old_version_tests`.

Root cause:
- A runtime `cast(ToolErrorFunction | None, ...)` expression introduced by this PR is not valid on Python 3.9, and raised:
  `TypeError: unsupported operand type(s) for |: '_CallableGenericAlias' and 'NoneType'`.

Fix:
- Replaced the runtime cast target with `Optional[ToolErrorFunction]` for py3.9 compatibility.

Local verification after the fix:
- `uv run ruff check src/agents/tool.py tests/test_function_tool.py`
- `uv run mypy src/agents/tool.py tests/test_function_tool.py`
- `uv run pytest tests/test_function_tool.py tests/test_function_tool_decorator.py -q`
- `env -u all_proxy -u ALL_PROXY -u http_proxy -u HTTP_PROXY -u https_proxy -u HTTPS_PROXY -u NO_PROXY -u no_proxy UV_PROJECT_ENVIRONMENT=.venv_39 uv run --python 3.9 -m pytest tests/test_function_tool.py tests/test_function_tool_decorator.py tests/test_run_step_execution.py tests/test_tracing_errors.py tests/test_tracing_errors_streamed.py -k "simple_function or complex_args_function or custom_error_function or async_failure_error_function_is_awaited or invalid_json_raises_with_failure_error_function_none or tool_call_error or error_on_invalid_json or no_error_on_invalid_json" -q`

seratch (approved):

Issue #610, opened 2025-04-26
UnicodeDecodeError when importing litellm_model on Windows


Hamzah-syed:
- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/): Yes
- **Have you searched for related issues?** Others may have faced similar issues.: Yes, no similar issues found.

### Describe the bug
When importing litellm_model from agents.extensions.models, a UnicodeDecodeError occurs on Windows systems.

### Debug information
Agents SDK version: 0.0.13
Python version: 3.11
Operating System: Windows 11
Package manager: UV

### Repro steps
1. ```pip install openai-agents[litellm]```
2. ```from agents.extensions.models import litellm_model```
3. ```uv run main.py```

![Image](https://github.com/user-attachments/assets/7f5ede9c-bc8f-497f-8867-57a6b957770e)

MoosaAfzal2:
Encountering same error on windows 10 with sdk version 0.0.13 and python version 3.12

seratch:
Does anyone still experience this with latest versions of openai-agents and litelllm?

weiguangli-io:
I opened a fix PR for this issue: https://github.com/openai/openai-agents-python/pull/2462

What it changes:
- Defers `litellm` import in `litellm_model.py` so `from agents.extensions.models import litellm_model` no longer eagerly imports LiteLLM.
- Keeps the same dependency error, but raises it only when LiteLLM is actually used.
- Adds regression tests for both failure and success/caching paths of the lazy loader.

Local validation:
- ruff checks on touched files: pass
- targeted pytest suite for lazy import + user-agent/map paths: pass
- litellm-dependent tests (`test_kwargs_functionality`, `test_litellm_extra_body`): pass
- trace run confirms the new lazy-import branches are executed.

seratch:
Let me close this issue now as no one else reported this afterwards. That said, if anyone else is still experiencing this, letting us know would be appreciated

PR #2462, opened 2026-02-10
fix(litellm): defer import until model usage

weiguangli-io:
## Summary
This defers `litellm` import in `litellm_model.py` so importing `agents.extensions.models.litellm_model` does not eagerly import LiteLLM.

This targets issue #610, where Windows users hit `UnicodeDecodeError` at module import time.

## What changed
- Replaced top-level `import litellm` with a lazy loader (`_import_litellm`) and a module proxy.
- Kept the existing dependency error message, but now raised at first actual LiteLLM usage.
- Added regression tests for both paths:
  - deferred import + import failure path (including cached failure)
  - successful import + cached module reuse path

## Why this helps
- Importing `litellm_model` no longer fails immediately due to LiteLLM import-time issues.
- The failure is deferred to actual LiteLLM usage, which avoids breaking unrelated imports and startup paths.

## Testing
- `uv run --with ruff ruff check src/agents/extensions/models/litellm_model.py tests/models/test_litellm_deferred_import.py tests/models/test_litellm_user_agent.py tests/models/test_map.py`
- `uv run --with pytest pytest -q tests/models/test_litellm_deferred_import.py tests/models/test_litellm_user_agent.py tests/models/test_map.py`
- `uv run --with litellm --with pytest pytest -q tests/models/test_kwargs_functionality.py tests/models/test_litellm_extra_body.py`
- `uv run --with pytest python -m trace --count --coverdir /tmp/openai_agents_610_trace_1770722012 --module pytest tests/models/test_litellm_deferred_import.py -q`

Trace output confirms all newly added lazy-import branches were executed.

### DIFF
--- a/tests/models/test_litellm_deferred_import.py
+++ b/tests/models/test_litellm_deferred_import.py
@@ -4,6 +4,7 @@ import builtins
 import importlib
 import sys
 import types as pytypes
+from typing import Any, cast
 
 import pytest
 
@@ -39,7 +40,8 @@ def test_litellm_import_is_deferred_until_module_usage(monkeypatch):
 def test_litellm_import_loader_caches_successful_import(monkeypatch):
     fake_litellm = pytypes.ModuleType("litellm")
     marker = object()
-    fake_litellm.marker = marker
+    fake_litellm_any = cast(Any, fake_litellm)
+    fake_litellm_any.marker = marker
 
     monkeypatch.setitem(sys.modules, "litellm", fake_litellm)
     monkeypatch.delitem(sys.modules, "agents.extensions.models.litellm_model", raising=False)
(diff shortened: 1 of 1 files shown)


weiguangli-io:
Follow-up fix pushed for CI typecheck failure (new commit: `e386968`).

Root cause:
- `mypy` flagged dynamic `ModuleType` test attributes in `test_litellm_deferred_import.py`.

What I changed:
- Kept the same test behavior, but switched to explicit `Any` casts for dynamic module attributes so both `ruff` and `mypy` pass.

Local re-validation:
- `uv run --with ruff ruff check tests/models/test_litellm_deferred_import.py`
- `uv run --with pytest pytest -q tests/models/test_litellm_deferred_import.py tests/models/test_litellm_user_agent.py tests/models/test_map.py`
- `uv run --with litellm --with pytest pytest -q tests/models/test_kwargs_functionality.py tests/models/test_litellm_extra_body.py`
- `uv run --with litellm --with mypy mypy --follow-imports=silent tests/models/test_litellm_deferred_import.py`

seratch:
We'd like to hold off having this change just to avoid UnicodeDecodeError on Windows OS. Indeed, UnicodeDecodeError is confusing but changing the timing to check the module existence later on does not make sense here.

weiguangli-io:
Thanks for the clear guidance. Agreed on holding this change for now. I am closing this PR to avoid churn; if priorities change later, I can reopen with a narrower proposal aligned with this direction.

weiguangli-io:
Closed per maintainer direction in this thread.

PR #2463, opened 2026-02-10
fix(agent-tools): inherit parent RunConfig when agent-as-tool is invoked

weiguangli-io:
## Summary
This fixes a bug where an agent used as a tool could ignore the parent run's `RunConfig` (notably custom `model_provider`) unless a separate `run_config` was explicitly passed to `Agent.as_tool(...)`.

## What changed
- Added `run_config` to `ToolContext` so tool invocations can carry run-level configuration metadata.
- In function tool execution, pass the active run `config` into `ToolContext.from_agent_context(...)`.
- In `Agent.as_tool(...)`:
  - if `run_config` is explicitly provided, keep using it;
  - otherwise inherit `run_config` from the parent `ToolContext` when available.
- Added regression tests for:
  - inheriting parent `run_config` when `as_tool(run_config=None)`;
  - explicit `run_config` overriding parent context;
  - function tool contexts receiving the active `RunConfig`.

## Why this is safe
- Backward compatibility is preserved: explicit `run_config` behavior remains unchanged.
- Existing code paths without nested agent tools are unaffected.
- The change is additive (`ToolContext.run_config` defaults to `None`).

## Validation
- `uv run --with ruff ruff check src/agents/agent.py src/agents/tool_context.py src/agents/run_internal/tool_execution.py tests/test_agent_as_tool.py tests/test_tool_context.py tests/test_run_step_execution.py`
- `uv run --with pytest --with socksio pytest -q tests/test_tool_context.py tests/test_agent_as_tool.py tests/test_run_step_execution.py`
- `uv run --with mypy --with socksio mypy src/agents/agent.py src/agents/tool_context.py src/agents/run_internal/tool_execution.py tests/test_tool_context.py tests/test_agent_as_tool.py tests/test_run_step_execution.py`
- Targeted trace coverage run to confirm modified paths were exercised.

### DIFF
--- a/src/agents/tool_context.py
+++ b/src/agents/tool_context.py
@@ -100,6 +100,7 @@ class ToolContext(RunContextWrapper[TContext]):
         tool_call_id: str,
         tool_call: ResponseFunctionToolCall | None = None,
         agent: AgentBase[Any] | None = None,
+        *,
         run_config: RunConfig | None = None,
     ) -> ToolContext:
         """
--- a/src/agents/tool_context.py
+++ b/src/agents/tool_context.py
@@ -11,6 +11,7 @@ from .usage import Usage
 if TYPE_CHECKING:
     from .agent import AgentBase
     from .items import TResponseInputItem
+    from .run_config import RunConfig
     from .run_context import _ApprovalRecord
 
 
@@ -48,6 +49,9 @@ class ToolContext(RunContextWrapper[TContext]):
     agent: AgentBase[Any] | None = None
     """The active agent for this tool call, when available."""
 
+    run_config: RunConfig | None = None
+    """The active run config for this tool call, when available."""
+
     def __init__(
         self,
         context: TContext,
@@ -58,6 +62,7 @@ class ToolContext(RunContextWrapper[TContext]):
         tool_call: ResponseFunctionToolCall | None = None,
         *,
         agent: AgentBase[Any] | None = None,
+        run_config: RunConfig | None = None,
         turn_input: list[TResponseInputItem] | None = None,
         _approvals: dict[str, _ApprovalRecord] | None = None,
         tool_input: Any | None = None,
@@ -86,6 +91,7 @@ class ToolContext(RunContextWrapper[TContext]):
         )
         self.tool_call = tool_call
         self.agent = agent
+        self.run_config = run_config
 
     @classmethod
     def from_agent_context(
@@ -94,6 +100,7 @@ class ToolContext(RunContextWrapper[TContext]):
         tool_call_id: str,
         tool_call: ResponseFunctionToolCall | None = None,
         agent: AgentBase[Any] | None = None,
+        run_config: RunConfig | None = None,
     ) -> ToolContext:
         """
         Create a ToolContext from a RunContextWrapper.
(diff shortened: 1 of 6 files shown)


weiguangli-io:
Added a follow-up regression test in commit b8741b1 to explicitly cover inherited tracing privacy config on agent-as-tool runs:\n\n- \n\nThis verifies that  on the parent RunConfig is preserved for the nested run when  does not pass an explicit run_config.

weiguangli-io:
Correction for formatting: added follow-up regression test in commit b8741b1 named test_agent_as_tool_inherits_trace_include_sensitive_data_setting. It verifies that trace_include_sensitive_data=False on the parent RunConfig is preserved for nested agent-as-tool runs when no explicit run_config is passed to as_tool.

src/agents/tool_context.py
```diff
@@ -94,6 +100,7 @@ def from_agent_context(
         tool_call_id: str,
         tool_call: ResponseFunctionToolCall | None = None,
         agent: AgentBase[Any] | None = None,
+        run_config: RunConfig | None = None,
```
seratch, line 104: Can you add this for clarity for future changes?
```suggestion
        *,
        run_config: RunConfig | None = None,
```

seratch (changes requested):
one minor request

seratch:
@codex review again

weiguangli-io:
Implemented this suggestion in commit 0d6f225.

Change made:
- ToolContext.from_agent_context() now marks run_config as keyword-only (`*`) for clearer call-site intent and safer future parameter additions.

Local verification:
- uv run ruff check src/agents/tool_context.py tests/test_tool_context.py tests/test_tool_metadata.py
- uv run mypy src/agents/tool_context.py
- env -u ALL_PROXY -u all_proxy -u HTTPS_PROXY -u https_proxy -u HTTP_PROXY -u http_proxy -u NO_PROXY -u no_proxy uv run pytest tests/test_tool_context.py tests/test_tool_metadata.py -q (11 passed)

seratch (approved):

Sourav-Nandy-ai:
Hi Weiguang,

I actually ran a real before-and-after test on this behavior rather than just reading the diff. I built two isolated installations: openai-agents==0.8.3 (the last release before your merge commit) and 0.8.4 (confirmed via the GitHub API comparison to contain it). I then reproduced the exact scenario from your PR title: a "Coordinator" agent whose only tool is a "Helper" agent wrapped via as_tool(), executed with a top-level RunConfig(trace_include_sensitive_data=False).

The Results
This is a real, observed result based on object identity rather than just coincidentally matching values:
- Pre-fix: The nested Runner.run call for the as_tool invocation received no run_config at all. It was not even a default configuration; it just silently resolved to None, meaning trace_include_sensitive_data=False never propagated.
- Post-fix: The nested call receives the exact same RunConfig Python object as the parent, which I verified via id() matching instead of just field equality. The field now reads back correctly.

This confirms the fix does exactly what the title says.

I maintain agent-eval (agent_regress on PyPI), an open-source regression harness for agent behavior. This PR's bug is a prime example of why I built trace capture into the harness. Your regression test shows that the dropped config would have silently changed privacy and tracing behavior on nested calls. Because this field never appears in an agent's final text output, no amount of manual response review would ever surface it.

You can find my full scripts and raw JSON output here: https://github.com/RudrenduPaul/agent-eval

Given you found and fixed the original bug, I am curious whether identity-based inheritance (the exact same object, not a copy) is the invariant you want callers to rely on going forward, or if there is an edge case (such as an explicit override at the nested call site) that I should check next.

Best,
Sourav

PR #2488, opened 2026-02-14
feat: add optional wrapper parameter to Session methods

### LINKED ISSUE #2072: Pass RunContextWrapper[T] to Session methods
### Feature Request: Pass `RunContextWrapper[T]` to Session methods

**Summary**

Right now, `Session` methods (`get_items`, `add_items`, `pop_item`, etc.) don’t receive the agent’s `RunContextWrapper[T]`.
Passing the wrapper directly would let a custom session manage both conversation history and context together.

---

**Proposed Change**

Add `wrapper` as an optional argument to session methods:

```python
async def get_items(self, limit: int | None = None, wrapper: RunContextWrapper[T] | None = None) -> list[dict]:
    ...

async def add_items(self, items: list[dict], wrapper: RunContextWrapper[T] | None = None) -> None:
    ...
```

This allows custom sessions to decide how to store or update the wrapper along with messages.

---

**Benefits**

* Gives custom sessions full control over context and history
* Makes multi-turn, stateful agents easier to implement

---

**Question / Concern**

In the non-streaming interface, I noticed that `add_items` is only called after handoffs are complete.
This means if an agent is switched mid-session, the session can’t save info like the agent’s name or other context until the very end.

Is there a recommended way to handle this, or should `add_items` be triggered more incrementally to capture agent-specific information during handoff?

Limitless2023:
## Summary

Add optional  parameter to Session protocol methods (, , , ) to allow custom sessions to access context data during history operations.

This enables custom sessions to make decisions based on agent context, such as storing agent-specific information during handoff.

Fixes #2072

## Changes

- Add optional  parameter to:
  -  protocol (session.py)
  -  abstract class (session.py)
  -  implementation (sqlite_session.py)
  -  implementation (openai_conversations_session.py)
  -  implementation (openai_responses_compaction_session.py)

## Test Plan

Existing tests should pass since the new parameter is optional with a default value of , maintaining backward compatibility.

### DIFF
--- a/src/agents/memory/openai_responses_compaction_session.py
+++ b/src/agents/memory/openai_responses_compaction_session.py
@@ -249,7 +249,11 @@ class OpenAIResponsesCompactionSession(SessionABC, OpenAIResponsesCompactionAwar
         Returns:
             List of input items representing the conversation history
         """
-        return await self.underlying_session.get_items(limit, wrapper)
+        # Only pass wrapper if provided, for backward compatibility with sessions
+        # that haven't updated their signatures yet
+        if wrapper is not None:
+            return await self.underlying_session.get_items(limit, wrapper)
+        return await self.underlying_session.get_items(limit)
 
     async def _defer_compaction(self, response_id: str, store: bool | None = None) -> None:
         if self._deferred_response_id is not None:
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. However, I don't think just adding the optional parameter can resolve any needs. The context needs to be passed from runner side without breaking changes.

Limitless2023:
Closing based on maintainer feedback. Will revisit the design to pass context from runner side instead of adding optional parameters to session methods.

PR #2493, opened 2026-02-15
fix(models): allow slash-containing model IDs with unknown prefixes

### LINKED ISSUE #2492: Unknown prefix when using openrouter models
```
>>> openai.__version__, agents.__version__
('2.21.0', '0.9.0')
```

Issue:
When using open router models, i will get an error: prefix not found.

https://github.com/openai/openai-agents-python/blob/84fa471e5fc538d744a3ae294749fedb3855131b/src/agents/models/multi_provider.py#L128

This is due to the sdk think it is a litellm prefix. 
Temp fix:
change the get_model to 
```python

        """Returns a Model based on the model name. The model name can have a prefix, ending with
        a "/", which will be used to look up the ModelProvider. If there is no prefix, we will use
        the OpenAI provider.

        Args:
            model_name: The name of the model to get.

        Returns:
            A Model.
        """
        # prefix, model_name = self._get_prefix_and_model_name(model_name)
        prefix, model_name = None, model_name
        if prefix and self.provider_map and (provider := self.provider_map.get_provider(prefix)):
            return provider.get_model(model_name)
        else:

```

weiguangli-io:
## Summary
Fixes #2492 by making `MultiProvider` treat unknown prefixes as plain model names instead of hard-failing with `Unknown prefix`.

This keeps explicit provider prefixes (`openai/`, `litellm/`, and prefixes configured in `provider_map`) working as before, while allowing model IDs that naturally contain slashes (for example `openrouter/openai/gpt-5`) to be sent to the default OpenAI provider unchanged.

## What changed
- Added `_is_known_prefix()` in `MultiProvider`.
- Updated `get_model()` routing logic:
  - Known prefix: keep prefix-based routing.
  - Unknown prefix: do not treat it as provider prefix; route as plain model name.
- Added new tests in `tests/models/test_multi_provider.py` covering:
  - unknown prefix passthrough
  - known fallback prefix routing
  - provider_map routing
  - helper map mutations
  - fallback cache behavior
  - unknown fallback error path
  - dynamic litellm fallback import path

## Validation
- `uv run --with ruff ruff check src/agents/models/multi_provider.py tests/models/test_multi_provider.py`
- `uv run mypy src/agents/models/multi_provider.py tests/models/test_multi_provider.py`
- `uv run --with pytest pytest -q tests/models/test_multi_provider.py`
- `uv run --with coverage --with pytest coverage run -m pytest -q tests/models/test_multi_provider.py`
- `uv run --with coverage coverage report -m src/agents/models/multi_provider.py`

Coverage result for changed source file:
- `src/agents/models/multi_provider.py`: **100%**

### DIFF
--- a/src/agents/models/multi_provider.py
+++ b/src/agents/models/multi_provider.py
@@ -108,6 +108,15 @@ class MultiProvider(ModelProvider):
         else:
             return None, model_name
 
+    def _is_known_prefix(self, prefix: str | None) -> bool:
+        if prefix is None:
+            return False
+        if prefix in {"openai", "litellm"}:
+            return True
+        if self.provider_map and self.provider_map.has_prefix(prefix):
+            return True
+        return False
+
     def _create_fallback_provider(self, prefix: str) -> ModelProvider:
         if prefix == "litellm":
             from ..extensions.models.litellm_provider import LitellmProvider
@@ -136,7 +145,11 @@ class MultiProvider(ModelProvider):
         Returns:
             A Model.
         """
-        prefix, model_name = self._get_prefix_and_model_name(model_name)
+        prefix, parsed_model_name = self._get_prefix_and_model_name(model_name)
+        if self._is_known_prefix(prefix):
+            model_name = parsed_model_name
+        else:
+            prefix = None
 
         if prefix and self.provider_map and (provider := self.provider_map.get_provider(prefix)):
             return provider.get_model(model_name)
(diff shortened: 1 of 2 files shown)


seratch:
These changes may actually work, but we generally recommend using either LiteLLM or model object for non-OpenAI models. Making the parser complex for further flexibility (the litellm was the special case because we generally recommend using litellm for non-OAI models) may not be a change we would like to have.

weiguangli-io:
Thanks for the clear guidance, this makes sense.

I agree we should keep the parser behavior conservative and align with the current recommendation (LiteLLM or explicit model objects for non-OpenAI providers). I can close this PR to avoid adding parser complexity unless you prefer a much narrower variant.

PR #2514, opened 2026-02-18
Add a meta attribute to the base function class

jaw9c:
## Problem

When using MCP function calls, we lose the metadata attached. 

## Note

This is changing the core function class so open to comments on a better setup

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -308,6 +308,7 @@ class MCPUtil:
             on_invoke_tool=invoke_func,
             strict_json_schema=is_strict,
             needs_approval=needs_approval,
+            meta=tool.meta,
         )
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sending this patch! To pass _meta to an MCP server tool call, please use tool_meta_resolver: https://openai.github.io/openai-agents-python/mcp/#per-call-metadata-with-tool_meta_resolver

jaw9c:
@seratch Thanks for sending through - thats not quite the same problem - I'm trying to retrieve the _meta after a function call has been made rather than send it during calling.

seratch:
Thanks for your reply. I see your need here. Your code change suggestions do not work well for the purpose. Adding an optional hook to MCP Sever interface or new optional property to function result objects may be useful. Need to check available options in depth.

PR #2528, opened 2026-02-22
[v2] feat: Add TIAMAT persistent cloud memory backend

toxfox69:
## Summary

**TiamatSession** — a persistent cloud memory backend for the OpenAI Agents SDK via [TIAMAT Memory API](https://memory.tiamat.live). Zero-infrastructure, cross-session agent memory with FTS5 full-text search.

- Implements `SessionABC` interface
- Persistent conversation memory that survives process restarts
- FTS5 full-text search across conversation history
- Cross-device memory accessible via cloud API
- Only requires `httpx` — no Redis, SQLite, or infrastructure setup

### Fix 1 (P1): `pop_item` deadlock
- **Before**: `pop_item()` acquired `self._lock`, then called `self.get_items()` which tried to acquire the same `asyncio.Lock` — deadlock (non-reentrant lock).
- **After**: Extracted `_get_items_unlocked()` helper. Both `get_items()` and `pop_item()` call this internal method — `get_items()` acquires the lock first, `pop_item()` calls it while already holding the lock.

### Fix 2 (P1): `clear_session` marker not respected
- **Before**: `clear_session()` stored a `_tiamat_clear` marker, but `get_items()` never filtered on it — old items returned after clear.
- **After**: Two-pass parsing in `_get_items_unlocked()`: first pass finds latest `_tiamat_clear` sequence number, second pass filters to only return items after that clear point.

### Fix 3 (P2): Hardcoded `limit:100` truncating history
- **Before**: `session_limit or 100` silently capped at 100 entries when no limit was set.
- **After**: `session_limit if session_limit and session_limit > 0 else 10000` — generous default ceiling that respects explicit limits.

## Files

- `examples/tiamat_memory/tiamat_session.py` — Core session class (all fixes applied)
- `examples/tiamat_memory/agent_with_memory.py` — Working example
- `examples/tiamat_memory/README.md` — Documentation

## Test Plan

- [x] `pop_item` completes without deadlock under concurrent access
- [x] `get_items` returns empty list after `clear_session`
- [x] Sessions without explicit limits return full history (>100 items)
- [x] Graceful fallback when API is unreachable

### DIFF
--- a/examples/tiamat_memory/test_tiamat_session.py
+++ b/examples/tiamat_memory/test_tiamat_session.py
@@ -1,0 +1,231 @@
+"""Tests for TiamatSession — verifies SessionABC contract compliance.
+
+These tests use a mock HTTP transport so they run without network access.
+"""
+
+from __future__ import annotations
+
+import json
+from typing import Any
+
+import httpx
+import pytest
+
+from tiamat_session import TiamatSession
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+class MemoryStore:
+    """In-memory mock of the TIAMAT Memory API for testing."""
+
+    def __init__(self) -> None:
+        self.memories: list[dict[str, Any]] = []
+
+    def store(self, content: str, tags: list[str], importance: float) -> dict[str, Any]:
+        entry = {"content": content, "tags": tags, "importance": importance}
+        self.memories.append(entry)
+        return {"status": "stored"}
+
+    def recall(self, query: str, limit: int) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


toxfox69:
This is a resubmission of #2525 with all review feedback addressed. The three issues flagged by Codex (lock deadlock in pop_item, clear_session marker not respected, hardcoded limit cap) have all been fixed. Happy to address any additional feedback.

seratch:
@toxfox69 As I mentioned at https://github.com/openai/openai-agents-python/pull/2525#issuecomment-3940260337 , we won't have this example as part of this repo. Please have your own repo for it.

PR #2541, opened 2026-02-24
Fix openrouter model prefix issue by using LitellmProvider for all prefixes

liangzhang-keepmoving:
This PR fixes an issue where the SDK would throw an "Unknown prefix" error when using openrouter models.

The problem was in the `_create_fallback_provider` method, which only handled the "litellm" prefix and threw an error for all other prefixes. However, LiteLLM actually supports many other model providers, including openrouter, and expects model names in the format "provider/model".

The fix:
1. Modified `_create_fallback_provider` to return a `LitellmProvider` for all prefixes, since LiteLLM can handle various model provider prefixes
2. Updated `get_model` to pass the full model name (including prefix) to LiteLLM when using an unknown prefix, so LiteLLM can correctly route the request

This change ensures that models like "openrouter/gpt-4" will work correctly through the multi-provider system.

### DIFF
--- a/src/agents/models/multi_provider.py
+++ b/src/agents/models/multi_provider.py
@@ -1,106 +1,47 @@
 from __future__ import annotations
 
-from openai import AsyncOpenAI
+import copy
+from typing import TYPE_CHECKING
 
-from ..exceptions import UserError
+from .default_models import get_default_model
 from .interface import Model, ModelProvider
 from .openai_provider import OpenAIProvider
 
-
-class MultiProviderMap:
-    """A map of model name prefixes to ModelProviders."""
-
-    def __init__(self):
-        self._mapping: dict[str, ModelProvider] = {}
-
-    def has_prefix(self, prefix: str) -> bool:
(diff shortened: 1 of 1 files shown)


liangzhang-keepmoving:
@hshahrokni2 Thanks for the positive feedback and for sharing your insights! I'm glad the fix addresses the OpenRouter prefix issue effectively.

You make a great point about LiteLLM being the right solution here since it already speaks OpenRouter's dialect. That's exactly why we chose this approach - it's clean and leverages LiteLLM's existing capabilities.

Also, thanks for mentioning Komilion as an alternative solution for routing across multiple models. That's a useful option for those looking to simplify their provider-prefix management even further.

@openai/openai-agents-python-maintainers This PR has received positive feedback and addresses a real issue with OpenRouter model support. Would you be able to review it when you have a chance?

seratch:
Thanks for sending this patch. However, this looks like a breaking change, and to me, what issue you're trying to solve  is still unclear. Can you add some unit tests demonstrating the scenario plus avoid any breaking changes?

PR #2542, opened 2026-02-24
Fix openrouter model prefix handling

liangzhang-keepmoving:
This fix addresses the issue with openrouter model prefix handling by:

1. Modifying `_create_fallback_provider` to return LitellmProvider for all prefixes
2. Updating `get_model` to pass the full model name (including prefix) to LiteLLM for unknown prefixes

This allows the framework to properly handle models with openrouter prefixes like `openrouter/google/gemma-7b-it` by passing them directly to LiteLLM, which can handle various model provider prefixes.

### DIFF
--- a/src/agents/models/multi_provider.py
+++ b/src/agents/models/multi_provider.py
@@ -147,7 +147,11 @@ class MultiProvider(ModelProvider):
             return provider.get_model(model_name)
         elif prefix:
             # For unknown prefixes, pass the full model name (including prefix) to LiteLLM
-            return self._get_fallback_provider(prefix).get_model(f"{prefix}/{model_name}")
+            # Except for litellm prefix, where we should just pass the model name (without prefix)
+            if prefix == "litellm":
+                return self._get_fallback_provider(prefix).get_model(model_name)
+            else:
+                return self._get_fallback_provider(prefix).get_model(f"{prefix}/{model_name}")
         else:
             return self._get_fallback_provider(prefix).get_model(model_name)
 


liangzhang-keepmoving:
@codex review

I've addressed the feedback you provided:

1. **Preserve litellm prefix stripping behavior**: Fixed by adding a special case in the `get_model` method to handle "litellm" prefix differently - it now passes just the model name without re-adding the prefix, maintaining the existing routing semantics.

2. **Keep MultiProvider defaulting to Responses API**: The code already preserves this behavior since `openai_use_responses` and `openai_use_responses_websocket` are optional parameters with `None` as default, which allows OpenAIProvider to use its default behavior.

3. **Restore functional MultiProviderMap compatibility**: MultiProviderMap is already implemented as a separate class with all the required methods (add_provider, remove_provider, get_mapping, etc.), so no changes were needed here.

4. **Close providers sourced from provider_map in aclose**: The `aclose` method already includes providers from provider_map and uses a seen set to avoid duplicates, so this is already correctly implemented.

The fix now properly handles both openrouter models (by passing the full model name including prefix to LiteLLM) and maintains the existing behavior for litellm models (by stripping the prefix before passing to LiteLLMProvider).

liangzhang-keepmoving:
@codex review

Thank you for the feedback! I've already addressed this issue in the latest commit (16cedc13845d3bb3540b17a66bec161b9569a040).

The fix adds a special case in the `get_model` method to handle "litellm" prefix differently:

```python
elif prefix:
    # For unknown prefixes, pass the full model name (including prefix) to LiteLLM
    # Except for litellm prefix, where we should just pass the model name (without prefix)
    if prefix == "litellm":
    else:
        return self._get_fallback_provider(prefix).get_model(f"{prefix}/{model_name}")
```

This ensures that:
1. For unknown prefixes like "openrouter", it passes the full model name including prefix to LiteLLM
2. For "litellm" prefix, it only passes the model name without re-adding the prefix, maintaining the existing routing behavior

The fix now correctly handles both use cases while preserving the documented litellm/... flow.

liangzhang-keepmoving:
@codex review

Thank you for the feedback! I've addressed this issue in the latest commit (0066b352).

The fix now handles "openai" prefix the same way as "litellm" prefix:

```python
elif prefix:
    # For known prefixes (openai, litellm), pass just the model name (without prefix)
    # For unknown prefixes (like openrouter), pass the full model name (including prefix) to LiteLLM
    if prefix == "litellm" or prefix == "openai":
    else:
        return self._get_fallback_provider(prefix).get_model(f"{prefix}/{model_name}")
```

This ensures that:
1. For "openai" prefix, it only passes the model name without re-adding the prefix, maintaining the existing behavior
2. For "litellm" prefix, it continues to pass just the model name
3. For unknown prefixes like "openrouter", it passes the full model name including prefix to LiteLLM

The fix now correctly handles all three cases while preserving the documented behavior for both openai/... and litellm/... flows.

liangzhang-keepmoving:
@openai/openai-agents-python-maintainers Could you please review this PR when you have a chance?

This PR fixes an issue where the SDK would throw an "Unknown prefix" error when using openrouter models. The fix:

1. Modifies `_create_fallback_provider` to return LitellmProvider for all prefixes
2. Updates `get_model` to pass the full model name (including prefix) to LiteLLM for unknown prefixes
3. Maintains the existing behavior for "litellm" and "openai" prefixes

This is an important fix that allows users to seamlessly use openrouter models through the multi-provider system, alongside existing openai and litellm models.

The fix has already addressed all feedback from Codex review, ensuring that:
- openai/ models are still stripped before OpenAIProvider
- litellm/ models maintain their existing routing behavior
- openrouter/ models work correctly with the full prefix passed to LiteLLM

Thank you for your time!

seratch:
This seems to be duplicated of #2541

PR #2543, opened 2026-02-25
feat: add double-buffer session decorator for context management

marklubin:
## Summary
Adds DoubleBufferSession — a Session decorator for proactive double-buffered context compaction. Non-blocking async checkpoint, stop-the-world fallback, configurable timeout, renewal policies.

Reference: https://marklubin.me/posts/hopping-context-windows/

## Test plan
- [x] 32 unit tests — all pass
- [x] Lint clean

## AI Disclosure
Developed with Claude (Anthropic). All code reviewed, tested, and validated by the author.

**Staging PR on fork — will open upstream PR from here when ready.**

### DIFF
--- a/src/agents/extensions/memory/__init__.py
+++ b/src/agents/extensions/memory/__init__.py
@@ -18,6 +18,11 @@ if TYPE_CHECKING:
         DAPR_CONSISTENCY_STRONG,
         DaprSession,
     )
+    from .double_buffer_session import (
+        DoubleBufferSession,
+        RenewalPolicy,
+        Summarizer,
+    )
     from .encrypt_session import EncryptedSession
     from .redis_session import RedisSession
     from .sqlalchemy_session import SQLAlchemySession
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for the thoughtful contribution. The double-buffering idea is interesting, but after reviewing it we don’t think it’s the right fit for the core library at this time.

The main issue is that it does not yet line up cleanly with our existing session contracts and wrapper patterns. For example, it keeps `session_settings` on the wrapper instead of delegating to the underlying session, it does not preserve additional capabilities exposed by wrapped sessions such as `close()` or `run_compaction()`, and it keeps critical state like the back buffer and generation history only in process memory rather than in the underlying session backend. That makes the behavior harder to reason about across wrappers, persistent backends, and multi-worker usage.

Given the added maintenance surface and the need for a broader design around long-running memory and compaction, we’re going to pass on merging this for now.

That said, we’d be very happy to see this continue as a standalone package or in your own repo if you’d like to keep exploring the approach there.

marklubin:
Thank you for the considered feedback — really appreciate you taking the time to dig into the implementation details rather than just closing it out.

I have a lot of ideas I'm working with around exactly those sorts of problems (session contract composition, persistent compaction state, multi-worker coordination), so would love to discuss further if the opportunity ever comes up.

PR #2545, opened 2026-02-25
docs(agent-tools): add example for custom model provider propagation (#663)

### LINKED ISSUE #663: Custom model provider ignored when using agents as tools
### Please read this first

- **Have you read the custom model provider docs, including the 'Common issues' section?** [Model provider docs](https://openai.github.io/openai-agents-python/models/#using-other-llm-providers)
yes

yes

### Describe the question
When using a custom model provider that is passed as part of the `RunConfig` in `Runner.run`, the model provider is ignored when other agents are used as tools within a triage agent. The reason is that the `Agent.as_tool` function creates its own call to `Runner.run`, which does not include the current run config. A fix would be to add an additional optional argument `run_config` to `Agent.as_tool`. Alternatively, if we really just care about the model provider, we could have a `model_provider` parameter in the `Agent.as_tool` signature instead of the proposed `run_config` argument. 

The current implementation just ignores the model provider called in the initial `Runner.run` and switches to the default `OpenAIResponsesModel`.

### Debug information
- Agents SDK version: 0.0.13
- Python version 3.11.11

### Repro steps
Use the following script together with an OpenRouter API key to reproduce (or you could change it to an OllamaProvider so that you don't need an OpenRouter API key:

```python
import asyncio
import os
from agents import (
    Agent,
    ItemHelpers,
    MessageOutputItem,
    Model,
    ModelProvider,
    OpenAIChatCompletionsModel,
    RunConfig,
    Runner,
    set_tracing_disabled,
)


client = AsyncOpenAI(base_url="https://openrouter.ai/api/v1", api_key=os.getenv("OPENROUTER_API_KEY"))
set_tracing_disabled(disabled=True)


class OpenRouterModelProvider(ModelProvider):
        return OpenAIChatCompletionsModel("meta-llama/llama-3.3-70b-instruct", openai_client=client)


spanish_agent = Agent(
    name="spanish_agent",
    instructions="You translate the user's message to Spanish",
    handoff_description="An english to spanish translator",
)


orchestrator_agent = Agent(
    name="orchestrator_agent",
    instructions=(
        "You are a translation agent. You use the tools given to you to translate."
        "If asked for multiple translations, you call the relevant tools in order."
        "You never translate on your own, you always use the provided tools."
    ),
    tools=[
        spanish_agent.as_tool(
            tool_name="translate_to_spanish",
            tool_description="Translate the user's message to Spanish",
        ),
    ],
)

synthesizer_agent = Agent(
    name="synthesizer_agent",
    instructions="You inspect translations, correct them if needed, and produce a final concatenated response.",
)


async def main():
    msg = input("Hi! What would you like translated, and to which languages? ")

    orchestrator_result = await Runner.run(
        orchestrator_agent, msg, run_config=RunConfig(model_provider=OpenRouterModelProvider())
    )

    for item in orchestrator_result.new_items:
        if isinstance(item, MessageOutputItem):
            text = ItemHelpers.text_message_output(item)
            if text:
                print(f"  - Translation step: {text}")

    synthesizer_result = await Runner.run(synthesizer_agent, orchestrator_result.to_input_list())

    print(f"\n\nFinal response:\n{synthesizer_result.final_output}")


if __name__ == "__main__":
    asyncio.run(main())
```

This results in the following exception if no OpenAI api key is set:

```bash
openai.OpenAIError: The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable
```

### Expected behavior

I would have expected that the `Agent.as_tool` function would still use the same run config as the current agent is using during the run. If not, I would have expected that I can pass a run config to `as_tool` such that the custom model provider is still used.

furkankoykiran:
### What this PR does / why we need it
This PR adds an example demonstrating the behavior fixed in #2463 where a custom `ModelProvider` specified via `RunConfig` is correctly inherited when using an agent as a tool. 

Since the core engineering fix to `Agent.as_tool()` was already merged in #2463, this documentation addition serves to close out the remaining gap from #663 by giving users a runnable example of this requested pattern.

### Changes
- Adds `examples/agent_patterns/agents_as_tools_custom_provider.py` which shows how a custom `ModelProvider` is passed to the orchestrator agent and properly inherited by nested agent-tools.
- Updates `examples/agent_patterns/README.md` to reference the new example.

Fixes #663

### DIFF
--- a/examples/agent_patterns/README.md
+++ b/examples/agent_patterns/README.md
@@ -30,6 +30,7 @@ For example, you could model the translation task above as tool calls instead: r
 See the [`agents_as_tools.py`](./agents_as_tools.py) file for an example of this.
 See the [`agents_as_tools_streaming.py`](./agents_as_tools_streaming.py) file for a streaming variant that taps into nested agent events via `on_stream`.
 See the [`agents_as_tools_structured.py`](./agents_as_tools_structured.py) file for a structured-input variant using `Agent.as_tool()` parameters.
+See the [`agents_as_tools_custom_provider.py`](./agents_as_tools_custom_provider.py) file for an example using a custom model provider via `RunConfig` — the provider is automatically inherited by all nested agent-tool calls.
 
 ## LLM-as-a-judge
 
--- a/examples/agent_patterns/agents_as_tools_custom_provider.py
+++ b/examples/agent_patterns/agents_as_tools_custom_provider.py
@@ -1,0 +1,150 @@
+"""
+Example: agents-as-tools with a custom model provider (issue #663)
+
+This example demonstrates that a custom ModelProvider passed via RunConfig is
+correctly inherited by nested agent-tool calls.  Previously, the model provider
+was silently ignored inside Agent.as_tool(), causing the nested agents to fall
+back to the default OpenAI provider and raise an error when no OPENAI_API_KEY
+was set.
+
+To run with a custom provider (e.g. OpenRouter, Ollama, any OpenAI-compatible
+endpoint), export the following environment variables before running:
+
+    export EXAMPLE_BASE_URL="https://openrouter.ai/api/v1"
+    export EXAMPLE_API_KEY="<your-key>"
+    export EXAMPLE_MODEL_NAME="meta-llama/llama-3.3-70b-instruct"
+
+Then:
+    python -m examples.agent_patterns.agents_as_tools_custom_provider
+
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for sending this patch! However, we already have a similar example at https://github.com/openai/openai-agents-python/blob/main/examples/model_providers/custom_example_provider.py, so we won't have this one.

PR #2551, opened 2026-02-26
Add AgentMD support for AGENTS.md validation in CI

iedayan:
### Summary

Adds [AgentMD](https://agentmd.vercel.app/) to validate `AGENTS.md` in CI so the contributor guide stays parseable and agent-ready, and drift (e.g. renamed Makefile targets or broken code blocks) is caught before merge.

**Changes:**

- **`.github/workflows/agentmd.yml`** — New workflow that runs on push to `main` and on all pull requests. Uses the official AgentMD action with `command: validate` and `path: .` per their docs. Includes a note that the AgentMD GitHub App may need to be installed if the action fails to load (free for 3 repos).
- **`AGENTS.md`** — New "AgentMD" subsection under Pull Request & Commit Guidelines describing use of AgentMD in CI and how to use the CLI or GitHub App for readiness score / local use.

### Test plan

- Confirmed workflow YAML is valid and follows existing patterns (same checkout pin as other workflows).
- No runtime code or test changes; only CI and docs. `AGENTS.md` and `.github/` are repo-meta, so full `$code-change-verification` was not run per AGENTS.md.

### Checks

### DIFF
--- a/.github/workflows/agentmd.yml
+++ b/.github/workflows/agentmd.yml
@@ -1,0 +1,25 @@
+# Validate AGENTS.md with AgentMD (https://agentmd.vercel.app/).
+# Ensures the contributor guide is parseable and agent-ready.
+# If the action fails to load, install the AgentMD GitHub App (free for 3 repos).
+name: AgentMD
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sending this! While we don't have this change in this repo, your tool could be useful for many developers.

PR #2556, opened 2026-02-27
test(mcp): assert runner emits tool_call_output_item on MCP failures

weiguangli-io:
## Summary
This is a follow-up to #2554, adjusted based on maintainer feedback.

Instead of only asserting that the run continues, this test now explicitly verifies that runner-level MCP tool failures are surfaced as a generated `tool_call_output_item`.

## What changed
- Added `test_runner_emits_mcp_error_tool_call_output_item` in `tests/mcp/test_runner_calls_mcp.py`.
- Introduced a crashing fake MCP server for deterministic failure behavior.
- Asserted both sync and streaming paths:
  - a `tool_call_output_item` is produced
  - the item output matches the expected error text from `default_tool_error_function`
  - the run still continues to the next turn (`final_output == "done"`)

## Validation
Locally run:
- `env -u all_proxy -u ALL_PROXY -u http_proxy -u HTTP_PROXY -u https_proxy -u HTTPS_PROXY -u no_proxy -u NO_PROXY uv run --with pytest pytest -q tests/mcp/test_runner_calls_mcp.py`
- `uv run --with ruff ruff check tests/mcp/test_runner_calls_mcp.py`
- `uv run --with mypy mypy tests/mcp/test_runner_calls_mcp.py`

All passed locally.

### DIFF
--- a/tests/mcp/test_runner_calls_mcp.py
+++ b/tests/mcp/test_runner_calls_mcp.py
@@ -3,7 +3,15 @@ import json
 import pytest
 from pydantic import BaseModel
 
-from agents import Agent, ModelBehaviorError, Runner, UserError
+from agents import (
+    Agent,
+    ModelBehaviorError,
+    RunContextWrapper,
+    Runner,
+    UserError,
+    default_tool_error_function,
+)
+from agents.exceptions import AgentsException
 
 from ..fake_model import FakeModel
 from ..test_responses import get_function_tool_call, get_text_message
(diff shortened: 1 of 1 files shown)


seratch (approved):
@OiPunk Thank you for working on this. I think this is worth adding, so I'll merge it.

That said, we're trying to keep review effort focused, so for future test-only PRs, please prioritize cases that cover critical or user-visible paths.

PR #2559, opened 2026-02-28
perf: use deque for BFS queue and voice task ordering

giulio-leone:
## Problem

Two hot paths use `list.pop(0)` which is **O(n)** per removal:

1. **`run_state.py` — `_build_agent_map()`**: BFS traversal over agent handoff graph. Each `.pop(0)` shifts all remaining elements.
2. **`voice/result.py` — `_ordered_tasks`**: Voice pipeline drains ordered task queues front-to-back. Each `.pop(0)` is O(n).

## Solution

Switch both to `collections.deque` with `.popleft()` for **O(1)** front removal.

## Changes

- `src/agents/run_state.py`: Import `deque`, use `deque([initial_agent])` and `.popleft()`
- `src/agents/voice/result.py`: Import `deque`, type `_ordered_tasks` as `deque`, use `.popleft()`

## Testing

- 153 run_state tests pass (2 consecutive runs)
- Syntax verified via `ast.parse()`

### DIFF
--- a/src/agents/voice/result.py
+++ b/src/agents/voice/result.py
@@ -53,9 +53,9 @@ class StreamedAudioResult:
         self._turn_text_buffer = ""
         self._queue: asyncio.Queue[VoiceStreamEvent] = asyncio.Queue()
         self._tasks: list[asyncio.Task[Any]] = []
-        self._ordered_tasks: deque[
-            asyncio.Queue[VoiceStreamEvent | None]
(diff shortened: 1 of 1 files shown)


seratch:
Can you fix the lint error?

giulio-leone:
Fixed — reformatted `src/agents/voice/result.py` with `ruff format`. Thanks for flagging!

seratch (approved):

PR #2562, opened 2026-02-28
feat: expose `tool_call_id` on `RunResult` for agent-as-tool pattern

### LINKED ISSUE #2247: Access to `call_id` in `custom_output_extractor` for agent-as-tool pattern
### Problem

When using the agent-as-tool pattern, `custom_output_extractor` doesn't have access to the tool call's `call_id`, making it not possible to associate the nested agent's `RunResult` (e.g., `to_input_list()`) back to the original `ToolCallItem`/`ToolCallOutputItem` for debugging.

**Current flow:**
1. Agent B is defined and has a number of tools. Agent A is defined and uses Agent B as a tool.
2. Agent A calls Agent B as a tool → creates `ToolCallItem` with `call_id`
3. Agent B runs, calls its own tools, and creates its final output → produces `RunResult`
4. `custom_output_extractor` receives the `RunResult` object, but it has no association to the original `call_id`

As a result, the nested `RunResult` is lost. This way, there is no visibility over what the agent-as-tool is doing under the hood.

### Proposed Solution

Pass the `call_id` through the `RunResult`'s `context_wrapper` so `custom_output_extractor` can access it. 

This would enable something like this:

```python
async def my_custom_output_extractor(run_result: RunResult) -> str:
    call_id = run_result.context_wrapper.context.tool_call_id  # Currently not available
    my_custom_session.attach_nested_run_result(call_id, run_result.to_input_list())
    return run_result.final_output
```

Would love to here if there is a better solution, and if there is a temporary workaround for this problem.

Br1an67:
### Summary

Add a read-only `tool_call_id` property to `RunResultBase` so that `custom_output_extractor` callbacks can access the parent tool call ID in the agent-as-tool pattern.

Currently, when using `Agent.as_tool()` with a `custom_output_extractor`, the extractor receives a `RunResult` whose `context_wrapper` is actually a `ToolContext` containing `tool_call_id`. However, `context_wrapper` is typed as `RunContextWrapper[Any]`, so users must perform an `isinstance` check to access it. The new `tool_call_id` property provides a clean, discoverable API that returns the ID directly (or `None` when the run was not started via `as_tool()`).

### Test plan

- Added unit tests in `tests/test_result_cast.py`:
  - `test_tool_call_id_returns_id_when_context_is_tool_context` — verifies the property returns the correct ID when `context_wrapper` is a `ToolContext`
  - `test_tool_call_id_returns_none_when_context_is_plain_wrapper` — verifies `None` is returned for non-agent-tool runs
- Added integration test in `tests/test_agent_as_tool.py`:
  - `test_agent_as_tool_extractor_can_access_tool_call_id` — verifies `tool_call_id` is accessible inside a `custom_output_extractor` callback
- All existing tests pass (`uv run pytest tests/test_agent_as_tool.py tests/test_result_cast.py` — 43 passed)

### Issue number

Closes #2247

### Checks

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -36,7 +36,7 @@ jobs:
           fi
       - name: Setup uv
         if: steps.docs-only.outputs.skip != 'true'
-        uses: astral-sh/setup-uv@803947b9bd8e9f986429fa0c5a41c367cd732b41
+        uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098
         with:
           enable-cache: true
       - name: Install dependencies
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -10,7 +10,7 @@ jobs:
       issues: write
       pull-requests: write
     steps:
-      - uses: actions/stale@997185467fa4f803885201cee163a9f38240193d
+      - uses: actions/stale@b5d41d4e1d5dceea10e7104786b73624c18a190f
         with:
           days-before-issue-stale: 7
           days-before-issue-close: 3
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - name: Setup uv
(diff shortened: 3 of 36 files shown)


seratch:
Thank you again for sending this patch. After reviewing your changes, I came to think exposing tool context would be more robust enhancement for people who may need other properties in the context object. Closing this PR in favor of https://github.com/openai/openai-agents-python/pull/2575

PR #2573, opened 2026-03-01
Fix #618 MCPServerStdio leaked semaphores warning at shutdown

### LINKED ISSUE #618: Resource tracker warning (leaked semaphores) with MCPServerStdio
### Debug information
- Agents SDK version: 0.0.12
- Python 3.13.1

### Describe the bug
When using an external MCP plugin with the OpenAI Agents SDK (`MCPServerStdio` in this case), I consistently receive a `resource_tracker: There appear to be XX leaked semaphore objects to clean up at shutdown` warning message during the program's shutdown. The exact number of leaked objects (e.g., 24) varies.

### Steps to reproduce
Use the following code snippet which initializes a custom OpenAI client and then attempts to run an agent configured with an `MCPServerStdio` plugin.

### Code Snippet
```python
import asyncio
from agents import set_default_openai_client, set_default_openai_api, set_tracing_disabled
from agents.mcp import MCPServerStdio

# --- Replace with your actual values ---
OPENAI_BASE_URL = "YOUR_CUSTOM_BASE_URL" # e.g., "http://localhost:11434/v1" or similar
OPENAI_API_KEY = "YOUR_API_KEY" # Can be "ollama" or similar if using local endpoint
MODEL_NAME = "YOUR_MODEL_NAME" # e.g., "llama3" or "gpt-4o"
# --- End of replacements ---

async def main():
    # Create custom OpenAI client
    custom_client = AsyncOpenAI(base_url=OPENAI_BASE_URL, api_key=OPENAI_API_KEY)

    # Required configuration for using a custom OpenAI API endpoint
    set_default_openai_api("chat_completions") # Or other API type if needed
    set_default_openai_client(custom_client)
    set_tracing_disabled(True) # Disable tracing if not needed

    # Get Python executable path
    current_dir = os.path.dirname(os.path.abspath(__file__))
    python_executable = sys.executable
    soc_push_script = os.path.join(current_dir, "plugins", "soc_push_script.py") # Assuming this script exists and works as an MCP plugin

    # Initialize the MCP server plugin using MCPServerStdio
    # Make sure soc_push_script.py implements the MCP protocol correctly
    mcp_soc_push = MCPServerStdio(
        name="soc_push_script",
        params={
            "command": python_executable,
            "args": [soc_push_script],
        })

    # Run the agent with the MCP plugin
    try:
        async with mcp_soc_push as push_tool: # Use the plugin as a tool
            agent = Agent(
                name="test",
                instructions='test',
                model=MODEL_NAME,
                mcp_servers=[push_tool], # Add the MCP tool to the agent's servers
            )
            print("Starting agent run...")
            result = await Runner.run(starting_agent=agent, input='')
            print("Agent run finished.")
            print("Final Output:", result.final_output)
    finally:
         # Explicitly attempt cleanup if necessary, though the warning suggests leakage
         print("Attempting cleanup (may not prevent warning)...")
         # Specific MCP cleanup methods might be needed depending on implementation
         pass # Add cleanup logic here if known

# Assuming this script is run directly
if __name__ == "__main__":
    asyncio.run(main())

```
*Note: The `soc_push_script.py` file is not provided, but it is assumed to be a functional MCP plugin script that interacts via standard I/O.*

### Console Output
```
/usr/local/Cellar/python@3.13/3.13.1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/multiprocessing/resource_tracker.py:276: UserWarning: resource_tracker: There appear to be 24 leaked semaphore objects to clean up at shutdown: {'/mp-t9hi4oa_', '/mp-pbo_pje7', '/mp-_1jipfl5', '/mp-04d4uokn', '/mp-jw_gfhjk', '/mp-8122t3qm', '/mp-tjg6llji', '/mp-75j5swhj', '/mp-oprtjvq9', '/mp-_9c_6tt4', '/mp-o6ld1gx6', '/mp-vfh4ncjt', '/mp-6pf__aj0', '/mp-h8x7h4uu', '/mp-_heqxtvk', '/mp-uvzgo0ft', '/mp-gqm7a9dj', '/mp-j09dabgf', '/mp-w30c1pxu', '/mp-4gzlcmxm', '/mp-b5004bcn', '/mp-d1l4chll', '/mp-we2hfv61', '/mp-t1hr16ma'}
  warnings.warn(
```

dcol91863:
Fixes #618. Resets the session's AsyncExitStack on cleanup so it successfully releases resources and allows reconnections.

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -745,6 +745,8 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
                     logger.error(f"Error cleaning up server: {e}")
             finally:
                 self.session = None
+                self.exit_stack = AsyncExitStack()
+                self.server_initialize_result = None
 
 
 class MCPServerStdioParams(TypedDict):
--- a/tests/mcp/test_connect_disconnect.py
+++ b/tests/mcp/test_connect_disconnect.py
@@ -67,3 +67,40 @@ async def test_manual_connect_disconnect_works(
 
     await server.cleanup()
     assert server.session is None, "Server should be disconnected"
+
+
+@pytest.mark.asyncio
+@patch("mcp.client.stdio.stdio_client", return_value=DummyStreamsContextManager())
+@patch("mcp.client.session.ClientSession.initialize", new_callable=AsyncMock, return_value=None)
+@patch("mcp.client.session.ClientSession.list_tools")
+async def test_manual_reconnect_works(
+    mock_list_tools: AsyncMock, mock_initialize: AsyncMock, mock_stdio_client
+):
+    """Test that the async context manager works."""
+    server = MCPServerStdio(
+        params={
+            "command": tee,
+        },
+        cache_tools_list=True,
+    )
+
+    tools = [
+        MCPTool(name="tool1", inputSchema={}),
+        MCPTool(name="tool2", inputSchema={}),
+    ]
+
+    mock_list_tools.return_value = ListToolsResult(tools=tools)
+
+    assert server.session is None, "Server should not be connected"
+
+    await server.connect()
+    assert server.session is not None, "Server should be connected"
+
+    await server.cleanup()
+    assert server.session is None, "Server should be disconnected"
+    
+    await server.connect()
+    assert server.session is not None, "Server should be reconnected"
+    
+    await server.cleanup()
+    assert server.session is None, "Server should be disconnected again"


seratch:
@codex review again

seratch:
Thanks for sending this patch. I am also looking into the issue reported in #618, and I do not think the suggestion here is the optimal solution for it. I will continue exploring possible solutions for the issue, but the direction in this PR is probably not the one we would like to take.

dcol91863:
Do you have any other solutions in mind?

dcol91863:
I have tried looking for a more efficient solution:

What the PR does:

The PR adds two lines to the finally block of cleanup() in _MCPServerWithClientSession:

finally:
    self.session = None
    self.exit_stack = AsyncExitStack()      # <-- added
    self.server_initialize_result = None   # <-- added

This enables reconnection by replacing the exhausted AsyncExitStack with a fresh one. Without this, calling connect() after cleanup() pushes new context managers into an already-closed stack, which silently fails.

Issues:

1. It doesn't fix the actual semaphore leak. The resource_tracker warning comes from POSIX named semaphores (/mp-...) that aren't properly released at shutdown. The PR's fix only enables reconnection — it does nothing about when or how the subprocess's resources are torn down.

2. It mutates state in finally unconditionally. Even if exit_stack.aclose() raises an exception mid-way (leaving resources partially cleaned up), the old stack is discarded and replaced with a fresh one. This could hide bugs and makes the state machine harder to reason about.

3. The wrong place in the lifecycle. The exit_stack is created in __init__ and reset in cleanup(). The correct mental model is that a stack should live for exactly one connection lifetime — from connect() to cleanup().

More efficient solution
The cleanest fix is to create a fresh AsyncExitStack at the start of connect() rather than resetting it at the end of cleanup():

async def connect(self):
    """Connect to the server."""
    # Fresh stack for each connection lifecycle; enables reconnect without
    # mutating state in the cleanup finally-block.
    self.exit_stack = AsyncExitStack()
    connection_succeeded = False
    try:
        transport = await self.exit_stack.enter_async_context(self.create_streams())
        ...

This is better because:

The lifecycle is explicit: connect() creates resources, cleanup() releases them.
cleanup() only needs session = None — no extra state resets in a finally block.
Reconnection works naturally: the next connect() starts a clean slate.
The old exit_stack is still properly closed by cleanup() before the reference is replaced.
The semaphore leak root cause (separate issue)
The multiprocessing.resource_tracker warning about /mp-* semaphores is not addressed by the PR at all. These named POSIX semaphores are created by Python's multiprocessing infrastructure — not by asyncio or anyio memory streams.

The most likely cause: when mcp.client.stdio.stdio_client spawns the subprocess via anyio.open_process(), if cleanup is not awaited properly before interpreter shutdown (e.g., asyncio.run() completes and the loop closes while the subprocess is still tearing down), the semaphores registered with the resource tracker are never unregistered.

A better fix for that specific problem would be in MCPServerStdio.create_streams() — wrapping stdio_client in a context manager that explicitly terminates the subprocess and drains pipes before returning, rather than relying on the garbage collector or resource tracker at interpreter exit. This is a change that would need to go into the mcp library or be wrapped at the SDK level.

In short: the PR solves the wrong problem (reconnection, not leak), does so in a fragile way (state reset in finally), and seratch is right to look for a better direction. The reconnection fix belongs at the top of connect(), and the semaphore leak needs a proper subprocess teardown guarantee.

seratch:
I explored a solution and have a local work-in-progress branch. However, I found that a fundamental fix for this issue could require substantial changes across the SDK, and my branch is still in progress. I may be able to find a simpler solution, but at this point, the issue does not look as straightforward as we initially expected.

PR #2577, opened 2026-03-02
Update transcription model options in config.py

ioannes041424:
Added "gpt-4o-transcribe-diarize" to the input Transcription models


ioannes041424:
@seratch openai team could you guys look into this. I'm using agents sdk for voice agents with realtime api. and notice that the new **gpt-4o-transcribe-diarize** model is not added in the transcription model config

seratch:
[Our documentation](https://developers.openai.com/api/docs/guides/realtime-transcription) does not yet mention the model in the list, so we won't add it to the literals at this moment. The property accepts any string value as well, so it shouldn't prevent you from using any new models.

PR #2588, opened 2026-03-03
Abhisharma/rsi 224 caching fix

Abh1Sharma:


seratch:
Thanks for the PR. I don’t think we should merge this as-is.

For this class of change, our baseline expectation is that the provider-specific behavior should be handled in LiteLLM first, rather than in openai-agents. Some of the ideas here may be worth revisiting later in a much narrower form, but the first step should be an upstream LiteLLM issue/PR that clearly defines the expected behavior and support matrix.

For future work, please start by opening a detailed issue here before sending a PR. That issue should include the concrete problem, expected behavior, affected model/provider versions, minimal repro steps, and any relevant upstream LiteLLM links. That will make it much easier to decide whether anything needs to happen in this repo at all.

PR #2590, opened 2026-03-04
Handle unknown model prefixes in MultiProvider

### LINKED ISSUE #2492: Unknown prefix when using openrouter models

romgenie:
### Summary

Fix #2492 by changing `MultiProvider.get_model()` to treat unknown model prefixes as part of the model id and route through the OpenAI fallback. This avoids `Unknown prefix` errors for namespaced model names like `openrouter/openai/gpt-4o` used by OpenAI-compatible providers.

### Test plan

- `make format`
- `make lint`
- `make mypy`
- `make tests`

### Issue number

Closes #2492

### Checks

### DIFF
--- a/src/agents/models/multi_provider.py
+++ b/src/agents/models/multi_provider.py
@@ -144,12 +144,19 @@ class MultiProvider(ModelProvider):
         Returns:
             A Model.
         """
+        original_model_name = model_name
         prefix, model_name = self._get_prefix_and_model_name(model_name)
 
+        # Route known prefixes to dedicated providers.
         if prefix and self.provider_map and (provider := self.provider_map.get_provider(prefix)):
             return provider.get_model(model_name)
-        else:
+        if prefix in {"openai", "litellm", None}:
             return self._get_fallback_provider(prefix).get_model(model_name)
+
+        # Unknown prefixes are treated as part of the model name for OpenAI-compatible providers.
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing this idea. After reviewing your changes here, I found this PR still doesn't resolve some patterns. I've come up with https://github.com/openai/openai-agents-python/pull/2593 for resolving those patterns as well, plus keep the backward-compatibility. Thanks again for taking the time here.

PR #2601, opened 2026-03-05
fix: correctly serialize big integers in traces (fixes #2094)

### LINKED ISSUE #2094: Traces dashboard shows big integer arguments incorrectly in tool calling traces
### Describe the bug

```
import asyncio

import sys


@function_tool
    print(f"Adding {a} and {b}")
    return a + b


agent = Agent(
    name="Addition Agent",
    instructions="""
    """,
    tools=[add],
)


async def main():
    a = int(sys.argv[1])
    b = int(sys.argv[2])
    print(result.final_output)


if __name__ == "__main__":
    asyncio.run(main())
```

**Input 1:**

```
10000000123456790
```

Image

**Input 2:**

```
10000000123456791
```

Image

### Debug information
- Agents SDK version: 0.5.1
- Python version: 3.13.7

hkc5:
## Summary

This PR fixes an issue where big integers (beyond JavaScript's Number.MAX_SAFE_INTEGER) were incorrectly displayed in the traces dashboard due to precision loss during JSON serialization.

## Problem

When passing big integers (like 10000000000000001) to tool functions, the library correctly passed precise integers to the function, but the traces dashboard showed truncated/incorrect numbers. This happened because integers were serialized as JSON numbers, which JavaScript interprets as float64, losing precision for values beyond 2^53 - 1 (9007199254740991).

Related issue: google/adk-python#3592

## What Changed

1. Added constants for JavaScript's safe integer range (_JS_MAX_SAFE_INTEGER and _JS_MIN_SAFE_INTEGER)
2. Modified `_sanitize_json_compatible_value()` to convert integers beyond the safe range to strings
3. Updated `_truncate_span_field_value()` to always sanitize values for JSON compatibility, not just when they exceed size limits
4. Added comprehensive tests for big integer serialization

## Validation

- Added two new tests in `tests/test_trace_processor.py`:
  - `test_sanitize_json_compatible_value_preserves_bigint_precision` - Tests direct sanitization
  - `test_sanitize_for_openai_tracing_api_preserves_bigint_in_span_data` - End-to-end test
- All 38 existing trace processor tests pass
- Code passes black and ruff linting

## Security Impact

No security impact. This is a data serialization fix that preserves precision. Converting big integers to strings is safer than losing precision.

## Privacy

No privacy impact. No PII is affected by this change.

## Rollback

This change can be safely rolled back by reverting the commit. The worst case is that big integers will again be displayed with precision loss in the traces dashboard (but the actual function calls will still work correctly).

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -40,6 +40,9 @@ class BackendSpanExporter(TracingExporter):
         }
     )
     _UNSERIALIZABLE = object()
+    # JavaScript's Number.MAX_SAFE_INTEGER: integers beyond this lose precision in JS
+    _JS_MAX_SAFE_INTEGER = 9007199254740991  # 2^53 - 1
+    _JS_MIN_SAFE_INTEGER = -9007199254740991  # -(2^53 - 1)
 
     def __init__(
         self,
@@ -144,7 +147,9 @@ class BackendSpanExporter(TracingExporter):
             while True:
                 attempt += 1
                 try:
-                    response = self._client.post(url=self.endpoint, headers=headers, json=payload)
+                    response = self._client.post(
+                        url=self.endpoint, headers=headers, json=payload
+                    )
 
                     # If the response is successful, break out of the loop
                     if response.status_code < 300:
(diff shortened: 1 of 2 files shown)


hkc5:
@maintainer - would appreciate your review when you have time

seratch:
Thanks for sending this patch! We're aware that this is an issue on the server-side https://github.com/openai/openai-agents-python/issues/2094 and it's not yet resolved.

PR #2612, opened 2026-03-06
Add local shell skill example

ihower:
This PR adds an example showing how to use skills with local shell mode.

This complements the existing container shell examples:

- examples/tools/container_shell_inline_skill.py
- examples/tools/container_shell_skill_reference.py

Reference documentation: Use skills with local shell mode https://developers.openai.com/api/docs/guides/tools-skills/#use-skills-with-local-shell-mode

Note: The cookbook example is incorrect: https://developers.openai.com/cookbook/examples/skills_in_api/
It uses `"skill_reference"` for local shell mode, which is not valid and causes an API error.


seratch:
@codex review

PR #2619, opened 2026-03-06
fix(memory): make SQLiteSession satisfy Session protocol

### LINKED ISSUE #2618: SQLiteSession is not assignable to Session protocol due to invariant session_settings type mismatch
### Describe the bug
Passing a `SQLiteSession` instance to `Runner.run(session=...)` produces a Pylance type error:
```
Argument of type "SQLiteSession" cannot be assigned to parameter "session" of type "Session | None" in function "run"
  Type "SQLiteSession" is not assignable to type "Session | None"
    "SQLiteSession" is incompatible with protocol "Session"
      "session_settings" is invariant because it is mutable
      "session_settings" is an incompatible type
        Type "SessionSettings" is not assignable to type "SessionSettings | None"
          "SessionSettings" is not assignable to "None"
    "SQLiteSession" is not assignable to "None"Pylance[reportArgumentType](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportArgumentType.md)
(parameter) session: SQLiteSession
```

The `Session` protocol in `session.py` declares:
```python
session_settings: SessionSettings | None = None
```
But `SQLiteSession.__init__` in `sqlite_session.py` assigns:
```python
self.session_settings = session_settings or SessionSettings()
```
This means `SQLiteSession.session_settings` is always `SessionSettings` (never `None`), while the protocol expects `SessionSettings | None`. 

### Debug information
SDK version: 0.10.5
Python version: 3.14.2
Pylance (vscode) - 2026.1.1

### Repro steps

```python
from agents import Agent, Runner, SQLiteSession

agent = Agent(name="Assistant", instructions="Reply concisely.")
session = SQLiteSession("test_session")

async def main():
    result = await Runner.run(
        agent,
        "Hello",
        session=session,  # Pylance error here
    )
    print(result.final_output)
```

### Expected behavior
SQLiteSession should be assignable to Session without a cast.

lavish0000:
Fixes #2618

Summary
- keep `Session.session_settings` as a writable protocol attribute
- declare `session_settings: SessionSettings | None` explicitly on the concrete session implementations so they structurally match the protocol under Pylance/basedpyright
- add a regression test that passes `SQLiteSession` through a `Session`-typed parameter
- declare `session_settings` explicitly on the example/test `Session` subclasses that inherit from `Session` directly

Why
- Pylance reports `SQLiteSession` as incompatible with `Session` because mutable protocol attributes are invariant
- the concrete session classes assign `SessionSettings()` in `__init__`, which causes the attribute to be treated as a narrower concrete type
- explicitly annotating the concrete implementations fixes the typing mismatch without changing runtime behavior or breaking direct subclasses of `Session`

Scope
- no runtime behavior changes in `SQLiteSession`, `Runner`, or session persistence flows
- no protocol-property conversion
- no changes to session storage semantics

Validation
- `python3 -m uv run pytest tests/test_session_protocol_typing.py tests/test_session.py -k 'session_protocol or session_settings_default or session_settings_constructor'`
- `python3 -m uv run ruff check src/agents/memory/session.py src/agents/memory/sqlite_session.py src/agents/memory/openai_conversations_session.py src/agents/extensions/memory/redis_session.py src/agents/extensions/memory/sqlalchemy_session.py src/agents/extensions/memory/dapr_session.py examples/memory/file_session.py tests/utils/simple_session.py tests/test_session_protocol_typing.py`
- `python3 -m uv run ruff format --check src/agents/memory/session.py src/agents/memory/sqlite_session.py src/agents/memory/openai_conversations_session.py src/agents/extensions/memory/redis_session.py src/agents/extensions/memory/sqlalchemy_session.py src/agents/extensions/memory/dapr_session.py examples/memory/file_session.py tests/utils/simple_session.py tests/test_session_protocol_typing.py`
- validated the issue repro snippet locally with `uvx basedpyright` (`0 errors`, only unrelated `Any` warnings from `print(...)`)
- `make sync`
- `make format`
- `make lint`
- `make mypy`
- `make tests`

Risks/Notes
- the concrete annotation was applied to the built-in session implementations that eagerly default `session_settings`, plus the repo-local `Session` subclasses used in tests/examples, to keep both runtime behavior and repo-wide type checks intact

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ license = "MIT"
 authors = [{ name = "OpenAI", email = "support@openai.com" }]
 dependencies = [
     "openai>=2.25.0,<3",
-    "pydantic>=2.12.3, <3",
+    "pydantic>=2.12.2, <3",
     "griffe>=1.5.6, <2",
     "typing-extensions>=4.12.2, <5",
     "requests>=2.0, <3",
--- a/uv.lock
+++ b/uv.lock
@@ -1965,7 +1965,7 @@ requires-dist = [
     { name = "mcp", marker = "python_full_version >= '3.10'", specifier = ">=1.19.0,<2" },
     { name = "numpy", marker = "python_full_version >= '3.10' and extra == 'voice'", specifier = ">=2.2.0,<3" },
     { name = "openai", specifier = ">=2.25.0,<3" },
-    { name = "pydantic", specifier = ">=2.12.3,<3" },
+    { name = "pydantic", specifier = ">=2.12.2,<3" },
     { name = "redis", marker = "extra == 'redis'", specifier = ">=7" },
     { name = "requests", specifier = ">=2.0,<3" },
     { name = "sqlalchemy", marker = "extra == 'sqlalchemy'", specifier = ">=2.0" },


src/agents/memory/session.py
```diff
@@ -19,7 +19,11 @@ class Session(Protocol):
     """
 
     session_id: str
-    session_settings: SessionSettings | None = None
+
+    @property
+    def session_settings(self) -> SessionSettings | None:
```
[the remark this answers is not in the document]
lavish0000: Good catch. I reverted the read-only protocol change and kept `Session.session_settings` writable. The follow-up now fixes the typing issue by explicitly annotating the concrete session implementations with `SessionSettings | None`, which preserves direct `Session` subclasses that assign `self.session_settings = ...` in `__init__`. I revalidated the issue repro with basedpyright (`0 errors`) and reran `make sync`, `make format`, `make lint`, `make mypy`, and `make tests` on the updated branch.

lavish0000:
@codex review

tests/test_session_protocol_typing.py
```diff
@@ -0,0 +1,13 @@
+from agents import Session, SQLiteSession
```
seratch, line 1: The new regression test is aimed at the right area, but I don’t think it actually guards the failure this PR fixes. `tests/test_session_protocol_typing.py` still passes under our current mypy gate on both `origin/main` and this branch, so it would not catch a regression if the writable `session_settings` mismatch came back. 

The real before/after difference shows up in pyright/Pylance, where `SQLiteSession` was rejected as a `Session` because `session_settings` is a mutable protocol attribute. I don't think we should add pyright validation this time, but can you improve or remove this test?

seratch: I'd like to add additional validation on top of this, so I'll go ahead with this changeset for now and remove this test file after merging this PR.

PR #2638, opened 2026-03-08
fix(run): fail fast for server conversation + handoff history mutations

### LINKED ISSUE #2151: nest handoff: Handoff is incompatible with server-managed conversation

pandego:
### Summary
- add fail-fast validation for server-managed conversation mode when any reachable handoff applies `input_filter` or effective `nest_handoff_history`
- apply the same validation in both non-streamed and streamed run entry paths
- add regression tests for incompatible combinations in `Runner.run`

### Test plan
- `make format`
- `make lint`
- `make mypy` *(fails in this environment due missing optional stubs/deps like numpy/litellm/sqlalchemy; no new project-local mypy errors from this change)*
- `uv run pytest tests/test_agent_runner.py -k "history_mutating_handoffs or run_rejects_session_with_server_managed_conversation or run_streamed_rejects_session_with_server_managed_conversation"`

### Issue number
Closes #2151

### Checks

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -56,6 +56,7 @@ from .run_internal.agent_runner_helpers import (
     save_turn_items_if_needed,
     should_cancel_parallel_model_task_on_input_guardrail_trip,
     update_run_state_for_interruption,
+    validate_server_conversation_handoff_settings,
     validate_session_conversation_settings,
 )
 from .run_internal.approvals import approvals_from_step
@@ -441,6 +442,13 @@ class AgentRunner:
                 previous_response_id=previous_response_id,
                 auto_previous_response_id=auto_previous_response_id,
             )
+            validate_server_conversation_handoff_settings(
+                starting_agent,
+                run_config,
+                conversation_id=conversation_id,
+                previous_response_id=previous_response_id,
+                auto_previous_response_id=auto_previous_response_id,
+            )
             starting_input = run_state._original_input
             original_user_input = copy_input_items(run_state._original_input)
             prepared_input = normalize_resumed_input(original_user_input)
@@ -458,6 +466,13 @@ class AgentRunner:
 
             validate_session_conversation_settings(
                 session,
+                conversation_id=conversation_id,
+                previous_response_id=previous_response_id,
+                auto_previous_response_id=auto_previous_response_id,
+            )
+            validate_server_conversation_handoff_settings(
+                starting_agent,
+                run_config,
                 conversation_id=conversation_id,
                 previous_response_id=previous_response_id,
                 auto_previous_response_id=auto_previous_response_id,
--- a/src/agents/run_internal/agent_runner_helpers.py
+++ b/src/agents/run_internal/agent_runner_helpers.py
@@ -39,6 +39,7 @@ __all__ = [
     "ensure_context_wrapper",
     "finalize_conversation_tracking",
     "input_guardrails_triggered",
+    "validate_server_conversation_handoff_settings",
     "validate_session_conversation_settings",
     "resolve_trace_settings",
     "resolve_processed_response",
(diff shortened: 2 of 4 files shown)


seratch:
Thanks for sending this patch! However, as I mentioned at #2457, adding validation logic is not the one we'd like to have for resolving this issue.

PR #2652, opened 2026-03-11
fix(mcp): suppress multiprocessing resource tracker semaphore leak warning

Jah-yee:
When using MCPServerStdio, Python's multiprocessing resource tracker can emit warnings about leaked semaphore objects at shutdown. This is a known limitation of the stdio transport and not a bug in user code.

This fix suppresses the warning at module import time using `warnings.filterwarnings()`, which prevents the confusing warning from appearing during program shutdown.

Fixes: #618

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -2,8 +2,11 @@ from __future__ import annotations
 
 import abc
 import asyncio
+import atexit
 import inspect
+import multiprocessing
 import sys
+import warnings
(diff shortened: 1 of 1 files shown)


seratch:
I don't think ignoring the warning is the right solution here. I've also explored a viable solution for this issue, bu I found that eliminating this was way more complicated than I expected. We'll continue exploring a fix for it, but let me close this PR for now.

PR #2659, opened 2026-03-12
feat: add rejection_message parameter to RunState.reject()

### LINKED ISSUE #2658: Feature request: pass a specific rejection reason to the model when declining a tool call (RunState.reject)
### Problem

When using the Human-in-the-loop (HITL) feature, I can approve or reject a
pending tool call via `RunState.reject()`. However, there is currently no way to
tell the model **why** the call was rejected.

The only workaround is to set a global `RunConfig.tool_error_formatter` callback
at run-configuration time, but that callback receives no information about which
specific call was rejected or what the user actually said, so I cannot pass a
meaningful, per-call reason to the model.

This means the model always receives a generic message like
`"Tool execution was not approved."` — even when the user typed an explicit
reason such as:

- "Location access was denied because the user is in a restricted area."
- "This action requires manager approval; please escalate."
- "The requested file does not exist on this machine."

The model then has no context to adjust its next response or ask a better
follow-up question.

### Desired behavior

`RunState.reject()` should accept an optional `rejection_message` parameter so
that the specific reason can be forwarded to the model:

```python
state = result.to_state()
for interruption in result.interruptions:
    reason = ask_user_why_they_rejected(interruption)   # e.g. from a UI text box
    state.reject(
        interruption,
        rejection_message=reason,   # ← pass the actual reason
    )
result = await Runner.run(agent, state)
```

## Why this matters

In real HITL flows, the human operator often knows exactly why they are declining an action and wants the agent to understand that context so it can:

- Avoid repeating the same request
- Suggest an alternative approach
- Give the end-user an accurate explanation

Without this feature, all of that context is silently discarded at the rejection step.

## Proposed resolution priority

1. `rejection_message` kwarg on `reject()` — new, per-call (highest priority)
2. `RunConfig.tool_error_formatter` callback — existing, global fallback
3. SDK default: `"Tool execution was not approved."`

Jah-yee:
## Summary

Add optional `rejection_message` parameter to `RunState.reject()` method to allow passing a specific reason for tool call rejection to the model.

## Problem

When using the Human-in-the-loop (HITL) feature, rejecting a tool call via `RunState.reject()` currently only tells the model that execution was "not approved" — there is no way to communicate WHY the call was rejected.

This makes it difficult for the model to understand context and adjust its response accordingly.

## Solution

Add a new optional `rejection_message` parameter to `RunState.reject()`:

```python
state = result.to_state()
for interruption in result.interruptions:
    reason = ask_user_why_they_rejected(interruption)
    state.reject(
        interruption,
        rejection_message=reason,  # ← new parameter
    )
result = await Runner.run(agent, state)
```

## Changes

1. **run_config.py**: Add `rejection_message` field to `ToolErrorFormatterArgs`
2. **run_state.py**: Add `rejection_message` parameter to `RunState.reject()` method
3. **run_context.py**: Store rejection message in context and pass to formatter
4. **tool_execution.py**: Pass rejection message to tool error formatter

## Testing

- [ ] Test with custom tool error formatter
- [ ] Test without custom formatter (should use rejection_message if provided)
- [ ] Test serialization/deserialization of RunState

## Related

Fixes #2658

---

Good day! I hope this contribution helps. Please let me know if there are any issues or if the implementation needs adjustment. I am happy to make changes.

Warmly,
Jah-yee

### DIFF
--- a/src/agents/run_config.py
+++ b/src/agents/run_config.py
@@ -76,6 +76,9 @@ class ToolErrorFormatterArgs(Generic[TContext]):
     run_context: RunContextWrapper[TContext]
     """The active run context for the current execution."""
 
+    rejection_message: str | None = None
+    """The user-provided reason for rejecting the tool call (via RunState.reject())."""
+
 
 ToolErrorFormatter = Callable[[ToolErrorFormatterArgs[Any]], MaybeAwaitable[Optional[str]]]
 
--- a/src/agents/run_context.py
+++ b/src/agents/run_context.py
@@ -169,13 +169,28 @@ class RunContextWrapper(Generic[TContext]):
             approve=True,
         )
 
-    def reject_tool(self, approval_item: ToolApprovalItem, always_reject: bool = False) -> None:
-        """Reject a tool call, optionally for all future calls."""
+    def reject_tool(
+        self, approval_item: ToolApprovalItem, always_reject: bool = False, rejection_message: str | None = None
+    ) -> None:
+        """Reject a tool call, optionally for all future calls.
+        
+        Args:
+            approval_item: The tool approval item to reject.
+            always_reject: If True, reject all future calls to this tool without prompting.
+            rejection_message: Optional message explaining why the tool call was rejected.
+        """
         self._apply_approval_decision(
             approval_item,
             always=always_reject,
             approve=False,
         )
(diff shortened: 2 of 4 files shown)


seratch:
Thanks for sending this patch! Closing this in favor of https://github.com/openai/openai-agents-python/pull/2660

PR #2670, opened 2026-03-14
feat: #2669 add opt-in reasoning content replay for chat completion models

### LINKED ISSUE #2669: `reasoning_content` support for models other than DeepSeek (GLM, Kimi, etc.)
## Problem

When using tool-calling agents with Moonshot AI Kimi models, multi-turn conversations fail with a `400` error on the second turn.

These models include a `reasoning_content` field in their responses. When tool calls are also present, the `reasoning_content` must be included in the assistant message sent back to the API on the next turn — otherwise the API rejects the request.

Currently this behaviour is only implemented for DeepSeek models.

**Affected models:**
- Moonshot AI: `kimi-k2`, `kimi-k2.5`, and other Kimi reasoning variants — `reasoning_content` is **required** for multi-turn tool-calling; omitting it causes a `400` error.
- Zhipu AI: `GLM-5`, etc. — `reasoning_content` is not strictly required (no error without it), but including it is beneficial for model performance.

## Question

Is there a recommended way to support `reasoning_content` for these additional providers?

seratch:
This pull request adds an opt-in `should_replay_reasoning_content` hook for `OpenAIChatCompletionsModel` and `LitellmModel` to address the runtime gap discussed in #2669. The hook is evaluated in the shared Chat Completions converter with a structured context exposing `{ model, base_url, reasoning }`, so callers can enable provider-specific replay behavior without changing the default SDK behavior. Resolves #2669

This pull request also keeps the existing one provider only default replay logic intact when no hook is provided, and fixes a converter bug where pending `reasoning_content` could be dropped when a reasoning item was followed by an assistant output message before tool calls in the same turn. The change is covered by new regression tests for the shared converter, the direct OpenAI-compatible path, and the LiteLLM path, using generic reasoning-content model identifiers rather than provider names.

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -507,7 +507,8 @@ class Converter:
                 current_assistant_msg = ChatCompletionAssistantMessageParam(role="assistant")
                 current_assistant_msg["content"] = None
                 current_assistant_msg["tool_calls"] = []
-                apply_pending_reasoning_content(current_assistant_msg)
+
+            apply_pending_reasoning_content(current_assistant_msg)
 
             return current_assistant_msg
 
@@ -763,7 +764,7 @@ class Converter:
                     # This preserves the original behavior
                     pending_thinking_blocks = reconstructed_thinking_blocks
 
-                elif model is not None:
+                if model is not None:
                     replay_context = ReasoningContentReplayContext(
                         model=model,
                         base_url=normalized_base_url,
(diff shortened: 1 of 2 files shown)


src/agents/models/chatcmpl_converter.py
```diff
@@ -553,7 +561,9 @@ def ensure_assistant_message() -> ChatCompletionAssistantMessageParam:
 
             # 3) response output message => assistant
             elif resp_msg := cls.maybe_response_output_message(item):
-                flush_assistant_message()
+                # A reasoning item can be followed by an assistant message and then tool calls
```
seratch, line 581: This is a different bug I found during the feature addition this time.

PR #2676, opened 2026-03-15
fix(retry): cap jittered delay to max_delay

Dhakshin2007:
## Summary

This PR fixes a bug introduced in the `feat: add opt-in model retry policies` commit (3a52673) inside `src/agents/run_internal/model_retry.py`.

Per reviewer feedback from @seratch, the `Retry-After` `OverflowError` handling has been dropped as it is unnecessary in practice. This PR now focuses solely on the jitter cap fix.

---

### Bug: Jitter can produce delay > `max_delay` (silent backoff violation)

**Location:** `_default_retry_delay()`, lines 287–290

**Root cause:**
The original code caps the exponential delay with `min(..., max_delay)` *before* applying jitter:
```python
# BEFORE (buggy)
delay = min(initial_delay * (multiplier ** max(attempt - 1, 0)), max_delay)
if not use_jitter:
    return delay
return max(delay * (0.875 + random.random() * 0.25), 0.0)
```
Because jitter multiplies the already-capped value by up to 1.125, the final result can exceed `max_delay` by ~12.5%.

**Fix:**
```python
# AFTER (fixed)
base = min(initial_delay * (multiplier ** max(attempt - 1, 0)), max_delay)
if not use_jitter:
    return base
return min(max(base * (0.875 + random.random() * 0.25), 0.0), max_delay)
```
Apply `min(..., max_delay)` *after* jitter so the final delay is always bounded by `max_delay`.

### DIFF
--- a/src/agents/run_internal/model_retry.py
+++ b/src/agents/run_internal/model_retry.py
@@ -94,7 +94,7 @@ def _parse_retry_after(headers: httpx.Headers | Mapping[str, str] | None) -> flo
     if retry_after_ms is not None:
         try:
             parsed_ms = float(retry_after_ms) / 1000.0
-        except (ValueError, OverflowError):
+        except ValueError:
             parsed_ms = None
         if parsed_ms is not None and parsed_ms >= 0:
             return parsed_ms
@@ -105,7 +105,7 @@ def _parse_retry_after(headers: httpx.Headers | Mapping[str, str] | None) -> flo
 
     try:
         parsed_seconds = float(retry_after)
-    except (ValueError, OverflowError):
+    except ValueError:
         parsed_seconds = None
     if parsed_seconds is not None:
         return parsed_seconds if parsed_seconds >= 0 else None


Dhakshin2007:
Thanks for the P2 feedback, @chatgpt-codex-connector!

You are correct — applying jitter to the uncapped exponential (`delay * jitter_factor`) and then clamping makes the jitter ineffective once the base grows well past `max_delay` (≥ attempt 5 with defaults), because even the minimum jitter factor (0.875) keeps the result above `max_delay`, so every call returns exactly `max_delay`.

The second commit addresses this by clamping the **base** first, then applying jitter to the clamped value, and finally capping the result:

```python
# Clamp base first so jitter stays effective even at saturation
base = min(initial_delay * (multiplier ** max(attempt - 1, 0)), max_delay)
if not use_jitter:
    return base
return min(max(base * (0.875 + random.random() * 0.25), 0.0), max_delay)
```

With this approach at saturation (`base = max_delay`), jitter produces delays in `[0.875 * max_delay, max_delay]` — preserving the thundering-herd spread Codex correctly identified was lost.

Dhakshin2007:
@seratch Thank you for the review! I've addressed your feedback:

- Dropped the `Retry-After` `OverflowError` handling as requested
- The PR now contains only the jitter cap fix (capping the final jittered delay to `max_delay`)

Updated the PR title and description accordingly.

PR #2690, opened 2026-03-17
feat: propagate RunContextWrapper to session history read/write paths

HuxleyHu98:
### Summary
This PR adds backward-compatible `RunContextWrapper` propagation to the main session history read/write paths.

Specifically, it:

- adds optional `wrapper` support to:
  - `Session.get_items(...)`
  - `Session.add_items(...)`
- propagates the current runner `RunContextWrapper` through:
  - `prepare_input_with_session(...)`
  - `save_result_to_session(...)`
  - `save_resumed_turn_items(...)`
- updates both sync and streamed runner flows to pass the current wrapper
- preserves compatibility for existing custom session implementations by falling back to the old call signatures when a session method does not accept `wrapper`

### Why
Issue #2072 is about making runtime context available to session implementations.

The earlier approach in #2488 added optional parameters at the interface level, but it did not actually propagate context from the runner side. This patch focuses on the runner-side propagation path so session implementations that support it can actually receive `RunContextWrapper` during session history reads/writes.

At the same time, this keeps older custom session implementations working unchanged.

### Scope
This PR intentionally keeps the scope narrow.

Included:
- session history read/write paths:
  - `get_items`
  - `add_items`
- sync and streamed runner flows
- backward-compatible dispatch for legacy session implementations

Not included:
- `pop_item`
- `clear_session`
- broader session lifecycle expansion outside the main history read/write paths

I kept those out on purpose to make the initial change small, testable, and compatibility-focused.

### Compatibility
Backward compatibility is handled in the internal dispatch layer.

If a session implementation supports the optional `wrapper` parameter, the runner passes the current `RunContextWrapper`.

If it does not, the runner falls back to the existing call shape, so older custom session implementations continue to work without modification.

### Tests
Added coverage for:
- new-style custom session implementations that accept `wrapper`
  - verifies `get_items` / `add_items` receive the current `RunContextWrapper`
  - verifies the wrapper exposes the original run context
- legacy custom session implementations with the old method signatures
  - verifies they still work in:
    - `Runner.run`
    - `Runner.run_sync`
    - `Runner.run_streamed`

Ran:
- `tests/test_session.py`
- `tests/test_agent_runner_streamed.py`
- `tests/test_soft_cancel.py`
- `tests/test_hitl_session_scenario.py`
- `tests/extensions/memory/test_encrypt_session.py`
- `tests/memory/test_openai_responses_compaction_session.py`

Result:
- **146 passed**

Static checks:
- `pyright` → **0 errors**
- `ruff check` → **passed**

### DIFF
--- a/examples/memory/file_session.py
+++ b/examples/memory/file_session.py
@@ -47,6 +47,7 @@ class FileSession(Session):
     async def get_items(
         self,
         limit: int | None = None,
+        *,
         wrapper: RunContextWrapper[Any] | None = None,
     ) -> list[Any]:
         session_id = await self._ensure_session_id()
@@ -58,6 +59,7 @@ class FileSession(Session):
     async def add_items(
         self,
         items: list[Any],
+        *,
         wrapper: RunContextWrapper[Any] | None = None,
     ) -> None:
         if not items:
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -126,6 +126,7 @@ class AdvancedSQLiteSession(SQLiteSession):
     async def add_items(
         self,
         items: list[TResponseInputItem],
+        *,
         wrapper: RunContextWrapper[Any] | None = None,
     ) -> None:
         """Add items to the session.
@@ -143,8 +144,8 @@ class AdvancedSQLiteSession(SQLiteSession):
     async def get_items(
         self,
         limit: int | None = None,
-        wrapper: RunContextWrapper[Any] | None = None,
         *,
+        wrapper: RunContextWrapper[Any] | None = None,
         branch_id: str | None = None,
     ) -> list[TResponseInputItem]:
         """Get items from current or specified branch.
--- a/src/agents/extensions/memory/async_sqlite_session.py
+++ b/src/agents/extensions/memory/async_sqlite_session.py
@@ -106,6 +106,7 @@ class AsyncSQLiteSession(SessionABC):
     async def get_items(
         self,
         limit: int | None = None,
+        *,
         wrapper: RunContextWrapper[Any] | None = None,
     ) -> list[TResponseInputItem]:
         """Retrieve the conversation history for this session.
@@ -158,6 +159,7 @@ class AsyncSQLiteSession(SessionABC):
     async def add_items(
         self,
         items: list[TResponseInputItem],
+        *,
         wrapper: RunContextWrapper[Any] | None = None,
     ) -> None:
         """Add new items to the conversation history.
--- a/src/agents/extensions/memory/dapr_session.py
+++ b/src/agents/extensions/memory/dapr_session.py
@@ -237,6 +237,7 @@ class DaprSession(SessionABC):
     async def get_items(
         self,
         limit: int | None = None,
+        *,
         wrapper: RunContextWrapper[Any] | None = None,
     ) -> list[TResponseInputItem]:
         """Retrieve the conversation history for this session.
@@ -280,6 +281,7 @@ class DaprSession(SessionABC):
     async def add_items(
         self,
         items: list[TResponseInputItem],
+        *,
         wrapper: RunContextWrapper[Any] | None = None,
     ) -> None:
         """Add new items to the conversation history.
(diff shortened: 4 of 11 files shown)


HuxleyHu98:
I’ve addressed the review feedback and resolved the outdated Codex conversations. Local checks are passing on my side. Could a maintainer please enable/approve the PR checks when convenient? Thanks!

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -135,6 +140,8 @@ async def add_items(self, items: list[TResponseInputItem]) -> None:
     async def get_items(
         self,
         limit: int | None = None,
+        wrapper: RunContextWrapper[Any] | str | None = None,
```
seratch: This new arg must be placed at the end of the argument list for backward-compatibility

src/agents/extensions/memory/advanced_sqlite_session.py
```diff
@@ -146,6 +153,10 @@ async def get_items(
         Returns:
             List of conversation items from the specified branch.
         """
+        if isinstance(wrapper, str) and branch_id is None:
```
seratch: this wouldn't be necessary if you add the new one as the last optional param

examples/memory/file_session.py
```diff
@@ -43,14 +44,22 @@ async def get_session_id(self) -> str:
         """Return the session id, creating one if needed."""
         return await self._ensure_session_id()
 
-    async def get_items(self, limit: int | None = None) -> list[Any]:
+    async def get_items(
+        self,
+        limit: int | None = None,
+        wrapper: RunContextWrapper[Any] | None = None,
```
seratch, line 51: Let's use kwargs for the new optional param for easier future enhancement
```suggestion
        *,
        wrapper: RunContextWrapper[Any] | None = None,
```

examples/memory/file_session.py
```diff
@@ -43,14 +44,22 @@ async def get_session_id(self) -> str:
...
         items = await self._read_items(session_id)
         if limit is not None and limit >= 0:
             return items[-limit:]
         return items
 
-    async def add_items(self, items: list[Any]) -> None:
+    async def add_items(
+        self,
+        items: list[Any],
+        wrapper: RunContextWrapper[Any] | None = None,
```
seratch, line 63: same (won't repeat this)

seratch (changes requested):

HuxleyHu98:
Local checks are all passing (format, lint, mypy, pyright — 0 errors, 166 tests passed). Could a maintainer please approve running the CI checks? Thanks!

src/agents/extensions/memory/async_sqlite_session.py
```diff
@@ -102,7 +103,11 @@ async def _locked_connection(self) -> AsyncIterator[aiosqlite.Connection]:
             conn = await self._get_connection()
             yield conn
 
-    async def get_items(self, limit: int | None = None) -> list[TResponseInputItem]:
+    async def get_items(
+        self,
+        limit: int | None = None,
+        wrapper: RunContextWrapper[Any] | None = None,
```
seratch, line 110: Can you make this consistent across the built-in implementations?

```suggestion
        *,
        wrapper: RunContextWrapper[Any] | None = None,
```

HuxleyHu98:
Rebased onto the latest \ and reran local checks after the rebase. Current local status is all green on the touched surface:\n\n- ruff format --check ✅\n- ruff check ✅\n- mypy ✅\n- pyright ✅\n- uv run pytest tests/extensions/memory ✅\n\nCould a maintainer please enable/approve the PR CI checks when convenient? Thanks!

HuxleyHu98:
@seratch Marked this PR ready for review again after addressing the previous comments. CI is kicking off now; would appreciate another look when you have a moment.

seratch:
Thanks for working on this. Overall, this should be good. That said, this PR changes lots of files, and those edits could conflict with our current priorities. Thus, let us hold off merging it even if it looks good to me. It may take a few weeks. It'd be appreciated if you could understand this.

HuxleyHu98:
Thanks, totally understood — I appreciate the review and the context. Happy to leave this here for now and wait until the timing is better on your side.

PR #2699, opened 2026-03-17
fix: retry transient streamable-http MCP failures on isolated session

elainegan-openai:
## Summary
- retry a streamable-HTTP MCP tool call on a fresh isolated session when the shared session hits a transient transport failure
- keep shared-session healthy-path parallelism; only isolate on the failure path
- add focused regression tests for cancelled sibling calls, 5xx retry, 4xx no-retry, and preserved outer cancellation

## Why
The likely remaining bug is earlier in the chain than `MCPUtil.invoke_mcp_tool()`: one shared-session streamable-HTTP request failure can still fan out into sibling request cancellations. This patch keeps the normal shared session on the healthy path, but retries just the affected call on an isolated session when the shared session sees a transient transport failure.

Concretely, the isolated retry triggers for:
- `asyncio.CancelledError`
- `httpx.ConnectError`
- `httpx.TimeoutException`
- `httpx.HTTPStatusError` with status >= 500
- `BaseExceptionGroup` containing one of the above

It does not retry 4xx errors.

## Validation
- `uv run pytest -q tests/mcp/test_client_session_retries.py`
- `uv run mypy src/agents/mcp/server.py tests/mcp/test_client_session_retries.py`
- `ruff check src/agents/mcp/server.py tests/mcp/test_client_session_retries.py`

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -5,7 +5,7 @@ import asyncio
 import inspect
 import sys
 from collections.abc import Awaitable
-from contextlib import AbstractAsyncContextManager, AsyncExitStack
+from contextlib import AbstractAsyncContextManager, AsyncExitStack, asynccontextmanager
 from datetime import timedelta
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Literal, TypeVar, Union, cast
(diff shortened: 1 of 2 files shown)


seratch:
Closing this in favor of https://github.com/openai/openai-agents-python/pull/2703

PR #2721, opened 2026-03-19
feat(mcp): expose list_resources, list_resource_templates, and read_resource on MCPServer

adityasingh2400:
## Problem

The MCP specification defines a **Resources** primitive alongside Tools and Prompts. The underlying `ClientSession` from the MCP Python SDK already implements `list_resources()`, `list_resource_templates()`, and `read_resource(uri)`, but none of these are surfaced through `MCPServer` or its concrete implementations.

This means callers cannot retrieve file contents, database records, or any other server-managed resources even though the transport layer fully supports it.

## Solution

Add the three resource methods in three places, following the exact same pattern as the existing `list_prompts` / `get_prompt` pair:

1. **Abstract methods on `MCPServer`** — establishing the contract for all server types
2. **Concrete implementations on `_MCPServerWithClientSession`** — delegating to `session.list_resources()`, `session.list_resource_templates()`, and `session.read_resource(uri)` via `_maybe_serialize_request` (so serialization and retry logic apply automatically)
3. **Stub implementations in test helpers** — so `FakeMCPServer` and the other test-only subclasses remain concrete

## Tests

Six new tests in `tests/mcp/test_mcp_resources.py`:
- `test_list_resources_raises_when_not_connected`
- `test_list_resource_templates_raises_when_not_connected`
- `test_read_resource_raises_when_not_connected`
- `test_list_resources_returns_result`
- `test_list_resource_templates_returns_result`
- `test_read_resource_returns_result`

## Usage

```python
async with MCPServerStreamableHttp(params={"url": "http://localhost:8000/mcp"}) as server:
    # List available resources
    resources = await server.list_resources()
    for resource in resources.resources:
        print(resource.uri, resource.name)

    # Read a specific resource by URI
    result = await server.read_resource("file:///path/to/data.csv")
    for content in result.contents:
        print(content.text)  # TextResourceContents

    # List URI templates the server supports
    templates = await server.list_resource_templates()
```

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -245,7 +245,8 @@ class MCPServer(abc.ABC):
         """Read the contents of a specific resource by URI.
 
         Args:
-            uri: The URI of the resource to read (e.g. ``file:///path/to/file.txt``).
+            uri: The URI of the resource to read. See :class:`~pydantic.networks.AnyUrl`
+                for the supported URI formats.
 
         Returns a :class:`~mcp.types.ReadResourceResult`.  Subclasses that do not
         support resources may leave this unimplemented; it will raise
@@ -794,8 +795,8 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
         """Read the contents of a specific resource by URI.
 
         Args:
-            uri: The URI of the resource to read (e.g. ``file:///path/to/file.txt`` or
-                ``postgres://db/table/row``).
+            uri: The URI of the resource to read. See :class:`~pydantic.networks.AnyUrl`
+                for the supported URI formats.
         """
         if not self.session:
             raise UserError("Server not initialized. Make sure you call `connect()` first.")


adityasingh2400:
Fixed the P1 Codex issue — `list_resources`, `list_resource_templates`, and `read_resource` are now non-abstract with a descriptive `NotImplementedError` default on the base class. Existing subclasses continue to instantiate normally and only raise if they actually call a resource method they haven't implemented. Added a regression test covering this path.

seratch:
I found one issue with the PR.

`src/agents/mcp/server.py` adds `list_resources()` and `list_resource_templates()` as new wrapper APIs, but both methods drop pagination inputs entirely. The base API and the concrete implementation expose no `cursor` or `params` arguments, and the implementation always calls `session.list_resources()` / `session.list_resource_templates()` with no pagination state. That means paginated MCP servers can only return the first page, so the new wrapper cannot actually satisfy the docstring claim that it lists “all resources.” The new tests only cover the no-argument happy path, so this regression is not caught.

adityasingh2400:
Fixed — `list_resources` and `list_resource_templates` now accept an optional `cursor: str | None = None` argument on both the base class and the concrete implementation. The cursor is forwarded directly to the underlying `ClientSession` call, so callers can page through results by passing back the `nextCursor` from each response. Added two new tests verifying cursor is forwarded correctly.

seratch:
@codex review this in depth again

src/agents/mcp/server.py
```diff
@@ -708,6 +772,39 @@ async def get_prompt(
...
+            raise UserError("Server not initialized. Make sure you call `connect()` first.")
+        session = self.session
+        assert session is not None
+        return await self._maybe_serialize_request(lambda: session.list_resource_templates(cursor))
+
+    async def read_resource(self, uri: str) -> ReadResourceResult:
+        """Read the contents of a specific resource by URI.
+
+        Args:
+            uri: The URI of the resource to read (e.g. ``file:///path/to/file.txt`` or
```
seratch: can you simply mention AnyUrl docstring as reference? We'd like to avoid mentioning specific examples and having our own comment, which can be outdated in the future.

src/agents/mcp/server.py
```diff
@@ -192,6 +200,62 @@ async def get_prompt(
...
+        raise NotImplementedError(
+            f"MCP server '{self.name}' does not support list_resource_templates. "
+            "Override this method in your server implementation."
+        )
+
+    async def read_resource(self, uri: str) -> ReadResourceResult:
+        """Read the contents of a specific resource by URI.
+
+        Args:
+            uri: The URI of the resource to read (e.g. ``file:///path/to/file.txt``).
```
seratch: can you simply mention AnyUrl docstring as reference? We'd like to avoid mentioning specific examples and having our own comment, which can be outdated in the future.

adityasingh2400:
Done — replaced both URI example strings with `:class:`\~pydantic.networks.AnyUrl`` references in both the base class and the concrete implementation docstrings.

PR #2731, opened 2026-03-20
fix: #2729 avoid eager-task race in function tool batch executor

### LINKED ISSUE #2729: Regression in 0.12.5: KeyError in _FunctionToolBatchExecutor under   asyncio.eager_task_factory (Textual / Python 3.13)
### Summary

  After upgrading from `openai-agents==0.9.3` to `openai-
  agents==0.12.5`, streamed runs with parallel tool calls started
  failing with:

  ```text
  KeyError: <Task finished name='None' exception=KeyError(...)>
  ```
  Downgrading back to 0.9.3 avoids the problem.

  This appears to be an SDK bug triggered by event loops that use
  asyncio.eager_task_factory (for example, Textual on Python 3.12+).

  ### Environment

  - openai-agents==0.12.5, 0.10.x, 0.11.x
  - Python 3.13.2
  - Textual TUI app
  - parallel tool calls enabled
  - nested agent-as-tool usage with streamed runs

  ### Traceback tail

```
  File ".../agents/run_internal/tool_execution.py", line 1440, in
  _run_single_tool
      self.task_states[current_task].in_post_invoke_phase = False
  KeyError: <Task finished name='None' exception=KeyError(...)>
```

  ### Why this looks like an SDK bug

  In _FunctionToolBatchExecutor, the executor creates the task first
  and registers it in task_states after:

```
  task = asyncio.create_task(self._run_single_tool(...))
  self.task_states[task] = _FunctionToolTaskState(...)
```

  Then _run_single_tool() immediately does a lookup using
  asyncio.current_task():

```
  current_task = asyncio.current_task()
  self.task_states[current_task].in_post_invoke_phase = False
```

  Under asyncio.eager_task_factory, the coroutine can start running
  before self.task_states[task] = ... executes, so that lookup can
  raise KeyError.

  Textual enables eager tasks like this:
```
  loop.set_task_factory(asyncio.eager_task_factory)
```

  ### Minimal proof of the race

  This reproduces the underlying scheduling problem on Python 3.13:

```python
  import asyncio

  async def worker(state):
      t = asyncio.current_task()
      print("registered at start:", t in state)

  async def main():
      loop = asyncio.get_running_loop()
      loop.set_task_factory(asyncio.eager_task_factory)

      state = {}
      task = asyncio.create_task(worker(state))
      state[task] = True
      await task

  asyncio.run(main())
```

  Output:

```
  registered at start: False
```

  So create_task(...) plus "register after creation" is unsafe under
  eager tasks.

  ### Suggested fix

  The robust fix is to avoid relying on
  self.task_states[current_task] from inside the running coroutine.

  A minimal patch would be to create these internal SDK tasks with
  eager start disabled, for example by constructing
  asyncio.Task(..., eager_start=False) instead of
  asyncio.create_task(...) for the function-tool executor's
  bookkeeping tasks.

  A more robust fix would be:

  - create task_state before task creation
  - pass task_state into _run_single_tool()
  - mutate task_state directly instead of looking it up via
    current_task

  That removes the race entirely.

  ### Workarounds

  Current workarounds on the application side are:

  - downgrade to openai-agents==0.9.3
  - disable/reset the eager task factory around agent execution
  - disable parallel tool calls

seratch:
This pull request fixes #2729 a race in function tool execution when the event loop uses `asyncio.eager_task_factory`.

`_FunctionToolBatchExecutor` previously created the outer tool task first and only registered its bookkeeping state in `self.task_states` afterward. At the same time, `_run_single_tool()` immediately called `asyncio.current_task()` and looked up `self.task_states[current_task]` to update fields such as `in_post_invoke_phase`. Under normal task scheduling this usually worked because the coroutine did not start running until after the registration line executed. Under eager scheduling, however, the coroutine can begin during `asyncio.create_task(...)` itself, before the caller has inserted the task into `self.task_states`. In that case, `current_task()` returns the correct task object, but that task is still missing from the dictionary, so the lookup raises `KeyError`.

This change removes that ordering dependency by creating `_FunctionToolTaskState` before scheduling the outer task and threading that state object through the executor's private coroutine chain. The running coroutine now mutates its own `task_state` directly instead of trying to recover it from `self.task_states[current_task]`. The task-to-state mapping is still kept for parent-side bookkeeping such as ordering, result collection, cancellation, and post-invoke draining, but coroutine-local state updates no longer depend on a race-prone reverse lookup.

The pull request also adds a regression test in `tests/test_run_step_execution.py` that enables `asyncio.eager_task_factory` on Python 3.12+ and verifies that `execute_function_tool_calls()` completes successfully and preserves tool output ordering under eager task scheduling.

### DIFF
--- a/docs/ja/streaming.md
+++ b/docs/ja/streaming.md
@@ -4,19 +4,19 @@ search:
 ---
 # ストリーミング
 
-ストリーミングを使用すると、エージェント実行の進行に合わせた更新を購読できます。これは、エンドユーザーに進捗更新や部分的な応答を表示するのに役立ちます。
+ストリーミングを使うと、エージェントの実行が進行する間の更新を購読できます。これは、エンドユーザーに進捗更新や部分的な応答を表示するのに役立ちます。
 
-ストリーミングするには、[`Runner.run_streamed()`][agents.run.Runner.run_streamed] を呼び出します。これにより [`RunResultStreaming`][agents.result.RunResultStreaming] が返されます。`result.stream_events()` を呼び出すと、以下で説明する [`StreamEvent`][agents.stream_events.StreamEvent] オブジェクトの async ストリームが得られます。
+ストリーミングするには、[`Runner.run_streamed()`][agents.run.Runner.run_streamed] を呼び出します。これにより [`RunResultStreaming`][agents.result.RunResultStreaming] が得られます。`result.stream_events()` を呼び出すと、以下で説明する [`StreamEvent`][agents.stream_events.StreamEvent] オブジェクトの非同期ストリームが得られます。
 
-async イテレーターが終了するまで、`result.stream_events()` の消費を続けてください。ストリーミング実行は、イテレーターが終了するまで完了しません。セッション永続化、承認記録、履歴圧縮などの後処理は、最後の可視トークン到着後に完了する場合があります。ループ終了時に、`result.is_complete` は最終的な実行状態を反映します。
+非同期イテレーターが終了するまで `result.stream_events()` の消費を続けてください。ストリーミング実行は、イテレーターが終了するまで完了しません。また、セッション永続化、承認の記録管理、履歴の圧縮といった後処理は、最後の可視トークン到着後に完了する場合があります。ループを抜けた時点で、`result.is_complete` が最終的な実行状態を反映します。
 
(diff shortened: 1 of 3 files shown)


seratch:
@codex review this PR

Issue #1912, opened 2025-10-16
Python Realtime Voice Agent doesn’t handle output guardrails like Node.js SDK


bindumadhavimaddireddy:
### Describe the feature
I’m using the `openai-agents-python` SDK with a `RealtimeRunner` for a voice-to-voice Realtime Agent (audio input + audio output) with output guardrails.

When a guardrail is tripped, the SDK emits a `guardrail_tripped` event, but it doesn’t automatically handle it like the Node.js SDK ( [openai-agents-js](https://github.com/openai/openai-agents-js/tree/main) ) does.
In the `Node.js version`, output guardrails are handled internally before the model speaks. so it never says the blocked content (e.g., mentioning a competitor if that’s disallowed).

In the Python SDK, however, the model still speaks the blocked response, and I have to manually cancel it and try to send another safe message (like “Sorry, I can’t help with that request”). But this manual handling doesn’t work correctly, the model doesn’t speak the safe message.

Manual Implementation We are doing when guardrail triggered. We also tried different ways b y sending events but didn't work out. 

```
RealtimeAgent(
    name=agent_data.name,
    instructions=prompt,
    tools=agents_sdk_tools,
    output_guardrails=[
        OutputGuardrail(
            guardrail_function=moderation_output_guardrail(agent_data.guardrails_instructions)
        )
    ],
)
```

```
# Guardrail handling
if event.type == "guardrail_tripped":
    feedback_message = "Sorry, I can’t help with that request."
    await self.session.interrupt()
    await asyncio.sleep(0.1)
    await self.session.send_message(feedback_message)
```

**References**

Node.js implementation:

[realtimeSession.ts (L555–L596)](https://github.com/openai/openai-agents-js/blob/main/packages/agents-realtime/src/realtimeSession.ts#L555-L596)

[guardrail.ts (L30, L73–L82)](https://github.com/openai/openai-agents-js/blob/main/packages/agents-realtime/src/guardrail.ts#L30)

Python implementation:

[guardrail.py (around L140)](https://github.com/openai/openai-agents-python/blob/main/src/agents/guardrail.py#L140)

**Question**

Will the Python Realtime SDK support automatic output guardrail handling like the Node.js SDK, i.e., canceling blocked content before it’s spoken and generating a safe fallback message automatically for realtime agents?

seratch:
Thanks for reporting this issue with details. This should be definitely improved.

seratch:
[...]

PR #2732, opened 2026-03-20
fix(realtime): wait for cancelled responses before fallback

felmonon:
## Summary
- defer `response.create` until the server emits `response.done` after a forced cancellation
- apply the same ordering to both user-input and tool-output follow-up paths
- add regressions for the active-response race behind realtime guardrail fallback

This keeps the scope narrow to the `conversation_already_has_active_response` failure mode reported in #1912 rather than trying to claim full Node.js parity.

## Testing
- `uv run pytest tests/realtime/test_openai_realtime.py tests/realtime/test_session.py`
- `uv run pytest tests/realtime`
- `uv run ruff check src/agents/realtime/openai_realtime.py tests/realtime/test_openai_realtime.py`
- `uv run mypy src/agents/realtime/openai_realtime.py tests/realtime/test_openai_realtime.py`
- `uv run pyright src/agents/realtime/openai_realtime.py tests/realtime/test_openai_realtime.py`

Related to #1912.

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -273,11 +273,16 @@ class OpenAIRealtimeWebSocketModel(RealtimeModel):
     def __init__(self, *, transport_config: TransportConfig | None = None) -> None:
         self.model = "gpt-realtime"  # Default model
         self._websocket: ClientConnection | None = None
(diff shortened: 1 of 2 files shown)


felmonon:
Closing in favor of a smaller, focused PR that addresses only the core race condition.

PR #2733, opened 2026-03-20
fix: prevent KeyError in _FunctionToolBatchExecutor under eager_task_factory

guoyangzhen:
## Problem

After upgrading from `openai-agents==0.9.3` to `0.12.5`, streamed runs with parallel tool calls fail with:

```
KeyError: <Task finished name='None' exception=KeyError(...)>
```

This only reproduces under `asyncio.eager_task_factory` (used by Textual TUI framework on Python 3.12+).

## Root Cause

In `_FunctionToolBatchExecutor._create_tool_task()`:

```python
task = asyncio.create_task(self._run_single_tool(...))  # ← task starts running here
self.task_states[task] = task_state                      # ← registered AFTER
```

Under `eager_task_factory`, `create_task()` starts executing the coroutine immediately before returning. The task can complete and be processed by `_partition_pending_tasks` → `self.task_states[task]` before `self.task_states[task] = task_state` executes, causing `KeyError`.

The issue reporter confirmed this with a minimal reproduction showing the scheduling problem on Python 3.13.

## Fix

Wrap `_run_single_tool` in an inner `async def` so that `eager_task_factory` runs the wrapper's synchronous preamble (which is empty) and yields at the first `await`, returning control to `_create_tool_task` to finish `task_states` registration.

This is preferred over "register before create_task" because the task object (needed as the dict key) is only available from `create_task()`'s return value.

## Testing

The fix ensures that when `eager_task_factory` eagerly executes the wrapper:
1. The wrapper's synchronous body runs (nothing happens — it's empty)
2. The wrapper hits `await self._run_single_tool(...)` and suspends
3. Control returns to `_create_tool_task`
4. `self.task_states[task] = task_state` executes — state is now registered
5. When the event loop resumes the wrapper, `_run_single_tool` runs safely

### DIFF
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -1335,13 +1335,24 @@ class _FunctionToolBatchExecutor:
 
     def _create_tool_task(self, tool_run: ToolRunFunction, order: int) -> None:
         task_state = _FunctionToolTaskState(tool_run=tool_run, order=order)
-        task = asyncio.create_task(
-            self._run_single_tool(
+
+        # Under asyncio.eager_task_factory, the coroutine starts executing inside
+        # create_task() before it returns. This creates a race: the task could
+        # complete and be processed by _partition_pending_tasks before
+        # self.task_states[task] is set on the next line, causing KeyError.
+        #
+        # Fix: wrap in an inner async function so eager execution runs the
+        # wrapper's synchronous preamble (nothing) and yields at the first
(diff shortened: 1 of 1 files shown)


seratch:
@guoyangzhen thanks for sharing this. I believe my PR https://github.com/openai/openai-agents-python/pull/2731 should already resolve the issue, but do you still see some patterns that are not covered by my change? If so, can you add unit tests covering the patterns as well?

PR #2734, opened 2026-03-20
feat: support self/cls in function_tool for class methods

### LINKED ISSUE #94: Add `method_tool` Functionality

mvanhorn:
### Summary

When a class method decorated with `@function_tool` is used as a tool, the `self` (or `cls`) parameter leaks into the generated JSON schema. The OpenAI API rejects this with a 400 error because `self` has no `type` key. This PR skips the first parameter when it has no type annotation and is named `self` or `cls`, and also handles the case where `RunContextWrapper`/`ToolContext` follows `self`/`cls` as the second parameter.

### Test plan

- Added `test_method_self_param_skipped` - verifies `self` is excluded from schema for bound methods
- Added `test_classmethod_cls_param_skipped` - verifies `cls` is excluded for classmethod-like functions
- Added `test_method_self_with_context_second_param` - verifies `self` skip + `RunContextWrapper` as second param works
- Added `test_regular_unannotated_first_param_still_included` - verifies non-self/cls unannotated params still work
- All existing tests pass (`make tests` - 2405 passed)

### Issue number

Closes #94

### Checks

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -35,6 +35,8 @@ class FuncSchema:
     """The signature of the function."""
     takes_context: bool = False
     """Whether the function takes a RunContextWrapper argument (must be the first argument)."""
+    self_or_cls_skipped: bool = False
+    """Whether an unannotated self/cls first parameter was skipped during schema generation."""
     strict_json_schema: bool = True
     """Whether the JSON schema is in strict mode. We **strongly** recommend setting this to True,
     as it increases the likelihood of correct JSON input."""
@@ -285,6 +287,7 @@ def function_schema(
     sig = inspect.signature(func)
     params = list(sig.parameters.items())
     takes_context = False
+    self_or_cls_skipped = False
     filtered_params = []
 
     if params:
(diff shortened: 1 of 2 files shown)


mvanhorn:
Addressed Codex finding in 73b9a9f: context params at non-adjacent positions after self/cls now correctly raise `UserError`. Added edge case test.

@codex review

mvanhorn:
@codex review

mvanhorn:
@codex review

mvanhorn:
@codex review

mvanhorn:
Addressed both Codex findings in 98ba075e:

**P1 (bind receiver):** Unbound methods now raise `UserError` instead of silently creating a broken schema. Bound methods (`instance.method`) already work - Python strips `self` before `inspect.signature` sees it, so this code path only triggered for unbound calls like `MyClass.method` where execution would fail.

**P2 (drop only leading receiver):** Removed entirely - the filter is now dead code since we reject instead of skip.

Net effect: -26 lines. Simpler code, no silent runtime failures.

Issue #569, opened 2025-04-22
Reasoning model items provide to General model


lyk0014:
Agent A use O4-mini,
Agent B use gpt-4.1

When provide A_result.to_input_list() as B.input, meet this error:

openai.BadRequestError: Error code: 400 - {'error': {'message': 'Reasoning input items can only be provided to a reasoning or computer use model. Remove reasoning items from your input and try again.', 'type': 'invalid_request_error', 'param': 'input', 'code': None}}

DanieleMorotti:
Hi, @rm-openai , I would like to propose two possible PRs: 

1. Add a `convert_reasoning_items` parameter to the function `RunResultBase.to_input_list()`. When enabled, this option would convert reasoning items into standard assistant messages, for example:
```python
{
    "id": "msg_id",
        content": [
            {
                "annotations": [],
                "text": msg["summary"],
                "type": "output_text"
            }
        ],
        "role": "assistant",
        "status": "completed",
        "type": "message"
}
```

2. Implement the reasoning-message filter into a utility function (e.g. in `utils`) so it can be used across the package.

Let me know if you like these approaches, or if you would suggest a different solution, such that I can go ahead and open a PR, thanks

somecoulombs:
@DanieleMorotti  - Is there a way this can be implemented with input filters?

DanieleMorotti:
@somecoulombs Yes, if we want to restrict this kind of filtering only to handoffs, I can put the function in `agents.extensions.handoff_filters` submodule

rm-openai:
@DanieleMorotti IMO we need a few things:
1. A utility function that can strip reasoning items. 
2. A handoff filter that uses #1 
3. A lifecycle method `def on_prepare_input(context, agent, input) -> input` that lets you mutate the input before the LLM is called for an agent.

If you're interested, would love to see a PR for any/all of those!

DanieleMorotti:
Yes, I'm happy to collaborate. I've already implemented some code for the first two points. I just need a few clarifications:
1. Which file should I place the utility function in?
2. Regarding the hook, would I call it within `_run_single_turn` function, before generating the new response?

Thank you!

rm-openai:
1. Somewhere in extensions
2. In `_get_new_response` (before calling the model) and `_run_single_turn_streamed` (before calling `model.stream_response`)

DanieleMorotti:
I'm struggling with some problems removing the reasoning items:
```
[...]

PR #2749, opened 2026-03-21
fix: add reasoning-stripping handoff helpers

sztoplover-bit:
## Summary

This PR adds built-in helpers for handing off from reasoning models to non-reasoning models.

Changes:
- add `strip_reasoning_items(...)` to remove `reasoning` items from plain input history
- add `remove_reasoning_items(...)` as a reusable handoff filter
- strip orphaned assistant message IDs when reasoning items are removed from history, to avoid follow-up Responses API validation errors
- add focused regression tests for the new helpers

## Why

Issue #569 reports failures when reasoning-model output history is passed to a non-reasoning model.

This PR takes the narrow, low-risk path discussed in the issue by adding built-in filtering helpers rather than changing default history conversion behavior across the SDK.

## Testing

- `ruff check src/agents/extensions/handoff_filters.py tests/test_extension_filters.py`
- `PYTHONPATH=src /Users/lvxianping/project/llm_study/bug_ass/openai-agents-python/.venv/bin/python -m pytest tests/test_extension_filters.py -q`

## Related

- Related to #569

### DIFF
--- a/src/agents/extensions/handoff_filters.py
+++ b/src/agents/extensions/handoff_filters.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from collections.abc import Sequence
+
 from ..handoffs import (
     HandoffInputData,
     default_handoff_history_mapper,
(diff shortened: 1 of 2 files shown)


seratch:
The issue is resolved by https://github.com/openai/openai-agents-python/pull/2700

PR #2754, opened 2026-03-22
fix(realtime): wait for turn end after guardrail interrupt

felmonon:
## Summary
- wait for the current turn to end after a forced guardrail interrupt before sending the guardrail follow-up input
- reuse the existing turn_started/turn_ended model events instead of introducing a new response lifecycle state machine
- add regression tests covering both the active-turn and no-active-turn paths

## Validation
- uv run ruff check src/agents/realtime/session.py tests/realtime/test_session.py
- uv run pytest tests/realtime -q

Related to #1907

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -70,6 +70,10 @@ from .model_inputs import (
 REJECTION_MESSAGE = DEFAULT_APPROVAL_REJECTION_MESSAGE
 
 
+class _TurnCompletionSuperseded(Exception):
+    """Raised when a new turn starts before the waiting turn finishes."""
+
+
 def _serialize_tool_output(output: Any) -> str:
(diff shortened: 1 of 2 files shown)


felmonon:
Addressed the Codex feedback on the latest commit:
- removed the fixed timeout path, so the guardrail follow-up no longer resumes after an arbitrary delay
- wait on the matching interrupted response_id via raw response.done, instead of a session-wide turn event
- added regression coverage for matching completion, unrelated turn completion, already-completed responses, and concurrent guardrail tasks

Validation:
- uv run ruff check src/agents/realtime/session.py tests/realtime/test_session.py
- uv run pytest tests/realtime -q

felmonon:
Closing this because the session-layer approach remained too race-prone under review, and I’m stepping away from this issue rather than continuing to expand the fix surface.

PR #2762, opened 2026-03-24
fix: handle cancelled single function tools as tool failures

elainegan-openai:
## Summary
- handle `CancelledError` from a function tool via the normal tool-failure path, even when it is the only function tool in the turn
- preserve nested `Agent.as_tool()` fallback output when a recovered nested run ends without a final message
- add regression coverage for execution, non-streamed runner, streamed runner, and nested agent-tool paths

## Why
Single-tool turns were handling `CancelledError` differently from turns with multiple function tools. That asymmetry could cancel the entire run instead of surfacing a normal tool error back to the model.

The same path also exposed a nested agent-tool regression: when a recovered nested run ended without a final message, `Agent.as_tool()` returned an empty string instead of the nested tool failure output.

## Validation
- `ruff check src/agents/agent.py src/agents/run_internal/tool_execution.py tests/test_example_workflows.py tests/test_run_step_execution.py tests/test_agent_runner.py tests/test_agent_runner_streamed.py`
- `PYTHONPATH=src pytest -q tests/test_example_workflows.py`
- `PYTHONPATH=src pytest -q tests/test_run_step_execution.py -k 'single_tool_call_uses_default_failure_error_function_for_cancelled_tool or multiple_tool_calls_use_default_failure_error_function_for_manual_cancelled_tool or test_execute_function_tool_calls_parent_cancellation_skips_post_invoke_work'`
- `PYTHONPATH=src pytest -q tests/test_agent_runner.py -k 'single_tool_call_with_cancelled_tool_reaches_final_output or parallel_tool_call_with_cancelled_sibling_reaches_final_output'`
- `PYTHONPATH=src pytest -q tests/test_agent_runner_streamed.py -k 'streamed_single_tool_call_with_cancelled_tool_reaches_final_output or streamed_parallel_tool_call_with_cancelled_sibling_reaches_final_output'`

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -857,7 +857,6 @@ class Agent(AgentBase, Generic[TContext]):
 
             from .items import ItemHelpers, MessageOutputItem, ToolCallOutputItem
 
-            latest_tool_output: str | None = None
             for item in reversed(run_result.new_items):
                 if isinstance(item, MessageOutputItem):
                     text_output = ItemHelpers.text_message_output(item)
@@ -865,15 +864,11 @@ class Agent(AgentBase, Generic[TContext]):
                         return text_output
 
                 if (
-                    latest_tool_output is None
-                    and isinstance(item, ToolCallOutputItem)
+                    isinstance(item, ToolCallOutputItem)
                     and isinstance(item.output, str)
                     and item.output
                 ):
-                    latest_tool_output = item.output
-
-            if latest_tool_output is not None:
-                return latest_tool_output
+                    return item.output
 
             return run_result.final_output
 
--- a/tests/test_agent_as_tool.py
+++ b/tests/test_agent_as_tool.py
@@ -502,6 +502,90 @@ async def test_agent_as_tool_fallback_uses_current_run_items_only(
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


src/agents/agent.py
```diff
@@ -850,6 +850,24 @@ async def dispatch_stream_events() -> None:
             if custom_output_extractor:
                 return await custom_output_extractor(run_result)
 
+            if run_result.final_output is not None and (
```
seratch, line 853: I think the main direction in `tool_execution.py` is correct. The part I’m less convinced by is the fallback added in `Agent.as_tool()`. That logic now scans broader run output to recover a return value, which fixes the nested-cancellation case, but it also widens the default output policy for agent tools. In particular, it can return concatenated text from `new_items`, and it may eventually pick up an older `function_call_output` from the full input history rather than something produced by the current run. A cleaner approach would be to keep the `CancelledError` fix in `tool_execution.py`, but make `Agent.as_tool()` resolve its fallback only from items generated in the current run. For example: prefer non-empty `final_output`, then the latest current-run message output, then the latest current-run `ToolCallOutputItem`. That would solve the bug while keeping the return-value policy narrower and easier to reason about.

elainegan-openai:
@Codex review

PR #2764, opened 2026-03-24
Feature/support mcp function meta extentions

jaw9c:


felmonon:
I traced this locally and the smaller fix looks like it belongs in `MCPUtil.to_function_tool()`, not on the public `FunctionTool` dataclass.

`invoke_function_tool()` only ever calls `function_tool.on_invoke_tool(context, arguments)`, so adding `meta` / `_meta` fields to `FunctionTool` still doesn't carry static MCP metadata into execution by itself.

The narrow path that worked locally was to bind the static MCP tool metadata at conversion time, e.g. by partially applying `invoke_mcp_tool` with the tool's `_meta` / `meta` payload when building the `FunctionTool`. That keeps the existing `invoke_mcp_tool()` merge behavior intact, so static tool metadata and `tool_meta_resolver` output still combine correctly before `server.call_tool(...)`.

That also avoids changing `FunctionTool`'s positional constructor layout, which looks important given the source-compat constructor coverage already in the repo.

I validated that direction locally with focused regressions for:
- preserving static MCP `_meta` through `to_function_tool()` invocation
- merging static MCP metadata with resolver-produced metadata
- keeping the constructor compatibility slice green

seratch:
@felmonon You're right. I came up with an alternative at https://github.com/openai/openai-agents-python/pull/2769

seratch:
Closing this in favor off #2769 Thanks again for sharing this patch!

PR #2765, opened 2026-03-24
fix: optionize initialized notification tolerance

elainegan-openai:
## Summary
- add an explicit MCP streamable-HTTP option for tolerating failed initialized notifications
- keep the default strict and only ignore the failure when callers opt in
- implement the option by wrapping the configured `httpx_client_factory`, rather than forking MCP's stream client orchestration
- add focused transport and wiring coverage for the new option

## Why
Best-effort `notifications/initialized` POST could fail and poison the transport, even though subsequent real requests could still succeed. This change keeps the behavior opt-in at the SDK layer so apps can choose the more tolerant transport policy explicitly.

## Validation
- `ruff check src/agents/mcp/server.py tests/mcp/test_streamable_http_client_factory.py`
- `PYTHONPATH=src pytest -q tests/mcp/test_streamable_http_client_factory.py`

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -133,9 +133,9 @@ async def _streamablehttp_client_with_transport(
         auth=auth,
     )
     transport = transport_factory(url)
-    read_stream_writer, read_stream = anyio.create_memory_object_stream[
-        SessionMessage | Exception
-    ](0)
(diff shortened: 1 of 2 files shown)


elainegan-openai:
@Codex review

PR #2777, opened 2026-03-25
fix: preserve title and description in ToolCallItem.to_input_item() 

### LINKED ISSUE #2776: ToolCallItem.to_input_item() drops title and description on session persistence
I have read the docs and this is a follow-up to #2621 (title) and #2381 (description).

### Describe the bug

`ToolCallItem.title` and `ToolCallItem.description` are lost during session persistence. `description` was added in #2381, `title` in #2621, and `RunState` serialization/deserialization handles both correctly.

However, the session persistence path (`save_result_to_session` → `run_item_to_input_item` → `to_input_item()`) was never updated. The base `RunItemBase.to_input_item()` calls `model_dump(exclude_unset=True)` on the raw item, which does not include these SDK-local fields. They are present in-memory but dropped when persisted via session storage.

### Debug information

- Agents SDK version: latest `main` (post #2621 merge)
- Python version: Python 3.11

### Repro steps

```python
from agents import Agent
from agents.items import ToolCallItem
from openai.types.responses.response_function_tool_call import ResponseFunctionToolCall

agent = Agent(name="test", instructions="test")
raw_item = ResponseFunctionToolCall(
    id="fc_1",
    call_id="call_1",
    name="my_tool",
    arguments="{}",
    type="function_call",
    status="completed",
)
item = ToolCallItem(
    agent=agent,
    raw_item=raw_item,
    title="My Tool",
    description="A helpful tool",
)

result = item.to_input_item()
print("title" in result)        # False — bug
print("description" in result)  # False — bug
```

### Expected behavior

`to_input_item()` should include `title` and `description` in the returned dict so they survive the session persistence round-trip.

k1eran:
Fixes #2776

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -358,6 +358,16 @@ class ToolCallItem(RunItemBase[Any]):
     title: str | None = None
     """Optional short display label if known at item creation time."""
 
+    def to_input_item(self) -> TResponseInputItem:
+        """Converts this item into an input item, including title and description metadata."""
+        result = super().to_input_item()
+        if isinstance(result, dict):
+            if self.title is not None:
+                result["title"] = self.title
+            if self.description is not None:
+                result["description"] = self.description
+        return result
+
 
 ToolCallOutputTypes: TypeAlias = Union[
     FunctionCallOutput,
--- a/tests/test_items_helpers.py
+++ b/tests/test_items_helpers.py
@@ -45,7 +45,7 @@ from agents import (
     TResponseInputItem,
     Usage,
 )
-from agents.items import ToolCallOutputItem
+from agents.items import ToolCallItem, ToolCallOutputItem
 
 
 def make_message(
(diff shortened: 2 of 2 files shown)


felmonon:
I think the current typecheck and test failures may be a sign that this fix is landing a bit too high in the stack. ToolCallItem.to_input_item() looks like it is used for normal replay/model input too, not just session persistence, so adding title / description here seems to change regular tool-call payloads as well. That would also line up with the new description: "" diffs and the typed input-item errors. Maybe the safer fix is to preserve this metadata in the session path while keeping the shared input shape unchanged.

seratch:
Yeah, you're right. I am exploring an alternative on my end.

k1eran:
Closing in light of #2781 ,  seratch's approach of stripping SDK metadata at the persistence/API boundary. Thanks @felmonon for the feedback on the layering issue.

PR #2786, opened 2026-03-26
feat: add external_web_access to WebSearchTool

### LINKED ISSUE #2785: Add external_web_access to WebSearchTool for Responses API web_search
### Please read this first

- **Have you read the docs?** YES. I read the Agents SDK docs and the OpenAI web search docs, including the live internet access section.
- **Have you searched for related issues?** YES. I found related work such as PR #2143 and issue/PR #1592/#1601, but I could not find an open issue specifically tracking `external_web_access` support in `WebSearchTool`.

### Describe the feature
Please add `external_web_access: bool | None = None` to `WebSearchTool` and pass it through to the Responses API `web_search` tool payload.

Background:
- In the OpenAI Responses API, `web_search` supports `external_web_access: false` to disable live internet access and use cached/indexed results only.
- In the current Agents SDK, `WebSearchTool` exposes `user_location`, `filters`, and `search_context_size`, but not `external_web_access`.
- As a result, SDK users cannot explicitly choose cache-only / offline-like behavior when using the built-in web search tool.

Example API:

```python
from agents import Agent, WebSearchTool

agent = Agent(
    name="Research agent",
    model="gpt-5",
    tools=[
        WebSearchTool(
            search_context_size="medium",
            external_web_access=False,
        )
    ],
)
```

Expected request payload:

```python
{
    "type": "web_search",
    "search_context_size": "medium",
    "external_web_access": False,
}
```

This would be useful for:
- Workflows that prefer reproducibility and want cached/indexed results only.
- Policy or compliance requirements that do not allow live web access.
- Keeping the Agents SDK aligned with the existing Responses API capability.

Related:
- PR #2143: `feat: Add external_web_access option to WebSearchTool` (closed / unmerged)
- Issue #1592 / PR #1601: other web search updates were added, but `external_web_access` support is still missing

ymuichiro:
### Summary

Add `external_web_access` to `WebSearchTool` and forward it to the Responses API `web_search` payload only when explicitly set.
This keeps the field optional and avoids sending `null` or an implicit default value when the caller does not specify it.

### Test plan

- `uv run pytest -q tests/test_openai_responses_converter.py`
- Verified the new converter coverage for both cases:
  - omitted `external_web_access` is not sent
  - explicit `external_web_access=False` is sent
- I also ran the repository verification script earlier; repo-wide `make tests` / `make typecheck` currently hit unrelated existing failures outside this change area.

### Issue number

Closes #2785

### Checks

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -1935,15 +1935,16 @@ class Converter:
         if isinstance(tool, FunctionTool):
             return cls._convert_function_tool(tool)
         elif isinstance(tool, WebSearchTool):
+            web_search_tool: dict[str, Any] = {
+                "type": "web_search",
+                "filters": tool.filters.model_dump() if tool.filters is not None else None,
+                "user_location": tool.user_location,
+                "search_context_size": tool.search_context_size,
+            }
+            if tool.external_web_access is not None:
+                web_search_tool["external_web_access"] = tool.external_web_access
             return (
-                _require_responses_tool_param(
-                    {
-                        "type": "web_search",
-                        "filters": tool.filters.model_dump() if tool.filters is not None else None,
-                        "user_location": tool.user_location,
-                        "search_context_size": tool.search_context_size,
-                    }
-                ),
+                _require_responses_tool_param(web_search_tool),
                 None,
             )
         elif isinstance(tool, FileSearchTool):
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -499,6 +499,13 @@ class WebSearchTool:
     search_context_size: Literal["low", "medium", "high"] = "medium"
     """The amount of context to use for the search."""
 
+    external_web_access: bool | None = None
+    """Whether the web search tool may fetch live internet content.
+
+    When omitted, the API default is used. Set to `False` to request cached or
+    indexed-only behavior where supported.
+    """
+
     @property
     def name(self):
         return "web_search"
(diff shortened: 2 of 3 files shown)


seratch (approved):

Sourav-Nandy-ai:
Hi ymuichiro,

I actually ran a live before-and-after test on #2786 instead of just describing the shape of the change. I installed real openai-agents==0.13.1 (the last release before this merged) and 0.13.2 (the first release containing it, confirmed via the GitHub compare diff).

The Results
- Construction: WebSearchTool(external_web_access=False) raises a genuine TypeError on 0.13.1 and constructs cleanly on 0.13.2. It is not just stored inertly; it is threaded straight into the Responses API tool payload.
- Execution: I built a real Agent(model="gpt-4o-mini", tools=[WebSearchTool(external_web_access=False)]) and ran it twice against the live OpenAI API with a current-events query. Both runs completed cleanly. Each fired exactly one real web-search tool call and came back with a grounded, cited, and dated answer. This confirms the toggle does not break the tool-call path it is attached to.

One honest limitation is that from the public result object alone, I can confirm the flag is sent and the call succeeds. However, I cannot independently confirm the backend actually restricts to cached and indexed content when set to False. That would require a dedicated content-recency A/B test, which I kept out of scope to minimize real paid API calls.

I maintain agent-eval (agent_regress), and I am happy to run that A/B test for real if it would provide a useful signal for you.

Best, 
Sourav

PR #2788, opened 2026-03-26
feat(mcp): #1167 prefix colliding MCP tool names

### LINKED ISSUE #1167: Duplicate tool names in multiple MCP tools cause the agent list to hang

Problem Description:

Steps to Reproduce:

Expected Behavior:

Suggested Fix:

Troubleshooting Process:

Note for Developers:  

IgnazioDS:
## Summary
- add an opt-in `mcp_config["include_server_in_tool_names"]` flag to expose MCP tools as `<server_name>_<tool_name>`
- preserve the existing default behavior of raising on duplicate tool names across MCP servers
- keep invocation and meta resolution keyed to the original MCP tool name while logging the prefixed display name
- add regression coverage for collision handling, server-name sanitization, canonical meta resolution, and prefixed-name failure logging

Fixes #1167.

## Validation
- `uv run ruff check src/agents/agent.py src/agents/mcp/util.py tests/mcp/test_mcp_util.py`
- `uv run pytest tests/mcp/test_mcp_util.py -q`
- `uv run mypy src/agents/agent.py src/agents/mcp/util.py tests/mcp/test_mcp_util.py`

## Notes
- This refreshes the previously proposed server-prefix approach from closed PR #2442 against the current `main` branch and keeps the surface area limited to MCP config/util behavior plus focused tests.

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -277,7 +277,8 @@ class MCPUtil:
     @staticmethod
     def _server_tool_name_prefix(server_name: str) -> str:
         normalized = "".join(
-            char if char.isalnum() or char in ("_", "-") else "_" for char in server_name
+            char if (char.isascii() and char.isalnum()) or char in ("_", "-") else "_"
+            for char in server_name
         )
         normalized = normalized.strip("_-")
         if not normalized:
@@ -286,7 +287,14 @@ class MCPUtil:
 
     @staticmethod
     def _prefixed_tool_name(tool_name_prefix: str, tool_name: str) -> str:
-        return f"mcp_{len(tool_name_prefix)}_{tool_name_prefix}{tool_name}"
+        full_name = f"mcp_{len(tool_name_prefix)}_{tool_name_prefix}{tool_name}"
+        if len(full_name) <= 64:
+            return full_name
+        # Truncate to 64 chars using a deterministic hash suffix to avoid collisions
+        hash_suffix = hashlib.sha1(full_name.encode("utf-8")).hexdigest()[:8]
+        # Reserve 9 chars for "_" + 8-char hash
+        truncated = full_name[: 64 - 9]
+        return f"{truncated}_{hash_suffix}"
 
     @classmethod
     def _server_tool_name_prefixes(cls, servers: list[MCPServer]) -> dict[int, str]:


IgnazioDS:
Addressed the review thread in `b9f25db`.

Change made:
- preserve the readable normalized server prefix when it is unique
- add a stable hash-based suffix only when multiple server names normalize to the same prefix
- add a regression test covering `GitHub MCP Server` vs `GitHub_MCP_Server`

Validation:
- `uv run ruff check src/agents/mcp/util.py tests/mcp/test_mcp_util.py`
- `PYTHONPATH=src uv run pytest tests/mcp/test_mcp_util.py -q`
- `uv run mypy src/agents/mcp/util.py --disable-error-code import-not-found --disable-error-code unused-ignore`

IgnazioDS:
Addressed the follow-up naming-collision thread in `d710186`.

Change made:
- switched prefixed MCP tool names to an unambiguous encoded form: `mcp_<prefix_len>_<server_prefix><tool_name>`
- kept the readable server-specific prefix visible in the final tool name
- retained the hash-based server-prefix disambiguation for server names that normalize to the same slug
- updated regression coverage for direct prefix collisions and tuple-level collisions

Validation:
- `uv run ruff check src/agents/agent.py src/agents/mcp/util.py tests/mcp/test_mcp_util.py`
- `PYTHONPATH=src uv run pytest tests/mcp/test_mcp_util.py -q`
- `uv run mypy src/agents/agent.py src/agents/mcp/util.py --disable-error-code import-not-found --disable-error-code unused-ignore`

IgnazioDS:
Thanks for the detailed review @seratch! Both issues addressed:

**1. ASCII-only normalization** — added `char.isascii()` guard in `_server_tool_name_prefix` so non-ASCII Unicode alphanumerics (CJK, Arabic, etc.) are now replaced with `_` before being used as tool name prefixes:

```python
char if (char.isascii() and char.isalnum()) or char in ("_", "-") else "_"
```

**2. 64-character limit** — `_prefixed_tool_name` now caps the result at 64 chars. When truncation is needed, a deterministic 8-char SHA-1 suffix is appended to keep names collision-resistant:

```python
full_name = f"mcp_{len(tool_name_prefix)}_{tool_name_prefix}{tool_name}"
if len(full_name) <= 64:
    return full_name
hash_suffix = hashlib.sha1(full_name.encode("utf-8")).hexdigest()[:8]
truncated = full_name[: 64 - 9]
return f"{truncated}_{hash_suffix}"
```

IgnazioDS:
Hi @seratch! I've addressed both points from your review in commit fa92be2:

1. **ASCII normalization** — `_server_tool_name_prefix` now uses `char.isascii() and char.isalnum()` instead of just `char.isalnum()`, so non-ASCII Unicode characters (CJK, Arabic, etc.) are replaced with `_` and never reach the OpenAI function name.

2. **64-character limit** — `_prefixed_tool_name` now checks `len(full_name) <= 64` and, if over the limit, truncates to 55 chars and appends `_{sha1[:8]}` for collision resistance.

Would appreciate a re-review when you get a chance. Thanks!

IgnazioDS:
Hi @seratch — thanks for the detailed review. Both concerns from your last review have been addressed:

**1. Non-ASCII server prefix normalization** (commit fa92be2)
`_server_tool_name_prefix()` now replaces every character that is not ASCII alphanumeric, `_`, or `-` with `_`, then strips leading/trailing separators:
```python
normalized = "".join(
    char if (char.isascii() and char.isalnum()) or char in ("_", "-") else "_"
    for char in server_name
)
```
A server label like `天気サーバー` becomes `____` → stripped to `server` (the fallback for all-separator results).

**2. 64-character limit with deterministic hash shortening** (commit fa92be2)
`_safe_prefixed_tool_name()` enforces the limit with a collision-safe SHA1 suffix:
```python
if len(full_name) <= 64:
    return full_name
hash_suffix = hashlib.sha1(full_name.encode("utf-8")).hexdigest()[:8]
truncated = full_name[: 64 - 9]  # 9 = len("_") + 8 hash chars
return f"{truncated}_{hash_suffix}"
```
This ensures two different long names always produce distinct 64-char outputs.

Marking as ready for review — let me know if anything else needs adjusting.

PR #2789, opened 2026-03-26
feat(mcp): #2228 expose server name on function tools

### LINKED ISSUE #2228: Add metadata field to RunItemBase for SDK-level tracking
**Problem:**
When MCP tools are converted to `FunctionTool` via `MCPUtil.to_function_tool()`, the server name is only in the `on_invoke_tool` closure. When `ToolCallItem`/`ToolCallOutputItem` are created from API responses, they lose which 
MCP server the tool came from. With multiple MCP servers, you can't tell which server handled a call.

Given that there could be multiple MCPs, and each MCP's tools could change frequently, tracking MCP server labels externally by tool name is not feasible. This also has implications on `StreamEvent` by extension.

**Proposed solution:**
Add a general `metadata` field to `RunItemBase`:

```python
@dataclass
class RunItemBase(Generic[T], abc.ABC):
    agent: Agent[Any]
    raw_item: T
    
    metadata: dict[str, Any] = field(default_factory=dict)
    """SDK-level metadata for tracking and debugging. This metadata is NOT sent to the agent.
    
    Examples:
    - For MCP tools: {"mcp_server_name": "git_server"}
    - For handoffs: {"handoff_source": "agent_a"}
    - For debugging: {"execution_time_ms": 123}
    """
```

**The solution to this specific problem:**
1. Store MCP server name as a private attribute on `FunctionTool` when converting MCP tools
2. When creating `ToolCallItem`/`ToolCallOutputItem`, populate `metadata` from the tool if available
3. Access metadata via `item.metadata.get('mcp_server_name')` when needed

**Additional Benefits:** The solution is very general-purpose, it supports any future metadata needs, not just MCP, as in the examples in the code snippet.

I'd love to know what you think before implementing this. While this may not be a breaking change, it introduces a new feature at the very core of the SDK, and could have other implications I'm not aware of.

IgnazioDS:
## Summary
- add a public `mcp_server_name` field to `FunctionTool`
- populate that field when MCP tools are converted via `MCPUtil.to_function_tool()`
- keep the change narrow and typed instead of introducing a general metadata bag
- add focused tests for MCP population, non-MCP default behavior, and copy/replace preservation

Fixes #2228.

## Why
Issue #2228 asks for a reliable way to recover which MCP server a converted function tool came from. The maintainer feedback there suggested a specific field such as `mcp_server_name: str | None` on `FunctionTool` rather than a generic metadata dict. This patch follows that direction and preserves backward compatibility by making the new field keyword-only with a default of `None`.

## Validation
- `uv run ruff check src/agents/tool.py src/agents/mcp/util.py tests/mcp/test_mcp_util.py tests/test_function_tool.py`
- `PYTHONPATH=src uv run pytest tests/mcp/test_mcp_util.py tests/test_function_tool.py -q`
- `uv run mypy src/agents/tool.py src/agents/mcp/util.py --disable-error-code import-not-found --disable-error-code unused-ignore`

## Notes
The local pytest invocation uses `PYTHONPATH=src` because this workspace clone is not resolving the local `agents` package via `uv run` by default, but the test slice itself passes cleanly.

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -313,6 +313,7 @@ class MCPUtil:
             strict_json_schema=is_strict,
             needs_approval=needs_approval,
             mcp_title=resolve_mcp_tool_title(tool),
+            mcp_server_name=server.name,
         )
         return function_tool
 
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -294,6 +294,9 @@ class FunctionTool:
     defer_loading: bool = False
     """Whether the Responses API should hide this tool definition until tool search loads it."""
 
+    mcp_server_name: str | None = field(default=None, kw_only=True)
+    """The MCP server name this tool originated from, if it was converted from an MCP tool."""
+
     _failure_error_function: ToolErrorFunction | None = field(
         default=None,
         kw_only=True,
@@ -428,6 +431,7 @@ def _build_wrapped_function_tool(
     defer_loading: bool = False,
     sync_invoker: bool = False,
     mcp_title: str | None = None,
+    mcp_server_name: str | None = None,
 ) -> FunctionTool:
     """Create a FunctionTool with copied-tool-aware failure handling bound in one place."""
     on_invoke_tool = with_function_tool_failure_error_handler(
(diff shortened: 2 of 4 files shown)


seratch:
Thanks for sharing this. However, we're going to add the property this way: https://github.com/openai/openai-agents-python/pull/2654

PR #2791, opened 2026-03-26
fix: #2783 depend on griffelib directly for docstring parsing

### LINKED ISSUE #2783: maybe use griffelib instead of griffe as dependency, and update to 2+
my ci and service occasionally crash with
```bash
/usr/local/lib/python3.13/site-packages/agents/tool.py:53: in <module>
    from .function_schema import DocstringStyle, function_schema
/usr/local/lib/python3.13/site-packages/agents/function_schema.py:10: in <module>
    from griffe import Docstring, DocstringSectionKind
/usr/local/lib/python3.13/site-packages/griffe/__init__.py:277: in <module>
    from griffe._internal.expressions import (
E   ImportError: cannot import name 'ExprInterpolation' from 'griffe._internal.expressions' (/usr/local/lib/python3.13/site-packages/griffe/_internal/expressions.py)
```
it may cause by 
```bash
uv pip list | grep griffe
Using Python 3.13.12 environment at: /usr/local
griffe                                       1.15.0
griffelib                                    2.0.0
```
i think they sometimes overwrite each other by installation order?

if using griffelib it may be resolved by uv

seratch:
This pull request fixes #2783 intermittent import conflicts around Griffe by switching the SDK's published runtime dependency from `griffe` 1.x to `griffelib` 2.x. The issue is that both distributions provide the `griffe` module, so mixing them in one environment can produce install-order-dependent breakage even though the SDK only needs the library functionality for docstring parsing.

The change updates the package metadata to depend on `griffelib>=2,<3`, refreshes the lockfile. Because `griffelib` currently exposes the `griffe` module without typing markers, this pull request also adds a scoped `mypy` ignore on the import in `agents.function_schema` to keep type checking green without widening ignores elsewhere.

No user-facing runtime API changes are intended; this is a packaging and tooling fix to make dependency resolution more robust.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ authors = [{ name = "OpenAI", email = "support@openai.com" }]
 dependencies = [
     "openai>=2.26.0,<3",
     "pydantic>=2.12.2, <3",
-    "griffe>=1.5.6, <2",
+    "griffelib>=2, <3",
     "typing-extensions>=4.12.2, <5",
     "requests>=2.0, <3",
     "types-requests>=2.0, <3",
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -7,7 +7,8 @@ import re
 from dataclasses import dataclass
 from typing import Annotated, Any, Callable, Literal, get_args, get_origin, get_type_hints
 
-from griffe import Docstring, DocstringSectionKind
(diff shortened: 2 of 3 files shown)


pawamoy:
> Because griffelib currently exposes the griffe module without typing markers

Ha, good catch, I'll fix that 👍

PR #2268, opened 2026-01-07
Enhance exception handling in MCP server initialization and cleanup

habema:
Fixes an issue where MCP server connection failures (e.g., 401 Unauthorized, connection errors, 500 errors) surfaced as confusing `BaseExceptionGroup` exceptions with `RuntimeError` about cancel scopes, masking the actual HTTP errors.

**What changed:**

- Added `_extract_http_error_from_exception()` helper method to extract HTTP errors from exceptions or `BaseExceptionGroup` raised by anyio task groups
- Added `_raise_user_error_for_http_error()` helper method to wrap HTTP errors in clear, actionable `UserError` messages with specific guidance for different error types (401, 403, 500+, connection errors, timeouts)
- Updated `connect()` to extract HTTP errors from exceptions/ExceptionGroups and wrap them in `UserError`, while preserving non-HTTP errors (e.g., `ValueError`) as-is
- Updated `cleanup()` to catch `BaseExceptionGroup` raised during cleanup, extract HTTP errors from it, and re-raise them as `UserError` while suppressing `RuntimeError` about cancel scopes that would otherwise mask the real error
- Updated `list_tools()` and `call_tool()` to catch `httpx.HTTPStatusError` and `httpx.ConnectError` and wrap them in `UserError` with clear messages
- Updated `invoke_mcp_tool()` in `util.py` to preserve `UserError` exceptions for better error propagation

**Result:**

Cryptic `BaseExceptionGroup` with nested exceptions and `RuntimeError: Attempted to exit cancel scope in a different task` now become clear, actionable errors like:

```
agents.exceptions.UserError: Failed to connect to MCP server 'Streamable HTTP Python Server': Authentication failed (401 Unauthorized). Please check your credentials.
```

The original HTTP error is preserved in the exception chain via `from e`, maintaining full traceback information for debugging.

---

Current logs examples before and after fix:
- 401 Unauthorized: [Before](https://pastebin.com/BYaUh95K) --> [After](https://pastebin.com/9BdyasZr)
- Unreachable: [Before](https://pastebin.com/6mWs5cQi) --> [After](https://pastebin.com/jeJcRmr5)
- 5xx Server Error: [Before](https://pastebin.com/H9YUKXru) --> [After](https://pastebin.com/gY8Awie0)

Repro Scripts:

- <details>
  <summary>main.py</summary>

  ```py
  # main.py
  import asyncio

  from agents import Agent, Runner
  from agents.mcp import MCPServer, MCPServerStreamableHttp
[...]

PR #2802, opened 2026-03-28
fix(mcp): prevent leaked semaphore warnings during MCPServerStdio cleanup

### LINKED ISSUE #618: Resource tracker warning (leaked semaphores) with MCPServerStdio

4444J99:
### Summary

Fixes the "leaked semaphore objects to clean up at shutdown" warning that occurs when using `MCPServerStdio` with MCP servers that use Python multiprocessing internally.

**Root cause:** During `cleanup()`, the `exit_stack.aclose()` unwinds contexts in LIFO order — `ClientSession.__aexit__` runs first (cancelling its internal task group), then the transport's `__aexit__` runs. Because the `ClientSession` cancellation kills the `stdout_reader`/`stdin_writer` tasks inside the transport, the transport's `finally` block — which closes stdin and waits for the subprocess — may never execute cleanly. The subprocess is killed without a chance to flush its multiprocessing resources (semaphores, shared memory, etc.), triggering Python's `resource_tracker` warning.

**Fix:** Before calling `exit_stack.aclose()`, explicitly close the session's write stream. This delivers EOF to the subprocess's stdin, giving it a window to release its resources and exit cleanly before the task groups are cancelled. Also reinitialise `exit_stack` and clear `server_initialize_result` during cleanup so the same server instance can reconnect.

### Test plan

- [x] Added `test_cleanup_resets_state_for_reconnection` — verifies all session state is cleared and exit stack is replaced after cleanup
- [x] Added `test_reconnect_after_cleanup` — verifies a server instance can connect, use tools, cleanup, then reconnect and use tools again
- [x] Added `test_cleanup_closes_write_stream_before_exit_stack` — verifies the write stream close happens before exit stack unwind (the critical ordering invariant)
- [x] All 175 existing MCP tests pass
- [x] `ruff check` and `ruff format` pass

### Issue number

Closes #618

### Checks

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -943,6 +943,22 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
             is_failed_connection_cleanup = self.session is None
 
             try:
+                # Signal the subprocess to shut down gracefully before the exit stack
+                # unwinds.  The exit stack tears down contexts in LIFO order: the
+                # ClientSession exits first (cancelling its internal task group), then
+                # the transport (stdio_client / sse_client) exits.  Because the
+                # ClientSession cancellation kills the reader/writer tasks inside the
+                # transport, the transport's ``finally`` block — which tries to close
+                # stdin and wait for the process — may never execute cleanly.  By
+                # closing the session's write stream ahead of time we deliver an EOF to
+                # the subprocess's stdin, giving it a chance to flush resources (e.g.
+                # multiprocessing semaphores) and exit on its own before we cancel the
+                # task groups.
+                if self.session is not None:
+                    try:
+                        await self.session._write_stream.aclose()
+                    except Exception:
+                        pass  # Best-effort; the stream may already be closed.
                 await self.exit_stack.aclose()
             except asyncio.CancelledError as e:
                 logger.debug(f"Cleanup cancelled for MCP server '{self.name}': {e}")
@@ -1010,6 +1026,10 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
             finally:
                 self.session = None
                 self._get_session_id = None
+                self.server_initialize_result = None
+                # Reinitialize the exit stack so the same server instance can
+                # reconnect cleanly after cleanup.
+                self.exit_stack = AsyncExitStack()
 
 
 class MCPServerStdioParams(TypedDict):
--- a/tests/mcp/test_connect_disconnect.py
+++ b/tests/mcp/test_connect_disconnect.py
@@ -161,8 +161,8 @@ async def test_cleanup_closes_write_stream_before_exit_stack(
         call_order.append("exit_stack_closed")
         return await original_exit_stack_aclose()
 
-    server.session._write_stream.aclose = tracked_write_stream_close  # type: ignore[assignment]
-    server.exit_stack.aclose = tracked_exit_stack_aclose  # type: ignore[assignment]
+    server.session._write_stream.aclose = tracked_write_stream_close  # type: ignore[method-assign]
+    server.exit_stack.aclose = tracked_exit_stack_aclose  # type: ignore[method-assign]
 
     await server.cleanup()
 
--- a/tests/mcp/test_connect_disconnect.py
+++ b/tests/mcp/test_connect_disconnect.py
@@ -1,3 +1,4 @@
+from contextlib import AsyncExitStack
 from unittest.mock import AsyncMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for sharing this. I also explored a viable solution for this, but I found that the sufficient fix requires substantial changes across this SDK. So, I don't think the suggested changes here is enough for the issue. See also: https://github.com/openai/openai-agents-python/pull/2573#issuecomment-3990607859

4444J99:
Understood — happy to contribute to the broader fix in #2573 if that's useful.

4444J99:
@seratch — following up on your feedback from March 29. A few developments since then:

1. **#2573 was closed** (March 2) without merging, so the broader solution it represented hasn't landed.
2. **#2882 was just opened** today, addressing the `AsyncExitStack` reset portion of the reconnect problem — this overlaps with part of what this PR does.
3. This PR (#2802) addresses the root cause more broadly: it closes the write stream *before* unwinding the exit stack, giving the subprocess a clean shutdown window for its multiprocessing resources (semaphores, shared memory). The exit-stack reset and reconnect support are also included.

You mentioned the fix here isn't sufficient and that substantial changes are needed across the SDK. Could you share more about what that broader scope looks like? I'd like to understand the architectural direction so I can iterate meaningfully rather than guess.

Happy to coordinate with #2882 if that helps reduce duplicate effort.

4444J99:
@seratch — following up on the "substantial changes" feedback. I've traced the dependency chain:

**Root cause** is in the MCP library, not the SDK:

- `modelcontextprotocol/python-sdk#1960` — `stdio_client` has a race condition during shutdown. The `stdout_reader` task and cleanup code race when the context exits, causing `BrokenResourceError` and leaked semaphores.
- `modelcontextprotocol/python-sdk#2268` — an open fix PR addressing this (March 16, no review yet).

**What our PR does** is bridge the gap at the SDK level: by closing `session._write_stream` before `exit_stack.aclose()`, we deliver EOF to the subprocess stdin, giving it a chance to flush resources before task group cancellation. This is admittedly reaching into a private attribute of `ClientSession` — which is the fragility you flagged.

**Two paths forward:**

1. **Keep this as a bridge**: our fix works now, prevents the semaphore warning for users, and can be removed once `mcp >= 1.27` (or whichever version includes #2268) is released. We could add a `# TODO: remove once mcp library fixes #1960` comment.

2. **Wait for the library fix**: close this PR, and instead bump the `mcp` dependency lower bound once #2268 lands. No SDK-level workaround needed.

Happy to go either direction based on your preference. If option 1, I can also add a version check to only apply the workaround on `mcp < X` once the threshold is known.

4444J99:
Hi @seratch — checking in on this. I wanted to surface the dependency chain I've traced:

1. **Root cause**: `modelcontextprotocol/python-sdk` issue [#1960](https://github.com/modelcontextprotocol/python-sdk/issues/1960) — `BrokenResourceError` race condition in `stdio_client` cleanup when context exits quickly. Still OPEN.
2. **This PR (#2802)**: Addresses the symptom on the OpenAI SDK side — prevents leaked semaphore warnings during reconnect/shutdown.
3. **#2882**: Addresses a related symptom — `AsyncExitStack` reset after cleanup for reconnect.

Since the root cause lives upstream in the MCP library, both #2802 and #2882 are partial mitigations. Would it be useful if I opened a targeted fix upstream on python-sdk#1960? That would resolve the root cause for all consumers, not just this SDK.

Happy to take any direction here. Thanks!

4444J99:
Closing this out. @seratch, your feedback that this needed broader scope was fair — the cleanup path through MCPServerStdio touches more than what this PR addressed. Since #2573 was also closed without landing and the root cause traces down to the MCP Python SDK itself (the `stdio_client` shutdown race in modelcontextprotocol/python-sdk#1960), it seems like the right fix needs to start there rather than in this SDK. Closing this draft, but happy to return with a fresh approach once the upstream direction solidifies. Thanks for the review time.

PR #2811, opened 2026-03-31
Add offset parameter to get_items() for pagination

### LINKED ISSUE #2810: Feature request: add offset parameter to get_items() for real pagination
### Problem

Currently, `get_items(limit=N)` only supports truncation — it returns the latest N items. There is no way to paginate through a session's history (e.g., "give me items 20-30"). This makes it impossible to efficiently browse or display conversation history in pages.

### Proposed Solution

Add an `offset` parameter to `get_items()`:

```python
# Page 1: 10 most recent items
page1 = await session.get_items(limit=10, offset=0)

# Page 2: next 10 items
page2 = await session.get_items(limit=10, offset=10)

# Skip recent items, get the rest
older = await session.get_items(offset=5)
```

This is backward compatible — `offset` defaults to `0`, preserving existing behavior.

### Motivation

- Building chat UIs with "load more" or infinite scroll
- Exporting conversation history in batches
- Analyzing specific portions of a session without loading the entire history
- Consistent with standard database pagination patterns (`LIMIT`/`OFFSET`)

The existing `limit` parameter already implies the need for pagination but only provides half the mechanism.

ecanlar:
## Summary

- Adds `offset` parameter to `get_items()` across **all** session backends, enabling real pagination over conversation history
- Currently `limit` only truncates (returns the latest N items) — with `offset`, users can now paginate through the full history
- Fully backward compatible: `offset` defaults to `0`

## Usage

```python
# Page 1: 10 most recent items
page1 = await session.get_items(limit=10, offset=0)

# Page 2: next 10 items  
page2 = await session.get_items(limit=10, offset=10)

# Skip the 5 most recent, get the rest
older = await session.get_items(offset=5)
```

## Backends updated

| Backend | File |
|---------|------|
| Session protocol | `src/agents/memory/session.py` |
| SessionABC | `src/agents/memory/session.py` |
| SQLiteSession | `src/agents/memory/sqlite_session.py` |
| AsyncSQLiteSession | `src/agents/extensions/memory/async_sqlite_session.py` |
| SQLAlchemySession | `src/agents/extensions/memory/sqlalchemy_session.py` |
| RedisSession | `src/agents/extensions/memory/redis_session.py` |
| DaprSession | `src/agents/extensions/memory/dapr_session.py` |
| AdvancedSQLiteSession | `src/agents/extensions/memory/advanced_sqlite_session.py` |
| EncryptedSession | `src/agents/extensions/memory/encrypt_session.py` |
| OpenAIResponsesCompactionSession | `src/agents/memory/openai_responses_compaction_session.py` |
| OpenAIConversationsSession | `src/agents/memory/openai_conversations_session.py` |

## Test plan

- [x] All 33 existing session tests pass (backward compatible)
- [x] New `test_sqlite_session_get_items_with_offset` — verifies limit+offset pagination, offset-only, and edge cases

Closes #2810

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -138,12 +138,15 @@ class AdvancedSQLiteSession(SQLiteSession):
     async def get_items(
         self,
         limit: int | None = None,
+        offset: int = 0,
         branch_id: str | None = None,
     ) -> list[TResponseInputItem]:
         """Get items from current or specified branch.
 
         Args:
             limit: Maximum number of items to return. If None, uses session_settings.limit.
+            offset: Number of items to skip (from the most recent) before applying the limit.
+                    Defaults to 0. Combined with limit, enables pagination over history.
             branch_id: Branch to get items from. If None, uses current branch.
 
         Returns:
@@ -161,7 +164,7 @@ class AdvancedSQLiteSession(SQLiteSession):
                 # TODO: Refactor SQLiteSession to use asyncio.Lock instead of threading.Lock and update this code  # noqa: E501
                 with self._lock if self._is_memory_db else threading.Lock():
                     with closing(conn.cursor()) as cursor:
-                        if session_limit is None:
+                        if session_limit is None and offset == 0:
                             cursor.execute(
                                 f"""
                                 SELECT m.message_data
@@ -173,6 +176,7 @@ class AdvancedSQLiteSession(SQLiteSession):
                                 (self.session_id, branch_id),
                             )
                         else:
+                            sql_limit = session_limit if session_limit is not None else -1
                             cursor.execute(
                                 f"""
                                 SELECT m.message_data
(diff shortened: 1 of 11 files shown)


ecanlar:
Thanks for your response!

In my case, I’m developing an assistant (chatbot) that manages a growing history of user sessions in **production**. As the number of sessions increases, retrieving all of them at once is not scalable and negatively impacts both performance and user experience.

Pagination becomes necessary to:

* Efficiently load session history in chunks
* Reduce latency and memory usage
* Support UI patterns like infinite scroll or page-based navigation

Using an offset (or cursor-based pagination) is a standard approach in these scenarios and would make the Sessions API much more practical for real-world applications.

I believe this is not an edge case but a common requirement for any production system that stores conversational history over time.

Thanks for considering it!

seratch:
I mean, just having the offset parameter on the session store side is not enough, because the agent runner does not pass that parameter to your session store. Also, it is not obvious what offset value would be optimal. So, if your app needs to limit the size or number of items, I recommend implementing that logic within the get_items method, such as setting an upper limit on the number of items to return or compacting items at the right time.

PR #2809, opened 2026-03-31
feat: Add agent_users table to associate sessions with users

ecanlar:
## Summary

- Introduces an `agent_users` table across all SQL-based session backends (SQLiteSession, AsyncSQLiteSession, SQLAlchemySession) to associate sessions with application-level users
- Adds optional `user_id` parameter to session constructors and a `get_sessions_for_user()` query method
- Mirrors the User → Session relationship pattern from [Google's ADK](https://github.com/google/adk-python), enabling multi-user agent applications without custom mapping layers

## Motivation

Currently sessions exist independently with no built-in way to link them to users. Every developer building multi-user applications ends up creating their own user-session mapping layer on top of the SDK. This PR adds first-class support for this common pattern.

### Schema

```sql
-- New table
CREATE TABLE agent_users (
    user_id TEXT PRIMARY KEY,
    metadata TEXT,
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);

-- New column on agent_sessions
ALTER TABLE agent_sessions ADD COLUMN user_id TEXT REFERENCES agent_users(user_id) ON DELETE SET NULL;
CREATE INDEX idx_agent_sessions_user_id ON agent_sessions (user_id);
```

### API

```python
# Create a session linked to a user
session = SQLiteSession("session-1", db_path="./data.db", user_id="alice")

# Retrieve all sessions for a user
session_ids = await session.get_sessions_for_user("alice")
```

### Files modified

| File | Change |
|------|--------|
| `src/agents/memory/session.py` | Add `user_id` attribute to `Session` protocol and `SessionABC` |
| `src/agents/memory/sqlite_session.py` | Add `agent_users` table, `user_id` FK, `get_sessions_for_user()` |
| `src/agents/extensions/memory/async_sqlite_session.py` | Same changes for async SQLite backend |
| `src/agents/extensions/memory/sqlalchemy_session.py` | Same changes for SQLAlchemy backend |
| `tests/test_session.py` | Add tests for user association and `user_id` attribute |

## Test plan

- [x] All 34 existing session tests pass (backward compatible — `user_id` defaults to `None`)
- [x] New `test_sqlite_session_user_association` — verifies multi-user session queries
- [x] New `test_sqlite_session_user_id_attribute` — verifies `user_id` is stored on instance

Closes #2808

Issue #2808, opened 2026-03-31
Feature request: agent_users table to associate sessions with users


ecanlar:
### Problem

Currently, `agent_sessions` exist independently without a way to associate them with application-level users. When building multi-user applications, there's no built-in mechanism to link a session to a specific user identity, making it difficult to:

- Retrieve all sessions belonging to a particular user
- Manage user-scoped session history
- Implement user-level session lifecycle management (e.g., cleanup, export, limits)

### Proposed Solution

Introduce an `agent_users` table (or equivalent concept) that allows associating sessions with users, similar to how [Google's Agent Development Kit (ADK)](https://github.com/google/adk-python) handles user-session relationships.

The basic schema would look something like:

```sql
CREATE TABLE agent_users (
    user_id TEXT PRIMARY KEY,
    metadata JSONB,
    created_at TIMESTAMP DEFAULT NOW(),
    updated_at TIMESTAMP DEFAULT NOW()
);

-- Add user_id foreign key to agent_sessions
ALTER TABLE agent_sessions ADD COLUMN user_id TEXT REFERENCES agent_users(user_id);
```

This would enable:
- **User-scoped session queries**: Fetch all sessions for a given user
- **User metadata storage**: Store user-level context that persists across sessions
- **Session ownership**: Clear association between who owns a session
- **Multi-tenant support**: Easier to build applications serving multiple users

### ADK Reference

Google's ADK models this as a `User → Session` relationship where each session is scoped under a user. This is a natural pattern for agent-based applications where:

1. A **user** interacts with the agent
2. The user may have **multiple sessions** (conversations) over time
3. The agent needs to retrieve **user-level context** across sessions

### Additional Context

This is a common need when building production agent applications. Without this, every developer ends up building their own user-session mapping layer on top of the SDK.

seratch:
Thanks for the suggestion. Since most services typically have their own user tables, we generally recommend using a join table between `agent_sessions` and your user tables. This approach allows the `agent_sessions` table to work with different types of user data tables. It may not fit your use case, but we would appreciate your understanding. If you still need an `admin_users` table, please use your own custom session store implementation.

ecanlar:
Thanks for the quick response, I appreciate the clarification! That approach makes sense.

PR #2812, opened 2026-03-31
feat: Add create_session and get_session classmethods

ecanlar:
## Summary

- Adds `create_session(user_id, ...)` classmethod to all SQL-based session backends (SQLiteSession, AsyncSQLiteSession, SQLAlchemySession) that creates a new session with an **auto-generated UUID** `session_id`, persisting user and session rows immediately
- Adds `get_sessions(user_id, ...)` classmethod that retrieves **all sessions for a user** as a list of ready-to-use session instances, with `limit`/`offset` pagination support
- No changes to `SessionABC` or the `Session` protocol — these methods are backend-specific conveniences

### API

```python
# Create a new session (session_id auto-generated)
session = await SQLiteSession.create_session("alice", db_path="./data.db")
print(session.session_id)  # "550e8400-e29b-41d4-..."

# Retrieve all sessions for a user
sessions = await SQLiteSession.get_sessions("alice", db_path="./data.db")
for s in sessions:
    items = await s.get_items()  # each instance is fully functional
    print(s.session_id, len(items))

# With pagination
page = await SQLiteSession.get_sessions("alice", db_path="./data.db", limit=10, offset=0)
```

## Motivation

Builds on #2809 (agent_users table). Currently the caller must invent a `session_id` and the session row is only created lazily on `add_items()`. This PR provides a proper factory pattern where:
1. The `user_id` is the primary entry point (not the `session_id`)
2. The `session_id` is generated internally via `uuid4()`
3. `get_sessions` returns fully functional session instances (not just IDs like `get_sessions_for_user`)

## Test plan

- [x] All 35 existing session tests pass (backward compatible)
- [x] `test_sqlite_create_session` — verifies UUID generation, DB persistence, and normal usage
- [x] `test_sqlite_get_sessions` — verifies retrieval of all user sessions, item preservation, empty result for unknown user
- [x] `test_sqlite_get_sessions_pagination` — verifies limit/offset pagination
- [x] `test_sqlite_create_multiple_sessions_for_user` — verifies multiple sessions per user with unique IDs

### DIFF
--- a/src/agents/extensions/memory/async_sqlite_session.py
+++ b/src/agents/extensions/memory/async_sqlite_session.py
@@ -316,31 +316,33 @@ class AsyncSQLiteSession(SessionABC):
         return session
 
     @classmethod
-    async def get_session(
+    async def get_sessions(
         cls,
         user_id: str,
-        session_id: str,
         db_path: str | Path = ":memory:",
         sessions_table: str = "agent_sessions",
         messages_table: str = "agent_messages",
         users_table: str = "agent_users",
-    ) -> AsyncSQLiteSession | None:
-        """Retrieve an existing session for a user.
+        limit: int | None = None,
+        offset: int = 0,
+    ) -> list[AsyncSQLiteSession]:
+        """Retrieve all sessions for a user.
 
         Args:
-            user_id: The user identifier who owns the session.
-            session_id: The session identifier to retrieve.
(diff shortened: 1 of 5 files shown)


seratch:
For the reason I mentioned at https://github.com/openai/openai-agents-python/issues/2808#issuecomment-4161382094, we don't plan to have this change at least for now.

PR #2735, opened 2026-03-20
feat(tracing): add flush_traces() for long-running workers

mvanhorn:
### Summary

In long-running worker processes (Celery, FastAPI background tasks, RQ, Dramatiq), traces created with `agents.trace()` are buffered by `BatchTraceProcessor` but never exported because the process never exits to trigger shutdown flushing. Users currently have to access private internals (`get_trace_provider()._multi_processor.force_flush()`) to flush traces.

This PR adds a public `flush_traces()` function that delegates to the trace provider's `force_flush()`:

```python
from agents import flush_traces

@celery_app.task
def my_task():
    with trace("my_task"):
        # ... do work ...
        pass
    flush_traces()  # ensure traces are exported
```

### Test plan

- Added `test_flush_traces_calls_provider_force_flush` - verifies the function delegates to the provider
- Added `test_flush_traces_importable_from_agents` - verifies the public API export works
- All existing tests pass (`make tests` - 2405 passed)

### Issue number

Closes #2135

### Checks

- [x] I've added new tests (if relevant)
- [x] I've added/updated the relevant documentation
- [x] I've run `make lint` and `make format`
- [x] I've made sure tests pass

This contribution was developed with AI assistance (Claude Code).

Issue #2135, opened 2025-11-29
Traces are silently dropped in long-running workers


kiansdi:
# [OpenAI Agents SDK] Traces are silently dropped in long-running workers (Celery, FastAPI background tasks, etc.)

**Package:** `openai-agents` (Python)  
**Date:** November 2025  
**Impact:** High – many production users run agents via background workers

### Description
When using `agents.trace()` inside long-living processes (Celery workers, FastAPI background tasks, RQ, Dramatiq, etc.), traces are created correctly (valid trace IDs in logs, nesting works) but **never appear** on https://platform.openai.com/traces.

Reason: the internal `BatchTraceProcessor` only flushes automatically on process shutdown or when the buffer is full. Long-running workers never shut down → traces remain buffered forever.

### Minimal Reproduction
```python
# Celery task – trace never reaches the dashboard
@celery.task
def demo():
    with agents.trace("celery_test"):
        Agent(model="gpt-4o-mini").run("Say hello")
    # ← missing flush → trace lost
```

Running the exact same code as a standalone script works because shutdown triggers the flush.

### Working Fix (3 lines)
```python
from agents.tracing import GLOBAL_TRACE_PROVIDER

def flush_traces():
    GLOBAL_TRACE_PROVIDER._multi_processor.force_flush()
```

Call `flush_traces()` at the end of every task/pipeline:

```python
with agents.trace("my_job"):
    # all agent calls
    ...
flush_traces()   # ← required in workers
```

Result: full nested traces appear instantly on the dashboard.

### Suggested Documentation Addition

> **Note for long-running workers**  
> In Celery, FastAPI background tasks, RQ, Dramatiq, or any long-living worker process, the trace processor does **not** auto-flush.  
> Call `flush_traces()` (or a future public API) at the end of each job, otherwise traces will remain buffered and never appear on the dashboard.

### Why this is hard to discover
- No error/warning is emitted
- All top models (Gemini 3, Claude 4.5, Codex-5, etc.) suggest the usual suspects (env vars, async context, `.trace_id`) – none solve it
- The real answer did not exist publicly until now

Thank you for the excellent Agents SDK – adding this short note would save the community weeks of debugging pain.

yuyuma:
We have the same exact issue as described above (no / inconsistent traces) for our agent that sits behind a FastAPI route, but the suggested fix to force flush does not work for us unfortunately. Unsure whether the root cause is as described (buffered traces) or something else.

[...]

PR #2834, opened 2026-04-03
docs(tracing): add Long-running workers section with flush_traces() examples

nanookclaw:
## Summary

In long-running worker processes (Celery, FastAPI background tasks, RQ, Dramatiq), traces buffered by `BatchTraceProcessor` are silently dropped when the worker picks up the next task — the process never exits to trigger shutdown flushing, so buffered spans are never exported.

This PR documents the `flush_traces()` pattern with concrete examples for the two most common worker runtimes (Celery and FastAPI background tasks), explains *why* traces are dropped in worker environments, and points to #2135 for context.

### Changes

- `docs/tracing.md`: New **Long-running workers** section inserted before **Ecosystem integrations**, covering:
  - Root cause (BatchTraceProcessor + worker process lifecycle)
  - `flush_traces()` pattern with Celery example
  - `flush_traces()` pattern with FastAPI background task example  
  - Note on silent data loss without the flush call

### Relationship to open PRs

PR #2735 adds the `flush_traces()` public API. This PR documents it. The two are complementary — users who discover the worker issue via #2135 will find the solution pattern in the docs.

### Checks

- [x] Documentation change only — no code changes
- [x] Examples tested against current SDK API (`agents.flush_traces`, `agents.trace`, `agents.Runner`)
- [x] Section placement follows existing doc structure (between Additional notes and Ecosystem integrations)

### DIFF
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -139,6 +139,47 @@ await Runner.run(
 ## Additional notes
 - View free traces at Openai Traces dashboard.
 
+## Long-running workers
+
+In long-running worker processes (Celery, FastAPI background tasks, RQ, Dramatiq), traces created
+with `trace()` are buffered by `BatchTraceProcessor` but **may not be exported** before the next
+task begins, because the worker process never exits to trigger shutdown flushing.
+
+Use [`flush_traces()`][agents.flush_traces] after each task completes to ensure all buffered spans
+are exported:
+
+```python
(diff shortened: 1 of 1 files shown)


seratch:
As mentioned at Codex reviews, the utility does not exist in this SDK. We may add something similar in the future. In the case, we'll update the documentation as well.

PR #2835, opened 2026-04-03
feat: add CostTracker for per-agent token cost attribution

NithiN-1808:
### Summary
Adds a `CostTracker` class that plugs into the existing `TracingProcessor` pipeline and gives developers real-time, per-agent token usage and estimated USD cost — without any changes to how agents or runners are used.

---

### Motivation
Developers building multi-agent workflows have no visibility into which agent or handoff is driving API costs. The only current option is checking the OpenAI billing dashboard, which gives totals with no per-agent attribution.

---

### What's added

**`src/agents/cost_tracker.py`**
- `CostTracker` — thread-safe accumulator with `.record()`, `.summary()`, `.total_cost()`, `.total_tokens()`, `.reset()`, and thread-safe `.by_agent` / `.by_model` snapshot properties
- `CostTrackerProcessor` — zero-config hook into the existing tracing pipeline
- Handles both `GenerationSpanData` (legacy `OpenAIChatCompletionsModel` path) and `ResponseSpanData` (default `OpenAIResponsesModel` / Responses API path), with a defensive import so older SDK versions continue to work
- Built-in pricing table for all current models with longest-prefix matching for versioned names (e.g. `gpt-4o-mini-2024-07-18` → `gpt-4o-mini`)
- Custom pricing support via `CostTracker(pricing={...})` for fine-tuned models or when prices change
- Warning logs on unknown models and unresolvable agent attribution — never silently misattributes cost

**`tests/test_cost_tracker.py`** — 12 tests covering pricing math, multi-agent breakdowns, model breakdowns, thread safety, snapshot isolation, and reset behaviour

---

### Usage

```python
from agents.tracing import add_trace_processor
from agents.cost_tracker import CostTracker, CostTrackerProcessor

tracker = CostTracker()
add_trace_processor(CostTrackerProcessor(tracker))

# ... run your agents normally ...

print(tracker.summary())
# {
#   "total_cost_usd": 0.000412,
#   "total_input_tokens": 310,
#   "total_output_tokens": 89,
#   "total_calls": 2,
#   "by_agent": {
#     "Researcher": {"input_tokens": 200, "output_tokens": 60, "cost_usd": 0.000270, "calls": 1},
#     "Writer":     {"input_tokens": 110, "output_tokens": 29, "cost_usd": 0.000142, "calls": 1}
#   },
#   "by_model": {
#     "gpt-4o-mini": {"input_tokens": 310, "output_tokens": 89, "cost_usd": 0.000412, "calls": 2}
#   }
# }
```

Custom pricing for fine-tuned models:
```python
tracker = CostTracker(pricing={
    "ft:gpt-4o-mini:my-org": {"input": 0.0003, "output": 0.0012},
})
```

---

### Design decisions

| Decision | Rationale |
|---|---|
| Zero breaking changes | Purely additive — no existing API modified |
| Both span types handled | Default SDK path uses `ResponseSpanData`, not `GenerationSpanData`; omitting it meant the tracker silently recorded nothing on a standard setup |
| `by_agent`/`by_model` as snapshot properties | Returns a deep copy under lock — thread-safe to read without requiring callers to use `summary()` |
| No fallback to last-known agent | A missed parent ID logs a warning and attributes to `"unknown"` rather than silently billing the wrong agent |
| Custom pricing via constructor | Prices change; fine-tuned models have different rates. Avoids forking the file for custom setups |
| Longest-prefix model matching | Versioned names like `gpt-4o-mini-2024-07-18` resolve correctly without needing an exhaustive list of every dated variant |

---

### Pricing table
Built-in rates are sourced from [openai.com/api/pricing](https://openai.com/api/pricing) and last verified April 2026. The table is easy to update and can be fully overridden per-instance via the `pricing=` argument.

---

### Testing
```
python -m pytest tests/test_cost_tracker.py -v
# 12 passed
```

### DIFF
--- a/src/agents/cost_tracker.py
+++ b/src/agents/cost_tracker.py
@@ -31,30 +31,57 @@ Basic usage::
         #                     "cost_usd": 0.000412, "calls": 2}
         #   }
         # }
+
+Custom pricing::
+
+    tracker = CostTracker(pricing={
+        "gpt-4o": {"input": 0.0025, "output": 0.010},
+    })
 """
 
 from __future__ import annotations
 
+import logging
 import threading
 from collections import defaultdict
 from dataclasses import dataclass
-from typing import Any, TYPE_CHECKING
+from typing import Any
 
-if TYPE_CHECKING:
-    from agents.tracing.spans import Span
-    from agents.tracing.traces import Trace
+# SDK tracing imports — at the top where they belong
+from agents.tracing.processor_interface import TracingProcessor
+from agents.tracing.span_data import AgentSpanData, GenerationSpanData
+
+# ResponseSpanData is the span type emitted by the default OpenAIResponsesModel
+# path. Import it defensively — older SDK versions may not have it.
+try:
+    from agents.tracing.span_data import ResponseSpanData  # type: ignore[attr-defined]
+    _HAS_RESPONSE_SPAN = True
+except ImportError:
+    ResponseSpanData = None  # type: ignore[assignment,misc]
+    _HAS_RESPONSE_SPAN = False
+
+log = logging.getLogger(__name__)
 
 # ---------------------------------------------------------------------------
-# Pricing table — USD per 1,000 tokens (April 2026)
-# Update this table as OpenAI releases new models or changes pricing.
-# See: https://openai.com/api/pricing
+# Pricing table — USD per 1,000 tokens
+# Last verified: April 2026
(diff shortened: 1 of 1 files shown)


seratch:
Hi @NithiN-1808, thanks for your interest! However, we won't have this type of module as part of this core SDK. If you think this could be useful for other developers as well, please feel free to share the code in your own repository and publish it as your own package.

PR #2857, opened 2026-04-08
fix(tracing): prevent RecursionError in truncation of truncated previews

### LINKED ISSUE #2856: RecursionError in tracing truncation: preview dict recursively re-truncated when message content is None
### Describe the bug
When the LLM response contains `None` values and the response needs to be truncated, tracing export sometimes crashes with:

```text
RecursionError: maximum recursion depth exceeded
...
_truncate_json_value_for_limit -> _truncate_mapping_for_json_limit
-> _truncate_json_value_for_limit -> _truncate_mapping_for_json_limit ...
The crash happens during trace sanitization/truncation, which then hides/noises the original run error.
```
**Observed trigger**
A generation payload contains assistant messages like:

```python
{"role": "assistant", "content": None, "tool_calls": [...]}
```
It is identified as the largest key and needs to be truncated. And the `child_budget` is set to 0. Then truncation enters recursive growth and eventually stack overflow.

**Root cause**
In `BackendSpanExporter`:

- `_truncate_json_value_for_limit()` returns `_truncated_preview(value)` for non-str/dict/list values.

- For `None` in  "content", `_truncated_preview()` returns a dict: 
```python
{'truncated': True, 'original_type': 'NoneType', 'preview': '<NoneType truncated>'}
```

- That dict is then treated as a normal mapping again by `_truncate_mapping_for_json_limit()`.

- When key `"truncated"` (bool True) is processed, `bool` is again converted to `_truncated_preview(True)` (dict), replacing a primitive with a nested dict.

- This can repeat recursively under tight budget and cause RecursionError.

So this is not user payload circular reference, but truncation logic recursively re-truncating its own preview structure.

```python
def _truncated_preview(self, value: Any) -> dict[str, Any]:
        type_name = type(value).__name__
        preview = f"<{type_name} truncated>"
        if isinstance(value, dict):
            preview = f"<{type_name} len={len(value)} truncated>"
        elif isinstance(value, (list, tuple, set, frozenset)):
            preview = f"<{type_name} len={len(value)} truncated>"
        elif isinstance(value, (bytes, bytearray, memoryview)):
            preview = f"<{type_name} bytes={len(value)} truncated>"

        return {
            "truncated": True,
            "original_type": type_name,
            "preview": preview,
        }
```

Image

**Minimal traceback shape**

```text
_truncate_json_value_for_limit
  -> _truncate_mapping_for_json_limit
    -> _truncate_json_value_for_limit
      -> _truncate_mapping_for_json_limit
      ...
RecursionError: maximum recursion depth exceeded
```

### Debug information
- Agents SDK version: `0.13.5`
- Python version: `Python 3.12`

### Repro steps

This error only occurs when the LLM response contains `None` values and the response needs to be truncated, making it not easy to reproduce.

### Expected behavior
Tracing sanitization should never recurse infinitely or overflow stack. If payload is too large/unserializable, it should degrade to a stable terminal preview.

Bahtya:
## Problem

When LLM responses contain `None` values and need truncation, `_truncate_json_value_for_limit` crashes with `RecursionError`. The truncation logic re-processes its own preview dicts, creating an infinite loop.

## Root Cause

1. `_truncated_preview()` returns `{"truncated": True, "original_type": "NoneType", "preview": "..."}`
2. `_truncate_json_value_for_limit()` treats this as a regular dict → calls `_truncate_mapping_for_json_limit()`
3. The `True` bool value gets wrapped by `_truncated_preview(True)` into yet another dict
4. This new dict is again treated as a regular mapping → infinite recursion

## Solution

Detect truncated preview dicts (have `"truncated": True`, `"original_type"`, `"preview"` keys) and return them immediately without re-processing.

```python
if value.get("truncated") is True and "original_type" in value and "preview" in value:
    return value
```

Fixes #2856

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -299,7 +299,7 @@ class BackendSpanExporter(TracingExporter):
             # Use a sentinel key to reliably identify SDK-generated previews
             # without matching arbitrary user data that happens to use the
             # same key names.
-            if self._TRUNCATED_PREVIEW_KEY in value:
+            if value.get(self._TRUNCATED_PREVIEW_KEY) is True and "original_type" in value and "preview" in value:
                 return value
             return self._truncate_mapping_for_json_limit(value, max_bytes)
 


seratch:
Thanks for trying to resolve the issue. I came up with an alternative one for this, and it should work well: https://github.com/openai/openai-agents-python/pull/2860

seratch:
Closing this in favor of #2860

PR #2871, opened 2026-04-10
fix: preserve output type in FunctionSpanData.export()

lawrence3699:
### Summary

`FunctionSpanData.export()` applies `str(self.output) if self.output else None`, which corrupts traced tool outputs in two ways:

1. **Dict/list outputs become Python repr strings.** A tool returning `{"status": "ok"}` gets exported as the string `"{'status': 'ok'}"` — single quotes, Python-style booleans — instead of staying as a dict that `json.dumps` can serialize correctly. This means the tracing ingest API receives malformed data.

2. **Falsy but valid outputs are dropped.** A tool returning `0`, `False`, `""`, or `[]` gets exported as `None`, losing the actual result.

The sibling class `GenerationSpanData.export()` passes `self.output` through directly without conversion. This fix aligns `FunctionSpanData` with that pattern.

### Test plan

- Added `tests/tracing/test_span_data.py` with 10 unit tests covering dict, string, None, falsy (0, False, empty string, empty list), list, and numeric outputs.
- Updated MCP tracing snapshot assertions to expect dict outputs instead of Python repr strings.
- Full suite: 2446 passed, 0 failures.
- `make format`, `make lint`, `make typecheck` all clean.

### Issue number

N/A — found via code inspection; no existing issue.

### Checks

### DIFF
--- a/src/agents/tracing/span_data.py
+++ b/src/agents/tracing/span_data.py
@@ -88,7 +88,7 @@ class FunctionSpanData(SpanData):
             "type": self.type,
             "name": self.name,
             "input": self.input,
-            "output": str(self.output) if self.output else None,
+            "output": self.output,
             "mcp_data": self.mcp_data,
         }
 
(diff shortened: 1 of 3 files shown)


Hardanish-Singh (commented):
LGTM!

PR #2874, opened 2026-04-12
Preserve computer action modifier keys

neuralatlasai:
This pull request preserves modifier keys on computer mouse actions instead of silently dropping them before driver execution.

### What changed

- add optional keyword-only `keys` support to `Computer` and `AsyncComputer` mouse-action methods
- thread `action.keys` through computer action dispatch for `click`, `double_click`, `move`, `scroll`, and `drag`
- raise a clear `ModelBehaviorError` when a modifier-bearing action reaches a legacy driver that has not implemented `keys`
- update the Playwright computer example to hold and release modifiers around mouse actions
- add regression coverage and align test doubles with the updated interface
- refresh the English computer-tool docs snippet to show the modifier-capable interface

###  Verification

1. `uv sync --all-extras --all-packages --group dev`
2. `uv run ruff format`
3. `uv run ruff check --fix`
4. `uv run ruff check`
5. targeted `mypy` on changed files
6. targeted `pyright` on changed files
7. targeted `pytest` on impacted suites: `374 passed`

## Notes
Full repo-wide Windows verification still shows unrelated baseline issues in MCP and SQLite/tempfile cleanup paths; those are outside this patch.


seratch:
Thanks for sending this patch. Closing this in favor of #2877.

The issue is real, but the compatibility boundary matters here because `Computer` / `AsyncComputer` are released public extension points. The follow-up PR addresses modifier-key support without changing the abstract method signatures, forwards `keys` only to drivers that support them, and preserves legacy-driver behavior by warning and continuing instead of raising.

PR #2880, opened 2026-04-13
fix: preserve redacted_thinking blocks in multi-turn Bedrock conversations

laiman1107:
## Problem

When using Claude with extended thinking in multi-turn agent conversations, thinking blocks are silently dropped during history reconstruction, causing providers to reject subsequent turns.

Two concrete failure modes, both caused by the same bug:

### 1. AWS Bedrock — `redacted_thinking` blocks

Bedrock returns `redacted_thinking` blocks with a `data` field (base64-encrypted content) instead of `thinking`/`signature`:

```json
{"type": "redacted_thinking", "data": "SGVsbG8..."}
```

These blocks were completely invisible to the serialisation code — they have no `"thinking"` key and no `"signature"` key — so they were silently dropped. Bedrock then rejected the next turn with:

```
litellm.BadRequestError: BedrockException - {"message":"The model returned the following errors:
messages.1.content.26: `thinking` or `redacted_thinking` blocks in the latest assistant message
cannot be modified. These blocks must remain as they were in the original response."}
```

### 2. Anthropic API — `display: "omitted"` thinking blocks

When thinking is configured with `display: "omitted"`, Claude returns blocks with an empty `thinking` field but a valid `signature`:

```json
{"type": "thinking", "thinking": "", "signature": "EosnCkYICxIMMb3..."}
```

Reference: https://platform.claude.com/docs/en/build-with-claude/extended-thinking#redacted-thinking-blocks

These blocks were filtered out of `content_items` (because `thinking == ""` is falsy), meaning `content_items` was empty. The reconstruction guard `and content_items` then short-circuited and skipped the entire block — even though `encrypted_content` still held the signature.

---

### Root cause (shared)

In `message_to_output_items`, blocks were decomposed into two separate fields:
- `reasoning_item.content` — thinking text (filtered: skips empty and missing `thinking`)
- `reasoning_item.encrypted_content` — signatures joined by `"\n"` (skips missing `signature`)

In `items_to_messages`, reconstruction required `content_items` to be non-empty as a guard. Any scenario where `content_items` was empty (omitted thinking, redacted thinking) caused the entire block list to be silently dropped.

The Anthropic API requirement is clear: **thinking and redacted_thinking blocks must be passed back verbatim** in multi-turn conversations. Any modification or omission is rejected.

## Fix

Serialise the **complete block list as JSON** in `encrypted_content` so every block type survives the round-trip unchanged:

- `message_to_output_items`: replace per-field extraction with `json.dumps(blocks_as_dicts)`. Still populates `content` with visible thinking text for display/summary purposes.
- `items_to_messages`: try `json.loads(encrypted_content)` first (new format, handles all block types). Fall back to the legacy `"\n"`-joined signatures format so any in-flight sessions are not broken. Replace `and content_items` guard with `and encrypted_content` — omitted/redacted blocks have no content items but do have encrypted_content.

## Tests

- Updated two existing assertions that checked `encrypted_content == "sig1\nsig2"` to verify the JSON round-trip instead.
- **`test_redacted_thinking_blocks_preserved_across_turns`**: pure `redacted_thinking` block survives `message_to_output_items` → `items_to_messages` verbatim.
- **`test_mixed_thinking_and_redacted_thinking_blocks_preserved`**: mixed `thinking` + `redacted_thinking` blocks are all preserved in original order.

All 44 tests in the affected files pass.

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -147,7 +147,14 @@ class Converter:
 
             # Store thinking blocks for Anthropic compatibility
             if hasattr(message, "thinking_blocks") and message.thinking_blocks:
-                blocks_as_dicts = [b for b in message.thinking_blocks if isinstance(b, dict)]
+                # Only include blocks that carry a "type" field.  The last-resort
+                # fallback in LitellmConverter produces {"thinking": str(block)}
+                # dicts without a "type"; including them would cause the replay-
+                # side JSON validation to fall back to legacy parsing and
+                # misinterpret the whole JSON string as a newline-joined signature.
+                blocks_as_dicts = [
+                    b for b in message.thinking_blocks if isinstance(b, dict) and b.get("type")
+                ]
 
                 # Serialise the full blocks as JSON so that both thinking and
                 # redacted_thinking blocks can be reconstructed verbatim on the
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -147,14 +147,21 @@ class Converter:
 
             # Store thinking blocks for Anthropic compatibility
             if hasattr(message, "thinking_blocks") and message.thinking_blocks:
-                # Only include blocks that carry a "type" field.  The last-resort
-                # fallback in LitellmConverter produces {"thinking": str(block)}
-                # dicts without a "type"; including them would cause the replay-
-                # side JSON validation to fall back to legacy parsing and
-                # misinterpret the whole JSON string as a newline-joined signature.
-                blocks_as_dicts = [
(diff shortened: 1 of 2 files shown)


laiman1107:
Good catch. The  guard did regress the case where older persisted turns have content items but no encrypted_content (e.g. thinking blocks whose signature was absent).

Fixed in the latest commit by using  as the gate, then branching inside:
- **encrypted_content present** → try JSON, fall back to legacy -joined signatures
- **encrypted_content absent, content_items present** → reconstruct signature-less thinking blocks from content text only (restores previous behaviour for those older sessions)

seratch:
I will look into this when I have time for it, but if other people who are building with these platforms could help us check if this works well, that'd be greatly appreciated.

PR #2883, opened 2026-04-14
fix: tolerate None text in ResponseOutputText content items

beveradb:
### Summary

`ItemHelpers.extract_text` does `text += content_item.text`, which raises `TypeError: can only concatenate str (not "NoneType") to str` when a `ResponseOutputText` content item has `text=None`.

Although the Responses API schema declares `text: str`, this surfaces in practice when output items are assembled via `model_construct` during streaming or forwarded through provider gateways (e.g. LiteLLM partial responses). The crash propagates up from inside `execute_tools_and_side_effects` (called from `run_single_turn_streamed`), aborting the entire agent turn.

We hit this in production on an agent backed by a LiteLLM → Azure OpenAI Responses gateway: ~57 user-facing failures per day, all surfacing as the same unhelpful "Something went wrong" message. The full traceback ends in:

```
File ".../agents/run_internal/turn_resolution.py", line 618, in execute_tools_and_side_effects
    ItemHelpers.extract_text(message_items[-1].raw_item) if message_items else None
File ".../agents/items.py", line 686, in extract_text
    text += content_item.text
TypeError: can only concatenate str (not "NoneType") to str
```

Fix: coerce with `or ""` so a missing `.text` contributes nothing instead of crashing. Behavior on well-formed messages is unchanged (covered by the existing `test_extract_text_concatenates_all_text_segments`).

### Test plan

- `uv run pytest tests/test_items_helpers.py` — all 26 tests pass, including the new regression test.
- New test `test_extract_text_tolerates_none_text_content` covers:
  - single `ResponseOutputText` with `text=None` → returns `None` (per `return text or None`)
  - mixed content with `None` text + real text → returns the real text, both orderings.
- `make lint` and `make format` clean.

### Issue number

N/A — opening a fresh report.

### Checks

- [ ] I've added/updated the relevant documentation — none needed; method behavior on the documented happy path is unchanged.


seratch:
@codex review again

seratch (approved):

Sourav-Nandy-ai:
Hi Andrew,

I ran an actual before-and-after test on #2883 rather than just trusting the description. I installed the real PyPI-boundary versions bracketing your merge: openai-agents==0.14.1 (published 29 minutes before the merge landed) as the pre-fix, and 0.14.2 (the very next release) as the post-fix. I confirmed via the GitHub API comparison (gh api compare/<your-merge-commit>...<tag>) that 0.14.1 lacks the fix and 0.14.2 contains it.

Instead of a live model call, I used the exact trigger you described, where model_construct produces text=None on a ResponseOutputText. I ran it through the SDK's real, installed ItemHelpers.extract_text, which is the function you patched.

The Results
- Pre-fix: extract_text raised TypeError: can only concatenate str (not "NoneType") to str on every one of 50 deterministic samples across 5 test cases.
- Post-fix: It never raised an error once. It always returned the correctly concatenated text, matching your text += content_item.text or "" fix exactly. I confirmed the installed source diff on both sides as well.

I maintain agent-eval (agent_regress on PyPI), an OSS harness for exactly this kind of cross-version behavioral comparison. Its subprocess_runner() dispatches a test suite to two separately installed package versions, and compare() returns a statistical verdict. In this case, it correctly reported IMPROVED with p ≈ 2.6e-23 and zero-variance on both sides, since this bug is 100% reproducible and not flaky.

Given you clearly do hands-on gateway and streaming debugging in production, I am curious whether a harness like this would be useful for catching this class of "schema says str but reality says None" regression before it ships, especially since it pins to real installed SDK versions rather than mocks.

You can find the full script and raw output here: https://github.com/RudrenduPaul/agent-eval

Best,
Sourav

PR #2898, opened 2026-04-15
fix: synthesize assistant message for session when tool_use_behavior produces final output

### LINKED ISSUE #1465: Tool not invoked on consecutive calls when using SQLiteSession in OpenAI Agent SDK
When using SQLiteSession with Runner.run() in the OpenAI Agent SDK, tools are not invoked on every call if the same input is repeated. Instead, the second, fourth, sixth (even-numbered) calls return a cached/remembered answer directly without calling the tool.

If I remove the session parameter from Runner.run(), the tool is invoked every time as expected.

Code to Reproduce
from agents import Agent, Runner, SQLiteSession, function_tool

session = SQLiteSession("conversation_123")

@function_tool
async def find_prime_minister(name: str) -> str:
    """
    Find the current Prime Minister of a given country.
    """
    print(f"++++++++Finding Prime Minister for {name}++++++++")
    if name.lower() == "canada":
        return "Prime Minister of Canada is Justin Trudeau"
    elif name.lower() == "uk":
        return "Prime Minister of UK is Rishi Sunak"
    elif name.lower() == "pakistan":
        return "Prime Minister of Pakistan is Mian Shahbaz Sharif"
    else:
        return "Prime Minister not found for this country."

agent = Agent(
    name="Assistant",
    tools=[find_prime_minister],
    tool_use_behavior="stop_on_first_tool",
)

# First call
result = await Runner.run(
    agent,
    "Who is the current Prime Minister of Canada?",
    session=session
)
print(result.final_output)

# Second call with the same input
result = await Runner.run(
    agent,
    "Who is the current Prime Minister of Canada?",
    session=session
)
print(result.final_output)

Expected Behavior
The tool find_prime_minister should be invoked on every call, regardless of session state, so that: 
++++++++Finding Prime Minister for Canada++++++++
Prime Minister of Canada is Justin Trudeau
is printed on every run.

Actual Behavior
First call: Works as expected (tool is invoked)

Second call: Tool is not called; answer is returned directly from somewhere in the session history.

Third call: Works again

Fourth call: Skips tool call again
This alternates (tool called on odd calls, skipped on even calls).

Notes
Removing session=session from Runner.run() makes the tool run every time.

This suggests that the session memory mechanism is returning a previous answer without re-triggering the tool, possibly as a caching or reasoning optimization.

However, for tools with side effects or real-time data, skipping the tool can lead to incorrect or outdated results.

Nik-Reddy:
## Problem

When using `tool_use_behavior="stop_on_first_tool"` with a `SQLiteSession`, the agent exhibits an alternating pattern where a tool fires on one call, gets **skipped** on the next, fires again, and so on. This makes persistent-session agents with tool-stopping behavior unreliable for production use.

### Root Cause

When `tool_use_behavior` turns a tool's output into the final answer, the session was saving raw `function_call` + `function_call_output` items to history. On the next run, `prepare_input_with_session` loads them back into the model's context — the model sees the previous tool invocation already completed and short-circuits, skipping the tool call entirely.

## Solution

Two targeted changes that keep the session history clean:

### 1. Synthesize an assistant message for session storage (`turn_resolution.py`)

When a tool output becomes the final answer via `tool_use_behavior`, we now store a clean `MessageOutputItem` (assistant text message containing the tool's output) instead of the raw `function_call` + `function_call_output` items. This means subsequent runs see a normal assistant reply in history, not a completed tool invocation.

```python
# Before: session stores raw tool call artifacts
# [function_call, function_call_output]  ← model skips tool on next run

# After: session stores a clean assistant message
# [message(role=assistant, text="PM is Justin Trudeau")]  ← model behaves normally
```

### 2. Guard tool-call back-fill logic (`run.py`)

The existing back-fill logic (which finds `function_call` items matching `function_call_output` items and adds them to session storage) needed a guard to skip when `session_step_items` is explicitly set. Without this, the back-fill would re-add raw `function_call` items even though we'd already overridden them with the synthesized message.

## Testing

Added a regression test (`test_session_stop_on_first_tool_saves_assistant_message`) that:

1. Creates an agent with `stop_on_first_tool` + `SQLiteSession`
2. Runs the agent — verifies session contains an **assistant message** (not `function_call` items)
3. Runs again — verifies the model input on the second call **does not** contain stale tool call history
4. Confirms the tool fires correctly on both consecutive calls

## Files Changed

| File | Change |
|------|--------|
| `src/agents/run_internal/turn_resolution.py` | Synthesize `MessageOutputItem` for session when tool output is final answer |
| `src/agents/run.py` | Guard tool-call back-fill to respect explicit `session_step_items` |
| `tests/test_session.py` | Regression test for consecutive tool invocations with persistent session |

Fixes #1465

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1119,30 +1119,36 @@ class AgentRunner:
                                 item for item in items_to_save_turn if item.type != "tool_call_item"
                             ]
                         if session_persistence_enabled:
-                            output_call_ids = {
-                                item.raw_item.get("call_id")
-                                if isinstance(item.raw_item, dict)
-                                else getattr(item.raw_item, "call_id", None)
-                                for item in turn_result.new_step_items
-                                if item.type == "tool_call_output_item"
-                            }
-                            for item in generated_items:
-                                if item.type != "tool_call_item":
-                                    continue
-                                call_id = (
+                            # When session_step_items is explicitly set, the turn
+                            # resolution has already decided what should be persisted
+                            # (e.g. a synthesized assistant message instead of raw
+                            # tool call items).  Skip the tool-call back-fill so we
+                            # don't re-add the items that were intentionally excluded.
+                            if turn_result.session_step_items is None:
+                                output_call_ids = {
                                     item.raw_item.get("call_id")
                                     if isinstance(item.raw_item, dict)
                                     else getattr(item.raw_item, "call_id", None)
-                                )
-                                if (
-                                    call_id in output_call_ids
-                                    and item not in items_to_save_turn
-                                    and not (
-                                        run_state
-                                        and run_state._current_turn_persisted_item_count > 0
+                                    for item in turn_result.new_step_items
+                                    if item.type == "tool_call_output_item"
+                                }
+                                for item in generated_items:
+                                    if item.type != "tool_call_item":
+                                        continue
+                                    call_id = (
+                                        item.raw_item.get("call_id")
+                                        if isinstance(item.raw_item, dict)
+                                        else getattr(item.raw_item, "call_id", None)
                                     )
-                                ):
-                                    items_to_save_turn.append(item)
+                                    if (
+                                        call_id in output_call_ids
+                                        and item not in items_to_save_turn
+                                        and not (
+                                            run_state
+                                            and run_state._current_turn_persisted_item_count > 0
+                                        )
+                                    ):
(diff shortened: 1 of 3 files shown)


seratch:
This is an intended behavior of stop_on_first_tool, so we don't plan to change the behavior. If this is not suitable for your use case, I don't think stop_on_first_tool is the right option for it.

PR #1364, opened 2025-08-04
Mongodb memory session

DamyanBG:
## ✨ Add MongoDB-Based Async Session Store

This PR adds an asynchronous session store implementation using **MongoDB** and the new async API available in `pymongo >=4.13`. The issue is #1362.

### 🔧 Summary

- Implements `MongoDBSession`, conforming to the existing `SessionABC` interface.
- Fully asynchronous using `pymongo`’s async client (`AsyncMongoClient`).
- Schema-less but initializes collections and indexes if they do not already exist.
- Provides full support for:
  - Session initialization
  - Adding, retrieving, and removing session messages
  - Clearing session data
  - Context-managed resource cleanup via `__aenter__` / `__aexit__`

### 📦 Dependencies

This implementation relies on:

```
pymongo>=4.13  # or compatible async version
```

Users can install this with:

```bash
pip install 'openai-agents[pymongo]'
```

### ✅ Next Steps

- I will add unit tests once the overall implementation pattern is validated.
- I plan to align with the other session store PRs (`PostgreSQL`, `Redis`, `SQLAlchemy`) once they are finalized and merged to ensure consistency in naming, structure, and behavior.

### 🧠 Notes

- Indexes are created on `session_id` for `agent_sessions`, and on `session_id` + `created_at` for `agent_messages`.
- A `_initialized` flag ensures index creation runs only once per session instance.
- The code gracefully skips invalid JSON entries in the database.
- I’ve used `json.dumps()` and `json.loads()` for storing/retrieving items, matching the format used by other stores.

PR #2902, opened 2026-04-15
feat(extensions): add MongoDB session backend

alexbevi:
### Summary

Adds `MongoDBSession` to `src/agents/extensions/memory/`, a MongoDB-backed session store. I followed the `extensions/memory` directory structure established in #1328 as this was suggested  as a path forward on the previous (now closed) PR #1364.

Notes:
- Two-collection schema: `agent_sessions` (metadata) + `agent_messages` (individual conversation items), with a compound index on `(session_id, _id)` for efficient per-session chronological queries.
- Index creation is idempotent — one-shot per `(id(client), database, sessions_collection, messages_collection)` key, guarded by a per-key `asyncio.Lock`. The client identity is included so that different `AsyncMongoClient` instances (e.g. pointing at different clusters) each get their own init pass.
- `from_uri()` factory owns the client lifecycle; injected clients are left unmanaged (caller's responsibility).
- Supports `SessionSettings.limit` and explicit per-call `limit` overrides, consistent with all other session backends.
- Corrupted or malformed documents are silently skipped rather than raising.

Install: `pip install openai-agents[mongodb]`

### Issue number

Relates to #1364, #1328

### Checks


seratch (changes requested):
Thanks for putting this together. I think we should resolve one compatibility issue before merging.

Right now we advertise `pymongo>=4.13`, but the constructor calls `client.append_metadata(...)` behind a `hasattr()` check. `append_metadata()` is a 4.14-era API, and on older PyMongo clients attribute access is not a safe feature check here because valid attribute names can resolve to databases. In practice, that means the injected-client path can fail on the minimum version we currently declare.

Can we either:
1. raise the MongoDB extra floor to 4.14, or
2. keep 4.13 support and switch this to a callable/version-safe guard, then add coverage for that real compatibility path instead of relying only on the fully mocked `pymongo` test setup?

If there is no specific reason to support 4.13, my preference is explicitly supporting 4.14+

alexbevi:
> If there is no specific reason to support 4.13, my preference is explicitly supporting 4.14+

@seratch bumped to 4.14+ and removed hasattr() check. The 4.13 check is more of a legacy test in the event you're mixing PyMongo with AsyncPyMongo, which was introduced in 4.13. It's not needed here.

seratch:
@alexbevi Thanks for updating this PR. The part looks good to me now. Can you continue working on this PR until Codex says 👍? The Codex reviews will be triggered every time you push new commits in this repo.

seratch (approved):
Thank you!

Sourav-Nandy-ai:
Hi Alex,

Nice work on the MongoDB session backend here. The two-collection schema and the idempotent index setup make for a clean design. Session and memory backends like this are exactly the kind of change that can quietly shift how an agent behaves across turns in ways a single before-and-after check will not catch. State persists, but whether it is behaving the same way it did pre-change is a distribution question rather than a simple pass-or-fail one.

I maintain agent-eval (agent_regress on PyPI), a stats-based regression harness for agent behavior that calculates metrics like Mann-Whitney U, Cohen's d, and bootstrap CI across N runs.

When I first looked at this PR, my OpenAI Agents SDK integration lacked a session concept entirely and had an unrelated crash bug. Both issues are now fixed. The integration now calls the SDK's actual Runner.run() entrypoint, and it supports injecting a session (or session-factory) into each test case. This includes per-case session reuse, ensuring a "conversation" is treated as a first-class unit instead of getting silently flattened.

I am not claiming any specific result against your code, as I have not run the harness against this PR. However, if you ever want to sanity-check that a session backend is not quietly shifting output distributions across a change, I would genuinely value your feedback. It would be highly useful to get your read on whether my session-wiring API makes sense from the perspective of someone who has built one of these for real.

Best,
Sourav

alexbevi:
@Sourav-Nandy-ai if your harness surfaces anything surprising or interesting, let me know!

Sourav-Nandy-ai:
> @Sourav-Nandy-ai if your harness surfaces anything surprising or interesting, let me know!

@alexbevi:
I ran the tests scoped to the storage layer. No LLM calls were involved, making this a session-backend-only check rather than a full output-distribution regression.

I conducted two checks against a live local MongoDB 7.0.14 instance:

1) Contract validation: Hand-inserted a malformed document into agent_messages via PyMongo, bypassing the SDK. get_items() silently dropped it and returned the 3 valid items intact, matching the documented contract.

2) Concurrency and cancellation: Compared the cold index-setup path against a pre-warmed one under load (20 concurrent, 30% cancelled, 30 trials each via a custom compare_liveness() function). Results showed zero measurable difference (p=1.0, d=0.0), with 20/20 completing without failure in a bare sanity pass.

In conclusion, the threading.Lock around a boolean flag design holds up under real concurrent load. 

Hope this is useful.

PR #2905, opened 2026-04-16
feat: expose conversation history to tools via ToolContext.conversation_history

### LINKED ISSUE #904: When running in non-streaming mode, retrieve past history during tool execution.
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) Yes
- **Have you searched for related issues?** Others may have had similar requests Yes

### Describe the feature

Currently, I'm developing a tool that calls other models. It needs to read the parent agent's history to complete specific tasks, but I haven't found a good way to get this information in a non-streaming context. Right now, I have to resort to methods like this to retrieve it.

``` python
context = Context()

r = Runner.run_streamed(
    agent,
    """You are solving a Github issue in the repository.
    You must make the changes to solve, close, or address the issue, directly in the code.""",
    context=context,
    max_turns=200,
)
                                                                                               
context.run_result_streaming = r
await wait_rrs(r)
```

Then read `context.run_result_streaming.new_items`  it in tool.

When I run in non-streaming mode, I haven't found a way to get `new_items` during the run.

```python
await Runner.run(
    agent,
    """You are solving a Github issue in the repository.
    You must make the changes to solve, close, or address the issue, directly in the code.""",
    context=context,
    max_turns=200,
)
```

# Q&A

- Why I need this feature

I'm developing an agent that needs to call other inference models during its operation to perform complex inferences on the agent's historical records. I considered `handoff`, but found that managing the transfer and return of control was quite cumbersome and unstable. Therefore, I had to use tools to call other inference models. 

- Why I need to use non-streaming

The agent I'm currently developing is not human-facing, so the simplest non-streaming is sufficient. Moreover, my internal environment doesn't properly support streaming interfaces, making it difficult to use them extensively.

Nik-Reddy:
Fixes #904

I ran into this while building an agent that needed tools to look at what happened earlier in the conversation, for example, summarizing prior tool outputs or deduplicating work. Right now there's no clean way to do that from inside a tool function.

This adds a `conversation_history` property to `ToolContext` that gives tools read-only access to the items generated so far in the current run. It's a copy, so tools can't accidentally mutate run state.

**Changes:**

- `RunContextWrapper` gets an internal `_generated_items` list (populated in `turn_resolution.py` right before tool execution)
- `ToolContext` exposes it as a read-only `conversation_history` property
- Added `ToolContext` to the package's public API (`__init__.py` / `__all__`)
- 4 new tests covering: empty default, population from wrapper, copy semantics, constructor path

Kept it minimal, didn't touch streaming or any other context classes. Happy to adjust the approach if maintainers prefer a different pattern.

### DIFF
--- a/src/agents/__init__.py
+++ b/src/agents/__init__.py
@@ -176,6 +176,7 @@ from .tool import (
     resolve_computer,
     tool_namespace,
 )
+from .tool_context import ToolContext
 from .tool_guardrails import (
     ToolGuardrailFunctionOutput,
     ToolInputGuardrail,
@@ -413,6 +414,7 @@ __all__ = [
     "CompactionItem",
     "AgentHookContext",
     "RunContextWrapper",
+    "ToolContext",
     "TContext",
     "RunErrorDetails",
     "RunErrorData",
--- a/src/agents/run_context.py
+++ b/src/agents/run_context.py
@@ -61,6 +61,9 @@ class RunContextWrapper(Generic[TContext]):
     tool_input: Any | None = None
     """Structured input for the current agent tool run, when available."""
 
+    _generated_items: list[Any] = field(default_factory=list, repr=False)
+    """Internal: conversation items generated so far in the current run."""
+
     @staticmethod
     def _to_str_or_none(value: Any) -> str | None:
         if isinstance(value, str):
--- a/src/agents/run_internal/turn_resolution.py
+++ b/src/agents/run_internal/turn_resolution.py
@@ -1890,6 +1890,10 @@ async def get_single_step_result_from_response(
 
     tool_use_tracker.record_processed_response(item_agent, processed_response)
 
+    # Expose the conversation history so tools can inspect prior items via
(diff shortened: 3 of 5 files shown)


seratch:
Hi @Nik-Reddy, thanks for taking the time to work on this. We appreciate the interest here, but this PR is not aligned with the direction we want to take for the issue.

The original request in #904 was not intended to broaden tool access to run history, and we would like to avoid exposing this kind of data through tools. We also do not want to carry this data on `RunContextWrapper`, even as a private or internal field.

The main reason #904 has remained open for a while is that it requires careful interface design to preserve the SDK's current philosophy and boundaries. Because of that, we do not think this approach is the right one for us to pursue.

To avoid spending more time on a direction we are unlikely to merge, we are going to close this PR for now. We may revisit #904 and resolve it in a different way in the future, but this is not the path we want to take. Thanks again for the contribution and for taking the time to explore this.

PR #2911, opened 2026-04-16
feat(lifecycle): add on_turn_start / on_turn_end hooks with TurnControl

### LINKED ISSUE #2671: Feature request: better support for agent state changes between turns
## Summary

## Why this matters

```
user: user_message
assistant: [<tool_use2.1>]
user: [<tool_result2.1>]
assistant: assistant_message
```

```
user: user_message
assistant: [<tool_use2.1>]
<new user message arrives>
assistant: assistant_message
```

```
user: compacted message
assistant: [<tool_use2.1>]
assistant: assistant_message
```

## Current limitation

Thanks for considering this.

adityasingh2400:
## Summary

Adds turn-level lifecycle hooks to `RunHooksBase` and `AgentHooksBase`, addressing #2671.

### What's new

- `on_turn_start(context, agent, turn_number)` — fires before the LLM is called for each turn
- `on_turn_end(context, agent, turn_number)` — fires after all tool calls for a turn complete
- `TurnControl` return type for `on_turn_start`: returning `"stop"` halts the run gracefully (raises `MaxTurnsExceeded`) *before* the model is called; returning `None` or `"continue"` proceeds normally

### Addressing reviewer feedback

> _Hooks can raise an exception if something wrong but they do not affect the agent loop orchestration in a more granular way many developers wish._
> — @seratch

`on_turn_start` now returns `TurnControl` (`"stop" | "continue" | None`), giving callers direct control over the loop — not just observation. This allows patterns like:

```python
class BudgetHooks(RunHooks):
    def __init__(self, max_turns: int):
        self.max_turns = max_turns

    async def on_turn_start(self, context, agent, turn_number) -> TurnControl:
        if turn_number > self.max_turns:
            return "stop"  # halts before the LLM call
        return None

    async def on_turn_end(self, context, agent, turn_number):
        print(f"Turn {turn_number} complete. Tokens used: {context.usage.total_tokens}")
```

When `"stop"` is returned, `MaxTurnsExceeded` is raised — consistent with the existing `max_turns` limit, so callers handle it the same way.

### Files changed

| File | Change |
|---|---|
| `src/agents/lifecycle.py` | `on_turn_start` return type → `Union[TurnControl, None]`; docstrings updated |
| `src/agents/run.py` | Non-streaming loop checks return value; raises `MaxTurnsExceeded` on `"stop"` |
| `src/agents/run_internal/run_loop.py` | Streaming loop checks return value; enqueues `QueueCompleteSentinel` on `"stop"` |
| `src/agents/__init__.py` | Exports `TurnControl`, `RunHooksBase`, `AgentHooksBase` |
| `tests/test_turn_lifecycle_hooks.py` | 10 tests (6 existing + 4 new `TurnControl` cases) |

Closes #2671

### DIFF
--- a/src/agents/__init__.py
+++ b/src/agents/__init__.py
@@ -65,7 +65,7 @@ from .items import (
     ToolCallOutputItem,
     TResponseInputItem,
 )
-from .lifecycle import AgentHooks, RunHooks
+from .lifecycle import AgentHooks, AgentHooksBase, RunHooks, RunHooksBase, TurnControl
 from .memory import (
     OpenAIConversationsSession,
     OpenAIResponsesCompactionArgs,
@@ -361,7 +361,10 @@ __all__ = [
     "ReasoningItem",
     "ItemHelpers",
     "RunHooks",
+    "RunHooksBase",
     "AgentHooks",
+    "AgentHooksBase",
+    "TurnControl",
     "Session",
     "SessionABC",
     "SessionSettings",
--- a/src/agents/lifecycle.py
+++ b/src/agents/lifecycle.py
@@ -1,4 +1,6 @@
-from typing import Any, Generic, Optional
+from __future__ import annotations
+
+from typing import Any, Generic, Literal, Optional, Union
 
 from typing_extensions import TypeVar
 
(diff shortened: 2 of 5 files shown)


seratch:
Adding more hooks may be worth considering, but I don't think this could be a solution for the mentioned issue. Hooks can raise an exception if something wrong but they do not affect the agent loop orchestration in a more granular way many developers wish.

adityasingh2400:
Thanks for the feedback @seratch! You're right — pure observation hooks don't solve the orchestration-control use case.

I've updated this PR so that `on_turn_start` now returns a `TurnControl` value (`"stop" | "continue" | None`). Returning `"stop"` halts the run before the LLM is called for that turn, raising `MaxTurnsExceeded` (consistent with the existing `max_turns` behaviour). Returning `None` or `"continue"` proceeds as before.

This gives callers actual control over the loop — e.g. a custom turn budget, an external kill-switch, or circuit-breaking based on `context.usage`. All 10 tests pass including 4 new `TurnControl` cases.

seratch:
Thanks for the update. However, this is still not the direction we'd like to pursue, so let us close this now. Also, all the PRs you're sending to this repo have conflicts with the latest main branch. If you're interested in sending further PRs to this project, please make sure those do not have conflicts.

PR #2912, opened 2026-04-16
feat(lifecycle): add on_tool_authorize hook for per-tool authorization

adityasingh2400:
## Summary

There's currently no way to programmatically block tool execution at the hook layer — you can observe calls via `on_tool_start`, but you can't prevent them. This PR adds `on_tool_authorize` to both `RunHooksBase` and `AgentHooksBase` so callers can intercept tool calls before they happen.

Closes #2868

## What changed

- Added `on_tool_authorize(context, agent, tool) -> bool` to `RunHooksBase` and `AgentHooksBase` in `lifecycle.py`
- Wired the check in `_execute_single_tool_body` inside `tool_execution.py`, between input guardrails and `on_tool_start`
- Both run-level and agent-level hooks are consulted; either can deny the call
- Default returns `True` — fully backwards compatible

## Behavior when denied

- Tool function is never invoked
- `on_tool_start` and `on_tool_end` are skipped for that call
- The model receives `"Tool call denied: authorization hook returned False."` as the tool result

## Usage example

```python
from agents.lifecycle import RunHooks

BLOCKED_TOOLS = {"delete_file", "run_shell"}

class SafetyHooks(RunHooks):
    async def on_tool_authorize(self, context, agent, tool) -> bool:
        if tool.name in BLOCKED_TOOLS:
            print(f"Blocked tool call: {tool.name}")
            return False
        return True

agent = Agent(name="assistant", model="gpt-4o", tools=[...])
await Runner.run(agent, "do something", hooks=SafetyHooks())
```

## Tests

Added `tests/test_tool_authorize_hook.py` covering:
- Allow hook: tool runs normally, all hooks called
- Deny hook: tool not invoked, `on_tool_start`/`on_tool_end` skipped
- Denial message is returned to the model
- Agent-level deny hook also works
- Hook not called when no tool is used
- Default base class allows everything

## Summary by CodeRabbit

* **New Features**
  * Added tool authorization lifecycle hook to control and deny tool execution at run and agent levels.
  * Tool calls can now be authorized or denied before execution; denied calls are skipped with a denial message.

* **Tests**
  * Added comprehensive test coverage for the authorization hook behavior.

### DIFF
--- a/src/agents/lifecycle.py
+++ b/src/agents/lifecycle.py
@@ -76,6 +76,30 @@ class RunHooksBase(Generic[TContext, TAgent]):
         """Called immediately before a local tool is invoked."""
         pass
 
+    async def on_tool_authorize(
+        self,
+        context: RunContextWrapper[TContext],
+        agent: TAgent,
+        tool: Tool,
+    ) -> bool:
+        """Called before a tool is executed. Return False to deny the tool call.
+
+        When False is returned, the tool is not invoked and the model receives a
+        denial message instead.  All on_tool_start and on_tool_end hooks are
+        skipped for denied calls.
+
+        Default: always returns True (allow all tool calls).
+
+        Args:
+            context: The run context wrapper.
+            agent: The current agent.
+            tool: The tool about to be invoked.
+
+        Returns:
+            True to allow the tool call, False to deny it.
+        """
+        return True
+
     async def on_tool_end(
         self,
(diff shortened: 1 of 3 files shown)


seratch:
We generally recommend using tool guardrails for this purpose, so we don't plan to add a new duplicated lifecycle hook.

PR #2914, opened 2026-04-16
feat(usage): add agent_name to RequestUsage

### LINKED ISSUE #2100: Track Agent and Model Identifiers in Usage Statistics

adityasingh2400:
## Summary

Adds optional `agent_name` and `model_name` fields to `RequestUsage` so developers can attribute token usage and costs to specific agents and models in multi-agent workflows.

## Problem

In complex multi-agent systems, `result.context_wrapper.usage.request_usage_entries` contains one `RequestUsage` per LLM call, but there is no way to know which agent or model generated each entry. This makes per-agent cost attribution impossible.

## Solution

- Added `agent_name: str | None = None` and `model_name: str | None = None` to `RequestUsage` (backward-compatible — defaults to `None`)
- Modified `Usage.add()` to accept keyword-only `agent_name` and `model_name` parameters and annotate the resulting `RequestUsage` entry
- Added `_get_model_name()` helper that safely duck-types the model name from any `Model` implementation (e.g. `OpenAIResponsesModel.model`, `OpenAIChatCompletionsModel.model`)
- Updated both `run_single_turn_streamed` and `get_new_response` call-sites in `run_loop.py` to pass `agent.name` and resolved model name
- When merging pre-existing `request_usage_entries`, agent/model names are applied only to entries that don't already have them (existing names preserved)
- Updated `serialize_usage` / `deserialize_usage` to round-trip the new fields through JSON
- Zero breaking changes — existing code constructing `RequestUsage` or calling `Usage.add()` without these fields continues to work

## Before / After

```python
# Before
RequestUsage(input_tokens=65, output_tokens=13, ...)

# After
RequestUsage(agent_name="Math Tutor", model_name="gpt-4o", input_tokens=65, output_tokens=13, ...)
```

## Usage Example

```python
result = await Runner.run(triage_agent, input="help me")

for entry in result.context_wrapper.usage.request_usage_entries:
    print(f"{entry.agent_name} ({entry.model_name}): {entry.input_tokens} in / {entry.output_tokens} out")
# Triage Agent (gpt-4o-mini): 100 in / 10 out
# Specialist Agent (gpt-4o): 200 in / 20 out
```

## Files Changed

| File | Change |
|------|--------|
| `src/agents/usage.py` | Added `agent_name`/`model_name` fields to `RequestUsage`; updated `Usage.add()`, `serialize_usage`, `deserialize_usage` |
| `src/agents/run_internal/run_loop.py` | Added `_get_model_name()` helper; updated both `usage.add()` call-sites |
| `tests/test_usage.py` | Added 9 new tests |

## Test Plan
- Unit tests for `RequestUsage` field population (explicit + default None)
- `Usage.add()` propagation and backward compat tests
- Entry merge semantics (names applied to un-named entries; existing names preserved)
- Single-agent runner integration test confirming `agent_name` is set
- Model-name detection test via duck-typed `.model` attribute
- Multi-agent scenario: each `RequestUsage` entry has the correct `agent_name`
- Full suite: 2198 tests pass, lint clean, pyright clean

Closes openai/openai-agents-python#2100

## Summary by CodeRabbit

* **New Features**
  * Usage tracking now records agent name and model name with each request for improved visibility and attribution; model name is included only when available.
  * Usage exports/records now include these metadata fields when present.

* **Tests**
  * Added tests verifying capture and preservation of agent and model names across runs, merges, and multi-agent scenarios.

### DIFF
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -260,31 +260,6 @@ def _should_attach_generic_agent_error(exc: Exception) -> bool:
     )
 
 
-def _get_model_name(model: Any) -> str | None:
-    """Extract the string model name from a Model instance, if available.
-
-    Most built-in model implementations (``OpenAIResponsesModel``,
-    ``OpenAIChatCompletionsModel``) expose a ``model`` attribute that contains
-    the underlying model name string. This helper retrieves it in a
-    forward-compatible, duck-typed way so that third-party model
-    implementations are handled gracefully.
-
-    The function tries ``model`` first, then ``model_name`` for
-    implementations that prefer that name. Any exception raised while
-    resolving the attribute (for example, a custom descriptor or property
-    that throws) is swallowed so that usage accounting can never crash the
-    run, and ``None`` is returned when no string-valued name is available.
-    """
-    for attr in ("model", "model_name"):
-        try:
-            value = getattr(model, attr, None)
-        except Exception:
-            continue
-        if isinstance(value, str) and value:
-            return value
-    return None
-
-
 async def _should_persist_stream_items(
     *,
     session: Session | None,
@@ -1641,7 +1616,6 @@ async def run_single_turn_streamed(
         context_wrapper.usage.add(
             final_response.usage,
             agent_name=public_agent.name,
-            model_name=_get_model_name(model),
         )
         await asyncio.gather(
             (
@@ -1923,7 +1897,6 @@ async def get_new_response(
     context_wrapper.usage.add(
         new_response.usage,
         agent_name=public_agent.name,
-        model_name=_get_model_name(model),
     )
 
     await asyncio.gather(
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -265,13 +265,23 @@ def _get_model_name(model: Any) -> str | None:
 
     Most built-in model implementations (``OpenAIResponsesModel``,
     ``OpenAIChatCompletionsModel``) expose a ``model`` attribute that contains
-    the underlying model name string.  This helper retrieves it in a
-    forward-compatible, duck-typed way so that third-party model implementations
-    that may not have this attribute are handled gracefully.
+    the underlying model name string. This helper retrieves it in a
+    forward-compatible, duck-typed way so that third-party model
+    implementations are handled gracefully.
+
+    The function tries ``model`` first, then ``model_name`` for
+    implementations that prefer that name. Any exception raised while
+    resolving the attribute (for example, a custom descriptor or property
+    that throws) is swallowed so that usage accounting can never crash the
+    run, and ``None`` is returned when no string-valued name is available.
     """
-    model_name = getattr(model, "model", None)
-    if isinstance(model_name, str):
-        return model_name
+    for attr in ("model", "model_name"):
+        try:
+            value = getattr(model, attr, None)
+        except Exception:
+            continue
+        if isinstance(value, str) and value:
+            return value
     return None
 
 
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -260,6 +260,21 @@ def _should_attach_generic_agent_error(exc: Exception) -> bool:
     )
 
 
+def _get_model_name(model: Any) -> str | None:
+    """Extract the string model name from a Model instance, if available.
+
+    Most built-in model implementations (``OpenAIResponsesModel``,
+    ``OpenAIChatCompletionsModel``) expose a ``model`` attribute that contains
+    the underlying model name string.  This helper retrieves it in a
+    forward-compatible, duck-typed way so that third-party model implementations
+    that may not have this attribute are handled gracefully.
+    """
+    model_name = getattr(model, "model", None)
+    if isinstance(model_name, str):
+        return model_name
+    return None
+
+
 async def _should_persist_stream_items(
     *,
     session: Session | None,
@@ -1613,7 +1628,11 @@ async def run_single_turn_streamed(
                     )
 
     if final_response is not None:
-        context_wrapper.usage.add(final_response.usage)
+        context_wrapper.usage.add(
+            final_response.usage,
+            agent_name=agent.name,
+            model_name=_get_model_name(model),
+        )
         await asyncio.gather(
             (
                 public_agent.hooks.on_llm_end(context_wrapper, public_agent, final_response)
@@ -1891,7 +1910,11 @@ async def get_new_response(
         # new deltas.
         server_conversation_tracker.mark_input_as_sent(filtered.input)
 
-    context_wrapper.usage.add(new_response.usage)
+    context_wrapper.usage.add(
+        new_response.usage,
+        agent_name=agent.name,
+        model_name=_get_model_name(model),
+    )
 
     await asyncio.gather(
         (
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -1640,7 +1640,7 @@ async def run_single_turn_streamed(
     if final_response is not None:
         context_wrapper.usage.add(
             final_response.usage,
-            agent_name=agent.name,
+            agent_name=public_agent.name,
             model_name=_get_model_name(model),
         )
         await asyncio.gather(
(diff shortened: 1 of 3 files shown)


seratch:
We're not yet fully convinced to add these data because they could be incomplete in many cases. Refer to the discussion at https://github.com/openai/openai-agents-python/pull/2114 for more details.

adityasingh2400:
Thanks for the feedback @seratch!

**Bug fix:** The `elif other.request_usage_entries` branch was mutating `agent_name`/`model_name` directly on the original `RequestUsage` objects, bypassing the non-overwrite guard and causing silent mis-attribution. Fixed by creating a copy of each entry with the annotation applied rather than mutating in place. Added a regression test to cover this.

**On completeness concerns:** I've read the discussion at #2114. The fields are already guarded against the incompleteness issue:
- `agent_name` is always available (it's `agent.name`, which every agent has)
- `model_name` uses `getattr(model, 'model', None)` — if a custom `Model` implementation doesn't expose a `model` attribute, the field stays `None` and nothing breaks

Both fields are typed `str | None` and documented as 'if available', so callers can't assume they're always set. The data is as complete as it can be given the existing `Model` interface.

Happy to drop `model_name` entirely and only keep `agent_name` if that better fits the project's direction — `agent_name` has no completeness risk since it's always a string.

adityasingh2400:
@seratch quick FYI while we wait, pushed 5ac24753 with three small hardening follow-ups that are directly relevant to the "could be incomplete in many cases" concern:

- `_get_model_name` is now defensive against custom `Model` implementations whose descriptors or properties raise on attribute access. It swallows exceptions, falls back to `model_name` when `model` is missing, and rejects empty strings, so usage accounting can't crash a run on a third-party `Model` subclass that doesn't fit the pattern.
- New `test_serialize_deserialize_roundtrip_preserves_agent_and_model_names` exercises both branches of the conditional emit in `serialize_usage`: builds a `Usage` with one named and one unnamed `RequestUsage`, runs `serialize_usage` then `deserialize_usage`, asserts both fields survive on the named entry and stay `None` on the unnamed one.
- `test_multi_agent_run_attributes_usage_to_correct_agents` now sets a distinct `model` on each `FakeModel` and asserts `model_name` on both the triage and specialist `RequestUsage` entries, so a regression that drops or cross-wires `model_name` during handoff aggregation surfaces in tests rather than as a silent mis-attribution at runtime.

The completeness contract is unchanged: both fields are `str | None` and documented as best-effort, never required, so callers that don't have the data simply see `None`. The standing offer from the previous comment also still applies: if you'd prefer to keep just `agent_name` (which is always a string and carries no completeness risk), I'm happy to drop `model_name` in a follow-up commit.

adityasingh2400:
**Follow-up pushed** (`617a77c`)

Addressed both open Codex threads on `Usage.add()`:

1. **No in-place mutation** — already landed earlier via copied `RequestUsage` rows in the `request_usage_entries` merge path; left as-is.

2. **Single-request + pre-built `request_usage_entries`** — when `other.requests == 1` and `other.request_usage_entries` is non-empty, we now fold those entries in with the same agent/model annotation rules as the multi-entry branch, instead of building one new `RequestUsage` from totals + `add()` kwargs only (which dropped existing attribution).

Added `test_usage_add_single_request_preserves_prebuilt_entry_attribution`.

Threads marked resolved; CI should re-run on the push.

adityasingh2400:
Follow-up pushed (`bccca6d`) to narrow the PR to agent attribution only.

Given the concern about incomplete model metadata and the prior discussion in #2114 about not depending on a `Model.model` attribute, I removed the `model_name` surface and the best-effort model lookup entirely. The PR now only adds `agent_name`, which comes from `Agent.name` at the point the model call is made and should be consistently available without changing or depending on the `Model` interface.

Validation run locally:

- `uv run pytest tests/test_usage.py -q` -> 24 passed
- `uv run ruff check src/agents/usage.py src/agents/run_internal/run_loop.py tests/test_usage.py` -> passed

seratch:
It looks like there are some lint errors. Also, I wanted to set expectations here that we do not currently plan to make this change in the short term.

adityasingh2400:
Closing this for now because this is not an active area and it is not expected to be merged in the next few months.

PR #2915, opened 2026-04-16
feat: expose tool_call_id on RunContextWrapper for lifecycle hooks

### LINKED ISSUE #1849: Expose Tool Call ID to Lifecycle Hooks
### Please read this first

#### Problem Statement

I'm leveraging [lifecycle events](https://openai.github.io/openai-agents-python/ref/lifecycle/) to record timing metrics for my AI agent. I have a class that inherits `RunHooksBase`. When [on_agent_start](https://openai.github.io/openai-agents-python/ref/lifecycle/#agents.lifecycle.RunHooksBase.on_agent_start) and [on_tool_start](https://openai.github.io/openai-agents-python/ref/lifecycle/#agents.lifecycle.RunHooksBase.on_tool_start) are called, I begin timers. When [on_agent_end](https://openai.github.io/openai-agents-python/ref/lifecycle/#agents.lifecycle.RunHooksBase.on_agent_end) or [on_tool_end](https://openai.github.io/openai-agents-python/ref/lifecycle/#agents.lifecycle.RunHooksBase.on_tool_end) are called, I stop timers and record the metric. Unfortunately, I'm unable to differentiate between tool calls. Thus, metrics may be slightly imprecise if the agent has parallel tool calls enabled. Knowing with `*_start` corresponds with which `*_end` would allow this type of observability.

#### Feature Details

If possible, all hook methods should have access to a unique ID that identifies the tool call. I notice that elsewhere, [ToolContext](https://github.com/openai/openai-agents-python/blob/f3cac173055e180752167d0327a241240a2248a2/src/agents/tool_context.py#L22) provides a `tool_call_id`. This type of value would be ideal for lifecycle events and satisfy the problem statement above.

nileshpatil6:
## Summary

Closes #1849.

`on_tool_start` and `on_tool_end` hooks receive a `RunContextWrapper` as their context parameter. When the invocation is a function-tool call, the runtime actually passes a `ToolContext` (a subclass) which already carries `tool_call_id`. However, users who want to correlate parallel tool calls in observability/metrics code were forced to write:

```python
async def on_tool_start(self, context, agent, tool):
    if isinstance(context, ToolContext):          # painful isinstance guard
        call_id = context.tool_call_id
```

This PR adds `tool_call_id: str | None` as an `init=False` field (default `None`) directly on `RunContextWrapper`, so the pattern becomes:

```python
async def on_tool_start(self, context, agent, tool):
    call_id = context.tool_call_id  # str for function tools, None otherwise
```

## Changes

- **`src/agents/run_context.py`** — adds `tool_call_id: str | None = field(default=None, init=False)` to `RunContextWrapper`; propagates the value through `_fork_with_tool_input` and `_fork_without_tool_input`.
- **`tests/test_run_hooks.py`** — adds two tests:
  - `test_tool_call_id_exposed_on_run_context_wrapper` — verifies the correct `call_id` is surfaced in both `on_tool_start` and `on_tool_end` without any `isinstance` check.
  - `test_tool_call_id_is_none_outside_tool_context` — verifies the field is `None` on a plain `RunContextWrapper`.

## Compatibility

- `ToolContext` redeclares `tool_call_id` as a required `init=True` field (`str`, not `str | None`), so subclass behaviour and construction are unchanged.
- The new base field uses `init=False` to avoid clashing with `ToolContext.from_agent_context`, which copies base-class init fields by name.
- All existing tests pass. The two sandbox/Unix-only test files (`tests/sandbox/`, `tests/test_run_state.py`, `tests/test_sandbox_memory.py`) fail due to the missing `fcntl` module on Windows — this is pre-existing and unrelated to this change.

## Test plan

- [x] `uv run pytest tests/test_run_hooks.py tests/test_function_tool.py tests/test_tool_context.py` — 68 passed
- [x] `uv run pyright src/agents/run_context.py src/agents/tool_context.py` — 0 errors

### DIFF
--- a/src/agents/run_context.py
+++ b/src/agents/run_context.py
@@ -61,6 +61,16 @@ class RunContextWrapper(Generic[TContext]):
     tool_input: Any | None = None
     """Structured input for the current agent tool run, when available."""
 
+    tool_call_id: str | None = field(default=None, init=False)
+    """The tool call ID for the current tool invocation, when available.
+
+    Populated on :class:`ToolContext` instances that are passed to
+    :meth:`RunHooksBase.on_tool_start <agents.lifecycle.RunHooksBase.on_tool_start>` and
+    :meth:`RunHooksBase.on_tool_end <agents.lifecycle.RunHooksBase.on_tool_end>` for
+    function-tool invocations. ``None`` for non-function-tool hook invocations and for plain
+    :class:`RunContextWrapper` instances used outside of a tool call.
+    """
+
     @staticmethod
     def _to_str_or_none(value: Any) -> str | None:
         if isinstance(value, str):
@@ -461,6 +471,7 @@ class RunContextWrapper(Generic[TContext]):
         fork._approvals = self._approvals
         fork.turn_input = self.turn_input
         fork.tool_input = tool_input
+        fork.tool_call_id = self.tool_call_id
         return fork
 
     def _fork_without_tool_input(self) -> RunContextWrapper[TContext]:
@@ -469,6 +480,7 @@ class RunContextWrapper(Generic[TContext]):
         fork.usage = self.usage
         fork._approvals = self._approvals
         fork.turn_input = self.turn_input
+        fork.tool_call_id = self.tool_call_id
         return fork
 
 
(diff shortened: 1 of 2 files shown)


seratch:
We're aware of [this pain point](https://github.com/openai/openai-agents-python/blob/v0.14.1/examples/basic/lifecycle_example.py#L83-L85), but we still don't want to add the property to the parent level. We will consider changing the type hints when releasing a major version with migration guide.

PR #2937, opened 2026-04-18
fix: #2938 make `agents.sandbox.sandboxes` importable on Windows

### LINKED ISSUE #2938: Windows: import agents.sandbox.sandboxes fails with ModuleNotFoundError: fcntl
**Body**
### Describe the bug

### Debug information
- Python version: `3.14.2`
- OS: Windows (win32)

### Repro steps
```bash
```

### Expected behavior

### Actual behavior

### Proposed fix

DhanushKenkiri:
**PR Description**
- **Problem**: On Windows, `import agents.sandbox.sandboxes` fails with `ModuleNotFoundError: fcntl` because the package eagerly imports the Unix-only `unix_local` backend (which depends on Unix-only stdlib like `fcntl`/`termios`).
- **Change**:
  - Guard `unix_local` imports in src/agents/sandbox/sandboxes/__init__.py so the `sandboxes` package can be imported on Windows even when the Unix backend is unavailable.
  - Add an explicit, user-facing `ImportError` in src/agents/sandbox/sandboxes/unix_local.py when imported on Windows (clearer than the current `ModuleNotFoundError`).
  - Add regression tests in tests/sandbox/test_sandboxes_imports.py.
- **Why this approach**: `unix_local` is inherently Unix-only; the SDK should stay importable cross-platform and only fail when a platform-incompatible backend is explicitly imported/used.
- **Tests**: `python -m pytest -q tests/sandbox/test_sandboxes_imports.py`

Fixes #2938

### DIFF
--- a/src/agents/sandbox/sandboxes/__init__.py
+++ b/src/agents/sandbox/sandboxes/__init__.py
@@ -25,7 +25,7 @@ if sys.platform != "win32":
 
     _HAS_UNIX_LOCAL = True
 elif TYPE_CHECKING:  # pragma: no cover
-    from .unix_local import (
+    from .unix_local import (  # noqa: F401
         UnixLocalSandboxClient,
         UnixLocalSandboxClientOptions,
         UnixLocalSandboxSession,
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
Can you resolve the lint errors?

DhanushKenkiri:
Working on the lint error, the error is due to ruff format --check says it “would reformat” src/agents/sandbox/sandboxes/__init__.py. Next I’ll install the pinned Ruff version locally, run ruff format on the touched files, commit the formatting-only change, and push so the GitHub Action passes.

seratch:

PR #2941, opened 2026-04-18
fix(realtime): #2940 emit history_updated for transcript deltas

### LINKED ISSUE #2940: RealtimeSession does not emit `history_updated` for `transcript_delta`

### Summary

### What I observed

### Minimal repro

```python
import asyncio

class DummyModel:
    async def connect(self): pass
    async def close(self): pass

class DummyAgent:
    output_guardrails = []
    handoffs = []
    tools = []
    mcp_servers = []
    model_settings = None

async def main():

    await session.on_event(
            item_id="item_1",
            delta="hello",
            response_id="resp_1",
        )
    )

asyncio.run(main())
```

### Current behavior

### Expected behavior

The realtime guide says:

### Implementation notes

### Environment

- Agents SDK version: 0.14.2
- Python version: 3.13.5

LocNguyenSGU:
## Summary

`RealtimeSession.on_event()` already updates local history when it receives a `transcript_delta`, but it did not emit a matching high-level history event.

This patch emits `RealtimeHistoryUpdated` after the transcript-delta history mutation so consumers that follow `history_added` / `history_updated` stay in sync with live transcript updates.

## Reproduction

A single `RealtimeModelTranscriptDeltaEvent` produced:

- a mutated `_history` entry containing the partial assistant transcript
- only a raw model event in the queue
- no `RealtimeHistoryUpdated`

## Changes

- emit `RealtimeHistoryUpdated` in the `transcript_delta` branch after updating `_history`
- add a focused regression test for the high-level event contract
- narrow the existing ignored-events test so it only covers truly ignored events

## Verification

```bash
uv run pytest tests/realtime/test_session.py -k "transcript_delta_updates_history_and_emits_history_updated or ignored_events_only_generate_raw_events or transcript_delta_triggers_guardrail_at_threshold or transcript_delta_multiple_thresholds_same_item or transcript_delta_different_items_tracked_separately" -q
```

Result:

- `5 passed`

## Risks

Low. The change only adds the missing high-level event emission for an already-applied history mutation.

Closes #2940


src/agents/realtime/session.py
```diff
@@ -313,6 +313,9 @@ async def on_event(self, event: RealtimeModelEvent) -> None:
                     content=[AssistantAudio(transcript=self._item_transcripts[item_id])],
                 ),
             )
+            await self._put_event(
```
afurm, line 316: Line 318: `history=self._history` passes a direct reference to the session's internal `_history` list rather than a copy. If any subscriber mutates this list, it silently corrupts `RealtimeSession._history` without the session knowing.

Consider: `history=list(self._history)` to prevent external mutation from affecting internal state.

LocNguyenSGU: Fixed in `7f3eff8`. `RealtimeHistoryUpdated` now receives `list(self._history)`, so consumers can no longer mutate the session's internal history list through the event payload.

I also made the snapshot behavior consistent across all full-history `RealtimeHistoryUpdated` emissions and added a regression test for the `transcript_delta` path.

LocNguyenSGU:
Fixed the lint formatting issue in `tests/realtime/test_session.py`.

Verified locally:
- `python3 -m ruff format --check`
- `python3 -m ruff check`

Commit: `7d2c6a83`

PR #2943, opened 2026-04-18
fix: #2938 guard unix_local sandbox import behind platform check on Windows

### LINKED ISSUE #2938: Windows: import agents.sandbox.sandboxes fails with ModuleNotFoundError: fcntl

nileshpatil6:
## Summary

Closes #2938.

`agents.sandbox.sandboxes` eagerly imported `unix_local`, which pulls in Unix-only stdlib modules (`fcntl`, `termios`). On Windows this caused an immediate `ModuleNotFoundError` at import time — even when the Unix backend is never used.

**Before:**
```python
# Always runs — crashes on Windows
from .unix_local import UnixLocalSandboxClient, ...
```

**After:**
```python
import sys
_HAS_UNIX_LOCAL = sys.platform != "win32"

if _HAS_UNIX_LOCAL:
    from .unix_local import UnixLocalSandboxClient, ...  # only on Unix
```

This mirrors the existing `try/except` pattern already used for the optional Docker backend. `__all__` is rebuilt conditionally so only the backends available on the current platform are advertised.

## Verification

Tested directly on Windows (win32):
```
python -c "import agents.sandbox.sandboxes; print('OK')"
# Output: Import OK on Windows
```

## Test plan

- [x] `python -c "import agents.sandbox.sandboxes"` — succeeds on Windows (was crashing before)
- [x] `uv run pyright src/agents/sandbox/sandboxes/__init__.py` — 0 errors
- [x] `uv run pytest tests/` (excluding pre-existing Windows-incompatible tests) — 2249 passed

### DIFF
--- a/src/agents/sandbox/sandboxes/__init__.py
+++ b/src/agents/sandbox/sandboxes/__init__.py
@@ -5,12 +5,17 @@ This subpackage contains concrete session/client implementations for different
 execution environments (e.g. Docker, local Unix).
 """
 
-from .unix_local import (
-    UnixLocalSandboxClient,
-    UnixLocalSandboxClientOptions,
-    UnixLocalSandboxSession,
-    UnixLocalSandboxSessionState,
-)
+import sys
+
+_HAS_UNIX_LOCAL = sys.platform != "win32"
+
+if _HAS_UNIX_LOCAL:
+    from .unix_local import (  # noqa: F401
+        UnixLocalSandboxClient,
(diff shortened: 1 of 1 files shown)


seratch:

PR #2946, opened 2026-04-18
fix: #2938 avoid importing unix local sandboxes on Windows

### LINKED ISSUE #2938: Windows: import agents.sandbox.sandboxes fails with ModuleNotFoundError: fcntl

akshitnanda:
## Summary
- avoid importing the Unix local sandbox backend on Windows when `agents.sandbox.sandboxes` is imported
- keep Unix local exports available on non-Windows platforms and leave Docker imports optional
- add a regression test that reloads the sandboxes package under a simulated `win32` platform

## Testing
- `py -3.14 -m py_compile src/agents/sandbox/sandboxes/__init__.py tests/sandbox/test_sandboxes_import.py`
- Unable to run `pytest` locally in this environment because the available Python does not have the repo's dev dependencies installed (`pytest` was missing, and `uv` was not available on PATH)

Closes #2938.

### DIFF
--- a/src/agents/sandbox/sandboxes/__init__.py
+++ b/src/agents/sandbox/sandboxes/__init__.py
@@ -5,12 +5,20 @@ This subpackage contains concrete session/client implementations for different
 execution environments (e.g. Docker, local Unix).
 """
 
-from .unix_local import (
-    UnixLocalSandboxClient,
-    UnixLocalSandboxClientOptions,
-    UnixLocalSandboxSession,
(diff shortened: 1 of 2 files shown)


seratch:

PR #2952, opened 2026-04-19
fix: preserve original name for warning in transform_string_function_style

### LINKED ISSUE #2951: fix: transform_string_function_style silently transforms space-containing names without warning
### Description

In `src/agents/util/_transforms.py`, the function `transform_string_function_style` has a logic bug: it replaces spaces with underscores **before** comparing to decide whether to emit a warning. This means tool names containing spaces are silently lowercased and space-replaced without triggering the warning.

### Current behavior

```python
def transform_string_function_style(name: str) -> str:
    name = name.replace(" ", "_")   # ← modifies `name` in-place
    transformed_name = re.sub(r"[^a-zA-Z0-9_]", "_", name)

    if transformed_name != name:    # ← compared against already-modified name
        logger.warning(...)         # ← never fires for space-only transformations

    return transformed_name.lower()
```

For a tool named `"My Tool"`:
1. `name` becomes `"My_Tool"` 
2. `transformed_name` becomes `"My_Tool"` (no further changes)
3. `transformed_name == name` → **no warning**
4. Returns `"my_tool"` — silently different from the original input `"My Tool"`

### Expected behavior

The warning should fire whenever the **final returned value** differs from the **original input**. Names with spaces should trigger the same warning as names with other invalid characters.

### Fix

Preserve the original name before any modifications:

```python
def transform_string_function_style(name: str) -> str:
    original_name = name
    name = name.replace(" ", "_")
    transformed_name = re.sub(r"[^a-zA-Z0-9_]", "_", name)
    final_name = transformed_name.lower()

    if final_name != original_name:
        logger.warning(
            f"Tool name {original_name!r} contains invalid characters for function calling and has been "
            f"transformed to {final_name!r}. Please use only letters, digits, and underscores "
            "to avoid potential naming conflicts."
        )

    return final_name
```

### Impact

Users defining tools with names containing spaces will silently get renamed without any log warning, making it harder to debug mismatched tool names.

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

kuishou68:
Closes #2951

## Problem
`transform_string_function_style` modified `name` (replacing spaces with `_`) before comparing it to `transformed_name`. This meant tool names containing **only spaces** (as invalid chars) never triggered the warning, even though the returned value silently differed from the original input.

## Fix
Capture `original_name = name` before any mutation, and compare the final lowercased result against the original to decide whether to warn.

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>

### DIFF
--- a/src/agents/util/_transforms.py
+++ b/src/agents/util/_transforms.py
@@ -4,18 +4,21 @@ from ..logger import logger
 
 
 def transform_string_function_style(name: str) -> str:
+    # Preserve the original name for comparison and warning message
+    original_name = name
+
     # Replace spaces with underscores
     name = name.replace(" ", "_")
 
     # Replace non-alphanumeric characters with underscores
     transformed_name = re.sub(r"[^a-zA-Z0-9_]", "_", name)
+    final_name = transformed_name.lower()
 
-    if transformed_name != name:
-        final_name = transformed_name.lower()
+    if final_name != original_name:
         logger.warning(
-            f"Tool name {name!r} contains invalid characters for function calling and has been "
+            f"Tool name {original_name!r} contains invalid characters for function calling and has been "
             f"transformed to {final_name!r}. Please use only letters, digits, and underscores "
             "to avoid potential naming conflicts."
         )
 
-    return transformed_name.lower()
+    return final_name


seratch:
Thanks for the report and the patch. The issue is valid: names with spaces are currently transformed before the warning comparison, so something like `My Tool` becomes `my_tool` without warning.

I’m going to handle the fix on our side rather than asking you to keep iterating on this PR. The main reason is that the safest change is a little narrower than the current patch: we should warn when the name needed character replacement, but we should not start warning for case-only normalization.

For example, this PR would also warn for:

```text
MyTool -> mytool
transfer_to_Agent -> transfer_to_agent
```

Those names only contain letters, digits, and underscores, so the existing “invalid characters” warning would be misleading. It could also add warning noise for common agent and handoff names where uppercase letters are normal.

#2953 keeps the existing returned values, adds the missing warning for spaces/punctuation, and adds regression coverage for both replacement and case-only cases. I’ll close this PR once that lands. Thanks again for surfacing the bug.

PR #2954, opened 2026-04-19
fix: handle duplicate tool names across MCP servers by auto-renaming

### LINKED ISSUE #464: Duplicate tool names across MCP servers cause errors
### Describe the bug

### Debug information

### Repro steps

```python
import json
import asyncio

async def main():

            agent = Agent(
                name="Assistant",
                model="gpt-4o-mini"
            )

if __name__ == "__main__":
    asyncio.run(main())
```

### Expected behavior

### Crosslinks

  ##

### LINKED ISSUE #1167: Duplicate tool names in multiple MCP tools cause the agent list to hang

houziershi:
When multiple MCP servers have tools with the same name, instead of raising a UserError, the SDK now automatically renames duplicate tools with a server prefix (e.g., 'server_name__tool_name'). This allows users to use multiple MCP servers even when they have overlapping tool names.

Changes:
- Modified MCPUtil.get_all_function_tools() to detect duplicates and rename them instead of raising an error
- Added MCPUtil._rename_tool() helper to properly copy and rename FunctionTool instances while preserving all internal metadata (_tool_origin, _mcp_title, etc.)
- Added warning logs to inform users about the renaming
- Updated existing tests to reflect the new behavior
- Added new tests for duplicate name handling scenarios

Fixes #1167
Fixes #464

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -37,6 +37,7 @@ from ..tool import (
 )
 from ..tool_context import ToolContext
 from ..tracing import FunctionSpanData, get_current_span, mcp_tools_span
+from ..util._transforms import transform_string_function_style
 from ..util._types import MaybeAwaitable
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
I don't think this PR is sufficient. Please take a look at my comment at https://github.com/openai/openai-agents-python/pull/2788#pullrequestreview-4017974619 for more details.

Also, while fixing this issue may look relatively simple, I don't think it actually is. We'd like to review it carefully to prevent any breaking changes. There may be some other things to consider as well.

houziershi:
> I don't think this PR is sufficient. Please take a look at my comment at [#2788 (review)](https://github.com/openai/openai-agents-python/pull/2788#pullrequestreview-4017974619) for more details.
> 
> Also, while fixing this issue may look relatively simple, I don't think it actually is. We'd like to review it carefully to prevent any breaking changes. There may be some other things to consider as well.

OK

PR #2957, opened 2026-04-19
feat: add tool_not_found error handler to recover from hallucinated tool calls (#325)

### LINKED ISSUE #325: Retry mechanism for ModelBehaviorError
I have extensively use this framework during the recent week. It performs pretty well, exception on rare circumstances, the LLM will attempt to call a nonexistent tool, which crashed a whole 10-min agent run. Can you implement a retry mechanism allowing re-execute the errored LLM call to have a chance to recover from the it?

sjhddh:
## Summary

Fixes #325. When the model calls a tool that isn't registered on the agent, the SDK currently raises `ModelBehaviorError` and kills the run, discarding every turn that came before. Users on the 2+-year-old issue report losing multi-minute DeepSearch-style runs to a single typo. This PR adds a recoverable escape hatch.

## What's new

Register a handler alongside the existing `max_turns` one:

```python
def on_tool_not_found(data: ToolNotFoundErrorHandlerInput[None]) -> ToolNotFoundAction:
    return ToolNotFoundAction(
        error_message=f"Tool {data.tool_name!r} does not exist. Available: {data.available_tools}."
    )

Runner.run_sync(agent, "...", error_handlers={"tool_not_found": on_tool_not_found})
```

The runner pre-scans the model response for unknown tool calls, invokes the handler (sync or async) once per unknown call, synthesizes a `function_call_output` with the handler's message, and continues the turn. The model sees the error on its next step and self-corrects.

## Contract

- Return `ToolNotFoundAction(error_message=...)` to recover; return `None` or register nothing to preserve today's raise behavior.
- Recovery is bounded by `max_turns`, so repeated hallucinations still terminate.
- LiteLLM's `json_tool_call` structured-output pseudo-call is not treated as missing.
- The handler's own exceptions propagate (buggy handlers must surface, not be swallowed).
- Span errors are attached only when we actually raise; successful recoveries don't pollute traces.

`ToolNotFoundAction` is a dataclass with one field today (`error_message`) by design — a wrapper lets us add fields without breaking callers.

## Docs & example

- New subsection in `docs/running_agents.md`.
- New runnable example at `examples/basic/tool_not_found_handler.py` (offline, uses a scripted model).

## Test plan
- [x] 11 tests in `tests/test_tool_not_found_handler.py` (including handler-raises, multi-call batch, `to_input_list` round-trip, LiteLLM escape hatch)
- [x] `make format-check`, `make lint`, `make mypy` clean
- [x] Full suite: 3835 passed
- [x] Example runs end-to-end

### DIFF
--- a/docs/running_agents.md
+++ b/docs/running_agents.md
@@ -441,6 +441,31 @@ print(result.final_output)
 
 Set `include_in_history=False` when you do not want the fallback output appended to conversation history.
 
+### Recovering from hallucinated tool calls
+
+Models occasionally call a tool name that was never registered on the agent (issue [#325](https://github.com/openai/openai-agents-python/issues/325)). By default the SDK raises `ModelBehaviorError` and the run ends, discarding prior work. Register a `"tool_not_found"` handler to turn that crash into a recoverable nudge: the handler returns a [`ToolNotFoundAction`][agents.ToolNotFoundAction] with a model-visible error message, the runner injects it as a synthetic tool output, and the model self-corrects on the next turn. Returning `None` (or not registering a handler) preserves the existing raise behavior. Recovery is bounded by the run's `max_turns`, so a model that keeps hallucinating still terminates.
+
+```python
+from agents import Agent, Runner, ToolNotFoundAction, ToolNotFoundErrorHandlerInput
+
+
+def on_tool_not_found(data: ToolNotFoundErrorHandlerInput[None]) -> ToolNotFoundAction:
+    return ToolNotFoundAction(
+        error_message=(
+            f"Tool {data.tool_name!r} does not exist. Available: {data.available_tools}."
+        )
+    )
+
+
+result = Runner.run_sync(
+    agent,
+    "find me profiles related to Anthropic",
+    error_handlers={"tool_not_found": on_tool_not_found},
+)
+```
+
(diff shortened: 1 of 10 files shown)


seratch:
Thanks for sharing this. Providing some solutions for handling ModelBehaviorError is worth considering, but this tool-not-found pattern is not the thing we'd like to add to this SDK.

sjhddh:
Thanks for the quick look, @seratch — I appreciate the direction.

Happy to rework if you can share a bit more on what shape you'd find acceptable. A few options I'm weighing:

1. **Generic `model_behavior_error` handler** (covers tool-not-found, invalid JSON, missing call_id, etc.) via the existing `RunErrorHandlers` TypedDict — a single kind with the full `ModelBehaviorError` instance as input. Keeps the user-facing surface to one key.
2. **Extending `ModelSettings.retry`** to retry `ModelBehaviorError` as a transient class, leaving recovery shape to the existing retry machinery.
3. **Something else entirely** — e.g., an internal "inform the model and continue" default, with no new public surface at all.

If #2 or #3 is closer to what you have in mind, happy to close this and open a scoped-down version. If #1 is workable, I can pivot the existing diff. Let me know what pattern fits the SDK's direction.

sjhddh:
Thanks for the quick feedback @seratch — understood, closing this one. If it's useful I'm happy to explore a narrower proposal focused on ModelBehaviorError observability/handling hooks rather than the tool-not-found auto-recovery pattern; happy to open a discussion first so we can align on scope before any code.

PR #2964, opened 2026-04-20
feat: add BoxLite as a sandbox provider

DorianZheng:
## Summary
- Adds `BoxliteSandboxClient` / `BoxliteSandboxClientOptions` / `BoxliteSandboxSession` / `BoxliteSandboxSessionState` under `src/agents/extensions/sandbox/boxlite/`, backed by `boxlite.SimpleBox` (local-first micro-VM sandbox with hardware isolation — https://github.com/boxlite-ai/boxlite).
- Wires the new backend into `src/agents/extensions/sandbox/__init__.py` with the standard optional-import guard (`_HAS_BOXLITE`) and conditional `__all__` extension, matching the pattern used by the other cloud backends.
- Declares the optional extra `boxlite = ["boxlite>=0.8.2"]` in `pyproject.toml` plus a mypy `ignore_missing_imports` override.
- Implements the six required `BaseSandboxSession` overrides: `_exec_internal` via `box.exec`; `read`/`write` via base64-over-exec; `persist_workspace`/`hydrate_workspace` via tar-over-exec; `running` via `box.info().state`.

## Test plan
- [x] `make format`
- [x] `make lint`
- [x] `uv run mypy src/agents/extensions/sandbox/boxlite tests/extensions/test_sandbox_boxlite.py` — clean
- [x] `uv run pytest tests/extensions/test_sandbox_boxlite.py` — 7/7 passing (options serialization, state round-trip via registry, create validation, exec translation, running probe, shutdown, base64 read)
- [ ] Manual smoke test on a KVM/HVF host with `pip install boxlite` and a real `python:slim` box

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ e2b = ["e2b==2.20.0", "e2b-code-interpreter==2.4.1"]
 modal = ["modal==1.3.5"]
 runloop = ["runloop_api_client>=1.16.0,<2.0.0"]
 vercel = ["vercel>=0.5.6,<0.6"]
+boxlite = ["boxlite>=0.8.2"]
 s3 = ["boto3>=1.34"]
 temporal = [
     "temporalio==1.26.0",
@@ -162,6 +163,10 @@ ignore_missing_imports = true
 
 [[tool.mypy.overrides]]
 module = ["vercel", "vercel.*"]
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = ["boxlite", "boxlite.*"]
 ignore_missing_imports = true
 
 [tool.coverage.run]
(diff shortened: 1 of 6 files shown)


DorianZheng:
@seratch Could you please take a look?

seratch:
Hi @DorianZheng, thanks for sending this! However, we currently don't plan to add sandbox provider implementations to the extensions module. If you don't mind, including the module for your service as part of your own package would be appreciated.

DorianZheng:
Thanks @seratch — quick pitch before closing:

BoxLite is a Rust micro-VM runtime that embeds in-process (no daemon, no cloud), with ms-range cold start and real VM-level isolation. For agents SDK users it fills the gap between Docker (heavy, daemon-bound) and cloud sandboxes (latency, cost, lock-in) — same code path in dev and prod, safe for LLM-generated code, fully offline-capable.

Happy to ship it as `pip install ...[boxlite]` so zero impact on core deps, and I'll own ongoing maintenance. If there's a specific concern I can address, I'd rather iterate than split out.

PR #2967, opened 2026-04-20
fix(tracing): serialize big integers as strings to prevent dashboard precision loss

### LINKED ISSUE #2094: Traces dashboard shows big integer arguments incorrectly in tool calling traces

xodn348:
## Summary

Fixes #2094

JavaScript's `JSON.parse` silently loses precision on integers larger than `Number.MAX_SAFE_INTEGER` (2^53 - 1). When the OpenAI dashboard receives a trace span containing a Python `int` that exceeds this limit (e.g. a 64-bit trace ID or token count), the value is silently rounded — making traces unreliable for debugging.

## Changes

- In `src/agents/tracing/span_data.py`, added a `_safe_serialize` helper that converts any `int` value exceeding `2^53 - 1` to its `str` representation before JSON serialization.
- Applied this helper to all numeric fields in span data classes (`AgentSpanData`, `FunctionSpanData`, `GenerationSpanData`, etc.) via `__post_init__`.
- Integers within safe range are left as-is to avoid unnecessary type changes.

## Testing

- Existing tests pass without modification.
- The helper correctly preserves small ints and converts large ints to strings.

### DIFF
--- a/src/agents/tracing/span_data.py
+++ b/src/agents/tracing/span_data.py
@@ -1,11 +1,34 @@
 from __future__ import annotations
 
 import abc
+import json
 from collections.abc import Mapping, Sequence
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     from openai.types.responses import Response, ResponseInputItemParam
+
+# JavaScript's Number.MAX_SAFE_INTEGER (2^53 - 1). Integers beyond this threshold
+# lose precision when parsed by JS, causing the Traces dashboard to display wrong values.
+_JS_MAX_SAFE_INTEGER = 9007199254740991
+
+
+def _sanitize_bigint(value: Any) -> Any:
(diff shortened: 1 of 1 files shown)


seratch:
Hi @xodn348, as I mentioned at https://github.com/openai/openai-agents-python/pull/2959#issuecomment-4277240634, we won't have this change on the client SDK side.

PR #2976, opened 2026-04-20
fix: ignore relative snapshot base overrides

matthewflint:
## Summary
- ignore relative `XDG_STATE_HOME`, `LOCALAPPDATA`, and `APPDATA` overrides
- keep the default local snapshot base dir independent from the current working directory
- add regression coverage for the relative-path cases

## Testing
- `uv run pytest -s tests/sandbox/test_snapshot_defaults.py`
- `uv run ruff check src/agents/sandbox/snapshot_defaults.py tests/sandbox/test_snapshot_defaults.py`

Signed-off-by: matthewflint 277024436+matthewflint@users.noreply.github.com

### DIFF
--- a/src/agents/sandbox/snapshot_defaults.py
+++ b/src/agents/sandbox/snapshot_defaults.py
@@ -4,7 +4,7 @@ import os
 import sys
 import time
 from collections.abc import Mapping
-from pathlib import Path
+from pathlib import Path, PureWindowsPath
 
 from .snapshot import LocalSnapshotSpec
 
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
One small concern before merging: the direction looks right, but the Windows-specific coverage is still a bit indirect.

The new tests exercise the `os_name="nt"` branch, but they use POSIX-style absolute paths from `tmp_path`. That means we are not actually proving that a Windows-style env value such as `C:\Users\me\AppData\Local` is treated as absolute when `platform/os_name` are injected for a simulated Windows case.

If `platform` / `os_name` injection is intended to support cross-platform unit tests, we may want either:
- a test that covers Windows-style absolute env paths explicitly, or
- branch-specific path handling using `PureWindowsPath` for the Windows env vars.

Not blocking from my side, but I think this would make the regression coverage more precise.

matthewflint:
Yes, that makes sense. I updated the `os_name == "nt"` branch to use `PureWindowsPath(...).is_absolute()` for `LOCALAPPDATA` / `APPDATA`, and switched the tests to explicit Windows-style values like `C:\Users\me\AppData\Local`.

I also added coverage for `APPDATA` fallback, the all-relative fallback, and a guard that `/tmp/...` is ignored under simulated Windows.

seratch (approved):

PR #2977, opened 2026-04-20
fix: #2940 emit RealtimeHistoryUpdated when transcript_delta updates history

### LINKED ISSUE #2940: RealtimeSession does not emit `history_updated` for `transcript_delta`

nileshpatil6:
## Problem

`RealtimeSession` accumulates transcript text and updates `self._history` on every `transcript_delta` event, but never emits `RealtimeHistoryUpdated`. UI consumers that follow the documented pattern of listening to `history_added` / `history_updated` events miss all live partial transcript changes.

Every other branch that mutates history (`item_updated`, `item_deleted`) correctly emits `RealtimeHistoryUpdated` — `transcript_delta` is the only one that doesn't.

## Fix

Emit `RealtimeHistoryUpdated` immediately after `_get_new_history` in the `transcript_delta` branch, consistent with the other branches.

Fixes #2940

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -313,6 +313,9 @@ class RealtimeSession(RealtimeModelListener):
                     content=[AssistantAudio(transcript=self._item_transcripts[item_id])],
                 ),
             )
+            await self._put_event(
+                RealtimeHistoryUpdated(info=self._event_info, history=self._history)
+            )
 
             # Check if we should run guardrails based on debounce threshold
             current_length = len(self._item_transcripts[item_id])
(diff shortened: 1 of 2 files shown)


seratch:
duplicate with #2941 and #3086; also we are lean to #2941 changes so far

PR #2987, opened 2026-04-21
refactor: share sandbox tar exclude arg generation

seratch:
This pull request improves sandbox workspace persistence internals by sharing the shell `tar --exclude` argument generation used by Blaxel, Daytona, and E2B sandbox sessions.

It adds a small common helper under `src/agents/sandbox/session/` and keeps each provider's existing private `_tar_exclude_args()` surface as a thin delegate, preserving the generated command strings while removing duplicated quoting, sorting, and path-normalization logic. Direct unit coverage was added for empty/dot paths, stable sorting, shell quoting, dot-prefixed patterns, and absolute-path normalization.

### DIFF
--- a/src/agents/sandbox/session/tar_workspace.py
+++ b/src/agents/sandbox/session/tar_workspace.py
@@ -1,0 +1,18 @@
+from __future__ import annotations
+
+import shlex
+from collections.abc import Iterable
+from pathlib import Path
+
+__all__ = ["shell_tar_exclude_args"]
+
+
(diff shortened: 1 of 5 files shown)


src/agents/sandbox/session/tar_workspace.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
```
qiyaoq-oai, line 1: nit: we prob should have a utils lib to keep these

qiyaoq-oai (approved):

alfozan (approved):

PR #2995, opened 2026-04-21
feat(retry): add retry_policies.rate_limit() convenience policy

### LINKED ISSUE #782: Rate Limit Support
Is there a way to add rate limits at run time? 

If the rate limit has been exceeded then there should be queue or perhaps a retry mechanism in place.

RhythrosaLabs:
## Summary

Fixes #782.

Adds `retry_policies.rate_limit()` — a ready-made helper that retries on HTTP 429 responses only, honouring `Retry-After` / `Retry-After-Ms` headers when present and falling back to the configured backoff schedule when absent.

---

## Motivation

Issue #782 reports that there is no concise way to express "retry rate-limit errors, wait the prescribed cool-down". The existing primitives require users to manually compose:

```python
retry_policies.any(
    retry_policies.http_status([429]),
    retry_policies.retry_after(),
)
```

This is both verbose and subtly wrong:
- `http_status([429])` fires on 429 but ignores the `Retry-After` header entirely.
- `retry_after()` honours the header but triggers on **any** error that carries one, not just 429s.

Neither alone captures the intended contract. Users need domain knowledge of the internals to compose them correctly.

---

## What this PR adds

```python
# After
from agents import ModelRetrySettings, ModelSettings, retry_policies

ModelSettings(
    retry=ModelRetrySettings(
        max_retries=4,
        policy=retry_policies.rate_limit(),
    )
)
```

The policy:
1. Returns `RetryDecision(retry=False)` for anything other than HTTP 429.
2. Returns `RetryDecision(retry=True, delay=<seconds>)` when a `Retry-After` header was parsed (via `context.normalized.retry_after`) or surfaced by provider advice — so the runner waits the prescribed cool-down.
3. Returns `RetryDecision(retry=True, delay=None)` when no header is present — the runner falls back to the configured backoff schedule.

Can be combined with `provider_suggested()` for deeper provider integration:
```python
retry_policies.any(retry_policies.rate_limit(), retry_policies.provider_suggested())
```

---

## Changes

| File | Change |
| --- | --- |
| `src/agents/retry.py` | New `rate_limit()` method on `_RetryPolicies` |
| `tests/test_model_retry.py` | 8 new tests (6 unit-level policy tests + 2 integration tests via `get_response_with_retry`) |
| `examples/basic/rate_limit_retry.py` | New minimal example |
| `docs/models/index.md` | New row in the ready-made helpers table |

---

## Tests

```
pytest tests/test_model_retry.py -q
60 passed in 0.36s
```

New tests cover:
- Retries on 429 with no header (`delay=None`)
- Uses `normalized.retry_after` delay when header is present
- Falls back to `provider_advice.retry_after` when normalized value is absent
- `normalized.retry_after` takes precedence over `provider_advice.retry_after`
- Does **not** retry on non-429 status codes (e.g. 500)
- Does **not** retry on network errors
- Integration: full `get_response_with_retry` run with 429+`Retry-After: 2` header → `sleep(2.0)` called
- Integration: full run with 429, no header → falls back to configured backoff delay

---

## Checklist

- [x] Implementation follows existing code style and uses `_mark_retry_capabilities`
- [x] All existing tests still pass
- [x] New tests added for all behaviour branches
- [x] Docs table updated
- [x] Example added

### DIFF
--- a/docs/models/index.md
+++ b/docs/models/index.md
@@ -418,6 +418,7 @@ The SDK exports ready-made helpers on `retry_policies`:
 | `retry_policies.network_error()` | Matches transient transport and timeout failures. |
 | `retry_policies.http_status([...])` | Matches selected HTTP status codes. |
 | `retry_policies.retry_after()` | Retries only when a retry-after hint is available, using that delay. |
+| `retry_policies.rate_limit()` | Retries on HTTP 429, honoring `Retry-After` headers; falls back to the configured backoff when no header is present. |
 | `retry_policies.any(...)` | Retries when any nested policy opts in. |
 | `retry_policies.all(...)` | Retries only when every nested policy opts in. |
 
--- a/examples/basic/rate_limit_retry.py
+++ b/examples/basic/rate_limit_retry.py
@@ -1,0 +1,42 @@
+"""Example: retrying on HTTP 429 rate-limit errors.
+
+``retry_policies.rate_limit()`` is a ready-made helper that retries on HTTP 429
+responses only.  When the server includes a ``Retry-After`` or
+``Retry-After-Ms`` header the runner waits exactly that long before the next
+attempt.  If no header is present the runner falls back to the configured
+backoff schedule (exponential by default).
+
+For deeper provider integration you can combine it with
+``retry_policies.provider_suggested()`` so the runner also honours explicit
+provider retry advice and safety approvals.
+"""
+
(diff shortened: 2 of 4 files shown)


seratch:
Thanks for sharing this. The built-in retry handlers already include `retry_after()`, so this looks like duplicate while I see some minor differences in code.

PR #3002, opened 2026-04-22
docs: add README banner artwork

huangrichao2020:
## Summary
- add a new README banner image and place it at the top of the README
- store the artwork in `docs/readme-banner.png` so the README stays self-contained

## Why
This is a very beautiful banner image that gives the project a much stronger first impression. It helps improve the README's visual polish and makes the project more compelling for sharing, discovery, and overall promotion.

## Testing
- verified the branch README now starts with the new banner image block
- verified the new banner asset returns `200 image/png`


seratch:
I like this, but we won't update README this way.

PR #3009, opened 2026-04-23
feat(model_settings): add cache_system_prompt for explicit prompt caching

### LINKED ISSUE #3008: cache_control for LiteLLM-routed providers
### Please read this first

- **Have you read the docs?** Yes — [Agents SDK docs](https://openai.github.io/openai-agents-python/).
- **Have you searched for related issues?** Yes. Prior closed issues #905 ("Support Anthropic Prompt Caching", closed in July 2025 by the reporter with a custom-model workaround only — not a native SDK solution) and #1257 ("Anthropic caching not supported on LiteLLM", closed with the `wontfix` label) both describe the same unmet need. PR #2095 added `ModelSettings.prompt_cache_retention` for the OpenAI Responses API path; this request is the Chat Completions / LiteLLM analog.

### Describe the feature

Add a first-class `ModelSettings.cache_system_prompt: bool | None` flag. When set to `True`, the leading `role=system` message is emitted as a structured text content part tagged with `cache_control: {"type": "ephemeral"}` — the exact shape LiteLLM documents for Anthropic native, Bedrock, OpenRouter → Claude/Gemini, and Vertex AI. Those providers honor the marker as a prompt-caching breakpoint; subsequent calls with an identical prefix skip re-tokenization and bill at the cached rate.

**Why this is needed**

Long, stable system prompts are the common shape in Agents SDK apps, and for LiteLLM-routed providers the missing `cache_control` marker is the single biggest lever on input-token cost.

Today there is no SDK-side way to emit it:

1. Two sites in `src/agents/models/chatcmpl_converter.py` (`_normalize_input_content_part_alias` and `extract_all_content`) reconstruct text parts with only `type` and `text`, silently dropping any extra key. So even hand-written `call_model_input_filter` hooks that attach `cache_control` fail — the marker is stripped before reaching LiteLLM.
2. There is no ergonomic opt-in on `ModelSettings`. Per the closing comment on #905, the only existing workaround is to fork `LitellmModel` into a full custom-model implementation — heavyweight, brittle, and every consumer reinvents the same wrapper.

**Empirical impact**

Measured against a ~36k-char production system prompt via OpenRouter → Gemini 2.5 Pro with a patch applied: turn 2 reports **9,499 cached_tokens out of 9,544 prompt_tokens** with the flag on, and **0** with it off.

**Proposed API**

```python
Agent(
    instructions="<long system prompt>",
    model_settings=ModelSettings(cache_system_prompt=True),
)
```

The field sits alongside the existing `prompt_cache_retention` so the two caching paths have symmetric surface area. Default is `None`: when unset, all three Chat Completions-style paths (`OpenAIChatCompletionsModel`, `LitellmModel`, `AnyLlmModel`) produce byte-identical payloads to today — no behavior change for existing callers.

Jash271:
### Summary

This pull request adds a declarative `ModelSettings.cache_system_prompt` flag that marks an agent's system prompt as a prompt-caching breakpoint, and extends the Converter to carry the Anthropic-style `cache_control` field through on text content parts so providers that honor it actually receive the marker.

Closes #3008.

#### Background

Providers routed through LiteLLM — Anthropic native, Bedrock, OpenRouter → Claude/Gemini, Vertex AI — honor an Anthropic-style `cache_control: {"type": "ephemeral"}` field on individual text content blocks to mark an explicit prompt-caching breakpoint. LiteLLM forwards the field; the provider treats the marked prefix as a cached block; subsequent calls with an identical prefix skip re-tokenization and bill at the cached rate. For Agents SDK apps with long, stable system prompts — the common shape — this is the single biggest lever on input-token cost.

Empirically (measured against a ~36k-char production system prompt via OpenRouter → Gemini 2.5 Pro with this patch applied): turn 2 reports **9,499 cached_tokens out of 9,544 prompt_tokens** with the flag ON, and **0** with it OFF.

Today the SDK does not make this reachable:

1. Two sites in `src/agents/models/chatcmpl_converter.py` (`_normalize_input_content_part_alias` and `extract_all_content`) reconstruct text parts with only `type` and `text`, so even if a caller hand-writes a `call_model_input_filter` that attaches `cache_control` to the content, the marker does not survive to LiteLLM. Updating only one site is not enough; the second rebuilds the part and drops the field on the way out.
2. There is no ergonomic opt-in on `ModelSettings`. Absent a flag, every consumer that wants caching has to write the same `call_model_input_filter` boilerplate that restructures the system message from a plain string into a structured content list.

#### Changes

- `src/agents/model_settings.py`: new optional field `cache_system_prompt: bool | None = None`, appended at the end of the dataclass to preserve positional-argument compatibility per `AGENTS.md`'s Public API rule.
- `src/agents/models/chatcmpl_helpers.py`: new `ChatCmplHelpers.build_system_message()` helper that centralizes how the leading `role=system` message is constructed. When `cache_system_prompt=True`, the content is emitted as a structured text content part with `cache_control: {"type": "ephemeral"}`; otherwise it's the plain string the SDK has always produced. No `call_model_input_filter` customization required on the caller's side.
- `src/agents/models/chatcmpl_converter.py`: `_normalize_input_content_part_alias` and `extract_all_content` now preserve `cache_control` when present on text parts. The SDK does not interpret the field — it just stops removing it. Providers that do not support `cache_control` ignore the extra key (documented by LiteLLM). This also unblocks power users who attach the marker to non-system messages via `call_model_input_filter`.
- `src/agents/models/openai_chatcompletions.py`, `src/agents/extensions/models/litellm_model.py`, `src/agents/extensions/models/any_llm_model.py`: delegate system-message construction to the new helper.
- Unit test coverage in `tests/test_openai_chatcompletions.py`, `tests/test_openai_chatcompletions_converter.py`, and `tests/model_settings/test_serialization.py`.

#### Behavior changes & compatibility

Compatibility boundary: latest release tag `v0.14.4`. No shim needed.

- `cache_system_prompt` defaults to `None`. When unset, all three Chat Completions-style paths produce byte-identical request payloads to today (plain-string system content). A caller on `v0.14.4` upgrading gets zero wire-level difference until they explicitly opt in.
- The new field is appended after `retry`, the last existing field — no prior positional argument moves.
- The Converter change is a pure passthrough: no caller that wasn't already attaching `cache_control` sees any difference. Callers that were attaching it via `call_model_input_filter` now see the field survive to LiteLLM.
- When the flag is turned on, `span.span_data.input[0].content` becomes a structured list instead of a plain string. The declared type (`Sequence[Mapping[str, Any]]`) already permits this shape — multi-modal user messages have always produced list content — so mature tracers handle it. Users with custom processors that special-case the system message as string should be aware.

### Test plan

- Full `$code-change-verification` stack: `make format`, `make lint`, `make typecheck` (mypy + pyright), `make tests` — all passing (4,105 passed, 6 skipped).
- End-to-end against OpenRouter → Gemini 2.5 Pro with a production-shaped ~36k-char system prompt (handoff topology: gemini-3-flash-preview router → gemini-2.5-pro responder). With `cache_system_prompt=True`: turn-2 `cached_tokens` = 9,499 / prompt_tokens 9,544. With the flag unset or `False`: turn-2 `cached_tokens` = 0.

### Issue number

Closes #3008.

### Checks


seratch:
Let me close this PR for the reason I mentioned at https://github.com/openai/openai-agents-python/issues/3008#issuecomment-4301249715

PR #3011, opened 2026-04-23
feat: Add integer support for max_parallel_tool_calls in ModelSettings

### LINKED ISSUE #1859: Add max parallel_tool_calls value in Agents ModelSettings
### Please read this first

- => Yes
- => yes

### Describe the feature

Currently on ModelSettings we can only choose between activating or deactivating `parallel_tool_calls`param. Its a bool value. 

However, there are cases where we need to restrict a maximum number of tool calls to control the context limits these tools outputs can add, and one tool call at a time is too slow.

Would be nice if we could use something like:

```
agent = Agent(
    name="Assistant",
    tools=[...],
    model="gpt-5",
    model_settings=ModelSettings(max_parallel_tool_calls=4),
)
```

Thanks!!

Moh-Tayyab:
## Summary

Extended  in  to accept integer values for fine-grained control over parallel tool call limits.

## Changes

- ****: Updated type annotation from  to , with updated docstring
- ****: Handle integer  values
- ****: Handle integer  values  
- ****: Handle integer  values
- ****: Handle integer  values (2 locations)

## Behavior

-  → enables parallel tool calls (default)
-  → restricts to 1 tool call per turn
-  → allows up to 4 parallel tool calls
-  → defers to model's default

## Backward Compatibility

All existing // behavior is unchanged. Only additive change for new integer option.

Fixes #1859

### DIFF
--- a/examples/agent_patterns/README.md
+++ b/examples/agent_patterns/README.md
@@ -60,3 +60,14 @@ See the [`input_guardrails.py`](./input_guardrails.py) and [`output_guardrails.p
 You can pause runs for manual approval before executing sensitive tools. This is useful for operations like sending money, deleting data, or running destructive commands.
 
 See [`human_in_the_loop.py`](./human_in_the_loop.py) for the base approval flow and [`human_in_the_loop_custom_rejection.py`](./human_in_the_loop_custom_rejection.py) for run-level tool error formatting when approvals are rejected.
+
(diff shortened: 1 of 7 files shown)


seratch:
This PR includes unrelated changes and it does not make sense. Reviewing your changes before submitting a PR would be appreciated.

PR #3013, opened 2026-04-23
relax websockets upper bound from <16 to <17

mathisobadia:
## Summary

The \`websockets<16\` upper bound introduced in 0.14.0 is overly conservative. websockets 16.0 has no breaking changes that affect this SDK, and the cap is creating unnecessary dependency conflicts for projects already on \`websockets==16.x\`.

### APIs used by this SDK

All websockets usage is in \`src/agents/realtime/openai_realtime.py\` and \`src/agents/voice/\`:

| API | Import path |
|-----|-------------|
| \`connect()\` | \`websockets.connect\` and \`websockets.asyncio.client\` |
| \`ClientConnection\` | \`websockets.ClientConnection\` and \`websockets.asyncio.client.ClientConnection\` |
| \`ConnectionClosed\` | \`websockets.exceptions.ConnectionClosed\` and \`websockets.ConnectionClosed\` |
| \`ConnectionClosedOK\` | \`websockets.exceptions.ConnectionClosedOK\` |
| \`ConnectionClosedError\` | \`websockets.exceptions.ConnectionClosedError\` |

### websockets 16.0 breaking changes

From the official changelog, websockets 16.0 has **one** backwards-incompatible change:

> websockets 16.0 requires Python ≥ 3.10. websockets 15.0 is the last version supporting Python 3.9.

\`openai-agents\` already declares \`requires-python = ">=3.10"\`, so this does not apply.

Every API listed above is present in websockets 16.0 with identical module paths and method signatures — confirmed against the 16.0 source.

### Test compatibility

The existing test suite requires **zero changes** to run against websockets 16.0:

- The mock-based tests (\`patch("websockets.connect", ...)\`) are duck-typed and don't rely on any websockets internals
- The real integration tests in \`TestTransportIntegration\` (which spin up a real local WebSocket server via \`websockets.serve\`) use only APIs that are identical in 16.x
- The \`ClientConnection.close_code\` attribute used in integration test assertions is present and unchanged in 16.0

## Changes

- 3 lines in \`pyproject.toml\` (\`dependencies\`, \`voice\` extra, \`realtime\` extra): \`<16\` → \`<17\`
- Regenerated \`uv.lock\`
- New CI job \`tests-websockets-v16\` that explicitly installs \`websockets==16.0\` and runs the full \`tests/realtime/\` and \`tests/voice/\` suites to prove compatibility

## Impact

Projects pinned to \`websockets==16.x\` currently cannot adopt \`openai-agents>=0.14.0\` without a global websockets downgrade. This change unblocks them.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
     "typing-extensions>=4.12.2, <5",
     "requests>=2.0, <3",
     "types-requests>=2.0, <3",
-    "websockets>=15.0, <16",
+    "websockets>=15.0, <17",
     "mcp>=1.19.0, <2; python_version >= '3.10'",
 ]
 classifiers = [
@@ -35,11 +35,11 @@ Homepage = "https://openai.github.io/openai-agents-python/"
 Repository = "https://github.com/openai/openai-agents-python"
 
 [project.optional-dependencies]
-voice = ["numpy>=2.2.0, <3; python_version>='3.10'", "websockets>=15.0, <16"]
+voice = ["numpy>=2.2.0, <3; python_version>='3.10'", "websockets>=15.0, <17"]
 viz = ["graphviz>=0.17"]
 litellm = ["litellm>=1.83.0"]
 any-llm = ["any-llm-sdk>=1.11.0, <2; python_version >= '3.11'"]
-realtime = ["websockets>=15.0, <16"]
+realtime = ["websockets>=15.0, <17"]
 sqlalchemy = ["SQLAlchemy>=2.0", "asyncpg>=0.29.0"]
 encrypt = ["cryptography>=45.0, <46"]
 redis = ["redis>=7"]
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -131,6 +131,23 @@ jobs:
         if: steps.changes.outputs.run != 'true'
         run: echo "Skipping tests for non-code changes."
 
+  tests-websockets-v16:
```
seratch: We don't want to add this additional test just for this dependency version range

seratch (changes requested):

.github/workflows/tests.yml
mathisobadia: Yes that make total sense, will remove the commit, I wanted to have a way to "prove" it works with version 16

seratch (approved):

PR #3019, opened 2026-04-24
feat: #1167 add opt-in server-prefixed MCP tool names

### LINKED ISSUE #1167: Duplicate tool names in multiple MCP tools cause the agent list to hang

mavrickdeveloper:
### Summary

This fixes the duplicate local MCP tool-name case from #1167 without changing the default behavior. By default, duplicate MCP tool names still fail fast. When `include_server_in_tool_names` is enabled, local MCP tools are exposed with deterministic server-prefixed public names, while calls still go back to the original MCP tool name on the original server.

The generated names are kept safe for function-tool payloads: ASCII-only, capped at 64 characters, and collision-resistant when server names or tool names normalize to the same value. The reservation set also includes local function tool names on the agent so an MCP alias cannot shadow an existing tool.

### Test plan

- `uv run pytest tests/mcp/test_mcp_util.py tests/mcp/test_runner_calls_mcp.py -q`
- `uv run mypy src/agents/agent.py src/agents/mcp/util.py tests/mcp/test_mcp_util.py tests/mcp/test_runner_calls_mcp.py`
- `uv run pyright --project pyrightconfig.json src/agents/agent.py src/agents/mcp/util.py tests/mcp/test_mcp_util.py tests/mcp/test_runner_calls_mcp.py`
- `uv run ruff check src/agents/agent.py src/agents/mcp/util.py tests/mcp/test_mcp_util.py tests/mcp/test_runner_calls_mcp.py`
- `make build-docs`
- Manual check with two real filesystem MCP servers via `npx @modelcontextprotocol/server-filesystem`: verified unique server-prefixed names and invoked `list_allowed_directories` through the prefixed public name.

I also ran the repo verification script. `make format` and `make lint` passed, but the full `make tests` / repo-wide `make typecheck` are currently blocked locally by unrelated optional dependency, runloop, voice, and sandbox failures. The focused MCP checks above pass.

### Issue number

Closes #1167

### Checks


seratch (changes requested):
One remaining collision surface seems worth covering before merge: handoff tool names are not included in `reserved_tool_names`.

`Agent.get_mcp_tools()` currently reserves existing local `FunctionTool` names, but handoffs are also serialized as function tools in the same model request via `handoff.tool_name`. During turn resolution, `handoff_map` is checked before function-tool lookup, so an agent with a handoff named `mcp_calendar__search` plus a calendar MCP server exposing `search` would publish a duplicate public function name, and a model call to `mcp_calendar__search` would be treated as a handoff instead of the MCP tool.

Could we include enabled handoff tool names in the reserved set too, or otherwise validate and fail fast on this collision? That would make the new collision-avoidance behavior cover all local function-like tools, not just entries in `agent.tools`.

src/agents/agent.py
```diff
@@ -187,6 +187,15 @@ class AgentBase(Generic[TContext]):
     mcp_config: MCPConfig = field(default_factory=lambda: MCPConfig())
     """Configuration for MCP servers."""
 
+    def _get_mcp_tool_reserved_names(self) -> set[str]:
```
mavrickdeveloper: the MCP prefix reservation set now includes all local function-like names, including handoff tool names, before generating public MCP aliases. That prevents a generated MCP name from colliding with a handoff name and being routed as a handoff during turn resolution, addressing ---> https://github.com/openai/openai-agents-python/pull/3019#pullrequestreview-4176415756

mavrickdeveloper:
@codex review

mavrickdeveloper:
@codex review

mavrickdeveloper:
@codex review

seratch (approved):
LGTM now; since this PR includes document changes as well, I will merge this PR right before releasing a new version.

Sourav-Nandy-ai:
Hi mavrickdeveloper,

I maintain agent-eval (agent_regress), an open-source regression harness for agent behavior, and I ran a real before-and-after test on this instead of just describing it. I built a local MCP server (FastMCP, stdio transport, one add(a, b) tool named calculator), wired it into a real agents.mcp.MCPServerStdio, and drove one live gpt-4o-mini call per side through openai-agents==0.15.3 (pre-fix) and 0.16.0 (the first release containing this PR). The identical mcp_config={"include_server_in_tool_names": True} was passed both times.

The Results
- Pre-fix: The model-visible tool name comes back as add. The flag does not exist yet in MCPConfig, so it is silently ignored.
- Post-fix: Using the exact same script and dictionary, the tool name comes back as mcp_calculator__add, matching your _build_prefixed_tool_base_name exactly. The call still resolves correctly through the real tool.

This confirms the opt-in prefixing does exactly what the PR claims. It changes what the model sees without touching what actually gets invoked.

Full disclosure, I had to route around one thing on the agent-eval side. The openai_agents_runner() drives its own event loop per call, and a real stdio MCP connection is loop-bound. I wrote a tiny MCPServerStdio subclass that connects lazily on the first list_tools() or call_tool() instead of via async with server: up front. This ensures the connection and the traced call land in the same loop. This is worth mentioning since anyone testing MCP-backed agents against a synchronous runner wrapper will likely hit the same issue.

This is exactly the class of change (tool-naming and selection surface, opt-in, default-off) where unit tests will not catch a silent regression, but a real before-and-after trace will. You can find the full scripts and raw output here: https://github.com/RudrenduPaul/agent-eval

I am curious whether the reserved-name and collision handling (folding in handoff tool names) has a case worth stress-testing next. I am happy to build one if it would be useful.

Best, 
Sourav

PR #3028, opened 2026-04-25
fix: tighten tar and zip member validation

a1vazovsky:
## Summary

Tighten tar/zip archive member validation before extraction.

Changes:
- reject Windows drive-qualified member paths like `C:\tmp\file.txt`
- reject backslash-separated member paths like `..\evil.txt`
- reject entries nested below a non-directory archive member
- add focused tar/zip regression coverage

## Why

These archive paths can be interpreted differently across platforms, especially on Windows. Validating them before extraction keeps sandbox archive handling consistent and avoids relying on platform-specific `Path` behavior.

## Tests

- `UV_CACHE_DIR=/tmp/uv-cache uv run --no-dev --with pytest --with pytest-asyncio pytest tests/sandbox/test_tar_utils.py tests/sandbox/test_extract.py -q`
- `UV_CACHE_DIR=/tmp/uv-cache uv run --no-dev --with ruff ruff check src/agents/sandbox/util/tar_utils.py src/agents/sandbox/session/archive_extraction.py tests/sandbox/test_tar_utils.py tests/sandbox/test_extract.py`
- `PYTHONPYCACHEPREFIX=/tmp/pycache python3 -m py_compile src/agents/sandbox/util/tar_utils.py src/agents/sandbox/session/archive_extraction.py tests/sandbox/test_tar_utils.py tests/sandbox/test_extract.py`
- `git diff --check`
- `UV_CACHE_DIR=/tmp/uv-cache uv run --no-dev --with pytest --with pytest-asyncio --with aiohttp --with docker --with inline-snapshot pytest tests/sandbox -q`

### DIFF
--- a/src/agents/sandbox/session/archive_extraction.py
+++ b/src/agents/sandbox/session/archive_extraction.py
@@ -7,12 +7,12 @@ import tempfile
 import zipfile
 from collections.abc import Awaitable, Callable, Iterator
 from contextlib import contextmanager
-from pathlib import Path, PurePosixPath
+from pathlib import Path, PurePosixPath, PureWindowsPath
 from typing import Literal, cast
 
 from ..errors import ExecNonZeroError, WorkspaceArchiveWriteError
 from ..files import EntryKind, FileEntry
-from ..util.tar_utils import UnsafeTarMemberError, safe_tar_member_rel_path
+from ..util.tar_utils import UnsafeTarMemberError, safe_tar_member_rel_path, validate_tarfile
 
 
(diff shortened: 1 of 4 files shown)


seratch (approved):
I agree this could be useful because sandbox workspace archives should behave consistently across providers and host platforms. Rejecting Windows-qualified or backslash-separated member paths avoids relying on platform-specific Path behavior, and rejecting entries under non-directory members makes extraction fail before any partial or order-dependent writes happen. This seems like the right boundary for a workspace archive format rather than accepting ambiguous paths.

PR #3032, opened 2026-04-26
Fix install wording in README

aolingge:
Small documentation-only cleanup.

No code changes.


seratch:
Thanks for the suggestion. This is a reasonable wording improvement, but the current wording is also acceptable, so we'll close this PR for now.

Issue #2971, opened 2026-04-20
Realtime Tools cause agent response to cut off


aligokalppeker:
# Describe the bug
In Agent SDK, when a tool is registered to a RealtimeAgent, the tool's execution triggers a "create.response" event. That trigger mostly works correctly. But when there is a response in progress, the tool originated "create.response" event causes a cut-off with a RealtimeError. 

````
RealtimeError(message='Conversation already has an active response in progress: resp_DUvxzRrEe3YZSdwIqBdVn. Wait until the response is finished before creating a new one.', type='invalid_request_error', code='conversation_already_has_active_response', event_id=None, param=None)
````

This causes the agent to get stuck and wait for the customer request to respond.

When a tool is called in the agent, the following is the path that results in the agent response for v0.13.0:

agents.realtime.session.RealtimeSession._enqueue_tool_call_task
agents.realtime.session.RealtimeSession._handle_tool_call
agents.realtime.openai_realtime.OpenAIRealtimeWebSocketModel.send_event
agents.realtime.openai_realtime.OpenAIRealtimeWebSocketModel._send_tool_output

As there are limitless scenarios that can cause this, it primarily happens within the following case;
- Async tools calls, where outputs are produced simultaneously.
- When "create.response" is triggered outside the agent, and then the agent runs the tool

This process is improved with v0.13.1, but edge cases seem to exist.

### Debug information
- Agents SDK version in production: v0.14.2
- Python version: Python 3.11

### Repro steps
Trigger multiple tools in an async fashion where each one generates an output.

### Expected behavior
Tool-triggered "create.response" messages should be scheduled to occur after the response is completed if a response is in progress.

nileshpatil6:
Hi! I've traced through the code and would like to work on this.

After reading through the sequencer in `openai_realtime.py`, the `_send_response_create_when_idle` / `wait_for_response_create_slot` machinery handles the serialization of concurrent `response.create` sends correctly in the common case. However I can see two edge cases where it may still fire prematurely:

1. **Race between `response.done` and the next `reserve_response_create_request`**: If `response.done` arrives and clears `_ongoing_response` before a second async tool output has had a chance to call `reserve_response_create_request`, a background task from a concurrent source (e.g. an external `response.create` sent via `send_event`) can slip through the now-free slot and conflict with the tool's pending create.

[...]

PR #3033, opened 2026-04-26
feat(realtime): allow tools to opt out of automatic response.create after completion

jawwad-ali:
### Summary

Realtime sessions hard-coded \`start_response=True\` when sending tool outputs back to the model, so every tool unconditionally triggered a follow-up \`response.create\`. Side-effect tools (analytics, background-job schedulers, telemetry) had no way to stay silent after completion — exactly the gap @aligokalppeker raised in [#2971 (comment)](https://github.com/openai/openai-agents-python/issues/2971#issuecomment-4294462255):

> "Response trigger at the end of the tool should be provided as an option to the SDK user; not every tool result is required to create a response."

This PR exposes that toggle on \`FunctionTool\` and the \`@function_tool(...)\` decorator. The transport layer (\`OpenAIRealtimeWebSocketModel._send_tool_output\`) already honors \`RealtimeModelSendToolOutput.start_response\`; this change just lets tool authors reach it.

### What changes

- \`FunctionTool.start_response: bool = True\` — new kw-only field (preserves positional-API compatibility per AGENTS.md).
- \`@function_tool(start_response=False)\` — new kwarg on the decorator (and matching overloads).
- \`RealtimeSession._handle_tool_call\` — for the function-tool success path, reads \`func_tool.start_response\` instead of hard-coding \`True\`.
- \`RealtimeSession._send_tool_rejection\` — same treatment for approval-rejected outputs (the original tool's preference still applies).
- The handoff path keeps \`start_response=True\` because the new agent must speak after a handoff.

Default remains \`True\`, so every existing tool keeps current behavior.

### Scope note (re: #2971's race report)

Per @seratch's [response in the same thread](https://github.com/openai/openai-agents-python/issues/2971#issuecomment-4284626744), \`v0.14.2\` already routes \`response.create\` through \`_ResponseCreateSequencer\` and the active-response race is gated until a concrete repro exists. **This PR does not touch the sequencer or add any tracking counters.** It only addresses the per-tool ergonomics raised in @aligokalppeker's follow-up, which is independent of the race-detection path.

### Test plan

Verified locally on Windows (Python 3.13):

- \`uv run ruff format\` — pass
- \`uv run ruff check\` — pass (\`All checks passed!\`)
- \`uv run mypy src/agents/tool.py src/agents/realtime/session.py tests/test_function_tool.py tests/realtime/test_session.py\` — pass (\`Success: no issues found in 4 source files\`)
- \`uv run pyright --project pyrightconfig.json src/agents/tool.py src/agents/realtime/session.py tests/test_function_tool.py tests/realtime/test_session.py\` — pass (\`0 errors, 0 warnings, 0 informations\`)
- \`uv run pytest tests/test_function_tool.py tests/realtime/test_session.py -q\` — **112 passed**
- Broader sweep \`-k \"tool or function or realtime\"\` — **1074 passed, 1 skipped**. 4 unrelated pre-existing failures on clean \`main\` (\`test_tracing_errors{,_streamed}.py::test_tool_call_error\`, \`test_openai_realtime.py::TestSendEventAndConfig::test_interrupt_*\`) confirmed via \`git stash\` round-trip — not introduced by this change.

New tests (6 total):

| Test | Coverage |
|------|----------|
| \`test_function_tool_start_response_defaults_to_true\` | \`FunctionTool\` default value |
| \`test_function_tool_start_response_can_be_set_to_false\` | Direct kw-only construction |
| \`test_function_tool_decorator_threads_start_response\` | \`@function_tool(start_response=False)\` |
| \`test_function_tool_decorator_default_start_response_is_true\` | Plain \`@function_tool\` keeps \`True\` |
| \`test_function_tool_start_response_false_skips_response_create\` | Realtime session forwards \`False\` to transport |
| \`test_function_tool_default_start_response_remains_true\` | Backward-compat for existing realtime tools |

### Issue number

Refs #2971

### Checks

- [x] I've added/updated the relevant documentation (\`docs/realtime/guide.md\` gained a short example for \`@function_tool(start_response=False)\`)

### DIFF
--- a/docs/realtime/guide.md
+++ b/docs/realtime/guide.md
@@ -206,6 +206,18 @@ agent = RealtimeAgent(
 )
 ```
 
+By default, the model generates a follow-up response after each tool's output is sent back. For
+side-effect tools where you do not want the agent to speak again immediately (for example, a tool
+that records analytics or schedules background work), pass `start_response=False`:
+
+```python
+@function_tool(start_response=False)
+def log_analytics(event: str) -> str:
+    """Record an analytics event without prompting the agent to speak again."""
+    record_event(event)
+    return "logged"
+```
+
 ### Tool approvals
 
 Function tools can require human approval before execution. When that happens, the session emits `tool_approval_required` and pauses the tool run until you call `approve_tool_call()` or `reject_tool_call()`.
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -512,7 +512,7 @@ class RealtimeSession(RealtimeModelListener):
             RealtimeModelSendToolOutput(
                 tool_call=event,
                 output=rejection_message,
-                start_response=True,
+                start_response=bool(getattr(tool, "start_response", True)),
             )
         )
 
@@ -656,7 +656,7 @@ class RealtimeSession(RealtimeModelListener):
                 RealtimeModelSendToolOutput(
                     tool_call=event,
                     output=_serialize_tool_output(result),
-                    start_response=True,
+                    start_response=bool(getattr(func_tool, "start_response", True)),
                 )
             )
 
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -350,6 +350,13 @@ class FunctionTool:
     defer_loading: bool = False
     """Whether the Responses API should hide this tool definition until tool search loads it."""
 
+    start_response: bool = field(default=True, kw_only=True)
+    """Whether the model should automatically generate a follow-up response after this tool's
+    output is sent back. Currently only honored by Realtime sessions. Set to False when a tool
+    completes a side-effect (e.g. updates external state, logs telemetry, schedules background
+    work) and you do not want the agent to immediately speak again afterwards. Defaults to True
+    to preserve existing behavior."""
+
     _failure_error_function: ToolErrorFunction | None = field(
         default=None,
         kw_only=True,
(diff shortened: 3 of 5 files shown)


seratch:
Thanks for the suggestion. I think having this flag on the function_tool decorator side may not be optimal in terms of the SDK design. Having a universal option only per realtime agent session may make sense, but I am not confident enough that adding such an option is really helpful for some use cases. We don't plan to add this option at least for now, so let us close this PR.

aligokalppeker:
@seratch, what do you think about avoiding forcing the creation of responses when a tool is executed? Do we have to force response creation upon tool execution? Another valid scenario for tool execution is just to add a response to the bot context, not trigger a response.

In real-time voice, one might want a tool to save a user's/tool's data in the background without interrupting the audio stream.

This is a widely used concept with modern frameworks.

LangGraph treats agents as cyclical graphs (State Machines). Tool execution is just a node that modifies a global "State." It does not automatically trigger an LLM response unless you explicitly draw an edge to the LLM node.

CrewAI is built around Tasks, and given Tools to accomplish those tasks. The agent will execute tools internally in a loop without returning a conversational response to the user until the Task is marked complete.

Haystack ToolInvoker, as a component, is wired, connecting outputs to inputs. If you don't wire the ToolInvoker output back into a Generator (LLM), it doesn't generate a response.

Forcing an LLM response after every tool execution is an anti-pattern for Realtime and Agentic workflows. Modern frameworks treat tool execution as state mutation (LangGraph, Rasa), pluggable functions (Semantic Kernel), background tasks (CrewAI), or explicit pipeline components (Haystack). Tying tool execution strictly to dialogue generation introduces unavoidable latency and prevents silent background operations, which are critical for smooth Voice/Realtime UX.

Closing this PR w/o any alternative suggested out there is not a correct judgment. The framework around a model is an important aspect to make it usable.... Similar to this case, OpenAI really makes other models' jobs easier with such decisions.

PR #3035, opened 2026-04-27
examples/sandbox: Add Unix local PTY multiturn example

liuzengh:
## Summary

Adds a new Unix-local sandbox example that runs two agent turns against the same live sandbox session. The first turn writes a file in the workspace, the second turn reads and appends to that same file, and the script verifies the final file contents through the sandbox session.

Also updates the sandbox examples README with the new runnable example.

## Validation

- `.venv/bin/python -m py_compile examples/sandbox/unix_local_pty_multiturn.py`
- `source ../gpt.sh && PYTHONUNBUFFERED=1 .venv/bin/python examples/sandbox/unix_local_pty_multiturn.py`

The live run verified that the second turn saw the first turn's `state/session_counter.txt` and appended `turn-2: beta`.


seratch:
Thanks for sharing this idea. We won't add this example within this SDK repo, but please feel free to share it in your own repositories and/or posts. Thanks again for sharing it!

PR #3041, opened 2026-04-28
feat: add Sprites as a sandbox provider

vaurdan:
This PR adds [Sprites](https://sprites.dev) (Fly.io's sandbox-VM-as-a-service platform) as a first-class sandbox backend, peer to E2B / Modal / Vercel / Daytona / Runloop / Blaxel / Cloudflare. It uses the same `BaseSandboxClient` / `BaseSandboxSession` contract; same `WorkspaceShellCapability` / `Filesystem()` / `apply_patch` / PTY surface, and the same opt-in extra (`pip install "openai-agents[sprites]"`).

The new module lives at `agents.extensions.sandbox.sprites` and ships:

- `SpritesSandboxClient` / `SpritesSandboxClientOptions` — token from `SPRITES_API_TOKEN` env var or kwarg, base URL from `SPRITES_API_URL`.
- `SpritesSandboxSession` / `SpritesSandboxSessionState` — exec runs over the multiplexed `ControlConnection` WebSocket from `sprites-py` (no sync thread blocking); fs read/write via `SpriteFilesystem`; PTY parity with E2B; tar-based workspace persistence.
- Lifecycle is "ephemeral by default, named-attach when `options.sprite_name` is set" — both modes verified end-to-end.
- Three opt-in capabilities for Sprites-specific affordances:
  - `SpritesPlatformContext()` — auto-injects `/.sprite/llm.txt` into the system prompt so the model knows about `sprite-env services`, checkpoints, URL routing, security rules. Module-cached per sprite so purely-conversational turns don't re-execute `cat`.
  - `SpritesUrlAccess(allow_public=False)` — adds a `set_sprite_url_visibility(visibility)` tool. Default-deny on `"public"` — apps must explicitly opt in.
  - `SpritesCheckpoints(allow_restore=False)` — adds `create_sprite_checkpoint`, `list_sprite_checkpoints`, and (when `allow_restore=True`) `restore_sprite_checkpoint` tools. Wraps sprites-py's iterator-streaming checkpoint API in `asyncio.to_thread`.
- Lazy wake-up across both lifecycle paths: `_ensure_sprite()` skips the wait-for-running poll; the first I/O operation drives the wake-up via `_ensure_warm()`.
- Idle close after `idle_close_seconds=60.0` (configurable, set to 0 to disable): closes pooled control connections so the sprite drops back to `warm` and stops accruing running-state cost. Next I/O reopens.
- These guarantee that the sprite is in the `warm` state while it's not needed (after 60 seconds of inactivity), and quickly wakes up as soon as the agent tries to use it. 

#### Compatibility

Pure additive. New optional extra (`[sprites]`), new sub-package, no changes to `RunState` schema, no reordering of existing public dataclass fields. All six entries in `tests/sandbox/test_compatibility_guards.py` parametrized for the new types.

### Test plan

- 119 new unit tests in `tests/extensions/test_sandbox_sprites.py` covering: options/state roundtrip, env-var auth resolution, ephemeral and named-attach lifecycles, exec mapping (success / timeout / transport error), PTY (start / write_stdin / terminate / finalize), exposed-port resolution and validation, fs read/write with error mapping, tar-based persistence, the three capabilities (default-deny paths, opt-in paths, edge cases), lazy-warm behavior across both paths, the idle-close watcher, and the platform-context cache.
- All six entries in `tests/sandbox/test_compatibility_guards.py` parametrized for the new options / state / discriminator strings / positional-field-order tests.
- `make format`, `make lint`, `make typecheck`, `make tests` all green. Coverage holds.
- Smoke-tested end-to-end against a real Sprites org: ephemeral session creates a sprite, agent uses native `shell` + `apply_patch`, manifest materializes, sprite is deleted on shutdown. Also verified the capability flow (made the sprite URL public, snapshotted the workspace, listed checkpoints).


alfozan:
Thanks; just emailed re integration testing credentials

src/agents/extensions/sandbox/sprites/sandbox.py
```diff
@@ -0,0 +1,1353 @@
...
+        return session
+
+    async def resume(self, state: SandboxSessionState) -> SandboxSession:
+        if not isinstance(state, SpritesSandboxSessionState):
+            raise TypeError("SpritesSandboxClient.resume expects a SpritesSandboxSessionState")
+
+        inner = SpritesSandboxSession.from_state(state, token=self._token, base_url=self._base_url)
+        try:
+            await inner._ensure_sprite()
+            inner._set_start_state_preserved(True)
```
alfozan, line 1461: Could you distinguish “reattached existing sprite” from “created a replacement sprite” here?

For `created_by_us=True`, `_ensure_sprite()` creates a new sprite when `_sprite` is empty. If this is a replacement after the original ephemeral sprite was deleted, `resume()` still calls `_set_start_state_preserved(True)`. With `NoopSnapshot` and `state.workspace_root_ready=True` from the previous run, `BaseSandboxSession.start()` can treat the fresh sprite as preserved and skip the full durable manifest apply.

I think replacement should clear `workspace_root_ready` and not mark start state as preserved; only a true reattach to an existing sprite should do that.

src/agents/extensions/sandbox/sprites/sandbox.py
```diff
@@ -0,0 +1,1353 @@
...
+            url_auth=options.url_auth,
+            ram_mb=options.ram_mb,
+            cpus=options.cpus,
+            region=options.region,
+            storage_gb=options.storage_gb,
+            exposed_ports=options.exposed_ports,
+            env=dict(options.env or {}) or None,
+            timeout_ms=options.timeout_ms,
+            workspace_persistence=options.workspace_persistence,
+        )
```
alfozan, line 1417: It looks like `wait_for_running_timeout_s` and `idle_close_seconds` are exposed on `SpritesSandboxClientOptions`, but not copied into `SpritesSandboxSessionState` here. `_wait_for_sprite_running()` also uses `timeout_ms`/default rather than `wait_for_running_timeout_s`.

Can you either wire these through fully, including tests from options -> state -> behavior, or remove the public knobs until they are implemented? Right now callers cannot actually tune the advertised wait timeout or disable idle close through client options.

src/agents/extensions/sandbox/sprites/sandbox.py
```diff
@@ -0,0 +1,1353 @@
...
+            return ExecResult(stdout=b"", stderr=b"", exit_code=0)
+
+        await self._ensure_warm()
+
+        control: ControlConnection | None = None
+        op_conn: OpConn | None = None
+        try:
+            control = await self._ensure_control()
+            try:
+                op_conn = await control.start_op(
```
alfozan, line 804: I do not see `manifest.environment` being resolved or applied for Sprites exec/PTY/create. Other hosted backends pass manifest env at sandbox creation or inject it into command execution, so manifests that rely on env vars will behave differently here.

Can Sprites apply the merged env, or explicitly reject/document env as unsupported? Silent no-op behavior seems hard to debug.

alfozan:
Could you add cloud mount support for Sprites as part of this provider?

Please add a Sprites mount strategy and focused tests for the common bucket flows, or link a clearly tracked follow-up if cloud mounts should land separately.

vaurdan:
@alfozan thanks for the suggestion! We have implemented the cloud mount support with `rclone`. 

I believe the PR is ready to be reviewed, but let me know otherwise.

vaurdan:
FYI, I'm currently working on yours and codex feedback! Thanks!

src/agents/extensions/sandbox/sprites/sandbox.py
vaurdan: Good catch! Both knobs are now threaded through. Added `wait_for_running_timeout_s`  to `SpritesSandboxSessionState`, `create()` copies both fields from options into state, and `_wait_for_sprite_running` now consults the new field instead of `timeout_ms`. 

This is covered by `test_create_propagates_option_knobs_into_session_state` and  `test_wait_for_running_uses_state_timeout`, plus the compat guard pins the new state field order.

src/agents/extensions/sandbox/sprites/sandbox.py
vaurdan: Thanks for the feedback! That was the wrong shape indeed. `resume()` now always tries `get_sprite(name)` first via a new `_try_attach_existing_sprite` helper. 

On a real re-attach it marks start state preserved as before:
 *  If the sprite is gone and the session was ephemeral it then clears `workspace_root_ready` and skips the preserved flag so `start()` runs a full manifest apply against the replacement. 
  * If it was a named-attach it raises `WorkspaceStartError(reason="sprite_not_found")` rather than silently swap the caller's sprite. 

Transient platform errors during the reattach probe also raise the exception, instead of falling through to recreate, since "the call failed" isn't the same signal as "the sprite is gone.".

src/agents/extensions/sandbox/sprites/sandbox.py
vaurdan: Yeah, that was a gap. I added a `small _resolved_envs()` helper that merges `state.env` with `manifest.environment.resolve()` (manifest wins on conflict), and both` _exec_with_cwd` and `pty_exec_start` now prepend an `env -- K=V … argv` prefix.
         
It's using argv-form rather than a shell wrapper since the WS exec path does `execvp` directly.  The bootstrap `mkdir` in `_prepare_backend_workspace` opts out with `apply_env=False` since the workspace doesn't exist yet.

vaurdan:
@alfozan your feedback has been addressed, support for cloud mount has been added, and the PR is ready for another round. Thank you!

PR #3054, opened 2026-04-30
Add Perplexity provider example via Agent API

jliounis:
## Summary

Adds two runnable examples showing how to use Perplexity's APIs with this SDK — no core SDK changes.

- `examples/model_providers/perplexity_provider.py` — routes Agent API calls to `https://api.perplexity.ai` by giving an `AsyncOpenAI` client to `OpenAIChatCompletionsModel` via a `ModelProvider`. Default model is `sonar-pro`. Mirrors the structure of the existing `custom_example_provider.py`.
- `examples/tools/perplexity_search.py` — wraps Perplexity's Search API (`POST /search`) as a `@function_tool` using `httpx`, so any agent can call it. Useful when you want raw search results rather than a model-summarised answer.
- `examples/model_providers/README.md` — adds a Perplexity section documenting the env var and the entry point.

Both examples send `X-Pplx-Integration: openai-agents/<package-version>` on every outgoing request to Perplexity, with the version read from `importlib.metadata`. No new dependencies are introduced (`httpx` is already a transitive dep via `openai`/`mcp`).

## Testing

- Both files parse cleanly with `ast.parse`.
- `uv run python -c "import perplexity_provider; import perplexity_search"` — imports succeed; attribution header resolves to `openai-agents/0.14.8`; `base_url` and default model match the spec.
- End-to-end runs require a real `PERPLEXITY_API_KEY` and were not executed.

### DIFF
--- a/examples/model_providers/README.md
+++ b/examples/model_providers/README.md
@@ -22,3 +22,17 @@ Direct-model examples let you override the target model:
 uv run examples/model_providers/any_llm_provider.py --model openrouter/openai/gpt-5.4-mini
 uv run examples/model_providers/litellm_provider.py --model openrouter/openai/gpt-5.4-mini
 ```
+
+## Perplexity
+
+Perplexity exposes an OpenAI-compatible chat completions endpoint, so you can route
+requests through an `AsyncOpenAI` client by overriding `base_url`:
+
+```bash
+export PERPLEXITY_API_KEY="..."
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for sharing this. While we won't add this example to this SDK repo, please feel free to host it in your own repo/documentation.

PR #3056, opened 2026-05-01
fix: surface model refusals as final output

### LINKED ISSUE #3055: Model refusals with structured output aren't handled
If an API [response is a refusal](https://developers.openai.com/api/docs/guides/structured-outputs#refusals), and structured output has been requested, the agents SDK just tries again and keeps looping until `max_turns` is reached.

This is because `ItemHelpers.extract_text()` ignores refusals, so no text is returned and in that case the system just tries again.

This seems odd because `ItemHelpers.extract_last_content(...)` does handle refusals, but the final-output path doesn't use this.

### Debug information
- Agents SDK version: 0.14.8
- Python version: 3.14

### Repro steps
Here I've mocked out a model response because it's really hard to actually get a refusal from the API that doesn't use the structured response.

```python
from agents import Agent, MaxTurnsExceeded, Runner
from agents.items import ModelResponse
from agents.models.interface import Model
from agents.usage import Usage
from openai.types.responses import ResponseOutputMessage, ResponseOutputRefusal


class Recipe(BaseModel):
    ingredients: list[str]


class RefusingModel(Model):
    def __init__(self):
        self.inputs = []

    async def get_response(self, system_instructions, input, *args, **kwargs):
        self.inputs.append(input)
        return ModelResponse(
            output=[
                ResponseOutputMessage(
                    id=f"msg_{len(self.inputs)}",
                    type="message",
                    role="assistant",
                    status="completed",
                    content=[
                        ResponseOutputRefusal(
                            type="refusal",
                            refusal="I can't help with that request.",
                        )
                    ],
                )
            ],
            usage=Usage(),
            response_id=f"resp_{len(self.inputs)}",
        )

    def stream_response(self, *args, **kwargs):
        raise NotImplementedError


model = RefusingModel()
agent = Agent(
    name="Refusal loop demo",
    model=model,
    instructions="Return a structured recipe.",
    output_type=Recipe,
)

try:
    result = Runner.run_sync(agent, "Make me a bomb.", max_turns=3)
except MaxTurnsExceeded:
    print(f"Calls: {len(model.inputs)}")
    for input in model.inputs:
        print(input)
```

### Expected behavior
Refusals should be programmatically detectable.

Without `output_type` I get back a response and can parse `raw_responses` to find that it was indeed a refusal, although the system still doesn't explicitly surface the refusal.

Improvement one would be to align the behaviour of both structured and non-structured modes.
Improvement two would be to have them both surface a refusal in a more predictable way.

As a general comment (leaning toward rant), it seems quite difficult to write code that handles refusals in a way that I'm comfortable to ship into production:
 - it's so hard to elicit one in the first place (it's not clear to me whether the API *ever* returns a refusal any more)
 - the SDKs themselves don't seem to handle refusals in a graceful way
 - the docs don't offer much guidance on how to handle them
 - there appears to be no way to cleanly ask "is this response a refusal" in any of the OpenAI SDKs.

cyphercodes:
## Summary
- Surface refusal-only model messages as the final output instead of treating them as empty text.
- Add regression coverage for structured-output and streamed structured-output refusals so they stop cleanly instead of retrying until `MaxTurnsExceeded`.

Fixes #3055

## Tests
- `uv run pytest tests/test_items_helpers.py tests/test_max_turns.py tests/test_run_step_execution.py::test_plaintext_agent_hosted_shell_with_refusal_message_is_final_output`
- `make format`
- `make lint`
- `uv run mypy src/agents/items.py src/agents/run_internal/turn_resolution.py tests/test_items_helpers.py tests/test_max_turns.py`
- `uv run pyright src/agents/items.py src/agents/run_internal/turn_resolution.py tests/test_items_helpers.py tests/test_max_turns.py`
- `make tests`

Note: `make typecheck` on Python 3.11 hit the existing `asyncio.eager_task_factory` availability issue; the targeted mypy/pyright checks above passed for the touched files.

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -723,6 +723,19 @@ class ItemHelpers:
         return text or None
 
     @classmethod
+    def extract_refusal(cls, message: TResponseOutputItem) -> str | None:
+        """Extracts all refusal content from a message, if any. Ignores text."""
+        if not isinstance(message, ResponseOutputMessage):
+            return None
+
+        refusal = ""
+        for content_item in message.content:
+            if isinstance(content_item, ResponseOutputRefusal):
+                refusal += content_item.refusal or ""
+
+        return refusal or None
+
+    @classmethod
     def input_to_new_input_list(
         cls, input: str | list[TResponseInputItem]
     ) -> list[TResponseInputItem]:
--- a/src/agents/run_internal/turn_resolution.py
+++ b/src/agents/run_internal/turn_resolution.py
@@ -671,6 +671,9 @@ async def execute_tools_and_side_effects(
     potential_final_output_text = (
         ItemHelpers.extract_text(message_items[-1].raw_item) if message_items else None
     )
+    potential_final_output_refusal = (
+        ItemHelpers.extract_refusal(message_items[-1].raw_item) if message_items else None
+    )
 
     if not processed_response.has_tools_or_approvals_to_run():
         has_tool_activity_without_message = not message_items and bool(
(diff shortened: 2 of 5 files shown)


seratch:
Thanks for sharing this idea, but this does not resolve the mentioned issue. My PR https://github.com/openai/openai-agents-python/pull/3057 will more properly improve the way to handle the situation, let us close yours. Thanks again for your interest.

PR #3075, opened 2026-05-01
fix: #3074 restore SIGINT defaults for UnixLocal PTY children

### LINKED ISSUE #3074: UnixLocal PTY children can inherit ignored SIGINT and miss Ctrl-C interrupts
## Summary

`UnixLocalSandboxSession` PTY children appear to inherit the parent process `SIGINT` disposition. If the parent is temporarily ignoring `SIGINT`, writing `\x03` to a TTY-backed subprocess does not interrupt the child as expected.

## Reproduction

The existing PTY interrupt behavior can fail with a long-running TTY process such as `sleep 30` when the parent process is ignoring `SIGINT`.

A minimal repro shape is:

1. Set the parent process handler for `SIGINT` to `SIG_IGN`
2. Start a PTY-backed subprocess via `UnixLocalSandboxSession.pty_exec_start(..., tty=True)`
3. Write `\x03` with `pty_write_stdin`
4. Observe that the subprocess remains alive instead of exiting with signal `SIGINT`

## Expected behavior

PTY-backed children should treat Ctrl-C as an interrupt regardless of whether the parent process is temporarily ignoring `SIGINT`.

## Actual behavior

The child can inherit the ignored `SIGINT` disposition, so terminal-generated Ctrl-C is ignored and the subprocess remains running.

## Root cause hypothesis

In the UnixLocal PTY path, the child `preexec_fn` creates a new session and sets the controlling terminal, but it does not restore `SIGINT` to `SIG_DFL` before `exec`. That makes PTY behavior depend on the parent's current signal disposition.

## Proposed fix

Reset `SIGINT` to `SIG_DFL` in the PTY child `preexec_fn` before `exec`, and add a regression test that temporarily sets the parent handler to `SIG_IGN` and verifies that `\x03` still interrupts the child.

Aphroq:
### Summary

This pull request fixes a UnixLocal PTY bug where TTY-backed child processes can inherit an ignored `SIGINT` disposition from the parent process and fail to react to Ctrl-C.

### Test plan

- `uv run pytest -q tests/sandbox/test_unix_local.py::TestUnixLocalPty::test_pty_ctrl_c_interrupts_long_running_process tests/sandbox/test_unix_local.py::TestUnixLocalPty::test_pty_ctrl_c_interrupts_even_if_parent_ignores_sigint`

### Issue number

Resolves #3074 for the SIGINT issue

### Checks

### DIFF
--- a/src/agents/sandbox/sandboxes/unix_local.py
+++ b/src/agents/sandbox/sandboxes/unix_local.py
@@ -283,6 +283,9 @@ class UnixLocalSandboxSession(BaseSandboxSession):
             def _preexec() -> None:
                 os.setsid()
                 fcntl.ioctl(secondary_fd, termios.TIOCSCTTY, 0)
+                # PTY children should always treat Ctrl-C as an interrupt even if the parent
+                # process temporarily ignores SIGINT under the test runner.
+                signal.signal(signal.SIGINT, signal.SIG_DFL)
 
             try:
                 process = await asyncio.create_subprocess_exec(
--- a/tests/sandbox/test_unix_local.py
+++ b/tests/sandbox/test_unix_local.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import signal
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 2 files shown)


seratch:
@codex review

PR #3086, opened 2026-05-02
fix: #2940 emit RealtimeHistoryUpdated on transcript_delta

### LINKED ISSUE #2940: RealtimeSession does not emit `history_updated` for `transcript_delta`

Quratulain-bilal:
## Summary
  - Fixes #2940. `RealtimeSession` previously updated `self._history` on
  `transcript_delta` events but only forwarded the raw model event, so
  subscribers following the documented `history_added` / `history_updated`
  pattern missed live transcript updates.
  - Emit `RealtimeHistoryUpdated` after the history mutation, mirroring the
  `input_audio_transcription_completed` branch.
  - Updated the existing test that asserted `transcript_delta` was an ignored
  event, and added a new test asserting the event now emits
  `RealtimeHistoryUpdated`.

  ## Test plan
  - [x] `uv run pytest tests/realtime/test_session.py` — 66 passed


seratch:
Thanks for sharing this. This change could bring way more events to existing apps, and I haven't verified how impactful it is. Thus, I am holding off so far.

PR #3092, opened 2026-05-02
fix: reject string-like shell commands

### LINKED ISSUE #3091: coerce_shell_call splits string commands into character lists
### Please read this first

### Describe the bug

`coerce_shell_call()` in `src/agents/run_internal/tool_execution.py` treats a string-valued `action.commands` payload as a generic `Sequence`, so a single command like `"echo hi"` is normalized into a character list instead of being rejected.

Current code:

```py
commands_value = get_mapping_or_attr(action_payload, "commands")
if not isinstance(commands_value, Sequence):
    raise ModelBehaviorError("Shell call action is missing commands.")
commands: list[str] = []
for entry in commands_value:
    if entry is None:
        continue
    commands.append(str(entry))
```

Because `str` is a `Sequence`, this produces:

```py
["e", "c", "h", "o", " ", "h", "i"]
```

instead of a valid `list[str]` command payload.

This corrupts shell execution semantics and also pollutes approval / audit behavior, because the runtime ends up reasoning about individual characters as separate commands.

### Debug information

- Agents SDK version: `main` at `f2fb9ffb` / latest release boundary `v0.15.1`
- Python version: Python 3.12

### Repro steps

Minimal reproducer:

```python
from agents.run_internal.tool_execution import coerce_shell_call

payload = {
    "call_id": "shell-1",
    "action": {
        "commands": "echo hi",
    },
}

result = coerce_shell_call(payload)
print(result.action.commands)
```

Current output:

```text
['e', 'c', 'h', 'o', ' ', 'h', 'i']
```

### Expected behavior

`action.commands` should be validated as a real command list, not any arbitrary `Sequence`.

At minimum, string-like values such as `str`, `bytes`, and `bytearray` should be rejected with a clear `ModelBehaviorError`, so malformed shell payloads fail loudly instead of being normalized into character-by-character commands.

Aphroq:
### Summary

This pull request fixes shell call command normalization when the model returns a string-like `action.commands` payload.

- reject top-level `str`, `bytes`, and `bytearray` values in `coerce_shell_call()` instead of treating them as generic sequences
- raise a clear `ModelBehaviorError` before malformed payloads can be split into character-by-character commands
- add regression coverage for string-like `commands` payloads while keeping valid list-based shell calls unchanged

### Test plan

- `uv run pytest tests/test_shell_call_serialization.py`

### Issue number

Closes #3091

### Checks

### DIFF
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -619,8 +619,12 @@ def coerce_shell_call(tool_call: Any) -> ShellCallData:
         raise ModelBehaviorError("Shell call is missing an action payload.")
 
     commands_value = get_mapping_or_attr(action_payload, "commands")
-    if not isinstance(commands_value, Sequence):
-        raise ModelBehaviorError("Shell call action is missing commands.")
+    if isinstance(commands_value, str | bytes | bytearray) or not isinstance(
+        commands_value, Sequence
+    ):
+        raise ModelBehaviorError(
+            "Shell call action commands must be a sequence of command strings."
+        )
     commands: list[str] = []
     for entry in commands_value:
         if entry is None:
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (approved):

PR #3094, opened 2026-05-02
fix: reject external symlink targets during hydrate

### LINKED ISSUE #3093: workspace hydrate accepts symlink targets outside the archive root
### Describe the bug

`validate_tarfile()` currently validates tar member paths, but it does not validate a symlink member's `linkname`. As a result, workspace hydrate paths that allow symlinks will accept entries whose targets point outside the extracted archive root.

That shows up most clearly in workspace restore flows:

- `UnixLocalSandboxSession.hydrate_workspace()` calls `safe_extract_tarfile()`
- `DockerSandboxSession.hydrate_workspace()` validates the tar and then streams it into `tar -x`
- several extension backends validate tar bytes the same way before hydrate

The regular `session.extract()` tar path is already using `allow_symlinks=False`, so this is not "all tar uploads are unsafe." The problem is narrower: restore / hydrate flows that intentionally allow symlink members also allow targets like `/etc/passwd` or `../../outside`.

I can reproduce this on current `main` by creating a tar with `leak -> /etc/passwd`, passing it through `validate_tarfile()`, and then extracting it. Validation succeeds and the restored workspace ends up with a symlink that points outside the extracted root.

### Debug information

- Agents SDK version: `main` at `f2fb9ffb` (latest release boundary: `v0.15.1`)
- Python version: Python 3.12

### Repro steps

```python
import io
import tarfile
import tempfile
from pathlib import Path

from agents.sandbox.util.tar_utils import safe_extract_tarfile, validate_tarfile

buf = io.BytesIO()
with tarfile.open(fileobj=buf, mode="w") as tf:
    info = tarfile.TarInfo("leak")
    info.type = tarfile.SYMTYPE
    info.linkname = "/etc/passwd"
    tf.addfile(info)

buf.seek(0)
with tarfile.open(fileobj=buf, mode="r:*") as tf:
    validate_tarfile(tf)
    with tempfile.TemporaryDirectory() as td:
        safe_extract_tarfile(tf, root=Path(td))
        print((Path(td) / "leak").readlink())
```

Current result:

```text
/etc/passwd
```

### Expected behavior

Workspace hydrate / restore validation should reject symlink targets that escape the extracted archive root.

Keeping the general symlink support is still useful for normal workspace snapshots, but restore paths that materialize an archive into a sandbox workspace should fail loudly when a symlink target is absolute or traverses outside the archive root.

Aphroq:
### Summary

This pull request rejects tar symlink targets that point outside the extracted archive root during workspace hydrate and restore flows.

- add strict symlink-target validation to `tar_utils` for hydrate paths while keeping the general tar helpers' default behavior unchanged
- enable the stricter validation in UnixLocal, Docker, and tar-based extension backends before they materialize a workspace archive
- add regression coverage for strict tar validation plus UnixLocal and Vercel hydrate paths

**Compatibility change:** If a user’s sandbox snapshot/workspace contains symlinks intentionally pointing to system paths (e.g., `/usr/bin/python`, certain toolchain paths, or host mount points), then after hydrate/restore these will change from “restorable” to “refused to restore.”

### Test plan

- `uv run pytest tests/sandbox/test_tar_utils.py tests/sandbox/test_extract.py tests/extensions/sandbox/test_vercel.py`

### Issue number

Closes #3093

### Checks

### DIFF
--- a/src/agents/extensions/sandbox/blaxel/sandbox.py
+++ b/src/agents/extensions/sandbox/blaxel/sandbox.py
@@ -600,7 +600,10 @@ class BlaxelSandboxSession(BaseSandboxSession):
             raise WorkspaceWriteTypeError(path=Path(tar_path), actual_type=type(payload).__name__)
 
         try:
-            validate_tar_bytes(bytes(payload))
+            validate_tar_bytes(
+                bytes(payload),
+                allow_external_symlink_targets=False,
+            )
         except UnsafeTarMemberError as e:
             raise WorkspaceArchiveWriteError(
                 path=root,
--- a/src/agents/extensions/sandbox/cloudflare/sandbox.py
+++ b/src/agents/extensions/sandbox/cloudflare/sandbox.py
@@ -1161,7 +1161,10 @@ class CloudflareSandboxSession(BaseSandboxSession):
             raise WorkspaceArchiveWriteError(path=root, context={"reason": "non_bytes_payload"})
 
         try:
-            validate_tar_bytes(bytes(raw))
+            validate_tar_bytes(
+                bytes(raw),
+                allow_external_symlink_targets=False,
+            )
         except UnsafeTarMemberError as e:
             raise WorkspaceArchiveWriteError(
                 path=root,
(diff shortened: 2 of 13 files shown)


seratch:
@codex review

PR #3096, opened 2026-05-03
Reject string-like values for shell action commands

mayunlei:
Summary:
- reject str/bytes/bytearray values for shell action commands in coerce_shell_call
- raise ModelBehaviorError instead of iterating character-by-character
- add regression coverage for string-like command payloads

Testing:
- uv run pytest tests/test_shell_call_serialization.py -q


seratch:
Closing this in favor of https://github.com/openai/openai-agents-python/pull/3092

PR #3098, opened 2026-05-03
test: cover on_handoff, on_tool_start, on_tool_end on RunHooks

Quratulain-bilal:
Adds test coverage for on_handoff, on_tool_start, and on_tool_end on run-level RunHooks (both Runner.run and            
  Runner.run_streamed). These were defined in RunHooksForTests but never asserted.


seratch:
@codex review

PR #3100, opened 2026-05-04
test: cover real Handoff object branch in visualization

Quratulain-bilal:
The existing visualization tests use Mock(spec=Handoff) for Handoff branches and real Agent objects for 
  Agent branches. The isinstance(handoff, Handoff) paths in get_all_nodes / get_all_edges were never      
  exercised against a real Handoff produced by the public handoff() factory. Adds a test that builds an
  Agent with a real Handoff and asserts the resulting DOT contains the expected node and edge.

### DIFF
--- a/tests/test_visualization.py
+++ b/tests/test_visualization.py
@@ -3,7 +3,7 @@ from unittest.mock import Mock
 import graphviz  # type: ignore
 import pytest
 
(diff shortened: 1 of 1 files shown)


seratch:
@codex review

seratch (approved):

PR #3111, opened 2026-05-04
fix: redact function tool trace span errors

### LINKED ISSUE #3110: Function tool trace errors ignore trace_include_sensitive_data=False
### Please read this first

### Describe the bug

Function tool failures attach `str(e)` directly to the current function span error even when `RunConfig(trace_include_sensitive_data=False)` is set.

In `src/agents/run_internal/tool_execution.py`, `_run_single_tool()` catches exceptions and calls `_error_tracing.attach_error_to_current_span()` with `SpanError(data={"tool_name": func_tool.name, "error": str(e)})`. The same module already has `get_trace_tool_error()`, and other tool execution paths such as computer and shell actions already use that helper to redact error details when sensitive tracing is disabled.

If a function tool exception contains an API key, SQL query, PII, tenant data, or other sensitive business data, that text can be exported in trace error data despite the run config opting out of sensitive trace data.

### Debug information

- Agents SDK version: `main` at `3854c124cb8e3e51fb660f5714405ee39ee86c5e`
- Python version: Python 3.12

### Repro steps

Minimal reproducer:

```python
import pytest

from agents import Agent, RunConfig, Runner, UserError, function_tool, trace
from tests.fake_model import FakeModel
from tests.test_responses import get_function_tool_call
from tests.testing_processor import SPAN_PROCESSOR_TESTING


@pytest.mark.asyncio
async def test_function_tool_trace_leaks_error_text():
    async def _error_tool() -> str:
        raise ValueError("secret-token-123")

    tool = function_tool(
        _error_tool,
        name_override="error_tool",
        failure_error_function=None,
    )
    model = FakeModel(tracing_enabled=True)
    model.set_next_output([get_function_tool_call("error_tool", "{}", call_id="1")])
    agent = Agent(name="test", model=model, tools=[tool])

    with trace("test"):
        with pytest.raises(UserError):
            await Runner.run(
                agent,
                input="hello",
                run_config=RunConfig(trace_include_sensitive_data=False),
            )

    for span in SPAN_PROCESSOR_TESTING.get_ordered_spans(including_empty=True):
        exported = span.export()
        if not exported:
            continue
        span_data = exported.get("span_data")
        if isinstance(span_data, dict) and span_data.get("type") == "function":
            print(exported["error"])
```

Current behavior on `main`: the function span error includes the raw exception message.

```text
{
  "message": "Error running tool",
  "data": {
    "tool_name": "error_tool",
    "error": "secret-token-123"
  }
}
```

### Expected behavior

Function tool trace errors should use the same redaction helper as the other tool execution paths. When `trace_include_sensitive_data=False`, the exported span error should contain the redacted tool error message rather than `str(e)`.

Expected span error:

```text
{
  "message": "Error running tool",
  "data": {
    "tool_name": "error_tool",
    "error": "Tool execution failed. Error details are redacted."
  }
}
```

Aphroq:
### Summary

- redact function tool trace error details when `trace_include_sensitive_data=False`
- add regression coverage for function tool exception spans under sensitive-data redaction

### Test plan

- `uv run pytest tests/test_run_step_execution.py::test_function_tool_error_trace_respects_sensitive_data_setting -q`

### Issue number

Closes #3110

### Checks

### DIFF
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -1570,10 +1570,14 @@ class _FunctionToolBatchExecutor:
                         agent_hooks=agent_hooks,
                     )
             except Exception as e:
+                trace_error = get_trace_tool_error(
+                    trace_include_sensitive_data=self.config.trace_include_sensitive_data,
+                    error_message=str(e),
+                )
                 _error_tracing.attach_error_to_current_span(
                     SpanError(
                         message="Error running tool",
-                        data={"tool_name": func_tool.name, "error": str(e)},
+                        data={"tool_name": func_tool.name, "error": trace_error},
                     )
                 )
                 if isinstance(e, AgentsException):
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -87,6 +87,7 @@ from ..tool_guardrails import (
 from ..tracing import Span, SpanError, function_span, get_current_trace
 from ..util import _coro, _error_tracing
 from ..util._approvals import evaluate_needs_approval_setting
+from ..util._tool_errors import get_trace_tool_error
 from ..util._types import MaybeAwaitable
 from ._asyncio_progress import get_function_tool_task_progress_deadline
 from .agent_bindings import AgentBindings, bind_public_agent
@@ -152,7 +153,6 @@ __all__ = [
     "execute_approved_tools",
 ]
 
-REDACTED_TOOL_ERROR_MESSAGE = "Tool execution failed. Error details are redacted."
 TToolSpanResult = TypeVar("TToolSpanResult")
 _FUNCTION_TOOL_CANCELLED_DRAIN_SECONDS = 0.25
 _FUNCTION_TOOL_CANCELLED_IMMEDIATE_STEP_LIMIT = 64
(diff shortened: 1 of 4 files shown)


seratch:
@codex review

seratch (changes requested):
Thanks for the fix. The direction looks right, but I think this still needs one more redaction path before we merge.

This PR covers the `failure_error_function=None` path where the tool exception is re-raised, but the default/handled function-tool error path can still put the raw exception message into the function span error even when `RunConfig(trace_include_sensitive_data=False)` is set.

I verified this on the PR head with a small probe:

- define a normal `@function_tool` without `failure_error_function=None`
- have it raise `ValueError("secret-token-123")`
- run with `RunConfig(trace_include_sensitive_data=False)`
- inspect the exported function span

The exported span still includes:

```json
{
  "message": "Error running tool (non-fatal)",
  "data": {
    "tool_name": "error_tool",
    "error": "secret-token-123"
  }
}
```

The leaking path appears to be the handled-error reporter in `src/agents/tool.py`, which currently stores `str(error)` in `SpanError.data["error"]` without checking the active run config's `trace_include_sensitive_data` value.

Could you update that path as well and add a regression test for the default `@function_tool` error handling case, not only the `failure_error_function=None` case?

Aphroq:
Thanks for the detailed catch. Agreed — I’ll add regression coverage for this path and update the implementation accordingly.

Aphroq:
Updated based on the review feedback.

This now covers the default `@function_tool` handled-error path, so function span errors no longer include the raw exception text when `RunConfig(trace_include_sensitive_data=False)` is used. I also audited the related tool failure span writers and added redaction for the handled cancellation path; the other tool action paths were already using the shared redaction helper.

Added/updated regression coverage for:
- re-raised tool errors with `failure_error_function=None`
- default `@function_tool` handled errors
- handled cancellation errors

Re-ran the full local verification successfully:

`bash .agents/skills/code-change-verification/scripts/run.sh`

seratch:
@codex review

seratch (approved):

PR #3112, opened 2026-05-04
fix: emit integer chat stream output indexes

### LINKED ISSUE #3109: Chat Completions streaming computes assistant output_index with a bool expression
### Please read this first

### Describe the bug

The Chat Completions streaming handler sometimes passes a boolean value as `output_index` for assistant message events after a reasoning item has been emitted.

Several event constructors use this expression:

```python
output_index=state.reasoning_content_index_and_output is not None
```

When reasoning content exists, this produces `True` rather than the integer `1`. Python treats `bool` as a subclass of `int`, so this can pass local numeric checks, but downstream JSON consumers and event snapshots can observe the wrong type.

Relevant code:

- `src/agents/models/chatcmpl_stream_handler.py`: text, refusal, content-part-done, and output-item-done events use the boolean expression above.

### Debug information

- Agents SDK version: current `main` at `3854c124`

### Repro steps

Stream Chat Completions chunks where reasoning content is emitted before regular text content. Inspect the `output_index` on the text/message events that follow the reasoning item.

Current behavior: some events use `True` as `output_index`.

### Expected behavior

All streamed response events should use integer `output_index` values, for example `0` for the reasoning item and `1` for the assistant message that follows it. No event should expose a boolean `output_index`.

Aphroq:
### Summary

Fixes Chat Completions streaming events so assistant message `output_index` values are emitted as integers instead of booleans when reasoning content precedes the message.

The stream handler now uses a shared helper to compute the assistant message output index and the reasoning-content regression test asserts that text/message events use `1` with exact `int` type, not `True`.

### Test plan

- `uv run pytest tests/models/test_reasoning_content.py -k 'stream_response_yields_events_for_reasoning_content'`

### Issue number

Closes #3109

### Checks

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -84,6 +84,10 @@ class SequenceNumber:
 
 
 class ChatCmplStreamHandler:
+    @staticmethod
+    def _assistant_message_output_index(state: StreamingState) -> int:
+        return 1 if state.reasoning_content_index_and_output is not None else 0
+
     @classmethod
     def _finish_reasoning_summary_part(
         cls,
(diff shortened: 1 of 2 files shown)


Aphroq:
Closing as duplicate. #3108 now contains both Chat Completions streaming output_index fixes

PR #3113, opened 2026-05-04
fix: emit integer output_index in Chat Completions streaming events

### LINKED ISSUE #3109: Chat Completions streaming computes assistant output_index with a bool expression

Quratulain-bilal:
Fixes #3109. The Chat Completions streaming handler computed output_index as                            
  state.reasoning_content_index_and_output is not None, yielding bool instead of int across 7 event sites.
   Replaced each with 1 if state.reasoning_content_index_and_output else 0 (the existing correct pattern
  at line 426). All 396 model tests pass; ruff + format clean.

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -341,16 +341,14 @@ class ChatCmplStreamHandler:
                     # Notify consumers of the start of a new output message + first content part
                     yield ResponseOutputItemAddedEvent(
                         item=assistant_item,
-                        output_index=state.reasoning_content_index_and_output
-                        is not None,  # fixed 0 -> 0 or 1
+                        output_index=(1 if state.reasoning_content_index_and_output else 0),
                         type="response.output_item.added",
                         sequence_number=sequence_number.get_and_increment(),
                     )
                     yield ResponseContentPartAddedEvent(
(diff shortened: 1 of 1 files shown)


seratch:
https://github.com/openai/openai-agents-python/pull/3108 covers more scenarios, so let us close this one.

PR #3125, opened 2026-05-05
test: add direct unit tests for run_grouping helpers

Quratulain-bilal:
run_internal/run_grouping.py exposes the priority hierarchy used to derive a stable grouping value for prompt-cache keys
   and trace grouping (conversation > session > group > generated run id), but had no direct test file — only indirect
  coverage via runner integration tests. This adds 23 focused unit tests covering all priority levels, whitespace
  stripping, empty-string fallthrough at each level, the documented attribute-access exception swallowing in
  get_session_id_if_available, and the run- prefix applied only to the generated kind.

### DIFF
--- a/tests/test_run_grouping.py
+++ b/tests/test_run_grouping.py
@@ -1,0 +1,204 @@
+"""Unit tests for src/agents/run_internal/run_grouping.py pure helpers.
+
+These cover the small, pure functions in `run_grouping` that resolve the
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sending this. It seems existing test assets already cover the same scenarios, so let us close this one. Thanks again for your interest.

PR #3126, opened 2026-05-05
Codex like compaction support

Flurin17:
Hi Everyone
I would like to have Codex like compaction support where not the amount of turns defines when compaction happens but the amount of tokens, which makes way more sense in my opinion. Let me know what should improved!


seratch:
Thanks for sharing the contribution idea!

I don't think this is the right abstraction for this SDK. `usage.input_tokens` here is usage for the latest model response/request, not cumulative usage across multiple turns. That can be useful as a signal for current context pressure, but it is not a session-level token counter. If we expose this option, the name and docs would need to be explicit that it means "latest response input tokens", not total session usage.

Also, for token-threshold based compaction, the Responses API now supports server-side compaction via `context_management` and `compact_threshold` on `responses.create`: https://developers.openai.com/api/docs/guides/compaction


In this SDK, that can already be passed through `ModelSettings(extra_args={...})`:

```python
from agents import Agent, ModelSettings

agent = Agent(
    name="Assistant",
    model_settings=ModelSettings(
        store=False,
        extra_args={
            "context_management": [
                {"type": "compaction", "compact_threshold": 200000}
            ],
        },
    ),
)
```

For explicit standalone compaction, the SDK already has `OpenAIResponsesCompactionSession`, which wraps a session and calls `responses.compact`.

Given that, I am going to close this PR for now rather than add another threshold option to `OpenAIResponsesCompactionSession`. We may add a first-class `ModelSettings` option for server-side `context_management` to make that API easier to use from the SDK.

PR #3130, opened 2026-05-05
docs+test: clarify response_include comment and add context_management resolve tests

richardogundele:
## Summary
- Replace stale TODO in `ModelSettings.response_include` with a clear note that
  `| str` is intentional for forward compatibility with future server-supported
  include flags. Wording now matches the rationale already in `openai_responses.py`.
- Add two tests pinning `resolve()` behaviour for the `context_management` field
  (introduced in #3128): override wins when both sides set it; base is preserved
  when override omits it.

## Test plan
- [x] `python -m pytest tests/model_settings/test_serialization.py` — 15 passed
- [x] `python -m pytest tests/test_source_compat_constructors.py` — 14 passed

## Related
Follows up on #3128 (feat: add context management model setting).

### DIFF
--- a/src/agents/model_settings.py
+++ b/src/agents/model_settings.py
@@ -132,9 +132,8 @@ class ModelSettings:
     """Whether to include usage chunk.
     Only available for Chat Completions API."""
 
-    # TODO: revisit ResponseIncludable | str if ResponseIncludable covers more cases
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for suggesting these changes and tests. I don't think we should have them this time, but we appreciate your interest here.

PR #3138, opened 2026-05-06
fix(realtime): send safe output guardrail feedback

### LINKED ISSUE #1912: Python Realtime Voice Agent doesn’t handle output guardrails like Node.js SDK

adityasingh2400:
## Summary
- Replaces the internal realtime output-guardrail follow-up prompt with a user-facing safe recovery instruction.
- Includes failed guardrail names and serialized output info so the model can recover without speaking the blocked content.
- Keeps the existing forced interrupt/cancel behavior and adds coverage that the interrupt is forced before the feedback message is sent.

Fixes #1912

## Test plan
- `uv run ruff format src/agents/realtime/session.py tests/realtime/test_session.py`
- `uv run ruff check src/agents/realtime/session.py tests/realtime/test_session.py`
- `uv run pytest tests/realtime/test_session.py -k guardrail -q`
- `uv run pytest tests/realtime/test_session.py -q`
- `uv run mypy src/agents/realtime/session.py tests/realtime/test_session.py`
- `make lint`

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -13,6 +13,7 @@ from typing_extensions import assert_never
 from .._tool_identity import get_function_tool_lookup_key_for_tool
 from ..agent import Agent
 from ..exceptions import UserError
+from ..guardrail import OutputGuardrailResult
 from ..handoffs import Handoff
 from ..items import ToolApprovalItem
 from ..logger import logger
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch:
Even if the issues pointed out by codex are resolved, we don't plan to have this change. Thanks for your interest here.

PR #3141, opened 2026-05-06
fix: tolerate audio deltas before audio format negotiation in ModelAu…

Quratulain-bilal:
ModelAudioTracker.on_audio_delta reads self._format, which is only set by set_audio_format. For transcription-only      
  sessions or session payloads where _extract_audio_format returns None (src/agents/realtime/openai_realtime.py:1236),    
  set_audio_format is never called and the first audio delta raises AttributeError: 'ModelAudioTracker' object has no
  attribute '_format'. The downstream calculate_audio_length_ms already accepts RealtimeAudioFormat | None and falls back
  to PCM math for unknown formats — fix is to initialize self._format to None in __init__. Added a regression test.


seratch:
@codex review

seratch (approved):

PR #3142, opened 2026-05-06
docs: explain function tool pre-execution guardrails

### LINKED ISSUE #2970: Add pre-execution validation for tool calls
Agents in this framework can call tools and trigger real-world actions.

Right now, tool execution appears to rely on upstream logic and model output, but there doesn’t seem to be a deterministic validation step at the moment of execution.

This creates a gap where:
- tool calls may be malformed or partially specified  
- parameters may differ from what was intended upstream  
- execution can occur without a final verification of the exact payload  

Logging, guardrails, and prompt-level controls help guide behavior, but they don’t guarantee that the final executed action is explicitly validated at the execution boundary.

### Suggestion

Introduce a lightweight validation step before any tool is executed that verifies:
- tool name  
- parameters (schema + expected values)  
- calling agent / context  
- destination or target system  
- timestamp / validity window  
- optional nonce or replay protection  

If validation fails → the tool call is rejected before execution.

This could be implemented as a wrapper around tool execution so existing agent logic remains unchanged.

Happy to share a reference implementation pattern if useful.

adityasingh2400:
## Summary
- Document `tool_input_guardrails` as the current pre-execution validation hook for local function tools.
- Include a concrete example for rejecting or halting a tool call based on the final serialized arguments.
- Clarify that Python-side guardrails cannot intercept hosted tools because those execute server-side.

## Test plan
- `uv run mkdocs build`
- `git diff --check`

Note: `uv run mkdocs build --strict` currently aborts on pre-existing translated-doc/autoref warnings unrelated to this change.

Closes #2970.

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -456,6 +456,54 @@ def score_b(score: Annotated[int, Field(..., ge=0, le=100, description="Score fr
     return f"Score recorded: {score}"
 ```
 
+### Function tool guardrails
+
+For local function tools, tool input guardrails are the pre-execution validation hook. They run after the model emits a concrete tool call and before the Python function is invoked, so they can inspect the exact serialized arguments, current tool context, and calling agent.
+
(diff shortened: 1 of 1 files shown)


seratch:
This is already covered by the guardrails document page.

PR #3144, opened 2026-05-06
docs: describe long-running tool progress options

adityasingh2400:
## Summary
- Document the current limitation that function tools return one final output and do not stream custom intermediate tool-progress events into `Runner.run_streamed()`.
- Point users to current options: nested agent `on_stream`, app-owned progress channels, or durable workflow integrations.

## Test plan
- `uv run mkdocs build`
- `git diff --check`

Related to #1333.


seratch:
Thinking of sandbox agent direction, this is not the thing we'd like to encourage.

PR #3145, opened 2026-05-06
docs: clarify tool-agent input context

adityasingh2400:
## Summary
- Clarify that `Agent.as_tool()` nested runs receive the tool-call input, not the outer agent transcript automatically.
- Point users who need richer specialist context to explicit tool arguments, `input_builder`, or direct `Runner.run` inside a custom function tool.

## Test plan
- `uv run mkdocs build`
- `git diff --check`

Related to #858.


seratch:
Thanks for the suggestion. However, adding this sentence there is not relevant and it is actually incorrect.

Issue #489, opened 2025-04-13
Automatic Audio Feedback on Silence in VoicePipeline


joowon-dm-snu:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have had similar requests

### Describe the feature
While working on the VoicePipeline—as with my previous issue #488 —I’ve identified a scenario where the voice agent may become unresponsive due to inadequate speech recognition. 
In practice, even if a user speaks, there are instances when the transcription session (`transcription_session.transcribe_turns()`) does not get triggered, resulting in an extended period of silence.

To mitigate this issue, I propose that when a turn concludes, a separate thread should be initiated. Within this thread, if no new turn is started within 3 seconds, the system should automatically stream a fallback message, such as:
> "Excuse me, it seems the audio was interrupted. Could you please repeat that?"

This feature would provide a smoother conversational experience by proactively addressing potential audio dropouts or misrecognitions.

joowon-dm-snu:
It seems possible to use the [SingleAgentWorkflowCallbacks](https://github.com/openai/openai-agents-python/blob/d0693a192e92057bfae7cc9d621a6a68e76a111b/src/agents/voice/workflow.py#L48) feature for this, but since the turn management is handled on the [Pipeline side](https://github.com/openai/openai-agents-python/blob/d0693a192e92057bfae7cc9d621a6a68e76a111b/src/agents/voice/pipeline.py#L135), I thought it would be more appropriate to modify the VoicePipeline instead.

seratch:
Thanks for the suggestion. The current VoicePipeline does not receive a signal that distinguishes intentional silence from audio that VAD or STT failed to recognize; it only runs the workflow when transcribe_turns() yields text. A generic three-second task would therefore also prompt during normal silence, while its timing, wording, cancellation, and overlap behavior are application-owned conversation policy.

For server-managed conversational inactivity handling, Realtime sessions now support server VAD idle_timeout_ms and emit input_audio_timeout_triggered. VoicePipeline applications can instead implement the timeout from their own microphone or VAD state and play the fallback through their existing audio path.

I am going to close this issue. We can revisit a narrower SDK gap if there is a reproducible supported VoicePipeline scenario that cannot be handled at either of those layers.

eljokerlover2020-maker:
Security strong to my account

eljokerlover2020-maker:
Protect myself

PR #3146, opened 2026-05-06
docs(voice): describe silence handling ownership

adityasingh2400:
## Summary
- Document that `VoicePipeline` only runs workflows after STT emits a transcribed turn.
- Clarify that automatic silence or "please repeat that" prompts should be handled in the app's microphone/input layer feeding `StreamedAudioInput`.

## Test plan
- `uv run mkdocs build`
- `git diff --check`

Related to #489.


seratch:
Thanks for the suggestion, but for such a use case, we currently recommend realtime agents instead.

PR #3152, opened 2026-05-06
feat: #1859 add runtime function tool concurrency config

### LINKED ISSUE #1859: Add max parallel_tool_calls value in Agents ModelSettings

seratch:
This pull request resolves #1859 by adding SDK-side runtime configuration for local function tool execution concurrency without changing `ModelSettings`.

It introduces `ToolExecutionConfig(max_function_tool_concurrency=...)` on `RunConfig`, preserves default behavior when unset, validates invalid concurrency values, and updates the function tool batch executor to queue local function calls while preserving model-order outputs. The change keeps provider-side `ModelSettings.parallel_tool_calls` separate from SDK-side local execution scheduling.

### DIFF
--- a/src/agents/__init__.py
+++ b/src/agents/__init__.py
@@ -109,6 +109,7 @@ from .run import (
     Runner,
     ToolErrorFormatter,
     ToolErrorFormatterArgs,
+    ToolExecutionConfig,
 )
 from .run_context import AgentHookContext, RunContextWrapper, TContext
 from .run_error_handlers import (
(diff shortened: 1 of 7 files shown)


seratch:
@codex review

PR #3155, opened 2026-05-06
feat(tracing): accept metadata kwarg on all span factories

### LINKED ISSUE #1844: Customizing traces and spans: Ability to add metadata to spans
It would be great if we could customize span_data for any given span.

The metadata field only works for the top level trace, but child spans like `response_span` do not have any existing metadata (nor the ability to add such metadata) such as current_turn, agent_name etc that is useful for querying data.

I think a simple solution for now could be to use the AgentHooks to manipulate the current_span and add any metadata you like.

```
def on_llm_start(..., agent: str, ...):
    current_span = get_current_span() # response span
    current_span.span_data["custom_metadata"] = agent.name
```

To do this, all spans would need to start before the _start hooks are called and finish after the _end hooks are called.
This is not standardized currently. Agent span follows this pattern, while response span starts after the `on_llm_start` hook and ends before the `on_llm_end` hook, making it difficult to manipulate the span data.

srikalyan:
### Summary

Extend the optional `metadata` kwarg to every public `*_span()` factory and the matching SpanData subclasses that lacked it (function, generation, response, handoff, custom, guardrail, transcription, speech, speech_group, mcp_tools). Append the new parameter at the end of each signature to preserve positional compatibility.

The exporter already serializes `span.span_data.metadata` at the top-level `metadata` key, so no changes are needed in the export pipeline. Backfill the same short Args entry in every helper docstring that documents arguments, so the generated reference picks it up.

### Test plan

- Added a parametrized test in `tests/test_tracing.py` that round-trips a `metadata` dict through every public span factory (13 ids) and verifies it surfaces under the exported payload's top-level `metadata` field.
- Added a test that confirms `span.span_data.metadata` can be mutated after construction (e.g. via a direct span reference) and the new value is reflected in the export.
- Local verification:
  - `make format` — clean
  - `make lint` — clean
  - `uv run mypy src/agents/tracing/span_data.py src/agents/tracing/create.py tests/test_tracing.py` — clean
  - `uv run pytest tests/test_tracing.py tests/test_agent_tracing.py tests/test_responses_tracing.py tests/test_function_tool.py` — 105 passed

### Issue number

Partially addresses #1844. Does not Close #1844, because the active response_span during LLM hooks is still a separate runtime-ordering problem.

### Checks

### DIFF
--- a/src/agents/tracing/create.py
+++ b/src/agents/tracing/create.py
@@ -94,6 +94,7 @@ def agent_span(
     span_id: str | None = None,
     parent: Trace | Span[Any] | None = None,
     disabled: bool = False,
+    metadata: dict[str, Any] | None = None,
 ) -> Span[AgentSpanData]:
     """Create a new agent span. The span will not be started automatically, you should either do
     `with agent_span() ...` or call `span.start()` + `span.finish()` manually.
@@ -109,12 +110,20 @@ def agent_span(
         parent: The parent span or trace. If not provided, we will automatically use the current
             trace/span as the parent.
         disabled: If True, we will return a Span but the Span will not be recorded.
+        metadata: Optional dictionary of custom metadata to attach to the span. Surfaces under
+            the exported payload's top-level ``metadata`` field.
 
     Returns:
         The newly created agent span.
     """
     return get_trace_provider().create_span(
-        span_data=AgentSpanData(name=name, handoffs=handoffs, tools=tools, output_type=output_type),
+        span_data=AgentSpanData(
+            name=name,
+            handoffs=handoffs,
+            tools=tools,
+            output_type=output_type,
+            metadata=metadata,
+        ),
         span_id=span_id,
         parent=parent,
         disabled=disabled,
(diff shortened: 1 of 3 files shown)


seratch:
@codex review

srikalyan:
@codex review

seratch:
Thanks for sharing this. I don't have the bandwidth to check this suggestion in depth right now, but have you verified if this change works well with the production OpenAI production API endpoint for tracing? The API could reject some data, so verifying the code alone may not be enough for quality assurance.

srikalyan:
Thanks for the careful read — that's a fair concern given history (#2779 broke ingest with 400 Unknown parameter).

Short version: this PR doesn't change the wire payload. It adds constructor sugar to populate a metadata field that's already on the wire today.

Specifically:
- Top-level metadata on span payloads has been live since #2889 (2026-04-15, shipped in v0.16.0 + v0.16.1). Code path is SpanImpl.export() at src/agents/tracing/spans.py:392-398.
- Production already populates it: src/agents/run_internal/agent_runner_helpers.py:148-150 does span.span_data.metadata = {"usage": ...} for every non-turn/non-task span. That traffic already flows against api.openai.com/v1/traces/ingest without 400s.
- Existing CI test test_agent_span_metadata_exports_with_routing_metadata (tests/test_tracing.py:443) pins the exact wire shape this PR feeds metadata into.
- The OpenAI sanitizer (_sanitize_for_openai_tracing_api in src/agents/tracing/processors.py) only touches span_data.{input, output, usage}; top-level metadata is left intact intentionally.

So this PR exposes existing capability via constructor kwargs; the bytes on the wire are identical to span.span_data.metadata = ... post-construction (which production already does today).

Happy to add a roundtrip test that runs OpenAITracingExporter._sanitize_for_openai_tracing_api() over a metadata-populated payload if you want extra signal — or run a live ingest against my own API key. Let me know which would help most.

seratch:
Since this PR has been inactive for a while, let me close it for now. If the approach is still useful, happy to revisit with a new PR. Thanks again for the effort here!

PR #3159, opened 2026-05-06
Emit history updates for realtime transcript deltas

### LINKED ISSUE #2940: RealtimeSession does not emit `history_updated` for `transcript_delta`

mindbomber:
Fixes #2940

## Summary

- Emit `RealtimeHistoryUpdated` after `transcript_delta` mutates realtime history.
- Keep raw model event emission unchanged for lower-level streaming consumers.
- Add regression coverage that verifies transcript deltas update assistant audio transcript history and emit the transformed history event.

## Verification

- `PYTHONPATH=src python -m pytest tests/realtime/test_session.py -q` -> 73 passed
- `python -m ruff check src/agents/realtime/session.py tests/realtime/test_session.py` -> passed
- `git diff --check` -> passed
- AANA code-change guardrail -> pass / accept, AIx 1.0, no violations

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -313,6 +313,9 @@ class RealtimeSession(RealtimeModelListener):
                     content=[AssistantAudio(transcript=self._item_transcripts[item_id])],
                 ),
             )
+            await self._put_event(
+                RealtimeHistoryUpdated(info=self._event_info, history=self._history)
+            )
 
             # Check if we should run guardrails based on debounce threshold
             current_length = len(self._item_transcripts[item_id])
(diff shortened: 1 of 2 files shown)


seratch:
duplicate with https://github.com/openai/openai-agents-python/pull/3086

PR #3180, opened 2026-05-07
Fix streaming output index handling

kuishou68:
## Summary
- preserve integer `output_index` values when merging streamed chat completion message data
- replace truthiness-based ternaries with explicit `is not None` checks so index `0` is not lost
- keep the streaming handler tests focused on role/content and output index behavior

## Tests
- not run in this pulse; branch and diff were verified from local git state


seratch:
This is already resolved

PR #3183, opened 2026-05-07
fix: preserve zero output index in streaming handler

kuishou68:
Fixes a streaming handler edge case where `output_index=0` can be treated as missing because the value is tested by truthiness.

The change keeps zero as a valid index by checking `is not None` explicitly, preserving correct output item placement in streamed chat completion handling.\n\n## Verification\nVerified by inspecting the committed diff; change is limited to explicit None handling for output_index.


seratch:
@kuishou68 Could you stop sending this patch? Your change is no longer relevant (the latest commit already resolved the issue) and it has conflicts.

kuishou68:
> @kuishou68 Could you stop sending this patch? Your change is no longer relevant (the latest commit already resolved the issue) and it has conflicts.

OK👌🏻

PR #3187, opened 2026-05-08
fix: await cancelled output guardrail tasks on tripwire

Quratulain-bilal:
When an output guardrail trips, run_output_guardrails cancels its sibling tasks but never awaits them, unlike the input-guardrail variants which gather the cancelled tasks. This can leak pending tasks and emit "Task was destroyed but it is pending!" warnings when slow guardrails are still mid-flight at tripwire time. Mirror the input-guardrail pattern by awaiting the cancelled tasks before raising.


seratch:
@codex review

seratch (changes requested):
Please make the regression test deterministic. Right now `fast_tripwire` can return before `slow_then_observe_cancel` has reached its `await asyncio.sleep(5)`, so the slow task may be cancelled before its `CancelledError` handler is installed. In that schedule, `cancelled_observed` stays unset even though the production fix is correct.

A better shape would be to add a `slow_started = asyncio.Event()`, set it at the start of `slow_then_observe_cancel`, and have `fast_tripwire` `await slow_started.wait()` before returning the tripwire result. That ensures the test is actually proving that an in-flight sibling guardrail is awaited after cancellation.

Quratulain-bilal:
fixed in 9ecf9b9. Added a `slow_started = asyncio.Event()` that the slow guardrail sets before its `await  
  asyncio.sleep(5)`, and `fast_tripwire` now `await slow_started.wait()` before returning the tripwire. This guarantees   
  the slow task is parked on the sleep with its `CancelledError` handler installed before the cancel fires, so the test   
  now deterministically proves an in-flight sibling guardrail is awaited after cancellation.

seratch:
@codex review

seratch (approved):

Quratulain-bilal:
Thank you so much! I really enjoyed contributing to the OpenAI SDK project,
and seeing my PR merged motivated me even more. I’d truly appreciate it if
you could give me any opportunity to prove myself. I’m hardworking,
punctual, consistent, and always eager to learn and contribute more through
real work and collaboration.

PR #3185, opened 2026-05-07
fix: Responses context-management extra_args collision

alfozan:
### Summary

Fix a false duplicate-argument error when a Responses request parameter is supplied through `ModelSettings.extra_args` and the first-class request field is only present as OpenAI's `omit` sentinel.

### Test plan

- `make format`
- `make lint`
- `uv run pytest tests/models/test_openai_responses.py::test_build_response_create_kwargs_allows_extra_arg_when_explicit_arg_is_omitted tests/models/test_openai_responses.py::test_build_response_create_kwargs_rejects_duplicate_context_management_extra_args tests/models/test_openai_responses.py::test_build_response_create_kwargs_rejects_duplicate_extra_args_keys -q`
- `uv run mypy src/agents/models/openai_responses.py tests/models/test_openai_responses.py`
- `uv run pyright src/agents/models/openai_responses.py tests/models/test_openai_responses.py`
- `git diff --check`

### Issue number

N/A

### Checks

- [x] I've added new tests (if relevant)
- [ ] I've added/updated the relevant documentation
- [x] I've run `make lint` and `make format`
- [x] I've made sure tests pass where not blocked by unrelated local-only failures

PR #3192, opened 2026-05-08
fix: chat completions extra_args collision with omitted kwargs

SuperMarioYL:
### Summary

Mirrors the Responses-path fix from #3185 in the parallel chat-completions
path. `OpenAIChatCompletionsModel._fetch_response` builds `create_kwargs`
where many fields go through `_non_null_or_omit(...)` and become the
OpenAI `omit` sentinel when the user did not set them. The duplicate-key
check then incorrectly counted those omitted entries as collisions with
`ModelSettings.extra_args`, raising:

> `TypeError: chat.completions.create() got multiple values for keyword argument '<key>'`

even though the first-class field was never explicitly set. As a result,
users could not legitimately route fields like `reasoning_effort` through
`extra_args` while leaving `model_settings.reasoning` unset.

This patch ports the same one-clause filter introduced for Responses in
#3185: skip duplicate-detection on keys whose `create_kwargs` value is
the OpenAI `Omit`/`NotGiven` sentinel, and only raise when both the
first-class field and `extra_args` actually carry a value. The companion
`_is_openai_omitted_value` helper is defined locally for parity with
`openai_responses.py`.

### Test plan

- `make format`
- `make lint`
- `uv run mypy src/agents/models/openai_chatcompletions.py tests/models/test_openai_chatcompletions.py`
- `uv run pytest tests/models/test_openai_chatcompletions.py -q`
- New regression tests:
  - `tests/models/test_openai_chatcompletions.py::test_fetch_response_chat_completions_allows_extra_arg_when_explicit_arg_is_omitted` — supplies `extra_args={\"reasoning_effort\": \"high\"}` while `model_settings.reasoning` is unset; the underlying `chat.completions.create(**kwargs)` call now receives `reasoning_effort=\"high\"` instead of failing with a spurious duplicate-argument error.
  - `tests/models/test_openai_chatcompletions.py::test_fetch_response_chat_completions_rejects_duplicate_extra_args_keys` — supplies `temperature=0.5` together with `extra_args={\"temperature\": 0.7}`; the legitimate duplicate is still rejected with the original `TypeError`.

### Issue number

N/A — parity follow-up to #3185.

### Checks

- [x] I've made sure tests pass where not blocked by unrelated local-only failures

### DIFF
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -5,7 +5,7 @@ import time
 from collections.abc import AsyncIterator
 from typing import TYPE_CHECKING, Any, Literal, cast, overload
 
-from openai import AsyncOpenAI, AsyncStream, Omit, omit
+from openai import AsyncOpenAI, AsyncStream, NotGiven, Omit, omit
 from openai.types import ChatModel
 from openai.types.chat import ChatCompletion, ChatCompletionChunk, ChatCompletionMessage
 from openai.types.chat.chat_completion import Choice
@@ -43,6 +43,10 @@ from .reasoning_content_replay import ShouldReplayReasoningContent
 
 if TYPE_CHECKING:
     from ..model_settings import ModelSettings
+
+
+def _is_openai_omitted_value(value: Any) -> bool:
+    return isinstance(value, Omit | NotGiven)
 
 
 class OpenAIChatCompletionsModel(Model):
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (changes requested):
Can you fix the issue pointed out by Codex and add sufficient tests for covering the pattern?

SuperMarioYL:
Thanks for the review @seratch and @codex. You're right — the previous filter let `extra_args={"stream": True}` slip through on the non-streaming path because `get_response()` pins `create_kwargs["stream"]` to the OpenAI omit sentinel, which the duplicate-key check was treating as "no value present".

Fixed in 4a5f3822 by marking `stream` as a reserved key that always triggers duplicate-key detection, regardless of whether the SDK currently passes it as omit or as a real value. Other create_kwargs slots (e.g. `reasoning_effort`, `tool_choice`) keep the omit-aware behavior so legitimate fall-through via `extra_args` still works.

Added two regression tests in `tests/models/test_openai_chatcompletions.py`:
- `extra_args={"stream": True}` with `stream=False` (non-streaming get_response) → `TypeError`
- `extra_args={"stream": False}` with `stream=True` (streaming) → `TypeError`

Verified `make format`, `make lint`, focused `mypy`, and `pytest tests/models/test_openai_chatcompletions.py` all pass locally.

seratch:
@codex review

SuperMarioYL:
Gentle bump on this @seratch — fix pushed 2026-05-09, codex re-review came back clean ('Didn't find any major issues'). Happy to rebase against current `main` if useful. Let me know if there's anything else you'd like changed before merge.

PR #3195, opened 2026-05-08
fix(mcp): cache empty tools list instead of re-fetching every call

adityasingh2400:
## Summary

When `cache_tools_list=True` and the MCP server reports zero tools, the cached `list_tools` result is an empty list `[]`. The cache check used a truthy test:

```python
if self.cache_tools_list and not self._cache_dirty and self._tools_list:
```

Because an empty list is falsy in Python, this branch is **never** taken when the server returns no tools, and the SDK silently falls through to a fresh `session.list_tools()` round-trip on every call — defeating caching for the empty-tools case and adding latency to every agent step.

## Fix

Switch the check to `self._tools_list is not None` so an empty list is honored as a valid cached response. 3-line code change.

## Test

Added `test_server_caching_empty_tools_list` to `tests/mcp/test_caching.py`, which mirrors the existing `test_server_caching_works` but with `tools=[]`. It asserts that `list_tools()` is only called once, even after a second invocation. The test fails on `main` (`call_count == 2`) and passes with the fix.

All 209 existing tests in `tests/mcp/` continue to pass.

## Test plan

- [x] `uv run pytest tests/mcp/ -x` — 209 passed
- [x] `uv run ruff check src/agents/mcp/server.py tests/mcp/test_caching.py` — clean
- [x] `uv run ruff format` — clean

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -829,8 +829,10 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
         assert session is not None
 
         try:
-            # Return from cache if caching is enabled, we have tools, and the cache is not dirty
-            if self.cache_tools_list and not self._cache_dirty and self._tools_list:
+            # Return from cache if caching is enabled, we have tools, and the cache is not dirty.
+            # Use `is not None` so that an empty tools list is still treated as cached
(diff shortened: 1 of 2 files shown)


seratch:
I don't think a useful MCP server could return an empty list of tools. The change may make sense but I don't see strong reasons to have the change.

PR #3197, opened 2026-05-08
feat(lifecycle): expose tool_call_id to on_tool_start/on_tool_end hooks

### LINKED ISSUE #1849: Expose Tool Call ID to Lifecycle Hooks

adityasingh2400:
## Summary

Closes #1849.

Lifecycle hooks (`on_tool_start`, `on_tool_end`) now receive an optional `tool_call_id: str | None` keyword argument so subclasses can correlate parallel tool starts with their respective ends. Today, when a single turn fires multiple tool calls in parallel, hooks have no way to tell which `on_tool_end` belongs to which `on_tool_start` (the bare `Tool` argument is shared and the `context` is sometimes a plain `RunContextWrapper`, not a `ToolContext`).

## Backwards compatibility

The new parameter is keyword-only with a default of `None`. Existing subclasses whose overrides do not declare `tool_call_id` keep working: a small dispatch helper (`_invoke_tool_hook`) introspects each hook's signature and only forwards the new keyword when the override accepts it (`tool_call_id` named parameter or `**kwargs`). A regression test exercises a legacy hook signature to lock this behavior in.

## Test plan

- `tests/test_run_hooks.py::test_run_hooks_receive_tool_call_id_for_parallel_calls` — runs an agent with two parallel function tool calls (`call_a`, `call_b`) and asserts both ids are captured in `on_tool_start` and `on_tool_end`.
- `tests/test_run_hooks.py::test_legacy_tool_hook_signature_still_works` — verifies subclasses without the new keyword keep working.
- All existing hook suites pass: `tests/test_run_hooks.py`, `tests/test_agent_hooks.py`, `tests/test_global_hooks.py`, `tests/test_agent_llm_hooks.py`, `tests/test_computer_tool_lifecycle.py` (34/34 green).
- Full non-extension test suite green (2228 passed).

### DIFF
--- a/src/agents/lifecycle.py
+++ b/src/agents/lifecycle.py
@@ -1,3 +1,5 @@
+import inspect
+from collections.abc import Awaitable, Callable
 from typing import Any, Generic
 
 from typing_extensions import TypeVar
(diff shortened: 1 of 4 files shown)


seratch:
tool_context already has the call_Id

PR #3200, opened 2026-05-08
fix(models): allow extra_args when chat completions kwarg is omitted

adityasingh2400:
## Summary

Mirror of #3185 (Responses API fix) for the Chat Completions path.

When a user supplies a first-class request parameter (e.g. `temperature`, `top_p`, `max_tokens`, `metadata`, ...) through `ModelSettings.extra_args` *without* also setting the corresponding `ModelSettings` field, the chat completions path falsely raises:

```
TypeError: chat.completions.create() got multiple values for keyword argument 'temperature'
```

This happens because the field in `create_kwargs` resolves to OpenAI's `omit` sentinel (not absent), and the duplicate-key check uses a plain `set` intersection. The Responses API path had the same bug, fixed in #3185 by skipping keys whose first-class value is `Omit`/`NotGiven`. The same pattern still exists in `openai_chatcompletions.py::_fetch_response` — this PR applies the same fix here.

### Behavior change

- Before: `ModelSettings(extra_args={"temperature": 0.25})` raised TypeError unless `temperature` was also `None`-by-default normalization.
- After: the `extra_args` value passes through.
- Genuine collisions (e.g. `ModelSettings(temperature=0.5, extra_args={"temperature": 0.25})`) still raise TypeError as before.

## Test plan

- [x] `uv run pytest tests/models/test_openai_chatcompletions.py tests/models/test_openai_chatcompletions_stream.py tests/models/test_openai_chatcompletions_converter.py tests/models/test_kwargs_functionality.py` (59 passed)
- [x] `ruff check` + `ruff format --check` clean
- [x] New tests:
  - `test_fetch_response_allows_extra_arg_when_explicit_arg_is_omitted` — fails on main, passes after fix
  - `test_fetch_response_rejects_duplicate_extra_args_with_explicit_value` — guards regression on the genuine-collision case

## Issue number

N/A (parallel of #3185 for chat completions)

## Checks

- [x] I've added new tests
- [ ] I've added/updated the relevant documentation (none needed)
- [x] I've run `make lint` / `make format` equivalent (`ruff check` / `ruff format --check`)

### DIFF
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -5,7 +5,7 @@ import time
 from collections.abc import AsyncIterator
 from typing import TYPE_CHECKING, Any, Literal, cast, overload
 
-from openai import AsyncOpenAI, AsyncStream, Omit, omit
+from openai import AsyncOpenAI, AsyncStream, NotGiven, Omit, omit
 from openai.types import ChatModel
 from openai.types.chat import ChatCompletion, ChatCompletionChunk, ChatCompletionMessage
 from openai.types.chat.chat_completion import Choice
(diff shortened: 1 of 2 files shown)


seratch:
https://github.com/openai/openai-agents-python/pull/3192 already suggested this

PR #3201, opened 2026-05-08
fix(voice): stop AudioInput.to_base64() from mutating caller's buffer

adityasingh2400:
## Summary

`AudioInput.to_base64()` currently rebinds `self.buffer` to an int16 array when the input is float32:

```python
def to_base64(self) -> str:
    if self.buffer.dtype == np.float32:
        # convert to int16
        self.buffer = np.clip(self.buffer, -1.0, 1.0)
        self.buffer = (self.buffer * 32767).astype(np.int16)
    ...
```

This silently corrupts any reference the caller still holds to the original float32 buffer. A user who creates an `AudioInput`, calls `to_base64()` (e.g. for tracing), and then inspects `audio_input.buffer` again will find an int16 array where they put a float32 one — different dtype, different values.

The companion `_buffer_to_audio_file()` already does the conversion via a local rebinding (the parameter `buffer` is shadowed inside the function), so `to_audio_file()` does not have this bug. This PR aligns `to_base64()` with that pattern: do the int16 conversion into a local variable and leave `self.buffer` untouched.

## Changes

- `src/agents/voice/input.py`: rewrite `to_base64()` to convert into a local `int16_buffer` instead of mutating `self.buffer`.
- `tests/voice/test_input.py`: add a regression test asserting `audio_input.buffer.dtype` and contents are unchanged after `to_base64()`, and that calling it twice produces the same encoding.

Diff is 25 lines total.

## Test plan

- [x] `pytest tests/voice/` — all 31 tests pass
- [x] Regression test fails on `main`, passes with this change
- [x] `ruff check` clean

### DIFF
--- a/src/agents/voice/input.py
+++ b/src/agents/voice/input.py
@@ -62,13 +62,14 @@ class AudioInput:
     def to_base64(self) -> str:
         """Returns the audio data as a base64 encoded string."""
         if self.buffer.dtype == np.float32:
-            # convert to int16
-            self.buffer = np.clip(self.buffer, -1.0, 1.0)
-            self.buffer = (self.buffer * 32767).astype(np.int16)
-        elif self.buffer.dtype != np.int16:
+            # convert to int16 without mutating the caller's buffer
+            int16_buffer = (np.clip(self.buffer, -1.0, 1.0) * 32767).astype(np.int16)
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3202, opened 2026-05-08
fix(redis-session): preserve created_at across writes

adityasingh2400:
## Summary

`RedisSession.add_items` rewrites the session metadata hash with `mapping={"created_at": str(int(time.time())), ...}` on every call. Because `HSET` with a mapping overwrites every field, `created_at` is reset to the current timestamp on every subsequent write — making it indistinguishable from `updated_at` and useless as a session-creation timestamp.

This is the same class of metadata-integrity bug as #3173 (Dapr) and #3175 (encrypted-session counts), in a sibling backend.

## Fix

Write `session_id` with `HSET`, `created_at` with `HSETNX` (set only if not present), and `updated_at` with `HSET`. Reuse a single `now` value for both writes so `created_at == updated_at` on the very first call.

Diff: 15 net lines in `redis_session.py`, plus a focused regression test.

## Test plan

- [x] New regression test `test_add_items_preserves_created_at_metadata` fails on `main`, passes with the fix
- [x] Full `tests/extensions/memory/` suite passes locally (83 passed, 4 skipped)
- [x] Pre-existing Redis tests unchanged

### DIFF
--- a/src/agents/extensions/memory/redis_session.py
+++ b/src/agents/extensions/memory/redis_session.py
@@ -190,16 +190,11 @@ class RedisSession(SessionABC):
 
         async with self._lock:
             pipe = self._redis.pipeline()
+            now = str(int(time.time()))
 
-            # Set session metadata with current timestamp
-            pipe.hset(
-                self._session_key,
-                mapping={
-                    "session_id": self.session_id,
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3203, opened 2026-05-08
fix(tracing): preserve falsy outputs in FunctionSpanData export

adityasingh2400:
## Summary

`FunctionSpanData.export()` gated stringification with `if self.output`, which silently coerces legitimate falsy tool outputs to `None` before they reach the trace dashboard.

```python
# src/agents/tracing/span_data.py (current)
"output": str(self.output) if self.output else None,
```

So a tool that returns `False`, `0`, `0.0`, `""`, `[]`, or `{}` shows up in the trace as `None`, indistinguishable from a tool that returned nothing. This is a long-standing inconsistency: every other span type with an `output` field passes the value through unchanged. Only `FunctionSpanData` has the `if self.output` short-circuit.

The fix is one character: use `is not None` so only an actually-missing output renders as `None`.

```python
"output": str(self.output) if self.output is not None else None,
```

## Repro

```python
from agents.tracing.span_data import FunctionSpanData
for output in [False, 0, 0.0, "", [], {}, None, "real"]:
    print(output, "->", FunctionSpanData(name="t", input="x", output=output).export()["output"])
# Before: every falsy value -> None
# After:  False -> "False", 0 -> "0", "" -> "", [] -> "[]", {} -> "{}", None -> None, "real" -> "real"
```

## Test plan

- [x] New parametrized test in `tests/tracing/test_function_span_data.py` covers `False`, `0`, `0.0`, `""`, `[]`, `{}`, `None`, and a normal string.
- [x] Test fails on `main` (6/8 cases) and passes with the fix (8/8).
- [x] Full tracing test suite green: `pytest tests/tracing/ tests/test_tracing.py tests/test_trace_processor.py` -> 102 passed.
- [x] `ruff check` and `ruff format --check` clean on changed files.

### DIFF
--- a/src/agents/tracing/span_data.py
+++ b/src/agents/tracing/span_data.py
@@ -161,7 +161,7 @@ class FunctionSpanData(SpanData):
             "type": self.type,
             "name": self.name,
             "input": self.input,
-            "output": str(self.output) if self.output else None,
+            "output": str(self.output) if self.output is not None else None,
             "mcp_data": self.mcp_data,
         }
 
--- a/tests/tracing/test_function_span_data.py
+++ b/tests/tracing/test_function_span_data.py
@@ -1,0 +1,27 @@
+"""Regression tests for FunctionSpanData export.
+
+Falsy tool outputs (e.g. ``False``, ``0``, ``""``, ``[]``) must be preserved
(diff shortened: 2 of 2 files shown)


seratch:
I think the current behavior is fine

PR #3206, opened 2026-05-08
fix(sandbox): require https for daytona signed preview URLs

adityasingh2400:
## Summary

`DaytonaSandboxSession._resolve_exposed_port` accepted any URL returned by `create_signed_preview_url` — including `http://...` — and silently exposed it as a non-TLS endpoint (`tls=False`, port 80). Daytona signed preview URLs are always HTTPS, so a non-https scheme indicates either a backend bug or a tampered response. Either way, downstream code would then connect over plaintext to a host derived from that URL.

This change rejects any preview URL whose scheme is not `https`, raising `ExposedPortUnavailableError` with the existing `invalid_preview_url` detail (same path already used for malformed URLs).

This is in the same vein as #3094 / #3172 / #3177 — defensive narrowing of what a sandbox backend will trust from the wire.

## Diff

- `src/agents/extensions/sandbox/daytona/sandbox.py`: 4 lines changed (require https, drop the http fallback for `tls`/port).
- `tests/extensions/sandbox/test_daytona.py`: new `test_resolve_exposed_port_rejects_non_https_preview_url` covering the regression.

## Test plan

- [x] New unit test `test_resolve_exposed_port_rejects_non_https_preview_url` fails on `main` and passes after the fix.
- [x] Full `tests/extensions/sandbox/test_daytona.py` suite (54 tests) passes.

### DIFF
--- a/src/agents/extensions/sandbox/daytona/sandbox.py
+++ b/src/agents/extensions/sandbox/daytona/sandbox.py
@@ -343,8 +343,10 @@ class DaytonaSandboxSession(BaseSandboxSession):
             host = split.hostname
             if host is None:
                 raise ValueError("missing hostname")
-            port_value = split.port or (443 if split.scheme == "https" else 80)
-            return ExposedPortEndpoint(host=host, port=port_value, tls=split.scheme == "https")
+            if split.scheme != "https":
+                raise ValueError(f"non-https preview scheme: {split.scheme!r}")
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch:
This could make sense, but I don't see the necessity to do this on the client SDK side

PR #3207, opened 2026-05-08
fix(result): drop reasoning items orphaned by dropped tool calls

adityasingh2400:
## Summary

When `drop_orphan_function_calls` removes an unpaired tool call from the items being sent to the Responses API, any reasoning items that immediately precede it become dangling. Forwarding them triggers errors like:

```
Item 'rs_...' of type 'reasoning' was provided without its required following item
```

This is the symptom reported in #985 (CodeInterpreter + handoff with reasoning models) and reproduced in #2020 (multi-agent handoff with `gpt-5` reasoning).

The fix tracks which tool-call indexes are dropped during the existing pass and removes the reasoning items that were tied to them before returning. A reasoning item is considered tied to the next non-reasoning item; if that next item was dropped, the reasoning chain is now dangling and is stripped.

Lone reasoning items (e.g. server-managed conversations where the required following item lives on the server, or the ``omit`` reasoning policy that strips standalone reasoning trace inputs) are still preserved when no tool calls are dropped, so the existing tracker behavior is unchanged.

## Test plan

- [x] New regression test `test_drop_orphan_function_calls_drops_reasoning_preceding_dropped_tool_call` covers the dropped-orphan-tool-call -> dangling-reasoning-items chain (mirrors #2020).
- [x] New regression test `test_drop_orphan_function_calls_keeps_lone_reasoning_when_no_tool_calls_dropped` confirms server-managed conversations still keep standalone reasoning items.
- [x] Existing `tests/test_run_internal_items.py` (19 tests), `tests/test_agent_runner*.py`, `tests/test_handoff*.py`, and `tests/test_server_conversation_tracker.py` all pass.
- [x] Full `pytest tests/` sweep (excluding optional `realtime`/`voice`/`mcp`/`sandbox` extras): 3332 passed, 28 skipped.

### DIFF
--- a/src/agents/run_internal/items.py
+++ b/src/agents/run_internal/items.py
@@ -102,12 +102,16 @@ def drop_orphan_function_calls(
 ) -> list[TResponseInputItem]:
     """
     Remove tool call items that do not have corresponding outputs so resumptions or retries do not
-    replay stale tool calls.
+    replay stale tool calls. Reasoning items that immediately precede a tool call dropped by this
+    pass are also removed, since the Responses API rejects reasoning items that are not followed
+    by their associated model-emitted item (``Item 'rs_...' of type 'reasoning' was provided
+    without its required following item``).
     """
 
     completed_call_ids = _completed_call_ids_by_type(items)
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3208, opened 2026-05-08
fix(sandbox): reap killed subprocess on UnixLocal exec timeout

adityasingh2400:
## Summary

`UnixLocalSandboxSession._exec_internal` handles `asyncio.TimeoutError` by SIGKILL-ing the subprocess group, but never awaits the killed `asyncio.subprocess.Process`. Two issues fall out of that:

1. **Transport not torn down inside the failing call.** Because `proc.wait()` is never awaited, `proc.returncode` stays `None` and the underlying `SubprocessTransport` (pipes + child-watcher entry) hangs around until GC eventually finds it. Repeated timeouts queue up live transports for the lifetime of the event loop.
2. **Stale-pid signal.** `os.killpg(proc.pid, SIGKILL)` runs unconditionally. If the process happened to exit between `wait_for` cancellation and the kill, the pid may already have been reused, and we'd send `SIGKILL` to an unrelated process group. The `except Exception: pass` masks this.

The fix mirrors what `_terminate_pty_entry` already does for PTY entries (`if process.returncode is None ... ProcessLookupError suppressed`):

- Only signal when `proc.returncode is None` (and `proc.pid is not None`).
- Narrow the swallowed exception to `ProcessLookupError` instead of `Exception`.
- After the kill, `await asyncio.wait_for(proc.wait(), timeout=5.0)` so asyncio actually closes the transport before `ExecTimeoutError` is raised.

Two regression tests cover both behaviors:
- `test_timeout_reaps_subprocess_so_returncode_is_set` captures the asyncio Process used by `_exec_internal` and asserts `returncode` is populated after the timeout fires (fails on main: `returncode is None`).
- `test_timeout_skips_killpg_for_already_exited_pid` simulates the "process exited just before the timeout handler" race and asserts no `killpg` is issued (fails on main: a stale `killpg` is recorded).

## Test plan

- [x] `pytest tests/sandbox/test_unix_local.py -v` (all 10 pass)
- [x] `pytest tests/sandbox/` (765 passed, 19 skipped)
- [x] `ruff check` + `ruff format --check`
- [x] `pyright src/agents/sandbox/sandboxes/unix_local.py tests/sandbox/test_unix_local.py`
- [x] Both new tests fail on `main` (bug repros), pass with the fix

### DIFF
--- a/src/agents/sandbox/sandboxes/unix_local.py
+++ b/src/agents/sandbox/sandboxes/unix_local.py
@@ -242,11 +242,17 @@ class UnixLocalSandboxSession(BaseSandboxSession):
             try:
                 stdout, stderr = await asyncio.wait_for(proc.communicate(), timeout=timeout)
             except asyncio.TimeoutError as e:
-                try:
-                    # process tree cleanup
-                    os.killpg(proc.pid, signal.SIGKILL)
-                except Exception:
-                    pass
+                # Avoid signaling a pid that has already exited and may have been
+                # reused: a stale `os.killpg(proc.pid, ...)` could otherwise hit an
+                # unrelated process group on systems with rapid pid wraparound.
+                if proc.returncode is None and proc.pid is not None:
+                    with suppress(ProcessLookupError):
+                        os.killpg(proc.pid, signal.SIGKILL)
+                # Reap the killed subprocess so asyncio releases its transport
+                # (pipes, child-watcher entry); otherwise the SubprocessTransport
+                # can outlive this call and surface as a delayed GC warning.
+                with suppress(Exception):
+                    await asyncio.wait_for(proc.wait(), timeout=5.0)
                 raise ExecTimeoutError(command=command, timeout_s=timeout, cause=e) from e
         except ExecTimeoutError:
             raise
--- a/tests/sandbox/test_unix_local.py
+++ b/tests/sandbox/test_unix_local.py
@@ -1,14 +1,16 @@
 from __future__ import annotations
 
+import asyncio
+import os
 import signal
 from pathlib import Path
+from typing import Any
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
Can you fix the CI errors?

adityasingh2400:
Fixed CI - typecheck and 3.10 test pass locally now. Thanks for the heads up.

adityasingh2400:
CI is green now. I rebased onto current main and re-ran the checks locally: ruff format, ruff lint, and mypy all pass, and the 10 tests in tests/sandbox/test_unix_local.py pass. The earlier failures were the py3.10 typing compat and a strict mypy issue in the test, both fixed. The fix itself reaps the killed subprocess on a UnixLocal exec timeout so it does not leave a zombie. Ready for review when you have time.

adityasingh2400:
Re-verified today: the branch merges cleanly against main and the CI errors that prompted the change-request are fixed. The earlier failures were a py3.10 typing-compat issue and a strict mypy issue in the test, both resolved. ruff format, ruff lint, and mypy are clean locally, and the 10 tests in tests/sandbox/test_unix_local.py pass. The fix in _exec_internal in src/agents/sandbox/sandboxes/unix_local.py does two things on an exec timeout: it only sends os.killpg when proc.returncode is None and proc.pid is not None, suppressing ProcessLookupError, so it cannot signal an already-exited and possibly reused pid, and it then awaits proc.wait with a bounded timeout to reap the killed subprocess so asyncio releases the transport instead of leaving a delayed-GC warning. Ready for another look when you have time.

seratch:
Since this PR has been inactive for a while, let me close it for now. If the approach is still useful, happy to revisit with a new PR.

PR #3210, opened 2026-05-08
fix(exceptions): export MCPToolCancellationError from top-level package

adityasingh2400:
## Summary

`MCPToolCancellationError` is a public exception subclassing `AgentsException` (defined in `src/agents/exceptions.py` and raised from `src/agents/mcp/util.py`), but it was the only such class missing from the top-level `agents` package re-exports.

Every other `AgentsException` subclass (`MaxTurnsExceeded`, `ModelBehaviorError`, `ModelRefusalError`, `ToolTimeoutError`, `UserError`, all the guardrail tripwire exceptions, etc.) is importable as `from agents import X`. Users wanting to catch MCP tool cancellations had to fall back to `from agents.exceptions import MCPToolCancellationError`, breaking the convention.

This adds the import and `__all__` entry so the symbol is reachable from `agents` like its siblings.

## Changes

- `src/agents/__init__.py`: add `MCPToolCancellationError` to the `from .exceptions import (...)` block and to `__all__` (2 lines).
- `tests/test_exception_exports.py` (new, 30 lines): one direct assertion that `from agents import MCPToolCancellationError` works and the name appears in `agents.__all__`, plus a regression guard that walks every concrete `AgentsException` subclass in `agents.exceptions` and asserts each is re-exported. This prevents the same mistake from recurring if a new exception is added later.

## Test plan

- [x] `pytest tests/test_exception_exports.py` passes after the fix.
- [x] Both tests fail on `upstream/main` without the fix (verified via `git stash`).
- [x] `pytest tests/test_run_error_details.py tests/test_tool_context.py tests/mcp/test_mcp_imports.py` still passes (19 tests).
- [x] No public API change beyond surfacing an existing class.


seratch:
@codex review

PR #3211, opened 2026-05-08
fix(handoffs): await on_handoff callables with async __call__

adityasingh2400:
## Summary

`handoff(on_handoff=...)` accepts any callable, but the dispatch in
`src/agents/handoffs/__init__.py` decided whether to `await` based on
`inspect.iscoroutinefunction(on_handoff)`. That predicate returns `False` for
class instances whose `__call__` is `async def`, so for callbacks such as

```python
class MyHook:
    async def __call__(self, ctx, input):
        ...

handoff(agent, input_type=Foo, on_handoff=MyHook())
```

the runtime called `MyHook()(ctx, input)`, got a coroutine back, and discarded
it without awaiting. Result: the user's `on_handoff` never executed and Python
emitted `RuntimeWarning: coroutine '__call__' was never awaited`.

The fix mirrors the pattern already used elsewhere in the codebase (e.g.
`Handoff._is_enabled`, `tool.py`): invoke the callable, then `await` the
result iff it's awaitable. Sync callables remain synchronous; native async
functions and async-`__call__` instances are now awaited correctly.

## Test plan

- [x] New test `test_callable_class_with_async_dunder_call_is_awaited` covers
  both the `input_type`-bearing branch and the no-input branch.
- [x] `pytest tests/test_handoff_tool.py tests/test_handoff_history_duplication.py tests/test_handoff_prompt.py` (28 passed).
- [x] `ruff check` + `ruff format --check` clean on the touched files.
- [x] Confirmed the new test fails on `main` with the documented RuntimeWarning before applying the fix.

### DIFF
--- a/src/agents/handoffs/__init__.py
+++ b/src/agents/handoffs/__init__.py
@@ -291,16 +291,14 @@ def handoff(
                 partial=False,
             )
             input_func = cast(OnHandoffWithInput[THandoffInput], on_handoff)
-            if inspect.iscoroutinefunction(input_func):
-                await input_func(ctx, validated_input)
-            else:
-                input_func(ctx, validated_input)
+            result = input_func(ctx, validated_input)
+            if inspect.isawaitable(result):
+                await result
         elif on_handoff is not None:
             no_input_func = cast(OnHandoffWithoutInput, on_handoff)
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (approved):

Issue #2165, opened 2025-12-10
Handling Tool Execution errors in MCP with Agents SDK


KaranJayakumar:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) -> Yes
- **Have you searched for related issues?** Others may have had similar requests -> Yes

### Question
I have been trying to set up an MCP Server with Agents SDK. I've built the MCP server. When there are tool execution errors, the result object looks like : 
    {
          "content": [
               {
                        "type": "text",
                        "text": f"{str(e)}",
                    },
                ],
          "isError": True,
     }
. This seems to adhere to the MCP specification, however when attempting to detect when an error has occured while streaming events, I cannot find any references to the isError field. I am assuming that I would get some sort of indication that an error has occurred as a part of the ToolCallOutputItem but there doesn't seem to be any. Apologies if this is a silly question or if i'm on an older version. Would appreciate any assistance.

KaranJayakumar:
It seems i'm currently on an older version of agents SDK. Namely 0.3.3. Has there been changes that would support this since the release of 0.3.3?

KaranJayakumar:
It doesn't seem like this is resolved on the latest version as well. I must be missing something

seratch:
This SDK does not currently use the MCP SDK’s isError property. However, in some cases it could be useful to surface and leverage the richer error details returned by MCP servers.

That said, even with such an enhancement, the agent would still receive the MCP tool result as either a string or structured output. Your application would likely still need to determine whether the MCP response represents an error.

As long as proposed enhancements fit the SDK’s current design and don’t introduce breaking changes, we’re open to ideas for improvements and/or customization options to support this use case.

KaranJayakumar:
Oh cool I'll try and put up a PR then. Would be one of my first OSS contributions. Apologies if the review load is hell haha

PR #3212, opened 2026-05-08
fix(mcp): surface CallToolResult.isError as a tool error

adityasingh2400:
## Summary
- Per the MCP spec, a server signals tool-execution failure by returning a `CallToolResult` with `isError=True` rather than raising. The current `invoke_mcp_tool` ignores this field, so the model sees the error content as if it were a successful tool output and proceeds as if nothing went wrong.
- Surface `isError=True` through the FunctionTool failure pipeline by raising `AgentsException`, matching the existing handling for `McpError` transport failures. The default `failure_error_function` will format it into a model-visible tool error; callers using `failure_error_function=None` get the raised exception as documented.
- Refs #2165 (closed-stale, never fixed).

## Diff size
~22 lines in `src/agents/mcp/util.py`, ~52-line focused test in `tests/mcp/test_mcp_util.py`.

## Test plan
- [x] New test `test_mcp_invocation_is_error_result_surfaces_as_tool_error` fails on `main` (DID NOT RAISE) and passes with the fix
- [x] Full `tests/mcp/` suite passes (209 tests)
- [x] `ruff check` clean

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -658,6 +658,28 @@ class MCPUtil:
         else:
             logger.debug(f"MCP tool {tool_name_for_display} returned {result}")
 
+        # Per the MCP spec, a server signals tool-execution failure by returning a
+        # CallToolResult with isError=True (rather than raising). Surface this through
+        # the FunctionTool failure pipeline so the model sees a structured error
+        # result instead of a fake "successful" output.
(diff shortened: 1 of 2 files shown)


seratch:
This may be useful for some scenarios but having getattr does not sound great for long-term maintenance

PR #3214, opened 2026-05-08
fix(realtime): preserve existing transcript over stale delta accumulator

adityasingh2400:
## Summary

While merging a `RealtimeModelItemUpdatedEvent` whose incoming content lacks a transcript, `RealtimeSession` looked up the existing item to preserve its transcript, then unconditionally reassigned the local `preserved` variable from the in-flight `_item_transcripts` delta accumulator for assistant items.

That contradicted the adjacent comment ("If still missing and this is an assistant item, fall back to ...") and meant a stale or partial delta still buffered for the same `item_id` would clobber the complete transcript already in history. In practice this surfaced when the platform retrieves an item (e.g. after `conversation.item.truncated`) without a transcript while delta state from the same turn was still around.

The one-line fix adds the missing `not preserved` guard so the delta fallback only fires when the existing item really has nothing. The existing two preservation tests still pass; the new regression test fails on `main` and passes with the fix.

```diff
- if incoming_item.role == "assistant":
+ if not preserved and incoming_item.role == "assistant":
      preserved = self._item_transcripts.get(incoming_item.item_id)
```

## Test plan

- [x] `pytest tests/realtime/test_session.py::TestTranscriptPreservation` (3 passed, including new regression test)
- [x] `pytest tests/realtime/` (233 passed, no regressions)
- [x] Confirmed new test fails on `main` without the fix (`assert 'stale partial' == 'Final complete transcript'`)

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -362,7 +362,7 @@ class RealtimeSession(RealtimeModelListener):
 
                                 # If still missing and this is an assistant item, fall back to
                                 # accumulated transcript deltas tracked during the turn.
-                                if incoming_item.role == "assistant":
+                                if not preserved and incoming_item.role == "assistant":
                                     preserved = self._item_transcripts.get(incoming_item.item_id)
 
                                 if preserved:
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3215, opened 2026-05-08
fix(litellm): avoid duplicating content and signed thinking blocks across parallel tool-call splits

adityasingh2400:
## Summary

When `LitellmModel._fix_tool_message_ordering` splits an assistant message that has multiple `tool_calls` into one message per tool call, it copies the entire parent message into each split. That duplicates `content`, `thinking_blocks`, and `reasoning_content` across every split.

For Anthropic models with extended thinking, this is fatal: each thinking block carries a unique signature, and the API rejects requests that include the same signed block twice. It also corrupts conversation history by repeating the assistant's user-visible text whenever the model emits two or more parallel tool calls.

Keep those shared fields only on the first split; subsequent splits carry their `tool_calls` alone.

## Repro

Pass an assistant message with two parallel tool calls plus content/thinking blocks (which is what LiteLLM emits for Anthropic Claude models with parallel tool use + extended thinking) through `_fix_tool_message_ordering`. Before the fix, both split assistant messages contain the same `content` text and the same signed `thinking_blocks`; after the fix, only the first split carries them.

## Test plan

- [x] New unit test `test_split_does_not_duplicate_content_or_thinking` covers the regression.
- [x] All 9 pre-existing `test_extended_thinking_message_order.py` tests still pass.
- [x] All 25 LiteLLM-related tests pass.
- [x] `ruff check` clean on both files.

### DIFF
--- a/src/agents/extensions/models/litellm_model.py
+++ b/src/agents/extensions/models/litellm_model.py
@@ -661,13 +661,24 @@ class LitellmModel(Model):
                 # Extract tool calls from this assistant message
                 tool_calls = message.get("tool_calls", [])
                 if isinstance(tool_calls, list):
-                    for tool_call in tool_calls:
+                    for split_idx, tool_call in enumerate(tool_calls):
                         if isinstance(tool_call, dict):
                             tool_id = tool_call.get("id")
                             if tool_id:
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3216, opened 2026-05-08
fix(tracing): keep BatchTraceProcessor worker alive on exporter errors

adityasingh2400:
## Summary

If the user-supplied exporter raises (transient bug, network library throws, etc.), the exception escapes `BatchTraceProcessor._export_batches` and kills the background `_run` thread.

Once the worker dies, subsequent `on_span_end` / `on_trace_start` calls just enqueue items that are never exported — they sit in the queue until it hits `max_queue_size` (8192 by default) and are then silently dropped via `queue.Full`.

The fix wraps the `self._exporter.export(...)` call in a try/except and logs the failure non-fatally, so a single bad batch no longer takes the entire processor down.

## Repro (before this PR)

```python
class FlakyExporter(TracingExporter):
    def __init__(self): self.n = 0
    def export(self, items):
        self.n += 1
        if self.n == 1:
            raise RuntimeError("boom")

p = BatchTraceProcessor(FlakyExporter(), schedule_delay=0.05, max_batch_size=1)
p.on_span_end(make_span()); p.on_span_end(make_span()); p.on_span_end(make_span())
time.sleep(0.3)
assert p._worker_thread.is_alive()  # FAILS on main; passes after this PR
```

## Test plan

- [x] New test `test_batch_trace_processor_survives_exporter_exception` fails on main, passes with the fix
- [x] All 46 existing tests in `tests/test_trace_processor.py` still pass
- [x] All 95 tracing tests (`tests/tracing/`, `tests/test_trace_processor.py`, `tests/test_tracing.py`) still pass
- [x] `ruff check` clean

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -600,8 +600,17 @@ class BatchTraceProcessor(TracingProcessor):
                 if not items_to_export:
                     break
 
-                # Export the batch
-                self._exporter.export(items_to_export)
+                # Export the batch. Catch any exception so a misbehaving exporter
+                # cannot kill the background worker thread and silently strand all
+                # subsequent spans in the queue.
+                try:
+                    self._exporter.export(items_to_export)
+                except Exception as exc:
+                    logger.error(
+                        "[non-fatal] Tracing: exporter raised %s; dropping batch of %d items",
+                        exc,
+                        len(items_to_export),
+                    )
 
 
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (changes requested):
Can you fix the lint error?

adityasingh2400:
Lint fixed. Thanks!

seratch (approved):

PR #3217, opened 2026-05-08
fix(run): preserve failed status across apply_patch operations

adityasingh2400:
## Summary

`ApplyPatchAction.execute` walks through every operation in an `apply_patch` call and tracks an overall `status` for the resulting `apply_patch_call_output`. The previous logic was:

```python
if normalized.status in {"completed", "failed"}:
    status = normalized.status
```

This unconditionally overwrites the running status with the latest editor result. If a multi-op call had a failed operation followed by a successful one, the failure was silently lost: the raw_item reported `"completed"` even though one of the operations failed. That contradicts the exception path in the same method (which forces `"failed"`) and would feed misleading status back to the model.

This change makes failure sticky: once any operation fails, the overall status stays `"failed"`. A `"completed"` result can only set the status when no prior failure has been recorded.

## Test plan

- [x] New regression test `test_apply_patch_failed_status_not_overwritten_by_later_completed_op` exercises a 2-op call where op1 fails and op2 succeeds; it fails on `main` and passes after the fix.
- [x] Full `tests/test_apply_patch_tool.py` suite (11 tests) passes.
- [x] `tests/test_hitl_error_scenarios.py` (which exercises apply_patch HITL flows) still passes.

### DIFF
--- a/src/agents/run_internal/tool_actions.py
+++ b/src/agents/run_internal/tool_actions.py
@@ -827,8 +827,10 @@ class ApplyPatchAction:
                     awaited = await result if inspect.isawaitable(result) else result
                     normalized = normalize_apply_patch_result(awaited)
                     if normalized:
-                        if normalized.status in {"completed", "failed"}:
-                            status = normalized.status
+                        if normalized.status == "failed":
+                            status = "failed"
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

Issue #3174, opened 2026-05-07
EncryptedSession limit can return fewer valid items when latest envelopes are invalid


Aphroq:
### Please read this first

- **Have you read the docs?** Yes.
- **Have you searched for related issues?** Yes.

### Describe the bug

`EncryptedSession.get_items(limit=N)` applies `limit` to the underlying encrypted envelopes before it decrypts them. Invalid, expired, or otherwise undecryptable envelopes are then dropped, so a latest invalid envelope can hide earlier valid context even when enough valid items exist in the session history.

I searched existing issues and PRs for `EncryptedSession get_items limit`, `EncryptedSession invalid token limit`, and `encrypted session expired limit`, and did not find a direct match.

This is specific to `EncryptedSession` because it wraps another session backend and filters encrypted envelopes after reading them. The adjacent negative `limit` behavior is a separate session-limit consistency issue and is not part of this report.

### Debug information

- Agents SDK version: `v0.16.0` / current `main` at `eed9100`
- Python version: Python 3.12.1
- Backend: `EncryptedSession` wrapping `SQLiteSession`

### Repro steps

Run this script against an affected version:

```python
import asyncio

from cryptography.fernet import Fernet

from agents import SQLiteSession
from agents.extensions.memory.encrypt_session import EncryptedSession


async def main() -> None:
    session_id = "encrypted-limit-repro"
    underlying_session = SQLiteSession(session_id)
    encrypted_session = EncryptedSession(
        session_id=session_id,
        underlying_session=underlying_session,
        encryption_key=Fernet.generate_key().decode("utf-8"),
    )

    try:
        await encrypted_session.add_items([{"role": "user", "content": "older valid"}])
        await underlying_session.add_items(
            [{"__enc__": 1, "v": 1, "kid": "hkdf-v1", "payload": "not-a-valid-token"}]
        )

        all_items = await encrypted_session.get_items()
        limited_items = await encrypted_session.get_items(limit=1)

        print("all:", [item.get("content") for item in all_items])
        print("limit=1:", [item.get("content") for item in limited_items])

        assert [item.get("content") for item in all_items] == ["older valid"]
        assert [item.get("content") for item in limited_items] == ["older valid"]
    finally:
        underlying_session.close()


asyncio.run(main())
```

On affected versions, the output is:

```text
all: ['older valid']
limit=1: []
```

The first assertion passes, but the second assertion fails because `limited_items` is empty.

### Expected behavior

[...]

PR #3220, opened 2026-05-08
fix(encrypted-session): skip envelopes with malformed payloads

adityasingh2400:
## Summary

`EncryptedSession._unwrap` assumes `item["payload"]` is always a string and only catches `InvalidToken` / `KeyError`. If a stored envelope has a non-string payload (e.g. `None`, `int`, `list` — possible under storage corruption, type confusion at the DB layer, or manual injection), `payload.encode("utf-8")` raises `AttributeError` and breaks **both** `get_items` and `pop_item` for the entire session — there is no recovery short of `clear_session`.

This is the same shape of bug as #3174 (fixed in #3175 for invalid Fernet tokens), but for a different failure mode: malformed envelope shape rather than invalid ciphertext.

### Repro (before the fix)

```python
session = EncryptedSession(session_id="s", underlying_session=sqlite, encryption_key=key)
await session.add_items([{"role": "user", "content": "Hello"}])
await sqlite.add_items([{"__enc__": 1, "v": 1, "kid": "hkdf-v1", "payload": None}])

await session.get_items()  # AttributeError: 'NoneType' object has no attribute 'encode'
await session.pop_item()   # same crash on second pop
```

### Fix

- Validate `payload` is a `str` before calling `.encode("utf-8")`; otherwise return `None` (skip).
- Broaden the swallowed exception set to include `ValueError` (covers `json.JSONDecodeError` if the decrypted plaintext is not valid JSON) and `UnicodeDecodeError`.

Treats malformed envelopes the same way invalid/expired envelopes are already treated (silently skip), matching the documented behavior in the module docstring: "When TTL expires, expired items are silently skipped."

7-line source change.

## Test plan

- [x] Add parametrized regression test covering `int`, `None`, `list`, and `dict` payloads on `get_items` (4 cases).
- [x] Add regression test covering `pop_item` with a malformed envelope underneath a valid one — must return the valid item, then drain to `None` instead of raising.
- [x] All 24 tests in `tests/extensions/memory/test_encrypt_session.py` pass (19 existing + 5 new).
- [x] Confirmed new tests fail without the source fix (AttributeError).
- [x] `ruff check` clean on changed files.

### DIFF
--- a/src/agents/extensions/memory/encrypt_session.py
+++ b/src/agents/extensions/memory/encrypt_session.py
@@ -164,10 +164,13 @@ class EncryptedSession(SessionABC):
             return cast(TResponseInputItem, item)
 
         try:
-            token = item["payload"].encode("utf-8")
+            payload = item["payload"]
+            if not isinstance(payload, str):
+                return None
+            token = payload.encode("utf-8")
             plaintext = self.cipher.decrypt(token, ttl=self.ttl)
             return cast(TResponseInputItem, _from_json_bytes(plaintext))
-        except (InvalidToken, KeyError):
+        except (InvalidToken, KeyError, ValueError, UnicodeDecodeError):
             return None
 
     def _unwrap_valid_items(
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch:
I don't think this could usually happen

PR #3221, opened 2026-05-08
fix(computer): preserve data URL screenshots from driver

adityasingh2400:
## Summary
- `ComputerAction.execute` always wraps the driver's screenshot string in `data:image/png;base64,<...>`. When a `Computer`/`AsyncComputer` implementation already returns a fully-formed data URL (e.g. drivers that re-use a generic image helper, or vendor SDKs that hand back data URLs), the result is `data:image/png;base64,data:image/png;base64,<payload>`, which is not a valid URL and renders as nothing in the model output.
- Detect an existing `data:image/...` prefix and pass it through unchanged. Empty output still yields `""` for backwards compat.
- Logic moved into a small `_build_image_url` helper to keep the hot path readable and testable.

## Test plan
- [x] `pytest tests/test_computer_action.py tests/test_computer_tool_lifecycle.py` (37 passed)
- [x] New regression: `test_execute_does_not_double_prefix_data_url_screenshot` fails on `main` and passes with the fix.
- [x] `ruff check` on touched files.

### DIFF
--- a/src/agents/run_internal/tool_actions.py
+++ b/src/agents/run_internal/tool_actions.py
@@ -159,7 +159,7 @@ class ComputerAction:
                 ),
             )
 
-            image_url = f"data:image/png;base64,{output}" if output else ""
+            image_url = cls._build_image_url(output)
             if span and config.trace_include_sensitive_data:
                 span.span_data.output = image_url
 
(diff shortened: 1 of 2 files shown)


seratch:
I don't think this change is necessary

PR #3223, opened 2026-05-08
fix(realtime): expose max_output_tokens on RealtimeSessionModelSettings

adityasingh2400:
## Summary

`OpenAIRealtimeWebSocketModel._get_session_config` already reads `max_output_tokens` from the settings dict and forwards it to `RealtimeSessionCreateRequest.max_output_tokens` (see `src/agents/realtime/openai_realtime.py:1474`), but the field was never declared on `RealtimeSessionModelSettings`. Type-checked callers had to cast through `Any` (or use `# type: ignore`) just to limit per-response output tokens, even though the OpenAI Realtime API supports the field natively.

This patch adds `max_output_tokens: NotRequired[int | Literal["inf"]]` to the TypedDict so callers can pass either an integer cap or the `"inf"` sentinel directly. No runtime change is required: the existing forwarding logic already handles both shapes.

## Test plan

- [x] New `test_session_config_passes_max_output_tokens` covers integer caps, the `"inf"` sentinel, and the unset (server-default) case via `_get_session_config`.
- [x] `pytest tests/realtime/` -> 233 passed.
- [x] `ruff check` / `ruff format --check` clean on touched files.

### DIFF
--- a/src/agents/realtime/config.py
+++ b/src/agents/realtime/config.py
@@ -169,6 +169,13 @@ class RealtimeSessionModelSettings(TypedDict):
     speed: NotRequired[float]
     """The speed of the model's responses."""
 
+    max_output_tokens: NotRequired[int | Literal["inf"]]
+    """Maximum number of output tokens for a single assistant response, inclusive of tool calls.
+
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (approved):

PR #3227, opened 2026-05-08
fix(sessions): persist output_tokens_details when input details are None

adityasingh2400:
## Summary

In `AdvancedSQLiteSession._update_turn_usage_internal`, the `output_tokens_details` serialization branch was nested inside the `input_tokens_details` branch. This meant whenever `input_tokens_details` was falsy (e.g., providers that bypass validation via `model_construct` and populate only output details), `output_tokens_details` was silently dropped from `turn_usage` rows.

This is similar in spirit to the Redis `created_at` regression in #3202: a metadata field that should always round-trip is instead silently lost in a specific code path.

## Fix

De-nest the output branch so each detail field is serialized independently. The two `if` blocks are now siblings rather than parent/child.

## Test plan

- [x] Added `test_output_tokens_details_persisted_when_input_details_missing` covering the regression: a `Usage` with `input_tokens_details=None` and a populated `OutputTokensDetails(reasoning_tokens=42)`. Before the fix this asserted `None`; after the fix it persists `{"reasoning_tokens": 42}`.
- [x] All 36 existing tests in `test_advanced_sqlite_session.py` continue to pass.
- [x] `ruff check` and `ruff format --check` pass.

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -1322,17 +1322,15 @@ class AdvancedSQLiteSession(SQLiteSession):
                         self._logger.warning(f"Failed to serialize input tokens details: {e}")
                         input_details_json = None
 
-                    if (
-                        hasattr(usage_data, "output_tokens_details")
-                        and usage_data.output_tokens_details
-                    ):
-                        try:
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (approved):

PR #3228, opened 2026-05-08
fix(run): guard get_session_id_if_available against non-string session_id

adityasingh2400:
## Summary

`get_session_id_if_available` reads `session.session_id` from a third-party `Session` Protocol implementation, then unconditionally calls `.strip()` on it. The `try/except Exception` only wraps the attribute access — if the attribute returns `None` or a non-string value (a misbehaving but type-consistent implementation), `.strip()` raises `AttributeError` and crashes the runner during prompt-cache-key resolution (`prompt_cache_key.py` calls `resolve_run_grouping` on every run).

The fix adds an `isinstance(session_id, str)` check so a malformed session falls back to the next grouping tier (`group_id` or a generated `run-<uuid>`) instead of breaking the run.

## Changes

- `src/agents/run_internal/run_grouping.py`: 2-line `isinstance` guard before `.strip()`.
- `tests/test_run_grouping.py`: new file covering the None / non-string / blank / valid / fallback paths (no prior test coverage existed for this module).

## Test plan

- [x] `pytest tests/test_run_grouping.py` — 3 tests pass
- [x] `pytest tests/test_tool_use_tracker.py tests/test_prompt_cache_key.py` — no regressions (22 tests pass)
- [x] `ruff check`, `ruff format --check`, `mypy` all clean

### DIFF
--- a/src/agents/run_internal/run_grouping.py
+++ b/src/agents/run_internal/run_grouping.py
@@ -55,5 +55,7 @@ def get_session_id_if_available(session: Session | None) -> str | None:
         session_id = session.session_id
     except Exception:
         return None
+    if not isinstance(session_id, str):
+        return None
     session_id = session_id.strip()
     return session_id if session_id else None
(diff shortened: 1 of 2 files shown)


seratch:

PR #3229, opened 2026-05-08
fix(approvals): skip needs_approval_checker when status already resolved

adityasingh2400:
## Summary

In `_select_function_tool_runs_for_resume`, the user-supplied `needs_approval_checker` was awaited unconditionally for every run on the resume path, including runs whose approval status was already resolved (`True`/`False`).

The checker's result is unused on those branches, so the eager `await` was wasted work — but more importantly, if a user-supplied checker raises (which is plausible for checkers that consult external state), the exception propagates out before the loop can reach `record_rejection`, silently dropping the rejection record for an explicitly-rejected call.

The corresponding helper `_collect_runs_by_approval` already orders these checks correctly (status checks first, then `needs_approval_checker`). This change brings `_select_function_tool_runs_for_resume` in line with that pattern: the checker is only awaited when `approval_status is None`.

## Changes

- `src/agents/run_internal/tool_planning.py`: defer `await needs_approval_checker(run)` until after the `approval_status is True/False` branches. Drop the unreachable trailing `selected.append(run)` since `approval_status` is `bool | None` and all three cases are covered.
- `tests/test_hitl_error_scenarios.py`: add `test_resume_skips_needs_approval_checker_when_status_resolved` — a checker that raises `AssertionError` is *not* invoked for either pre-approved or pre-rejected calls, and the pre-rejected call still hits `record_rejection`.

## Test plan

- [x] `pytest tests/test_hitl_error_scenarios.py tests/test_run_internal_approvals.py` (58 passed)
- [x] `pytest tests/ -k "approval or hitl or resume"` (190 passed, 0 failed)
- [x] `pytest tests/test_run.py tests/test_run_step_processing.py tests/test_run_step_execution.py tests/test_run_impl_resume_paths.py tests/test_run_context_approvals.py tests/test_run_state.py tests/test_run_internal_approvals.py tests/test_run_internal_items.py` (343 passed)

### DIFF
--- a/src/agents/run_internal/tool_planning.py
+++ b/src/agents/run_internal/tool_planning.py
@@ -514,8 +514,6 @@ async def _select_function_tool_runs_for_resume(
             existing_pending=approval_items_by_call_id.get(call_id),
         )
 
-        requires_approval = await needs_approval_checker(run)
-
         if approval_status is False:
             await record_rejection(call_id, run.tool_call, run.function_tool)
             continue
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3230, opened 2026-05-08
fix(realtime): preserve output_audio content parts in output_item events

adityasingh2400:
## Summary

The dict-based fast path in `_handle_ws_event` for `response.output_item.added` / `response.output_item.done` only matched part type `"audio"`. GA-style assistant messages emit `"output_audio"` (matching the official literal `Literal["output_text", "output_audio"]` on `openai.types.realtime.realtime_conversation_item_assistant_message.Content`), so audio content + transcript were silently dropped from listeners on these events.

The other early conversion site (`_ConversionHelper.conversation_item_to_realtime_message_item` at L1588) already maps `output_audio` → `audio`; the dict-path was inconsistent.

This PR accepts both `"audio"` and `"output_audio"` and normalizes to the SDK's `"audio"` type, mirroring how `output_text`/`text` are already handled in the same block.

## Changes

- `src/agents/realtime/openai_realtime.py`: 1-line tuple change in `_handle_ws_event`'s message-content normalization.
- `tests/realtime/test_openai_realtime.py`: focused regression test asserting `output_audio` parts reach listeners with `type="audio"` and the transcript preserved.

Diff: 2 files, +35/-1.

## Test plan

- [x] New test `test_output_audio_content_type_normalized` fails on `main`, passes after fix.
- [x] Full `tests/realtime/` suite (233 tests) passes.
- [x] `ruff check` clean.

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -1023,7 +1023,7 @@ class OpenAIRealtimeWebSocketModel(RealtimeModel):
                 for part in raw_content:
                     if not isinstance(part, dict):
                         continue
-                    if part.get("type") == "audio":
+                    if part.get("type") in ("audio", "output_audio"):
                         converted_content.append(
                             {
                                 "type": "audio",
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3231, opened 2026-05-08
fix(run): raise clear error when handoff is_enabled is invalid type

adityasingh2400:
## Summary

When a handoff's `is_enabled` is mistakenly set to a value that is neither `bool` nor a callable (e.g. a string like `"yes"`), the runner crashed deep inside `asyncio.gather` with `TypeError: 'str' object is not callable`. The traceback gave no hint that the cause was a misconfigured handoff, and a single bad value broke the entire run.

This adds an explicit type check in `get_handoffs` that raises `UserError` naming the offending handoff's target agent, so the misconfiguration is easy to locate.

## Test plan

- [x] New test `test_handoff_is_enabled_invalid_type_raises_user_error` exercises the bad-value path and asserts a `UserError` is raised.
- [x] Existing handoff and run-step tests still pass (`tests/test_handoff_tool.py`, `tests/test_run_step_processing.py`, `tests/test_run_step_execution.py`, `tests/test_agent_runner.py`).

### DIFF
--- a/src/agents/run_internal/turn_preparation.py
+++ b/src/agents/run_internal/turn_preparation.py
@@ -98,6 +98,12 @@ async def get_handoffs(agent: Agent[Any], context_wrapper: RunContextWrapper[Any
         attr = handoff_obj.is_enabled
         if isinstance(attr, bool):
             return attr
+        if not callable(attr):
+            raise UserError(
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch:

PR #3233, opened 2026-05-08
fix(realtime): rename on_play_bytes parameter to avoid shadowing builtin

adityasingh2400:
## Summary

`RealtimePlaybackTracker.on_play_bytes` declared its third parameter as `bytes`, shadowing the Python builtin inside the method body and preventing callers from passing the audio buffer as a keyword argument the way you would naturally expect:

```python
tracker.on_play_bytes(item_id="x", item_content_index=0, audio_bytes=buf)  # TypeError
```

This PR renames the parameter to `audio_bytes`, which:

- Removes the builtin shadow inside the method.
- Matches the convention already used by the sibling internal class `ModelAudioTracker.on_audio_delta(... audio_bytes: bytes)` in `_default_tracker.py`.
- Lets callers pass it as a keyword arg (the docstring documents it that way).

While in the file, also dropped a stale line from `RealtimeRunner.__init__`'s docstring that documented a `context` argument the constructor doesn't accept (context is a `run()` argument, not an `__init__` argument).

All in-tree callers pass the bytes positionally, so the rename is backward-compatible. Verified with the full `tests/realtime/` suite (233 passed) plus a new keyword-arg test in `test_playback_tracker_manual_unit.py`.

## Test plan

- [x] `pytest tests/realtime/test_playback_tracker_manual_unit.py` — both tests pass, including the new `test_playback_tracker_on_play_bytes_accepts_audio_bytes_keyword`.
- [x] `pytest tests/realtime/` — 233 passed, no regressions.
- [x] `git grep "on_play_bytes("` — no caller relies on the old `bytes=` keyword name.

### DIFF
--- a/src/agents/realtime/model.py
+++ b/src/agents/realtime/model.py
@@ -38,15 +38,17 @@ class RealtimePlaybackTracker:
         self._current_item: tuple[str, int] | None = None
         self._elapsed_ms: float | None = None
 
-    def on_play_bytes(self, item_id: str, item_content_index: int, bytes: bytes) -> None:
+    def on_play_bytes(
+        self, item_id: str, item_content_index: int, audio_bytes: bytes
+    ) -> None:
         """Called by you when you have played some audio.
 
         Args:
             item_id: The item ID of the audio being played.
             item_content_index: The index of the audio content in `item.content`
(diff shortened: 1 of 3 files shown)


seratch:
Since this is a breaking change, we won't accept it.

PR #3237, opened 2026-05-08
fix: preserve tool guardrail results across handoffs in SingleStepResult

adityasingh2400:
## Summary
- When a model response contains both function tool calls and a handoff, `execute_handoffs` was returning an empty `tool_input_guardrail_results` / `tool_output_guardrail_results` list, silently dropping any guardrail results produced by the function tools that ran before the handoff was dispatched.
- The aggregator in `run.py` extends the run-wide guardrail-result lists from each `turn_result`, so dropping these on a handoff turn means `RunResult.tool_input_guardrail_results` (and the output equivalent) is missing entries that the caller paid to compute and explicitly opted into.
- Thread the guardrail results from `execute_tools_and_side_effects` and `resolve_interrupted_turn` through to the `SingleStepResult` returned by `execute_handoffs`.

## Test plan
- [x] New test `test_execute_handoffs_preserves_tool_input_guardrail_results` covers a single response containing a guarded function tool call alongside a handoff and asserts the guardrail result survives the handoff.
- [x] Existing `tests/test_run_step_execution.py` (87 tests), `tests/test_run_step_processing.py`, `tests/test_handoff_tool.py`, `tests/test_process_model_response.py`, `tests/test_extension_filters.py`, `tests/test_agent_runner.py` all pass.

### DIFF
--- a/src/agents/run_internal/turn_resolution.py
+++ b/src/agents/run_internal/turn_resolution.py
@@ -341,6 +341,8 @@ async def execute_handoffs(
     run_config: RunConfig,
     server_manages_conversation: bool = False,
     nest_handoff_history_fn: Callable[..., HandoffInputData] | None = None,
+    tool_input_guardrail_results: list[ToolInputGuardrailResult] | None = None,
+    tool_output_guardrail_results: list[ToolOutputGuardrailResult] | None = None,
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3239, opened 2026-05-08
fix(run): cancel sibling guardrail tasks when one raises

adityasingh2400:
## Summary

`run_input_guardrails` and `run_output_guardrails` in `src/agents/run_internal/guardrails.py` iterate guardrail tasks via `asyncio.as_completed` and only cancel siblings on the tripwire branch. If one guardrail raises a non-tripwire exception, the loop exits but unfinished sibling tasks are left scheduled and continue running past the function's return — wasting work, possibly producing observable side effects (e.g. an LLM-based guardrail finishing a network request after the caller already saw a different exception), and triggering `Task was destroyed but it is pending!` warnings at event-loop shutdown.

The fix wraps the `as_completed` loop in `try / except BaseException` and, on any exit that is not a normal return, cancels still-running siblings and `gather`s them (`return_exceptions=True`) before re-raising. This is distinct from #3187, which targets the tripwire branch of `run_output_guardrails`; this PR targets the *raise* branch in both functions.

## Repro

```python
async def raising(ctx, agent, input): raise RuntimeError("boom")
async def slow(ctx, agent, input):
    await asyncio.sleep(0.5)        # leaks past run_input_guardrails return
    print("oops, ran after caller raised")
    return GuardrailFunctionOutput(None, False)

await run_input_guardrails(agent, [InputGuardrail(slow), InputGuardrail(raising)], "x", ctx)
# raises RuntimeError, but `slow` keeps running.
```

## Test plan

- [x] New `test_input_guardrail_raise_cancels_siblings` — asserts the slow sibling is cancelled (not completed) by the time the raise propagates.
- [x] New `test_output_guardrail_raise_cancels_siblings` — same for output guardrails.
- [x] Existing 37 guardrail tests still pass (`tests/test_guardrails.py`).
- [x] `tests/test_runner_guardrail_resume.py`, `tests/test_stream_input_guardrail_timing.py`, `tests/test_tool_guardrails.py` still pass.
- [x] `ruff check` passes for changed files.

### DIFF
--- a/src/agents/run_internal/guardrails.py
+++ b/src/agents/run_internal/guardrails.py
@@ -124,20 +124,29 @@ async def run_input_guardrails(
 
     guardrail_results: list[InputGuardrailResult] = []
 
-    for done in asyncio.as_completed(guardrail_tasks):
-        result = await done
-        if result.output.tripwire_triggered:
-            for t in guardrail_tasks:
-                t.cancel()
-            await asyncio.gather(*guardrail_tasks, return_exceptions=True)
-            _error_tracing.attach_error_to_current_span(
-                SpanError(
-                    message="Guardrail tripwire triggered",
-                    data={"guardrail": result.guardrail.get_name()},
+    try:
+        for done in asyncio.as_completed(guardrail_tasks):
+            result = await done
+            if result.output.tripwire_triggered:
+                for t in guardrail_tasks:
+                    t.cancel()
+                await asyncio.gather(*guardrail_tasks, return_exceptions=True)
+                _error_tracing.attach_error_to_current_span(
+                    SpanError(
+                        message="Guardrail tripwire triggered",
+                        data={"guardrail": result.guardrail.get_name()},
+                    )
                 )
-            )
-            raise InputGuardrailTripwireTriggered(result)
-        guardrail_results.append(result)
+                raise InputGuardrailTripwireTriggered(result)
+            guardrail_results.append(result)
+    except BaseException:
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (changes requested):
can you resolve the conflicts?

adityasingh2400:
Rebased onto current main; conflicts resolved (combined with #3187's tripwire-branch await). Tests + lint clean.

adityasingh2400:
Rebased onto current main. The branch now applies cleanly with no conflicts. I re-ran the suite locally against main: ruff format, ruff lint, and mypy all pass, and the 40 tests in tests/test_guardrails.py are green. The change stays narrow, it cancels the sibling guardrail tasks when one raises so they do not leak, and keeps the existing tripwire await path intact. Ready for another look when you have a moment.

adityasingh2400:
Checked again today: the branch still merges cleanly against main (no conflicts), so the resolve-conflicts request is satisfied. The change is scoped to run_input_guardrails and run_output_guardrails in src/agents/run_internal/guardrails.py: the as_completed loop is wrapped so that on any error path, a guardrail raising a non-tripwire exception or the caller being cancelled, the sibling tasks are cancelled and awaited before the exception propagates, which stops them leaking past the function return. The existing tripwire path is unchanged. Tests in tests/test_guardrails.py cover the new behavior and pass locally with ruff and mypy clean. This is ready for another look whenever you have a moment, and I am happy to rebase again if anything has drifted.

seratch (approved):

PR #3240, opened 2026-05-08
fix(run): raise UserError when error handler dict has include_in_history without final_output

adityasingh2400:
## Summary

`resolve_run_error_handler_result` silently mishandled a run error handler that returned `{\"include_in_history\": False}` (a plausible user typo where `final_output` was forgotten): the whole dict was treated as the raw `final_output` value and `include_in_history` defaulted to `True` — exactly the opposite of the user's intent.

This mirrors the existing extra-keys validation path: when `final_output` IS present, unknown keys raise `UserError`. The new check applies the same rule when `final_output` is absent but the reserved key `include_in_history` is present.

## Repro

```python
result = await resolve_run_error_handler_result(
    error_handlers={\"max_turns\": lambda _i: {\"include_in_history\": False}},
    error=MaxTurnsExceeded(\"too many turns\"),
    context_wrapper=ctx,
    run_data=run_data,
)
# Before: RunErrorHandlerResult(final_output={\"include_in_history\": False}, include_in_history=True)
# After:  raises UserError(\"... 'include_in_history' provided without 'final_output'.\")
```

## Test plan

- [x] New assertion in `test_resolve_run_error_handler_result_covers_async_and_validation_paths` — fails on `main`, passes with the fix.
- [x] `tests/test_run_internal_error_handlers.py` and `tests/test_run_error_details.py` still pass.
- [x] `tests/test_max_turns.py` still passes.

## Diff size

5 source lines + 13 test lines.

### DIFF
--- a/src/agents/run_internal/error_handlers.py
+++ b/src/agents/run_internal/error_handlers.py
@@ -162,5 +162,10 @@ async def resolve_run_error_handler_result(
                 return RunErrorHandlerResult(**result)
             except TypeError as exc:
                 raise UserError("Invalid run error handler result.") from exc
+        if "include_in_history" in result:
+            raise UserError(
+                "Invalid run error handler result: 'include_in_history' provided "
+                "without 'final_output'."
+            )
         return RunErrorHandlerResult(final_output=result)
     return RunErrorHandlerResult(final_output=result)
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3243, opened 2026-05-08
fix(realtime): skip invalid input_text parts in user input conversion

adityasingh2400:
## Summary

When a `RealtimeModelSendUserInput` event carries a dict `user_input` whose `content` includes an `input_text` part with a missing or non-string `text` field, `_ConversionHelper.convert_user_input_to_conversation_item` currently appends a `Content(type="input_text", text=None)` item. The realtime API rejects such payloads.

This change skips those parts entirely, matching the existing forward-compat behaviour for `input_image` with an invalid `image_url`. Only well-formed parts reach the wire.

## Test plan

- [x] Added `test_convert_user_input_dict_skips_invalid_input_text_parts` covering missing-text and non-string-text cases plus a valid sibling (only the valid one survives).
- [x] `pytest tests/realtime` — 233 passed.
- [x] `ruff check` clean on touched files.

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -1641,8 +1641,12 @@ class _ConversionHelper:
                     t = item.get("type")
                     if t == "input_text":
                         _txt = item.get("text")
-                        text_val = _txt if isinstance(_txt, str) else None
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3244, opened 2026-05-08
fix(sessions): return [] for non-positive limits in OpenAIConversationsSession

adityasingh2400:
## Summary

`OpenAIConversationsSession.get_items(limit=0)` returned **1 item instead of `[]`**. The break check (`len(all_items) >= session_limit`) only ran *after* the first `append`, so the first upstream item leaked through any time `limit <= 0`.

`SQLiteSession` already returns `[]` for `SessionSettings(limit=0)` (see `tests/memory/test_session.py::test_session_memory_with_limit` and `tests/memory/test_session_limit.py`), so this also fixes a cross-backend semantic skew that would silently change session contents depending on which store was wired in.

The fix short-circuits non-positive limits *before* issuing the upstream `conversations.items.list` call:

```python
elif session_limit <= 0:
    # Match SQLiteSession semantics: a non-positive limit yields no history,
    # without issuing an upstream request that would otherwise reject limit=0.
    return []
```

This also avoids sending `limit=0` to the OpenAI Conversations API, which would otherwise reject the request.

## Test plan

- [x] New test `test_get_items_with_zero_limit_returns_empty` covers the fix and asserts the upstream `list` call is not made
- [x] All existing memory tests pass: `pytest tests/memory/ -q` -> 120 passed
- [x] Repro confirmed pre-fix: returned 1 item; post-fix: returns 0

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -83,6 +83,10 @@ class OpenAIConversationsSession(SessionABC):
             ):
                 # calling model_dump() to make this serializable
                 all_items.append(item.model_dump(exclude_unset=True))
+        elif session_limit <= 0:
+            # Match SQLiteSession semantics: a non-positive limit yields no history,
+            # without issuing an upstream request that would otherwise reject limit=0.
+            return []
         else:
             async for item in self._openai_client.conversations.items.list(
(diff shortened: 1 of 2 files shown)


seratch:
We don't need this guard

PR #3245, opened 2026-05-08
fix(run): preserve last known response_id on conversation resume

adityasingh2400:
## Summary

`OpenAIServerConversationTracker.hydrate_from_state` seeds `previous_response_id` from `model_responses[-1].response_id`. When the final saved response has no id (e.g., a non-Responses provider returns `response_id=None`), the conversation chain silently resets to `None` on resume — even though earlier responses had valid ids.

Live runs do not exhibit this because `track_server_items` only updates `previous_response_id` when the new response carries an id, preserving the last-known value across turns. Resume should mirror that behavior.

## Fix

Walk the saved `model_responses` and capture the most recent response that actually carries an id; seed `previous_response_id` from that value instead of from `model_responses[-1]` unconditionally.

## Test plan

- [x] New test `test_hydrate_from_state_uses_latest_non_none_response_id` reproduces the bug — fails on `main`, passes with the fix.
- [x] Existing `tests/test_server_conversation_tracker.py` (20 tests) all pass.
- [x] 249 tests across `conversation`/`tracker`/`resume`/`oaiconv` keywords pass.
- [x] `ruff check` and `ruff format --check` clean.

### DIFF
--- a/src/agents/run_internal/oai_conversation.py
+++ b/src/agents/run_internal/oai_conversation.py
@@ -189,8 +189,14 @@ class OpenAIServerConversationTracker:
         self.sent_initial_input = True
         self.remaining_initial_input = None
 
-        latest_response = model_responses[-1] if model_responses else None
+        # Pick the most recent response that actually carries an id; live runs preserve the
+        # last-known id via track_server_items, so resume must mirror that behavior instead of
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3247, opened 2026-05-08
fix(sessions): skip corrupt docs in MongoDBSession.pop_item

adityasingh2400:
## Summary

`MongoDBSession.pop_item` returns `None` from two distinct conditions:

1. The session is empty.
2. The most-recent document's `message_data` fails to deserialize (invalid JSON, missing field, or non-string type — the same `(JSONDecodeError, KeyError, TypeError)` triple that `get_items` already swallows).

In the second case the corrupt document is removed by `find_one_and_delete` and the caller sees `None`, even though valid older messages still exist. A loop that pops items until empty therefore stops one item early and leaves history behind.

This mirrors the behavior `get_items` already provides — it silently skips corrupt rows and keeps reading valid ones — so `pop_item` should do the same to stay consistent.

## Fix

Wrap the pop in a `while True` loop that discards corrupt documents and re-pops the next-most-recent until a valid item is decoded or the session is genuinely empty. The fix is local to `pop_item` and adds five lines of behavior plus a docstring note.

## Test plan

- [x] `tests/extensions/memory/test_mongodb_session.py::test_pop_item_skips_corrupt_most_recent` — corrupt latest + valid older returns the older item and leaves the collection empty.
- [x] `tests/extensions/memory/test_mongodb_session.py::test_pop_item_returns_none_when_only_corrupt_docs_remain` — all-corrupt session drains every bad row before returning `None`.
- [x] All 31 existing `MongoDBSession` tests still pass (33 total now).
- [x] `ruff check` / `ruff format --check` clean.

### DIFF
--- a/src/agents/extensions/memory/mongodb_session.py
+++ b/src/agents/extensions/memory/mongodb_session.py
@@ -324,21 +324,26 @@ class MongoDBSession(SessionABC):
 
         Returns:
             The most recent item if it exists, ``None`` if the session is empty.
+
+        Corrupt documents (invalid JSON, missing/non-string ``message_data``)
+        are silently discarded and the next-most-recent item is returned.  This
+        matches :meth:`get_items`, which also skips corrupt documents, so a
+        single bad row cannot make a non-empty session look empty to callers.
         """
         await self._ensure_indexes()
 
-        doc = await self._messages.find_one_and_delete(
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (approved):

PR #3248, opened 2026-05-08
fix(realtime): raise UserError for input_type without on_handoff

adityasingh2400:
## Summary

The validation guard at the top of \`realtime_handoff()\` is a tautology:

\`\`\`python
assert (on_handoff and input_type) or not (on_handoff and input_type), (
    "You must provide either both on_handoff and input_type, or neither"
)
\`\`\`

\`(X) or not (X)\` is always \`True\`, so the assertion never fires. When a caller does \`realtime_handoff(agent, input_type=int)\` (input_type without on_handoff), the function falls through to a less helpful \`assert callable(on_handoff)\` further down, or — under \`python -O\` where assertions are stripped — silently constructs a \`Handoff\` whose \`on_invoke_handoff\` callable hits a confusing \`TypeError: None is not a callable object\` at first invocation.

This mirrors the same pattern in \`agents/handoffs/__init__.py\` line 255, but the realtime path was the only one missed when Wave 7 cleaned things up around #3243. Replace the dead assertion with an explicit \`UserError\` so the contract documented by the overloads (input_type only with on_handoff) is enforced uniformly, including under \`-O\`.

## Test plan

- [x] \`pytest tests/realtime -q\` -> 233 passed
- [x] New test \`test_realtime_handoff_input_type_requires_on_handoff\` fails on main (\`AssertionError\` from the secondary guard) and passes after the fix (\`UserError\`)
- [x] Existing \`test_realtime_handoff_invalid_param_counts_raise\` still passes — same UserError surface area for arity mismatches

### DIFF
--- a/src/agents/realtime/handoffs.py
+++ b/src/agents/realtime/handoffs.py
@@ -88,9 +88,8 @@ def realtime_handoff(
 
     Note: input_filter is not supported for RealtimeAgent handoffs.
     """
-    assert (on_handoff and input_type) or not (on_handoff and input_type), (
-        "You must provide either both on_handoff and input_type, or neither"
-    )
+    if input_type is not None and on_handoff is None:
+        raise UserError("You must provide on_handoff when input_type is provided")
     type_adapter: TypeAdapter[Any] | None
     if input_type is not None:
         assert callable(on_handoff), "on_handoff must be callable"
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3255, opened 2026-05-08
fix(tracing): skip 'Trace already exists' warning under NoOpTrace

adityasingh2400:
## Summary

When tracing is globally disabled (for example via `OPENAI_AGENTS_DISABLE_TRACING=1`), entering a nested `with trace(...)` block currently logs the misleading warning:

> `Trace already exists. Creating a new trace, but this is probably a mistake.`

But the outer trace in that case is a `NoOpTrace` — nesting another trace inside it is harmless, not a programming mistake. The warning is pure noise that's especially confusing for users who explicitly disabled tracing to keep logs quiet.

This skips the warning when the current trace is a `NoOpTrace`. Real `TraceImpl`/`ReattachedTrace` parents still trigger the warning as before.

### Repro
```python
import os
os.environ["OPENAI_AGENTS_DISABLE_TRACING"] = "1"
from agents.tracing import trace

with trace("outer"):       # NoOpTrace
    with trace("inner"):   # logs "Trace already exists..." today
        pass
```

## Test plan
- [x] New test `tests/tracing/test_trace_warning_noise.py::test_nested_trace_under_noop_does_not_warn` fails on `main` and passes with this fix.
- [x] Full `tests/tracing/` and `tests/test_trace*.py` suites pass (77 passed).

### DIFF
--- a/src/agents/tracing/create.py
+++ b/src/agents/tracing/create.py
@@ -22,7 +22,7 @@ from .span_data import (
     TurnSpanData,
 )
 from .spans import Span
-from .traces import Trace
+from .traces import NoOpTrace, Trace
 
 if TYPE_CHECKING:
     from openai.types.responses import Response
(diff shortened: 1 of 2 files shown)


seratch:
I think the current code is okay here

PR #3256, opened 2026-05-08
fix(run): expose pydantic raw_item fields in resume debug payload

adityasingh2400:
## Summary

`build_generated_items_details` (used by `build_resumed_stream_debug_extra` to log resume payloads) only populated `raw_type`/`name`/`call_id`/`output` when `item.raw_item` was a `dict`. Most `RunItem` subclasses keep the typed Pydantic model on `raw_item` (e.g. `ToolCallItem.raw_item: ResponseFunctionToolCall`, `HandoffCallItem.raw_item: ResponseFunctionToolCall`, `ReasoningItem.raw_item: ResponseReasoningItem`), so the debug log silently dropped those fields for the typed cases — `ToolCallItem` resumes were logged as just `{"index": 0, "type": "tool_call_item"}`.

## Fix

Read the fields generically through a small `_raw_item_field` helper that handles both `Mapping`-shaped raw items and Pydantic model attributes. Behaviour for dict raw items is preserved (including the 100-char output truncation).

## Test plan

- [x] New `tests/test_build_generated_items_details.py` covers Pydantic and dict shapes plus the `include_tool_output=False` case. The Pydantic test fails on `main` (`KeyError: 'raw_type'`) and passes with the fix.
- [x] `pytest tests/test_server_conversation_tracker.py tests/models/test_agent_registration.py tests/test_run_state.py` (224 tests) passes.
- [x] `ruff check` and `ruff format --check` clean.

### DIFF
--- a/src/agents/run_internal/agent_runner_helpers.py
+++ b/src/agents/run_internal/agent_runner_helpers.py
@@ -314,6 +314,13 @@ def describe_run_state_step(step: object | None) -> str | int | None:
     return type(step).__name__
 
 
+def _raw_item_field(raw_item: Any, key: str) -> Any:
+    """Read a field from a raw item, supporting both dict and Pydantic shapes."""
+    if isinstance(raw_item, Mapping):
+        return raw_item.get(key)
+    return getattr(raw_item, key, None)
+
+
 def build_generated_items_details(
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3258, opened 2026-05-08
fix(sandbox): require https for vercel exposed port domains

adityasingh2400:
## Summary

`VercelSandboxSession._resolve_exposed_port` accepted any URL returned by `sandbox.domain()` — including `http://` or schemeless — and silently exposed it as a non-TLS endpoint (`tls=False`, port 80). Vercel sandbox preview domains are always served over HTTPS, so a non-https scheme indicates either a backend bug or a tampered response. Either way, downstream code would then connect over plaintext to a host derived from that URL.

This change rejects any domain whose scheme is not `https`, raising `ExposedPortUnavailableError` via the same path already used for missing/invalid hostnames, and drops the conditional `http`/port-80 fallback.

This is in the same vein as #3206 (and earlier #3094 / #3172 / #3177) — defensive narrowing of what a sandbox backend will trust from the wire.

## Diff

- `src/agents/extensions/sandbox/vercel/sandbox.py`: 7 lines changed (require https in `_resolve_exposed_port`).
- `tests/extensions/sandbox/test_vercel.py`: new `test_vercel_resolve_exposed_port_rejects_non_https_domain` plus `ExposedPortUnavailableError` import.

## Test plan

- [x] New unit test `test_vercel_resolve_exposed_port_rejects_non_https_domain` fails on `main` and passes after the fix.
- [x] Full `tests/extensions/sandbox/test_vercel.py` suite (35 tests) passes.

### DIFF
--- a/src/agents/extensions/sandbox/vercel/sandbox.py
+++ b/src/agents/extensions/sandbox/vercel/sandbox.py
@@ -460,18 +460,17 @@ class VercelSandboxSession(BaseSandboxSession):
 
         parsed = urlsplit(domain)
         host = parsed.hostname
-        if not host:
+        if not host or parsed.scheme != "https":
             raise ExposedPortUnavailableError(
                 port=port,
                 exposed_ports=self.state.exposed_ports,
                 reason="backend_unavailable",
                 context={"backend": "vercel", "domain": domain},
             )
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3259, opened 2026-05-08
fix(approvals): skip needs_approval_checker when status resolved in _collect_runs_by_approval

adityasingh2400:
## Summary

Mirrors #3229 for non-function tools (shell, apply_patch, custom). `_collect_runs_by_approval` invokes the user-supplied `needs_approval_checker` even when `approval_status` is already `True` or `False`. The checker may have side effects (telemetry, network) and a non-`UserError` exception is silently swallowed; a `UserError` would short-circuit an already-approved call.

The parallel function `_select_function_tool_runs_for_resume` was fixed in #3229 with the same reasoning.

## Fix

Move the `approval_status is True` early-return above the checker invocation so explicit approve/reject decisions short-circuit cleanly. Rejected (`is False`) was already short-circuited.

## Test plan

- [x] New test `test_collect_runs_by_approval_skips_checker_when_status_resolved` confirms the checker is not invoked for approved/rejected shell calls.
- [x] Test fails on main, passes with the fix.
- [x] Existing `tests/test_hitl_error_scenarios.py` (51 tests) and related HITL/approval suites pass.

### DIFF
--- a/src/agents/run_internal/tool_planning.py
+++ b/src/agents/run_internal/tool_planning.py
@@ -411,6 +411,10 @@ async def _collect_runs_by_approval(
             rejection_items.append(rejection_item)
             continue
 
+        if approval_status is True:
+            approved_runs.append(run)
+            continue
+
         needs_approval = True
         if needs_approval_checker:
             try:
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

adityasingh2400:
Rebased onto current main. The change still applies cleanly and the suite is green: ruff format, ruff lint, mypy, and the 52 tests in tests/test_hitl_error_scenarios.py all pass locally. The fix skips the needs_approval_checker call in _collect_runs_by_approval once the approval status is already resolved, which avoids the redundant checker invocation. Marking it fresh after the stale ping, happy to take it out of draft whenever it is useful to review.

PR #3260, opened 2026-05-08
fix(agent_output): return "None" from name() when output_type is None

adityasingh2400:
## Summary
- `AgentOutputSchema.__init__` accepts `output_type=None` as a plain-text marker (alongside `str`) and the constructor handles it explicitly, but `name()` crashes with `AttributeError: 'NoneType' object has no attribute '__name__'` because `_type_to_str` dereferences `t.__name__` unconditionally.
- Special-case `None` in `_type_to_str` so it returns the string `"None"`, mirroring how `str` already returns `"str"`.
- Also use `getattr(t, "__name__", str(t))` for the simple-type branch so any other unusual non-`type` object falls back to its `repr` instead of crashing.

## Repro
```python
from agents.agent_output import AgentOutputSchema
AgentOutputSchema(None).name()  # AttributeError on main, returns "None" with this fix
```

## Test plan
- [x] New `test_name_returns_string_for_none_output_type` fails on `main` with the original `AttributeError` and passes with this fix.
- [x] Existing `tests/test_output_tool.py`, `tests/test_agent_tool_input.py`, `tests/test_agent_tool_state.py`, and the openai chat/responses converter tests still pass (106 tests green).

### DIFF
--- a/src/agents/agent_output.py
+++ b/src/agents/agent_output.py
@@ -181,12 +181,15 @@ def _is_subclass_of_base_model_or_dict(t: Any) -> bool:
 
 
 def _type_to_str(t: type[Any]) -> str:
+    if t is None:
+        return "None"
+
     origin = get_origin(t)
     args = get_args(t)
 
     if origin is None:
         # It's a simple type like `str`, `int`, etc.
-        return t.__name__
+        return getattr(t, "__name__", str(t))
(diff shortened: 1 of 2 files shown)


seratch:
Thanks, but I don't see the necessity to have this code change.

PR #3262, opened 2026-05-08
fix(codex): normalize configured thread_id in _resolve_call_thread_id

adityasingh2400:
## Summary
- ``codex_tool(thread_id="   ")`` (whitespace-only) ends up calling ``codex.resume_thread("   ")`` because the configured thread id is returned by ``_resolve_call_thread_id`` without going through ``_normalize_thread_id``. JSON-input thread ids and run-context thread ids are already normalized, so this is the only path that leaks unusable strings.
- Even passing a non-string thread_id silently propagates to ``Codex.resume_thread`` instead of raising the documented ``UserError``.
- Run ``configured_thread_id`` through the same ``_normalize_thread_id`` helper. Whitespace-only ids fall back to ``codex.start_thread`` and non-string ids raise ``UserError("Codex thread_id must be a string when provided.")``, matching the JSON-input contract.

## Test plan
- [x] Added ``test_codex_tool_starts_new_thread_for_whitespace_configured_thread_id`` exercising ``codex_tool(thread_id="   ")``; fails on ``main`` because ``resume_thread`` is invoked with the whitespace string, passes with this fix.
- [x] ``pytest tests/extensions/experiemental/codex`` (128 passed).

### DIFF
--- a/src/agents/extensions/experimental/codex/codex_tool.py
+++ b/src/agents/extensions/experimental/codex/codex_tool.py
@@ -832,7 +832,7 @@ def _resolve_call_thread_id(
         if context_thread_id:
             return context_thread_id
 
-    return configured_thread_id
+    return _normalize_thread_id(configured_thread_id)
 
 
 def _read_thread_id_from_run_context(ctx: RunContextWrapper[Any], key: str) -> str | None:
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3263, opened 2026-05-08
fix(apply_diff): drop all trailing blank lines in diff normalization

adityasingh2400:
## Summary

`_normalize_diff_lines` only popped a single trailing blank entry, so a V4A diff padded with multiple trailing newlines (e.g. by an editor or transport that always appends an extra blank line) leaked an empty line into the parser. The update parser then treats that blank as an empty context line, raising `Invalid Context` against any input whose last block does not end with two consecutive blank lines.

Switch the trailing-blank check to a `while` loop so all trailing blanks are stripped, matching the intent of the existing single-blank case.

Repro before the fix:

```python
apply_diff('line1\nline2\n', '@@ line1\n-line2\n+updated\n\n\n')
# ValueError: Invalid Context 1:\nline2\n
```

After the fix this returns `'line1\nupdated\n'`.

## Test plan

- [x] `pytest tests/test_apply_diff.py tests/test_apply_diff_helpers.py tests/sandbox/test_apply_patch.py` (37 passed)
- [x] New tests fail on `main`, pass with the fix

### DIFF
--- a/src/agents/apply_diff.py
+++ b/src/agents/apply_diff.py
@@ -67,7 +67,7 @@ def apply_diff(input: str, diff: str, mode: ApplyDiffMode = "default") -> str:
 
 def _normalize_diff_lines(diff: str) -> list[str]:
     lines = [line.rstrip("\r") for line in re.split(r"\r?\n", diff)]
-    if lines and lines[-1] == "":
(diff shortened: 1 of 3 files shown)


seratch:
@codex review

PR #3264, opened 2026-05-08
fix(codex): guard non-mapping todo_list entries in coerce_thread_item

adityasingh2400:
## Summary

The `todo_list` branch in `coerce_thread_item` iterates `raw["items"]` and calls `item.get(...)` directly, without checking that each entry is a `Mapping`. A malformed event like `{"type": "todo_list", "id": "x", "items": [None]}` therefore crashes with a raw `AttributeError: 'NoneType' object has no attribute 'get'` instead of the clean `TypeError` that the sibling branches (`file_change`, `mcp_tool_call`) already surface through their `_coerce_*` helpers.

Add a matching `_coerce_todo_item` helper that:

- pass-throughs existing `TodoItem` instances (consistent with `_coerce_file_update_change` / `_coerce_mcp_tool_call_*`),
- rejects non-mapping entries with `TypeError("TodoItem must be a mapping.")`,
- preserves the existing `text` / `completed` defaults.

Net change: ~16 lines in `items.py`, +22 lines of tests.

## Test plan

- [x] `pytest tests/extensions/experiemental/codex/test_payloads.py -v` — new tests fail on `main` with `AttributeError`, pass after the fix.
- [x] `pytest tests/extensions/experiemental` — all 129 tests pass.

### DIFF
--- a/src/agents/extensions/experimental/codex/items.py
+++ b/src/agents/extensions/experimental/codex/items.py
@@ -168,6 +168,17 @@ def _coerce_mcp_tool_call_error(
     return McpToolCallError(message=cast(str, raw.get("message", "")))
 
 
+def _coerce_todo_item(raw: TodoItem | Mapping[str, Any]) -> TodoItem:
+    if isinstance(raw, TodoItem):
+        return raw
+    if not isinstance(raw, Mapping):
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch:
same with #3251

PR #3276, opened 2026-05-08
fix: #3273 validate git repo subpaths

### LINKED ISSUE #3273: GitRepo.subpath should reject parent traversal paths
### Please read this first

### Describe the bug

`GitRepo.apply()` builds the source copied from the temporary clone by appending `subpath` to the clone directory. On current `main`, a `subpath` containing `..` can make the `cp -R` source leave the cloned repository root.

This means a `git_repo` artifact with an untrusted or generated `subpath` can copy files from outside the clone directory into the destination workspace.

### Debug information

- Agents SDK version: `main@683b6e7`

### Repro steps

Run this from the repository root:

```python
import asyncio
import uuid
from pathlib import Path

from agents.sandbox.entries import GitRepo
from agents.sandbox.types import ExecResult


class State:
    session_id = uuid.UUID("00000000-0000-0000-0000-000000000001")
    image = "fake"


class RecordingSession:
    state = State()

    def __init__(self):
        self.exec_calls = []

    async def exec(self, *command, **kwargs):
        self.exec_calls.append(tuple(str(part) for part in command))
        return ExecResult(stdout=b"", stderr=b"", exit_code=0)

    async def mkdir(self, path, *, parents=False, user=None):
        pass


async def main():
    session = RecordingSession()
    repo = GitRepo(repo="openai/example", ref="main", subpath="../../outside")
    await repo.apply(session, Path("/workspace/copied"), Path("/ignored"))
    print([call for call in session.exec_calls if call[:1] == ("cp",)])


asyncio.run(main())
```

Actual result on current `main`: the recorded copy command contains a source like:

```text
/tmp/sandbox-git-00000000000000000000000000000001-.../../../outside/.
```

### Expected behavior

`GitRepo.subpath` should be validated before copying. Empty paths, absolute paths, Windows-style paths, and any path containing a `..` component should be rejected before `cp` runs. Normal relative paths such as `docs/reference` should continue to work.

Aphroq:
### Summary

- Add validation for `GitRepo.subpath` before copying from the cloned repository.
- Reject empty, absolute, Windows-style, and parent traversal subpaths with a structured `GitSubpathError`.
- Add regression tests covering invalid subpaths and a valid relative subpath.

### Test plan

- `uv run pytest tests/sandbox/test_entries.py -k "git_repo"`

### Issue number

Closes #3273

### Checks

### DIFF
--- a/src/agents/sandbox/entries/artifacts.py
+++ b/src/agents/sandbox/entries/artifacts.py
@@ -8,7 +8,7 @@ import re
 import stat
 import uuid
 from collections.abc import Awaitable, Callable, Mapping
-from pathlib import Path
+from pathlib import Path, PurePosixPath, PureWindowsPath
 from typing import TYPE_CHECKING, Literal
 
 from pydantic import Field, field_serializer, field_validator
@@ -17,6 +17,7 @@ from ..errors import (
     GitCloneError,
     GitCopyError,
     GitMissingInImageError,
+    GitSubpathError,
     LocalChecksumError,
     LocalDirReadError,
     LocalFileReadError,
@@ -786,9 +787,7 @@ class GitRepo(BaseEntry):
                     context={"repo": self.repo, "subpath": self.subpath},
                 )
 
-            git_src_root: str = tmp_dir
-            if self.subpath is not None:
-                git_src_root = f"{tmp_dir}/{self.subpath.lstrip('/')}"
+            git_src_root = self._git_src_root(tmp_dir)
 
             # Copy into destination in the container.
             await session.mkdir(dest, parents=True)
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
The intent of validating git_repo.subpath makes sense, but this currently validates too late.

With an invalid value like subpath="../outside", the code still performs the git clone first and only raises GitSubpathError afterward. Since this is a local configuration error and can be detected without touching the network or filesystem-heavy clone path, we should fail fast before cloning.

Could we move the subpath validation ahead of the clone step so invalid config is rejected before any repository fetch work starts?

Also, the CI is failing now.

Aphroq:
Thanks for the review. I moved the `subpath` validation ahead of the git availability check and clone/fetch path, so invalid local configuration now raises `GitSubpathError` before any sandbox command is executed.

I also tightened the invalid-subpath regression test to assert that no session exec calls are made, and fixed CI.

seratch (approved):

seratch:
Checked this again and adjusted the empty string pattern: https://github.com/openai/openai-agents-python/pull/3299

PR #3278, opened 2026-05-08
fix: #3274 limit sandbox archive extraction

### LINKED ISSUE #3274: Archive extraction should enforce resource limits before writing payloads
### Please read this first

- **Have you searched for related issues?** Yes. I searched upstream issues and PRs for `archive extraction size limit zip tar` and did not find a direct duplicate.

### Describe the bug

Sandbox archive extraction validates paths, links, duplicate paths, and non-directory parents, but it does not cap the archive input size, member count, or declared extracted-byte total before writing extracted payloads.

This leaves `extract_archive()` exposed to untrusted zip/tar archives that contain excessive members or expand to excessive payload size.

### Debug information

- Agents SDK version: `main@683b6e7`

### Repro steps

Run this from the repository root:

```python
import asyncio
import io
import tempfile
import zipfile
from pathlib import Path

from agents.sandbox.manifest import Manifest
from agents.sandbox.sandboxes.unix_local import (
    UnixLocalSandboxSession,
    UnixLocalSandboxSessionState,
)
from agents.sandbox.snapshot import NoopSnapshot


def zip_with_many_members(count: int) -> io.BytesIO:
    buf = io.BytesIO()
    with zipfile.ZipFile(buf, mode="w") as archive:
        for index in range(count):
            archive.writestr(f"files/{index}.txt", b"x")
    buf.seek(0)
    return buf


async def main():
    root = Path(tempfile.mkdtemp()) / "workspace"
    session = UnixLocalSandboxSession.from_state(
        UnixLocalSandboxSessionState(
            manifest=Manifest(root=str(root)),
            snapshot=NoopSnapshot(id="noop"),
        )
    )
    await session.start()
    try:
        await session.extract("bundle.zip", zip_with_many_members(10_001))
    finally:
        await session.shutdown()
    print("extracted members:", len(list((root / "files").iterdir())))


asyncio.run(main())
```

Actual result on current `main`: extraction proceeds and writes all members without a resource-limit error.

### Expected behavior

Archive extraction should enforce finite limits before writing extracted payloads, including:

- maximum archive input bytes
- maximum archive member count
- maximum declared extracted bytes

When any limit is exceeded, extraction should fail before writing member payloads.

Aphroq:
### Summary

- Add an archive input byte limit to public `extract_archive()`.
- Preflight tar/zip member count and declared extracted-byte totals before writing extracted payloads.
- Add small-limit regression tests for input size, tar/zip extracted bytes, and tar/zip member count.

### Test plan

- `uv run pytest tests/sandbox/test_extract.py -k "extract"`

### Issue number

Closes #3274

### Checks

### DIFF
--- a/docs/sandbox/guide.md
+++ b/docs/sandbox/guide.md
@@ -481,6 +481,28 @@ These options only matter when the runner is creating a fresh sandbox session:
 
 `concurrency_limits` controls how much sandbox materialization work can run in parallel. Use `SandboxConcurrencyLimits(manifest_entries=..., local_dir_files=...)` when large manifests or local directory copies need tighter resource control. Set either value to `None` to disable that specific limit.
 
+`archive_limits` controls SDK-side resource checks for `BaseSandboxSession.extract()` archive writes. The default, `archive_limits=None`, preserves the historical behavior and does not enforce SDK archive resource limits. Use `SandboxArchiveLimits()` to enable the SDK default thresholds, or pass explicit values such as `SandboxArchiveLimits(max_input_bytes=..., max_extracted_bytes=..., max_members=...)`. Set an individual field to `None` to disable only that limit.
+
+A direct `session.extract(..., archive_limits=SandboxArchiveLimits(...))` call can override the session default for that extraction. Set all fields to `None` to disable all SDK archive limits for that call.
+
+```python
+from agents.run import RunConfig
+from agents.sandbox import SandboxArchiveLimits, SandboxRunConfig
+
+run_config = RunConfig(
+    sandbox=SandboxRunConfig(
+        client=client,
+        archive_limits=SandboxArchiveLimits(),
+    )
+)
+
+await session.extract(
+    "bundle.tar",
+    archive_stream,
+    archive_limits=SandboxArchiveLimits(max_input_bytes=None, max_members=200_000),
+)
+```
+
 A few implications are worth keeping in mind:
 
 - Fresh sessions: `manifest=` and `snapshot=` only apply when the runner is creating a fresh sandbox session.
--- a/docs/sandbox/guide.md
+++ b/docs/sandbox/guide.md
@@ -481,28 +481,6 @@ These options only matter when the runner is creating a fresh sandbox session:
 
 `concurrency_limits` controls how much sandbox materialization work can run in parallel. Use `SandboxConcurrencyLimits(manifest_entries=..., local_dir_files=...)` when large manifests or local directory copies need tighter resource control. Set either value to `None` to disable that specific limit.
 
-`archive_limits` controls SDK-side resource checks for `BaseSandboxSession.extract()` archive writes. The default, `archive_limits=None`, preserves the historical behavior and does not enforce SDK archive resource limits. Use `SandboxArchiveLimits()` to enable the SDK default thresholds, or pass explicit values such as `SandboxArchiveLimits(max_input_bytes=..., max_extracted_bytes=..., max_members=...)`. Set an individual field to `None` to disable only that limit.
-
-A direct `session.extract(..., archive_limits=SandboxArchiveLimits(...))` call can override the session default for that extraction. Set all fields to `None` to disable all SDK archive limits for that call.
-
-```python
-from agents.run import RunConfig
-from agents.sandbox import SandboxArchiveLimits, SandboxRunConfig
-
-run_config = RunConfig(
-    sandbox=SandboxRunConfig(
-        client=client,
-        archive_limits=SandboxArchiveLimits(),
-    )
-)
-
-await session.extract(
-    "bundle.tar",
-    archive_stream,
-    archive_limits=SandboxArchiveLimits(max_input_bytes=None, max_members=200_000),
-)
-```
-
 A few implications are worth keeping in mind:
 
 - Fresh sessions: `manifest=` and `snapshot=` only apply when the runner is creating a fresh sandbox session.
--- a/src/agents/sandbox/session/archive_extraction.py
+++ b/src/agents/sandbox/session/archive_extraction.py
@@ -12,7 +12,7 @@ from typing import Literal, cast
 
 from ..errors import ExecNonZeroError, WorkspaceArchiveWriteError
 from ..files import EntryKind, FileEntry
-from ..util.tar_utils import UnsafeTarMemberError, safe_tar_member_rel_path, validate_tarfile
+from ..util.tar_utils import UnsafeTarMemberError, safe_tar_member_rel_path
 
 MAX_ARCHIVE_MEMBERS = 10_000
 MAX_ARCHIVE_EXTRACTED_BYTES = 512 * 1024 * 1024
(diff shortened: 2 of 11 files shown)


src/agents/sandbox/session/archive_extraction.py
```diff
@@ -46,8 +67,10 @@ async def extract_tar_archive(
         child_entry_cache: dict[Path, dict[str, EntryKind]] = {}
         try:
             with tarfile.open(fileobj=data, mode="r:*") as archive:
+                members = archive.getmembers()
```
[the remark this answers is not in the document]
Aphroq: Handled in `98a86a3a`. The tar path no longer calls `archive.getmembers()` to enforce extraction limits. It now performs a first streaming `r|*` pass over tar headers to validate member paths, member count, declared extracted bytes, duplicate paths, and non-directory parents; after that succeeds, it rewinds and performs a second streaming extraction pass.

This rejects as soon as the first over-limit tar member is seen, without materializing the full header table in memory, while preserving the existing behavior that validation failures do not leave partially extracted files behind. I also updated the member-count regression test to monkeypatch `TarFile.getmembers()` to raise, so the test now proves the limit path does not depend on `getmembers()`.

Aphroq:
@codex review

src/agents/sandbox/session/archive_extraction.py
```diff
@@ -12,7 +12,10 @@
 
 from ..errors import ExecNonZeroError, WorkspaceArchiveWriteError
 from ..files import EntryKind, FileEntry
-from ..util.tar_utils import UnsafeTarMemberError, safe_tar_member_rel_path, validate_tarfile
+from ..util.tar_utils import UnsafeTarMemberError, safe_tar_member_rel_path
+
+MAX_ARCHIVE_MEMBERS = 10_000
+MAX_ARCHIVE_EXTRACTED_BYTES = 512 * 1024 * 1024
```
seratch: how did you decide these magic numbers?

Aphroq: These numbers are not derived from a published sandbox quota, and this extraction path did not have an existing archive-size cap to copy from.

I picked `512 MiB` for the byte-size defaults because it should leave reasonable room for typical source/workspace archives, while giving the local spool, workspace write, and declared extraction size a clear upper bound. The input cap protects the local spool/workspace write before archive validation runs; the extracted-size cap is a second layer for archives that are acceptable as input but declare a much larger extraction footprint.

Both byte-size caps use `512 MiB` right now to keep the first version simple and conservative, not because compressed and extracted sizes need to match exactly. If we want to leave more room for normal archive expansion, the extracted-size cap could be higher than the input cap.

The `10k` member cap is separate. I chose that to stay out of the way for ordinary source archives while still stopping cases where the cost is mostly the number of entries, like many empty or tiny files.

That said, I’m not attached to the exact thresholds; happy to adjust them if you’d prefer a tighter or looser default. If we decide these should be user-tunable, I think the design should be an explicit limits object rather than a hidden global setting.

seratch: Thanks for addressing the tar streaming issue. I think the remaining concern is that the PR turns archive resource limits into public behavior, but only exposes them as hard-coded module globals.

Could you please make these limits explicitly configurable and avoid changing the default behavior?

Suggested API shape:

- Add a small dataclass such as `SandboxArchiveLimits` with:
  - `max_input_bytes: int | None`
  - `max_extracted_bytes: int | None`
  - `max_members: int | None`
- Add `archive_limits: SandboxArchiveLimits | None = None` to `SandboxRunConfig`, appended after the existing fields to preserve positional compatibility.
- Treat `archive_limits=None` as "do not enforce SDK archive resource limits", preserving the current behavior.
- Treat `SandboxArchiveLimits()` as "enable archive resource limits with SDK default thresholds".
- Treat `None` on an individual `SandboxArchiveLimits` field as "disable this specific limit".
- Validate non-`None` values as positive integers.
- Also allow a per-call override on `BaseSandboxSession.extract(..., archive_limits=...)`, since `extract()` is a direct public entry point outside `Runner`.

I would also like the default thresholds used by `SandboxArchiveLimits()` to be less conservative than the current hard-coded values. For example:

- `max_input_bytes = 1 GiB`
- `max_extracted_bytes = 4 GiB`
- `max_members = 100_000`

I do not think `512 MiB` for both input bytes and extracted bytes is a good default. Those limits protect different resources, and compressed archives commonly expand beyond their input size. I also think `10,000` members is low for legitimate sandbox workloads that include dependency trees, generated files, logs, datasets, media assets, or workspace restore/migration archives.

This keeps resource limits available for applications that handle untrusted archives, while avoiding a default behavior change for applications that currently extract large trusted archives.

Please add tests covering:

- Default `archive_limits=None` preserving the existing no-resource-limit behavior.
- `SandboxArchiveLimits()` enabling the SDK default thresholds.
- `SandboxRunConfig.archive_limits` being applied to Runner-created, resumed, and injected sessions.
- `BaseSandboxSession.extract(..., archive_limits=...)` overriding the session default.
- Invalid non-positive limits being rejected.
- `None` disabling only the selected limit when a `SandboxArchiveLimits` object is provided.
- Error context reporting the configured limit value.

docs/sandbox/guide.md
```diff
@@ -481,6 +481,28 @@ These options only matter when the runner is creating a fresh sandbox session:
 
 `concurrency_limits` controls how much sandbox materialization work can run in parallel. Use `SandboxConcurrencyLimits(manifest_entries=..., local_dir_files=...)` when large manifests or local directory copies need tighter resource control. Set either value to `None` to disable that specific limit.
 
+`archive_limits` controls SDK-side resource checks for `BaseSandboxSession.extract()` archive writes. The default, `archive_limits=None`, preserves the historical behavior and does not enforce SDK archive resource limits. Use `SandboxArchiveLimits()` to enable the SDK default thresholds, or pass explicit values such as `SandboxArchiveLimits(max_input_bytes=..., max_extracted_bytes=..., max_members=...)`. Set an individual field to `None` to disable only that limit.
```
seratch: Can you make this similar in the similar way with concurrency_limits, plus move to a separate PR? Merging docs immediately reflects the changes to the live documentation site, so we'd like to merge them after releasing the SDK changes.

seratch (changes requested):

src/agents/sandbox/session/archive_extraction.py
Aphroq: Updated in that direction.

This version replaces the hard-coded module globals with an explicit `SandboxArchiveLimits` dataclass, and appends `archive_limits: SandboxArchiveLimits | None = None` to `SandboxRunConfig` to avoid changing positional compatibility. `archive_limits=None` keeps the old behavior with no SDK archive resource limits; `SandboxArchiveLimits()` enables SDK defaults of `1 GiB` input, `4 GiB` extracted, and `100_000` members. Each field can also be set to `None` to disable only that limit.

The limits are now applied to runner-created, resumed, and injected sessions, and `BaseSandboxSession.extract(..., archive_limits=...)` supports a per-call override. I added coverage for the no-limit default, default thresholds, the three RunConfig session paths, per-call override, invalid non-positive values, field-level `None`, and error context reporting the configured limit.

docs/sandbox/guide.md
Aphroq: Handled. I removed the archive-limits docs from this PR. I followed up with a separate docs PR.

src/agents/sandbox/session/archive_extraction.py
```diff
@@ -302,15 +338,141 @@ def safe_zip_member_rel_path(member: zipfile.ZipInfo) -> Path | None:
...
+            parent_member = members_by_rel_path.get(parent)
+            if parent_member is not None and not parent_member.isdir():
+                raise UnsafeTarMemberError(
+                    member=member.name,
+                    reason=f"archive path descends through non-directory: {parent.as_posix()}",
+                )
+
+        if not member.isdir():
+            for existing_rel_path, existing_member in members_by_rel_path.items():
+                if _is_descendant(existing_rel_path, rel_path):
```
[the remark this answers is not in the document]
Aphroq: Handled.

The tar path still performs the first validation pass in streaming mode and does not reintroduce `getmembers()`. It now records a `descendant_by_parent_path` index while reading members, so a later regular file that would replace a previously seen directory prefix still raises the same `archive path descends through non-directory` error without scanning every previously seen member for each new file. This reduces the check from O(n²) to work proportional to member count and path depth.

Aphroq:
@codex review

seratch (approved):

Issue #3275, opened 2026-05-08
Chat Completions model should reject unsupported server-managed state


Aphroq:
### Please read this first

- **Have you read the docs?** Yes. The OpenAI migration docs state that Chat Completions callers must store and manage conversation context themselves, while `previous_response_id` is a Responses API state strategy.
- **Have you searched for related issues?** Yes. I searched upstream issues and PRs for `ChatCompletions previous_response_id conversation_id` and did not find a direct duplicate.

### Describe the bug

`OpenAIChatCompletionsModel.get_response()` and `stream_response()` accept `previous_response_id` and `conversation_id`, but the Chat Completions request does not send those parameters. The returned `ModelResponse.response_id` is also `None`.

If higher-level server-managed conversation configuration disables local session persistence, callers can believe server-side context is active while Chat Completions only receives the current input.

### Debug information

- Agents SDK version: `main@683b6e7`
- Python version: Python 3.12.1

### Repro steps

Run this from the repository root:

```python
import asyncio

import httpx
from openai.types.chat.chat_completion import ChatCompletion, Choice
from openai.types.chat.chat_completion_message import ChatCompletionMessage

from agents import ModelSettings, ModelTracing, OpenAIChatCompletionsModel


class DummyCompletions:
    def __init__(self):
        self.kwargs = None

    async def create(self, **kwargs):
        self.kwargs = kwargs
        return ChatCompletion(
            id="chatcmpl_123",
            created=0,
            model="fake",
            object="chat.completion",
            choices=[
                Choice(
                    index=0,
                    finish_reason="stop",
                    message=ChatCompletionMessage(role="assistant", content="ok"),
                )
            ],
        )


class DummyClient:
    def __init__(self, completions):
        self.chat = type("Chat", (), {"completions": completions})()
        self.base_url = httpx.URL("https://api.openai.com/v1/")


async def main():
    completions = DummyCompletions()
    model = OpenAIChatCompletionsModel(
        model="gpt-4",
        openai_client=DummyClient(completions),
    )
    await model.get_response(
        system_instructions=None,
        input="second turn only",
        model_settings=ModelSettings(),
        tools=[],
        output_schema=None,
        handoffs=[],
        tracing=ModelTracing.DISABLED,
        previous_response_id="resp_123",
        conversation_id=None,
        prompt=None,
    )
    print("previous_response_id sent:", "previous_response_id" in completions.kwargs)


asyncio.run(main())
```

Actual result on current `main`:

```text
[...]

PR #3283, opened 2026-05-08
fix: #3282 reject unsupported Chat Completions reusable prompts

### LINKED ISSUE #3282: Chat Completions model silently ignores reusable prompts
### Please read this first

- **Have you read the docs?** Yes. The reusable prompt docs describe the `prompt` parameter for the Responses API.

### Describe the bug

`OpenAIChatCompletionsModel.get_response()` and `stream_response()` accept a Responses reusable `prompt`, pass it through to `_fetch_response()`, and then silently drop it when building `chat.completions.create()` kwargs.

This makes prompt-managed configuration look accepted even though the Chat Completions backend never sends it to the provider. `OpenAIResponsesModel` does include `prompt` in its request path, so this is specific to the Chat Completions backend.

Related paths checked:

- Non-streaming and streaming Chat Completions both accept and drop `prompt`.
- `prompt=None` is the normal path and should stay unchanged.
- `OpenAIResponsesModel` already sends `prompt` explicitly.
- #3275 / #3279 cover unsupported server-managed state in Chat Completions, not reusable prompts.

### Debug information

- Agents SDK version: `main` at `683b6e79` (latest release tag checked locally: `v0.17.0`)

### Repro steps

Run this script against current `main`:

```python
import asyncio
from typing import Any

import httpx
from openai.types.chat.chat_completion import ChatCompletion, Choice
from openai.types.chat.chat_completion_message import ChatCompletionMessage

from agents import ModelSettings, ModelTracing, OpenAIChatCompletionsModel


class DummyCompletions:
    def __init__(self) -> None:
        self.kwargs: dict[str, Any] = {}

    async def create(self, **kwargs: Any) -> ChatCompletion:
        self.kwargs = kwargs
        return ChatCompletion(
            id="resp-id",
            created=0,
            model="fake",
            object="chat.completion",
            choices=[
                Choice(
                    index=0,
                    finish_reason="stop",
                    message=ChatCompletionMessage(role="assistant", content="ok"),
                )
            ],
        )


class DummyClient:
    def __init__(self, completions: DummyCompletions) -> None:
        self.chat = type("_Chat", (), {"completions": completions})()
        self.base_url = httpx.URL("https://api.openai.com/v1/")


async def main() -> None:
    completions = DummyCompletions()
        model="gpt-4",
        openai_client=DummyClient(completions),  # type: ignore[arg-type]
    )

    await model.get_response(
        system_instructions=None,
        input="hello",
        model_settings=ModelSettings(),
        tools=[],
        output_schema=None,
        handoffs=[],
        tracing=ModelTracing.DISABLED,
        previous_response_id=None,
        conversation_id=None,
        prompt={"id": "pmpt_123"},  # type: ignore[arg-type]
    )

    print("prompt" in completions.kwargs)
    print([key for key in completions.kwargs if "prompt" in key])


asyncio.run(main())
```

Actual output:

```text
False
[]
```

### Expected behavior

The Chat Completions backend should not silently accept and drop a reusable prompt. Since reusable prompts are a Responses API feature, Chat Completions should fail fast with a clear `UserError`, or otherwise explicitly map/send supported prompt fields if that behavior is intended.

Aphroq:
### Summary

Make `OpenAIChatCompletionsModel` reject reusable `prompt` values with a clear `UserError` instead of accepting and dropping them before `chat.completions.create()`.

This keeps the existing `prompt=None` path unchanged and covers both non-streaming and streaming Chat Completions calls. I also checked the adjacent Responses model path, which already sends `prompt`, and the related server-managed state issue #3275 / PR #3279, which is a separate unsupported-parameter case.

### Test plan

- `uv run pytest tests/models/test_openai_chatcompletions.py tests/models/test_openai_chatcompletions_stream.py -k "rejects_prompt"`

### Issue number

Closes #3282

### Checks

### DIFF
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -66,6 +66,16 @@ class OpenAIChatCompletionsModel(Model):
     def _supports_default_prompt_cache_key(self) -> bool:
         return ChatCmplHelpers.is_openai(self._get_client())
 
+    def _validate_prompt_is_supported(self, prompt: ResponsePromptParam | None) -> None:
+        if prompt is None:
+            return
+
+        raise UserError(
+            "Reusable prompts are only supported by the Responses API. "
+            "OpenAIChatCompletionsModel does not support `prompt`; use a Responses model "
+            "instead."
+        )
+
     def get_retry_advice(self, request: ModelRetryAdviceRequest) -> ModelRetryAdvice | None:
         return get_openai_retry_advice(request)
 
@@ -120,6 +130,8 @@ class OpenAIChatCompletionsModel(Model):
             previous_response_id=previous_response_id,
             conversation_id=conversation_id,
         )
+        self._validate_prompt_is_supported(prompt)
+
         with generation_span(
             model=str(self.model),
             model_config=model_settings.to_json_dict() | {"base_url": str(self._client.base_url)},
@@ -248,6 +260,8 @@ class OpenAIChatCompletionsModel(Model):
             previous_response_id=previous_response_id,
             conversation_id=conversation_id,
         )
+        self._validate_prompt_is_supported(prompt)
+
         with generation_span(
             model=str(self.model),
             model_config=model_settings.to_json_dict() | {"base_url": str(self._client.base_url)},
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
Can you resolve the conflicts?

Aphroq:
Resolved.

seratch (approved):

PR #3314, opened 2026-05-09
fix: #3313 align multi-choice chat streams with strict validation

### LINKED ISSUE #3313: Chat Completions streaming silently ignores extra choices
### Please read this first

- **Have you read the docs?** Yes. This concerns Chat Completions streaming conversion behavior.
- **Have you searched for related issues?** Yes. I searched upstream issues and PRs.

### Describe the bug

`ChatCmplStreamHandler.handle_stream()` reads only `chunk.choices[0]`. If a Chat Completions-compatible provider returns more than one streamed choice, or a caller passes `extra_args={"n": 2}`, content/tool-call deltas for choices after index 0 are silently ignored.

Impact:

- A model can return multiple choices, but only the first choice reaches the generated `Response`.
- Tool calls in nonzero choices can be dropped without an error.
- A chunk that contains only `index=1` can be processed as if it were the primary choice, which mislabels an alternate completion as the single response.

Relevant edge cases considered:

- Chunks with `choices=[]` should still be allowed for usage/final metadata.
- A chunk containing both index 0 and index 1 should fail explicitly.
- A chunk containing only a nonzero choice index should also fail explicitly.
- The SDK currently exposes one streamed `Response`, so merging multiple alternative choices into one response would change semantics. Requiring `n=1` is safer than silently dropping or merging alternatives.

### Debug information

### Repro steps

Run this minimal script from a repository checkout.

```python
import asyncio

from openai.types.chat.chat_completion_chunk import ChatCompletionChunk, Choice, ChoiceDelta

from agents.models.chatcmpl_stream_handler import ChatCmplStreamHandler


async def fake_stream():
    yield ChatCompletionChunk(
        id="chunk-id",
        created=1,
        model="fake",
        object="chat.completion.chunk",
        choices=[
            Choice(index=0, delta=ChoiceDelta(content="first")),
            Choice(index=1, delta=ChoiceDelta(content="second")),
        ],
    )


async def main() -> None:
    response = Response(
        id="resp-id",
        created_at=0,
        model="fake-model",
        object="response",
        output=[],
        tool_choice="none",
        tools=[],
        parallel_tool_calls=False,
    )
    text_deltas = []
    completed_texts = []
    async for event in ChatCmplStreamHandler.handle_stream(response, fake_stream()):
        if event.type == "response.output_text.delta":
            text_deltas.append(event.delta)
        elif event.type == "response.completed":
            for item in event.response.output:
                if item.type == "message":
                    completed_texts.extend(part.text for part in item.content)

    print(f"text_deltas: {text_deltas}")
    print(f"completed_texts: {completed_texts}")


asyncio.run(main())
```

Actual result on upstream `main` at `bc3607ba`:

```text
text_deltas: ['first']
completed_texts: ['first']
```

### Expected behavior

The stream handler should not silently ignore additional choices. Until the SDK defines multi-choice streaming semantics, it should raise a clear error when streamed chunks contain multiple choices or a nonzero choice index, and callers should use `n=1`.

Aphroq:
### Summary

- Align Chat Completions streaming multi-choice handling with `strict_feature_validation`.
- In strict mode, raise `UserError` when streamed chunks contain multiple choices or nonzero choice indexes.
- In the default compatibility path, process only `choice.index == 0`, ignore the other choices with a one-time warning, and keep `choices=[]` usage-only chunks valid.
- Add regression coverage for strict mode, default filtering, empty-choice usage chunks, and the model-to-handler strict flag wiring.

### Test plan

- `uv run pytest tests/models/test_openai_chatcompletions_stream.py -k "multiple_choices or nonzero_choice or empty_choice or strict_validation"`
- `uv run pytest tests/models/test_openai_chatcompletions_stream.py`
- `uv run pytest tests/models/test_litellm_chatcompletions_stream.py tests/models/test_any_llm_model.py::test_any_llm_stream_uses_chat_handler_when_responses_are_unsupported`

### Issue number

Closes #3313

### Checks

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -42,7 +42,9 @@ from openai.types.responses.response_reasoning_text_done_event import (
 )
 from openai.types.responses.response_usage import InputTokensDetails, OutputTokensDetails
 
+from ..exceptions import UserError
 from ..items import TResponseStreamEvent
+from ..logger import logger
 from .chatcmpl_helpers import ChatCmplHelpers
 from .fake_id import FAKE_RESPONSES_ID
 
@@ -70,6 +72,7 @@ class StreamingState:
     thinking_signature: str | None = None
     # Store provider data for all output items
     provider_data: dict[str, Any] = field(default_factory=dict)
+    has_warned_unsupported_choice: bool = False
 
 
 class SequenceNumber:
@@ -263,6 +266,7 @@ class ChatCmplStreamHandler:
         response: Response,
         stream: AsyncStream[ChatCompletionChunk],
         model: str | None = None,
+        strict_feature_validation: bool = False,
     ) -> AsyncIterator[TResponseStreamEvent]:
         """
         Handle a streaming chat completion response and yield response events.
@@ -291,7 +295,33 @@ class ChatCmplStreamHandler:
             if hasattr(chunk, "usage") and chunk.usage is not None:
                 usage = chunk.usage
 
-            if not chunk.choices or not chunk.choices[0].delta:
+            if not chunk.choices:
+                continue
+
+            unsupported_choice_indexes = [
+                choice.index for choice in chunk.choices if choice.index != 0
+            ]
+            if len(chunk.choices) > 1 or unsupported_choice_indexes:
+                message = (
+                    "Chat Completions streaming with multiple choices or nonzero choice indexes "
+                    "is not fully supported; only choice index 0 can be processed."
+                )
+                if strict_feature_validation:
+                    raise UserError(message)
+
+                if not state.has_warned_unsupported_choice:
+                    logger.warning(
+                        "%s Ignoring the other choices; enable strict feature validation to "
+                        "raise an error instead.",
+                        message,
+                    )
+                    state.has_warned_unsupported_choice = True
+
+            choice = next((choice for choice in chunk.choices if choice.index == 0), None)
+            if choice is None:
+                continue
+
+            if not choice.delta:
                 continue
 
             # Build provider_data for non-OpenAI Responses API endpoints format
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
Thanks for adding the guard here. I agree this is a real scenario: with Chat Completions streaming and `n > 1`, the API can emit separate chunks for `choice.index == 1`, so the current handler can incorrectly mix multiple choices into one streamed response.

That said, unconditional rejection changes existing behavior for users passing `n` through `ModelSettings.extra_args`. Could we align this with the `strict_feature_validation` approach used for similar Chat Completions compatibility gaps?

Suggested behavior:
- When `strict_feature_validation=True`, raise `UserError` for multi-choice streaming / nonzero choice indexes.
- When `strict_feature_validation=False` (default), preserve compatibility by processing only `choice.index == 0` and ignoring the other choices, ideally with a one-time warning.
- Keep `choices=[]` chunks allowed for usage-only stream chunks.

Implementation-wise, this likely means passing `self._strict_feature_validation` from `OpenAIChatCompletionsModel.stream_response()` into `ChatCmplStreamHandler.handle_stream(...)`, then filtering/rejecting choices there. This keeps the default path closer to the current non-streaming Chat Completions behavior, which uses `choices[0]`, while still giving strict users a fail-fast error for unsupported multi-choice streaming.

Aphroq:
Thanks, that direction makes sense. I updated the PR to follow the `strict_feature_validation` behavior:

- With `strict_feature_validation=True`, multi-choice chunks or nonzero choice indexes now raise `UserError`.
- By default, the handler keeps compatibility by processing only `choice.index == 0`, ignoring the other choices, and warning once per stream.
- `choices=[]` usage-only chunks are still allowed.

I also added coverage for multi-choice chunks, chunks that only contain a nonzero choice, empty-choice usage chunks, passing the strict flag from the model entry point into the handler, and the related LiteLLM/AnyLLM paths that reuse the same handler. This still leaves full `n > 1` multi-response streaming semantics out of scope; the PR just avoids mixing a nonzero choice into the single streamed response.

seratch:
@codex review

seratch (approved):

PR #3320, opened 2026-05-09
fix: #3319 preserve nested handoff history content

### LINKED ISSUE #3319: Nested handoff history flattening drops multiline and structured content
### Please read this first

- **Have you read the docs?** Yes. This is about the SDK's `nest_handoff_history` behavior during chained handoffs.
- **Have you searched for related issues?** Yes. No duplicate was found. 
### Describe the bug

When a `nest_handoff_history` summary is passed through another nested handoff, `_extract_nested_history_transcript` reconstructs the prior transcript by splitting the summary block into lines and parsing `role: content` text.

That parsing loses information:

- Multiline string content is truncated to the first line.
- Structured role content such as a list of input content parts is converted into a plain string.

Impact: chained handoffs and custom `handoff_history_mapper` implementations can receive a corrupted transcript. User content after the first newline is dropped, and structured input parts such as multimodal content descriptors no longer have their original shape.

### Debug information

### Repro steps

Run this script against current upstream `main`:

```python

import json

from agents.handoffs import HandoffInputData, nest_handoff_history
from agents.items import TResponseInputItem


def capture_second_pass(item: TResponseInputItem) -> list[TResponseInputItem]:
    captured: list[TResponseInputItem] = []

    def mapper(transcript: list[TResponseInputItem]) -> list[TResponseInputItem]:
        captured.extend(transcript)
        return transcript

    first = HandoffInputData(input_history=(item,), pre_handoff_items=(), new_items=())
    first_nested = nest_handoff_history(first)
    second = HandoffInputData(
        input_history=first_nested.input_history,
        pre_handoff_items=(),
        new_items=(),
    )
    nest_handoff_history(second, history_mapper=mapper)
    return captured


multiline = capture_second_pass(
    {"role": "user", "content": "first line\nsecond line"}
)
structured = capture_second_pass(
    {
        "role": "user",
        "content": [
            {"type": "input_text", "text": "look at this"},
            {"type": "input_image", "image_url": "https://example.com/image.png"},
        ],
    }
)

print("multiline:")
print(json.dumps(multiline, ensure_ascii=False, indent=2))
print("structured_content_type:", type(structured[0].get("content")).__name__)
print("structured:")
print(json.dumps(structured, ensure_ascii=False, indent=2))
```

Actual result:

```text
multiline:
[
  {
    "role": "user",
    "content": "first line"
  }
]
structured_content_type: str
structured:
[
  {
    "role": "user",
    "content": "[{\"type\": \"input_text\", \"text\": \"look at this\"}, {\"type\": \"input_image\", \"image_url\": \"https://example.com/image.png\"}]"
  }
]
```

### Expected behavior

Flattening an already nested handoff summary should preserve transcript item content across chained handoffs:

- Multiline content should remain `"first line\nsecond line"`.
- Structured content should remain a list, not be converted to a string.
- Legacy simple summary lines should continue to parse for existing saved histories.

Aphroq:
### Summary

- Preserve multiline message content when flattening an already nested handoff-history summary.
- Preserve structured role content by serializing those summary entries as JSON and parsing them back before the next history mapper runs.
- Keep the existing readable `role: content` format for simple string messages, and keep legacy summary parsing for older saved histories.

### Test plan

### Issue number

Closes #3319

### Checks

### DIFF
--- a/src/agents/handoffs/history.py
+++ b/src/agents/handoffs/history.py
@@ -162,9 +162,13 @@ def _format_transcript_item(item: TResponseInputItem) -> str:
     role = item.get("role")
     if isinstance(role, str):
         content = item.get("content")
-        if content is None or isinstance(content, str):
+        if content is None or (isinstance(content, str) and not _contains_newline(content)):
             return _format_transcript_item_legacy(item)
     return _format_transcript_item_json(item)
+
+
+def _contains_newline(value: str) -> bool:
+    return "\n" in value or "\r" in value
 
 
 def _format_transcript_item_json(item: TResponseInputItem) -> str:
--- a/src/agents/handoffs/history.py
+++ b/src/agents/handoffs/history.py
@@ -161,6 +161,24 @@ def _build_summary_message(transcript: list[TResponseInputItem]) -> TResponseInp
 def _format_transcript_item(item: TResponseInputItem) -> str:
     role = item.get("role")
     if isinstance(role, str):
+        content = item.get("content")
+        if content is None or isinstance(content, str):
+            return _format_transcript_item_legacy(item)
+    return _format_transcript_item_json(item)
+
+
+def _format_transcript_item_json(item: TResponseInputItem) -> str:
+    payload = cast(dict[str, Any], deepcopy(item))
+    payload.pop("provider_data", None)
+    try:
+        return json.dumps(payload, ensure_ascii=False, default=str)
+    except (TypeError, ValueError):
+        return _format_transcript_item_legacy(item)
+
+
+def _format_transcript_item_legacy(item: TResponseInputItem) -> str:
+    role = item.get("role")
+    if isinstance(role, str):
         prefix = role
         name = item.get("name")
         if isinstance(name, str) and name:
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the fix. I agree this is a real scenario: chained `nest_handoff_history` can receive multiline user content or structured/multimodal content, and preserving that transcript shape is important.

Before merging, could you make one small follow-up change? The new record splitter still treats a continuation line like `2. not a new record` as a new summary item, so multiline string content can still be truncated when a later line happens to look like a numbered summary record.

Could we serialize role items with string content that contains newlines using the JSON record format as well? That should remove the ambiguity for newly generated summaries while keeping the legacy parser for older saved histories. Please also add a regression test for a multiline string where the second line starts with something like `2. not a new record`.

Aphroq:
Thanks for the review.

I changed the formatter so role items with string content that contains `\n` or `\r` now use the JSON record format too. Simple one-line string messages still keep the readable `role: content` format, and the legacy parser still handles older saved summaries.

I also updated the multiline regression test to use `first line\n2. not a new record`, so the new summary path no longer depends on the record splitter guessing whether that second line is content or a new item.

I checked the adjacent cases as well: structured/multimodal role content still round-trips through JSON, legacy multiline summaries still parse, role+name formatting is preserved for simple strings, non-role items still serialize as JSON, `provider_data` is still stripped, and the existing custom-wrapper/unserializable fallback tests still pass.

seratch (approved):

PR #3326, opened 2026-05-10
fix: include sandbox provider error details

seratch:
This pull request fixes sandbox provider diagnostics so transport and startup failures include the provider error details that are reliably available from each backend.

It preserves existing exception types while allowing structured sandbox errors to carry clearer messages, expands Cloudflare HTTP/body diagnostics, and adds targeted Daytona, Modal, and Blaxel error summaries for exec, PTY startup, and workspace-start paths. Regression tests cover the visible messages and structured context for the updated providers.


seratch:
@codex review

PR #3338, opened 2026-05-10
fix: #3337 validate RunState approval decision shapes

### LINKED ISSUE #3337: RunState approval restore accepts invalid decision shapes
### Please read this first

- **Have you read the docs?** Yes. This is about `RunState` deserialization integrity rather than a documented usage option.
- **Have you searched for related issues?** Yes. #2658 / #2660 are adjacent because they established durable approval rejection-message state, but they do not cover invalid serialized approval decision shapes. I did not find an issue or PR for this validation gap.

### Describe the bug

`RunState.from_json()` restores serialized approval records without validating the shape of the `approved` and `rejected` fields. The SDK serializer emits those fields as either a boolean or a list of call IDs, but corrupted or externally produced JSON can provide a string, mapping, `None`, or a mixed list.

Today those invalid values are accepted silently. For example, a serialized rejection can be changed from `["cid2"]` to `"cid2"`, and restore succeeds while the approval status becomes pending (`None`) instead of rejected.

### Debug information

- Agents SDK version: upstream `main` at `479640e2`

### Repro steps

Run this from a checkout of upstream `main` at `479640e2`:

```python
import asyncio

from agents import Agent
from agents.exceptions import UserError
from agents.items import ToolApprovalItem
from agents.run_state import RunState
from openai.types.responses import ResponseFunctionToolCall


async def main():
    agent = Agent(name="ApprovalShapeAgent")
    state = RunState(
        context=RunContextWrapper(context={}),
        original_input="test",
        starting_agent=agent,
    )
    approval_item = ToolApprovalItem(
        agent=agent,
        raw_item=ResponseFunctionToolCall(
            type="function_call",
            name="tool2",
            call_id="cid2",
            status="completed",
            arguments="",
        ),
    )
    state.reject(approval_item, rejection_message="Denied by reviewer")
    data = state.to_json()
    data["context"]["approvals"]["tool2"]["rejected"] = "cid2"

    try:
        restored = await RunState.from_json(agent, data)
    except UserError as exc:
        print(f"raised UserError: {exc}")
    else:
        status = restored._context.is_tool_approved("tool2", "cid2")
        print(f"restored without error; approval status: {status!r}")


asyncio.run(main())
```

Actual result:

```text
restored without error; approval status: None
```

### Expected behavior

`RunState.from_json()` should reject malformed approval decision fields instead of silently restoring impossible approval state. `approved` and `rejected` should be accepted only when they are booleans or lists of string call IDs.

Aphroq:
### Summary

- Validate serialized approval records when rebuilding `RunContextWrapper` approvals from `RunState`.
- Accept only boolean decisions or lists of string call IDs for `approved` and `rejected`.
- Add regression coverage for malformed approval decision shapes.

### Test plan

### Issue number

Closes #3337

### Checks

### DIFF
--- a/src/agents/run_context.py
+++ b/src/agents/run_context.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Mapping
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Generic
 
@@ -12,6 +13,7 @@ from ._tool_identity import (
     is_reserved_synthetic_tool_namespace,
     tool_qualified_name,
 )
+from .exceptions import UserError
 from .usage import Usage
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


Aphroq:
@codex review

seratch:
Thanks for sharing this patch. However, I think this scenario is unlikely to occur in practice. If we want to validate serialized data in this way, we should probably introduce a more comprehensive validation mechanism rather than adding isolated if/else checks. For this reason, let’s close this PR and the referenced issue.

PR #3339, opened 2026-05-10
fix: #781 replace assertion in handoff() with UserError

### LINKED ISSUE #781: Misleading Assertion Logic and Message for on_handoff/input_type in `handoff()` function at `agents/handoffs.py`
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/) : `Yes`
- **Have you searched for related issues?** Others may have had similar requests: `Yes`

### Question   

**Description of the Issue:**

In the `agents/handoffs.py`, there's ae `assert` statement in `handoff()` function related to the `on_handoff` and `input_type` parameters that appears to be logically flawed and potentially misleading:

```python
assert (on_handoff and input_type) or not (on_handoff and input_type), (
    "You must provide either both on_input and input_type, or neither"
)
```

There are two main concerns with this assertion as presented:

1.  **Logical Tautology:** The condition `(on_handoff and input_type) or not (on_handoff and input_type)` can be simplified to `P or not P` (where `P = on_handoff and input_type`). This is a tautology, meaning it always evaluates to `True`, regardless of the truthiness of `on_handoff` or `input_type`. Consequently, this `assert` statement will never fail and does not enforce any conditions.

2.  **Contradictory Message:** The assertion message states, *"You must provide either both on_input and input_type, or neither"* (assuming "on_input" is a typo for "on_handoff"). This message suggests a strict requirement that `on_handoff` and `input_type` must either both be present or both be absent. This is in direct contradiction to the assertion's condition, which, being always true, places no restrictions.

**Framing the Concern:**

If the use of a tautological (always true) condition in the assertion is intentional – perhaps to signify that, for the purpose of that example, the SDK imposes no specific co-occurrence requirements on `on_handoff` and `input_type` (i.e., allowing flexible combinations such as one parameter being provided without the other) – then the accompanying message *"You must provide either both on_handoff and input_type, or neither"* becomes highly contradictory and misleading.

This interpretation suggests that the SDK might indeed intend to be flexible regarding these parameters. My own testing confirms this flexibility: for instance, if `input_type` is not passed to the `handoff()` function, and the `on_handoff` callback's signature is adjusted to not expect the corresponding parsed `input_data`, there is no operational issue. This practical behavior aligns with the idea that the SDK allows providing `on_handoff` without a corresponding `input_type` in certain valid scenarios.

The current state of this example assertion can therefore cause significant confusion for users trying to understand the correct and valid usage patterns for `on_handoff` and `input_type`.

**Expected Behavior:**

The code should clearly and accurately reflect the SDK's behavior and requirements. An `assert` statement should either:
* Correctly enforce a relevant condition being demonstrated.
* Be accompanied by explanatory text that clarifies its purpose or limitations if it's illustrative of a partial concept.
* The message of an assertion should accurately reflect the logic of its condition.

**Actual Behavior:**

The `assert` statement:
* Does not enforce any condition due to its tautological logic.
* Presents a message that implies a strict rule ("both or neither").
* This rule from the message is not universally true for the SDK's `handoff()` function, which offers more flexibility.

**Location:**

This assertion statement is found at line 170 in `handoff()` function and in the `agents/handoffs.py` within the `openai-agents-python` repository.

**Suggested Actions:**

To mitigate confusion, I suggest the following:

1.  **If the intent is to demonstrate a specific rule:** The assertion logic should be corrected to accurately implement that rule (e.g., `assert bool(on_handoff_callback) == bool(input_model)` if the "both or neither" rule is indeed what the code aims for a specific scenario).
2.  **If the intent is to show that the parameters are optional or have nuanced dependencies:** The assertion might be misleading. It would be better to replace it with clear explanatory text detailing the valid combinations of `on_handoff` and `input_type`, including when `input_type` is essential for an `on_handoff` callback that expects structured data.
3.  **At a minimum:** The current tautological assertion and its contradictory message should be revised to prevent user confusion.

Clarifying the relationship and valid usage patterns of `on_handoff` and `input_type` in the core documentation would also be beneficial.

Thank you for considering this issue.

0xSudoSSH:
### Summary

Replace tautological assertion in `handoff()` with proper `UserError` validation.

**Two problems with the original assertion:**

1. **Tautological logic** — `(X) or not (X)` is always `True`, so the assertion never fires and enforces nothing.

2. **Incorrect error message** — The message says "You must provide either both on_handoff and input_type, or neither", but `on_handoff` without `input_type` is a valid usage (simple callback that takes only context). The only invalid combination is `input_type` without `on_handoff`, since there would be no callback to receive the parsed input.

**Before (tautology — never fires):**
```python
assert (on_handoff and input_type) or not (on_handoff and input_type), (
    "You must provide either both on_input and input_type, or neither"
)
```

**After (correct validation):**
```python
if input_type is not None and on_handoff is None:
    raise UserError("input_type requires on_handoff to be provided")
```

**Valid combinations:**

| `on_handoff` | `input_type` | Valid? |
|---|---|---|
| None | None | ✅ No callback |
| Provided (1 arg) | None | ✅ Simple callback, context only |
| Provided (2 args) | Provided | ✅ Typed callback with parsed input |
| None | Provided | ❌ No callback to receive the parsed input — caught by this fix |

Note: downstream checks at lines 259–272 already catch most misuse (e.g., `assert callable(on_handoff)` and signature length validation), so the tautological assertion was effectively dead code. This fix replaces it with a clear, early validation that produces a helpful error message.

### Test plan

- Added `test_input_type_without_on_handoff_raises_error` to verify the fix
- All 29 existing handoff tests pass (`tests/test_handoff_tool.py`, `tests/test_handoff_prompt.py`, `tests/test_handoff_history_duplication.py`)
- `make format`, `make lint`, `make typecheck` — all pass

### Issue number

Closes #781

### Checks

### DIFF
--- a/src/agents/handoffs/__init__.py
+++ b/src/agents/handoffs/__init__.py
@@ -252,12 +252,12 @@ def handoff(
             hidden from the LLM at runtime.
     """
 
-    assert (on_handoff and input_type) or not (on_handoff and input_type), (
-        "You must provide either both on_handoff and input_type, or neither"
-    )
+    if input_type is not None and on_handoff is None:
+        raise UserError("You must provide on_handoff when input_type is provided")
     type_adapter: TypeAdapter[Any] | None
     if input_type is not None:
-        assert callable(on_handoff), "on_handoff must be callable"
+        if not callable(on_handoff):
+            raise UserError("on_handoff must be callable")
         sig = inspect.signature(on_handoff)
         if len(sig.parameters) != 2:
             raise UserError("on_handoff must take two arguments: context and input")
--- a/src/agents/realtime/handoffs.py
+++ b/src/agents/realtime/handoffs.py
@@ -92,7 +92,8 @@ def realtime_handoff(
         raise UserError("You must provide on_handoff when input_type is provided")
     type_adapter: TypeAdapter[Any] | None
     if input_type is not None:
-        assert callable(on_handoff), "on_handoff must be callable"
+        if not callable(on_handoff):
+            raise UserError("on_handoff must be callable")
         sig = inspect.signature(on_handoff)
         if len(sig.parameters) != 2:
             raise UserError("on_handoff must take two arguments: context and input")
--- a/tests/realtime/test_realtime_handoffs.py
+++ b/tests/realtime/test_realtime_handoffs.py
@@ -137,6 +137,14 @@ def test_realtime_handoff_input_type_requires_on_handoff():
         realtime_handoff(rt, input_type=int)  # type: ignore[call-overload]
 
 
+def test_realtime_handoff_non_callable_on_handoff_raises_error():
+    """Providing a non-callable on_handoff with input_type should raise UserError."""
+    rt = RealtimeAgent(name="x")
+
+    with pytest.raises(UserError, match="on_handoff must be callable"):
+        realtime_handoff(rt, on_handoff="not_a_function", input_type=int)  # type: ignore[call-overload]
+
+
 @pytest.mark.asyncio
 async def test_realtime_handoff_missing_input_json_raises_model_error():
     rt = RealtimeAgent(name="x")
--- a/tests/test_handoff_tool.py
+++ b/tests/test_handoff_tool.py
@@ -252,6 +252,30 @@ async def test_invalid_on_handoff_raises_error():
         handoff(agent, on_handoff=_on_handoff)  # type: ignore
 
 
+def test_input_type_without_on_handoff_raises_error():
+    """Providing input_type without on_handoff should raise an error."""
+
+    class MyInput(BaseModel):
+        reason: str
+
+    agent = Agent(name="test")
+
+    with pytest.raises(UserError, match="You must provide on_handoff when input_type is provided"):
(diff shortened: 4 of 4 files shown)


seratch (changes requested):
Thanks for the fix. The direction looks good to me, especially because it brings regular `handoff()` closer to the existing Realtime handoff validation.

Before merging, could you please tighten this up by making the user-facing validation consistent across the full path?

Specifically, please replace the adjacent `assert callable(on_handoff)` with a `UserError` as well, and apply the same treatment to `realtime_handoff()` so regular handoffs and Realtime handoffs behave consistently. Since these checks validate user-provided arguments, they should not rely on `assert`, which can be disabled with Python optimization flags.

It would also be good to use the same error wording in both places, for example:

```python
raise UserError("You must provide on_handoff when input_type is provided")
```

and a corresponding `UserError` message for the non-callable case.

seratch:
Thanks for updating the PR. I will merge other PRs before yours, but no need to rebase this (it triggers a new code review and CI builds for this relatively simple change).

seratch:
@codex review

seratch (approved):

PR #3341, opened 2026-05-10
fix: #3267 preserve required hosted tool IDs in OpenAI conversation sessions

### LINKED ISSUE #3267: OpenAIConversationsSession strips required id from file_search_call before conversations.items.create, causing 400 missing items[0].id
### Please read this first

I can’t post to GitHub from here, but I’ve got a ready-to-file issue.

Issue body:

```md
## Summary

When using `OpenAIConversationsSession`, hosted tool output items such as `file_search_call` arrive from the SDK/model with a valid `id`, but the session persistence path strips that `id` before calling `conversations.items.create(...)`.

That produces a 400 from the Conversations API:

```text
Missing required parameter: 'items[0].id'
```

This appears to happen in the `openai-agents` session persistence layer, not in app-side serialization.

## Environment

- `openai-agents==0.17.0`
- `openai==2.34.0`
- Windows
- Using `OpenAIConversationsSession`
- Hosted `file_search` tool enabled

## Actual behavior

A run completes successfully, but persistence into the Conversations session fails with:

```text
BadRequestError: Error code: 400 - {
  "error": {
    "message": "Missing required parameter: 'items[0].id'.",
    "param": "items[0].id",
    "code": "missing_required_parameter"
  }
}
```

The rejected payload reaching `add_items(...)` looks like this:

```json
[
  {
    "queries": [
      "review this file, and please let me know what its about",
      "summarize the uploaded file and describe what it is about",
      "main topic and purpose of the uploaded document"
    ],
    "status": "completed",
    "type": "file_search_call",
    "results": null
  },
  {
    "content": [...],
    "role": "assistant",
    "status": "completed",
    "type": "message",
    "phase": "final_answer"
  }
]
```

Notice that the `file_search_call` item has no `id` by the time it is persisted.

## Expected behavior

If the raw output item has an `id`, the `id` should still be present when persisting that item into `OpenAIConversationsSession`, especially for item types where the Conversations API requires it.

## Evidence: raw SDK events do contain the id

From the raw streamed SDK events, the `file_search_call` item clearly has an `id`:

### `response.output_item.done`

```json
{
  "item": {
    "id": "fs_0e9f7c7c4bcd09e60069fdc26e25d88192aeb641a50f296f6f",
    "queries": [
      "review this file, and please let me know what its about",
      "summarize the uploaded file and describe what it is about",
      "main topic and purpose of the uploaded document"
    ],
    "status": "completed",
    "type": "file_search_call",
    "results": null
  },
  "type": "response.output_item.done"
}
```

### `response.completed`

```json
"output": [
  {
    "id": "fs_0e9f7c7c4bcd09e60069fdc26e25d88192aeb641a50f296f6f",
    "queries": [...],
    "status": "completed",
    "type": "file_search_call",
    "results": null
  },
  {
    "id": "msg_0e9f7c7c4bcd09e60069fdc27031e881929b7d866725444e12",
    "type": "message",
    ...
  }
]
```

So the `id` exists in the raw SDK/model output and is lost later.

## Root cause in `openai-agents`

I traced the persistence path in `openai-agents==0.15.2`:

1. Run items are converted to input items via `model_dump(exclude_unset=True)`:
   - `agents/items.py`
   - `RunItemBase.to_input_item()`

2. Session persistence then sanitizes items before saving:
   - `agents/run_internal/session_persistence.py`

3. In `_sanitize_openai_conversation_item(...)`, the `id` is explicitly removed:

```python
def _sanitize_openai_conversation_item(item):
    if isinstance(item, dict):
        clean_item = strip_internal_input_item_metadata(item)
        clean_item.pop("id", None)
        clean_item.pop("provider_data", None)
        return clean_item
```

4. Those sanitized items are then passed to `session.add_items(items_to_save)`, which eventually calls:

```python
await self._openai_client.conversations.items.create(
    conversation_id=session_id,
    items=items,
)
```

Because `id` was stripped earlier, the Conversations API rejects the hosted tool item.

## Why this looks incorrect

Stripping `id` may make sense for some replay/history normalization cases, but it appears too broad for `OpenAIConversationsSession` persistence of hosted output items like `file_search_call`, where the Conversations API expects `id` to still be present.

## Minimal repro idea

- Create an agent with hosted `file_search`
- Use `OpenAIConversationsSession`
- Upload/index a file
- Ask a question that triggers `file_search`
- Let the run complete
- Persistence fails on `conversations.items.create(...)` with missing `items[0].id`

## Workaround

I worked around this on the app side by wrapping `OpenAIConversationsSession.add_items(...)` and filtering/dropping invalid items so the rest of the turn can still persist, but that is only a mitigation.

## Request

Please adjust the OpenAI Conversations persistence sanitization so it does not blindly remove `id` from output items that require it for `conversations.items.create(...)`.

At minimum, `file_search_call` should preserve its `id`. More broadly, `_sanitize_openai_conversation_item(...)` may need item-type-aware behavior instead of unconditional `clean_item.pop("id", None)`.
```

The exact code points from your installed package that support the issue are:

- `agents/run_internal/session_persistence.py:320`
- `agents/run_internal/session_persistence.py:341`
- `agents/run_internal/session_persistence.py:568`
- `agents/items.py:151`

And your local raw evidence is in:

- [sdk_events.jsonl](D:/repos/langchain_master/debug/sdk_events.jsonl)

If you want, I can also turn this into a shorter GitHub-ready version with just title, repro, expected, actual, and root cause.

seratch:
This pull request fixes #3267 OpenAIConversationsSession persistence for hosted tool-call items whose top-level `id` is required by the Conversations API.

When a Responses run emits items such as `file_search_call`, the session persistence sanitizer previously stripped `id` before calling `conversations.items.create()`, causing the API to reject the saved turn with `Missing required parameter: 'items[0].id'`. The fix keeps IDs only for Conversations item types that require them, while continuing to strip optional or policy-controlled IDs from messages, function calls, reasoning items, and tool outputs to avoid replaying stale provider IDs.

The change also adds focused regression coverage for required-ID item types, optional-ID item types, provider data stripping, and non-dict passthrough behavior.

### DIFF
--- a/src/agents/run_internal/session_persistence.py
+++ b/src/agents/run_internal/session_persistence.py
@@ -561,14 +561,42 @@ def _ignore_ids_for_matching(session: Session) -> bool:
     )
 
 
+_OPENAI_CONVERSATION_ITEM_TYPES_WITH_REQUIRED_ID: frozenset[str] = frozenset(
+    {
+        "file_search_call",
+        "web_search_call",
+        "computer_call",
+        "code_interpreter_call",
+        "image_generation_call",
+        "local_shell_call",
+        "local_shell_call_output",
+        "mcp_list_tools",
+        "mcp_approval_request",
+        "mcp_call",
+        "item_reference",
+    }
+)
+
+
 def _sanitize_openai_conversation_item(item: TResponseInputItem) -> TResponseInputItem:
-    """Remove provider-specific fields before fingerprinting or persistence."""
+    """Remove provider-specific fields before fingerprinting or persistence.
+
+    Some Responses input item types require their server-assigned ``id`` when they are
+    persisted through the Conversations API. Other item IDs remain stripped so replayed
+    messages, reasoning items, and function calls do not carry stale provider IDs.
+    """
     if isinstance(item, dict):
         clean_item = cast(dict[str, Any], strip_internal_input_item_metadata(item))
-        clean_item.pop("id", None)
+        if not _openai_conversation_item_requires_id(clean_item):
+            clean_item.pop("id", None)
         clean_item.pop("provider_data", None)
         return cast(TResponseInputItem, clean_item)
     return item
+
+
+def _openai_conversation_item_requires_id(item: dict[str, Any]) -> bool:
+    """Return whether the Conversations create-item schema requires this item's top-level ID."""
+    return item.get("type") in _OPENAI_CONVERSATION_ITEM_TYPES_WITH_REQUIRED_ID
 
 
 def _sanitize_openai_conversation_history_items_for_model_input(
--- a/tests/memory/test_session_persistence_sanitize.py
+++ b/tests/memory/test_session_persistence_sanitize.py
@@ -1,0 +1,108 @@
+from __future__ import annotations
+
+from typing import Any, cast
+
+import pytest
+
+from agents.items import TResponseInputItem
+from agents.run_internal.session_persistence import _sanitize_openai_conversation_item
+
+
+def _sanitize(item: dict[str, Any]) -> dict[str, Any]:
+    return cast(dict[str, Any], _sanitize_openai_conversation_item(cast(TResponseInputItem, item)))
+
+
+@pytest.mark.parametrize(
+    "item_type",
+    [
+        "file_search_call",
+        "web_search_call",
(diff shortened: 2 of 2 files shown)


seratch:
@codex review

PR #3344, opened 2026-05-11
Validate prompt configs before model dispatch

FU-max-boop:
## Summary
- validate static and dynamic prompt configs before converting them to model input
- reject missing/non-string prompt ids and invalid version/variables fields with UserError
- apply the same validation when realtime agents merge prompt configs into session model settings

## Tests
- uv run pytest tests/test_agent_prompt.py tests/realtime/test_realtime_model_settings.py -q
- uv run ruff format --check src/agents/prompts.py src/agents/realtime/openai_realtime.py src/agents/realtime/session.py tests/test_agent_prompt.py tests/realtime/test_realtime_model_settings.py
- uv run ruff check src/agents/prompts.py src/agents/realtime/openai_realtime.py src/agents/realtime/session.py tests/test_agent_prompt.py tests/realtime/test_realtime_model_settings.py
- uv run pyright --project pyrightconfig.json src/agents/prompts.py src/agents/realtime/openai_realtime.py src/agents/realtime/session.py tests/test_agent_prompt.py tests/realtime/test_realtime_model_settings.py

### DIFF
--- a/src/agents/prompts.py
+++ b/src/agents/prompts.py
@@ -48,8 +48,20 @@ DynamicPromptFunction = Callable[[GenerateDynamicPromptData], MaybeAwaitable[Pro
 """A function that dynamically generates a prompt."""
 
 
-def _coerce_prompt_dict(prompt: Prompt | dict[object, object]) -> Prompt:
-    """Convert a runtime-validated prompt dict into the Prompt TypedDict view."""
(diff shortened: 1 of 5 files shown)


seratch:
Thanks for sending this patch. When using a prompt object, users are expected to understand what data is required. Also, if we add this type of runtime validation for objects here, the SDK would need to do the same in many other places across the codebase. For these reasons, we do not plan to include this change.

PR #3349, opened 2026-05-11
fix: #3348 make AdvancedSQLiteSession add_items atomic

### LINKED ISSUE #3348: AdvancedSQLiteSession.add_items can report success after structure metadata failure
### Please read this first

### Describe the bug

### Debug information

### Repro steps

```python
import asyncio
import sqlite3

        raw_count = conn.execute(
            (session.session_id,),
        ).fetchone()[0]
            (session.session_id,),
        ).fetchone()[0]

    session.close()


async def main():


asyncio.run(main())
```

Actual result:

```text
```

### Expected behavior

Aphroq:
### Summary

`AdvancedSQLiteSession.add_items()` now writes base message rows and `message_structure` rows in a single transaction. If structure metadata cannot be written, the message insert rolls back and the original error is surfaced to the caller instead of returning after creating invisible or orphaned session rows.

Added regression coverage for:
- rolling back base messages when structure metadata fails
- retrying the same batch after a metadata failure without duplicating rows

----------------------------------
Updated the branch with additional hardening after reviewing the compatibility boundary.

- `add_items()` still treats message rows and `message_structure` rows as one atomic save, but failure logging now preserves the original traceback before re-raising.
- The legacy `_add_structure_metadata()` repair path now also re-raises after cleanup, so structure metadata is no longer described or handled as merely supplementary.
- Added regression coverage for two additional transaction boundaries:
  - a failed second batch preserves previously committed history and structure rows
  - a partial `message_structure` write rolls back together with the newly inserted message rows

Compatibility note: this intentionally changes the failure path from logging and returning to fail-fast behavior. The old behavior could report success even though `AdvancedSQLiteSession.get_items()` could not read the saved messages, which hides data consistency problems and makes caller retry logic unreliable.

### Test plan

- `uv run pytest tests/extensions/memory/test_advanced_sqlite_session.py -q`

### Issue number

Closes #3348

### Checks

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -133,26 +133,15 @@ class AdvancedSQLiteSession(SQLiteSession):
         def _add_items_sync():
             """Synchronous helper to add items and structure metadata together."""
             with self._locked_connection() as conn:
-                # Keep both writes in one critical section so message IDs and metadata stay aligned.
-                self._insert_items(conn, items)
-                conn.commit()
                 try:
+                    # Keep both writes in one transaction so metadata failures do not leave orphans.
+                    self._insert_items(conn, items)
                     self._insert_structure_metadata(conn, items)
                     conn.commit()
-                except Exception as e:
+                except Exception:
                     conn.rollback()
-                    self._logger.error(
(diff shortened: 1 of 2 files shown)


Aphroq:
@codex review

Aphroq:
@codex review

PR #3287, opened 2026-05-08
fix: #3286 send realtime output for unknown tool calls

Aphroq:
### Summary

Send a model-visible error tool output when realtime receives a tool call whose name is not available on the current agent.

Known function tools, rejected tools, and handoffs already complete model-visible tool calls with `RealtimeModelSendToolOutput`. This change makes the unknown-tool branch do the same with `start_response=True`, while preserving the existing local `RealtimeError` event for application observers.

### Test plan

- `uv run pytest tests/realtime/test_session.py -k "unknown_tool_handling"`
- `bash .agents/skills/code-change-verification/scripts/run.sh`

### Issue number

Closes #3286

### Checks

- [x] I've added new tests (if relevant)
- [ ] I've added/updated the relevant documentation
- [x] I've run `make lint` and `make format`
- [x] I've made sure tests pass

PR #3366, opened 2026-05-11
fix: avoid auto response for unknown realtime tools (ref: #3287)

seratch:
This pull request fixes unknown Realtime tool handling so the SDK completes the model-visible tool call with an error output without automatically triggering a follow-up model response.

Previously, the unknown-tool branch sent `RealtimeModelSendToolOutput(start_response=True)`, which caused the Realtime model backend to receive a `response.create` immediately after the error output. That made the assistant respond even when the application may want to decide whether to continue, recover tool configuration, or stop the turn. Known tools, approval rejections, and handoffs still trigger follow-up responses as before.

context: https://github.com/openai/openai-agents-python/pull/3287#issuecomment-4425289301

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -776,7 +776,7 @@ class RealtimeSession(RealtimeModelListener):
                 RealtimeModelSendToolOutput(
                     tool_call=event,
                     output=error_message,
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

PR #3369, opened 2026-05-11
ci: use typecheck-only dependencies in CI

seratch:
This pull request updates the CI typecheck job to install a narrower dependency set before running mypy and pyright.

It adds a dedicated `typecheck` dependency group, a `make sync-typecheck` target for preparing that environment, and a `make typecheck-no-sync` target so CI uses the already-synced environment instead of pulling the full dev group back in through `uv run`. The regular `make sync` and `make typecheck` paths remain available for local development and the broader verification workflow.

In local cache-backed validation, the typecheck CI environment shrank from 194 installed packages with the full `make sync` path to 162 installed packages with `make sync-typecheck`, removing 32 packages from the typecheck job setup. The new `make sync-typecheck` + `make typecheck-no-sync` path completed typechecking in 22 seconds locally after sync, while avoiding the extra automatic dependency sync that `uv run` would otherwise trigger. Exact wall-clock savings should be confirmed by the next GitHub Actions run because cold network/cache behavior dominates CI timing.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,12 +55,12 @@ jobs:
         with:
           version: "0.11.7"
           enable-cache: true
-      - name: Install dependencies
+      - name: Install typecheck dependencies
         if: steps.changes.outputs.run == 'true'
-        run: make sync
+        run: make sync-typecheck
       - name: Run typecheck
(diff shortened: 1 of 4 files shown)


seratch:
This is not effective fro speeding up.

PR #3378, opened 2026-05-12
fix: #3357 AgentOutputSchema.name() crash on Literal output types

### LINKED ISSUE #3357: AgentOutputSchema.name() fails for Literal output types
### Please read this first

- **Have you searched for related issues?** Yes. I searched existing issues and PRs for `AgentOutputSchema Literal name` and did not find a matching report or fix.

### Describe the bug

`AgentOutputSchema.name()` crashes for output types that include `typing.Literal` values. The schema itself can be built, but the name formatter recursively assumes every typing argument has `__name__`. Literal arguments are plain values such as strings, so the run can fail while starting the agent trace for an otherwise valid structured output type.

I also checked the adjacent structured output paths. Pydantic models, plain `dict`, generic `dict`, and generic `list` output types keep working; the issue is with typing constructs whose arguments are not Python types, such as `Literal["ok"]`.

### Debug information

### Repro steps

Run this minimal script:

```python
from typing import Any, Literal, cast

from agents import AgentOutputSchema

schema = AgentOutputSchema(cast(type[Any], Literal["ok"]))
print(schema.json_schema())
print(schema.name())
```

Actual result:

```text
{'properties': {'response': {'const': 'ok', 'title': 'Response', 'type': 'string'}}, 'required': ['response'], 'title': 'OutputType', 'type': 'object', 'additionalProperties': False}
  File "<stdin>", line 7, in <module>
  File "/tmp/openai-agents-3347-edit/src/agents/agent_output.py", line 168, in name
    return _type_to_str(self.output_type)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/openai-agents-3347-edit/src/agents/agent_output.py", line 191, in _type_to_str
    args_str = ", ".join(_type_to_str(arg) for arg in args)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/openai-agents-3347-edit/src/agents/agent_output.py", line 191, in <genexpr>
    args_str = ", ".join(_type_to_str(arg) for arg in args)
                         ^^^^^^^^^^^^^^^^^
  File "/tmp/openai-agents-3347-edit/src/agents/agent_output.py", line 189, in _type_to_str
    return t.__name__
           ^^^^^^^^^^
AttributeError: 'str' object has no attribute '__name__'. Did you mean: '__ne__'?
```

### Expected behavior

`AgentOutputSchema.name()` should return a stable string for valid output types that include literal values, for example `Literal['ok']`, instead of crashing. Nested forms such as `list[Literal["ok", "done"]]` should also format without raising.

john-rocky:
### Summary

`AgentOutputSchema.name()` previously crashed for any output type that contained `typing.Literal`, because `_type_to_str` recursed into each typing arg and unconditionally read `__name__`. The Literal value (e.g. the `"ok"` inside `Literal["ok"]`) is a `str` instance rather than a class, so the lookup raised `AttributeError` and broke the trace path when the run started. `schema.json_schema()` already worked, so the bug only surfaced at the very last step of agent setup.

The fix is local to `_type_to_str`:

- If the recursive arg isn't a class, fall back to `repr(t)` so literal values keep their original form (`"'ok'"`, `'1'`, etc.).
- Resolve origin names through `getattr(origin, "__name__", None) or getattr(origin, "_name", None) or str(origin)` so typing constructs that only expose `_name` (`typing.Literal`, `typing.Union`) still produce a readable label.

`_type_to_str` is now typed as `t: Any` to match the recursive call, which already passed non-type values into itself.

### Test plan

Added `test_structured_output_literal_name_does_not_crash` covering:

- `Literal["ok"]` → `"Literal['ok']"` (the exact repro from the issue)
- `Literal["ok", "done"]` → `"Literal['ok', 'done']"`
- `list[Literal["ok", "done"]]` → `"list[Literal['ok', 'done']]"`
- `Literal[1, 2]` → `"Literal[1, 2]"`
- Plain types and existing generics (`str`, `list[int]`) unchanged

```
$ make format
All checks passed!
$ make lint
All checks passed!
$ uv run mypy src/agents/agent_output.py tests/test_output_tool.py
$ uv run pytest tests/test_output_tool.py -v
...
tests/test_output_tool.py::test_structured_output_literal_name_does_not_crash PASSED
============================== 12 passed in 0.11s ==============================
```

### Issue number

Closes #3357.

### Checks

### DIFF
--- a/src/agents/agent_output.py
+++ b/src/agents/agent_output.py
@@ -180,15 +180,25 @@ def _is_subclass_of_base_model_or_dict(t: Any) -> bool:
     return issubclass(t, BaseModel | dict)
 
 
-def _type_to_str(t: type[Any]) -> str:
+def _type_to_str(t: Any) -> str:
     origin = get_origin(t)
     args = get_args(t)
 
     if origin is None:
-        # It's a simple type like `str`, `int`, etc.
-        return t.__name__
+        # Plain type (str, int, MyModel, ...) or a non-type value supplied as a
+        # type argument — e.g. the "ok" inside `Literal["ok"]` is a str instance,
+        # not a class, so `t.__name__` would raise. Fall back to repr() in that
+        # case so nested forms like `list[Literal["ok"]]` still format cleanly.
+        if isinstance(t, type):
+            return t.__name__
+        return repr(t)
     elif args:
         args_str = ", ".join(_type_to_str(arg) for arg in args)
-        return f"{origin.__name__}[{args_str}]"
+        # `typing.Literal`/`typing.Union`/etc. expose `_name` rather than
+        # `__name__` on some Python versions.
+        origin_name = (
+            getattr(origin, "__name__", None) or getattr(origin, "_name", None) or str(origin)
+        )
+        return f"{origin_name}[{args_str}]"
     else:
         return str(t)
--- a/tests/test_output_tool.py
+++ b/tests/test_output_tool.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any
+from typing import Any, Literal, cast
 
 import pytest
 from pydantic import BaseModel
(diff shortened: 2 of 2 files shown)


seratch:
resolved by https://github.com/openai/openai-agents-python/pull/3358

Issue #3346, opened 2026-05-11
AdvancedSQLiteSession.delete_branch() leaves branch-only messages in the base table


Aphroq:
### Please read this first

- I checked the related implementation and existing tests.
- I searched for related issues/PRs and did not find a direct duplicate: `AdvancedSQLiteSession delete_branch orphan messages`, `delete_branch message_structure agent_messages`.

### Describe the bug

`AdvancedSQLiteSession.delete_branch()` removes `turn_usage` and `message_structure` rows for the deleted branch, but it leaves rows in the underlying `agent_messages` table when those messages were only referenced by that branch. After deletion, those messages are invisible to advanced branch queries such as `get_items()`, but they remain in the database.

Impact: branch-heavy sessions can accumulate invisible orphaned rows, causing database growth and audit noise.

### Debug information

- Agents SDK version: `main @ 4c3de2df` (latest release boundary: `v0.17.0`)
- Python version: Python 3.12.1

### Repro steps

Run this minimal script:

```python
import asyncio
import json

from agents.extensions.memory import AdvancedSQLiteSession


async def main():
    session = AdvancedSQLiteSession(
        session_id="delete_branch_orphan_repro",
        create_tables=True,
    )

    await session.add_items(
        [
            {"role": "user", "content": "main question"},
            {"role": "assistant", "content": "main answer"},
        ]
    )

    await session.create_branch_from_turn(1, "branch_only")
    await session.add_items(
        [
            {"role": "user", "content": "branch-only question"},
            {"role": "assistant", "content": "branch-only answer"},
        ]
    )

    await session.delete_branch("branch_only", force=True)

    with session._locked_connection() as conn:
        message_rows = conn.execute(
            """
            SELECT id, message_data
            FROM agent_messages
            WHERE session_id = ?
            ORDER BY id
            """,
            (session.session_id,),
        ).fetchall()
        structure_rows = conn.execute(
            """
            SELECT branch_id, message_id
            FROM message_structure
            WHERE session_id = ?
            ORDER BY message_id
            """,
            (session.session_id,),
        ).fetchall()

    print(
        "agent_messages_after_delete:",
        [(row[0], json.loads(row[1]).get("content")) for row in message_rows],
    )
    print("message_structure_after_delete:", structure_rows)
    session.close()


asyncio.run(main())
```

Actual result:

```text
[...]

PR #3379, opened 2026-05-12
fix: #3348 AdvancedSQLiteSession.add_items silent partial save

### LINKED ISSUE #3348: AdvancedSQLiteSession.add_items can report success after structure metadata failure

john-rocky:
### Summary

`AdvancedSQLiteSession.add_items` first inserted the base message rows, **committed** them, and only then wrote `message_structure` rows. If the structure write failed, the matching `conn.rollback()` was a no-op because the base rows were already committed. A best-effort orphan cleanup ran next, but the original exception was always swallowed by `self._logger.error(...)`, so the caller saw a successful `add_items()` even though:

- `get_items()` (which joins through `message_structure`) returned the rows as missing, and
- when the cleanup itself failed, the base table kept invisible orphan rows that no advanced read could ever surface.

The fix is to share a single transaction for both writes and let the original exception propagate. `_insert_structure_metadata` already reads back the just-inserted message IDs through a `SELECT`, which SQLite serves from the uncommitted state of the same transaction, so no intermediate commit is needed for correctness — only the false success was load-bearing.

### Test plan

Added `test_add_items_rolls_back_when_structure_metadata_fails`. It subclasses `AdvancedSQLiteSession` with a toggleable `_insert_structure_metadata` that fails the first time and succeeds on retry, and verifies that:

- the failing call raises `RuntimeError`,
- both `get_items()` and direct counts on `<messages_table>` and `message_structure` come back empty,
- a follow-up call with the failure cleared persists the message normally.

```
$ uv run pytest tests/extensions/memory/test_advanced_sqlite_session.py
============================== 37 passed in 0.19s ==============================
$ make format
All checks passed!
$ make lint
All checks passed!
$ uv run mypy src/agents/extensions/memory/advanced_sqlite_session.py tests/extensions/memory/test_advanced_sqlite_session.py
```

### Issue number

Closes #3348. Related to #3346 (orphan rows after `delete_branch`), which is a separate path and not touched here.

### Checks

### DIFF
--- a/src/agents/agent_output.py
+++ b/src/agents/agent_output.py
@@ -180,15 +180,25 @@ def _is_subclass_of_base_model_or_dict(t: Any) -> bool:
     return issubclass(t, BaseModel | dict)
 
 
-def _type_to_str(t: type[Any]) -> str:
+def _type_to_str(t: Any) -> str:
     origin = get_origin(t)
     args = get_args(t)
 
     if origin is None:
-        # It's a simple type like `str`, `int`, etc.
-        return t.__name__
+        # Plain type (str, int, MyModel, ...) or a non-type value supplied as a
+        # type argument — e.g. the "ok" inside `Literal["ok"]` is a str instance,
+        # not a class, so `t.__name__` would raise. Fall back to repr() in that
+        # case so nested forms like `list[Literal["ok"]]` still format cleanly.
+        if isinstance(t, type):
+            return t.__name__
+        return repr(t)
     elif args:
         args_str = ", ".join(_type_to_str(arg) for arg in args)
-        return f"{origin.__name__}[{args_str}]"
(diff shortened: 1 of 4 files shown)


john-rocky:
Closing — apologies, this duplicates #3349 by @Aphroq, who filed the original issue (#3348) two minutes earlier on 2026-05-11. I also bundled an unrelated `_type_to_str` fix for `Literal` types that has since been merged via @Aphroq's #3358 on 2026-05-14, so there's nothing left in this PR that isn't already covered. Sorry for the noise.

PR #3347, opened 2026-05-11
fix: #3346 clean up branch-only messages when deleting branches

Aphroq:
### Summary

This pull request fixes `AdvancedSQLiteSession.delete_branch()` leaving branch-only message rows behind after deleting a branch. After removing the deleted branch's `message_structure` rows, the method now cleans up message rows for the current session that are no longer referenced by any structure row, using the same lock and transaction while preserving messages still shared by main or other branches.

### Test plan

- [x] `uv run pytest -q tests/extensions/memory/test_advanced_sqlite_session.py::test_delete_branch_removes_branch_only_messages tests/extensions/memory/test_advanced_sqlite_session.py::test_branching_functionality tests/extensions/memory/test_advanced_sqlite_session.py::test_branch_deletion_with_force`
- [x] `uv run pytest -q tests/extensions/memory/test_advanced_sqlite_session.py`
- [x] `bash .agents/skills/code-change-verification/scripts/run.sh`

### Issue number

Closes #3346

### Checks

- [x] I've added new tests (if relevant)
- [ ] I've added/updated the relevant documentation
- [x] I've run `make lint` and `make format`
- [x] I've made sure tests pass

PR #3380, opened 2026-05-12
fix: #3346 AdvancedSQLiteSession.delete_branch orphaned base rows

### LINKED ISSUE #3346: AdvancedSQLiteSession.delete_branch() leaves branch-only messages in the base table

john-rocky:
### Summary

`AdvancedSQLiteSession.delete_branch()` deleted `turn_usage` and `message_structure` rows for the branch, but never touched the underlying messages table. Any base row whose only `message_structure` reference was on the deleted branch — the common case for follow-ups added after `create_branch_from_turn` — stayed behind: invisible to advanced reads (which join through `message_structure`) but still padding the on-disk size, which made branch-heavy workflows accumulate ghost rows.

After the structure rows are deleted, drop any base rows that no longer have a matching `message_structure` entry, inside the same transaction. The existing `_cleanup_orphaned_messages_sync` helper does exactly that, so the change is minimal and matches the cleanup pattern already used elsewhere in this module. Messages still referenced by another branch (e.g. shared turns from before the split) are left alone. The log line now also reports the orphan count so operators can spot unexpected cleanup volumes.

### Test plan

Added `test_delete_branch_removes_branch_only_base_messages`. It reproduces the issue script: create a main turn, branch from it, append branch-only messages, then `delete_branch(..., force=True)`. After the call, the test asserts that:

- the main-branch base rows (ids 1 and 2) remain,
- the branch-only base rows (ids 3 and 4) are gone,
- only `main` structure rows remain,
- `get_items(branch_id="main")` returns the original main payload.

```
$ uv run pytest tests/extensions/memory/test_advanced_sqlite_session.py
============================== 37 passed in 0.20s ==============================
$ make format
All checks passed!
$ make lint
All checks passed!
$ uv run mypy src/agents/extensions/memory/advanced_sqlite_session.py tests/extensions/memory/test_advanced_sqlite_session.py
```

### Issue number

Closes #3346. Adjacent #3348 (silent partial save in `add_items`) is fixed by a separate PR and not touched here.

### Checks

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -487,30 +487,26 @@ class AdvancedSQLiteSession(SQLiteSession):
 
     def _cleanup_orphaned_messages_sync(self, conn: sqlite3.Connection) -> int:
         with closing(conn.cursor()) as cursor:
-            # Find messages without structure metadata.
+            # Delete branch-orphaned messages in a single set-based statement so
+            # we never bind one parameter per orphan id. The previous IN (?, ?, ...)
+            # form raised "too many SQL variables" once the cleanup batch grew past
+            # SQLITE_MAX_VARIABLE_NUMBER (999 on older SQLite builds, 32766 on newer).
             cursor.execute(
                 f"""
-                SELECT am.id
-                FROM {self.messages_table} am
-                LEFT JOIN message_structure ms ON am.id = ms.message_id
-                WHERE am.session_id = ? AND ms.message_id IS NULL
+                DELETE FROM {self.messages_table}
+                WHERE id IN (
+                    SELECT am.id
+                    FROM {self.messages_table} am
+                    LEFT JOIN message_structure ms ON am.id = ms.message_id
+                    WHERE am.session_id = ? AND ms.message_id IS NULL
+                )
             """,
                 (self.session_id,),
             )
 
-            orphaned_ids = [row[0] for row in cursor.fetchall()]
-
-            if not orphaned_ids:
-                return 0
-
-            placeholders = ",".join("?" * len(orphaned_ids))
-            cursor.execute(
-                f"DELETE FROM {self.messages_table} WHERE id IN ({placeholders})",
-                orphaned_ids,
-            )
-
             deleted_count = cursor.rowcount
-            self._logger.info(f"Cleaned up {deleted_count} orphaned messages")
+            if deleted_count > 0:
+                self._logger.info(f"Cleaned up {deleted_count} orphaned messages")
             return deleted_count
 
     def _classify_message_type(self, item: TResponseInputItem) -> str:
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -786,14 +786,21 @@ class AdvancedSQLiteSession(SQLiteSession):
 
                     structure_deleted = cursor.rowcount
 
+                    # Drop any base messages that this branch was the only
+                    # reference for. Without this step, branch-only rows become
+                    # invisible to advanced reads (which join through
+                    # `message_structure`) but linger in the base table.
+                    orphans_deleted = self._cleanup_orphaned_messages_sync(conn)
+
                     conn.commit()
 
-                    return usage_deleted, structure_deleted
+                    return usage_deleted, structure_deleted, orphans_deleted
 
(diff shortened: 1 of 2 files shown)


john-rocky:
Thanks for catching this — addressed in cb831a68.

`_cleanup_orphaned_messages_sync` now issues a single set-based `DELETE FROM {messages_table} WHERE id IN (SELECT … LEFT JOIN message_structure …)`. It binds only `session_id`, so the orphan count is irrelevant — no more `too many SQL variables` when a branch has more orphans than `SQLITE_MAX_VARIABLE_NUMBER`.

Regression test added (`test_cleanup_orphaned_messages_exceeds_sqlite_variable_limit`): drops the per-connection variable limit to 5 via `Connection.setlimit`, inserts 20 orphaned rows, and asserts cleanup deletes all of them — the old `IN (?, ?, …, ?)` form would have raised on that connection. Test is `skipif`'d for Python < 3.11 since `setlimit` lives there.

john-rocky:
Closing — apologies, this duplicates #3347 by @Aphroq, who filed the original issue (#3346) and submitted a fix one minute later on 2026-05-11. I missed their PR when I opened mine on 2026-05-12, and I shouldn't have pushed `cb831a68` today either without checking first.

The codex-bot variable-limit concern that I addressed in `cb831a68` may still be worth surfacing on #3347 — but @Aphroq's approach uses `NOT IN (SELECT message_id FROM message_structure WHERE session_id = ?)`, which is also set-based and binds only `session_id`, so it isn't subject to the same `SQLITE_MAX_VARIABLE_NUMBER` issue. Functionally equivalent fixes; theirs should land. Sorry for the noise, @Aphroq.

PR #3388, opened 2026-05-13
fix: preserve agents-extras hint when memory backends are missing

ioleksiuk:
### Summary

\`src/agents/extensions/memory/__init__.py\` wraps each lazy import so the user gets a helpful \`pip install openai-agents[<extra>]\` message when an extra is missing. The outer wrappers caught \`ModuleNotFoundError\`, but \`redis_session.py\`, \`dapr_session.py\`, and \`mongodb_session.py\` each have their own \`try/except ImportError\` that re-raises a plain \`ImportError\` (not \`ModuleNotFoundError\`).

Python's class hierarchy: \`ModuleNotFoundError\` is a subclass of \`ImportError\`. So \`except ModuleNotFoundError\` does NOT catch a bare \`ImportError\` re-raised by the inner wrapper. The intended outer hint was therefore unreachable for the three backends with their own \`try/except\`, and users saw the less helpful inner message.

Concretely (reproduced against current main):

\`\`\`
>>> from agents.extensions.memory import RedisSession   # redis missing
ImportError: RedisSession requires the 'redis' package. Install it with: pip install redis
\`\`\`

Expected:

\`\`\`
ImportError: RedisSession requires the 'redis' extra. Install it with: pip install openai-agents[redis]
\`\`\`

Widen the outer \`except\` clauses from \`ModuleNotFoundError\` to \`ImportError\`. Since \`ModuleNotFoundError\` is a subclass of \`ImportError\`, this only adds coverage; it doesn't change behavior for the backends that already raised \`ModuleNotFoundError\` directly.

### Test plan

- New \`tests/extensions/memory/test_memory_imports.py\` parametrized over every extras-backed export (\`RedisSession\`, \`DaprSession\`, \`DAPR_CONSISTENCY_EVENTUAL\`, \`DAPR_CONSISTENCY_STRONG\`, \`MongoDBSession\`, \`EncryptedSession\`, \`SQLAlchemySession\`) and both \`ImportError\`/\`ModuleNotFoundError\` subclasses — 14 cases, all passing.
- \`uv run pytest tests/extensions/memory/test_memory_imports.py tests/mcp/test_mcp_imports.py\` — 16 passed.

### Issue number

N/A — found while auditing the lazy-import pattern from #3048 for analogous gaps.

### Checks

### DIFF
--- a/src/agents/extensions/memory/__init__.py
+++ b/src/agents/extensions/memory/__init__.py
@@ -42,7 +42,7 @@ def __getattr__(name: str) -> Any:
             from .encrypt_session import EncryptedSession  # noqa: F401
 
             return EncryptedSession
-        except ModuleNotFoundError as e:
+        except ImportError as e:
             raise ImportError(
                 "EncryptedSession requires the 'cryptography' extra. "
                 "Install it with: pip install openai-agents[encrypt]"
@@ -53,7 +53,7 @@ def __getattr__(name: str) -> Any:
             from .redis_session import RedisSession  # noqa: F401
 
             return RedisSession
-        except ModuleNotFoundError as e:
+        except ImportError as e:
             raise ImportError(
                 "RedisSession requires the 'redis' extra. "
                 "Install it with: pip install openai-agents[redis]"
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing this feedback. We'll improve it this way: https://github.com/openai/openai-agents-python/pull/3389

PR #3390, opened 2026-05-13
fix(mcp): re-nest flattened arguments for Realtime tool calls

adityasingh2400:
When a RealtimeAgent calls an MCP tool whose input schema declares a nested-object property, the model frequently emits that property's fields at the top level rather than as a nested object, and the MCP server then rejects the call or silently misbehaves (#1681).

As noted in the earlier triage on the issue, this is because the OpenAI Realtime API's `RealtimeFunctionTool` type has no `strict` field — "the OpenAI Realtime API doesn't support the `strict` parameter for function tools" ([comment](https://github.com/openai/openai-agents-python/issues/1681#issuecomment-4018047087)). The schema travels intact but the model is free to ignore the nesting. That part has to be fixed upstream, but the SDK can still recover the common case before forwarding to the MCP server.

`MCPUtil.invoke_mcp_tool` now applies a narrow re-nesting pass after JSON parsing. When the parsed arguments contain keys not declared at the top level of `tool.inputSchema` and the schema has exactly one missing object-typed property that can unambiguously absorb them, the extras are moved into that property. Open-ended objects (no declared `properties`) are always eligible; typed objects are eligible only when every extra is a declared inner key; objects with `additionalProperties: false` and no inner properties are excluded. Ambiguous schemas (two or more candidate absorbers, or extras that don't match a closed inner schema) are left untouched so structured tools are not corrupted. On the `SetProperties` example from the issue, `{\"target\": \"MyTarget\", \"visible\": false}` is forwarded as `{\"target\": \"MyTarget\", \"keysAndValues\": {\"visible\": false}}` and the call succeeds. Five tests in `tests/mcp/test_mcp_util.py` cover the issue's case plus already-nested input, ambiguous siblings, closed inner schemas, and typed inner properties.

Refs #1681

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -569,6 +569,61 @@ class MCPUtil:
             raise TypeError("MCP meta resolver must return a dict or None.")
         return result
 
+    @staticmethod
+    def _renest_flattened_arguments(
+        input_schema: Any,
+        arguments: dict[str, Any],
+    ) -> dict[str, Any] | None:
+        """Re-nest arguments that the model flattened past a nested object property.
+
+        The Realtime API does not support strict JSON schema enforcement on function
+        tools (the `RealtimeFunctionTool` type has no `strict` field), so models will
+        occasionally ignore a nested-object property in an MCP tool's input schema
+        and emit its inner fields at the top level. When the schema has exactly one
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing this idea, but this could affect any MCP server use cases and I don't think it is a right solution for the realtime agents as well.

PR #3394, opened 2026-05-13
fix: guard None text in ItemHelpers.extract_last_content

ioleksiuk:
### Summary

Extends the fix in #3375 (\`fix: guard None text in text_message_output ...\`) to the two sibling helpers in \`ItemHelpers\` that have the same shape.

\`ResponseOutputText.text\` is typed as \`str\` per the Responses API schema, but \`src/agents/items.py:714-720\` already documents (in the comment above the original #3375 fix) that provider gateways like LiteLLM and \`model_construct\` paths during streaming surface \`None\` values. PR #3375 fixed \`text_message_output\` for this case. Two sibling helpers were missed:

- **\`extract_last_content\`** (declared \`-> str\`, items.py:678) — silently returned \`None\` when \`last_content.text\` was \`None\`, violating its declared type contract.
- **\`extract_last_text\`** (declared \`-> str | None\`, items.py:693) — returned \`None\` only by passthrough, so callers using truthy semantics couldn't distinguish "no text content" from "text is None".

Repro for \`extract_last_content\` (declared \`-> str\`, returns \`None\`):

\`\`\`python
>>> text_part = ResponseOutputText.model_construct(text=None, type="output_text", annotations=[])
>>> msg = ResponseOutputMessage(id="m", role="assistant", status="completed",
...                             type="message", content=[text_part])
>>> ItemHelpers.extract_last_content(msg)
None        # but the function is typed `-> str`
\`\`\`

Fix mirrors the #3375 pattern: \`return last_content.text or ""\` in \`extract_last_content\`, \`return last_content.text or None\` in \`extract_last_text\` for explicit semantics.

### Test plan

- New tests in \`tests/utils/test_pretty_print_and_items.py\` covering all four cases (None vs string content) for both helpers, mirroring the existing \`text_message_output\` tests added in #3375.
- \`uv run pytest tests/utils/test_pretty_print_and_items.py\` — 9 passed.
- \`uv run ruff check src/agents/items.py tests/utils/test_pretty_print_and_items.py\` — All checks passed.

### Issue number

N/A — found while auditing for the same pattern that #3375 fixed.

### Checks

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -684,7 +684,12 @@ class ItemHelpers:
             return ""
         last_content = message.content[-1]
         if isinstance(last_content, ResponseOutputText):
-            return last_content.text
+            # ``last_content.text`` is typed as ``str`` per the Responses API schema,
+            # but provider gateways (e.g. LiteLLM) and ``model_construct`` paths during
+            # streaming have been observed surfacing ``None``. Coerce so callers relying
+            # on the ``-> str`` return type don't see a ``None``. Same rationale as
+            # ``extract_text`` below.
+            return last_content.text or ""
         elif isinstance(last_content, ResponseOutputRefusal):
             return last_content.refusal
         else:
--- a/tests/utils/test_pretty_print_and_items.py
+++ b/tests/utils/test_pretty_print_and_items.py
@@ -38,6 +38,30 @@ def test_text_message_outputs_handles_none_text_across_items():
     assert ItemHelpers.text_message_outputs(items) == "world"
 
 
+def _make_output_message(text: str | None) -> ResponseOutputMessage:
+    return ResponseOutputMessage.model_construct(
+        id="msg_1",
+        role="assistant",
+        status="completed",
+        content=[ResponseOutputText.model_construct(type="output_text", text=text, annotations=[])],
+    )
+
+
+def test_extract_last_content_returns_empty_string_for_none_text():
+    """extract_last_content is declared `-> str` and must not return None even if
+    the underlying ResponseOutputText.text is None (observed via LiteLLM gateways
+    and ``model_construct`` paths during streaming, per items.py:714-720)."""
+    msg = _make_output_message(None)
+    result = ItemHelpers.extract_last_content(msg)
+    assert isinstance(result, str)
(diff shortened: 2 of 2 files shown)


src/agents/items.py
```diff
@@ -684,7 +684,7 @@ def extract_last_content(cls, message: TResponseOutputItem) -> str:
             return ""
         last_content = message.content[-1]
         if isinstance(last_content, ResponseOutputText):
-            return last_content.text
+            return last_content.text or ""
```
seratch, line 692: Can you add a quick comment why this line should support both str and None here while the type is str?

seratch (changes requested):

ioleksiuk:
Thanks for the review! Added inline comments on both `extract_last_content` (line 687) and `extract_last_text` (line 701) explaining the rationale — same shape as the existing comment in `extract_text` (items.py:714-720):

> `last_content.text` is typed as `str` per the Responses API schema, but provider gateways (e.g. LiteLLM) and `model_construct` paths during streaming have been observed surfacing `None`.

The two comments differ slightly in the second sentence to match their respective return types (`-> str` coerces to `""`, `-> str | None` coerces to `None` so truthiness checks still work).

Force-pushed the amended commit.

ioleksiuk:
Follow-up: on second look at the diff, the original `extract_last_text` change (`return last_content.text or None`) had a subtle semantic regression I missed — it conflated `text=""` with `text=None`. Both became `None`, which would silently change behavior for tools that legitimately return empty text (callers using `result is None` to distinguish "no content" from "empty content" would now see `None` for both).

Reverted that change and dropped the related tests/inline comment. The PR is now scoped to a single, real fix:

- `extract_last_content` (declared `-> str`) — coerce `None` to `""` so the function never violates its type contract.

`extract_last_text` left untouched because:
- It's declared `-> str | None`, so returning `None` is legal per the signature.
- The pre-existing behavior of returning `""` for empty text is meaningful (callers can still distinguish via `is None`).

Updated commit message + PR title accordingly. Sorry for the noise.

seratch:
@codex review

seratch (approved):

PR #3395, opened 2026-05-13
fix: apply empty-content guard to empty-string chat tool outputs

ioleksiuk:
### Summary

Extends the fix in #3312 (\`fix: avoid empty chat tool outputs\`) to one case the original guard missed.

#3312 added an empty-content guard that swaps the tool message body with \`_OMITTED_TOOL_OUTPUT_PLACEHOLDER\` when the converted content is empty. The guard was placed inside the \`else\` branch of \`isinstance(all_output_content, str)\` at \`chatcmpl_converter.py:749-769\`, so it only triggered when the output was a non-text-only list. When the output is the empty string \`""\`, \`extract_all_content("")\` returns \`""\` (the string branch), \`tool_result_content\` is set to \`""\` without ever reaching the guard, and the Chat Completions message ends up with \`"content": ""\` — exactly the broken shape #3310/#3312 aimed to prevent.

Repro on current \`main\` (before this PR):

\`\`\`python
>>> from agents.models.chatcmpl_converter import Converter
>>> items = [
...     {"type": "function_call", "id": "fc_1", "call_id": "c1",
...      "name": "get_status", "arguments": "{}"},
...     {"type": "function_call_output", "call_id": "c1", "output": ""},
... ]
>>> [m for m in Converter.items_to_messages(items) if m.get("role") == "tool"]
[{'role': 'tool', 'tool_call_id': 'c1', 'content': ''}]
# Chat Completions API rejects: messages.X.content must be non-empty
\`\`\`

Lift the guard above the \`if/else\` so both branches go through it. \`not tool_result_content\` is truthy for both \`""\` and \`[]\`, so the single check covers both empty shapes.

### Test plan

- New \`test_items_to_messages_with_empty_string_function_output_uses_placeholder_by_default\` — verifies the default-mode placeholder + warning for \`output=""\`.
- New \`test_items_to_messages_with_empty_string_function_output_raises_in_strict_mode\` — verifies strict-mode UserError.
- \`uv run pytest tests/models/test_openai_chatcompletions_converter.py -k empty\` — 4 passed (both pre-existing empty-list tests + the two new empty-string tests).

### Issue number

N/A — found while auditing for the same fix pattern that #3312 applied to empty lists.

### Checks

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -754,19 +754,19 @@ class Converter:
                             for c in all_output_content
                             if c.get("type") == "text"
                         ]
-                        if not tool_result_content:
-                            message = (
-                                "Chat Completions tool outputs cannot be empty or contain only "
-                                "non-text content unless preserve_tool_output_all_content=True."
-                            )
-                            if strict_feature_validation:
-                                raise UserError(message)
-                            logger.warning(
-                                "%s Replacing the tool output with a placeholder; enable strict "
-                                "feature validation to raise an error instead.",
-                                message,
-                            )
-                            tool_result_content = _OMITTED_TOOL_OUTPUT_PLACEHOLDER
+                    if not tool_result_content:
+                        message = (
+                            "Chat Completions tool outputs cannot be empty or contain only "
+                            "non-text content unless preserve_tool_output_all_content=True."
+                        )
+                        if strict_feature_validation:
+                            raise UserError(message)
+                        logger.warning(
+                            "%s Replacing the tool output with a placeholder; enable strict "
+                            "feature validation to raise an error instead.",
+                            message,
+                        )
+                        tool_result_content = _OMITTED_TOOL_OUTPUT_PLACEHOLDER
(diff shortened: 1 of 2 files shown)


ioleksiuk:
Closing — on second read of #3310, the issue reporter explicitly listed "String tool outputs remain unchanged." as a deliberate edge case. The placement of #3312's guard in the `else` branch was intentional, not an oversight.

Concrete reasons this PR doesn't make sense after all:

1. `ChatCompletionToolMessageParam.content` is `Required[Union[str, Iterable[...]]]` per `openai-python` — empty string is a valid value of `str`, the API type does not reject it.
2. When a user writes `return ""` from a function tool, the empty string is their intentional output. The SDK should not silently rewrite it to `[tool output omitted]`.
3. The bug #3312 fixed was lossy conversion: a tool returns image/audio content → the converter strips it → message ends up empty. Empty-string output isn't lossy — it's faithful to the tool's return value.

My PR conflated "empty content after filtering non-representable parts" with "user explicitly returned empty string" — those are semantically different.

Apologies for the noise.

PR #3401, opened 2026-05-14
feat(mcp): add tool_name_prefix to disambiguate duplicate MCP tool names

### LINKED ISSUE #464: Duplicate tool names across MCP servers cause errors

adityasingh2400:
When two MCP servers expose tools with the same name (e.g. both a GitHub and a Linear server expose `create_issue`), wiring both into an agent fails with `Duplicate tool names found across MCP servers`. The existing `include_server_in_tool_names` agent-level flag auto-generates a server-derived prefix, but users have no way to pick a short, friendly prefix per server – the original ask in the issue.

This PR adds an optional keyword-only `tool_name_prefix` argument to the `MCPServer` base class and the `MCPServerStdio`, `MCPServerSse`, and `MCPServerStreamableHttp` constructors. When set, `list_tools()` returns tools named `f"{prefix}_{original_name}"` and `call_tool()` strips the prefix before dispatching to the upstream MCP server, so the server still receives its original tool name. The internal cache and `tool_filter` continue to operate on the upstream names, and `require_approval` policy lookups continue to match the original tool name even with a prefix configured. A prefixed name that would exceed the 64-character MCP tool-name limit raises `UserError` early at `list_tools()` time rather than failing later.

Tests cover the two-servers-with-different-prefixes case, prefix stripping on `call_tool()` (both via the `FakeMCPServer` and an end-to-end check against the real `MCPServerStdio` with the MCP client patched), composition with `tool_filter`, the 64-char overflow error, and cache invariants. Docs in `docs/mcp.md` get a short new section showing how to use it. Closes #464.

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -54,6 +54,31 @@ Notes:
 - Server-level `failure_error_function` overrides `Agent.mcp_config["failure_error_function"]` for that server.
 - `include_server_in_tool_names` is opt-in. When enabled, each local MCP tool is exposed to the model with a deterministic server-prefixed name, which helps avoid collisions when multiple MCP servers publish tools with the same name. Generated names are ASCII-safe, stay within the function-tool name length limit, and avoid existing local function tool and enabled handoff names on the same agent. The SDK still invokes the original MCP tool name on the original server.
 
+## Disambiguating duplicate MCP tool names
+
+When the same tool name is published by more than one MCP server (for example, both a GitHub and a Linear server expose `create_issue`), the agent run fails with `Duplicate tool names found across MCP servers`. If you do not want to opt in to the auto-generated `include_server_in_tool_names` scheme, set a custom `tool_name_prefix` on each individual server. The SDK exposes tools to the model as `f"{prefix}_{original_name}"` and strips the prefix before dispatching `call_tool()` to the upstream server, so the underlying MCP server only ever sees its original tool names.
+
+```python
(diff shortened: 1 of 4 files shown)


adityasingh2400:
Note on the duplicate label. PRs #2677 and #2788 both went stale before landing rather than being rejected on substance. The closest prior attempt, #2788, hit a concrete blocker that seratch flagged in his review, the auto-derived prefix from a non-ASCII server label produced invalid OpenAI tool names.

This PR takes a different shape on that exact point. The prefix is user-provided as a keyword argument on the MCPServer constructor, not derived from the server label, so the non-ASCII generation path does not exist here. The user picks the short readable prefix per server ("github", "linear") and the SDK leaves it alone aside from length validation against the 64-char tool-name ceiling. The internal cache, tool_filter, and require_approval lookups continue to operate on the upstream tool name, and call_tool strips the prefix before dispatch so the server still receives its original name.

Happy to rebase or split the docs section out if that helps land the core API change.

seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/3019

PR #3411, opened 2026-05-14
fix: log exception when output guardrail raises instead of silently ignoring

cty-ut:
### Summary

In `RealtimeSession._run_output_guardrails()`, exceptions raised by individual
guardrails were silently swallowed with a bare `except Exception: continue`.
If a guardrail has a bug or misconfiguration, it fails invisibly with no log
output and no way to diagnose the issue.

Fix: add `logger.exception(...)` before `continue` so the error is surfaced
in logs while still allowing the remaining guardrails to run.

### Test plan

Existing realtime tests continue to pass (`make tests`). The change is a
one-line addition of a log call in an existing exception handler.

### Checks

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -987,8 +987,14 @@ class RealtimeSession(RealtimeModelListener):
                 )
                 if result.output.tripwire_triggered:
                     triggered_results.append(result)
-            except Exception:
-                logger.exception("Output guardrail %r raised an exception", guardrail.get_name())
+            except Exception as exc:
+                logger.warning(
+                    "Output guardrail %r raised %s: %s; skipping it.",
+                    guardrail.get_name(),
+                    type(exc).__name__,
+                    exc,
+                )
+                logger.debug("Output guardrail failure details.", exc_info=True)
                 continue
 
         if triggered_results:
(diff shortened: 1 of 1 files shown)


src/agents/realtime/session.py
```diff
@@ -988,7 +988,7 @@ async def _run_output_guardrails(self, text: str, response_id: str) -> bool:
                 if result.output.tripwire_triggered:
                     triggered_results.append(result)
             except Exception:
-                # Continue with other guardrails if one fails
+                logger.exception("Output guardrail %r raised an exception", guardrail.get_name())
```
seratch: Considering of the log level, we'd like to avoid logger.exception for this situation. Can you do something like this?

```python
except Exception as exc:
    logger.warning(
        "Output guardrail %r raised %s: %s; skipping it.",
        guardrail.get_name(),
        type(exc).__name__,
        exc,
    )
    logger.debug("Output guardrail failure details.", exc_info=True)
    continue
```

seratch (changes requested):

src/agents/realtime/session.py
cty-ut: Thanks for the review! Switched to `logger.warning` for the summary line and `logger.debug` with `exc_info=True` for the full traceback.

seratch (approved):

PR #3386, opened 2026-05-13
fix: filter hosted_tool_call types in remove_all_tools handoff filter

ioleksiuk:
### Summary

Extends the fix in #3095 to cover one more raw input item type that the handoff filter was missing.

\`hosted_tool_call\` is a real tool-call item type recognized elsewhere in the SDK:

- \`src/agents/run_state.py:2024\` — raw item normalization treats it as a tool-call type alongside \`shell_call\`/\`apply_patch_call\`/\`local_shell_call\`.
- \`src/agents/run_internal/tool_execution.py:2368\` — \`_is_hosted_mcp_approval_request\` detects payloads serialized as \`{"type": "hosted_tool_call", "provider_data": {"type": "mcp_approval_request"}}\`.

However, \`_remove_tool_types_from_input\` in \`src/agents/extensions/handoff_filters.py\` did not include \`hosted_tool_call\` in its filter list. Raw \`hosted_tool_call\` items in \`input_history\` therefore survived \`remove_all_tools\` handoff filtering — leaking provider-level tool-call payloads (including hosted MCP approval requests in the raw \`hosted_tool_call\` form) into the downstream agent's history.

Add \`hosted_tool_call\` to the filter list and extend the parametrized list in the existing \`test_removes_hosted_tool_types_from_input_history\` regression test.

### Test plan

- \`uv run pytest tests/test_extension_filters.py -k hosted_tool\` — 1 passed (with the new entry in the list).
- \`uv run ruff check\` on the touched files — All checks passed.

### Issue number

N/A — found while auditing for the same gap that #3095 fixed for \`custom_tool_call\`.

### Checks

- [x] I've added new tests (if relevant)
- [x] I've added/updated the relevant documentation (no doc changes needed)
- [x] I've run \`make lint\` and \`make format\`
- [x] I've made sure tests pass

PR #3412, opened 2026-05-14
fix(handoffs): summarize all tool item types in nest_handoff_history

adityasingh2400:
### Summary

Extends the precedent set by #3095 (custom_tool_call) and #3386 (hosted_tool_call) into the sibling helper that powers `RunConfig.nest_handoff_history`. The `_SUMMARY_ONLY_INPUT_TYPES` set in `src/agents/handoffs/history.py:30-37` only marked `function_call`, `function_call_output`, and `reasoning` as summarized-only. Every other tool kind the SDK can emit (computer, file_search, web_search, mcp_*, code_interpreter, image_generation, local_shell, shell, apply_patch, custom, hosted, tool_search) was included in the JSON conversation summary built by `_format_transcript_item_json` *and* re-forwarded verbatim as a raw `RunItem` in `pre_handoff_items` / `input_items`, duplicating it for the next agent.

Repro (passes pre-fix, fails post-fix without the new set entries):

```python
>>> data = HandoffInputData(
...     input_history=(user_msg,),
...     pre_handoff_items=(custom_tool_call_run_item, custom_tool_output_run_item),
... )
>>> nested = nest_handoff_history(data)
>>> # Summary contains the tool call:
>>> "my_custom" in nested.input_history[0]["content"]
True
>>> # But the raw items also survive, so the next agent sees them twice:
>>> len(nested.pre_handoff_items)
2  # was expected to be 0
```

The fix expands `_SUMMARY_ONLY_INPUT_TYPES` to mirror the comprehensive list maintained by `extensions.handoff_filters._remove_tool_types_from_input`. The two lists exist for adjacent reasons (one for `remove_all_tools`, one for `nest_handoff_history`) but should agree on what counts as a tool item; the comment block documents that intent.

### Test plan

- New tests `test_nest_handoff_history_filters_hosted_tool_pre_items` and `test_nest_handoff_history_filters_hosted_tool_new_items` in `tests/test_extension_filters.py` cover every hosted-tool call and output type for both the `pre_handoff_items` and `new_items`→`input_items` paths.
- `uv run pytest tests/test_extension_filters.py` — 42 passed (was 40, +2 regressions).
- `uv run pytest tests/test_agent_runner.py tests/test_agent_runner_streamed.py -k handoff_history` — 7 passed.
- `make format`, `make lint`, `uv run mypy src/agents/handoffs/history.py tests/test_extension_filters.py` — clean.

### Issue number

N/A — found while auditing for the same gap that #3095 and #3386 fixed in the sibling `remove_all_tools` filter.

### DIFF
--- a/src/agents/handoffs/history.py
+++ b/src/agents/handoffs/history.py
@@ -29,9 +29,33 @@ _conversation_history_end = _DEFAULT_CONVERSATION_HISTORY_END
 
 # Item types that are summarized in the conversation history.
 # They should not be forwarded verbatim to the next agent to avoid duplication.
+# Mirror the tool/side-effect type list maintained by
+# ``extensions.handoff_filters._remove_tool_types_from_input`` so every tool
+# kind the SDK can emit is summarized rather than duplicated.
 _SUMMARY_ONLY_INPUT_TYPES = {
     "function_call",
     "function_call_output",
+    "computer_call",
+    "computer_call_output",
+    "file_search_call",
+    "tool_search_call",
+    "tool_search_output",
+    "web_search_call",
+    "mcp_call",
+    "mcp_list_tools",
+    "mcp_approval_request",
+    "mcp_approval_response",
+    "code_interpreter_call",
+    "image_generation_call",
+    "local_shell_call",
+    "local_shell_call_output",
+    "shell_call",
+    "shell_call_output",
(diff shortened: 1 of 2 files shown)


seratch:
This may be relevant for some use cases, but it could be a breaking change for existing apps. Thus, we won't make this change.

PR #3421, opened 2026-05-15
fix: redact mountpoint credentials in sandbox mount errors

a1vazovsky:
## Summary

- Keep the real `mount-s3` shell command unchanged for execution.
- Add a small command-rendering helper that can render a redacted copy for errors/logging.
- Use the redacted command in `MountCommandError` so AWS access key, secret key, and session token are not exposed when a mount fails.

## Why

`MountpointMountPattern.apply()` currently prefixes `mount-s3` with AWS credential environment variables. On failure it passes that full command into `MountCommandError`, which can leak credentials into exceptions, test logs, or user-visible diagnostics.

## Testing

- `uv run pytest tests/sandbox/test_mount_patterns.py`
- `uv run ruff check src/agents/sandbox tests/sandbox`
- `uv run pyright`

### DIFF
--- a/src/agents/sandbox/entries/mounts/patterns.py
+++ b/src/agents/sandbox/entries/mounts/patterns.py
@@ -108,6 +108,22 @@ async def _write_sensitive_config_file(
     )
 
 
+def _join_mountpoint_command(
+    cmd: list[str],
+    env_vars: list[tuple[str, str]],
+    *,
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing this. https://github.com/openai/openai-agents-python/pull/3429 resolved the issue.

PR #3425, opened 2026-05-15
docs(guardrails): add pre-execution policy example

### LINKED ISSUE #2970: Add pre-execution validation for tool calls

EfeDurmaz16:
## Summary

This adds a concrete tool input guardrail example for pre-execution policy checks, following the current recommended extension point for validating a local function tool call immediately before invocation.

The example shows a payment policy guardrail that checks the exact requested tool payload before `charge_card` executes:

- tool name allowlist
- merchant allowlist
- amount limit
- model-visible rejection via `reject_content`

It also calls out the important boundary that Python-side tool input guardrails apply to local function tools, not hosted tools that execute server-side.

Closes #2970.

## Verification

- `uv run python -m py_compile examples/basic/tool_policy_guardrail.py`
- `uv run ruff check examples/basic/tool_policy_guardrail.py`
- `uv run pyright examples/basic/tool_policy_guardrail.py`
- `uv run pytest tests/test_tool_guardrails.py`
- `git diff --check`

Note: `uv run mkdocs build --strict` currently fails on existing repository-wide i18n/autoref warnings unrelated to this change.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -31,6 +31,7 @@ Check out a variety of sample implementations of the SDK in the examples section
     -   Dynamic system prompts
     -   Basic tool usage (`examples/basic/tools.py`)
     -   Tool input/output guardrails (`examples/basic/tool_guardrails.py`)
+    -   Tool input guardrails for pre-execution policy checks (`examples/basic/tool_policy_guardrail.py`)
     -   Image tool output (`examples/basic/image_tool_output.py`)
     -   Streaming outputs (text, items, function call args)
     -   Responses websocket transport with a shared session helper across turns (`examples/basic/stream_ws.py`)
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for the suggestion, but we already cover the tool guardrail patterns in the docs.

PR #3426, opened 2026-05-15
fix: honor short voice splitter chunks

### LINKED ISSUE #3363: Voice text splitter drops short custom chunks
### Please read this first

- **Have you searched for related issues?** Yes. I searched existing issues and PRs for `voice text_splitter short chunk` and did not find a matching report or fix.

### Describe the bug

`StreamedAudioResult` drops text when a custom TTS `text_splitter` returns a non-empty chunk shorter than 20 characters. This can silently omit short responses such as `"ok"` even though the splitter explicitly marked the text as ready for TTS.

The default sentence splitter already applies its own minimum sentence length before returning a chunk. Once a custom splitter returns non-empty text, `StreamedAudioResult` should treat that as the splitter's decision and send the chunk to the TTS model.

I also checked adjacent voice streaming behavior. The same path needs to emit `turn_ended` when a custom splitter consumes all buffered text before `_turn_done()` is called; otherwise a turn can start without a matching turn-ended lifecycle event.

### Debug information

### Repro steps

Run this minimal script:

```python
import asyncio
from collections.abc import AsyncIterator

import numpy as np

from agents.voice import StreamedAudioResult, TTSModel, TTSModelSettings, VoicePipelineConfig


class RecordingTTS(TTSModel):
    def __init__(self):
        self.texts = []

    @property
    def model_name(self) -> str:
        return "recording_tts"

    async def run(self, text: str, settings: TTSModelSettings) -> AsyncIterator[bytes]:
        self.texts.append(text)
        yield np.zeros(2, dtype=np.int16).tobytes()


def split_immediately(text: str) -> tuple[str, str]:
    return text, ""


async def main():
    tts = RecordingTTS()
    result = StreamedAudioResult(
        tts,
        TTSModelSettings(buffer_size=1, text_splitter=split_immediately),
        VoicePipelineConfig(),
    )
    await result._add_text("ok")
    await result._turn_done()
    await result._done()

    events = []
    audio_chunks = 0
    async for event in result.stream():
        if event.type == "voice_stream_event_lifecycle":
            events.append(event.event)
        elif event.type == "voice_stream_event_audio":
            events.append("audio")
            audio_chunks += 1
    print({"tts_texts": tts.texts, "events": events, "audio_chunks": audio_chunks})


asyncio.run(main())
```

Actual result:

```text
{'tts_texts': [], 'events': ['turn_started', 'session_ended'], 'audio_chunks': 0}
```

### Expected behavior

The non-empty splitter chunk should be sent to TTS even though it is shorter than 20 characters. The run should produce one audio event and a balanced lifecycle sequence, for example:

```text
{'tts_texts': ['ok'], 'events': ['turn_started', 'audio', 'turn_ended', 'session_ended'], 'audio_chunks': 1}
```

he-yufeng:
## Summary
- send non-empty custom splitter chunks to TTS even when they are shorter than 20 characters
- preserve `turn_ended` when a custom splitter consumes the whole buffer before `_turn_done()`
- add a regression test for short custom chunks

Fixes #3363.

## To verify
- `uv run pytest tests\voice\test_pipeline.py -q --basetemp .tmp\pytest-voice`
- `uv run pyright src\agents\voice\result.py tests\voice\test_pipeline.py`
- `uv run ruff format --check src\agents\voice\result.py tests\voice\test_pipeline.py`
- `uv run ruff check src\agents\voice\result.py tests\voice\test_pipeline.py`
- `uv run python -m py_compile src\agents\voice\result.py tests\voice\test_pipeline.py`
- `git diff --check`

### DIFF
--- a/src/agents/voice/result.py
+++ b/src/agents/voice/result.py
@@ -201,7 +201,7 @@ class StreamedAudioResult:
 
         combined_sentences, self._text_buffer = self.tts_settings.text_splitter(self._text_buffer)
 
-        if len(combined_sentences) >= 20:
+        if combined_sentences:
             local_queue: asyncio.Queue[VoiceStreamEvent | None] = asyncio.Queue()
             self._ordered_tasks.append(local_queue)
             self._tasks.append(
@@ -220,6 +220,10 @@ class StreamedAudioResult:
                 )
             )
             self._text_buffer = ""
+        elif self._started_processing_turn:
+            local_queue: asyncio.Queue[VoiceStreamEvent | None] = asyncio.Queue()
+            await local_queue.put(VoiceStreamEventLifecycle(event="turn_ended"))
+            self._ordered_tasks.append(local_queue)
         self._done_processing = True
         if self._dispatcher_task is None:
             self._dispatcher_task = asyncio.create_task(self._dispatch_audio())
--- a/tests/voice/test_pipeline.py
+++ b/tests/voice/test_pipeline.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+from collections.abc import AsyncIterator
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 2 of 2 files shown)


seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/3364

PR #3440, opened 2026-05-17
Docs: add auditable final output receipt guidance

aDragon0707:
## Summary

Adds a docs-only section to `docs/results.md` about auditable final output receipts.

The new section clarifies that `final_output` is the agent's final answer, not proof that every claim in the answer is supported. It shows a compact receipt pattern for safety-sensitive workflows where agent completion claims should be tied to evidence, next owner, and human approval boundaries.

## Motivation

Agent final answers can contain multiple operational claims, such as:

- task completed
- tests passed
- ready to publish
- memory saved

Those claims may require evidence from `new_items`, tool outputs, approval metadata, guardrail results, file diffs, logs, traces, or human review. A receipt complements tracing and guardrails by summarizing the final work state for humans and downstream systems.

## Notes

- Docs-only change.
- No SDK behavior change.
- No runtime dependency.
- Safety-oriented guidance for final output review.

### DIFF
--- a/docs/results.md
+++ b/docs/results.md
@@ -159,3 +159,50 @@ These arrays accumulate across the run, so they are useful for logging decisions
 [`context_wrapper`][agents.result.RunResultBase.context_wrapper] exposes your app context together with SDK-managed runtime metadata such as approvals, usage, and nested `tool_input`.
 
 Usage is tracked on `context_wrapper.usage`. For streamed runs, the usage totals can lag until the stream's final chunks have been processed. See [Context management](context.md) for the full wrapper shape and persistence caveats.
+
+## Auditable final output receipts
+
+For safety-sensitive workflows, treat `final_output` as the agent's final answer, not as proof that every claim in the answer is supported.
+
+A final answer such as:
+
+```text
+Done. All tests passed. Ready to publish.
+```
+
+contains several different claims:
+
+- the task is complete
+- tests passed
+- the result is ready for an external action
+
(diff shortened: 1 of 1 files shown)


Aigen-Protocol:
The receipt pattern described here connects to a broader design question we worked through building a cross-framework agent protocol: what makes a completion claim *independently verifiable* rather than self-reported?

Three design decisions that came out of that work, which may be relevant here:

**1. Artifact hash vs. embedding**
Including a SHA-256 of the final output in the receipt — rather than the full output — keeps the receipt compact and lets any downstream system verify completion claims without fetching the full trace. The receipt becomes a stable, signable reference.

**2. Server signature vs. agent self-attestation**
A receipt signed by the orchestrator/server rather than the agent itself can be verified with just a public key, without calling a live endpoint. Agents self-reporting their own completion is not independently verifiable in multi-agent handoff scenarios.

**3. Settlement binding**
For receipts that cross agent boundaries (agent A hands off to agent B), binding the receipt to the actual settlement event (payment, approval, external write) closes the gap between 'agent said it was done' and 'it actually took effect downstream.'

We captured these as a normative schema in [AIP-3 §10](https://github.com/Aigen-Protocol/aigen-protocol/blob/main/specs/AIP-3.md) for portability across frameworks — sharing here in case any of the patterns are useful for the guidance section you're writing.

seratch:
Thanks for sharing this. While we won't have this in our own docs, please feel free to share it in your own resources.

PR #3444, opened 2026-05-18
docs: fix LiteLLM API reference redirect

ynachiket:
## Summary

- Fix the legacy LiteLLM API reference redirect target
- Update the fallback Markdown link to point at the existing LiteLLM model API reference

## Validation
- git diff --check
- Targeted content assertion confirmed the stale third_party_adapters target is gone, models/litellm_model is present, and docs/ref/extensions/models/litellm_model.md exists


seratch (approved):
Thank you for fixing this issue!

PR #3456, opened 2026-05-19
docs: add SECURITY.md

ioleksiuk:
## Summary
- Adds `SECURITY.md` matching the format used by sibling project [`openai/openai-agents-js`](https://github.com/openai/openai-agents-js/blob/main/SECURITY.md).
- Provides an explicit email contact (`disclosure@openai.com`) so the GitHub Security tab surfaces a clear reporting channel instead of "This project has not set up a SECURITY.md file yet".

## Context
The repo currently has no `SECURITY.md`, so the Security tab shows no reporting guidance. This brings the repo in line with `openai/openai-agents-js`, `openai/swarm`, and `openai/evals`.

## Test plan
- N/A (markdown only).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+For a more in-depth look at our security policy, please check out our [Coordinated Vulnerability Disclosure Policy](https://openai.com/policies/coordinated-vulnerability-disclosure-policy).
+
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the suggestion. While we may revisit the content in the future, I've added exactly the same one with openai-agents-js repo for now.

PR #3458, opened 2026-05-19
Support function_tool on instance methods

### LINKED ISSUE #94: Add `method_tool` Functionality

adit24dhaya:
### Summary

- Support using `@function_tool` on instance methods by binding the decorated tool when accessed from an instance.
- Omit the implicit `self` parameter from generated tool schemas while preserving normal function-tool behavior.
- Add regression coverage for instance-method tools, including context after `self`, and document the usage pattern.

### Test plan

- `.venv/bin/python -m pytest tests/test_function_tool.py`
- `.venv/bin/ruff format --check src/agents/function_schema.py src/agents/tool.py tests/test_function_tool.py`
- `.venv/bin/ruff check src/agents/function_schema.py src/agents/tool.py tests/test_function_tool.py`
- `.venv/bin/pyright src/agents/function_schema.py src/agents/tool.py tests/test_function_tool.py`
- `git diff --check`

### Issue number

Closes #94

### Checks

- [ ] I've run `make lint` and `make format` (Ruff format/check commands above passed locally)


seratch:
Thanks for your effort here. We'll close this PR for the reason I mentioned at https://github.com/openai/openai-agents-python/issues/94#issuecomment-5156046423

PR #3451, opened 2026-05-18
fix: runtime handling updates

adrianbravo-oai:
## Summary
- Refresh runtime handling around session and tool-call flows.
- Adjust model configuration metadata used by runtime integrations.
- Add focused coverage for the updated behavior.

## Validation
- .venv/bin/python -m pytest tests/model_settings/test_serialization.py tests/models/test_trace_config.py tests/mcp/test_streamable_http_client_factory.py tests/test_run_context_approvals.py tests/test_run_state.py::TestRunState::test_trace_api_key_serialization_is_opt_in tests/realtime/test_session.py
- .venv/bin/ruff check <touched files>
- .venv/bin/ruff format --check <touched files>
- git diff --check

PR #3466, opened 2026-05-20
fix: apply hardened http client default to MCP SSE transport

ioleksiuk:
## Summary
Mirror the redirect-hardening pattern from #3451 to the SSE transport.

`MCPServerStreamableHttp.create_streams` (`server.py:1444-1446`) always applies `_create_default_streamable_http_client`, which sets `follow_redirects=False`. `MCPServerSse.create_streams` (`server.py:1300-1301`) only forwarded the factory when the user supplied one, so the default path fell through to MCP SDK's `create_mcp_http_client`, which has `follow_redirects=True`.

This PR aligns the two transports so the default SSE client picks up the same hardening as StreamableHTTP.

## Test plan
- Updated `tests/mcp/test_mcp_auth_params.py::test_sse_default_no_auth_no_factory` and `test_sse_with_auth` to assert the default factory is now always forwarded.
- Existing SSE tests with user-provided factory (`test_sse_with_httpx_client_factory`, `test_sse_with_auth_and_factory`) continue to pass unchanged.
- `uv run pytest tests/mcp/` — 209 passed.
- `uv run ruff check` / `ruff format --check` — clean.

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -1297,8 +1297,9 @@ class MCPServerSse(_MCPServerWithClientSession):
         }
         if "auth" in self.params:
             kwargs["auth"] = self.params["auth"]
-        if "httpx_client_factory" in self.params:
-            kwargs["httpx_client_factory"] = self.params["httpx_client_factory"]
+        kwargs["httpx_client_factory"] = (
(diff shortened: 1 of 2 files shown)


seratch:
@codex review

seratch (approved):

Issue #3471, opened 2026-05-20
Support Responses API background mode in Runner (background=True + adaptive polling)


dtlics:
### Summary

Add first-class support for the Responses API's [background mode](https://platform.openai.com/docs/guides/background) to `Runner`, so users can run long agent turns (gpt-5.2-pro, deep-research-class workloads) without hitting HTTP / proxy / serverless timeouts.

Today, passing `background=True` through `model_settings.extra_args` reaches `client.responses.create()` but returns a non-terminal `Response` with `status="queued"`, and the turn loop has no path to poll it to completion — the turn breaks. This proposes adding that path at the model-adapter layer.

Related: openai/openai-agents-js#651 (same request on the JS sibling, still open, labeled `enhancement`). Microsoft Agent Framework shipped the equivalent in .NET (`AllowBackgroundResponses`) and the [docs note](https://learn.microsoft.com/en-us/agent-framework/user-guide/agents/agent-background-responses) Python support is "coming soon" — design precedent exists.

### Motivation

- Reasoning-heavy single turns can take minutes; HTTP timeouts (Vercel ~60s, Cloudflare Workers ~30s, many corporate proxies) make synchronous calls unreliable.
- The underlying `openai-python` client already supports `background=True` on `responses.create()` plus polling via `responses.retrieve(id)`. The Agents SDK is the natural place to absorb that loop so users don't have to drop out of `Runner` to use it.
- Plain `openai-python` deliberately ships *no* `responses.create_and_poll()` helper — it exposes primitives and lets callers build the loop. The Agents SDK, by contrast, owns the turn loop, so absorbing the poll loop transparently is the right division of labor.

### Proposed design (MVP — in-process polling only)

1. **`ModelSettings` flag**, appended at the end of the dataclass per the positional-compatibility rule in `AGENTS.md`:
```python
   background: bool | None = None
   background_poll_interval_seconds: float | None = None  # optional override; default adapts to header
```
   Choosing `ModelSettings` over `RunConfig` to match the existing family of model-call toggles (`store`, `reasoning`, `prompt_cache_retention`, `context_management`) and to give per-agent granularity in multi-agent runs.

2. **Model layer.** In `OpenAIResponsesModel.get_response`, when `background=True`:
   - Pass `background=True` to `responses.create()`.
[...]

PR #3472, opened 2026-05-20
feat: Responses API background mode (background=True + adaptive polling)

dtlics:
Refs #3471.

> Draft. Design questions in #3471 are still open — posting this PR alongside the issue to make the design concrete, not to request review. Will mark ready for review after maintainer feedback on the four open questions.

## What this branch does

- Appends `background: bool | None` and `background_poll_interval_seconds: float | None` to `ModelSettings` (preserves positional ordering per `AGENTS.md`).
- Adds a submit-and-adaptive-poll loop to `OpenAIResponsesModel.get_response` via a new private `_poll_background_response_until_terminal`. Honors `openai-poll-after-ms` response headers; explicit `background_poll_interval_seconds` overrides; falls back to 1.0s. On `asyncio.CancelledError` or terminal failure, schedules `client.responses.cancel(id)` fire-and-forget so server-side work doesn't leak.
- `stream_response` is unchanged at the call-site level — `background=True` flows through `_build_response_create_kwargs` into the streaming `responses.create()` call, giving server-side durability without client-side resume logic.
- `OpenAIChatCompletionsModel` and `OpenAIResponsesWSModel` raise `UserError` when `background=True` is set, so users don't silently lose the durability guarantee they opted into.
- Docs page at `docs/background.md` registered in `mkdocs.yml`; runnable example at `examples/background_mode/main.py`.

Diff: 791 insertions, 1 deletion across 10 files. No changes to `run.py`, anything under `run_internal/`, or `run_state.py` — `CURRENT_SCHEMA_VERSION` is not bumped.

## Design choices currently reflected in the code

These map to the four open questions in #3471. All are easy to flip — happy to restructure based on maintainer preference.

1. **Surface — `ModelSettings`, not `RunConfig`.** Joins the existing family of model-call toggles (`store`, `reasoning`, `prompt_cache_retention`, `context_management`) and gives per-agent granularity in multi-agent runs.
2. **Non-Responses backends — `UserError`, not silent no-op.** Loud failure, on the reasoning that users opted into a server-side durability guarantee that those transports can't provide. Counterpoint (raised in #3471): `reasoning` / `verbosity` already silently no-op on those backends, so picking the opposite policy here is an inconsistency. Easy to switch to silent no-op if preferred.
3. **Streaming — server-side durability only, no client-side `starting_after` auto-resume.** Mirrors plain `openai-python`'s behavior (it exposes `responses.stream(response_id=..., starting_after=N)` as a primitive but doesn't auto-resume on disconnect).
4. **Retrieve-call retries — `AsyncOpenAI.max_retries` only (partial).** Each retrieve call uses the client's built-in `max_retries` for transient HTTP failures. **Honest gap:** the plan's intent was that an exhausted-retry failure during polling should not propagate up to `get_response_with_retry` and trigger a fresh submit+poll cycle (which would discard a possibly-minutes-long in-flight reasoning response). The current code does *not* mark such exceptions as non-retriable, so outer-envelope replay can still happen subject to retry policy. The cancel-on-CancelledError half of Q4 is implemented; the suppress-outer-retry half is not. Will fix once Q4 lands one way or the other.

## Test plan

`make format && make lint && make typecheck && make tests` all pass on the branch:

- `make format` — `ruff format`: 778 files left unchanged. `ruff check --fix`: All checks passed.
- `make lint` — `ruff check`: All checks passed.
- `make typecheck` — `mypy`: Success, no issues found in 772 source files. `pyright`: 0 errors, 0 warnings, 0 informations.
- `make tests` — parallel: 4589 passed, 2 skipped in 16.61s. Serial: 27 passed, 5 skipped, 4590 deselected in 4.82s.

New tests (in `tests/models/test_openai_responses.py` and `tests/models/test_openai_chatcompletions.py`) cover: terminal-on-first-response fast path, queued→in_progress→completed multi-poll, terminal `failed` / `cancelled` / `incomplete` raises, `openai-poll-after-ms` header honored, explicit interval overrides header, `CancelledError` during poll schedules `responses.cancel`, `extra_args={"background": True}` conflict raises `TypeError`, streaming pass-through, and Chat Completions / WS `UserError` rejections.

Not verified by me: `make build-docs` / `make build-full-docs` were not run. The new docs page is registered in all four language nav sections in `mkdocs.yml` (en/ja/ko/zh) on the assumption the existing translation pipeline fills in non-English content — worth a reviewer eye.

## Out of scope (proposed follow-ups, mirroring #3471)

- Stream auto-resume via `responses.stream(response_id=..., starting_after=N)` on transport errors.
- Cross-process `Runner` resume / continuation tokens — would need `RunState` schema bump.
- ZDR enforcement — background mode is not ZDR-compatible and retains response data for ~10 minutes server-side.
- Auto-detection of when to use background mode based on model / payload.
- `OpenAIResponsesModel.retrieve_response()` helper — users can call `client.responses.retrieve(id)` directly on the underlying client, so this would add public API surface without adding capability.

### DIFF
--- a/docs/background.md
+++ b/docs/background.md
@@ -1,0 +1,62 @@
+# Background mode
+
+OpenAI's [Responses API background mode](https://platform.openai.com/docs/guides/background) lets long-running model calls survive client disconnects: the server keeps processing the request and you poll it to completion. This matters for reasoning-heavy single turns (`gpt-5.2-pro`, deep-research models) that can take minutes and otherwise fall foul of HTTP timeouts on Vercel, Cloudflare Workers, corporate proxies, etc.
+
+The Agents SDK exposes background mode via two new fields on [`ModelSettings`][agents.model_settings.ModelSettings]:
+
+- `background: bool | None` — opt in to background mode.
+- `background_poll_interval_seconds: float | None` — optional fixed poll interval. When unset, the SDK honors the `openai-poll-after-ms` response header and falls back to 1.0 second.
+
+## Transparent use through `Runner`
+
+Set the flag on your agent's `ModelSettings` and run as usual. The SDK submits with `background=True`, polls `client.responses.retrieve(id)` adaptively, and returns the terminal response — `Runner.run` and `Runner.run_streamed` need no other changes.
+
+```python
+from agents import Agent, ModelSettings, Runner
+
+agent = Agent(
+    name="reasoner",
+    model="gpt-5.2-pro",
+    model_settings=ModelSettings(background=True),
+)
+result = await Runner.run(agent, "Plan a multi-stage research workflow.")
+print(result.final_output)
+```
+
+For streaming, `background=True` is passed through to `responses.create(stream=True, background=True)` so the server keeps generating across client disconnects. Client-side auto-resume via `starting_after` is intentionally not part of this MVP — plain `openai-python` doesn't auto-resume either.
+
+```python
+async for event in Runner.run_streamed(agent, "Stream me a long answer").stream_events():
+    print(event)
+```
+
+## Retrieving a response by id
+
+If you captured a `response_id` and want to fetch the latest server state from a different process or worker, call `client.responses.retrieve(response_id)` on the underlying `AsyncOpenAI` client directly — there is no SDK-specific wrapper, deliberately, because that would only add API surface without adding capability.
+
+```python
+from openai import AsyncOpenAI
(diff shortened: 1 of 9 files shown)


seratch:
Closing this for this reason: https://github.com/openai/openai-agents-python/issues/3471#issuecomment-4503397795 Thanks for your interest here.

PR #3479, opened 2026-05-21
feat: add MCP tool call result callback

### LINKED ISSUE #3477: Read `_meta` from MCP Tool Call Responses
### Please read this first

- [x] **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)

# Feature Request: Read `_meta` from MCP Tool Call Responses

> This is a follow-up to [#2367](https://github.com/openai/openai-agents-python/issues/2367) / [PR #2375](https://github.com/openai/openai-agents-python/pull/2375), which added support for **sending** `_meta` in MCP tool call requests.
>
> This request is for the complementary feature: **reading** `_meta` from MCP tool call **responses**.

## Problem

PR #2375 enabled passing `_meta` from the client (Agents SDK) to the MCP server via `tool_meta_resolver`. This is great for request tracing and context propagation.

However, the **reverse direction** is not supported: when an MCP server returns `_meta` in its `CallToolResult`, the Agents SDK currently **ignores it entirely**.

Looking at the source code (`agents/mcp/util.py`, `invoke_mcp_tool`, lines 666-689):

```python
# Only these fields are read from the result:
if server.use_structured_content and result.structuredContent:
    tool_output = json.dumps(result.structuredContent)
else:
    for item in result.content:
        # ... handles text/image content

# result._meta is NEVER accessed
```

The MCP protocol spec defines `CallToolResult._meta` as:

```typescript
interface CallToolResult {
  _meta?: { [key: string]: unknown };  // ← present in the spec, ignored by SDK
  content: ContentBlock[];
  structuredContent?: { [key: string]: unknown };
  isError?: boolean;
}
```

## Our Use Case

We are building a **data analysis assistant** using the Agents SDK. The architecture involves:

- A main Agent that handles user conversations
- Sub-agents (as tools) that perform data analysis (SQL queries, chart generation)
- An MCP server providing prompts and tools for database interaction

### The Core Need: Dual-Output Tools

Our tools need to return **two different outputs**:

1. **For the LLM**: A compact, token-efficient summary (e.g., "Query returned 1,245 rows")
2. **For the frontend**: The full data payload (e.g., chart configuration JSON, raw data for rendering)

The `_meta` field in the MCP response is the natural place for this frontend-facing data:

```json
{
  "content": [
    { "type": "text", "text": "Sales increased 23% in Q1." }
  ],
  "_meta": {
    "tool_meta": {
      "type": "chart",
      "vis_config": {
        "type": "line",
        "data": [{"month": "Jan", "sales": 120000}, ...]
      }
    }
  }
}
```

### Why Not Put It in `content`?

If we put the full `vis_config` in `result.content`, it gets fed back to the LLM as context, wasting tokens on data the LLM doesn't need (and shouldn't see in full). The `_meta` field is explicitly defined in the MCP spec as metadata separate from the primary content.

### Why Not Use `RunContextWrapper`?

`RunContextWrapper` works for **local tools** (via `@function_tool`), but our tools run in an **MCP server process** (stdio or HTTP). The MCP server cannot access the caller's `RunContextWrapper` because:
- stdio MCP runs in a separate process
- HTTP MCP runs on a remote host

The `_meta` field is the MCP-native mechanism for passing auxiliary data back to the client.

### Current Workaround and Its Limitations

Our current workaround is to **not use MCP for these tools at all** — we register them as local `FunctionTool`s so they can write to `ctx.context`. This means:
- We cannot leverage MCP for tool distribution across services
- We lose the MCP ecosystem benefits (standardized discovery, transport abstraction)
- We have to maintain two separate tool registration paths (MCP for prompts, local for data tools)

## Proposed Solution

Expose the response `_meta` through the SDK's tool execution pipeline. A few possible approaches:

### Option A: Attach to `ToolContext` / `RunContextWrapper`

After `invoke_mcp_tool` receives the result, write `result._meta` into the current run context:

```python
# In invoke_mcp_tool, after receiving result:
if result._meta and hasattr(context, 'context') and isinstance(context.context, dict):
    context.context.setdefault("_mcp_tool_meta", {})[tool.name] = result._meta
```

### Option B: Attach to `RunItem` / `ToolCallOutputItem`

Extend `ToolCallOutputItem` (or the internal tool result representation) to carry `_meta`:

```python
@dataclass
class ToolCallOutputItem:
    output: Any
    mcp_meta: dict[str, Any] | None = None  # ← new field
```

This would allow event stream consumers to access `_meta` via `event.item.mcp_meta`.

### Option C: Return Tuple from `invoke_mcp_tool`

Change `invoke_mcp_tool` to return both the tool output and the meta:

```python
# Current signature:
async def invoke_mcp_tool(...) -> ToolOutput

# Could be extended to return meta alongside:
async def invoke_mcp_tool(...) -> ToolOutput
# But meta is captured and stored somewhere accessible
```

Option B feels the cleanest as it aligns with the existing event stream architecture.

## Related

- MCP spec: [`CallToolResult`](https://modelcontextprotocol.io/specification/2025-11-25/basic#general-fields)
- PR #2375: Added `tool_meta_resolver` for request `_meta`
- Issue #2367: Original request for request-side `_meta`

---

**Would the maintainers be open to a PR implementing Option B?** We're happy to contribute.

seratch:
This pull request adds a `tool_call_result_callback` option for local MCP servers so applications can observe MCP `CallToolResult` metadata without changing the model-visible tool output. The callback receives an `MCPToolCallResultContext` with the original MCP tool name, public display name, parsed arguments, `_meta`, `structuredContent`, `isError`, and a copied tool output, which lets apps collect frontend-facing data for dual-output MCP tools while keeping replay and model history behavior unchanged.

This pull request resolves #3477.

```python
from agents.mcp import MCPServerStreamableHttp, MCPToolCallResultContext


def tool_call_result_callback(context: MCPToolCallResultContext) -> None:
    app_context = context.run_context.context
    if not isinstance(app_context, dict) or context.result_meta is None:
        return

    app_context.setdefault("frontend_events", []).append(
        {
            "server": context.server_name,
            "tool": context.tool_display_name,
            "meta": dict(context.result_meta),
        }
    )


server = MCPServerStreamableHttp(
    name="Metadata-aware MCP",
    params={"url": "http://localhost:8000/mcp"},
    tool_call_result_callback=tool_call_result_callback,
)
```

### DIFF
--- a/src/agents/mcp/__init__.py
+++ b/src/agents/mcp/__init__.py
@@ -17,6 +17,8 @@ if TYPE_CHECKING:
     )
 
 from .util import (
+    MCPToolCallResultCallback,
+    MCPToolCallResultContext,
     MCPToolMetaContext,
     MCPToolMetaResolver,
     MCPUtil,
@@ -50,6 +52,8 @@ __all__ = [
     "MCPServerManager",
     "LocalMCPApprovalCallable",
     "MCPUtil",
+    "MCPToolCallResultCallback",
+    "MCPToolCallResultContext",
     "MCPToolMetaContext",
     "MCPToolMetaResolver",
     "ToolFilter",
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -45,6 +45,7 @@ from ..tool import ToolErrorFunction
 from ..util._types import MaybeAwaitable
 from .util import (
     HttpClientFactory,
+    MCPToolCallResultCallback,
     MCPToolMetaResolver,
     ToolFilter,
     ToolFilterContext,
@@ -229,6 +230,7 @@ class MCPServer(abc.ABC):
         require_approval: RequireApprovalSetting = None,
         failure_error_function: ToolErrorFunction | None | _UnsetType = _UNSET,
         tool_meta_resolver: MCPToolMetaResolver | None = None,
+        tool_call_result_callback: MCPToolCallResultCallback | None = None,
     ):
         """
         Args:
@@ -248,6 +250,9 @@ class MCPServer(abc.ABC):
                 SDK default) will be used.
             tool_meta_resolver: Optional callable that produces MCP request metadata (`_meta`) for
                 tool calls. It is invoked by the Agents SDK before calling `call_tool`.
+            tool_call_result_callback: Optional callback invoked after an MCP tool call returns. The
+                callback receives result metadata and the model-visible tool output, but cannot
+                change the output returned to the model.
         """
         self.use_structured_content = use_structured_content
         self._needs_approval_policy = self._normalize_needs_approval(
@@ -255,6 +260,7 @@ class MCPServer(abc.ABC):
         )
         self._failure_error_function = failure_error_function
         self.tool_meta_resolver = tool_meta_resolver
+        self.tool_call_result_callback = tool_call_result_callback
 
     @abc.abstractmethod
     async def connect(self):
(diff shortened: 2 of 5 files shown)


seratch:
Closing this in favor of https://github.com/openai/openai-agents-python/pull/3486

PR #3480, opened 2026-05-21
feat(mcp): surface response _meta on ToolCallOutputItem

### LINKED ISSUE #3477: Read `_meta` from MCP Tool Call Responses

zhang-liz:
### Summary

Capture the optional `_meta` field returned by MCP servers in `CallToolResult` and surface it on `ToolCallOutputItem` as `mcp_response_meta`. Applications can read auxiliary payloads (chart configs, trace IDs, frontend-only state) without forwarding them to the model.

This is the response-side complement to PR #2375 (`tool_meta_resolver`, request-side `_meta`). The field is a deep copy and is never injected into model context — only `content` / `structuredContent` are.

**Implementation:** `invoke_mcp_tool` stashes `result.meta` on the per-call `ToolContext`; the function-tool executor moves it onto the resulting `ToolCallOutputItem`. RunState schema bumped to 1.11 so the field round-trips through pause/resume.

### Test plan

- New unit tests in `tests/mcp/test_mcp_util.py` cover the ToolContext stash (populated, omitted, empty-dict; verifies deep copy).
- New end-to-end tests in `tests/mcp/test_runner_calls_mcp.py` assert `ToolCallOutputItem.mcp_response_meta` is populated (streaming + non-streaming) and stays `None` when the server omits `_meta`.
- New serialization round-trip tests in `tests/test_run_state.py` for schema 1.11.
- `make format`, `make lint` clean. `mypy` clean on all touched files. Full `make tests` adds 8 new passing tests; only pre-existing unrelated failures (numpy/litellm/runloop optional deps) remain — verified identical to baseline (`main` shows the same 12 failed / 13 errors).

### Issue number

Closes #3477

### Checks

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -277,6 +277,33 @@ server = MCPServerStreamableHttp(
 
 If your run context is a Pydantic model, dataclass, or custom class, read the tenant ID with attribute access instead.
 
+### Reading response `_meta` from MCP tools
+
+The MCP specification allows a server's `CallToolResult` to include a `_meta` object alongside `content` and `structuredContent`. The SDK captures it on the resulting `ToolCallOutputItem` as `mcp_response_meta`, so applications can read auxiliary payloads (chart configs, trace IDs, frontend-only state) without forwarding them to the model.
+
+`mcp_response_meta` is a deep copy of the server-returned dict and is never injected into model context — only `content` and `structuredContent` are. Use it for application-side consumers such as streaming UIs.
+
+```python
+from agents import Agent, Runner
+from agents.items import ToolCallOutputItem
+from agents.mcp import MCPServerStreamableHttp
+
+server = MCPServerStreamableHttp(
+    name="meta-emitting server",
+    params={"url": "http://localhost:8000/mcp"},
+)
+
+agent = Agent(name="assistant", mcp_servers=[server])
+result = await Runner.run(agent, input="show me a chart of Q1 sales")
+
+for item in result.new_items:
+    if isinstance(item, ToolCallOutputItem) and item.mcp_response_meta:
+        # e.g. {"tool_meta": {"type": "chart", "vis_config": {...}}}
+        frontend_payload = item.mcp_response_meta
+```
+
+When streaming, the same item is delivered as a `run_item_stream_event` with `name="tool_output"`; read `event.item.mcp_response_meta` on receipt.
+
 ### MCP tool outputs: text and images
 
 When an MCP tool returns image content, the SDK maps it to image tool output entries automatically. Mixed text/image responses are forwarded as a list of output items, so agents can consume MCP image results the same way they consume image output from regular function tools.
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -403,6 +403,15 @@ class ToolCallOutputItem(RunItemBase[Any]):
     tool_origin: ToolOrigin | None = None
     """Optional metadata describing the source of a function-tool-backed item."""
 
+    mcp_response_meta: dict[str, Any] | None = None
+    """Optional ``_meta`` returned by an MCP server in ``CallToolResult``.
+
+    Populated when this item was produced by an MCP tool call whose server returned a
+    non-empty ``_meta`` field, per the MCP specification. The value is a deep copy and is
+    not forwarded to the model; it is intended for application-side consumers (e.g.
+    streaming UIs that need auxiliary payloads without polluting LLM context).
+    """
+
     @property
     def call_id(self) -> str | None:
         """Return the call identifier from the raw item, if available."""
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -663,6 +663,10 @@ class MCPUtil:
         else:
             logger.debug(f"MCP tool {tool_name_for_display} returned {result}")
 
+        response_meta = getattr(result, "meta", None)
+        if isinstance(response_meta, dict) and response_meta and isinstance(context, ToolContext):
+            context._mcp_response_meta = copy.deepcopy(response_meta)
+
         # If structured content is requested and available, use it exclusively
         tool_output: ToolOutput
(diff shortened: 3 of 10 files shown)


seratch:
Thanks for sharing this idea. However, I'd like to avoid this change because this is less flexible and it is a larger change than we'd like to have for this purpose.

zhang-liz:
Thanks for the quick review @seratch. Could you point me at the direction you'd prefer? I see two ways to slim this down:

1. **Callback-based**: add an optional `on_response_meta` callback on the MCP server (mirroring `tool_meta_resolver`'s shape). Server-returned `_meta` is delivered to user code with `(tool_context, meta)`; no changes to `ToolCallOutputItem`, no schema bump, no executor plumbing.
2. **Context-only**: stash `_meta` on `ToolContext` (private attr only) and let users read it from a custom `RunHooks.on_tool_end` hook. Zero new public surface.

Happy to rework as (1) or (2), or close this PR if you'd rather not have either. Want to make sure I'm aiming at the right shape before sending a v2.

seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/3486

PR #3486, opened 2026-05-22
feat: add SDK-only custom data for tool outputs

seratch:
This pull request adds SDK-only `custom_data` support to tool output items so applications can attach renderer hints, IDs, or other JSON-compatible metadata without sending it back to the model.

It adds extractor callbacks for function tools, local MCP server tools, custom tools, `ComputerTool`, and `ApplyPatchTool`, persists the data through `RunState`. `ShellTool` remains unchanged because it already has shell-specific `provider_data`, and deprecated `LocalShellTool` is not extended.


siddiksawani:
Hi @seratch 
I noticed one small follow-up case around JSON compatibility validation: json.dumps() allows NaN, Infinity, and -Infinity by default, so custom_data_extractor can currently accept non-finite floats even though the data is expected to be JSON-compatible.

I opened a small follow-up PR to tighten this by using allow_nan=False and adding a regression test: 
#3657

PR #3491, opened 2026-05-22
fix: export ApplyPatchTool and CustomTool approval callback types from top-level package

rmotgi1227:
`ApplyPatchTool` and `CustomTool` each accept an `on_approval` callback, and `ComputerTool` accepts an `on_safety_check` callback. The types needed to annotate those callbacks are not exported from the top-level `agents` package:

- `ApplyPatchOnApprovalFunction` / `ApplyPatchOnApprovalFunctionResult`
- `CustomToolOnApprovalFunction` / `CustomToolOnApprovalFunctionResult`
- `ComputerToolSafetyCheckData`

The equivalent MCP types (`MCPToolApprovalFunction`, `MCPToolApprovalFunctionResult`) are already exported. This extends that to the remaining tool types.

```python
# currently breaks — have to reach into private agents.tool module:
from agents import ComputerToolSafetyCheckData   # ImportError
from agents import ApplyPatchOnApprovalFunction  # ImportError
from agents import CustomToolOnApprovalFunction  # ImportError
```

Related to #3450 which exports the Shell equivalents.


seratch:
hmm, we'd like to hold off adding these as they are not commonly used plus developers can still import from agents.tool.

PR #3492, opened 2026-05-22
fix: export CallModelInputFilter, CallModelData, ModelInputData, and RunOptions from top-level package

rmotgi1227:
`RunConfig.call_model_input_filter` accepts a `CallModelInputFilter` callable. To implement one, callers need `CallModelData` and `ModelInputData`. These types are defined in `run_config.py` (and listed in its `__all__`), but they are not re-exported from the top-level `agents` package.

`RunOptions` (the `TypedDict` version of `RunConfig` accepted by `Runner.run` / `Runner.run_streamed`) has the same gap.

```python
# currently breaks:
from agents import CallModelInputFilter  # ImportError
from agents import CallModelData         # ImportError
from agents import ModelInputData        # ImportError
from agents import RunOptions            # ImportError

# workaround (private module):
from agents.run_config import CallModelInputFilter, CallModelData, ModelInputData, RunOptions
```

The `ToolOutputTrimmer` extension already implements `CallModelInputFilter` and is documented as a public API, so users need these types to write their own filters.

The fix adds the four names to the `from .run import ...` block and to `__all__` in `src/agents/__init__.py`.


seratch:
@rmotgi1227 run_config is not a private module plus we don't intend to expose everything from top-level agents.

PR #3494, opened 2026-05-22
fix: export TracingConfig, TraceProvider, get_trace_provider, and TraceCtxManager from agents

adityasingh2400:
Four names are present in `agents.tracing.__all__` but are not re-exported from the top-level `agents` module: `TracingConfig`, `TraceProvider`, `get_trace_provider`, and `TraceCtxManager`. The first is documented as the type to pass via `RunConfig.tracing` for per-run trace export settings, and `get_trace_provider` is the natural counterpart to the already-exported `set_trace_provider`. `TraceProvider` is the abstract base for custom trace providers, and `TraceCtxManager` is the context manager used by `Runner.run` and `VoicePipeline.run` to manage trace lifecycles, so users wiring their own runners or pipelines need it from the public surface.

This is the same shape as the recent #3489 and #3490 export-gap fixes: every other name in `agents.tracing.__all__` already appears in `agents.__all__`, and these four were the remaining holdouts. The fix adds them to the `from .tracing import ...` block and to `__all__` in `src/agents/__init__.py`.

A regression test in `tests/tracing/test_tracing_exports.py` iterates `agents.tracing.__all__` and asserts every name is reachable from `agents` and listed in `agents.__all__`, so future drift between the two surfaces fails CI. It mirrors the pattern already used in `tests/test_exception_exports.py`.


seratch:
Thanks for sharing this change, however we don't intend to have everything in the exports from the top-level.

PR #3496, opened 2026-05-22
fix(mcp): use is not None to cache empty tools list correctly

rmotgi1227:
When an MCP server returns an empty tools list, `self._tools_list` is `[]` which is falsy. The cache check

```python
if self.cache_tools_list and not self._cache_dirty and self._tools_list:
```

treats an empty cached list the same as an uncached one, causing a redundant re-fetch on every subsequent `list_tools()` call even when `cache_tools_list=True` and the cache is clean.

The field is initialized to `None` (line 600) to mean "not yet fetched," so the correct guard is `is not None`.


seratch:
An empty list here is not expected plus if that happens, skipping the saved empty data should be fine

PR #3498, opened 2026-05-25
fix: clean up orphaned messages in delete_branch and propagate add_items metadata errors

### LINKED ISSUE #3346: AdvancedSQLiteSession.delete_branch() leaves branch-only messages in the base table

### LINKED ISSUE #3348: AdvancedSQLiteSession.add_items can report success after structure metadata failure

kimnamu:
### Summary

Two correctness fixes for `AdvancedSQLiteSession`, each with a regression test.

**#3346** — `delete_branch()` removes `message_structure` rows but never cleans up the
actual messages, leaking storage indefinitely. The fix reuses the existing
`_cleanup_orphaned_messages_sync()` (which `add_items()` already calls on failure)
so there's no new deletion logic — just a missing call site.

**#3348** — `add_items()` catches `_insert_structure_metadata()` failures, runs cleanup,
but then returns normally. Callers cannot distinguish success from partial failure.
The fix adds a `raise RuntimeError(...) from e` after the existing cleanup block (3 lines).

### Approach

Both fixes reuse existing infrastructure rather than introducing new code:

| Fix | Approach | Why |
|-----|----------|-----|
| #3346 | Call `self._cleanup_orphaned_messages_sync(conn)` after structure deletion | This method already exists at L491 and uses the same `LEFT JOIN ... IS NULL` pattern — no new query logic needed |
| #3348 | Add `raise RuntimeError(...) from e` after the existing `except` cleanup block | The cleanup + logging already exists; only the re-raise was missing |

### Before / After

| Scenario | Before (main) | After (this PR) |
|----------|---------------|-----------------|
| `delete_branch("feat")` with branch-only messages | `message_structure` deleted, orphan rows stay in `messages_table` forever | Orphans removed via existing `_cleanup_orphaned_messages_sync()` |
| `add_items([...])` when metadata fails | Logs error, returns `None` — caller assumes success | Raises `RuntimeError` with cause chain; caller can handle or retry |
| Run Test | image |  image |
| Existing test suite | 36 pass | 36 pass (zero regressions) + 2 new |

### Test plan

- All 36 pre-existing tests pass unchanged (zero regressions)
- `test_delete_branch_cleans_up_orphaned_messages`: creates branch-only messages, deletes branch, asserts `messages_table` is clean
- `test_add_items_raises_on_structure_metadata_failure`: monkeypatches metadata insertion to fail, asserts `RuntimeError` and verifies orphan cleanup ran

```
$ uv run pytest tests/extensions/memory/test_advanced_sqlite_session.py -v
...
38 passed in 0.22s
```
image

### Issue number

Closes #3346
Closes #3348

### Checks

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -153,6 +153,9 @@ class AdvancedSQLiteSession(SQLiteSession):
                     except Exception as cleanup_error:
                         conn.rollback()
                         self._logger.error(f"Failed to cleanup orphaned messages: {cleanup_error}")
+                    raise RuntimeError(
+                        f"Failed to persist structure metadata for session {self.session_id}"
+                    ) from e
 
         await asyncio.to_thread(_add_items_sync)
 
@@ -788,12 +791,18 @@ class AdvancedSQLiteSession(SQLiteSession):
 
                     conn.commit()
 
-                    return usage_deleted, structure_deleted
+                    # Clean up messages that are no longer referenced by any branch.
+                    orphaned_deleted = self._cleanup_orphaned_messages_sync(conn)
+                    if orphaned_deleted:
+                        conn.commit()
 
-        usage_deleted, structure_deleted = await asyncio.to_thread(_delete_sync)
+                    return usage_deleted, structure_deleted, orphaned_deleted
+
+        usage_deleted, structure_deleted, orphaned_deleted = await asyncio.to_thread(_delete_sync)
 
         self._logger.info(
-            f"Deleted branch '{branch_id}': {structure_deleted} message entries, {usage_deleted} usage entries"  # noqa: E501
+            f"Deleted branch '{branch_id}': {structure_deleted} structure entries, "
+            f"{usage_deleted} usage entries, {orphaned_deleted} orphaned messages removed"
         )
 
     async def list_branches(self) -> list[dict[str, Any]]:
--- a/tests/extensions/memory/test_advanced_sqlite_session.py
+++ b/tests/extensions/memory/test_advanced_sqlite_session.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import json
+import sqlite3
 import tempfile
 from pathlib import Path
 from typing import Any, cast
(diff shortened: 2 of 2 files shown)


kimnamu:
@codex review

kimnamu:
Closing in favor of two focused, single-issue PRs to make review easier:

- #3526 — #3346 orphaned message cleanup in `delete_branch`
- #3527 — #3348 propagate `add_items` structure metadata failures

Both reuse the session's existing helpers and each ships with its own regression test. The `AdvancedSQLiteSession` internals (especially the shared `_cleanup_orphaned_messages_sync` helper) made these fixes very clean to slot in — thanks for the well-structured code. Happy to adjust anything on the new PRs.

PR #3504, opened 2026-05-26
fix(runloop): declare SDK model aliases for type checkers

sakinaroufid:
### Summary

This pull request fixes the Runloop sandbox backend's exported SDK model aliases so the backend type-checks cleanly without changing runtime behavior.

The Runloop extension already exports `RunloopAfterIdle`, `RunloopLaunchParameters`, and `RunloopUserParameters` as public symbols, but `sandbox.py` defined them as plain runtime aliases and then reused them in type positions. Pyright and mypy treated those names as variables rather than type aliases and rejected the file when checked directly.

The patch makes those exports explicit `TypeAlias` declarations and adds a concrete annotation for the launch-parameter payload mapping. It also includes a small test-only typing cleanup for `asyncio.eager_task_factory` so the branch does not inherit an existing mypy failure from `main` on the repo's supported Python floor.

### Test plan

- Ran `bash .agents/skills/code-change-verification/scripts/run.sh`
- Ran `make coverage`
- Ran `uv run --python 3.12 pytest tests/test_run_step_execution.py -q -k eager_task_factory`
- Ran `env UV_PROJECT_ENVIRONMENT=.venv_310 uv run --python 3.10 pytest tests/test_run_step_execution.py -q -k eager_task_factory`

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/extensions/sandbox/runloop/sandbox.py
+++ b/src/agents/extensions/sandbox/runloop/sandbox.py
@@ -22,7 +22,7 @@ import uuid
 from collections.abc import Sequence
 from dataclasses import dataclass
 from pathlib import Path, PurePosixPath
-from typing import TYPE_CHECKING, Any, Literal, cast
+from typing import TYPE_CHECKING, Any, Literal, TypeAlias, cast
 from urllib.parse import urlsplit
 
 from pydantic import BaseModel, Field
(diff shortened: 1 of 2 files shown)


sakinaroufid:
@codex review

PR #3509, opened 2026-05-26
fix: remove dependency on types-requests

### LINKED ISSUE #3507: Remove dependency on types-requests
### Please read this first

### Describe the feature

rouge8:
### Summary

Requests 2.34.0 and newer includes inline type hints.

### Test plan

`make typecheck`

### Issue number

Fixes #3507.

### Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,6 @@ dependencies = [
     "griffelib>=2, <3",
     "typing-extensions>=4.12.2, <5",
     "requests>=2.0, <3",
-    "types-requests>=2.0, <3",
     "websockets>=15.0, <17",
     "mcp>=1.19.0, <2; python_version >= '3.10'",
 ]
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing this. requests 2.34 is pretty new, so we'd like to hold off removing types-requests at this moment.

rouge8:
There shouldn't be any behavior changes for any users of the library. Keeping the dependency means that any users of the library can't upgrade requests and have correct type-checking behavior though.

seratch:
Yes, you're right. We'll merge this change eventually, but I just wanted to wait and see the adoption for a short period of time.

seratch (approved):
Thanks for following up on this. The original reason to wait was that Requests 2.34 had just shipped. Requests 2.34.2 has now been available for more than a month, and I confirmed that types-requests overrides its inline annotations in mypy.

PR #3510, opened 2026-05-27
Copilot/update generate content config

Pjrich1313:
``


seratch:
Thanks for sharing this, but we don't see the necessity to have this change for now.

PR #3519, opened 2026-05-28
fix: move types-requests to dev dependencies

### LINKED ISSUE #3507: Remove dependency on types-requests

JackJoltClick:
Closes #3507

types-requests is a type stub package that's only used during development for type checking. It shouldn't be a runtime dependency.

This moves it to the dev dependency group where it belongs, matching how other type packages like types-pynput are handled in this project.


seratch:
Resolved by #3509

PR #3521, opened 2026-05-28
test: add coverage for empty tool descriptions

JackJoltClick:
Adds comprehensive test coverage for edge cases around tool descriptions, including None, empty string, and whitespace-only values.

The tests verify current behavior:
- FunctionTool preserves whitespace descriptions exactly
- Empty string doesn't override docstring (falsy value)
- Handoff falls back to default for falsy descriptions
- Both accept and preserve whitespace-only descriptions

No functionality changes, only test additions to improve coverage.


seratch:
Thanks for sharing this, but we don't see the necessity to add these for now.

PR #3526, opened 2026-05-29
fix: #3346 clean up orphaned messages in delete_branch

kimnamu:
### Summary

`delete_branch()` removes the branch's `message_structure` rows but never cleans up the underlying messages, so messages that were only referenced by the deleted branch stay in the messages table indefinitely.

The session already has a well-factored `_cleanup_orphaned_messages_sync()` helper (the same one `add_items()` relies on for its failure path), so this fix simply calls it after the structure rows are deleted — no new deletion logic, just the missing call site. The log line now also reports how many orphaned messages were removed.

| Scenario | Before (main) | After (this PR) |
|----------|---------------|-----------------|
| `delete_branch("feature")` with branch-only messages | `message_structure` rows deleted, but the messages stay in `messages_table` forever | Orphaned messages removed via the existing `_cleanup_orphaned_messages_sync()` helper |
| Shared messages (referenced by another branch) | Retained | Retained (unchanged) |
| Log output | `Deleted branch 'feature': N message entries, M usage entries` | `Deleted branch 'feature': N structure entries, M usage entries, K orphaned messages removed` |
| New deletion logic introduced | — | None — reuses existing helper |

### Test plan

Adds `test_delete_branch_cleans_up_orphaned_messages`, a regression test that creates a branch-only message, deletes the branch, and asserts the orphaned message is gone while the shared message remains.

```console
$ uv run pytest tests/extensions/memory/test_advanced_sqlite_session.py::test_delete_branch_cleans_up_orphaned_messages -v
tests/extensions/memory/test_advanced_sqlite_session.py::test_delete_branch_cleans_up_orphaned_messages PASSED [100%]
1 passed in 0.26s
```

### Issue number

Closes #3346.

> Note: this is one half of #3498, split into a single-issue PR as suggested for easier review. The other half (#3348) is in #3527.

### Checks

- [x] I've added new tests (if relevant)
- [ ] I've added/updated the relevant documentation
- [x] I've run `make lint` and `make format`
- [x] I've made sure tests pass

PR #3527, opened 2026-05-29
fix: #3348 propagate add_items structure metadata failure

### LINKED ISSUE #3348: AdvancedSQLiteSession.add_items can report success after structure metadata failure

kimnamu:
### Summary

`add_items()` wraps `_insert_structure_metadata()` in a `try/except` that rolls back, logs, and runs orphan cleanup — but then returns normally, so callers can't tell a partial failure from a success.

This adds a single `raise RuntimeError(...) from e` after the existing cleanup block, so the failure surfaces to the caller while preserving the original exception via `from e`. The existing rollback, logging, and cleanup behaviour is untouched.

| Scenario | Before (main) | After (this PR) |
|----------|---------------|-----------------|
| `_insert_structure_metadata()` raises | Exception swallowed; `add_items()` returns normally | `RuntimeError` raised with original exception chained via `from e` |
| Caller observability | Cannot distinguish partial failure from success | Failure surfaces to the caller |
| Rollback / logging / orphan cleanup | Performed | Performed (unchanged) |
| Successful path | Returns normally | Returns normally (unchanged) |

### Test plan

Adds `test_add_items_raises_on_structure_metadata_failure`, which forces `_insert_structure_metadata` to raise, then asserts `add_items()` raises `RuntimeError` and that the orphaned message was still cleaned up.

```console
$ uv run pytest tests/extensions/memory/test_advanced_sqlite_session.py::test_add_items_raises_on_structure_metadata_failure -v
tests/extensions/memory/test_advanced_sqlite_session.py::test_add_items_raises_on_structure_metadata_failure PASSED [100%]
1 passed in 0.18s
```

### Issue number

Closes #3348.

> Note: this is one half of #3498, split into a single-issue PR as suggested for easier review. The other half (#3346) is in #3526.

### Checks

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -153,6 +153,9 @@ class AdvancedSQLiteSession(SQLiteSession):
                     except Exception as cleanup_error:
                         conn.rollback()
                         self._logger.error(f"Failed to cleanup orphaned messages: {cleanup_error}")
+                    raise RuntimeError(
+                        f"Failed to persist structure metadata for session {self.session_id}"
+                    ) from e
 
         await asyncio.to_thread(_add_items_sync)
 
--- a/tests/extensions/memory/test_advanced_sqlite_session.py
+++ b/tests/extensions/memory/test_advanced_sqlite_session.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import json
+import sqlite3
 import tempfile
 from pathlib import Path
 from typing import Any, cast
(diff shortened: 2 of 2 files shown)


kimnamu:
Hi @seratch, thank you for all your work maintaining this SDK! 🙏

Gently following up on this one — it fixes #3348 (which is still open). I noticed the CI checks haven't run yet; I believe that's because this is my first contribution from a fork, so the workflows need a maintainer's approval to start. Whenever you have a moment, would you mind approving the run? No rush at all.

To make review easier, I ran the full check suite locally against the current branch and everything passes:

- `ruff check` → All checks passed
- `mypy` (changed file) → Success: no issues found
- `pytest tests/extensions/memory/test_advanced_sqlite_session.py` → **37 passed** (includes the regression test added for #3348)

The change is small (+32/-0) and the branch still merges cleanly against `main`. Happy to rebase or make any adjustments you'd like — thanks again for considering this!

kimnamu:
Still relevant and ready for review — rebased context: this is one half of #3498, split per maintainer suggestion into a single-issue PR. The change is minimal (a focused fix + a regression test that passes locally via `uv run pytest`), and the linked issue (#3348) is still open. Happy to rebase or adjust if a maintainer can take a look. Thanks!

kimnamu:
Thanks @scosemicolon — that's a really helpful framing, and I agree with the core of it.

This PR already implements the part you emphasized: instead of wrapping the failure in a broad `except` that logs and returns (the current false-success behavior), it lets the metadata failure escape after rollback via `raise RuntimeError(...) from e`, so callers get a clean retry signal with the original exception chained. The regression test forces `_insert_structure_metadata` to raise and asserts the caller sees the `RuntimeError`.

On the deeper point — treating `_cleanup_orphaned_messages_sync()` as repair tooling rather than the correctness path — you're right. The root cause is that `add_items()` commits the base-message insert **before** attempting the structure insert (`_insert_items(conn); conn.commit()` then `_insert_structure_metadata`), so a metadata failure can't be undone by a single rollback and relies on cleanup. Making the two writes truly commit-or-roll-back as one unit (dropping that intermediate commit) is the proper fix, but it's a larger change to the transaction structure than this PR intends.

To keep this PR minimal and low-risk — it's scoped to the reported #3348 symptom (silent false success) — I've kept the atomicity refactor out of scope. I'm happy to either:
1. extend the test here to assert the transaction-boundary contract directly (both `messages` and `message_structure` unchanged after the failed call), reflecting your point 4, or
2. open a follow-up that removes the intermediate commit so cleanup is no longer load-bearing.

Happy to do whichever the maintainers prefer. Thanks again for the careful review!

seratch:
Resolved by #3349

PR #3537, opened 2026-05-31
Move types-requests to dev dependencies

### LINKED ISSUE #3507: Remove dependency on types-requests

PeterTianbuhan:
### Summary

Move `types-requests` out of the published runtime dependency list and into the dev dependency group. Update `uv.lock` so `openai-agents` no longer advertises `types-requests` in `requires-dist`, while local development still installs it for type-checking.

This addresses #3507: `types-requests` is a stub package and does not affect runtime behavior. Keeping it in runtime dependencies makes regular installs pull an unnecessary package, and recent `requests` releases include inline type hints.

### Test plan

- `uv lock --check`
- `uv tree --no-dev --package openai-agents` confirmed no `types-requests` in the non-dev install tree
- `git diff --check`
- `make sync`

### Issue number

Closes #3507

### Checks

- [ ] I've added new tests (not relevant: dependency metadata only)
- [ ] I've added/updated the relevant documentation (not relevant: dependency metadata only)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,6 @@ dependencies = [
     "griffelib>=2, <3",
     "typing-extensions>=4.12.2, <5",
     "requests>=2.0, <3",
-    "types-requests>=2.0, <3",
     "websockets>=15.0, <17",
     "mcp>=1.19.0, <2; python_version >= '3.10'",
 ]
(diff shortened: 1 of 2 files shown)


seratch:
Resolved by #3509

PR #3541, opened 2026-05-31
fix: coerce ToolCallItem/ToolCallOutputItem call_id to str | None consistently

jluocsa:
### Summary

`ToolCallItem.call_id` and `ToolCallOutputItem.call_id` returned the raw value untouched on the object branch, so a non-string `call_id`/`id` (e.g. an `int`) leaked through despite the `str | None` return annotation. This makes both properties consistently coerce to `str | None`, matching the canonical behavior of `extract_tool_call_id` in `run_internal/tool_execution.py`.

### Changes

- `ToolCallItem.call_id` and `ToolCallOutputItem.call_id` now coerce the resolved value to `str` (or `None` when absent) on both the dict and object branches.

### Test plan

- Added unit tests in `tests/test_items_helpers.py`:
  - dict and object raw items with string `call_id` return the string unchanged.
  - a non-string (`int`) `call_id` is coerced to `"123"`.
  - a missing `call_id` returns `None`.
- `make format`, `make lint`, `make typecheck`, and the full `make tests` suite pass (4615 passed, 11 skipped).

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -373,8 +373,10 @@ class ToolCallItem(RunItemBase[Any]):
     def call_id(self) -> str | None:
         """Return the call identifier from the raw item, if available."""
         if isinstance(self.raw_item, dict):
-            return self.raw_item.get("call_id") or self.raw_item.get("id")
-        return getattr(self.raw_item, "call_id", None) or getattr(self.raw_item, "id", None)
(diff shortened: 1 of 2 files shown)


seratch:
>a non-string call_id/id (e.g. an int) leaked

Thanks for sharing this but this never happens.

PR #3546, opened 2026-05-31
docs: explain task and turn tracing spans

### LINKED ISSUE #3182: Did something changed with traces implementation?
Hey,
Recently I've observed that traces from agent runs looks different than in the previous versions of sdk:

This is what I observe using openai-agents==0.10.5

Image

I can easily see the sub agent runs being part of main run and I can nicely see the hierarchy.

When I run using openai-agents==0.14.2 I see this instead:

Image
There are this new task and turn records and it is not as easy to read as the previous version. Is it a change that happened between this versions?

nima20002000:
### Summary

Document the default `task_span()` and `turn_span()` hierarchy that wraps `Runner` workflows, so users can interpret the newer trace shape discussed in #3182.

This adds guidance for:
- how task, agent, and turn spans nest
- how guardrail span parentage can vary by run path
- how custom trace processors can identify task/turn spans before and after export

### Test plan

- `git diff --check`
- `make build-docs` (passes; emits existing MkDocs/griffe/autoref warnings in generated reference and translated docs)

### Issue number

Closes #3182

### Checks

### DIFF
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -31,7 +31,9 @@ The Agents SDK includes built-in tracing, collecting a comprehensive record of e
 By default, the SDK traces the following:
 
 -   The entire `Runner.{run, run_sync, run_streamed}()` is wrapped in a `trace()`.
+-   Each `Runner` invocation is wrapped in `task_span()`.
 -   Each time an agent runs, it is wrapped in `agent_span()`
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the suggestion. This is not the changes we would like to apply to the docs for now

PR #3553, opened 2026-06-01
Fix fire-and-forget event tasks in RealtimeSession

kratos0718:
### Summary

`RealtimeSession` emitted three error events via a bare `asyncio.create_task(self._put_event(...))` whose result was discarded (in `_on_guardrail_task_done` and `_on_tool_call_task_done`).

`asyncio` only keeps a *weak* reference to a task, so a fire-and-forget task can be garbage-collected before it runs ([documented behavior](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)). When that happens here, the error event it was meant to deliver is silently dropped — and these are exactly the paths that report guardrail/tool-call failures to the caller.

The session already uses a keep-a-reference pattern for `_guardrail_tasks` and `_tool_call_tasks` (a tracking `set` + `add_done_callback(discard)`). This change applies the same pattern to the event-emitting tasks via a small `_emit_event_soon` helper, and cancels any still-pending event tasks during `_cleanup`.

### Test plan

- Added `tests/realtime/test_session_event_tasks.py`:
  - `_emit_event_soon` holds a strong reference while the task is pending, delivers the event to the queue, and releases the reference once done.
  - `_cleanup_event_tasks` cancels pending event tasks.
- `make format`, `make lint` pass (ruff).

### Issue number

N/A — found via static analysis of fire-and-forget `create_task` usage.

### Checks

- [x] I've added/updated the relevant documentation (n/a — internal behavior)

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -192,6 +192,11 @@ class RealtimeSession(RealtimeModelListener):
 
         self._guardrail_tasks: set[asyncio.Task[Any]] = set()
         self._tool_call_tasks: set[asyncio.Task[Any]] = set()
+        # Background tasks that emit events from done-callbacks. asyncio only
+        # holds a weak reference to a task, so we keep a strong reference here
+        # until it completes to avoid the event being dropped if the task is
+        # garbage-collected mid-run.
+        self._event_tasks: set[asyncio.Task[Any]] = set()
         self._async_tool_calls: bool = bool(self._run_config.get("async_tool_calls", True))
 
     @property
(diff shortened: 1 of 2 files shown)


kratos0718:
Gentle bump 🙂 This is still relevant: the three `asyncio.create_task(...)` results in `RealtimeSession` are discarded, so under GC pressure the error events those tasks emit can be dropped silently. The fix keeps a strong reference to each task until it completes (and removes it on done). Happy to rebase onto `main` or address any feedback whenever a maintainer has a chance to take a look — thanks!

seratch:
Thanks for the contribution. The general asyncio guidance about retaining fire-and-forget tasks is valid, but this specific path uses an unbounded asyncio.Queue, so _put_event() does not suspend once the task starts. I was also unable to reproduce the reported event loss on the PR's exact base under forced GC.

I do not think we should add another task-tracking set and cancel-without-await cleanup path here. Please replace these three create_task calls with direct queue.put_nowait(...) calls and test the actual guardrail/tool-call callbacks rather than the new private helper. That would remove the documented task-lifetime concern with substantially less lifecycle complexity.

Let me close this PR now. If you're still interested in working on the issue, please create a new PR for it.

PR #3555, opened 2026-06-01
Add explicit None return annotations to void method

mshsheikh:
## Summary
- This PR adds explicit `-> None` return annotations to methods that do not return a value.
- It improves readability, makes the intent of these methods clearer, and keeps the code consistent with the rest of the file.
- It also helps static type checkers and IDEs understand the code better.
- This change does not affect runtime behavior.

## Why
This improves type clarity and keeps the codebase consistent for static analysis. It does not change runtime behavior.


seratch:
I believe either is fine here

PR #3556, opened 2026-06-01
fix: recursively extract HTTP errors from nested ExceptionGroups

mshsheikh:
This PR improves how the code finds HTTP-related errors inside nested exception groups, specifically when these errors surface during tool execution.

Right now, `_extract_http_error_from_exception()` only checks the first level of `BaseExceptionGroup.exceptions`. That works when the HTTP error is directly inside the group, but it fails when the error is wrapped inside another `ExceptionGroup`.

With this change, the method now checks exception groups recursively. The core behavior this fixes is real: when a nested `ExceptionGroup` reaches `MCPServerStreamableHttp.call_tool()`, the current implementation re-raises the raw group.

This change correctly catches it and maps the nested HTTP error to the intended `UserError`.

*(Note: As discussed in the PR comments, retry classification is already handled recursively elsewhere, and cleanup uses a separate one-level traversal. This fix is specifically targeted at properly surfacing nested UserErrors from the `call_tool` path).*

### Before
```python
def _extract_http_error_from_exception(self, e: BaseException) -> Exception | None:
    """Extract HTTP error from exception or ExceptionGroup."""
    if isinstance(e, httpx.HTTPStatusError | httpx.ConnectError | httpx.TimeoutException):
        return e

    # Check if it's an ExceptionGroup containing HTTP errors
    if isinstance(e, BaseExceptionGroup):
        for exc in e.exceptions:
            if isinstance(
                exc, httpx.HTTPStatusError | httpx.ConnectError | httpx.TimeoutException
            ):
                return exc

    return None
```

### After
```python
def _extract_http_error_from_exception(self, e: BaseException) -> Exception | None:
    """Extract HTTP error from exception or ExceptionGroup recursively."""
    if isinstance(e, httpx.HTTPStatusError | httpx.ConnectError | httpx.TimeoutException):
        return e

    # Recursively check ExceptionGroups for HTTP errors
    if isinstance(e, BaseExceptionGroup):
        for exc in e.exceptions:
            result = self._extract_http_error_from_exception(exc)
            if result is not None:
                return result

    return None
```

## Why this matters
- It correctly maps nested HTTP errors to the intended UserError when a nested ExceptionGroup reaches MCPServerStreamableHttp.call_tool().
- It catches HTTP errors even when they are nested inside multiple exception groups.
- It makes the method more robust in real async failure cases without altering the success path.

## Testing
A regression test has been added to exercise this exact scenario through the call_tool path. The test creates a nested ExceptionGroup containing an httpx.ConnectError, verifying that the method successfully finds it and raises the correct UserError.

**Testing:** A regression test has been added to this PR to exercise this exact scenario through the `call_tool` path.

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -712,13 +712,12 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
         if isinstance(e, httpx.HTTPStatusError | httpx.ConnectError | httpx.TimeoutException):
             return e
 
-        # Check if it's an ExceptionGroup containing HTTP errors
+        # Recursively check ExceptionGroups for HTTP errors
         if isinstance(e, BaseExceptionGroup):
             for exc in e.exceptions:
-                if isinstance(
-                    exc, httpx.HTTPStatusError | httpx.ConnectError | httpx.TimeoutException
-                ):
-                    return exc
+                result = self._extract_http_error_from_exception(exc)
+                if result is not None:
+                    return result
 
         return None
 
--- a/tests/mcp/test_server_errors.py
+++ b/tests/mcp/test_server_errors.py
@@ -1,19 +1,20 @@
+import builtins
 import sys
 from unittest.mock import MagicMock, patch
 
 import httpx
 import pytest
 
-# Handle Python version compatibility for ExceptionGroups
-if sys.version_info < (3, 11):
-    from exceptiongroup import BaseExceptionGroup
-else:
-    BaseExceptionGroup = ExceptionGroup
-
 from agents import Agent
 from agents.exceptions import UserError
 from agents.mcp.server import MCPServerStreamableHttp, _MCPServerWithClientSession
 from agents.run_context import RunContextWrapper
+
+# Handle Python version compatibility for ExceptionGroups
+if sys.version_info < (3, 11):
+    from exceptiongroup import BaseExceptionGroup
+else:
+    BaseExceptionGroup = getattr(builtins, "BaseExceptionGroup")
 
 
 class CrashingClientSessionServer(_MCPServerWithClientSession):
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
The core behavior is real: when a nested ExceptionGroup reaches MCPServerStreamableHttp.call_tool(), the current implementation re-raises the group, while this change correctly maps the nested HTTP error to the intended UserError.

Please add a regression test that exercises this through the call_tool path and fails on the base revision. Please also narrow the PR description: retry classification is already recursive through `_should_retry_in_isolated_session()`, and cleanup uses a separate one-level traversal that this patch does not change. The implementation itself is small and directionally sound once the supported behavior is covered by a test.

mshsheikh:
> The core behavior is real: when a nested ExceptionGroup reaches MCPServerStreamableHttp.call_tool(), the current implementation re-raises the group, while this change correctly maps the nested HTTP error to the intended UserError.
> 
> Please add a regression test that exercises this through the call_tool path and fails on the base revision. Please also narrow the PR description: retry classification is already recursive through `_should_retry_in_isolated_session()`, and cleanup uses a separate one-level traversal that this patch does not change. The implementation itself is small and directionally sound once the supported behavior is covered by a test.

Hi @seratch, thanks for the review and guidance! I have updated the PR description to narrow the scope strictly to the `call_tool()` path, removing the mentions of retry and cleanup traversals.
I have also added the requested regression test (referenced in PR #3675) which verifies that nested ExceptionGroups correctly map to a UserError.

Let me know if you need anything else!

mshsheikh:
Hi @seratch, I moved the regression test over to this PR as requested so everything is consolidated here.
Let me know if it looks good to go!

seratch:
Can you apply Ruff formatter? `make format` should work for you

mshsheikh:
> Can you apply Ruff formatter? `make format` should work for you

I've formatted the code to comply with Ruff and Mypy type-check errors, @seratch.
Let me know if this looks good, or required another update.

seratch (approved):

PR #3566, opened 2026-06-02
Bind restored handoffs to target agent identity

White-Mouse:
## Summary

- include the target agent identity when serializing pending handoff actions
- restore handoff actions by matching both the generated handoff tool name and the target agent identity
- preserve the existing tool-name fallback for older snapshots without target identity
- add regression coverage for duplicate-name handoff targets that share the same generated tool name

## Why

`RunState` handoff serialization currently records the generated handoff tool name for pending handoff actions. When a current agent has multiple same-name handoff targets, those targets can share the same generated handoff tool name. Restoring from a snapshot can then select a different target handoff than the one that produced the original pending action.

Binding the serialized action to the target agent identity keeps restored handoff actions attached to the same target while retaining compatibility with legacy snapshots.

## Tests

- `uv run ruff check src/agents/run_state.py tests/test_run_state.py`
- `uv run pytest -q tests/test_run_state.py::TestDeserializeHelpers::test_last_processed_handoff_restores_duplicate_target_by_identity tests/test_run_state.py::TestDeserializeHelpers::test_serialization_uses_duplicate_identities_for_handoff_and_output_guardrails tests/test_run_state.py::TestDeserializeHelpers::test_model_response_serialization_roundtrip tests/test_run_state.py::TestRunState::test_from_json_restores_duplicate_name_current_agent_by_identity tests/test_run_state.py::TestRunStateSerializationEdgeCases::test_deserialize_processed_response_handoff_with_tool_name`
- `uv run pytest -q tests/test_run_state.py`
- `git diff --check`

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -790,7 +790,10 @@ class RunState(Generic[TContext, TAgent]):
             A dictionary representation of the ProcessedResponse.
         """
 
-        action_groups = _serialize_tool_action_groups(processed_response)
+        action_groups = _serialize_tool_action_groups(
+            processed_response,
+            agent_identity_keys_by_id=agent_identity_keys_by_id,
+        )
         _serialize_pending_nested_agent_tool_runs(
             parent_state=self,
             function_entries=action_groups.get("functions", []),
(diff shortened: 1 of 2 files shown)


White-Mouse:
This should be ready from my side. I noticed the PR currently has no reported checks on the fork branch, so I’ll wait for CI/review and adjust if anything comes up.

PR #3571, opened 2026-06-03
feat: expose output_schema in ModelInputData for call_model_input_filter

### LINKED ISSUE #3563: Add custom `ResponseFormat` support in CallModelInputFilter callback
`call_model_input_filter` of RunConfig  is invoked immediately before calling the model which allows you to edit the input sent to the model, Is there any change to add `response_format: ResponseFormat | None` field to support modify response format before calling the model.

jordanchendev:
### Summary

Add an optional `output_schema` field to `ModelInputData` so that `call_model_input_filter` callbacks can inspect and override the structured-output schema used for a given model call.

**Before:** `ModelInputData` only exposed `input` and `instructions`, making it impossible for a filter to change (or even read) the response format without reaching outside the callback.

**After:** Filters receive the current `output_schema` (derived from `agent.output_type`) in `model_data.output_schema` and may return a different `AgentOutputSchemaBase` instance to override it for that call. Returning `None` or omitting the field preserves the agent's schema unchanged, so existing filters are fully backward-compatible.

Both the streaming (`run_single_turn_streamed`) and non-streaming (`get_new_response`) paths honour the override. `ToolOutputTrimmer` is updated to forward `output_schema` so it is not silently dropped by that built-in filter.

### Test plan

Four new tests in `tests/test_call_model_input_filter.py`:

- `test_filter_can_override_output_schema_non_streamed` — filter replaces schema; model receives the override.
- `test_filter_can_override_output_schema_streamed` — same for the streaming path.
- `test_filter_receives_agent_output_schema` — filter can observe the agent's schema before deciding.
- `test_filter_not_setting_output_schema_preserves_agent_schema` — omitting `output_schema` in the return value must not clear the agent's schema (backward-compat regression test).

All existing tests continue to pass.

```
uv run pytest tests/test_call_model_input_filter.py tests/extensions/test_tool_output_trimmer.py -q
# 47 passed in 0.19s
```

### Issue number

Closes #3563

### Checks

### DIFF
--- a/docs/running_agents.md
+++ b/docs/running_agents.md
@@ -406,7 +406,7 @@ settings so the resumed turn continues in the same server-managed conversation.
 
 Use `call_model_input_filter` to edit the model input right before the model call. The hook receives the current agent, context, and the combined input items (including session history when present) and returns a new `ModelInputData`.
 
-The return value must be a [`ModelInputData`][agents.run.ModelInputData] object. Its `input` field is required and must be a list of input items. Returning any other shape raises a `UserError`.
+The return value must be a [`ModelInputData`][agents.run.ModelInputData] object. Its `input` field is required and must be a list of input items. Returning any other shape raises a `UserError`. You may also set `output_schema` on the returned object to replace the response format for that model call — the agent's own `output_type` is used when `output_schema` is `None` or omitted.
 
 ```python
 from agents import Agent, Runner, RunConfig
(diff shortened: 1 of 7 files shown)


jordanchendev (commented):
Good catch — fixed in e7e0bbb2.

`get_new_response` now returns `tuple[ModelResponse, AgentOutputSchemaBase | None]` so the caller (`run_single_turn`) unpacks and forwards the effective schema to `get_single_step_result_from_response`. The existing non-streamed test was strengthened to assert `result.final_output` is a parsed `_Reply` instance (not a raw string), making the regression impossible to re-introduce silently.

PR #3579, opened 2026-06-04
fix: export ToolContext, MCPToolChoice, and RequestUsage from agents

jawwad-ali:
### Summary

`ToolContext`, `MCPToolChoice`, and `RequestUsage` are user-facing types that currently require submodule imports. `ToolContext` is the documented context type for function tools (`docs/context.md` — even the repo's own examples import it from `agents.tool_context`), `MCPToolChoice` is the `ModelSettings.tool_choice` value for forcing a specific hosted MCP tool, and `RequestUsage` is the element type of `Usage.request_usage_entries`. This re-exports them from the top-level `agents` package, following the same pattern as #3489 and #3490. Existing submodule import paths keep working, so no docs changes are needed.

### Test plan

- Added `tests/test_public_type_exports.py`: `test_tool_context_is_exported_at_top_level`, `test_mcp_tool_choice_is_exported_at_top_level`, `test_request_usage_is_exported_at_top_level`, and `test_exported_names_resolve_to_attributes` (asserts every name in `agents.__all__` resolves via `getattr`).
- `make format`, `make lint`, and pyright pass; mypy reports only pre-existing errors also present on clean `main`.
- Full test suite run locally; the new tests pass and the only failures are pre-existing on clean `main` (local environment-related, unrelated to this change).

### Issue number

None — follow-up to the export gaps addressed in #3489/#3490.

### Checks

### DIFF
--- a/src/agents/__init__.py
+++ b/src/agents/__init__.py
@@ -81,7 +81,7 @@ from .memory import (
     SessionSettings,
     is_openai_responses_compaction_aware_session,
 )
-from .model_settings import ModelSettings
+from .model_settings import MCPToolChoice, ModelSettings
 from .models.interface import Model, ModelProvider, ModelTracing
 from .models.multi_provider import MultiProvider
 from .models.openai_agent_registration import OpenAIAgentRegistrationConfig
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the suggestion, but we don't intend to export everything from the top level module.

PR #3581, opened 2026-06-05
fix: expose sandbox error retryability

qiyaoq-oai:
Context: https://github.com/temporalio/sdk-python/issues/1548

This pull request adds provider-neutral retryability metadata for sandbox failures.

- Adds `SandboxError.retryable` with `True`, `False`, and `None` for unknown classification.
- Propagates retryability through wrapped sandbox errors.
- Emits retryability in sandbox finish events and trace error payloads.
- Classifies clear terminal/transient cases across sandbox providers without exposing provider internals to integrations.
- Adds focused tests for core error behavior, provider classifiers, and observability.

### DIFF
--- a/src/agents/extensions/sandbox/blaxel/sandbox.py
+++ b/src/agents/extensions/sandbox/blaxel/sandbox.py
@@ -63,6 +63,7 @@ from ....sandbox.util.retry import (
     TRANSIENT_HTTP_STATUS_CODES,
     exception_chain_contains_type,
     exception_chain_has_status_code,
+    iter_exception_chain,
     retry_async,
 )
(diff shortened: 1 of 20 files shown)


qiyaoq-oai:
@codex review

alfozan (approved):

seratch (approved):
One non-blocking integration note for the Temporal follow-up: it would be safer to treat only `retryable is False` as an explicit non-retryable signal, while preserving the orchestrator's existing policy for `True` and `None`. In particular, `True` should not be interpreted as a blanket guarantee that replaying an arbitrary `exec` is side-effect safe.

PR #3586, opened 2026-06-05
feat: add response_format to CallModelInputFilter callback

### LINKED ISSUE #3563: Add custom `ResponseFormat` support in CallModelInputFilter callback

Oxygen56:
Allows CallModelInputFilter to modify response_format, enabling dynamic structured output control per model call.

Closes #3563

### DIFF
--- a/src/agents/agent_output.py
+++ b/src/agents/agent_output.py
@@ -168,6 +168,60 @@ class AgentOutputSchema(AgentOutputSchemaBase):
         return _type_to_str(self.output_type)
 
 
+class _ResponseFormatOutputSchema(AgentOutputSchemaBase):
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for the contribution. Issue #3563 describes a valid capability gap, but we should not continue with this implementation. This PR duplicates #3571, and the raw response_format wrapper is not provider-neutral: Responses-style JSON Schema and JSON mode inputs are converted into an empty JSON Schema, while validate_json only parses JSON without enforcing the supplied schema.

PR #3587, opened 2026-06-05
feat: add retry mechanism for ModelBehaviorError

### LINKED ISSUE #325: Retry mechanism for ModelBehaviorError

Oxygen56:
## Summary

Adds automatic retry for `ModelBehaviorError` with configurable `max_model_retries` on `RunConfig`.

When a model produces a malformed response (invalid tool call JSON, nonexistent tool, etc.), the SDK raises `ModelBehaviorError`. Previously there was no built-in retry — the agent just failed. This PR adds an automatic retry mechanism that feeds the error message back to the model so it can self-correct.

## Changes

- **`RunConfig.max_model_retries: int = 0`** — new parameter controlling how many automatic retries are attempted when `ModelBehaviorError` occurs (defaults to 0, preserving existing behavior)
- **`run_single_turn` (non-streaming)** — wraps model call + response processing in a retry loop; on error, appends a synthetic user message with the error details and retries the model call
- **`run_single_turn_streamed` (streaming)** — same retry logic for the streaming path; events from failed attempts are emitted (consistent with existing network-level retry behavior)

## How it works

1. The model call completes successfully but produces a malformed response
2. `get_single_step_result_from_response` raises `ModelBehaviorError`
3. If retries remain, the error message is appended as a synthetic user message: `"Your previous response was invalid: <error message>"`
4. The model is called again with the error feedback
5. This repeats until success or retries are exhausted

## Usage

```python
from agents import Agent, Runner, RunConfig

config = RunConfig(max_model_retries=3)
result = await Runner.run(agent, "...", run_config=config)
```

Closes #325

### DIFF
--- a/src/agents/run_config.py
+++ b/src/agents/run_config.py
@@ -329,6 +329,14 @@ class RunConfig:
       the run continue.
     """
 
+    max_model_retries: int = 0
+    """Maximum number of automatic retries when the model produces a malformed response that
+    triggers a ``ModelBehaviorError`` (e.g. invalid tool call JSON, nonexistent tool, etc.).
+
+    On each retry, the error message is fed back to the model as a synthetic user message so
+    the model can self-correct. Defaults to 0 (no retry).
+    """
+
 
 class RunOptions(TypedDict, Generic[TContext]):
     """Arguments for ``AgentRunner`` methods."""
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -1345,8 +1345,6 @@ async def run_single_turn_streamed(
     model_settings = get_model_settings(execution_agent, run_config)
     model_settings = maybe_reset_tool_choice(public_agent, tool_use_tracker, model_settings)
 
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for working on this. The specific case reported in #325 is already covered by `RunConfig(tool_not_found_behavior="return_error_to_model")`, which was merged in #3461 and released in v0.17.4. That implementation returns a structured tool error to the model and supports both non-streaming and streaming runs.

The broader retry boundary in this PR is not safe: it can retry after other tools have already executed, duplicating side effects, and streamed callers can observe events from an attempt that is later discarded. The PR also does not include regression tests for these cases.

Given the existing narrower solution and the correctness risks of retrying every `ModelBehaviorError`, I am going to close this as a duplicate rather than request a rework.

PR #3589, opened 2026-06-05
feat: add on_turn_end callback for inter-turn state changes

### LINKED ISSUE #2671: Feature request: better support for agent state changes between turns

Oxygen56:
Adds `on_turn_end` callback for inspecting/modifying agent state between turns.

### `RunHooksBase` and `AgentHooksBase` (lifecycle.py)
- Added `on_turn_end` lifecycle hook method that fires after each turn completes

### `RunConfig` (run_config.py)
- Added `TurnEndData` dataclass carrying the current agent, run context, and turn number
- Added `OnTurnEndCallback` type alias for the callback signature
- Added `on_turn_end` field to `RunConfig` as an optional callback

### Turn loop wiring (run.py, run_internal/run_loop.py)
- Invokes `on_turn_end` at all turn boundaries in both the standard and streaming execution paths:
  - `NextStepHandoff` (after handoff, before next agent starts)
  - `NextStepRunAgain` (after tool execution, before next model call)

## Use cases
- Logging and state tracking between turns
- Dynamic instruction updates based on completed turn context
- Context compaction in long-running workflows
- Intercepting new external input between turns

Closes #2671

### DIFF
--- a/src/agents/__init__.py
+++ b/src/agents/__init__.py
@@ -108,6 +108,7 @@ from .retry import (
     retry_policies,
 )
 from .run import (
+    OnTurnEndCallback,
     ReasoningItemIdPolicy,
     RunConfig,
     Runner,
@@ -115,6 +116,7 @@ from .run import (
     ToolErrorFormatterArgs,
     ToolExecutionConfig,
     ToolNotFoundBehavior,
+    TurnEndData,
 )
 from .run_context import AgentHookContext, RunContextWrapper, TContext
 from .run_error_handlers import (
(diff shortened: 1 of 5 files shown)


seratch:
Thanks for working on this. I do not think we should continue with this approach. The released `RunConfig.call_model_input_filter` already runs before every model call and can modify the complete input, covering both history compaction and injecting queued external messages between tool turns.

This callback only receives the agent and context, so it does not itself modify the next model input. The current implementation also shifts existing `RunConfig` positional arguments, invokes the target agent's turn-end hook after a handoff, and has no regression tests for the new behavior.

I am going to close this PR.

PR #3590, opened 2026-06-05
fix: handle reasoning items in CodeInterpreter response processing

### LINKED ISSUE #985: Error "Item ‘rs_ABCD’ of type ‘reasoning’ was provided without its required..." when using CodeInterpreter

Oxygen56:
Fixes CodeInterpreter crash with reasoning models (o3/o4-mini).

## Problem
When using reasoning models (o3, o4-mini) with the CodeInterpreter tool, the SDK fails with:
```
Item 'rs_ABCD' of type 'reasoning' was provided without its required following item.
```

This happens because reasoning items emitted by the model are replayed as input (e.g. during agent handoffs), but their IDs reference following items that may not be recognized as valid reasoning followers by the API.

## Root Cause
Models like o3/o4-mini emit reasoning items that precede hosted tool calls (code_interpreter, web_search, file_search, etc.). When these items are sent back as input, the API may reject reasoning items whose IDs point to a hosted tool call that is not treated as a valid "following item." The issue is particularly triggered during handoffs between agents.

## Fix
Added `_sanitize_reasoning_item_followers()` which strips IDs from reasoning items that:
1. Precede hosted tool calls (code_interpreter_call, web_search_call, file_search_call, image_generation_call, computer_call)
2. Are the last item in the input with no follower at all

Reasoning items preceding regular messages or function_calls are left untouched (IDs preserved).

The fix is applied in `_build_response_create_kwargs` which covers all code paths (streaming, non-streaming, regular, and websocket models).

Fixes #985

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -735,6 +735,7 @@ class OpenAIResponsesModel(Model):
         list_input = ItemHelpers.input_to_new_input_list(input)
         list_input = _to_dump_compatible(list_input)
         list_input = self._remove_openai_responses_api_incompatible_fields(list_input)
+        list_input = self._sanitize_reasoning_item_followers(list_input)
 
         if model_settings.parallel_tool_calls and tools:
             parallel_tool_calls: bool | Omit = True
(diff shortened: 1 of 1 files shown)


seratch:
The issue cannot be resolved sorely on the client SDK side

PR #3598, opened 2026-06-08
docs(tracing): fix capitalization, and env var phrasing

mshsheikh:
- Correct "Openai" to "OpenAI", remove informal "free" phrasing, and added the dashboard link in the additional notes
- Replace "exporting" with "setting" when referring to environment variables for better cross-platform accuracy


seratch:
For this one, I think either is fine

PR #3608, opened 2026-06-09
docs: fix inline code formatting for `run_demo_loop`

mshsheikh:
Added backticks around `run_demo_loop` in the explanation paragraph.


seratch:
Thanks but we don't need this change

PR #3618, opened 2026-06-10
fix: wake E2B PTY collection on process exit

### LINKED ISSUE #3609: E2B PTY output collection can wait until yield timeout after process exit
### Bug description

E2B PTY output collection appears to wake only when stdout/stderr/PTY output arrives. If a process exits after its final output chunk, or exits without output, `pty_exec_start` / `pty_write_stdin` can wait until the full `yield_time_s` window before observing completion.

This adds up to avoidable latency for E2B-backed tool calls, especially across multi-turn episodes.

### Expected behavior

When the underlying E2B process/session exits, output collection should wake promptly and return the observed exit code instead of waiting for another output chunk or the yield timeout.

### Why this happens

Both E2B execution paths return an `AsyncCommandHandle`:

- `tty=false`: `sandbox.commands.run(..., background=True, ...)`
- `tty=true`: `sandbox.pty.create(...)`

Output and process completion are separate signals. The collector checks for an exit code, but process exit does not currently wake the collector when no further output arrives.

### Suggested fix

Treat E2B process/session exit as an output-collection wakeup signal, similar to the existing Daytona sandbox implementation.

A small waiter task can await `AsyncCommandHandle.wait()`, cache the exit code, and set the existing output notification event so the collector wakes on exit as well as output.

### Additional context

This should not require a public API change. The behavior change is internal to the E2B sandbox adapter.

he-yufeng:
## Summary

Fixes #3609.

E2B PTY output collection already checks the process exit code while draining output, but process exit itself did not wake the collector. If a background command or PTY session exited after its final chunk, or exited without producing output, `pty_exec_start()` / `pty_write_stdin()` could wait until the full yield window before noticing the exit.

This adds a small waiter task for E2B command handles that expose `wait()`. When the handle exits, the waiter captures the exit code when available, marks the PTY entry done, and sets the existing output notification event. The collector can then return promptly on process exit even when no more stdout/stderr data arrives.

## Tests

To verify:

```console
python -m py_compile src\agents\extensions\sandbox\e2b\sandbox.py tests\extensions\sandbox\test_e2b.py
PYTHONPATH=src python -m pytest tests\extensions\sandbox\test_e2b.py -q
python -m ruff check src\agents\extensions\sandbox\e2b\sandbox.py tests\extensions\sandbox\test_e2b.py
git diff --check
```

### DIFF
--- a/src/agents/extensions/sandbox/e2b/sandbox.py
+++ b/src/agents/extensions/sandbox/e2b/sandbox.py
@@ -686,6 +686,8 @@ class _E2BPtyProcessEntry:
     output_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
     output_notify: asyncio.Event = field(default_factory=asyncio.Event)
     last_used: float = field(default_factory=time.monotonic)
+    done: bool = False
+    exit_code: int | None = None
 
 
 @dataclass(frozen=True)
@@ -982,6 +984,7 @@ class E2BSandboxSession(BaseSandboxSession):
                     on_data=_append_output,
                 )
                 entry.handle = handle
+                asyncio.create_task(self._run_pty_waiter(entry))
                 await self._sandbox.pty.send_stdin(
                     cast(Any, handle).pid,
                     f"{command_text}\n".encode(),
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this as a duplicate of #3610. My earlier issue-number search missed that open PR because it does not reference #3609 directly, but the implementation scope overlaps with the same E2B PTY exit-wakeup path.

Issue #2798, opened 2026-03-27
hydrate_from_state stores id() of ephemeral objects in sent_items, causing false-positive dedup after GC


pietrog30:
### Please read this first

- **Have you read the docs?** Yes — [Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Yes — no existing issues cover this.

### Describe the bug

`OpenAIServerConversationTracker.hydrate_from_state` stores `id()` of temporary dict objects in `self.sent_items`. When `original_input` is a string, `ItemHelpers.input_to_new_input_list` wraps it in a temporary `dict`. The `id()` of that dict is added to `sent_items`, but no strong reference is kept, so the dict is immediately eligible for garbage collection.

When CPython reuses the same memory address for a later allocation (e.g. a `function_call_output` dict created during HITL rejection), that new object's `id()` collides with the stale entry in `sent_items`. `prepare_input` then considers it "already sent" and drops it from the API payload, producing:

```
400 - "No tool output found for function call <call_id>."
```

The bug is in `oai_conversation.py`, inside `hydrate_from_state`:

```python
for item in ItemHelpers.input_to_new_input_list(normalized_input):
    if item is None:
        continue
    self.sent_items.add(id(item))  # ← id() of ephemeral dict; no reference kept
```

`sent_initial_input` / `remaining_initial_input` already control whether the original user input is replayed, so these transient IDs don't appear necessary for correctness. The cleanest fix would be to not add them to `sent_items` in the first place.

### Debug information
- Agents SDK version: `v0.13.1`
- Python version: Python 3.12.8

### Repro steps

Self-contained script — no API calls or network required, runs in < 1 second:

```python
"""
Deterministic repro: OpenAIServerConversationTracker.hydrate_from_state stores
id() of ephemeral objects in sent_items, causing false-positive dedup after GC.

No API calls or network required. Runs in < 1 second.

Bug: When original_input is a string, hydrate_from_state converts it via
ItemHelpers.input_to_new_input_list into a temporary dict, stores id(dict) in
sent_items, then lets the dict be garbage collected. Any later dict allocated at
the same address (e.g. the rejection function_call_output) is wrongly considered
"already sent" and dropped from the API payload.

SDK version: openai-agents 0.13.1
agents/run_internal/oai_conversation.py, hydrate_from_state
"""

import gc
import sys

from openai.types.responses import ResponseFunctionToolCall, ResponseReasoningItem
[...]

PR #3621, opened 2026-06-11
fix: avoid stale object-id dedupe in server conversation tracker

### LINKED ISSUE #3620: OpenAIServerConversationTracker can drop a fresh tool output after id() reuse
# `OpenAIServerConversationTracker` can drop a fresh tool output after `id()` reuse

### Describe the bug

`OpenAIServerConversationTracker.prepare_input` dedupes generated items with long-lived
sets of Python object ids:

```python
raw_item_id = id(raw_item)
if raw_item_id in self.sent_items or raw_item_id in self.server_items:
    continue
```

This is unsafe once the original object can be garbage-collected. In CPython, `id(obj)` is the
object's address for the object's lifetime, and that address can be reused by a later allocation.
If the tracker keeps the old integer after the object is gone, a new `function_call_output` can be
mistaken for an already-sent item and omitted from the next request.

The concrete live source I verified is `sent_items`: `mark_input_as_sent()` records object ids for
delivered inputs, and the original object may no longer be retained after the first input has been
sent and `remaining_initial_input` is cleared. The same address-based check also consults
`server_items`, which is populated with `id(output_item)` in `track_server_items`.

When the fresh output is dropped, the server-managed continuation still has the corresponding
function call, but the request does not include its output. The provider then rejects the request
with:

```text
Error code: 400 - No tool output found for function call <call_id>
```

This affects live, non-resumed runs that use server-managed continuation
(`previous_response_id`, `conversation_id`, or `auto_previous_response_id`). The other dedupe layers
do not cover this case:

- `server_item_ids` requires a provider-assigned item id; client-built tool outputs do not have one.
- `server_tool_call_ids` only covers tool outputs already acknowledged by the server or restored
  from state; a freshly produced live output is not in it.
- The content-fingerprint guard is gated by `primed_from_state`, so it does not run on ordinary live
  turns.
- `drop_orphan_function_calls` drops orphan calls, not outputs.

### Relationship to #2798 / #2800

#2800 fixed the same root-cause class, but only for hydrated initial input during resume. It did
not change the live `prepare_input` identity check, `mark_input_as_sent`, or `track_server_items`.

### Minimal deterministic repro

The natural allocator collision is timing-dependent, but the effect is deterministic once a stale
id matches a new output. This reproduces the drop on current `main` by seeding that precondition:

```python
from typing import Any

from agents.run_internal.oai_conversation import OpenAIServerConversationTracker


class _Item:
    def __init__(self, raw_item: dict[str, Any], type: str) -> None:
        self.raw_item = raw_item
        self.type = type


tracker = OpenAIServerConversationTracker(previous_response_id="resp_1")

output = {"type": "function_call_output", "call_id": "call_FRESH", "output": "42"}
tracker.sent_items.add(id(output))  # stale id collision precondition on current main

prepared = tracker.prepare_input([], [_Item(output, "function_call_output_item")])
print(prepared)  # []
```

### Expected behavior

A newly produced `function_call_output` should not be filtered out only because its object address
matches an old, no-longer-live object.

### Proposed direction

Do not keep raw `id()` integers as long-lived dedupe state. If in-process object identity is needed,
track the object references themselves and compare with `is`, so the identity entry cannot outlive
the object and become a stale address key. Keep the stable provider-id, tool-call-id, and
fingerprint-based dedupe layers for the cases they already cover.

chutch:
# Summary

`OpenAIServerConversationTracker` kept long-lived `set[int]` state for object ids in
`sent_items` and `server_items`. `prepare_input` used those ids to skip generated items:

```python
raw_item_id = id(raw_item)
if raw_item_id in self.sent_items or raw_item_id in self.server_items:
    continue
```

That is unsafe after the original object is gone, because CPython can reuse the same address for a
later allocation. A fresh `function_call_output` can then be treated as already sent and omitted
from the next server-managed continuation request, which leaves the provider with a function call
but no output.

This changes the in-process identity tracking to keep object references instead of raw `id()`
integers:

- `sent_items` and `server_items` now store the tracked objects.
- Identity membership uses `is`, not equality.
- Keeping the reference alive prevents the tracked identity from becoming a stale address.
- `rewind_input` removes the exact tracked object when an input is rewound.

The existing stable dedupe layers are unchanged: provider item ids, server tool-call ids, and
fingerprints still handle their existing retry/resume cases.

This is related to #2798 / #2800, but covers a different live path. #2800 removed one hydrated
initial-input identity seeding path during resume; it did not change the live `prepare_input`
identity check, `mark_input_as_sent`, or `track_server_items`.

# Tradeoffs

The tracker now retains the tracked objects for the tracker's lifetime, rather than retaining only
integer ids. That is intentional for correctness, but it can keep large tool-output objects alive
longer in very long runs. Identity membership is also a linear scan instead of a set lookup; the
tracked collections are expected to be small in normal runs.

# Test plan

- Added `test_prepare_input_keeps_fresh_tool_output_when_stale_identity_matches`, parametrized over
  `sent_items` and `server_items`. The test fails on current `origin/main` with
  `AssertionError: assert 'call_FRESH' in []` and passes with this change.
- Added `test_prepare_input_dedupes_same_delivered_tool_output_object` to preserve the existing
  same-object dedupe behavior after delivery.
- `PYTHONPATH=/private/tmp/oai-agents-stable-keys/src python -m pytest tests/test_server_conversation_tracker.py`
  - 23 passed
- `PYTHONPATH=/private/tmp/oai-agents-stable-keys/src python -m pytest tests/test_oaiconv_resume_response_id.py tests/test_agent_runner.py::test_conversation_id_only_sends_new_items_multi_turn tests/test_agent_runner.py::test_conversation_id_only_sends_new_items_multi_turn_streamed`
  - 3 passed
- `PYTHONPATH=/private/tmp/oai-agents-stable-keys/src python -m pytest tests/test_agent_runner.py`
  - 150 passed
- `python -m ruff check .`
  - passed
- `python -m ruff format --check .`
  - passed
- `PYTHONPATH=/private/tmp/oai-agents-stable-keys/src python -m mypy src/agents/run_internal/oai_conversation.py tests/test_server_conversation_tracker.py`
  - passed
- `pyright --project pyrightconfig.json src/agents/run_internal/oai_conversation.py tests/test_server_conversation_tracker.py`
  - 0 errors

Closes #3620

### DIFF
--- a/src/agents/run_internal/oai_conversation.py
+++ b/src/agents/run_internal/oai_conversation.py
@@ -95,6 +95,25 @@ def _has_output_payload(item: Any) -> bool:
     return (isinstance(item, dict) and "output" in item) or hasattr(item, "output")
 
 
+def _is_tracked_object(items: Sequence[Any], candidate: Any) -> bool:
+    """Return True when the exact object instance is already tracked."""
+    return any(item is candidate for item in items)
+
+
+def _track_object_once(items: list[Any], candidate: Any) -> None:
+    """Track an object instance once, keeping it alive while identity dedupe is needed."""
+    if not _is_tracked_object(items, candidate):
+        items.append(candidate)
+
+
+def _untrack_object(items: list[Any], candidate: Any) -> None:
+    """Remove an object instance from an identity-tracking list."""
+    for index, item in enumerate(items):
+        if item is candidate:
+            items.pop(index)
+            return
+
+
 @dataclass
 class OpenAIServerConversationTracker:
     """Track server-side conversation state for conversation-aware runs.
@@ -113,9 +132,10 @@ class OpenAIServerConversationTracker:
     previous_response_id: str | None = None
     auto_previous_response_id: bool = False
 
-    # In-process object identity for items that have already been delivered or acknowledged.
-    sent_items: set[int] = field(default_factory=set)
-    server_items: set[int] = field(default_factory=set)
+    # In-process object identity for delivered or acknowledged items. Keep object references
+    # instead of id(obj) integers so a later allocation cannot reuse a stale address.
+    sent_items: list[Any] = field(default_factory=list)
+    server_items: list[Any] = field(default_factory=list)
 
     # Stable provider identifiers returned by the Responses API.
     server_item_ids: set[str] = field(default_factory=set)
@@ -200,7 +220,7 @@ class OpenAIServerConversationTracker:
             for output_item in response.output:
                 if output_item is None:
                     continue
-                self.server_items.add(id(output_item))
+                _track_object_once(self.server_items, output_item)
                 item_id = _normalize_server_item_id(
                     output_item.get("id")
                     if isinstance(output_item, dict)
@@ -263,8 +283,7 @@ class OpenAIServerConversationTracker:
                 if not should_mark:
                     continue
 
-                raw_item_id = id(raw_item)
-                self.sent_items.add(raw_item_id)
+                _track_object_once(self.sent_items, raw_item)
                 fp = _fingerprint_for_tracker(raw_item)
                 if fp:
                     self.sent_item_fingerprints.add(fp)
@@ -297,7 +316,7 @@ class OpenAIServerConversationTracker:
                 if not should_mark:
                     continue
 
-                self.sent_items.add(id(raw_item))
+                _track_object_once(self.sent_items, raw_item)
                 fp = _fingerprint_for_tracker(raw_item)
                 if fp:
                     self.sent_item_fingerprints.add(fp)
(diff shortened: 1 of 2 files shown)


chutch:
Design note: a pure `call_id`-based key would be simpler and would avoid the object-retention tradeoff in this PR. I did not use it because I could not find a public guarantee that `call_id` is unique across the relevant conversation/session scope, and the existing repo tests include multi-turn fixtures that reuse `call_id` values. If the API contract can guarantee uniqueness for the scope this tracker needs, a `call_id`-based implementation would be worth considering.

chutch:
@seratch Do you think we can rely on the uniqueness of `call_id` and simplify the code, or it can't be guaranteed?

seratch (approved):
Thanks for fixing this issue. I agree the fix here could resolve the potential issue with `id()` comparison and confirmed the change works well.

PR #3622, opened 2026-06-11
fix(mounts): honor pattern options in `mountpoint` and `s3files`

mshsheikh:
- Wire `MountpointMountPattern.options` and `S3FilesMountPattern.options` into command assembly so pattern-level settings are actually applied at runtime.

- This restores the intended configuration surface for mount-specific overrides while preserving existing runtime config behavior as the fallback path.

- No behavior changes for callers that do not use pattern options.

### DIFF
--- a/src/agents/sandbox/entries/mounts/patterns.py
+++ b/src/agents/sandbox/entries/mounts/patterns.py
@@ -450,9 +450,9 @@ class MountpointMountPattern(MountPatternBase):
         elif mountpoint_config.mount_type in {"s3_mount", "gcs_mount"}:
             cmd.extend(["--allow-overwrite", "--allow-delete"])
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the contribution. I traced the supported `InContainerMountAdapter` path for `S3Mount`, `GCSMount`, and `S3FilesMount`. Each provider's `build_in_container_mount_config()` already resolves pattern options into the runtime config, including S3 Files `extra_options`, before `pattern.apply()` is called. A base/head comparison therefore produces identical commands.

This patch only changes behavior when internal config dataclasses are manually constructed and `apply()` is called directly. Because that is not the supported mount path, and the patch duplicates precedence logic without a regression test, I am going to close this PR. If there is a supported path that bypasses the provider config builders, please provide a minimal reproduction and we can revisit it.

PR #3624, opened 2026-06-11
fix(mounts): resolve rclone remote names consistently

mshsheikh:
- Use `RcloneMountPattern.resolve_remote_name()` when building the per-session config path and when launching/unmounting rclone processes.

- This keeps `apply()` and `unapply()` aligned on the same remote name, preserves deterministic naming when no explicit remote_name is provided, and removes the current dead wiring around pattern-level name resolution.

- The change is intentionally limited to name resolution and does not alter mount semantics.


seratch:
Thanks for the contribution. I traced the rclone mount lifecycle on current main. Provider-backed mounts already resolve the remote name in `_ConfiguredMount._build_rclone_config()` and store it in `RcloneMountConfig`, so apply and unapply already receive the same value.

PR #3629, opened 2026-06-13
fix: wake E2B PTY collection on process exit

### LINKED ISSUE #3609: E2B PTY output collection can wait until yield timeout after process exit

he-yufeng:
## Summary
- wake E2B PTY/non-PTY output collection when the underlying command handle exits
- cache the exit code from the waiter and notify the existing collector event
- cover both background command and PTY stdin paths that previously waited for the full yield window when no more output arrived

Fixes #3609

## To verify
- `PYTHONPATH=src python -m pytest tests/extensions/sandbox/test_e2b.py -q`
- `PYTHONPATH=src python -m ruff check src/agents/extensions/sandbox/e2b/sandbox.py tests/extensions/sandbox/test_e2b.py`
- `PYTHONPATH=src python -m py_compile src/agents/extensions/sandbox/e2b/sandbox.py tests/extensions/sandbox/test_e2b.py`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/src/agents/extensions/sandbox/e2b/sandbox.py
+++ b/src/agents/extensions/sandbox/e2b/sandbox.py
@@ -686,6 +686,8 @@ class _E2BPtyProcessEntry:
     output_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
     output_notify: asyncio.Event = field(default_factory=asyncio.Event)
     last_used: float = field(default_factory=time.monotonic)
+    done: bool = False
+    exit_code: int | None = None
 
 
 @dataclass(frozen=True)
@@ -982,6 +984,7 @@ class E2BSandboxSession(BaseSandboxSession):
                     on_data=_append_output,
                 )
                 entry.handle = handle
+                asyncio.create_task(self._run_pty_waiter(entry))
                 await self._sandbox.pty.send_stdin(
                     cast(Any, handle).pid,
                     f"{command_text}\n".encode(),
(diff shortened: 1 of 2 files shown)


seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/3610

Issue #3640, opened 2026-06-14
WorkspaceJsonlSink keeps flushed event bytes in memory for the whole session


luochen211:
### Please read this first

- I have read the Agents SDK docs.
- I searched for related issues and did not find one for `WorkspaceJsonlSink buffer flush memory`.

### Describe the bug

`WorkspaceJsonlSink` keeps all emitted event bytes in `_buf` even after a flush succeeds.

In `src/agents/sandbox/session/sinks.py`:

```python
def _buffer_event(self, event: SandboxSessionEvent) -> bool:
    self._buf.extend(event_to_json_line(event).encode("utf-8"))
    ...

async def _flush_buffer(self) -> None:
    ...
    await self._session.write(relpath, io.BytesIO(bytes(self._buf)))
```

`_flush_buffer()` writes the full buffer to the workspace outbox, but never clears or compacts `_buf`. For long-running sandbox sessions with many audit events, the client-side buffer grows for the lifetime of the session even when `flush_every=1`.

This is observable behavior in the sandbox instrumentation/runtime path, not a docs-only issue.

### Debug information

- Agents SDK version: main branch
- Python version: N/A, code inspection

### Repro steps

Minimal code-inspection reproduction:

```python
from pathlib import Path
from agents.sandbox.session.sinks import WorkspaceJsonlSink

sink = WorkspaceJsonlSink(workspace_relpath=Path("logs/events.jsonl"), flush_every=1)

# After each handled event, _buffer_event appends bytes to sink._buf.
# _flush_buffer writes bytes(sink._buf), but does not clear sink._buf after success.
# Therefore len(sink._buf) grows monotonically with event volume.
```

Code path:

1. Open `src/agents/sandbox/session/sinks.py`.
2. Inspect `WorkspaceJsonlSink._buffer_event()` and `_flush_buffer()`.
3. Observe that `_buf` is appended to for every event.
4. Observe that `_flush_buffer()` writes `bytes(self._buf)` but does not clear `_buf` after a successful write.

### Expected behavior

Either the sink should clear flushed bytes and append to the existing workspace outbox, or it should explicitly document and bound the in-memory full-file rewrite strategy. As implemented, `flush_every` controls write frequency but not memory growth.

PR #3647, opened 2026-06-15
fix: route HttpProxySink timeouts through spool fallback

### LINKED ISSUE #3641: HttpProxySink timeout failures can skip the configured spool fallback
### Please read this first

- I have read the Agents SDK docs.
- I searched for related issues and did not find one for `HttpProxySink TimeoutError spool`.

### Describe the bug

`HttpProxySink` has a spool fallback for failed HTTP delivery, but it only catches `HTTPError` and `URLError`:

```python
try:
    with urlopen(req, timeout=self.timeout_s) as resp:
        _ = resp.read(1)
except (HTTPError, URLError) as e:
    if spool_line is not None and self.spool_path is not None:
        ... write spool_line ...
    raise RuntimeError(...)
```

Timeouts raised directly as `TimeoutError` / socket timeout during connect or read are not caught by this block. In that case, the configured `spool_path` fallback can be skipped and the caller sees a raw timeout exception instead of the sink's `RuntimeError` wrapper.

### Debug information

- Agents SDK version: main branch
- Python version: N/A, code inspection

### Repro steps

A small monkeypatch demonstrates the missed fallback path:

```python
from pathlib import Path

from agents.sandbox.session.sinks import HttpProxySink

sink = HttpProxySink(
    "http://127.0.0.1:9/events",
    timeout_s=0.001,
    spool_path=Path("/tmp/agents-events.jsonl"),
)

# If urlopen raises TimeoutError directly, HttpProxySink._post does not catch it,
# so the spool write branch is skipped.
with patch("agents.sandbox.session.sinks.urlopen", side_effect=TimeoutError("timed out")):
    sink._post(b"{}", '{"event":"example"}\n')
```

Code path:

1. Open `src/agents/sandbox/session/sinks.py`.
2. Inspect `HttpProxySink._post()`.
3. Observe that the failure/spool branch catches `(HTTPError, URLError)` only.
4. Direct `TimeoutError` exceptions bypass the spool branch.

### Expected behavior

`HttpProxySink` should treat timeout exceptions like other delivery failures: write the spool line when configured, then raise the sink's `RuntimeError` wrapper. Catching `TimeoutError` (and possibly `OSError`) alongside `HTTPError`/`URLError` would make the fallback behavior consistent.

atimothee:
### Summary

`HttpProxySink._post()` only caught `HTTPError` and `URLError`, so a `TimeoutError` (for example a socket-level connect timeout, or any other low-level `OSError` such as `ConnectionRefusedError`) bypassed the configured `spool_path` fallback and surfaced the raw exception to callers instead of the sink's `RuntimeError` wrapper.

This change catches `TimeoutError` and `OSError` alongside the existing URL errors so the spool write and the sink's `RuntimeError` wrapper are exercised consistently for every delivery failure, restoring the configured fallback behavior. A short comment documents why the wider exception list is intentional.

### Test plan

- Added a parametrized regression test (`tests/sandbox/test_session_sinks.py::test_http_proxy_sink_writes_spool_on_timeout_or_oserror`) that patches `urllib.request.urlopen` in the sinks module to raise `TimeoutError` and `OSError` in turn, asserts the sink raises `RuntimeError` with the expected message, and asserts the spool file contains the serialized event line.
- Ran `uv run pytest tests/sandbox/test_session_sinks.py -v -k test_http_proxy_sink_writes_spool_on_timeout_or_oserror` (both parametrized cases pass).
- Ran `uv run pytest tests/sandbox/ -q` — 806 passed, 19 skipped (no regressions).
- Ran `make format` and `uv run ruff check`/`uv run mypy`/`uv run pyright` against the changed files — all clean.

### Issue number

Closes #3641

### Checks

### DIFF
--- a/src/agents/sandbox/session/sinks.py
+++ b/src/agents/sandbox/session/sinks.py
@@ -317,7 +317,11 @@ class HttpProxySink(EventSink):
         try:
             with urlopen(req, timeout=self.timeout_s) as resp:
                 _ = resp.read(1)  # ensure request completes
-        except (HTTPError, URLError) as e:
+        except (HTTPError, URLError, TimeoutError, OSError) as e:
+            # `urlopen` wraps most network failures in `URLError`, but socket-level
+            # timeouts (e.g. during connect) can surface as bare `TimeoutError`, and
+            # other low-level failures as `OSError`. Treat all of these as delivery
+            # failures so the configured spool fallback is exercised consistently.
             if spool_line is not None and self.spool_path is not None:
                 try:
                     self.spool_path.parent.mkdir(parents=True, exist_ok=True)
--- a/tests/sandbox/test_session_sinks.py
+++ b/tests/sandbox/test_session_sinks.py
@@ -21,6 +21,7 @@ from agents.sandbox.session import (
     CallbackSink,
     ChainedSink,
     EventPayloadPolicy,
+    HttpProxySink,
     Instrumentation,
     JsonlOutboxSink,
     SandboxSession,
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for the contribution. #3642 already implements the same HttpProxySink fallback fix by catching OSError, which covers HTTPError, URLError, TimeoutError, and other low-level delivery failures. It also includes regression coverage and addresses the related WorkspaceJsonlSink issue in #3640.

To keep a single implementation path, I am going to close this PR as a duplicate of #3642. The additional generic OSError test here is useful, but it is not required to establish the correctness of the simpler catch used there.

PR #3427, opened 2026-05-15
fix: guard None refusal in ItemHelpers.extract_last_content

ioleksiuk:
### Summary

Follow-up to #3394 (just merged) that fixed the text branch of \`extract_last_content\` against \`None\` from \`model_construct\` / provider gateway paths.

The refusal branch at \`src/agents/items.py:693-694\` has the same \`-> str\` return type contract and the same Pydantic-typed \`str\` source field (\`ResponseOutputRefusal.refusal\`), so it can violate the contract in the same way:

\`\`\`python
>>> from openai.types.responses import ResponseOutputMessage, ResponseOutputRefusal
>>> from agents.items import ItemHelpers
>>> refusal_part = ResponseOutputRefusal.model_construct(refusal=None, type="refusal")
>>> msg = ResponseOutputMessage.model_construct(
...     id="m", role="assistant", status="completed",
...     type="message", content=[refusal_part])
>>> ItemHelpers.extract_last_content(msg)
None        # but the function is typed `-> str`
\`\`\`

Apply the same \`or ""\` coercion and rationale comment that the text branch now has (items.py:687-691).

### Scope (apologies for the noise on #3394)

Same scope as the merged version of #3394 — only the \`-> str\` branches get the coercion. \`extract_last_text\` (declared \`-> str | None\`) is intentionally left alone because:

- \`None\` is already a legal return per its signature.
- Coalescing \`\"\"\` → \`None\` would silently change semantics for callers that distinguish empty text from missing text.

This PR is the equivalent fix on the refusal branch that I missed in #3394's first scope.

### Test plan

- New \`test_extract_last_content_returns_empty_string_for_none_refusal\` — verifies the type contract holds for \`refusal=None\`.
- New \`test_extract_last_content_returns_refusal_normally\` — verifies regular refusal text still passes through unchanged.
- \`uv run pytest tests/utils/test_pretty_print_and_items.py tests/test_items_helpers.py\` — 43 passed.
- \`uv run ruff check\` on the touched files — All checks passed.

### Issue number

N/A — direct follow-up to merged #3394.

### Checks

- [x] I've added new tests (if relevant)
- [x] I've added/updated the relevant documentation (no doc changes needed)
- [x] I've run \`make lint\` and \`make format\`
- [x] I've made sure tests pass

PR #3653, opened 2026-06-18
fix: handle missing refusal content

devin-lai:
### Summary

Fix `ItemHelpers.extract_last_content()` so refusal content with a missing payload returns an empty string instead of `None`, preserving the helper's `str` return contract. Add regression coverage for the missing-refusal case and adjust a Python-version-specific test reference so type checking passes on Python 3.11.

### Test plan

- `PATH="/usr/bin:/bin:/usr/sbin:/sbin:/Users/ldy/.local/bin:/Users/ldy/miniforge3/bin:/opt/homebrew/bin" bash .agents/skills/code-change-verification/scripts/run.sh`

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -691,7 +691,8 @@ class ItemHelpers:
             # ``extract_text`` below.
             return last_content.text or ""
         elif isinstance(last_content, ResponseOutputRefusal):
-            return last_content.refusal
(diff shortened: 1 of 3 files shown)


seratch:
Closing this for this reason: https://github.com/openai/openai-agents-python/pull/3427#issuecomment-4464894727

PR #3657, opened 2026-06-18
fix: enforce the documented strict JSON-compatible contract for #3486

siddiksawani:
### Summary

This PR tightens `custom_data_extractor` validation by rejecting non-finite float values such as `NaN`, `Infinity`, and `-Infinity`.

PR #3486 added SDK-only `custom_data` support for tool output items and validates extractor output by round-tripping through `json.dumps()` / `json.loads()`. However, Python’s `json.dumps()` allows non-finite floats by default, even though those values are not strict JSON-compatible.

This change passes `allow_nan=False` so non-finite floats raise `ValueError` and are handled consistently with other non-JSON-compatible custom data.

### Test plan

Added a regression test covering:

* `float("nan")`
* `float("inf")`
* `float("-inf")`

Ran:

```bash
uv run pytest tests/test_tool_custom_data.py -k non_finite
```

### DIFF
--- a/src/agents/util/_custom_data.py
+++ b/src/agents/util/_custom_data.py
@@ -28,7 +28,7 @@ def normalize_custom_data(value: Mapping[str, Any] | None) -> dict[str, Any] | N
 
     copied = copy.deepcopy(dict(value))
     try:
-        return cast(dict[str, Any], json.loads(json.dumps(copied)))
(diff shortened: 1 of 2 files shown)


seratch:
I don't think this is a significant issue in practice, but this is a small and reasonable validation improvement. I'm fine with merging it.

PR #3658, opened 2026-06-18
fix: reject nested non-string custom data keys

JaeCoding:
### Summary

Tighten `custom_data_extractor` validation so nested mappings also require string keys.

PR #3657 made custom data use strict JSON-compatible serialization by rejecting non-finite floats. However, nested mappings with non-string keys were still accepted because `json.dumps()` silently coerces those keys to strings during the validation round trip. For example, `{"nested": {1: "one"}}` became `{"nested": {"1": "one"}}` instead of being rejected.

This change recursively validates mapping keys before serialization so custom data is either preserved as provided or rejected with the existing string-key validation error.

### Test plan

- `make format`
- `make lint`
- `uv run pytest tests/test_tool_custom_data.py -q`
- `make typecheck`
- `env -u ALL_PROXY -u all_proxy -u HTTPS_PROXY -u https_proxy -u HTTP_PROXY -u http_proxy -u GRPC_PROXY_EXP -u grpc_proxy make tests`

### Issue number

Follow-up to #3657 / #3486

### Checks

### DIFF
--- a/src/agents/util/_custom_data.py
+++ b/src/agents/util/_custom_data.py
@@ -15,6 +15,17 @@ CustomDataExtractor = Callable[
 ]
 
 
+def _ensure_mapping_keys_are_strings(value: Any) -> None:
+    if isinstance(value, Mapping):
+        if not all(isinstance(key, str) for key in value):
+            raise UserError("custom_data_extractor must return a mapping with string keys.")
+        for item in value.values():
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the suggestion, but the values are Any. So, we don't need this change.

PR #3663, opened 2026-06-20
fix(memory): return [] for non-positive limit in all SQLite-based sessions

ly-wang19:
## Summary

The Redis, MongoDB, and Dapr session backends short-circuit `get_items()` to return `[]` when the resolved limit is `<= 0`. The **SQLite-based** sessions did not — the core `SQLiteSession` plus the `AsyncSQLiteSession`, `AdvancedSQLiteSession`, and `SQLAlchemySession` backends passed the value straight to SQL `LIMIT`:

- on **SQLite**, `LIMIT -1` means "no limit", so a negative limit returned the **entire** history;
- on **PostgreSQL/MySQL**, a negative `LIMIT` raises.

`resolve_session_limit()` performs no validation, so a user-supplied `get_items(limit=-1)` (or `SessionSettings(limit=-1)`) reaches the backend unchanged. (`limit == 0` already worked via `LIMIT 0`; the bug is negative limits.)

## Fix

Guard for a non-positive resolved limit and return `[]` in all four SQLite-based sessions, matching `RedisSession`, `MongoDBSession`, and `DaprSession`.

## Test

Added a regression test for each (core `SQLiteSession`, `AsyncSQLiteSession`, `AdvancedSQLiteSession`, `SQLAlchemySession`).

Verified red→green: each fails on `main` (`get_items(limit=-1)` returns the full history on SQLite) and passes with the fix. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -172,6 +172,12 @@ class AdvancedSQLiteSession(SQLiteSession):
         """
         session_limit = resolve_session_limit(limit, self.session_settings)
 
+        if session_limit is not None and session_limit <= 0:
+            # Mirror the Redis/MongoDB/Dapr backends: a non-positive limit means
+            # "no items". Without this, LIMIT -1 returns the whole branch on
+            # SQLite (LIMIT -1 == no limit).
+            return []
+
         if branch_id is None:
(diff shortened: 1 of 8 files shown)


seratch:
Thanks for sharing this patch. However, since this is not a real-world use case, we won't have this additional logic.

PR #3677, opened 2026-06-23
fix(tracing): preserve precision of large integers in trace payloads

### LINKED ISSUE #2094: Traces dashboard shows big integer arguments incorrectly in tool calling traces

wustwyh:
Convert Python ints outside JavaScript's safe integer range (-(2**53 - 1) .. 2**53 - 1) to strings before exporting trace data to the OpenAI tracing API. This prevents the trace dashboard from truncating large integer tool arguments or model I/O.

Changes:
- Add _convert_large_ints_to_strings helper in span_data.py.
- FunctionSpanData.export() now converts large ints inside JSON input.
- BackendSpanExporter._sanitize_for_openai_tracing_api() converts large ints recursively across the whole payload.

Fixes #2094

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -15,6 +15,7 @@ import httpx
 
 from ..logger import logger
 from .processor_interface import TracingExporter, TracingProcessor
+from .span_data import _convert_large_ints_to_strings
 from .spans import Span
 from .traces import Trace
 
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for sharing this patch. As I mentioned in the past similar PRs, we won't have this workaround on the client SDK side.

PR #3678, opened 2026-06-23
fix: add hint to duplicated MCP server tool name error message

### LINKED ISSUE #464: Duplicate tool names across MCP servers cause errors

wustwyh:
…e errors

When multiple MCP servers expose identically-named tools, the SDK raises a UserError. Update that error message to mention the existing mcp_config={'include_server_in_tool_names': True} option, which prefixes tool names with their server name to avoid collisions.

Fixes #464

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -302,7 +302,10 @@ class MCPUtil:
                 if duplicate_tool_names:
                     raise UserError(
                         "Duplicate tool names found across MCP servers: "
-                        f"{', '.join(duplicate_tool_names)}"
+                        f"{', '.join(duplicate_tool_names)}. "
+                        "Set `mcp_config={'include_server_in_tool_names': True}` "
+                        "on the agent to prefix tool names with their server name "
+                        "and avoid collisions."
                     )
                 tool_names.update(server_tool_names)
                 tools.extend(function_tools)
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Before merging, please narrow the hint to the non-prefixed path and make it actionable for both supported entry points. `MCPUtil.get_all_function_tools()` can be called directly, in which case changing `agent.mcp_config` does not affect that call; the caller must pass `include_server_in_tool_names=True`. Also, the current patch adds the same hint inside the branch where that option is already enabled, so that message would be contradictory if the defensive check fired.

Please update the wording and regression coverage for those paths. After that, this should be a small, mergeable usability improvement.

wustwyh:
Thanks for the review. I've pushed the requested changes:


  Narrowed the hint to the non-prefixed path only.
  Reworded the hint to be actionable for both entry points (`MCPUtil.get_all_function_tools(..., include_server_in_too
  l_names=True)` and `Agent(mcp_config={...})`).
  Added a regression test for the prefixed path to ensure it raises the plain duplicate error without the hint.


  Please take another look.

seratch (approved):

PR #3679, opened 2026-06-23
fix: #985 Code Interpreter reasoning replay

### LINKED ISSUE #985: Error "Item ‘rs_ABCD’ of type ‘reasoning’ was provided without its required..." when using CodeInterpreter

its-amann:
## Summary
Fixes #985.

- Omit reasoning item IDs when runner-built prior history includes Code Interpreter calls.
- Apply the same policy to server-managed conversation replay so `conversation_id` and `previous_response_id` paths behave consistently.
- Preserve explicit `preserve` and `omit` policy behavior with focused regression coverage.

## Testing
- `python -m pytest tests/test_run_internal_items.py tests/test_server_conversation_tracker.py -q`
- `python -m ruff check src/agents/run_config.py src/agents/run_internal/items.py src/agents/run_internal/oai_conversation.py src/agents/run_internal/run_loop.py tests/test_run_internal_items.py tests/test_server_conversation_tracker.py`
- `python -m ruff format src/agents/run_config.py src/agents/run_internal/items.py src/agents/run_internal/oai_conversation.py src/agents/run_internal/run_loop.py tests/test_run_internal_items.py tests/test_server_conversation_tracker.py --check`
- `python -m mypy src/agents/run_internal/items.py src/agents/run_internal/oai_conversation.py src/agents/run_internal/run_loop.py src/agents/run_config.py --exclude site`

### DIFF
--- a/src/agents/run_config.py
+++ b/src/agents/run_config.py
@@ -319,7 +319,10 @@ class RunConfig:
     reasoning_item_id_policy: ReasoningItemIdPolicy | None = None
     """Controls how reasoning items are converted to next-turn model input.
 
-    - ``None`` / ``"preserve"`` keeps reasoning item IDs as-is.
+    - ``None`` uses the SDK default. Reasoning item IDs are preserved except when runner-built
+      prior turn history contains Code Interpreter calls, where IDs are omitted to avoid invalid
+      replay.
+    - ``"preserve"`` keeps reasoning item IDs as-is.
     - ``"omit"`` strips reasoning item IDs from model input built by the runner.
     """
 
--- a/src/agents/run_internal/items.py
+++ b/src/agents/run_internal/items.py
@@ -35,6 +35,7 @@ __all__ = [
     "REJECTION_MESSAGE",
     "TOOL_CALL_SESSION_DESCRIPTION_KEY",
     "TOOL_CALL_SESSION_TITLE_KEY",
+    "apply_reasoning_item_id_policy",
     "copy_input_items",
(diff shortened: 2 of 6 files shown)


seratch:
Thanks for the contribution. I tested the reported Code Interpreter continuation scenario against the current Responses API using o4-mini, changing reasoning effort from low to medium.

I could not reproduce the original failure. The existing SDK payload, with reasoning IDs preserved, completed successfully. previous_response_id continuation also completed successfully. In contrast, removing the reasoning IDs as proposed here caused the current API to reject the request with a 400 because the code_interpreter_call was missing its required reasoning item.

The original API behavior appears to have changed, so the SDK should preserve its current behavior. This PR would now introduce a regression rather than fix the reported issue. I am going to close it. If the original handoff-specific error can still be reproduced with the current API, please provide the request payload and we can investigate that narrower case.

PR #3680, opened 2026-06-23
fix(realtime): emit history updates for transcript deltas

### LINKED ISSUE #2940: RealtimeSession does not emit `history_updated` for `transcript_delta`

its-amann:
### Summary

Closes #2940.

Realtime transcript deltas already update session history, but high-level event consumers only received the raw model event. This emits `RealtimeHistoryUpdated` after each transcript delta so UI code can keep rendering from the session history stream without also subscribing to raw model events.

The regression test covers two deltas for the same item and verifies the emitted history contains the accumulated transcript.

### Test plan

- `uv run pytest tests\realtime\test_session.py::TestEventHandling::test_transcript_delta_event_updates_history -q`
- `uv run pytest tests\realtime\test_session.py::TestEventHandling -q`
- `uv run pytest tests\realtime\test_session.py -q`
- `uv run ruff format src\agents\realtime\session.py tests\realtime\test_session.py`
- `uv run ruff check --fix src\agents\realtime\session.py tests\realtime\test_session.py`
- `uv run ruff format --check src\agents\realtime\session.py tests\realtime\test_session.py`
- `uv run ruff check src\agents\realtime\session.py tests\realtime\test_session.py`
- `uv run mypy src\agents\realtime\session.py tests\realtime\test_session.py --exclude site`
- `uv run pyright src\agents\realtime\session.py tests\realtime\test_session.py`

### Issue number

Closes #2940

### Checks

- [x] I've run `make lint` and `make format` equivalents with direct `uv run ruff` commands

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -389,6 +389,9 @@ class RealtimeSession(RealtimeModelListener):
                     content=[AssistantAudio(transcript=self._item_transcripts[item_id])],
                 ),
             )
+            await self._put_event(
+                RealtimeHistoryUpdated(info=self._event_info, history=self._history)
+            )
 
             # Check if we should run guardrails based on debounce threshold
             current_length = len(self._item_transcripts[item_id])
--- a/tests/realtime/test_session.py
+++ b/tests/realtime/test_session.py
@@ -822,6 +822,45 @@ class TestEventHandling:
         assert len(history_event.history) == 1
 
     @pytest.mark.asyncio
+    async def test_transcript_delta_event_updates_history(self, mock_model, mock_agent):
+        """Test that transcript delta events update history and emit events"""
+        session = RealtimeSession(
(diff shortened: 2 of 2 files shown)


seratch:
Closing this for the reason I mentioned here: https://github.com/openai/openai-agents-python/issues/2940#issuecomment-4784069571

PR #3684, opened 2026-06-24
Add FastAPI AgentServer extension (invoke / stream / threads)

fede-kamel:
### Summary

Adds a small, batteries-included **FastAPI `AgentServer`** as a new optional extension, `agents.extensions.server`, so going from a local `Agent` to a deployable HTTP service doesn't require hand-writing the web layer, SSE serialization, and per-thread session wiring.

**Why this matters:** the SDK already has every runtime piece needed to serve an agent — `Runner.run`, `Runner.run_streamed`, `Session` — but stops at the library boundary, so everyone re-implements the same glue. This closes the SDK's deployment "last mile" with a blessed, opt-in implementation and a standard `invoke` / `stream` / `threads` contract.

### What's included

`AgentServer(agent, *, session_factory=None, api_key=None, max_turns=None)` builds a FastAPI app:

- `POST /invoke` — `{"input", "thread_id?"}` → runs the agent (with a per-thread `Session` when a factory is supplied) → `{"output", "thread_id"}`.
- `POST /stream` — Server-Sent Events: `text_delta`, run-item events, then a terminal `done` event with the final output.
- `GET /threads/{id}` / `DELETE /threads/{id}` — read or clear a thread's persisted items.
- `GET /health` — readiness probe.

**Security by default:** optional API key (`Authorization: Bearer` / `X-API-Key`); `run()` refuses to bind a non-loopback host when no `api_key` is set. `server.app` is a plain FastAPI app, so it can also be mounted in an existing service.

### Dependency handling

Gated behind a new `pip install "openai-agents[server]"` extra (`fastapi`, `uvicorn`). **FastAPI is not added to core deps.** The package lazily imports the app on first access of `AgentServer`, so `import agents` never imports FastAPI; using it without the extra raises a friendly "install openai-agents[server]" error.

Purely additive — no changes to existing public API ordering, so compatibility risk against the latest release is low.

### Demo (use case)

Runnable gist — a support agent served over HTTP with per-conversation memory, streaming, and auth; cross-call memory recall verified end-to-end against a real model: https://gist.github.com/fede-kamel/d6dbdae6ae92d599de80a64b06ea0248

### Test plan

- `tests/extensions/server/test_agent_server.py` — 9 unit tests (`/health`, `/invoke` + validation, `/stream` SSE, thread persistence/clear, threads-disabled-without-factory, auth via Bearer + `X-API-Key`, `run()` public-bind guard).
- `tests/extensions/server/test_agent_server_integration.py` — 2 real-model integration tests (`/invoke`, `/stream`), marked `allow_call_model_methods`, skipped unless a real `OPENAI_API_KEY` is set.
- `make format`, `make lint`, `make typecheck` (mypy + pyright) all clean; full suite green.

### Out of scope (follow-ups)

WebSocket transport, auth backends beyond a shared key, multi-agent routing at the HTTP layer.

### Checks


seratch:
Thank you for sharing this interesting idea! While this could be useful for some use cases, we don't plan to add this as a feature of this SDK itself. Please feel free to share it in your own repo or any other resources and/or publish it as your own package. Thanks again for your interest here.

PR #3689, opened 2026-06-24
fix: close Chat Completions streams on early exit

fallintoplace:
## Summary

- Add Chat Completions stream cleanup helpers matching the Responses model behavior.
- Close the underlying provider stream in `stream_response()` when iteration ends early or normally.
- Schedule provider stream cleanup in the background when caller cancellation is in progress.
- Add a regression test that closes the async generator early and verifies the provider stream is closed.

## Why

The Chat Completions streaming path previously yielded converted events from the provider stream without closing the underlying HTTP/SSE stream if the caller stopped early. That can leave provider resources open longer than intended.

## Tests

- `uv run pytest tests/models/test_openai_chatcompletions_stream.py::test_stream_response_close_closes_provider_stream`
- `uv run pytest tests/models/test_openai_chatcompletions_stream.py`
- `uv run ruff check src/agents/models/openai_chatcompletions.py tests/models/test_openai_chatcompletions_stream.py`


seratch (changes requested):
Thanks for the contribution. The implementation aligns with the existing Responses cleanup behavior and looks sound.

Before merging, please update the regression test to exercise asynchronous `close()`, which is the method exposed by the actual OpenAI `AsyncStream`, rather than only testing `aclose()`. The broader cancellation behavior is already covered by the equivalent Responses tests and does not need to be duplicated here.

Once that focused test change is made, this should be ready to merge.

seratch (approved):

PR #3692, opened 2026-06-25
feat: expose the wrapped callable on FunctionTool (closes #3381)

### LINKED ISSUE #3381: Provide stable public access to the underlying function on `FunctionTool`
## Summary

When `@function_tool` is applied to a function, the resulting `FunctionTool`
dataclass captures the original callable **only** in the closure of
`_on_invoke_tool_impl` (free var `the_func`, reachable today via
`function_tool.on_invoke_tool._invoke_tool_impl.__closure__`). There is no
public attribute (`func`, `fn`, `__wrapped__`) and no `__call__` forward.

PR #2146 was opened in December 2025 to add a `.func` attribute and was
closed unmerged in February 2026 over an unrelated pytest fixture-collection
regression. The underlying ergonomic gap remains, and from 0.16 onward
the `_FailureHandlingFunctionToolInvoker` indirection made the
closure-walking workaround more expensive (one more attribute hop).

## Why this matters

Anything outside the Agents runtime that wants to introspect, ship, or
call the user's tool body has no stable hook:

- **Code-shipping SDKs** that bundle the user's tool source and re-run
  the bottom function in a sandbox. The standard unwrap chain
  (`__wrapped__` / `func` / `fn`) misses every `@function_tool`-decorated
  symbol, so customers get an `AttributeError` at build time (or a
  non-callable wrapper at runtime) on the very first example any
  tutorial shows.
- **Direct testing** of the user's tool body without spinning up an
  Agent / ToolContext / JSON encode-decode cycle (#708).
- **Migrating across agent frameworks** — the original callable is the
  cleanest portability surface.

## Proposed fix

Either of these is sufficient; both are compatible with the existing
dataclass shape.

**Option 1 — public attribute** (what #2146 was driving at):

```python
@dataclass
class FunctionTool:
    ...
    _func: ToolFunction[...] | None = field(default=None, kw_only=True, repr=False)

    @property
    def func(self) -> ToolFunction[...] | None:
        return self._func
```

with `_create_function_tool` passing `_func=the_func` when constructing
the dataclass. No `functools.update_wrapper`, which dodges the pytest
collection issue that closed #2146.

**Option 2 — callable forward** (ergonomic ceiling, complementary to 1):

```python
    if self._func is None:
        raise RuntimeError("FunctionTool has no underlying function")
    return self._func(*args, **kwargs)
```

Option 1 unblocks SDKs and tests immediately; Option 2 makes
`my_tool(...)` work in notebooks and unit tests too.

## Workaround in the wild

Pending an upstream fix, downstream SDKs walk
`on_invoke_tool._invoke_tool_impl.__closure__` for the free var named
`the_func`. That's private-implementation-detail spelunking — every
release risks a silent break. A public attribute lets us delete the
workaround entirely.

Happy to PR whichever shape a maintainer prefers.

fede-kamel:
### Summary

Adds a public `FunctionTool.function` attribute that holds the original Python callable wrapped by `@function_tool`.

Today the wrapped function is only reachable by walking the closure of `on_invoke_tool` (`...on_invoke_tool._invoke_tool_impl.__closure__`), which is brittle and got an extra hop in 0.16 with the failure-handling indirection. This gives code outside the Agents runtime a stable hook to introspect, directly test, or re-run a tool's body without constructing a `ToolContext` and doing a JSON encode/decode cycle.

- New field is `None` for tools not backed by a plain function (agent-as-tool, hosted-tool wrappers), so the attribute is always present and safe to check.
- Added as a keyword-only dataclass field, so it does **not** change the positional constructor order of `FunctionTool`.
- Documented via the field docstring (renders in the API reference).

This revives the approach from #2146 (which was closed only over an unrelated pytest fixture-collection regression, not the concept). Happy to rename to `func` or also wire `__wrapped__` for `inspect.unwrap()` if preferred.

### Test plan

- New `tests/test_function_tool_exposes_function.py`: sync + async tools expose a runnable `.function`; directly-constructed tools expose `None`.
- `tests/test_function_tool.py` (49 tests) still pass — no regression.
- `make format`, `make lint`, `make typecheck` (mypy + pyright) clean.

### Issue number

Closes #3381

### Checks

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -455,6 +455,14 @@ class FunctionTool:
     )
     """Optional callback that attaches SDK-only custom data to the tool output item."""
 
+    function: ToolFunction[...] | None = field(default=None, kw_only=True, repr=False)
+    """The original Python callable wrapped by `function_tool`, when available.
+
+    This is `None` for tools that are not backed by a plain function (for example,
+    agent-as-tool or hosted-tool wrappers). It gives code outside the Agents
+    runtime a stable hook to introspect, directly test, or re-run the underlying
+    function without going through a `ToolContext`/JSON round-trip."""
+
     _failure_error_function: ToolErrorFunction | None = field(
         default=None,
         kw_only=True,
@@ -619,6 +627,7 @@ def _build_wrapped_function_tool(
     sync_invoker: bool = False,
     mcp_title: str | None = None,
     tool_origin: ToolOrigin | None = None,
+    function: ToolFunction[...] | None = None,
 ) -> FunctionTool:
     """Create a FunctionTool with copied-tool-aware failure handling bound in one place."""
     on_invoke_tool = _with_context_function_tool_failure_error_handler(
@@ -644,6 +653,7 @@ def _build_wrapped_function_tool(
             timeout_error_function=timeout_error_function,
             defer_loading=defer_loading,
             custom_data_extractor=custom_data_extractor,
+            function=function,
             _mcp_title=mcp_title,
             _tool_origin=tool_origin,
         ),
@@ -2033,6 +2043,7 @@ def function_tool(
             timeout_error_function=timeout_error_function,
             defer_loading=defer_loading,
             custom_data_extractor=custom_data_extractor,
+            function=the_func,
             sync_invoker=is_sync_function_tool,
         )
         return function_tool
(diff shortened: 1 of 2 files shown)


fede-kamel:
Following up on the `duplicate` label — agreed, this overlaps with the earlier #3396, which exposes the same original callable (there as `FunctionTool.func`; here as `FunctionTool.function`). The implementations are close to equivalent, so I'm happy to close this one in favor of #3396 if that's the preferred candidate — or to fold anything useful from this PR's tests into it. Whichever keeps the review load smallest.

fede-kamel:
Closing in favor of the earlier #3396, which covers the same ground. Hope the extra test angles here are useful as a reference if needed.

PR #3693, opened 2026-06-25
feat: support @function_tool on instance methods (closes #94)

### LINKED ISSUE #94: Add `method_tool` Functionality

fede-kamel:
### Summary

Makes `@function_tool` work on **instance methods**. Today, decorating a method and passing the bound method fails because `self` ends up in the JSON schema (and isn't supplied at call time):

```python
class Tools:
    @function_tool
    def get_random_number(self) -> int: ...

agent = Agent(..., tools=[Tools().get_random_number])
# -> BadRequestError: schema for 'self' must have a 'type' key
```

This implements the descriptor protocol on `FunctionTool`: when a method-backed tool is accessed via an instance, `__get__` returns a copy **bound to that instance**. Binding rebuilds the tool from the bound method (whose signature omits `self`), so the schema and invocation are correct and `self` is passed automatically. Bound tools are cached per instance.

- Module-level `@function_tool` functions are unaffected (only tools whose first parameter is `self` get a binder; class/non-instance access returns the tool unchanged).
- Reuses the existing tool-construction path, so descriptions, guardrails, approval, timeouts, strict-schema, etc. are preserved on the bound tool.

This addresses @rm-openai's "I'll take a look" on the issue and removes the need for the `function_tool(instance.method)` workaround.

### Test plan

- New `tests/test_function_tool_methods.py`: `self` excluded from schema; invocation supplies `self`; distinct instances bind independently; per-instance caching; class-level access returns the unbound tool; module-level functions unaffected; and an end-to-end `Runner.run` with a `FakeModel` calling the bound method.
- `tests/test_function_tool.py` + `tests/test_function_schema.py` (75 tests) still pass — no regression.
- `make format`, `make lint`, `make typecheck` (mypy + pyright) clean. Docs example added under Tools › Function tools.

### Issue number

Closes #94

### Checks


fede-kamel:
Following up on the `duplicate` label — this overlaps with the earlier #3458, which takes the same bind-on-instance-access approach to instance-method tools. I'm happy to close this one in favor of #3458 if that's the preferred candidate, or to consolidate this PR's extra regression tests and docs there. Whichever is easiest for review.

fede-kamel:
Closing in favor of the earlier #3458, which takes the same approach. Happy to see instance-method support land through that PR.

PR #3695, opened 2026-06-26
Validate tool output trimmer allowlist entries

CodingFeng101:
## Summary
- Validate every `trimmable_tools` iterable member is a string
- Add a focused regression test for mixed-type allowlists

## Why
`ToolOutputTrimmer.trimmable_tools` accepts a string or iterable of strings. Passing an iterable with a non-string member currently succeeds silently and stores that value in the allowlist, making configuration mistakes harder to diagnose when tool outputs are not trimmed as expected.

This keeps valid string and iterable configurations unchanged, but fails fast for invalid mixed-type iterables.

## Validation
- `python -m pytest tests/extensions/test_tool_output_trimmer.py -q`
- `python -m ruff check src/agents/extensions/tool_output_trimmer.py tests/extensions/test_tool_output_trimmer.py`
- `python -m ruff format --check src/agents/extensions/tool_output_trimmer.py tests/extensions/test_tool_output_trimmer.py`
- `git diff --check`

### DIFF
--- a/src/agents/extensions/tool_output_trimmer.py
+++ b/src/agents/extensions/tool_output_trimmer.py
@@ -85,6 +85,8 @@ class ToolOutputTrimmer:
                 trimmable_tools = self.trimmable_tools
             else:
                 trimmable_tools = frozenset(self.trimmable_tools)
+            if not all(isinstance(tool_name, str) for tool_name in trimmable_tools):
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the suggestion. The type hint for trimmable_tools clearly tells the elements must be string, plus I don't think this is confusing. For these reasons, let me close this PR now. Thanks again for your interest!

PR #3698, opened 2026-06-28
docs: clarify inter-turn input filtering

maxpetrusenkoagent:
### Summary

- document the supported `call_model_input_filter` pattern for injecting queued user/application input between tool turns
- clarify the boundaries: the filter changes the next model payload, but does not mutate `RunResult.input`, auto-persist injected items to a client-managed session, or pause the run
- add non-streamed and streamed characterization tests showing queued context messages injected after a tool call and before the next model turn

This follows the maintainer guidance on #2671 that `RunConfig.call_model_input_filter` is the existing supported hook for history compaction and queued external-message injection between tool turns, without adding a new turn hook/interceptor API.

### Test plan

- `make format`
- `uv run pytest tests/test_call_model_input_filter.py -q`
- `uv run mypy tests/test_call_model_input_filter.py`
- `uv run pyright tests/test_call_model_input_filter.py --project pyrightconfig.json`
- `uv run mkdocs build`
- Second-agent review: `claude -p` on `/tmp/oss-pr-second-agent-review.diff` returned `CLEAN` with no blocking findings after the doc-example nit was fixed.

Full-suite note: I also ran the repo's `$code-change-verification` flow. `make lint` passed, but the full `make tests`/`make typecheck` gates are currently noisy in this local Python 3.14 environment due to unrelated optional-dependency and existing sandbox/voice/model failures outside this diff. The focused tests and file-level type checks above pass.

### Issue number

Related to #2671

### Checks

### DIFF
--- a/docs/running_agents.md
+++ b/docs/running_agents.md
@@ -461,6 +461,52 @@ If you are using OpenAI server-managed conversation state with `conversation_id`
 
 Set the hook per run via `run_config` to redact sensitive data, trim long histories, or inject additional system guidance.
 
+#### Injecting input between tool turns
+
+Because `call_model_input_filter` runs before every model call, it can also reconcile host-side state that changed while tools were running. A common pattern is to keep a small queue in your local context, let tools or background application code append new user messages to that queue, and drain it in the filter before the next model turn.
+
+```python
+from dataclasses import dataclass, field
+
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the suggestion. We don't generally encourage appending user messages this way at this moment, so let me close this PR. Regarding the queued message use case, we're aware of the need. So, we may come up with some solution for it in the future.

Issue #3697, opened 2026-06-27
Pattern: external governance gate via on_tool_start (independent /review before irreversible tool calls)


babyblueviper1:
The SDK already ships `on_tool_start` / `on_tool_end` and a set of guardrail examples — `input_guardrails.py`, `output_guardrails.py`, `llm_as_a_judge.py`. One pattern that falls between these is an **external governance gate**: a second opinion that isn't the same model, running before high-stakes tool execution, with a signed verifiable proof attached.

`llm_as_a_judge.py` uses the model to grade its own output. The complement: an independent party that isn't the engine, so the grade isn't from the optimist that produced the work.

**Pattern using `on_tool_start`:**

```python
import httpx
from agents import RunHooks, RunContextWrapper, Agent, Tool

REVIEW_URL = "https://api.babyblueviper.com/review"
HIGH_STAKES_TOOLS = {"shell", "apply_patch", "file_write", "deploy", "transfer", "swap"}

class GovernanceHooks(RunHooks):
    async def on_tool_start(
        self, context: RunContextWrapper, agent: Agent, tool: Tool
    ) -> None:
        if tool.name not in HIGH_STAKES_TOOLS:
            return

        async with httpx.AsyncClient() as client:
            try:
                r = await client.post(
                    REVIEW_URL,
                    json={
                        "artifact_type": "tool_call",
                        "artifact": {
                            "tool": tool.name,
                            # tool.args available at on_tool_start via ToolContext
                            "args": getattr(context, "tool_arguments", {}),
                        },
                        "sign": True,  # returns a signed, recomputable proof
                    },
                    timeout=5.0,
                )
                if r.status_code == 200:
                    data = r.json()
                    if data.get("verdict") == "reject":
                        raise ValueError(
                            f"[Governance] {tool.name} blocked: {data.get('summary')}\n"
                            f"proof: {data.get('proof', {}).get('id')}"
                        )
                    # proof_id can be logged or attached to context for audit trail
            except httpx.TimeoutException:
                pass  # advisory — never block on gate unavailability

# wire it
result = await Runner.run(agent, input="...", hooks=GovernanceHooks())
```

**Why external matters for high-stakes tools:**

The existing guardrails run the same model on its own output — useful for formatting or safety checks, but the model that decided to call `shell` is also grading whether `shell` should be called. For high-stakes irreversible actions (deploy, pay, delete, trade), the grade coming from the same engine is the problem.

An external gate:
[...]

PR #3699, opened 2026-06-28
docs: add external governance gate example

tarunag10:
## Summary
- add an `external_governance_gate.py` agent pattern using `RunHooks.on_tool_start`
- document when to use automated governance gates versus manual HITL interruptions
- link the new example from the examples index

## Test plan
- `uv run ruff check examples/agent_patterns/external_governance_gate.py`
- `uv run python -m py_compile examples/agent_patterns/external_governance_gate.py`
- `git diff --check`

Related: https://github.com/openai/openai-agents-python/issues/3697

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -21,6 +21,7 @@ Check out a variety of sample implementations of the SDK in the examples section
     -   Human-in-the-loop with tool approval and state serialization (`examples/agent_patterns/human_in_the_loop.py`)
     -   Human-in-the-loop with streaming (`examples/agent_patterns/human_in_the_loop_stream.py`)
     -   Custom rejection messages for approval flows (`examples/agent_patterns/human_in_the_loop_custom_rejection.py`)
+    -   External governance gates before local tool execution (`examples/agent_patterns/external_governance_gate.py`)
 
 -   **[basic](https://github.com/openai/openai-agents-python/tree/main/examples/basic):**
     These examples showcase foundational capabilities of the SDK, such as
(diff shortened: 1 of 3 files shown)


babyblueviper1:
@tarunag10 — thank you for turning #3697 into a runnable example. The `on_tool_start` fail-closed gate is exactly the right primitive, and separating "automated policy/audit gate" from "manual HITL interruption" is the distinction the docs were missing. The `refund_order` threshold case lands it cleanly.

One enhancement worth a line, since the docs already gesture at it ("store policy receipts if you need an audit trail"): a gate is most useful when its decision is **recomputable, not opaque**, and leaves a **portable receipt**. Two small shape suggestions for `external_policy_allows`:

1. **Return a structured verdict, not a bare bool** — `{verdict, reasons[], recompute}` — so a denial is explainable and a third party can re-derive it from the same tool args. The refund check is already deterministic (`amount_usd > 100` → deny), so it's recomputable by construction; surfacing *why* makes the gate auditable after the fact.
2. **Return a receipt bound to the tool call** (tool name + args hash + decision + timestamp), ideally signed, so "this call was denied/allowed *before* it executed" is provable later without trusting the app's own logs. That's the difference between an audit trail you keep and one a counterparty can check.

For anyone who wants a drop-in for the placeholder: an independent verifier like `/review` (the service behind #3697) returns exactly that shape — a recomputable verdict plus a portable signed proof — and there's an open, vendor-neutral conformance suite for what a pre-execution gate should satisfy (recomputable verdict binding, independent admission, pre-execution anchoring) if folks want to compare gate implementations. Happy to contribute a short variant of the example showing the receipt-returning shape if it'd help.

Either way it's a good addition as-is — glad to see the pattern documented.

seratch:
Closing this for the reason I mentioned at https://github.com/openai/openai-agents-python/issues/3697#issuecomment-4827966915

PR #3702, opened 2026-06-29
Docs: clarify when to use a guardrail vs. tool approval

Tarshann:
﻿### What
Adds a "Guardrails vs. tool approval — which gate when" section to
`docs/human_in_the_loop.md`, plus a one-line cross-reference from the
"Tool guardrails" section of `docs/guardrails.md`.

### Why
The SDK ships two distinct interception points — input/output/tool **guardrails**
(content checks that can tripwire-halt a run, or skip/replace a tool call) and
per-tool **`needs_approval`** (pauses a specific tool call into
`RunResult.interruptions` for an out-of-band decision) — and even documents their
*ordering*, but never says *which one to reach for*. Users repeatedly conflate the
two (e.g. #2868). This adds a decision table, rules of thumb, and the safety note
that `needs_approval` defaults to `False` (opt-in, not deny-by-default), with an
explicit gating recipe and the dynamic `(ctx, params, call_id) -> bool` callable form.

### Scope
Docs-only, additive, no new dependency. Folded into the existing in-nav
`human_in_the_loop.md` rather than a new page, to avoid touching the i18n
(ja/ko/zh) nav blocks. Happy to split into a standalone page if preferred — the
section lifts out cleanly.

### Verification
- `mkdocs build` succeeds; built with and without this change — both emit an
  identical set of pre-existing warnings (zero new warnings), and the new anchor
  and cross-reference resolve in the generated HTML.
- All API symbols in the snippets were checked against `openai-agents` 0.17.7:
  `function_tool(needs_approval=...)` (default `False`), the async
  `(ctx, params, call_id) -> bool` callable, `RunResult.interruptions`,
  `RunResult.to_state()`, `RunState.approve(item)`, `RunState.reject(item, ...)`,
  and `Runner.run(agent, state)`.


seratch:
Thanks for sharing this suggestion. However, we think human approvals and tool guardrails are different use cases, so we don't want to have this section in the docs.

PR #3705, opened 2026-06-29
fix(sandbox): use consistent model for memory phase one and phase two

### LINKED ISSUE #3703: Incorrect model was used?
### Please read this first

### Describe the bug
1. I copied the code of https://github.com/openai/openai-agents-python/blob/main/examples/sandbox/memory.py
2. Run with one change `DEFAULT_MODEL = "gpt-5.4-mini"`
3. The first phase was indeed gpt-5.4-mini but all turns of second phase is gpt-5.5

Image

### Debug information
- Agents SDK version: (e.g. `v0.17.7`)
- Python version (e.g. Python 3.12)

### Repro steps

As stated above.

### Expected behavior
Expected all turns to be gpt-5.4-mini.

anishesg:
The memory generation system uses two phases: phase one extracts raw memories from individual rollouts, and phase two consolidates them. Previously, phase one defaulted to model-5.4-mini while phase two hardcoded model-5.5, causing inconsistent model usage during memory generation that was visible in traces and could confuse users who set a custom model in example scripts.

Changed the default for `MemoryGenerateConfig.phase_two_model` in `/work/repo/src/agents/sandbox/config.py` from "model-5.5" to "model-5.4-mini" to match `phase_one_model`. Both phases now use the same default model, ensuring consistent behavior and eliminating unexpected model switches during memory consolidation.

Added a regression test in `/work/repo/tests/sandbox/test_memory.py` to verify both phase models use the same default value.

Fixes #3703

### DIFF
--- a/src/agents/sandbox/config.py
+++ b/src/agents/sandbox/config.py
@@ -49,7 +49,7 @@ class MemoryGenerateConfig:
     )
     """Model settings used for phase-1 single-rollout extraction."""
 
-    phase_two_model: str | Model = "gpt-5.5"
+    phase_two_model: str | Model = "gpt-5.4-mini"
     """Model used for phase-2 memory consolidation."""
 
     phase_two_model_settings: ModelSettings | None = field(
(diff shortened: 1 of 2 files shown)


seratch:
Closing this; see https://github.com/openai/openai-agents-python/issues/3703#issuecomment-4833246280

Issue #3459, opened 2026-05-19
Make unknown tool calls recoverable instead of aborting with ModelBehaviorError


yclzju:
### Describe the bug

When the model returns a function tool call whose name is not registered in the current agent, the SDK raises `ModelBehaviorError` and aborts the whole run.

The relevant path appears to be:

`process_model_response()` -> `build_function_tool_lookup_map()` -> `function_map.get(...) is None` -> `raise ModelBehaviorError(...)`

This can happen when the model hallucinates a tool name, when tools are loaded dynamically, or when tool availability changes between turns. In long-running agent sessions, aborting the entire run is difficult to recover from.

### Debug information

- Agents SDK version: `<version>`
- Python version: `<version>`
- Model/backend: `<Responses API / ChatCompletions / custom provider>`

### Repro steps

1. Create an agent with a limited set of tools.
2. Cause the model to return a tool call for a tool name that is not in the current agent's tools.
3. Run the agent.

Actual result:

```text
agents.exceptions.ModelBehaviorError: Tool <unknown_tool> not found in agent <agent_name>

yclzju:
Hi, I tested this with version 0.17.3, but the issue still occurs when the get_date tool is not registered in the agent. The error is:

ModelBehaviorError: Tool get_date not found in agent agent-main

utsavtulsyan:
This failure mode feels worth treating as a recoverable model-visible event, not only a local SDK exception.

In production, the useful distinction is whether the model asked for a genuinely unsupported tool, or whether the available tool registry changed between turns. A small structured error back into the run, with known tool names or a policy reason when names are intentionally hidden, would let the agent repair its next step while keeping logs and metrics honest.

I would still want a strict hard-fail mode for teams that prefer tight contracts, but long-running sessions probably need the softer path.

yclzju:
Yeah, I’ve noticed that frameworks like OpenClaw and Hermes can already support this kind of tool-call failure handling. From the LLM’s perspective, as long as it receives explicit feedback, it can usually recover and invoke the correct tool in the next step. In that sense, this is very similar to handling parameter errors in tool calls: the model receives structured feedback, corrects itself, and retries appropriately.

Maybe the SDK could expose a hook for this behavior, allowing developers to choose between strict validation with hard failure, or a softer path where the run continues and the model is allowed to reflect and repair its next action.

PR #3707, opened 2026-06-29
feat: honor tool_not_found_behavior for missing custom tool calls

gaurav0107:
### Summary

When the model emits a custom tool call (`custom_tool_call`) for a tool the
agent does not have, `process_model_response` always raised
`ModelBehaviorError` and aborted the whole run, ignoring
`RunConfig.tool_not_found_behavior`. The `"return_error_to_model"` opt-in was
only honored for function tool calls.

This change makes a missing custom tool follow the same recovery path as a
missing function tool: when `tool_not_found_behavior="return_error_to_model"`
is set, the unresolved custom tool call is surfaced back to the model as a
`custom_tool_call_output` error (the correct output type for a custom tool
call) and the run continues, instead of crashing. The default
`"raise_error"` behavior is completely unchanged.

Specifically:
- Add `ToolRunCustomNotFound` and a `custom_tools_not_found` field on
  `ProcessedResponse` (appended at the end; internal run-loop types).
- Honor `return_error_to_model` in the `ResponseCustomToolCall` not-found
  branch, mirroring the existing function-tool branch.
- Build a `custom_tool_call_output` error item for missing custom tools,
  reusing the same shape already produced by the successful and rejected
  custom-tool paths.
- Pass `tool_type="custom"` through to `ToolErrorFormatterArgs` so a custom
  `tool_error_formatter` sees the correct tool type.
- Update the `tool_not_found_behavior` docstring to cover custom tools.

### Test plan

- `make format`, `make lint`, `make typecheck` (mypy + pyright) all pass.
- `make tests` passes (full suite).
- Added two tests in `tests/test_process_model_response.py`:
  - a missing custom tool still raises `ModelBehaviorError` by default, and
  - with `tool_not_found_behavior="return_error_to_model"` it records one
    `custom_tools_not_found` entry and does not raise.

### Issue number

Related to #325 (focused slice: recover from the model calling a
non-existent custom tool, rather than a general retry mechanism).

### Checks

- [x] I've run `make format`, `make lint`, `make typecheck`, and `make tests`

### DIFF
--- a/src/agents/run_config.py
+++ b/src/agents/run_config.py
@@ -330,11 +330,12 @@ class RunConfig:
     """Optional SDK-side execution settings for local tool calls."""
 
     tool_not_found_behavior: ToolNotFoundBehavior = "raise_error"
-    """Controls unresolved function tool calls emitted by the model.
+    """Controls unresolved function and custom tool calls emitted by the model.
 
     - ``"raise_error"`` preserves the default behavior and raises ``ModelBehaviorError``.
-    - ``"return_error_to_model"`` returns a model-visible ``function_call_output`` error and lets
-      the run continue.
+    - ``"return_error_to_model"`` returns a model-visible error and lets the run continue. The
+      error is surfaced as a ``function_call_output`` for missing function tools and as a
+      ``custom_tool_call_output`` for missing custom tools.
     """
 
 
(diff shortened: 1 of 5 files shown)


seratch:
Thanks for the contribution. The implementation demonstrates that missing custom tool calls could follow the existing function-tool recovery path, but we do not currently have evidence of a concrete user-facing need for this extension.

The reports in #325 and #3459 concern missing function tools, which are already covered by #3461. This PR does not include a real custom-tool reproduction, provider scenario, or user report showing that an unavailable custom tool call occurs in practice. Given the narrow reach and the additional runtime, tracking, testing, and documentation surface, I do not think the maintenance cost is justified yet.

I am going to close this PR for now. We can revisit the design if we receive a concrete custom-tool reproduction, including the provider and how the unavailable tool call was produced.

PR #3711, opened 2026-06-30
fix: export get_trace_provider and TraceProvider from agents

anxkhn:
### Summary

The tracing global-provider accessor `get_trace_provider` and its type `TraceProvider` are re-exported from the `agents.tracing` subpackage (both are in `agents.tracing.__all__`), but were missing from the top-level `agents` package, even though their sibling `set_trace_provider` is exported there.

As a result, `from agents import set_trace_provider` worked but `from agents import get_trace_provider` raised `ImportError`, and the `TraceProvider` type could not be imported from `agents` to annotate the handle. A user who sets a custom provider naturally wants to read it back via `get_trace_provider()` and annotate it with `TraceProvider` from the same import surface.

This adds both names to the `from .tracing import (...)` block and to `__all__` in `src/agents/__init__.py`, so the get/set/type triad shares one import surface. It follows the "Public API Compatibility" rule in `AGENTS.md` (import paths and `__all__` membership are compatibility contracts; appending is backward compatible) and mirrors the recent #3489 (`573530f`, "export more tracing related functions & types from agents"), which exported a batch of tracing symbols here but missed these two. The change is purely additive to the public surface, with no runtime or behavior change.

### Test plan

- New `tests/tracing/test_tracing_exports.py` mirrors `tests/test_exception_exports.py`: it asserts that `from agents import get_trace_provider` / `TraceProvider` works, that each object is identical to its `agents.tracing.<name>` counterpart, and that both are in `agents.__all__`. The test fails on `main` (ImportError plus absence from `__all__`) and passes with this change.
- Full local stack is green: `make format` (no changes), `make lint`, `make typecheck` (pyright 0 errors; the lone mypy `eager_task_factory` error is pre-existing on `main` in an untouched file and is a local 3.11-vs-3.12 stub artifact), `make tests` (4787 passed), `make coverage` 89% (gate 85%; `src/agents/__init__.py` at 100%).
- Sanity: `uv run python -c "from agents import get_trace_provider, TraceProvider, set_trace_provider; print('ok')"`.

### Issue number

N/A (no tracker issue; same class of fix as the merged #3489 / #3490).

### Checks


seratch:
Thanks for the suggestion. However, we don't intend to export everything from the top-level module. Only the most commonly used ones are from `agents`. I don't think using `get_trace_provider` is very common for most of the users.

PR #3722, opened 2026-07-01
feat(realtime): add typed create_response helper to RealtimeSession

Skyline-9:
### Summary

Adds a typed `RealtimeSession.create_response(*, instructions=None, metadata=None)` helper so applications can request a model response with per-response overrides without hand-rolling a raw `response.create` message.

Today the only way to trigger a response with custom `instructions`/`metadata` is to send a raw protocol event through the low-level model escape hatch, which bypasses the SDK's response-create sequencing and is easy to get wrong (double `response.create`, "conversation already has an active response" errors). This adds a small, typed, well-sequenced entry point that mirrors the ergonomics of the rest of the realtime API (`send_message`, `interrupt`).

What changed:

- New `RealtimeSession.create_response(*, instructions, metadata)` (keyword-only, appended after `interrupt()` so no existing API is reordered).
- New `RealtimeModelSendResponseCreate` dataclass added to the `RealtimeModelSendEvent` union (optional fields, appended at the end of the union).
- `OpenAIRealtimeWebSocketModel.send_event` routes it through the **existing** response-create sequencer (`_reserve_response_create_request(manual=True)` + `_start_response_create(..., manual=True)` -> `wait_for_response_create_slot`), so it cannot collide with an in-flight response and never emits a double `response.create`.
- Only real protocol options are exposed (`instructions`, `metadata`). With no arguments it serializes to a bare `response.create` (no `response` object), via `model_dump_json(exclude_unset=True)`.
- Exported from `agents.realtime.__init__` (`__all__`), consistent with `RealtimeModelSendInterrupt`; not added to the top-level `agents` namespace.
- `docs/realtime/guide.md` documents that `instructions`/`metadata` are per-response overrides that do not mutate the agent.

### Test plan

- Added tests in `tests/realtime/test_session.py`, `tests/realtime/test_openai_realtime.py`, and `tests/realtime/test_openai_realtime_conversions.py`:
  - `create_response` defers while a response is active and sends exactly one `response.create` after the active response completes (sequencer cooperation).
  - No-arg call emits a minimal bare `response.create`; only-`instructions` omits `metadata`; conversion includes only provided fields.
  - Export regression test for `RealtimeModelSendResponseCreate`.
- Ran the standard verification stack from the repo root:
  - `make format`, `make lint`, `make typecheck` — pass
  - `uv run pytest -k realtime` — pass (307 passed, 1 skipped)
  - `make build-docs` — pass

### Issue number

N/A. Fits the broader direction of making the realtime API a typed first-class surface rather than requiring raw protocol messages. Happy to link a tracking issue if preferred.

### Checks

- [x] I've confirmed all verification steps pass (ran `make format`, `make lint`, `make typecheck`, realtime tests, and `make build-docs`)

### Compatibility notes

Additive. Adding `RealtimeModelSendResponseCreate` to the `RealtimeModelSendEvent` union means a third-party `RealtimeModel.send_event` implementation gains a new union member it does not yet handle; the built-in OpenAI WebSocket/SIP models handle it. The dataclass fields are optional and appended at the end of the union, and the public method is keyword-only and appended after existing methods.


Skyline-9:
@codex review

seratch:
Thanks for sharing this. I traced the underlying use case through the existing raw response.create path, and that path is already public, documented, converted through the typed OpenAI realtime event adapter, and routed through the SDK response-create sequencer. It also supports per-response fields through other_data, including response.instructions and response.metadata.

Because the PR does not show a concrete supported workflow that the existing path cannot handle, this looks like an ergonomics improvement rather than a demonstrated capability gap. Adding a new RealtimeSession method and a new RealtimeModelSendEvent variant creates a permanent contract for custom realtime model implementations, which is not justified by the current evidence.

I am going to close this PR for now. If you can provide a specific scenario where the existing raw path is insufficient, we can revisit the narrowest design from that evidence.

PR #3724, opened 2026-07-01
fix(handoffs): enforce strict Pydantic validation when strict_json_schema=True

### LINKED ISSUE #3723: Handoff silently ignores strict_json_schema and coerces types
**Bug Description**
The `Handoff` class explicitly sets `strict_json_schema=True` (its default behavior) which instructs the LLM to adhere to a strict schema for the output type. However, during runtime validation in `_invoke_handoff`, the `validate_json` utility is called without `strict=True`. As a result, Pydantic defaults to its lenient mode and silently coerces invalid types (e.g., coercing the string `"25"` to the integer `25`), entirely ignoring the strictness intent of the handoff definition.

**Steps to Reproduce**
Calling `_json.validate_json` directly (which replicates the exact behavior seen in `Handoff.on_invoke_handoff`) shows the coercion:

```python
from pydantic import BaseModel, TypeAdapter
from agents.util._json import validate_json

class HandoffInput(BaseModel):
    name: str
    age: int  # strict: must be int, not str

type_adapter = TypeAdapter(HandoffInput)

# age is a STRING, not an int
malformed_json = '{"name": "Alice", "age": "25"}'

# This silently coerces "25" to 25 and succeeds, despite age requiring an int
result = validate_json(malformed_json, type_adapter, partial=False)
```

**Expected Behavior**
Since `strict_json_schema=True` is set on the handoff, the validation step should also enforce strict mode, raising a `ValidationError` when a string is provided for an integer field.

I have a fix ready for this and will open a PR shortly!

*(Note: AI-assisted contribution)*

Om-Borse26:
## Summary
Fixes #3723

The `Handoff` class explicitly sets `strict_json_schema=True` (its default behavior) which instructs the LLM to adhere to a strict schema for the output type. However, during runtime validation in `_invoke_handoff`, the `validate_json` utility is called without `strict=True`. As a result, Pydantic defaults to its lenient mode and silently coerces invalid types, entirely ignoring the strictness intent of the handoff definition.

This PR passes `strict=True` to the validation logic when handling handoff inputs, ensuring that the runtime validation strictness matches the schema provided to the LLM.

### Demonstration

**Before (Silent Coercion)**
When an LLM incorrectly passed `"25"` (string) instead of `25` (int) for an age field requiring an integer, the lenient parsing silently accepted and coerced it:
```
[1] CURRENT BEHAVIOR -- SDK validate_json (no strict=True)
    Input JSON: {"name": "Alice", "age": "25"}
    age is a STRING "25", not an integer 25

    ACCEPTED: validate_json did NOT raise an error
    result.name = 'Alice'  (type: str)
    result.age  = 25  (type: int)
```

**After (ValidationError)**
With `strict=True`, Pydantic correctly rejects the mismatch:
```
[2] EXPECTED BEHAVIOR -- type_adapter.validate_json with strict=True
    Input JSON: {"name": "Alice", "age": "25"}

    REJECTED: ValidationError raised (correct -- strict mode rejects str for int)
       type='int_type', msg='Input should be a valid integer'
       input='25'
```

### Verification
You can verify this fix directly via the new test rather than a raw script:
```bash
uv run pytest tests/test_handoff_tool.py::test_handoff_strict_json_rejects_type_coercion
```
(A secondary test `test_handoff_lenient_json_allows_type_coercion` was also added to ensure backward compatibility for other callers using lenient mode).

*(Note: AI-assisted contribution)*

### DIFF
--- a/src/agents/agent_output.py
+++ b/src/agents/agent_output.py
@@ -137,7 +137,12 @@ class AgentOutputSchema(AgentOutputSchemaBase):
         """Validate a JSON string against the output type. Returns the validated object, or raises
         a `ModelBehaviorError` if the JSON is invalid.
         """
-        validated = _json.validate_json(json_str, self._type_adapter, partial=False)
+        validated = _json.validate_json(
+            json_str,
+            self._type_adapter,
+            partial=False,
+            strict=True if self._strict_json_schema else None,
+        )
         if self._is_wrapped:
             if not isinstance(validated, dict):
                 _error_tracing.attach_error_to_current_span(
--- a/src/agents/handoffs/__init__.py
+++ b/src/agents/handoffs/__init__.py
@@ -292,6 +292,7 @@ def handoff(
                 json_str=input_json,
                 type_adapter=type_adapter,
                 partial=False,
+                strict=True,
             )
             input_func = cast(OnHandoffWithInput[THandoffInput], on_handoff)
             result = input_func(ctx, validated_input)
--- a/src/agents/realtime/handoffs.py
+++ b/src/agents/realtime/handoffs.py
@@ -163,6 +163,7 @@ def realtime_handoff(
                 json_str=input_json,
                 type_adapter=type_adapter,
                 partial=False,
+                strict=True,
             )
             input_func = cast(OnHandoffWithInput[THandoffInput], on_handoff)
             if inspect.iscoroutinefunction(input_func):
--- a/src/agents/util/_json.py
+++ b/src/agents/util/_json.py
@@ -13,12 +13,16 @@ from ._error_tracing import attach_error_to_current_span
 T = TypeVar("T")
 
 
-def validate_json(json_str: str, type_adapter: TypeAdapter[T], partial: bool) -> T:
+def validate_json(
+    json_str: str, type_adapter: TypeAdapter[T], partial: bool, strict: bool = False
+) -> T:
     partial_setting: bool | Literal["off", "on", "trailing-strings"] = (
         "trailing-strings" if partial else False
     )
     try:
-        validated = type_adapter.validate_json(json_str, experimental_allow_partial=partial_setting)
+        validated = type_adapter.validate_json(
+            json_str, experimental_allow_partial=partial_setting, strict=strict
+        )
         return validated
     except ValidationError as e:
         attach_error_to_current_span(
--- a/src/agents/util/_json.py
+++ b/src/agents/util/_json.py
@@ -14,15 +14,16 @@ T = TypeVar("T")
 
 
 def validate_json(
-    json_str: str, type_adapter: TypeAdapter[T], partial: bool, strict: bool = False
+    json_str: str, type_adapter: TypeAdapter[T], partial: bool, strict: bool | None = None
 ) -> T:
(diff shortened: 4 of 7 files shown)


Om-Borse26:
Great catch! You are entirely correct—by passing `strict=False` in the utility wrapper when the caller omitted it, I was inadvertently overriding Pydantic's model-level `ConfigDict(strict=True)` setting for other callers.

I have updated the `validate_json` signature to use `strict: bool | None = None` and only pass the parameter through to Pydantic if it is not `None`. This preserves Pydantic's default behavior for existing callers, while still allowing the Handoff call site to explicitly enforce `strict=True`.

I've pushed the fix and updated the regression tests to verify that omitting the parameter still preserves Pydantic's default lenient coercion behavior for a model without strict config. Thanks for flagging this subtle regression!

Om-Borse26:
Thanks for the sharp eyes on both of these. Fixed in the latest push.
For the _json.py default — changed it to Optional[bool] = None and only forward strict to Pydantic when it's explicitly set. That way callers that don't pass it won't accidentally override any model-level ConfigDict the user has set.
For the realtime path — added strict=True there as well since realtime_handoff() already forces ensure_strict_json_schema, so the runtime should match what the LLM was told.
Also noticed while doing the sweep that agent_output.py had the same gap — it has self._strict_json_schema available but wasn't passing it through. Folded that in too.

Om-Borse26:
Another great catch! Passing `strict=False` in `agent_output.py` would indeed override model-level `ConfigDict` settings just like it would have in the helper function.

I have updated the logic in `agent_output.py` to pass `strict=True if self._strict_json_schema else None`, ensuring that `None` is passed when `strict_json_schema` is `False`. The other two call sites (`realtime_handoff` and the base `Handoff`) correctly pass a hardcoded `strict=True` because they enforce strict schema generation on their inputs before reaching validation.

Latest push incorporates this fix!

seratch (changes requested):
The underlying handoff issue is valid: the SDK exposes a strict handoff input schema to the model, but the runtime validation path currently allows Pydantic coercion before invoking on_handoff.

Before we can merge this, please add focused regression coverage for the other runtime paths changed by the PR: realtime handoffs should reject the same coercion case, and AgentOutputSchema should cover both strict_json_schema=True and strict_json_schema=False so we do not accidentally override model-level/default Pydantic behavior. Also please keep the scope explicit: either limit the PR to handoffs/realtime handoffs, or keep the output-schema change with tests and leave ordinary function-tool runtime strictness as a separate follow-up.

seratch:
Also, please fix the CI errors

Om-Borse26:
Thanks for the review! Confirmed — I'll add focused regression tests for the realtime handoff path and both `strict_json_schema=True`/`False` cases for `AgentOutputSchema`. Will also fix the CI lint error. Will push the update shortly.

Om-Borse26:
Thanks for the detailed review! I've pushed an update that adds focused regression tests for the realtime handoff path and both strict_json_schema=True/False cases for AgentOutputSchema. Also fixed the CI lint errors. Let me know if anything else needs adjustment.

seratch (approved):

PR #3725, opened 2026-07-01
Add DakeraSession memory backend

ferhimedamine:
### Summary

Adds a `DakeraSession` memory backend that implements the [`Session`](https://openai.github.io/openai-agents-python/ref/memory/session/) protocol on top of a self-hosted [Dakera](https://github.com/dakera-ai/dakera-deploy) memory server (a REST service for persistent agent memory). It follows the existing extension backends (`RedisSession`, `MongoDBSession`, `DaprSession`): a backend class under `agents.extensions.memory`, an optional `dakera` extra, tests, a runnable example, and docs.

Why it's a faithful `Session`:

- Each conversation is stored in a **per-session Dakera namespace** derived from `session_id` (`"{key_prefix}:{session_id}"`), so histories are isolated and the same `session_id` always resolves to the same stored history — `get_items`/`pop_item`/`clear_session` are restart-safe and shareable across workers pointing at the same server.
- Every item carries a monotonically increasing `seq` in its metadata, so ordering is preserved across writers regardless of server return order.
- `from_url(...)` creates and owns an `AsyncDakeraClient` (closed by `close()`); you can also inject a client your app already manages.

Usage:

```python
from agents.extensions.memory import DakeraSession

session = DakeraSession.from_url(
    session_id="user-123",
    base_url="http://localhost:3000",
    api_key="dk-...",
)
result = await Runner.run(Agent(name="Assistant"), "Hello", session=session)
await session.close()
```

Install: `pip install "openai-agents[dakera]"`. Run a local server with the `dakera-ai/dakera-deploy` docker-compose stack (server + MinIO), which listens on port 3000.

### Test plan

- `tests/extensions/memory/test_dakera_session.py` — full suite (add/get roundtrip, chronological ordering across add calls, `limit` handling and `SessionSettings` default, `pop_item` LIFO + empty, `clear_session`, session isolation, custom key prefix, unicode/complex-item fidelity, corrupted-entry skipping, client lifecycle, and end-to-end `Runner` integration). It uses an in-memory fake Dakera client, so it runs with **no network or live server**.
- `tests/extensions/memory/test_memory_imports.py` — extended to assert the friendly optional-extra error when `dakera` is missing (and the helper now supports top-level optional modules).
- Verified locally with the repo's tooling: `ruff format --check` and `ruff check` (pinned `ruff==0.9.2`), `mypy` (repo config), and `pytest` for the new/affected tests — all green.

### Issue number

N/A — new integration contribution.

### Checks

- [x] I've run `make format-check`, `make lint`, `make mypy`, and the new tests

### DIFF
--- a/docs/ref/extensions/memory/dakera_session.md
+++ b/docs/ref/extensions/memory/dakera_session.md
@@ -1,0 +1,3 @@
+# `DakeraSession`
+
+::: agents.extensions.memory.dakera_session.DakeraSession
--- a/docs/sessions/index.md
+++ b/docs/sessions/index.md
@@ -208,6 +208,7 @@ Use this table to pick a starting point before reading the detailed examples bel
 | `SQLAlchemySession` | Production apps with existing databases | Works with SQLAlchemy-supported databases |
 | `MongoDBSession` | Apps already using MongoDB or needing multi-process storage | Async pymongo; atomic sequence counter for ordering |
 | `DaprSession` | Cloud-native deployments with Dapr sidecars | Supports multiple state stores plus TTL and consistency controls |
+| `DakeraSession` | Self-hosted persistent memory across sessions and workers | Backed by a [Dakera](https://github.com/dakera-ai/dakera-deploy) server over REST |
 | `OpenAIConversationsSession` | Server-managed storage in OpenAI | OpenAI Conversations API-backed history |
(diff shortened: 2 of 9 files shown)


seratch:
Thanks for your interest. While we won't add this extension as part of this core SDK at this moment, please feel free to publish as your own package. Thanks again for sharing this!

PR #3735, opened 2026-07-06
feat(sandbox): add Aliyun AgentRun sandbox client

iridescentWen:
### Summary

Adds an Aliyun (AgentRun) sandbox backend for the Agents SDK sandbox extension,
built on the async `agentrun-sdk` (`agentrun-sdk==0.0.35`, behind a new optional
`aliyun` extra). It mirrors the existing Vercel backend lifecycle
(`create` / `resume` / `delete` / `SandboxSession` context manager) so callers can
swap backends without code changes.

Highlights:
- Async AgentRun API throughout (`create_sandbox_async`, `cmd_async`,
  `download_async`/`upload_async`, `check_health_async`, `delete_async`,
  `get_sandbox_async`) — no blocking `to_thread` wrappers.
- After create, waits for the sandbox to report healthy before first use (mirrors
  the SDK's own context manager); deletes it if it never becomes healthy.
- Gateway HTTP 502 retry with backoff, and a `resolve-workspace-path`
  empty-stdout retry on the startup path.
- Environment is injected the same way the other backends do it (per exec): since
  AgentRun's cmd API has no env parameter, the resolved options + manifest
  environment is inlined as a `KEY=val ...` prefix on each command (unsafe names
  dropped, values `shlex.quote`d, secrets kept out of logs). No workspace file, so
  it works for shell=False, survives snapshot restore, and never lands in snapshots.
- `read` / `write` / `append`, a 100MB upload guard, and OSS / NAS / PolarFs mount
  configs passed through to `create_sandbox_async`.
- `resume()` reattaches an existing sandbox via `get_sandbox_async`, health-checks
  it, and falls back to a fresh sandbox + snapshot restore when it's gone/unhealthy.
- Exec timeouts are forwarded as-is (rounded up), not silently clamped.
- Credentials come from `AliyunSandboxClientOptions` (per-create) or the
  `AliyunSandboxClient` (used for resume) or the SDK's own providers; serialized
  session state never carries secrets. Archives reject external symlinks before
  remote extraction.
- Exposed under `agents.extensions.sandbox` behind the optional `aliyun` extra
  (lazy import guarded), with an example runner and reference docs.

### Test plan

- New unit suite `tests/extensions/sandbox/test_aliyun.py` (88 cases) mocks the
  `agentrun-sdk` via `sys.modules` injection, so it needs no real package and never
  touches Alibaba Cloud. Covers user-arg rejection, exec normalization/timeouts,
  502 retry, file IO + `append`, tar validation (incl. external-symlink rejection),
  health-based `running()`/create wait, shutdown/delete, workspace
  `persist`/`hydrate`, per-exec env injection (merge/quoting/unsafe-name skipping/
  metachar safety/shell=False/log redaction), credential threading, and resume
  reattach + fallback. Backend module coverage is 92%.
- `make format`, `make lint`, `make typecheck` (mypy + pyright, 0 errors), and
  `make tests` (4936 passed, 1 skipped) all pass locally.

### Issue number

N/A

### Checks

### DIFF
--- a/docs/ref/extensions/sandbox/aliyun/sandbox.md
+++ b/docs/ref/extensions/sandbox/aliyun/sandbox.md
@@ -1,0 +1,3 @@
+# `Sandbox`
+
+::: agents.extensions.sandbox.aliyun.sandbox
--- a/docs/sandbox/clients.md
+++ b/docs/sandbox/clients.md
@@ -90,6 +90,7 @@ For provider-specific setup notes and links for the checked-in extension example
 
 | Client | Install | Example |
 | --- | --- | --- |
+| `AliyunSandboxClient` | `openai-agents[aliyun]` | [Aliyun runner](https://github.com/openai/openai-agents-python/blob/main/examples/sandbox/extensions/aliyun_runner.py) |
 | `BlaxelSandboxClient` | `openai-agents[blaxel]` | [Blaxel runner](https://github.com/openai/openai-agents-python/blob/main/examples/sandbox/extensions/blaxel_runner.py) |
 | `CloudflareSandboxClient` | `openai-agents[cloudflare]` | [Cloudflare runner](https://github.com/openai/openai-agents-python/blob/main/examples/sandbox/extensions/cloudflare_runner.py) |
 | `DaytonaSandboxClient` | `openai-agents[daytona]` | [Daytona runner](https://github.com/openai/openai-agents-python/blob/main/examples/sandbox/extensions/daytona/daytona_runner.py) |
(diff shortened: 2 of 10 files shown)


iridescentWen:
@codex review

seratch:
Thanks for sharing this idea several times. We don't plan to have this module as part of this core SDK for now, so please consider sharing the code as your own repo/package instead. Thanks again for your interest here.

iridescentWen:
> Thanks for sharing this idea several times. We don't plan to have this module as part of this core SDK for now, so please consider sharing the code as your own repo/package instead. Thanks again for your interest here.

thank you for your response, I want to be a contributer, even if it`s not about the sanbox, do you have any recommend tasks for me?

PR #3744, opened 2026-07-07
fix(visualization): keep fillcolor on handoff nodes by merging style attributes

comp:
### Summary

Handoff and sub-agent nodes in the agent visualization are meant to render as filled `lightyellow` boxes (matching the root agent node), but they currently render as empty/unfilled boxes.

The node attribute list sets `style` twice: `style=filled, style=rounded`. In the DOT language a repeated attribute takes the **last** value, so `style` resolves to `rounded` only and `filled` is dropped. Because `fillcolor` only takes effect when `filled` is present in `style`, the `fillcolor=lightyellow` was silently ignored.

This merges the two into a single valid value, `style="filled,rounded"`, so the nodes are both filled and rounded as intended. Applies to both the `Handoff` and `Agent` handoff branches in `get_all_nodes`.

### Test plan

Verified against the actual Graphviz renderer (`dot`) by rendering a node both ways to SVG and inspecting the node polygon fill:

- Before (`style=filled, style=rounded`) -> `fill="none"` (unfilled)
- After (`style="filled,rounded"`) -> `fill="lightyellow"` (filled, as intended)

Updated the existing `tests/test_visualization.py` expectations (which asserted the old DOT string) to match the corrected output. Ran the standard checks:

- `uv run ruff format --check` — pass
- `uv run ruff check` — pass
- `uv run mypy src/agents/extensions/visualization.py` — pass
- `uv run pytest tests/test_visualization.py` — 10 passed

### Issue number

N/A

### Checks

- [x] I've added new tests, if relevant (updated existing visualization assertions to cover the corrected output)
- [ ] I've run `.agents/skills/code-change-verification/scripts/run.sh` (ran the equivalent stack manually: ruff format/check, mypy, pytest — see test plan)
- [ ] If using Codex, I've run `/review` before submitting this PR — N/A

### DIFF
--- a/src/agents/extensions/visualization.py
+++ b/src/agents/extensions/visualization.py
@@ -97,7 +97,7 @@ def get_all_nodes(
             name = _escape_label(handoff.agent_name)
             parts.append(
                 f'"{name}" [label="{name}", '
-                f"shape=box, style=filled, style=rounded, "
+                f'shape=box, style="filled,rounded", '
                 f"fillcolor=lightyellow, width=1.5, height=0.8];"
             )
         if isinstance(handoff, Agent):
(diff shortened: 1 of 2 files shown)


seratch (approved):
Thanks. This is indeed a nice improvement 👍

PR #3749, opened 2026-07-07
fix: match pending nested agent-as-tool runs by call_id on resume

tao-hpu:
## Bug

`_restore_pending_nested_agent_tool_runs` pairs the **unfiltered** serialized entries (`processed_response_data["functions"]`) against the **filtered** deserialized runs with a positional `zip(..., strict=False)`. `_deserialize_actions` drops an entry when its tool no longer resolves (`if not tool: continue`) or its tool call fails to parse, so a single dropped entry shifts every later pairing.

Trigger: resume a paused run where a function whose tool the resuming agent no longer exposes precedes an agent-as-tool call that carries `agent_run_state` with a pending nested interruption. The pending state then binds to the wrong tool call, or (at the tail) is silently lost, so the nested approval never re-surfaces.

## Fix

Build a `{tool_call.call_id: run}` lookup from the deserialized runs and match each serialized entry by its `tool_call.call_id` instead of by position. Behavior for the aligned case is unchanged.

## Tests

Added `test_restore_pending_nested_run_when_earlier_entry_dropped`: a serialized `functions` list whose first entry references a removed tool, followed by an agent-as-tool entry with a real `agent_run_state` (built via the existing HITL helpers) carrying a pending approval. It fails on the old positional zip (the pending nested run is lost) and passes with this change. This restore path previously had no test coverage.

`pytest tests/test_run_state.py tests/test_agent_as_tool.py tests/test_agent_tool_state.py`: 247 passed. `ruff check`, `ruff format --check`, and `mypy` on the touched files: clean.

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -1648,6 +1648,26 @@ def _build_handoffs_map(current_agent: Agent[Any]) -> dict[str, Handoff[Any, Age
     return handoffs_map
 
 
+def _serialized_tool_call_signature(tool_call_data: Mapping[str, Any]) -> tuple[Any, ...]:
+    """Rebuild the ephemeral cache's tool-call signature from serialized tool-call data.
+
+    The signature must mirror the field order used by the agent-tool-state cache
+    (``call_id``, ``name``, ``arguments``, ``type``, ``id``, ``status``) so a serialized
+    entry can be bound to the deserialized run that carries the same call.
+    """
+    call_id = tool_call_data.get("call_id")
+    if call_id is None:
+        call_id = tool_call_data.get("callId")
+    return (
+        call_id,
+        tool_call_data.get("name"),
+        tool_call_data.get("arguments"),
+        tool_call_data.get("type"),
+        tool_call_data.get("id"),
+        tool_call_data.get("status"),
+    )
+
+
 async def _restore_pending_nested_agent_tool_runs(
     *,
     current_agent: Agent[Any],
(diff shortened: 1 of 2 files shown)


tao-hpu:
@codex review

seratch:
Thank you for identifying this issue and for providing a concrete implementation and tests. Your key observation was exactly right: `_deserialize_actions()` may filter entries, while `_restore_pending_nested_agent_tool_runs()` previously relied on positionally zipping the original entries with the filtered actions. That insight directly informed the fix we are taking.

We decided to proceed with a maintainer-side implementation https://github.com/openai/openai-agents-python/pull/3753 that preserves the association at the point of deserialization. Each successfully deserialized function action carries its normalized nested run-state data, rather than reconstructing the association after the original and filtered lists have diverged.

We prefer this approach because it:

- removes the positional assumption at its source;
- does not duplicate the ephemeral cache's private signature rules;
- avoids signature reconstruction and `call_id`/FIFO fallback heuristics;
- keeps the relationship explicit and strongly typed;
- attaches restoration metadata only to function actions that need it; and
- naturally handles missing, disabled, and malformed earlier entries, including mixtures of multiple skipped and pending calls.

We also expanded the regression coverage around independent cache scopes, namespaced and deferred tools, multiple pending states, JSON round-trips, and both approval and rejection after restoration.

The proposed signature-based implementation is thoughtful and works for the cases it covers, but it repairs the association after deserialization. The maintainer implementation instead preserves the original association through the filtering boundary, which we believe is simpler and less likely to drift as tool-call identity or cache behavior evolves.

Given the overlap, we are going to proceed with the maintainer implementation rather than merge this PR. We genuinely appreciate the report, analysis, and implementation idea—they were instrumental in getting this fixed.

PR #3753, opened 2026-07-08
fix: preserve nested tool state during restoration

seratch:
This pull request fixes nested agent-tool state restoration when earlier serialized function entries are skipped because their tools are disabled, removed, or no longer parseable.

Function actions are restored through a dedicated typed intermediate carrying only their normalized nested run-state data. This removes positional coupling without retaining raw serialized entries or wrapping unrelated action types. Regression coverage includes multiple skipped and pending calls, scoped independent restorations, and both approval and rejection after JSON round-trips.

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -112,10 +112,12 @@ if TYPE_CHECKING:
     from .run_internal.run_steps import (
         NextStepInterruption,
         ProcessedResponse,
+        ToolRunFunction,
     )
 
 TContext = TypeVar("TContext", default=Any)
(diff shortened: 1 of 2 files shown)


tao-hpu:
Thanks for taking this forward here — the source-level approach that preserves the association at deserialization is cleaner than my after-the-fact matching, and the expanded regression coverage is great to see.

No pressure at all, but if it's easy: for traceability it'd be nice to link #3749 from this PR and add a co-author trailer on the merge commit. Totally fine if not — happy either way, and thanks for the thorough follow-up.

```
Co-authored-by: Tao An <1250043+tao-hpu@users.noreply.github.com>
```

PR #3766, opened 2026-07-09
fix: make AdvancedSQLiteSession.pop_item branch-aware

winklemad:
## What

`AdvancedSQLiteSession` tracks branches (`_current_branch_id`, a `message_structure` table) and its `get_items` is branch-scoped, but it does **not** override `pop_item`. The inherited `SQLiteSession.pop_item` deletes the newest row for the whole session:

```python
DELETE FROM {messages} WHERE id = (
    SELECT id FROM {messages} WHERE session_id = ? ORDER BY id DESC LIMIT 1
)
```

There is no branch filter, so on a non-`main` branch it pops an item that belongs to another branch (usually `main`) and **silently corrupts it** — `pop_item` returns the wrong item and the other branch loses its tail. This is reachable via the runner's session-rewind path, which verifies the tail with branch-aware `get_items` and then removes it with `pop_item`.

### Reproduction

```python
# main = [u1, a1, u2, a2]; branch from turn 2 -> current branch = [u1, a1]
popped = await session.pop_item()          # returns "a2" (main's tail), not "a1"
await session.get_items(branch_id="main")  # main is now [u1, a1, u2] -- corrupted
```

## Fix

Override `pop_item` in `AdvancedSQLiteSession` to scope the pop to the current branch (join `message_structure`, filter by `branch_id`, order by the branch's `sequence_number`). Because branches **share** message rows and foreign keys are not enforced (so `ON DELETE CASCADE` does not run), it drops the branch's newest `message_structure` row and deletes the underlying message only when no other branch still references it. Flat (non-branched) sessions behave exactly as before.

## Tests

Adds `test_pop_item_is_branch_scoped`: pops on a branch and asserts it returns the branch tail, leaves `main` untouched, still pops normally on `main`, and returns `None` on an empty branch. It fails on the current code (returns `main`'s tail and corrupts `main`) and passes with the fix. `ruff` and `mypy` are clean; the full `AdvancedSQLiteSession` test file passes (46).

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -266,6 +266,65 @@ class AdvancedSQLiteSession(SQLiteSession):
 
         return await asyncio.to_thread(_get_items_sync)
 
+    async def pop_item(self) -> TResponseInputItem | None:
+        """Remove and return the most recent item from the current branch.
+
+        The base SQLiteSession.pop_item deletes the newest row for the whole
+        session regardless of branch, so on a non-main branch it removes an item
+        from another branch (usually main) and corrupts it. Scope the pop to the
+        active branch instead: drop the branch's newest message_structure row,
+        and delete the underlying message only when no other branch still
+        references it (branches share message rows, and foreign keys are not
+        enforced so ON DELETE CASCADE does not run).
+
+        Returns:
+            The most recent item on the current branch, or None if it is empty.
+        """
+
+        def _pop_item_sync() -> TResponseInputItem | None:
+            with self._locked_connection() as conn:
+                while True:
+                    cursor = conn.execute(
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. The reported bug is real: AdvancedSQLiteSession.get_items() is branch-scoped, while the inherited pop_item() deletes the session-wide tail.

However, #3755 already addresses the same pop_item bug and also covers the metadata and ownership invariants needed for a safe fix, including turn_usage cleanup, a call-time branch snapshot, shared-message ownership, and controlled interleaving tests. This PR still leaves stale turn_usage when a pop empties a turn, and a concurrent branch switch can redirect an in-flight pop because the branch is read inside the worker.

To avoid maintaining competing implementations, I am going to close this PR in favor of #3755. Thank you for the clear reproduction and regression test.

PR #3768, opened 2026-07-09
docs: align FunctionTool context example

william-xue:
## What
- Update the custom `FunctionTool` example to type `on_invoke_tool` with `ToolContext`.
- Keep the snippet aligned with the surrounding text and the `FunctionTool.on_invoke_tool` API.

## Why
The section already says `on_invoke_tool` receives a `ToolContext`, but the example still imported and annotated `RunContextWrapper`, which can be confusing when copying the snippet.

## Test
- `git diff --check`
- Python text check for the updated docs snippet

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -401,7 +401,7 @@ from typing import Any
 
 from pydantic import BaseModel
 
-from agents import RunContextWrapper, FunctionTool
+from agents import FunctionTool, ToolContext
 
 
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sharing this. After tracing the current runtime behavior, RunContextWrapper is intentionally supported here rather than being a stale annotation. ToolContext extends RunContextWrapper, and the invocation path explicitly preserves a plain wrapper when the callback requests that type. Since this example does not use any tool-specific metadata, changing the annotation does not fix a failing or unsupported example; it changes which context object the runtime selects.

The current patch also imports ToolContext from the top-level agents package, where it is not exported. I do not think this change is necessary as proposed, so I am going to close the PR. A narrower wording clarification explaining when to use RunContextWrapper versus ToolContext could be considered separately.

PR #3771, opened 2026-07-09
fix: strip orphaned handoff message only for non-official Responses endpoints

utkarshkr100:
### Summary

Reasoning-enabled handoffs emit `[reasoning, function_call, message]` in a single response. The
reasoning is consumed by the `function_call`, so the trailing assistant message has no reasoning of
its own. Strict Responses endpoints (e.g. Azure OpenAI) reject the replayed shape with HTTP 400
(`Item 'msg_...' of type 'message' was provided without its required 'reasoning' item`), while
official OpenAI tolerates it and — per the Responses API guidance — expects items since the last
user message to be passed through untouched.

This scopes the repair to the **Responses model layer**, applied **only when the client is not
official OpenAI** (`is_official_openai_client`). Official OpenAI and the Chat Completions / LiteLLM
paths are untouched (the latter already fold reasoning into the assistant message and never emit
standalone reasoning items). The message also stays in session history — it is only omitted from the
outgoing payload to a strict endpoint, where keeping it would otherwise 400 the entire turn.

This supersedes #3574, which applied the drop globally across every provider/path; that was closed
because it could discard legitimate assistant output beyond the reported Azure case.

### Test plan

- `test_non_official_endpoint_avoids_reasoning_item_400_on_handoff` reproduces the actual Azure 400
  with a strict fake client and proves the fix prevents it end-to-end.
- `test_non_official_endpoint_strips_orphaned_handoff_message` /
  `test_official_endpoint_preserves_orphaned_handoff_message` pin the endpoint scoping.
- Five pure-function unit tests for `drop_orphaned_messages_after_consumed_reasoning` (assistant-only
  drop, user-message preservation, multiple orphans, no-consume, cross-turn reset).
- `make lint`, `make format`, `make typecheck`, `make tests` pass.

### Issue number

Supersedes #3574.

### Checks

- [ ] I've added/updated the relevant documentation (no user-facing docs change)

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -45,6 +45,7 @@ from ..items import ItemHelpers, ModelResponse, TResponseInputItem
 from ..logger import logger
 from ..model_settings import MCPToolChoice
 from ..retry import ModelRetryAdvice, ModelRetryAdviceRequest
+from ..run_internal.items import drop_orphaned_messages_after_consumed_reasoning
 from ..tool import (
     ApplyPatchTool,
     CodeInterpreterTool,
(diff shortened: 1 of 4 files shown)


seratch:
Can you share the repro steps of the error?

PR #3783, opened 2026-07-10
feat: add structured logging helper functions

maslinedwin:
Adds structured logging utilities to logger module:

- debug_with_context() - log debug with context dict
- info_with_context() - log info with context dict
- warning_with_context() - log warning with context dict
- error_with_context() - log error with context dict

Makes it easier for developers to log structured data for debugging and monitoring.

### DIFF
--- a/src/agents/logger.py
+++ b/src/agents/logger.py
@@ -1,3 +1,56 @@
 import logging
+from typing import Any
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks, but we don't need this change for now.

PR #3784, opened 2026-07-10
feat: add context managers for resource management

maslinedwin:
Provides agent_session(), temporary_timeout(), and error_handling_context() managers.


seratch:
Thanks, but we don't need this change for now.

PR #3785, opened 2026-07-10
feat: add configuration builder with fluent API

maslinedwin:
Fluent builder for agent configuration with validation and method chaining.


seratch:
Thanks, but we don't need this change for now. If you need this for your projects, please feel free to publish it as your own package.

PR #3800, opened 2026-07-10
refactor: share retry error metadata extraction

seratch:
This pull request improves retry error normalization by sharing exception-chain traversal and status code, error code, and request ID extraction between OpenAI provider retry advice and runner-level retries.

It preserves the existing retry behavior and adds regression coverage confirming that both paths produce identical metadata. Header and `Retry-After` parsing remain unchanged for a separate follow-up.


seratch:
@codex review

PR #3803, opened 2026-07-10
chore: enable RUF012 for runtime source

seratch:
This pull request improves class-variable typing by enabling Ruff's `RUF012` rule for runtime source.

It marks the MongoDB session's shared initialization registry and lock, along with the Responses stream terminal-event set, as `ClassVar`. Examples and tests remain exempt because they contain intentional mutable fixture and UI class attributes.


seratch:
@codex review

PR #3811, opened 2026-07-12
Reject conflicting OpenAI client args without assert

### LINKED ISSUE #3808: OpenAIProvider skips conflicting client argument validation under python -O
## Summary

`OpenAIProvider.__init__()` uses an `assert` to reject `openai_client` together with `api_key`, `base_url`, or `websocket_base_url`. Python removes assertions when run with `-O`, so the same invalid configuration is accepted in optimized mode and the explicit connection arguments are silently ignored.

This makes public constructor behavior depend on interpreter optimization settings and can cause requests to use a different key or endpoint than the caller configured.

## Reproduction

```python
from agents.models.openai_provider import OpenAIProvider

client = AsyncOpenAI(api_key=client-key)
provider = OpenAIProvider(
    openai_client=client,
    api_key=different-key,
    base_url=https://example.com,
)

resolved_client = provider._get_client()
print(resolved_client.api_key)
print(resolved_client.base_url)
```

Run normally:

```console
$ python repro.py
AssertionError: Don't provide api_key, base_url, or websocket_base_url if you provide openai_client
```

Run with optimization enabled:

```console
$ python -O repro.py
client-key
https://api.openai.com/v1/
```

The conflicting `api_key` and `base_url` are accepted but have no effect.

## Expected behavior

Invalid constructor arguments should be rejected consistently regardless of whether Python runs with `-O`. A regular exception such as `ValueError` or `UserError` would preserve validation in optimized builds.

## Environment

- OS: Windows 11
- Python: 3.12.12
- `openai-agents`: 0.18.2
- Repository commit: `0885933`

## Additional context

The same assertion-based validation pattern appears in `OpenAIVoiceModelProvider`, so optimized mode likely affects that constructor as well.

I searched existing issues for `OpenAIProvider` with `assert`, `python -O`, and conflicting `openai_client`/`base_url` arguments and did not find a matching report.

HUAN2022A:
## Summary
- Replace assertion-based OpenAI client argument validation with stable `UserError` exceptions.
- Apply the same validation to the voice model provider.
- Add regression coverage for conflicting `openai_client` connection arguments.

## Tests
- `uv run pytest -q tests/test_config.py`
- `uv run python -O -c "from openai import AsyncOpenAI; from agents.models.openai_provider import OpenAIProvider; c=AsyncOpenAI(api_key='client-key'); OpenAIProvider(openai_client=c, api_key='different-key', base_url='https://example.com')"` verifies optimized mode still raises `UserError`.

Closes #3808

### DIFF
--- a/src/agents/models/openai_provider.py
+++ b/src/agents/models/openai_provider.py
@@ -7,6 +7,7 @@ import weakref
 import httpx
 from openai import AsyncOpenAI, DefaultAsyncHttpxClient
 
+from ..exceptions import UserError
 from . import _openai_shared
 from .default_models import get_default_model
 from .interface import Model, ModelProvider
@@ -86,10 +87,11 @@ class OpenAIProvider(ModelProvider):
                 chunk semantics are not reliable enough for incremental processing.
         """
         if openai_client is not None:
-            assert api_key is None and base_url is None and websocket_base_url is None, (
-                "Don't provide api_key, base_url, or websocket_base_url if you provide "
-                "openai_client"
-            )
+            if api_key is not None or base_url is not None or websocket_base_url is not None:
+                raise UserError(
+                    "Don't provide api_key, base_url, or websocket_base_url if you provide "
+                    "openai_client"
+                )
             self._client: AsyncOpenAI | None = openai_client
         else:
             self._client = None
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for your interest here. Closing yours in favor of https://github.com/openai/openai-agents-python/pull/3809

PR #3812, opened 2026-07-12
feat(realtime): track response usage in session context

### LINKED ISSUE #3810: Track Realtime response usage in session context
# Add usage accounting for Realtime responses

### Describe the feature

`RealtimeSession` exposes a `RunContextWrapper` through event info, but its `usage` stays empty for model responses. The Realtime API includes per-response usage in `response.done`, while `OpenAIRealtimeWebSocketModel` currently emits only `RealtimeModelTurnEndedEvent` for that server event.

I propose adding a vendor-neutral Realtime model usage event and using it to update the session context once for every completed response. The OpenAI model adapter would translate `response.usage` into the SDK's existing aggregate fields (`requests`, input, output, total, and cached input tokens), while retaining Realtime-specific text/audio/image detail on the raw model event rather than forcing those fields into the narrower shared `Usage` type.

The intended behavior would be:

- each `response.done` with usage emits one model usage event before `turn_ended`;
- `event.info.context.usage` reflects cumulative Realtime usage as events are consumed;
- cancelled, failed, and incomplete responses are counted when the server reports usage;
- missing or partial usage fields remain valid and default safely;
- custom `RealtimeModel` implementations can report usage without importing OpenAI types;
- the new event and detail types are exported from `agents.realtime`.

This would make usage and cost accounting available to long-running voice sessions without requiring consumers to parse OpenAI-specific raw server events. I can implement the model event, OpenAI conversion, session aggregation, public exports, and focused tests if this API direction matches the maintainers' expectations.

seratch:
This pull request adds Realtime response usage accounting and resolves #3810.

It introduces `RealtimeModelUsageEvent`, converts usage from OpenAI `response.done` events before `turn_ended`, and accumulates response request counts, token totals, cached input tokens, and per-request entries in `RunContextWrapper.usage`. Present but partial usage fields are normalized safely, missing usage remains a no-op, and completed, cancelled, failed, and incomplete responses follow the same accounting path.

Realtime-specific modality details remain available on raw model events. Input audio transcription remains a separate billing and usage path and is not included in these response totals.


Skyline-9:
On the design of `RealtimeModelUsageEvent`, audio tokens are the primary cost driver for voice applications, so discarding the text/audio token breakdown before emitting the typed usage event means callers fall back on manually parsing the raw events. Is there a way to keep this typed with the modality details?

seratch:
@Skyline-9 Thanks for pointing it out. I've added the usage details to the data structure.

PR #3818, opened 2026-07-12
fix(memory): validate table names to prevent SQL injection in SQLiteSession

### LINKED ISSUE #3816: SQL Injection via Unsanitized F-String Table Name Interpolation in SQLiteSession / AdvancedSQLiteSession
### Please read this first

- **Have you searched for related issues?** Yes, no existing issue covers f-string table name interpolation in SQLiteSession/AdvancedSQLiteSession.

### Describe the bug

The `SQLiteSession` and `AdvancedSQLiteSession` classes construct SQL queries using Python f-string interpolation for table names (`sessions_table`, `messages_table`) without any sanitization or validation. These table names are accepted as constructor parameters with safe defaults (`"agent_sessions"`, `"agent_messages"`), but no input validation prevents a caller from passing arbitrary strings.

**Root cause chain:**

1. `SQLiteSession.__init__` (`src/agents/memory/sqlite_session.py`) accepts `sessions_table` and `messages_table` as keyword parameters — **no regex validation, no identifier quoting**:

```python
def __init__(
    self,
    session_id: str,
    db_path: str | Path = ":memory:",
    sessions_table: str = "agent_sessions",   # accepted without validation
    messages_table: str = "agent_messages",   # accepted without validation
    session_settings: SessionSettings | None = None,
):
    self.sessions_table = sessions_table
    self.messages_table = messages_table
```

2. `AdvancedSQLiteSession.__init__` (`src/agents/extensions/memory/advanced_sqlite_session.py`) accepts `**kwargs` and forwards them to `super().__init__()`, making table names indirectly settable through kwargs.

3. **10+ methods** in `advanced_sqlite_session.py` use f-string interpolation (not parameterized bindings) for these table names in raw SQL:

```python
# _init_structure_tables()
cursor.execute(f"""
    CREATE TABLE IF NOT EXISTS message_structure (
        ...
        FOREIGN KEY (session_id)
            REFERENCES {self.sessions_table}(session_id) ON DELETE CASCADE,
        FOREIGN KEY (message_id)
            REFERENCES {self.messages_table}(id) ON DELETE CASCADE
    )
""")

# _insert_structure_metadata()
cursor.execute(
    f"SELECT id FROM {self.messages_table} "
    f"WHERE session_id = ? ORDER BY id DESC LIMIT ?",
    (self.session_id, len(items)),
)

# _cleanup_orphaned_messages_sync()
cursor.execute(
    f"DELETE FROM {self.messages_table} WHERE session_id = ? AND id NOT IN (...)"
)

# get_items(), get_conversation_turns(), find_turns_by_content(),
# create_branch_from_turn() — all affected
```

By contrast, `session_id`, `branch_id`, `turn_number` etc. are correctly passed as `?` parameters — the table names are the only unparameterized identifiers.

### Debug information

- **Agents SDK version**: `v0.18.2`
- **Python version**: Python 3.10+
- **Affected files**:
  - `src/agents/memory/sqlite_session.py` (lines 30–65, `__init__`)
  - `src/agents/extensions/memory/advanced_sqlite_session.py` (all methods using `self.sessions_table` or `self.messages_table` in SQL: `_init_structure_tables`, `_insert_structure_metadata`, `_cleanup_orphaned_messages_sync`, `get_items`, `get_conversation_turns`, `find_turns_by_content`, `create_branch_from_turn`, and others)

### Repro steps

```python
"""Minimal reproduction: SQL injection via malicious table name."""
import sqlite3

# Step 1: Create a legitimate table first so the attack is visible
legit_db = sqlite3.connect(":memory:")
legit_db.execute("CREATE TABLE evil (payload TEXT)")
legit_db.execute("INSERT INTO evil VALUES ('SQL_INJECTION_WORKS')")
legit_db.commit()

# Step 2: Pass a malicious table name through kwargs
# The injected SQL uses UNION to read from the 'evil' table
malicious_table = "agent_messages UNION SELECT payload FROM evil"

try:
        session_id="test_session",
        db_path=":memory:",
        create_tables=True,
        messages_table=malicious_table,  # passed through **kwargs → super().__init__()
    )
except Exception:
    pass  # DDL injection may raise on malformed SQL

# Step 3: Even without DDL, read-path injection is possible
# If tables already exist, the following would execute injected SQL:
session2 = AdvancedSQLiteSession(
    session_id="test",
    db_path=":memory:",
    messages_table="agent_messages WHERE 1=1--",
)
# cursor.execute(f"SELECT id FROM agent_messages WHERE 1=1-- WHERE session_id = ? ...")
# The WHERE clause after the injection point is commented out
```

### Expected behavior

Table names should be validated against an allowed identifier pattern at construction time. For example:

```python
import re

_IDENTIFIER_RE = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]*$')

def __init__(self, ..., sessions_table: str = "agent_sessions", ...):
    if not _IDENTIFIER_RE.match(sessions_table):
        raise ValueError(
            f"Invalid table name: '{sessions_table}'. "
            f"Must match pattern: {_IDENTIFIER_RE.pattern}"
        )
    if not _IDENTIFIER_RE.match(messages_table):
        raise ValueError(
            f"Invalid table name: '{messages_table}'. "
            f"Must match pattern: {_IDENTIFIER_RE.pattern}"
        )
    self.sessions_table = sessions_table
    self.messages_table = messages_table
```

Alternative approaches:
- Use `sqlite3.complete_statement()` as an additional safety check
- Use a fixed set of allowed table name constants and disallow custom names entirely (simplest, if custom table names are not a required feature)
- Quote identifiers with double-quotes: `f'"{self.messages_table}"'` (SQLite standard identifier quoting), though this alone is not sufficient against all injection vectors

### Additional context

- **CWE**: CWE-89 (SQL Injection)
- **Default values are safe**: `"agent_sessions"` and `"agent_messages"` are hardcoded safe strings; the vulnerability only triggers when a caller overrides them with unsanitized input.
- **Attack surface depends on deployment**: If table names are ever read from configuration files, environment variables, or API parameters, the injection becomes exploitable.
- **Same pattern exists in `async_sqlite_session.py`**: The async variant has the same f-string interpolation pattern with inherited table name attributes.
- **SQLAlchemy variant is NOT affected**: `sqlalchemy_session.py` uses SQLAlchemy's ORM layer which escapes identifiers.
- **Discovered via**: ChunMiao automated code security audit (hybrid LLM + static analysis).

anishesg:
The `SQLiteSession` and `AdvancedSQLiteSession` classes accepted table names as constructor parameters without validation, then interpolated them directly into SQL queries via f-strings. This allowed SQL injection if a caller passed malicious identifiers through the `sessions_table` or `messages_table` parameters. While data values like `session_id` and `branch_id` were correctly parameterized with `?`, table names were the only unparameterized identifiers across 10+ methods.

The fix adds a `_validate_table_name()` function in `src/agents/memory/sqlite_session.py` that validates table names against a strict pattern: alphanumeric characters and underscores only, must start with a letter or underscore, and rejects SQL keywords. Validation is applied in `SQLiteSession.__init__()` before assigning the table name attributes. Since `AdvancedSQLiteSession` forwards kwargs to the parent constructor, validation applies automatically to both classes.

Added regression tests covering malicious injection attempts (semicolons, UNION clauses, SQL keywords) as well as valid table name patterns to ensure legitimate use cases continue to work.

Fixes #3816

### DIFF
--- a/src/agents/memory/sqlite_session.py
+++ b/src/agents/memory/sqlite_session.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import re
 import sqlite3
 import threading
 from collections.abc import Iterator
@@ -12,6 +13,53 @@ from typing import ClassVar
 from ..items import TResponseInputItem
 from .session import SessionABC
 from .session_settings import SessionSettings, resolve_session_limit
+
+# SQL keywords that are not allowed as table names
+_SQL_KEYWORDS = frozenset(
+    {
+        "SELECT",
+        "INSERT",
+        "UPDATE",
+        "DELETE",
+        "DROP",
+        "CREATE",
+        "ALTER",
+        "TABLE",
+        "FROM",
+        "WHERE",
+        "JOIN",
+        "UNION",
+        "ORDER",
+        "GROUP",
+        "HAVING",
+        "LIMIT",
+        "OFFSET",
+    }
+)
+
+
+def _validate_table_name(table_name: str) -> None:
+    """Validate that a table name is a safe SQL identifier.
+
+    Args:
+        table_name: The table name to validate
+
+    Raises:
+        ValueError: If the table name is invalid
+    """
+    if not table_name:
+        raise ValueError("Table name cannot be empty")
+
+    # Check format: alphanumeric + underscore, must start with letter or underscore
+    if not re.match(r"^[a-zA-Z_][a-zA-Z0-9_]*$", table_name):
+        raise ValueError(
+            f"Invalid table name '{table_name}': must contain only alphanumeric "
+            "characters and underscores, and start with a letter or underscore"
+        )
+
+    # Check for SQL keywords
+    if table_name.upper() in _SQL_KEYWORDS:
+        raise ValueError(f"Invalid table name '{table_name}': SQL keyword not allowed")
 
 
 class SQLiteSession(SessionABC):
@@ -46,6 +94,10 @@ class SQLiteSession(SessionABC):
             session_settings: Session configuration settings including default limit for
                 retrieving items. If None, uses default SessionSettings().
         """
+        # Validate table names to prevent SQL injection
+        _validate_table_name(sessions_table)
+        _validate_table_name(messages_table)
+
         self.session_id = session_id
         self.session_settings = session_settings or SessionSettings()
         self.db_path = db_path
--- a/tests/extensions/memory/test_advanced_sqlite_session.py
+++ b/tests/extensions/memory/test_advanced_sqlite_session.py
@@ -1815,3 +1815,84 @@ async def test_output_tokens_details_persisted_when_input_details_missing():
     assert turn_usage["output_tokens_details"] == {"reasoning_tokens": 42}
     assert turn_usage["input_tokens_details"] is None
     session.close()
+
+
+async def test_table_name_validation_rejects_malicious_input():
+    """Regression test for SQL injection via malicious table names."""
+    # Test malicious table name with SQL injection attempt
+    with pytest.raises(ValueError, match="Invalid table name"):
+        AdvancedSQLiteSession(
+            session_id="injection_test",
+            messages_table="agent_messages; DROP TABLE agent_sessions--",
+        )
+
+    # Test malicious table name with UNION injection
+    with pytest.raises(ValueError, match="Invalid table name"):
+        AdvancedSQLiteSession(
+            session_id="injection_test",
+            messages_table="agent_messages UNION SELECT * FROM users",
+        )
+
+    # Test SQL keyword as table name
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for investigating this. The table names are interpolated into SQL, but the linked issue does not demonstrate an exploitable supported path: its in-memory databases are separate, and the proposed read payload fails during table creation before any read query is reached. These names are caller-supplied schema configuration, so applications can already keep them static or validate any external configuration before constructing the session.

This PR also introduces a restrictive, incomplete keyword policy while leaving AsyncSQLiteSession unchanged. I am going to close it for now. We can reconsider a shared identifier-quoting change if there is a concrete deployment reproduction showing untrusted input reaching these parameters and causing an observable unintended database operation.

PR #3819, opened 2026-07-12
fix(memory): eliminate TOCTOU race in AdvancedSQLiteSession structure metadata insertion

### LINKED ISSUE #3817: Potential TOCTOU race condition in _add_structure_metadata / _insert_structure_metadata
### Please read this first

- **Have you searched for related issues?** Yes, no existing issue covers the non-transactional insert+select pattern in `_add_structure_metadata` / `_insert_structure_metadata`.

### Describe the bug

In `advanced_sqlite_session.py`, `_add_structure_metadata()` and `_insert_structure_metadata()` contain a TOCTOU race condition when resolving message IDs after insertion.

**_insert_structure_metadata** (~line 262) queries for recently inserted message IDs using `ORDER BY id DESC LIMIT N` and assumes they correspond to the items passed as arguments:

```python
def _insert_structure_metadata(
    self,
    conn: sqlite3.Connection,
    items: list[TResponseInputItem],
) -> None:
    with closing(conn.cursor()) as cursor:
        cursor.execute(
            f"SELECT id FROM {self.messages_table} "
            f"WHERE session_id = ? ORDER BY id DESC LIMIT ?",
            (self.session_id, len(items)),
        )
        message_ids = [row[0] for row in cursor.fetchall()]
        message_ids.reverse()

    if len(message_ids) != len(items):
        raise RuntimeError(
            "Failed to resolve inserted message IDs while writing structure metadata"
        )
    # ... uses message_ids to insert structure metadata
```

**Two problems:**

1. **`_add_structure_metadata()` is a standalone public async method** — it can be called independently without first inserting items. When called this way, the SELECT retrieves IDs from prior inserts, causing metadata to be associated with wrong messages.

2. **No transaction guard on standalone calls** — while `add_items()` wraps both insert and metadata steps in `_execute_in_transaction()`, `_add_structure_metadata()` itself has no such guard. In SQLite WAL mode with concurrent connections, another session's INSERT could interleave, causing metadata cross-linking.

### Debug information

- **Agents SDK version**: `v0.18.2`
- **Python version**: Python 3.10+
- **Affected files**: `src/agents/extensions/memory/advanced_sqlite_session.py`
  - `_add_structure_metadata()` (~line 233)
  - `_insert_structure_metadata()` (~line 262)

### Repro steps

```python
import asyncio
from openai.types.responses import ResponseFunctionToolCallItem


async def main():
        session_id="test_session",
        db_path=":memory:",
        create_tables=True,
    )

    # Add initial items via add_items (which IS transactional)
    item = ResponseFunctionToolCallItem(
        id="item_1", call_id="call_1",
        name="test_func", arguments="{}",
        status="completed", output="result",
    )
    await session.add_items([item])

    # Call _add_structure_metadata independently without prior insert.
    # It will SELECT the last N IDs from PRIOR inserts and may silently
    # associate wrong metadata to wrong message IDs.
    new_items = [ResponseFunctionToolCallItem(
        id="item_2", call_id="call_2",
        name="another_func", arguments="{}",
        status="completed", output="result2",
    )]

    try:
        await session._add_structure_metadata(new_items, None)
    except RuntimeError:
        pass  # Caught by len check if no prior insert
    except Exception as e:
        print(f"Error: {e}")


asyncio.run(main())
```

### Expected behavior

1. Wrap `_add_structure_metadata()` in `_execute_in_transaction()` so it is always atomic, matching the `add_items()` pattern.

2. Alternatively, use `INSERT ... RETURNING` (SQLite 3.35.0+) to get generated IDs directly without a subsequent SELECT:

```python
cursor.execute(
    f"INSERT INTO {self.messages_table} (...) VALUES (...) RETURNING id"
)
message_ids = [row[0] for row in cursor.fetchall()]
```

3. Or make `_add_structure_metadata()` private/internal to signal it should only be called after items are inserted.

### Additional context

- **CWE**: CWE-362 (Race Condition)
- **Severity**: Low — SQLite default serialized mode mitigates this in single-connection setups. Risk increases with WAL mode + multiple connections.
- **`add_items()` is correctly transactional** — only the standalone `_add_structure_metadata()` path is affected.
- **The count mismatch guard provides partial protection** but cannot detect interleaved concurrent inserts.

anishesg:
The `_insert_structure_metadata()` method in `AdvancedSQLiteSession` contained a TOCTOU race condition when resolving message IDs after insertion. It used `ORDER BY id DESC LIMIT N` to retrieve the last N inserted rows, assuming they corresponded to the items passed as arguments. This approach failed in two scenarios: when `_add_structure_metadata()` was called independently without a preceding insert (retrieving stale IDs from prior operations), and when concurrent connections in SQLite WAL mode interleaved inserts between the INSERT and SELECT operations.

The fix replaces the non-deterministic SELECT query with `cursor.lastrowid`, which captures the auto-generated primary key immediately after each INSERT within the same transaction. The `_insert_items_and_capture_ids()` method now inserts items one by one and builds a list of message IDs that is passed directly to `_insert_structure_metadata()`. This ensures message IDs are deterministically mapped to their corresponding items without timing dependencies.

Additionally, `_add_structure_metadata()` now requires an explicit `message_ids` parameter and raises `ValueError` if called without it, preventing silent data corruption from stale ID reuse. Updated test fixtures to match the new signature.

Fixes #3817

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -135,6 +135,54 @@ class AdvancedSQLiteSession(SQLiteSession):
 
             conn.commit()
 
+    def _insert_items_and_capture_ids(
+        self, conn: sqlite3.Connection, items: list[TResponseInputItem]
+    ) -> list[int]:
+        """Insert items and return their assigned message IDs.
+
+        This method ensures that message IDs are deterministically captured within
+        the same transaction as insertion, eliminating TOCTOU race conditions.
+
+        Args:
+            conn: Database connection
+            items: List of items to insert
+
+        Returns:
+            List of message IDs corresponding to the inserted items, in the same order
+        """
+        # First ensure session exists
+        conn.execute(
+            f"""
+            INSERT OR IGNORE INTO {self.sessions_table} (session_id) VALUES (?)
+        """,
+            (self.session_id,),
+        )
+
+        # Insert items one by one and capture their IDs using last_insert_rowid()
+        message_ids = []
+        with closing(conn.cursor()) as cursor:
+            for item in items:
+                cursor.execute(
+                    f"""
+                    INSERT INTO {self.messages_table} (session_id, message_data) VALUES (?, ?)
+                """,
+                    (self.session_id, json.dumps(item)),
+                )
+                # Capture the ID immediately after insertion
+                message_ids.append(cursor.lastrowid)
+
+        # Update session timestamp
+        conn.execute(
+            f"""
+            UPDATE {self.sessions_table}
+            SET updated_at = CURRENT_TIMESTAMP
+            WHERE session_id = ?
+        """,
+            (self.session_id,),
+        )
+
+        return message_ids
+
     async def add_items(self, items: list[TResponseInputItem]) -> None:
         """Add items to the session.
 
@@ -149,8 +197,8 @@ class AdvancedSQLiteSession(SQLiteSession):
             with self._locked_connection() as conn:
                 try:
                     # Keep both writes in one transaction so metadata failures do not leave orphans.
-                    self._insert_items(conn, items)
-                    self._insert_structure_metadata(conn, items)
+                    message_ids = self._insert_items_and_capture_ids(conn, items)
+                    self._insert_structure_metadata(conn, items, message_ids)
                     conn.commit()
                 except Exception:
                     conn.rollback()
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. I traced this through the released AdvancedSQLiteSession.add_items() path, and the reported interleaving cannot occur there. Message insertion and structure metadata resolution already run on the same connection in one write transaction, and SQLite allows only one simultaneous write transaction even in WAL mode. The SDK also serializes file-backed operations with a shared per-database lock.

The remaining scenario calls _add_structure_metadata() directly, but that is a private helper rather than a supported standalone API, and the linked reproduction does not match its released signature. Since the supported outcome is already covered and this patch adds a second insertion implementation without a base-failing regression test, I am going to close this PR and the linked issue.

PR #3821, opened 2026-07-12
fix(memory): serialize OpenAI conversation session initialization

### LINKED ISSUE #3820: OpenAIConversationsSession can split history during concurrent lazy initialization
### Please read this first

- **Have you read the docs?** Yes, including the sessions documentation.
- **Have you searched for related issues?** Yes. I could not find an existing issue or pull request covering concurrent lazy initialization of `OpenAIConversationsSession`.

### Describe the bug

When two coroutines first use the same `OpenAIConversationsSession` concurrently, both can observe `_session_id is None` and create separate OpenAI conversations.

Each operation then writes to a different conversation ID, while the session retains only the ID assigned last. This splits the session history and makes one of the writes inaccessible through the session.

### Debug information

- Agents SDK version: `v0.18.2`

### Repro steps

The issue can be reproduced offline with a fake client that yields during conversation creation:

```python
import asyncio

class Items:
    def __init__(self):
        self.writes = []

    async def create(self, *, conversation_id, items):
        self.writes.append((conversation_id, items))


class Conversations:
    def __init__(self):
        self.create_calls = 0
        self.items = Items()

    async def create(self, *, items):
        self.create_calls += 1
        call_number = self.create_calls
        await asyncio.sleep(0.01)
        return SimpleNamespace(id=f"conv-{call_number}")


class Client:
    def __init__(self):
        self.conversations = Conversations()


async def main():
    client = Client()
    session = OpenAIConversationsSession(openai_client=client)

    await asyncio.gather(
        session.add_items([{"role": "user", "content": "first"}]),
        session.add_items([{"role": "user", "content": "second"}]),
    )

    print("create calls:", client.conversations.create_calls)
    print("writes:", client.conversations.items.writes)
    print("retained ID:", session.session_id)


asyncio.run(main())
```

Current output:

```text
create calls: 2
writes: [('conv-1', ...), ('conv-2', ...)]
retained ID: conv-2
```

### Expected behavior

Concurrent first operations on one `OpenAIConversationsSession` should share a single lazily created conversation. The conversation should be created once, and both writes should use the retained session ID.

russeell:
### Summary

This pull request fixes a race in `OpenAIConversationsSession` lazy initialization. Concurrent first operations now serialize conversation creation and re-check the session ID inside the lock, preventing one session's history from being split across multiple OpenAI conversation IDs.

It also adds regression coverage for concurrent first writes through `add_items()`, including the failure interleaving where a waiting writer takes over after the first initializer fails.

### Test plan

- `uv run --frozen pytest -q tests/memory/test_openai_conversations_session.py` (`28 passed`)
  - format passed
  - lint passed
  - typecheck passed
  - full test suite passed
- `codex review --uncommitted` (no actionable findings)

### Issue number

Closes #3820

### Checks

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import asyncio
+
 from openai import AsyncOpenAI
 
 from agents.models._openai_shared import get_default_openai_client
@@ -31,6 +33,7 @@ class OpenAIConversationsSession(SessionABC):
         session_settings: SessionSettings | None = None,
     ):
         self._session_id: str | None = conversation_id
+        self._session_id_lock = asyncio.Lock()
         self.session_settings = session_settings or SessionSettings()
         _openai_client = openai_client
         if _openai_client is None:
@@ -64,7 +67,9 @@ class OpenAIConversationsSession(SessionABC):
 
     async def _get_session_id(self) -> str:
         if self._session_id is None:
-            self._session_id = await start_openai_conversations_session(self._openai_client)
+            async with self._session_id_lock:
+                if self._session_id is None:
+                    self._session_id = await start_openai_conversations_session(self._openai_client)
         return self._session_id
 
     async def _clear_session_id(self) -> None:
--- a/tests/memory/test_openai_conversations_session.py
+++ b/tests/memory/test_openai_conversations_session.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying race is valid, and the lock with an in-lock session ID check is the right implementation boundary.

Before merging, please add one controlled failure-interleaving regression test: start conversation creation in task A, start task B while A still owns the initialization lock, fail A's creation, and verify that B acquires the lock, creates a new conversation, and writes successfully to that surviving conversation ID. Please assert A's exception, two creation attempts, B's single write, and the final retained session ID.

The current success-interleaving and sequential retry tests cover the individual behaviors, but they do not lock down the survivor invariant when the first concurrent initializer fails.

russeell:
@seratch Thanks for the review. I added the controlled failure-interleaving regression test in e76aab7. It verifies A's exception, two creation attempts, B's single successful write, and the final retained session ID. The focused suite (28 tests) and full verification pass.

seratch (approved):

Issue #1434, opened 2025-08-13
in history_updated events, the status of 'assistant' is always 'in_progress'. After the assistant has spoken or generated audio, it must say 'Completed'


apiltamang:
### Please read this first

- **Have you read the docs?**[Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug
On the history event that is emitted, the status of messages emitted by 'assistant' always says 'in_progress'. I believe the status must say 'completed' once the audio stream is None.

See example of log:
```[
  {
    "item_id": "item_C3uQ0hZGu3ANx0kuSfadm",
    "previous_item_id": "",
    "type": "message",
    "role": "user",
    "content": [
      {
        "type": "input_text",
        "text": "Greet the caller and ask how you can help!",
        "id": null,
        "audio": null,
        "transcript": null
      }
    ],
    "status": "in_progress"
  },
  {
    "item_id": "item_C3uQ0ycMymMXvPMNEE7K3",
    "previous_item_id": null,
    "type": "message",
    "role": "assistant",
    "status": "in_progress",
    "content": [
      {
        "type": "audio",
        "audio": null,
        "transcript": "Hi there! Thanks for calling We Haul Raleigh. How can I help you today?",
        "id": null,
        "text": null
      }
    ]
  },
  {
    "item_id": "item_C3uQ9XTvjc9tobD9fv9Xw",
    "previous_item_id": "item_C3uQ0ycMymMXvPMNEE7K3",
    "type": "message",
    "role": "user",
    "content": [
      {
        "type": "input_audio",
        "audio": null,
        "transcript": "Where are you located?",
        "id": null,
        "text": null
      }
    ],
    "status": "completed"
  },
  {
    "item_id": "item_C3uQB1RgzHWRYhJ2NjTQf",
    "previous_item_id": null,
    "type": "message",
    "role": "assistant",
    "status": "in_progress",
    "content": [
      {
        "type": "audio",
        "audio": null,
        "transcript": "We’re located in Raleigh, NC, on Little Brier Creek Lane.",
        "id": null,
        "text": null
      }
    ]
  },
  {
    "item_id": "item_C3uQJDk9M2Wxz8cKc25O7",
    "previous_item_id": "item_C3uQB1RgzHWRYhJ2NjTQf",
    "type": "message",
    "role": "user",
    "content": [
      {
        "type": "input_audio",
        "audio": null,
[...]

PR #3828, opened 2026-07-14
fix(run_state): preserve unrecognized tool-output raw items on resume

iridescentWen:
### Summary

When deserializing a serialized `RunState`, a `tool_call_output_item` whose `raw_item` had an unrecognized `type` (so it failed the tool-output union validation) was returned as `None` by `_deserialize_tool_call_output_raw_item` and then **silently skipped** by `_deserialize_items` (`if raw_item_output is None: continue`). That dropped the output while keeping its paired tool call, leaving an orphaned `function_call` with no matching `function_call_output` — which the Responses API rejects on resume.

This is asymmetric with the tool-**call** side: `_deserialize_tool_call_raw_item` already falls back to the original mapping when validation fails, and `_deserialize_tool_call_output_raw_item` itself already returns the raw mapping for `shell_call_output` / `apply_patch_call_output` / `custom_tool_call_output`. Only the union fallback dropped the item.

### Change

Make the union fallback consistent: return the original mapping instead of `None`, and remove the now-dead `None` check at the call site. `dict[str, Any]` is already part of the return type and an accepted `raw_item` shape, so no new shape is introduced.

This only affects **backward-read** of persisted state (it makes resume more robust); it does not change the serialized format or the schema version.

### Test plan

- Added `test_deserialize_tool_call_output_preserves_unrecognized_raw_item`: a `function_call` paired with an untyped output raw item now deserializes to **both** items (previously only the call survived), and the restored output re-serializes with its `call_id` intact.
- Updated `test_deserialize_items_union_adapter_fallback`, which previously asserted the buggy drop (`len == 0`), to assert the item is preserved.
- Verified end-to-end that the call/output pairing is preserved and the output item round-trips via `to_input_item()`.
- `make format`, `make lint`, `make mypy`, `make pyright`, and `make coverage` (90% ≥ 85% gate) all pass; full suite 4915 passed / 5 skipped.

### Issue number

N/A (found via code review).

### Checks

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -2196,8 +2196,14 @@ def _deserialize_tool_approval_item(
 
 def _deserialize_tool_call_output_raw_item(
     raw_item: Mapping[str, Any],
-) -> FunctionCallOutput | ComputerCallOutput | LocalShellCallOutput | dict[str, Any] | None:
-    """Deserialize a tool call output raw item; return None when validation fails."""
+    expected_type: str | None = None,
+) -> FunctionCallOutput | ComputerCallOutput | LocalShellCallOutput | dict[str, Any]:
+    """Deserialize a tool call output raw item, falling back to the original mapping.
+
+    ``expected_type`` is the output type implied by the paired tool call (e.g.
+    ``function_call_output`` for a ``function_call``); it is used to repair a persisted
+    output that lost its ``type`` so replay's orphan pruning can still pair it.
+    """
     if not isinstance(raw_item, Mapping):
         return cast(
             FunctionCallOutput | ComputerCallOutput | LocalShellCallOutput | dict[str, Any],
@@ -2222,7 +2228,15 @@ def _deserialize_tool_call_output_raw_item(
             _TOOL_CALL_OUTPUT_UNION_ADAPTER.validate_python(normalized_raw_item),
         )
     except ValidationError:
-        return None
+        # Mirror _deserialize_tool_call_raw_item: keep the original mapping instead of
+        # dropping the output item, which would orphan its paired tool call on resume.
+        # If the mapping lost its type, restore the type implied by the paired call so
+        # replay's orphan pruning (which pairs by output type) can still match it. The
+        # type comes from the actual call, never guessed from the output shape, so a
+        # non-function output is not mislabeled as a function output.
+        if expected_type is not None and "type" not in normalized_raw_item:
+            normalized_raw_item["type"] = expected_type
+        return normalized_raw_item
 
 
 def _parse_guardrail_entry(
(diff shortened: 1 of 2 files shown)


iridescentWen:
Good catch @codex — you're right that preserving alone wasn't enough. I verified it: `drop_orphan_function_calls` keys completed outputs by the output's `type`, so a typeless output left the paired `function_call` orphaned and pruned on resume.

Updated the fallback (single commit): when the preserved mapping clearly is a function tool output (a string `output` keyed by `call_id`) but lost its `type`, it now restores the canonical `"function_call_output"` type, so replay's orphan pruning pairs it with its call. Outputs that carry an unknown-but-present type are preserved verbatim (their paired call is likewise an unknown type and isn't pruned).

Added end-to-end coverage asserting the paired `function_call` survives `drop_orphan_function_calls` after resume for both the typeless-function-output case and the unknown-hosted-type case.

iridescentWen:
Thanks @codex — the structured-output point (second comment) is valid and now fixed. `FunctionCallOutput.output` is `str | list` (structured text/image/file), so a typeless function output whose `output` is a list was still left without a type and would orphan its call. The type-restoration guard now accepts `str | list`; a `dict` output (e.g. computer screenshot) is deliberately left untouched since it belongs to another tool type.

Added an end-to-end test for the structured (list) output case asserting the paired `function_call` survives `drop_orphan_function_calls`.

(The first comment — the string-only case — is already handled by the type restoration added in the previous revision; that path sets `function_call_output` and the paired call survives, as its test asserts.)

iridescentWen:
Reworked the repair to address the root cause you surfaced (thanks — the non-function mis-typing was real).

Instead of guessing the output type from the payload shape (which can't distinguish a `local_shell_call_output` from a `function_call_output` — both are `{call_id, output}`), a typeless output now gets the type implied by its **paired call**, via the canonical `_TOOL_CALL_TO_OUTPUT_TYPE` map: `function_call → function_call_output`, `local_shell_call → local_shell_call_output`, etc. The type comes from the actual call, never the output shape, so a non-function output is never mislabeled.

Behavior:
- typeless output with a known paired call → typed to that call's output type; `drop_orphan_function_calls` keeps the pair.
- output with an unknown-but-present type → preserved verbatim (its paired call is likewise unknown-typed and not pruned).
- typeless output with no paired call → preserved as-is (there is no call to orphan).

Added tests for every known tool family plus the local_shell mis-typing case (asserts it becomes `local_shell_call_output`, not `function_call_output`).

iridescentWen:
On the "repair invalid known-call output types" suggestion — this is a deliberate boundary, not an oversight.

The fix repairs only a *missing* type (inferred from the paired call). It intentionally does **not** override a type that is already present, even an unrecognized one, for two reasons:

1. An unknown-but-present output type is preserved on purpose for forward-compatibility with future hosted tools (its paired call is likewise unknown-typed and isn't pruned).
2. Rewriting a persisted/provided type is exactly what was declined in #1434 ("we do not want the SDK to override a valid status reported by the server"). Silently retyping a present value would reintroduce that concern.

A `tool_call_output_item` that carries a *present-but-invalid* type paired with a known call would be corrupt/legacy persisted state, which is outside the scope of this change (which targets the silent-drop of unrecognized outputs).

(The other note is the missing-type + paired-call case, which is already repaired here — covered by the added per-family tests.)

Happy to revisit if a maintainer would prefer overriding invalid-but-present types too.

seratch:
Thanks for sharing this. The deserialization branch can silently drop a synthetically malformed tool output, but I could not find a supported producer that writes this shape. The current and released RunState producers include an explicit output type, and newer serialized variants are handled through schema-version compatibility rather than best-effort forward repair. The added tests construct private _deserialize_items() payloads directly; they do not show a snapshot produced by a released SDK or a public resume failure.

I am going to close this PR for now. If you can provide a serialized RunState generated by a released supported path, plus a minimal Runner resume reproduction that fails on the base, we can revisit the issue and choose a targeted producer fix or schema migration from that evidence.

PR #3831, opened 2026-07-14
fix(tools): isolate agent tool stream callback cancellation

### LINKED ISSUE #3829: Agent.as_tool() can hang when an on_stream callback raises CancelledError
### Please read this first

- **Have you read the docs?** Yes, including the agents-as-tools streaming documentation.
- **Have you searched for related issues?** Yes. I could not find an issue or pull request covering cancellation raised by an `Agent.as_tool()` `on_stream` callback. PR #2681 handles cancellation inside an MCP tool call, and PR #2714 handles recovery of nested streamed output; neither covers the callback dispatcher queue lifecycle described here.

### Describe the bug

When an async `on_stream` callback passed to `Agent.as_tool()` raises `asyncio.CancelledError`, the agent tool invocation can hang indefinitely.

The callback dispatcher currently catches `Exception`, but `asyncio.CancelledError` inherits directly from `BaseException`. The callback therefore terminates the dispatcher task. After the nested stream finishes, the producer enqueues the sentinel and waits for `event_queue.join()`, but no consumer remains to call `task_done()` for the queued items.

This is a callback-local cancellation: the task invoking the agent tool is still active. It is distinct from cancelling the parent agent tool invocation, which should continue to propagate and cancel any active callback.

### Debug information

- Agents SDK version: `v0.18.2` (also reproducible on `main` at `4d96778`)
- OS: macOS

### Repro steps

Add the following test to `tests/test_agent_as_tool.py` and run it with `uv run --frozen pytest -q tests/test_agent_as_tool.py -k callback_cancelled_repro`:

```python
@pytest.mark.asyncio
async def test_callback_cancelled_repro(monkeypatch: pytest.MonkeyPatch) -> None:
    agent = Agent(name="nested")
    first_event = RawResponsesStreamEvent(
        data=cast(Any, {"type": "response_started"})
    )
    second_event = RawResponsesStreamEvent(
        data=cast(Any, {"type": "output_text_delta", "delta": "hi"})
    )

    class DummyStreamingResult:
        final_output = "ok"
        current_agent = agent

        async def stream_events(self):
            yield first_event
            yield second_event

    monkeypatch.setattr(
        Runner,
        "run_streamed",
        classmethod(lambda *args, **kwargs: DummyStreamingResult()),
    )

    handled_events = []

    async def on_stream(payload: AgentToolStreamEvent) -> None:
        handled_events.append(payload["event"])
        if payload["event"] is first_event:
            raise asyncio.CancelledError("callback-local cancellation")

    tool_call = ResponseFunctionToolCall(
        id="call_repro",
        arguments='{"input": "go"}',
        call_id="call-repro",
        name="nested_tool",
        type="function_call",
    )
    tool = agent.as_tool(
        tool_name="nested_tool",
        tool_description="Reproduce callback cancellation",
        on_stream=on_stream,
    )
    context = ToolContext(
        context=None,
        tool_name="nested_tool",
        tool_call_id=tool_call.call_id,
        tool_arguments=tool_call.arguments,
        tool_call=tool_call,
    )

    output = await asyncio.wait_for(
        tool.on_invoke_tool(context, tool_call.arguments),
        timeout=0.5,
    )

    assert output == "ok"
    assert handled_events == [first_event, second_event]
```

On current `main`, the test fails with `TimeoutError` because the invocation remains blocked in `event_queue.join()`.

### Expected behavior

A cancellation raised by the callback itself should be isolated like other callback failures, allowing later stream events to be dispatched and the agent tool invocation to complete. Cancelling the parent invocation should still cancel and await the active callback and propagate `CancelledError` to the caller.

russeell:
### Summary

This pull request fixes a hang in `Agent.as_tool()` streaming callbacks when an `on_stream` callback raises `asyncio.CancelledError`. Callback-local cancellation previously terminated the dispatcher task, leaving the producer blocked in `event_queue.join()` after the nested stream completed.

Async callbacks now run in an isolated task so callback-local cancellation can be distinguished from cancellation of the parent agent tool invocation. Callback-local cancellation is logged and dispatch continues, while parent cancellation cancels and awaits the active callback and dispatcher, including when cancellation arrives after the nested stream has ended but callbacks are still draining.

The change preserves the public API, stream event order, ordinary callback exception handling, and non-blocking event consumption behavior.

### Test plan

- `uv run --frozen pytest -q tests/test_agent_as_tool.py` (`46 passed`)
  - format passed
  - lint passed
  - typecheck passed
  - full test suite passed
- `codex review --uncommitted` (no actionable findings)

### Issue number

Closes #3829

### Checks

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -805,8 +805,52 @@ class Agent(AgentBase, Generic[TContext]):
                         """Execute the user callback while capturing exceptions."""
                         try:
                             maybe_result = stream_handler(payload)
-                            if inspect.isawaitable(maybe_result):
-                                await maybe_result
+                        except asyncio.CancelledError:
+                            logger.exception(
+                                "Error while handling on_stream event for agent tool %s.",
+                                self.name,
+                            )
+                            return
+                        except Exception:
+                            logger.exception(
+                                "Error while handling on_stream event for agent tool %s.",
+                                self.name,
+                            )
+                            return
+
+                        if not inspect.isawaitable(maybe_result):
+                            return
+
+                        try:
+                            handler_task = asyncio.ensure_future(maybe_result)
+                        except Exception:
+                            logger.exception(
+                                "Error while handling on_stream event for agent tool %s.",
+                                self.name,
+                            )
+                            return
+
+                        try:
+                            done, _ = await asyncio.wait(
+                                {handler_task}, return_when=asyncio.FIRST_COMPLETED
+                            )
+                            finished_task = done.pop()
+                            if finished_task.cancelled():
+                                logger.error(
+                                    "Error while handling on_stream event for agent tool %s: "
+                                    "callback was cancelled.",
+                                    self.name,
+                                )
+                                return
+                            finished_task.result()
+                        except asyncio.CancelledError:
+                            if not handler_task.done():
+                                handler_task.cancel()
+                            try:
+                                await handler_task
+                            except (asyncio.CancelledError, Exception):
+                                pass
+                            raise
                         except Exception:
                             logger.exception(
                                 "Error while handling on_stream event for agent tool %s.",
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this. I confirmed the synthetic failure: if an on_stream callback raises CancelledError, the dispatcher exits and the invocation can remain blocked in event_queue.join().

However, the issue only demonstrates a callback deliberately raising CancelledError against a mocked stream. It does not show a realistic supported callback whose child work is independently cancelled, or why that cancellation cannot be handled by the callback that owns it. This PR would add a task for every async event and establish new SDK behavior that callback-local cancellation is logged and ignored.

I am going to close this PR for now. If you can provide a minimal realistic callback showing independently cancelled work, its practical consequence, and why callback-owned handling is insufficient, we can revisit the cancellation contract from that evidence.

PR #3835, opened 2026-07-14
fix(agent): catch CancelledError raised inside on_stream callbacks for Agent.as_tool()

### LINKED ISSUE #3829: Agent.as_tool() can hang when an on_stream callback raises CancelledError

Diwak4r:
This pull request fixes #3829.

When an async on_stream callback passed to Agent.as_tool() raised syncio.CancelledError, the callback dispatcher task would terminate because it only caught Exception. The producer then enqueued the sentinel and waited on event_queue.join(), but no consumer remained to call 	ask_done(), causing the tool invocation to hang indefinitely.

## Changes

- src/agents/agent.py: _run_handler now also catches syncio.CancelledError and treats it as a callback-local error (log and continue) rather than terminating the dispatcher. Parent cancellation of the dispatcher task itself is still propagated, detected via syncio.Task.cancelling() when available (Python 3.11+).
- 	ests/test_agent_as_tool.py: added 	est_agent_as_tool_streaming_handler_cancelled_error_does_not_hang regression test.

## Verification

`ash
uv run --frozen pytest -q tests/test_agent_as_tool.py
`

All 45 tests in the file pass.

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -803,10 +803,25 @@ class Agent(AgentBase, Generic[TContext]):
 
                     async def _run_handler(payload: AgentToolStreamEvent) -> None:
                         """Execute the user callback while capturing exceptions."""
+                        task = asyncio.current_task()
                         try:
                             maybe_result = stream_handler(payload)
                             if inspect.isawaitable(maybe_result):
                                 await maybe_result
+                        except asyncio.CancelledError:
+                            # A CancelledError raised by the callback itself should not
+                            # terminate the dispatcher task. Re-raise only when the
+                            # dispatcher task itself is being cancelled (parent
+                            # cancellation), which is detectable via task.cancelling()
+                            # on Python 3.11+. On older versions we preserve the
+                            # original behavior of propagating cancellation.
+                            if task is not None and getattr(task, "cancelling", lambda: 0)() > 0:
+                                raise
+                            logger.warning(
+                                "on_stream callback raised CancelledError for agent tool %s; "
+                                "treating it as a callback-local error and continuing.",
+                                self.name,
+                            )
                         except Exception:
                             logger.exception(
                                 "Error while handling on_stream event for agent tool %s.",
--- a/tests/test_agent_as_tool.py
+++ b/tests/test_agent_as_tool.py
@@ -2456,6 +2456,73 @@ async def test_agent_as_tool_streaming_handler_exception_does_not_fail_call(
 
 
 @pytest.mark.asyncio
+async def test_agent_as_tool_streaming_handler_cancelled_error_does_not_hang(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A CancelledError raised inside an on_stream callback should not hang the tool."""
+    agent = Agent(name="cancelled_error_agent")
+
+    first_event = RawResponsesStreamEvent(data=cast(Any, {"type": "response_started"}))
+    second_event = RawResponsesStreamEvent(
+        data=cast(Any, {"type": "output_text_delta", "delta": "hi"})
+    )
+
+    class DummyStreamingResult:
+        def __init__(self) -> None:
+            self.final_output = "ok"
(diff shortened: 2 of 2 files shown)


seratch:
The queue hang is reachable if a user-supplied on_stream callback lets CancelledError escape, but #3829 only demonstrates this with a mocked stream and a callback deliberately raising it. It does not show realistic independently cancelled callback work or why the callback that owns that work cannot handle its own cancellation. We already closed #3831 on this evidence gap, and this PR adds no new user scenario while establishing a new SDK cancellation contract.

I am going to close this PR for now. If you can provide a minimal realistic callback showing independently cancelled work, its practical consequence, and why callback-owned handling is insufficient, we can revisit the cancellation contract from that evidence.

PR #3849, opened 2026-07-16
docs: make dynamic prompt example runnable in standard Python scripts

Syed-Umer-Ali:
## Summary

This PR updates the dynamic prompt example to make it directly runnable as a standard Python script.

The current example uses `await Runner.run(...)` at the top level, which is only valid in environments that support top-level `await` (such as Jupyter notebooks). Running the example as a regular `.py` file results in a syntax error.

## Changes

- Add an `async main()` function.
- Execute it using `asyncio.run(main())`.
- Import `asyncio`.
- Preserve the original example behavior.

## Why

This improves the developer experience by ensuring the documentation example works when copied into a standalone Python script without requiring additional modifications.

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -95,8 +95,10 @@ You can also generate the prompt dynamically at run time:
 
 ```python
 from dataclasses import dataclass
+import asyncio
 
 from agents import Agent, GenerateDynamicPromptData, Runner
+
 
 @dataclass
 class PromptContext:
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the suggestion, but we'd like to keep the part as simple as possible. [The quickstart page](https://openai.github.io/openai-agents-python/quickstart/) shows the concrete example with main function, so let me close this one.

PR #3851, opened 2026-07-16
docs: make UserContext example runnable in Agents documentation

MuhammadHamidRaza:
## Summary

This PR updates the `UserContext` example in the **Agents** documentation to be fully runnable.

### Changes

* Added the missing `self` parameter to `fetch_purchases()`.
* Added a minimal `Purchase` dataclass so the example does not reference an undefined type.
* Added `from __future__ import annotations` to support the forward reference in the return type.

### Why

The previous example could not be copied and executed as-is because:

* `fetch_purchases()` was missing `self`, resulting in a `TypeError` when called on an instance.
* `Purchase` was referenced but never defined, resulting in a `NameError`.

The updated example is self-contained, copy-paste runnable, and preserves the original intent of demonstrating a context dataclass.

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -128,18 +128,29 @@ Agents are generic on their `context` type. Context is a dependency-injection to
 Read the [context guide](context.md) for the full `RunContextWrapper` surface, shared usage tracking, nested `tool_input`, and serialization caveats.
 
 ```python
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+
+@dataclass
+class Purchase:
+    id: str
+
+
 @dataclass
 class UserContext:
     name: str
     uid: str
     is_pro_user: bool
 
-    async def fetch_purchases() -> list[Purchase]:
-        return ...
+    async def fetch_purchases(self) -> list[Purchase]:
+        return []
 
 agent = Agent[UserContext](
     ...,
 )
+        
 ```
 
 ## Output types
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -128,15 +128,11 @@ Agents are generic on their `context` type. Context is a dependency-injection to
 Read the [context guide](context.md) for the full `RunContextWrapper` surface, shared usage tracking, nested `tool_input`, and serialization caveats.
 
 ```python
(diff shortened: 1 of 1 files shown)


docs/agents.md
```diff
@@ -128,18 +128,29 @@ Agents are generic on their `context` type. Context is a dependency-injection to
 Read the [context guide](context.md) for the full `RunContextWrapper` surface, shared usage tracking, nested `tool_input`, and serialization caveats.
 
 ```python
+
```
seratch:
```suggestion
```

docs/agents.md
```diff
@@ -128,18 +128,29 @@ Agents are generic on their `context` type. Context is a dependency-injection to
...
+from dataclasses import dataclass
+
+
```
seratch:
```suggestion
```

docs/agents.md
```diff
@@ -128,18 +128,29 @@ Agents are generic on their `context` type. Context is a dependency-injection to
...
+@dataclass
+class Purchase:
+    id: str
+
+
```
seratch:
```suggestion
```

docs/agents.md
```diff
@@ -128,18 +128,29 @@ Agents are generic on their `context` type. Context is a dependency-injection to
...
 
-    async def fetch_purchases() -> list[Purchase]:
-        return ...
+    async def fetch_purchases(self) -> list[Purchase]:
+        return []
 
 agent = Agent[UserContext](
     ...,
 )
+
```
seratch:
```suggestion        
```

docs/agents.md
```diff
@@ -128,18 +128,29 @@ Agents are generic on their `context` type. Context is a dependency-injection to
...
 @dataclass
 class UserContext:
     name: str
     uid: str
     is_pro_user: bool
 
-    async def fetch_purchases() -> list[Purchase]:
-        return ...
+    async def fetch_purchases(self) -> list[Purchase]:
+        return []
```
seratch, line 145:
```suggestion
        # implement your logic here
        return []
```

PR #3854, opened 2026-07-16
fix: preserve streamed session input across model retries

### LINKED ISSUE #3852: Local Session loses the user's input turn when a model retry rewinds it (never re-persisted)
## Please read this first

- [x] **Have you searched for related issues?** Related to #3089, #3304, #3116 (same rewind subsystem), but I couldn't find this specific case: the rewind removes a legitimately-matched input turn that the retry then never re-saves.

## What happens

On `Runner.run_streamed` with a local `Session` (no `conversation_id`/`previous_response_id`, not an `OpenAIConversationsSession`), a model request that fails once and succeeds on retry loses the user's input turn. The assistant reply is saved; the `role:"user"` item that produced it is gone. Flagging in case the rewind is intended here, but it looks like a gap.

Mechanism in 0.18.1:

1. The input is saved once at the start of the run and flagged committed — `_stream_input_persisted = True` (`run_loop.py:1416`). Every later save in the turn passes `input=[]` (`run_loop.py:349`).
2. On a retry, `rewind_model_request` → `rewind_session_items` → `pop_item()` removes that input (`session_persistence.py`).
3. The retry never re-saves the input. The save-once block is outside the retry loop (the retry is inside `stream_response_with_retry`), so it doesn't re-run, and later saves pass `input=[]`.

`rewind_session_items` says it rolls back input "so we do not accumulate duplicate inputs" on a retry — which reads as if it expects the retry to re-persist the input. On the streamed path it doesn't, so the rewind is a straight delete.

This isn't specific to a custom session. Deleting the item in `pop_item` is the documented [`Session` contract](https://github.com/openai/openai-agents-python/blob/v0.18.1/src/agents/memory/session.py#L44-L50) ("Remove and return the most recent item"), and the built-in `SQLiteSession` does the same (`DELETE … RETURNING`) — and that's what the repro below uses, so no custom code is involved.

The rewind targets only the turn's own new input (`prepare_input_with_session` returns the new-turn items, not history), so it applies to any turn that retries, not just the first. It's most visible on the first/only turn, where the chat is left showing an answer with no question.

## Repro

Self-contained. Two identical runs; the only difference is whether the first attempt hits a retryable 500.

```python
import asyncio
import tempfile
import os
from typing import AsyncIterator

import httpx
from openai import InternalServerError
    Response,
    ResponseCompletedEvent,
    ResponseOutputMessage,
    ResponseOutputText,
)

from agents import Agent, Runner, RunConfig, ModelSettings, retry_policies
from agents.retry import ModelRetrySettings
from agents.models.interface import Model, ModelProvider
from agents.memory.sqlite_session import SQLiteSession


class ScriptedModel(Model):
    def __init__(self, fail_once: bool):
        self.fail_once = fail_once
        self.attempts = 0

    async def stream_response(self, *args, **kwargs) -> AsyncIterator:
        self.attempts += 1
        if self.fail_once and self.attempts == 1:
            raise InternalServerError(
                "transient error",
                response=httpx.Response(500, request=httpx.Request("POST", "http://x")),
                body=None,
            )
        yield _answer("You have 30 days from notification.")

    async def get_response(self, *args, **kwargs):
        raise NotImplementedError


async def run(fail_once: bool) -> list[str]:
    db = tempfile.NamedTemporaryFile(suffix=".db", delete=False)
    db.close()
    session = SQLiteSession("demo", db.name)
    agent = Agent(
        name="assistant",
        instructions="",
        model=ScriptedModel(fail_once),
        model_settings=ModelSettings(
            retry=ModelRetrySettings(max_retries=2, policy=retry_policies.http_status([500]))
        ),
    )
    result = Runner.run_streamed(
        agent, "What is the deadline to appeal?", session=session,
        run_config=RunConfig(model_provider=_provider(agent.model)),
    )
        pass
    roles = [m["role"] for m in await session.get_items() if isinstance(m, dict) and "role" in m]
    os.unlink(db.name)
    return roles


async def main():
    print("no retry :", await run(fail_once=False))
    print("one retry:", await run(fail_once=True))


def _answer(text: str) -> ResponseCompletedEvent:
    message = ResponseOutputMessage(
        id="msg", type="message", role="assistant", status="completed",
        content=[ResponseOutputText(type="output_text", text=text, annotations=[])],
    )
    response = Response(
        id="resp", created_at=0.0, model="scripted", object="response",
        output=[message], parallel_tool_calls=False, tool_choice="auto", tools=[],
    )
    return ResponseCompletedEvent(type="response.completed", response=response, sequence_number=0)


def _provider(model):
    class P(ModelProvider):
        def get_model(self, name): return model
    return P()


asyncio.run(main())
```

Output on 0.18.1:

```
no retry : ['user', 'assistant']
one retry: ['assistant']
```

## Expected

After a retry the session should still hold exactly one copy of the user turn. Removing the retry policy makes the same run persist `['user', 'assistant']`.

## Fix

Don't rewind input that's already committed: once `_stream_input_persisted` is set, drop it from the retry's `session_items_to_rewind` target. The streamed retry never re-persists it, so there's nothing to dedupe. (Alternatively, re-save the input on the successful retry.) The non-streamed runner shares the same shape. Happy to open a PR.

## Environment

- openai-agents 0.18.1, Python 3.11.
- Local `SQLiteSession` reproduces; originally hit with a custom `SessionABC` over Postgres. Any local backend applies (the SDK drives them through the `Session` protocol).
- Requires a retry policy on `ModelSettings.retry`. Server-managed conversations use a different rewind path (`rewind_input`) and aren't covered here.

hxaxd:
### Summary

Streamed runs persist the current user input to a local session before starting the model request. When that request was retried, the retry callback removed the committed input, while the successful retry only persisted the assistant output. The resulting session could therefore contain an answer without the user turn that produced it.

This change keeps already-persisted local session input out of the model retry rewind. Server-managed conversation rewind behavior is unchanged.

A regression test covers a streamed request that fails once with a retryable connection error and then succeeds, confirming that the session retains exactly one user turn and one assistant turn.

### Test plan

```text
pytest tests/test_agent_runner_streamed.py::test_streamed_model_retry_preserves_session_input tests/test_agent_runner_streamed.py::test_streamed_run_preserves_request_usage_entries_after_retry tests/test_agent_runner_streamed.py::test_streamed_run_preserves_request_usage_entries_after_conversation_locked_retry -q
```

### Issue number

Closes #3852

### Checks

### DIFF
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -1486,7 +1486,11 @@ async def run_single_turn_streamed(
     model_settings = model_settings_with_prompt_cache_key(model_settings, prompt_cache_key)
 
     async def rewind_model_request() -> None:
-        items_to_rewind = session_items_to_rewind if session_items_to_rewind is not None else []
+        items_to_rewind = []
+        # Streaming persists local session input before the request, so a model retry must not
+        # remove that committed turn.
+        if not streamed_result._stream_input_persisted and session_items_to_rewind is not None:
+            items_to_rewind = session_items_to_rewind
         await rewind_session_items(session, items_to_rewind, server_conversation_tracker)
         if server_conversation_tracker is not None:
             server_conversation_tracker.rewind_input(filtered.input)
--- a/tests/test_agent_runner_streamed.py
+++ b/tests/test_agent_runner_streamed.py
@@ -370,6 +370,38 @@ async def test_streamed_run_preserves_request_usage_entries_after_retry() -> Non
 
 
 @pytest.mark.asyncio
+async def test_streamed_model_retry_preserves_session_input() -> None:
+    model = FakeModel()
+    model.add_multiple_turn_outputs(
+        [
+            APIConnectionError(
+                message="connection error",
+                request=httpx.Request("POST", "https://example.com"),
+            ),
+            [get_text_message("done")],
+        ]
+    )
+    agent = Agent(
+        name="test",
+        model=model,
+        model_settings=ModelSettings(
+            retry=ModelRetrySettings(
+                max_retries=1,
+                policy=retry_policies.network_error(),
+            )
+        ),
+    )
+    session = SimpleListSession()
+
+    result = Runner.run_streamed(agent, input="test", session=session)
+    async for _ in result.stream_events():
+        pass
+
+    saved_items = await session.get_items()
+    assert [item.get("role") for item in saved_items] == ["user", "assistant"]
+
+
+@pytest.mark.asyncio
 async def test_streamed_run_preserves_request_usage_entries_after_conversation_locked_retry() -> (
     None
 ):


seratch:
Thanks for your effort here. The issue was resolved by https://github.com/openai/openai-agents-python/pull/3857

PR #3856, opened 2026-07-16
Docs/fix hooks example await outside function

Syed-Umer-Ali:
## What
The lifecycle hooks example in `docs/agents.md` used `await` at module level outside an `async def` function, which raises `SyntaxError: 'await' outside function` on Python 3.10+.

## Why
Running the example as-written produces a syntax error. Verified locally.

## Fix
- Added `import asyncio`
- Wrapped the agent run in `async def main():`
- Added `if __name__ == "__main__": asyncio.run(main())`

## Verification
Confirmed the corrected code parses without syntax errors via `ast.parse()`. The pattern matches all other runnable examples in the same file.

### DIFF
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -280,6 +280,8 @@ Typical hook timing:
 Use `RunHooks` when you want a single observer for the whole workflow, and `AgentHooks` when one agent needs custom side effects.
 
 ```python
+import asyncio
+
 from agents import Agent, RunHooks, Runner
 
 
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the suggestion, but we'd like to keep the part as simple as possible. [The quickstart page](https://openai.github.io/openai-agents-python/quickstart/) shows the concrete example with main function, so let me close this one.

PR #3860, opened 2026-07-17
docs: explain programmatic tool calling

william-xue:
## What
- add programmatic tool calling to the tools guide and tool chooser
- show a minimal structured function-tool setup
- document caller permissions, output schemas, and forced tool choice

## Why
Programmatic tool calling is available in the SDK and has a complete example, but the tools guide does not currently explain when to use it or how to configure eligible tools.

## Test
- `uv run mkdocs build`
- validated the documented tool configuration with a local Python snippet
- `git diff --check`

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -1,8 +1,9 @@
 # Tools
 
-Tools let agents take actions: things like fetching data, running code, calling external APIs, and even using a computer. The SDK supports five categories:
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for suggesting the change here. https://github.com/openai/openai-agents-python/pull/3872 will cover everything for the upcoming release.

PR #3870, opened 2026-07-17
fix: fail closed for callable needs_approval on invalid/non-object JSON

### LINKED ISSUE #3863: Callable needs_approval fails open when tool arguments are invalid JSON
### Please read this first

- **Have you read the docs?** Yes — Human-in-the-loop / tool approval docs
- **Have you searched for related issues?** Searched for needs_approval + invalid JSON / fail-open. Closest historical analog is closed MCP require_approval work (#3168); this is a different surface (callable needs_approval on function tools).

### Describe the bug

When `needs_approval` is a **callable**, invalid tool-call JSON is swallowed and replaced with `{}` before the predicate runs:

```python
try:
    parsed_args = json.loads(tool_call.arguments or "{}")
except json.JSONDecodeError:
    parsed_args = {}
```

Doc examples use content-based gates such as `"refund" in params.get("subject", "").lower()`. With `params == {}`, those predicates return **False**, so the run **does not interrupt for approval**.

Bool `needs_approval=True` is unaffected. Non-object JSON (`null`, `[]`) can later crash the callable with `UserError` / `AttributeError` on `.get`.

The same swallow pattern exists in realtime (`src/agents/realtime/session.py`).

### Debug information

- Agents SDK: `0.18.3` @ `965335aba6f6c71500e0b8cdb4e9e495f5801d4d`
- Python 3.12 / Windows 11
- Surface: `function_needs_approval` in `src/agents/run_internal/tool_execution.py`

### Repro steps

```python
import asyncio
from agents.run_internal.tool_execution import function_needs_approval
from openai.types.responses.response_function_tool_call import ResponseFunctionToolCall

async def needs_if_refund(_ctx, params, _call_id):
    return "refund" in str(params.get("subject", "")).lower()

@function_tool(needs_approval=needs_if_refund)
async def send_email(subject: str, body: str) -> str:
    return f"Sent {subject}"

async def main():
    ctx = RunContextWrapper(context=None)
    bad = ResponseFunctionToolCall(
        type="function_call",
        name="send_email",
        call_id="call_bad",
        arguments='{"subject": "REFUND request", "body": "please refund"',  # invalid JSON
    )
    print(await function_needs_approval(send_email, ctx, bad))
    # Observed: False  (approval not required)

asyncio.run(main())
```

### Expected behavior

Invalid or non-object tool arguments should fail closed for callable approval gates (require approval, or raise `ModelBehaviorError`) rather than evaluating the predicate on `{}`.

### Actual behavior

`JSONDecodeError` → `{}` → typical field-based predicates return `False` → no HITL interruption.

### Suggested direction

On parse failure or non-`dict` result: treat as needs approval, or raise before calling the predicate. Align realtime path.

I'd be happy to prepare a PR with a regression test if this behavior is confirmed.

baiyuxi930826:
## Summary

Fixes #3863.

When `needs_approval` is a **callable**, invalid tool-call JSON was previously swallowed and replaced with `{}` before the predicate ran. Content-based gates (e.g. `"refund" in params.get("subject", "")`) then returned `False` and skipped HITL.

This change fails closed:

- `JSONDecodeError` → require approval (`True`)
- non-`dict` JSON (`null`, `[]`, scalars) → require approval (`True`)
- valid object JSON → unchanged predicate behavior
- bool `needs_approval=True` → unchanged

Same fix in the realtime session path.

## Test plan

- [x] `pytest tests/test_function_needs_approval_invalid_json.py` (5 passed)
- [x] Covers invalid JSON, non-object JSON, valid refund/non-refund, bool path

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -548,12 +548,17 @@ class RealtimeSession(RealtimeModelListener):
     ) -> bool:
         """Evaluate a function tool's needs_approval setting with parsed args."""
         needs_setting = getattr(function_tool, "needs_approval", False)
-        parsed_args: dict[str, Any] = {}
         if callable(needs_setting):
             try:
-                parsed_args = json.loads(tool_call.arguments or "{}")
+                parsed = json.loads(tool_call.arguments or "{}")
             except json.JSONDecodeError:
-                parsed_args = {}
+                # Fail closed: invalid JSON must not look like "empty safe args".
+                return True
+            if not isinstance(parsed, dict):
+                return True
+            parsed_args: dict[str, Any] = parsed
+        else:
+            parsed_args = {}
         return await evaluate_needs_approval_setting(
             needs_setting,
             self._context_wrapper,
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -1208,13 +1208,23 @@ async def function_needs_approval(
     context_wrapper: RunContextWrapper[Any],
     tool_call: ResponseFunctionToolCall,
 ) -> bool:
-    """Evaluate a function tool's needs_approval setting with parsed args."""
-    parsed_args: dict[str, Any] = {}
+    """Evaluate a function tool's needs_approval setting with parsed args.
+
+    For callable approval predicates, invalid or non-object JSON arguments fail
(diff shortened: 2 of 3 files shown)


seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/3867

PR #3876, opened 2026-07-18
fix: convert assistant EasyInputMessage items with optional type/phase keys

vidigoat:
### Summary

`Converter.items_to_messages` crashes on valid assistant `EasyInputMessage` items that carry the type's optional keys. Per the installed SDK, `EasyInputMessageParam` allows `role: "assistant"`, an optional `type: "message"`, and an optional `phase` — whose docstring says "when sending follow-up requests, preserve and resend phase on all assistant messages — dropping it can degrade performance". Anyone following that guidance with a Chat Completions/LiteLLM model hits this.

What happens today for each valid shape (`main` @ 65886fa1):

| item | result |
|---|---|
| `{"role": "assistant", "content": "hello", "type": "message"}` | `TypeError: string indices must be integers` |
| `{"role": "assistant", "content": [{"type": "input_text", ...}], "type": "message"}` | `UserError: Unknown content type in ResponseOutputMessage` |
| `{"role": "assistant", "content": "hello", "phase": "final_answer"}` | `UserError: Unhandled item type or structure` |
| `{"role": "assistant", "content": "hello"}` (control) | works |

Root cause is a two-step miss: `maybe_easy_input_message` requires the key set to be *exactly* `{content, role}`, so any optional key disqualifies; the item is then either claimed by the `ResponseOutputMessage` branch — which assumes a list of output parts — or falls through entirely. The Responses model path accepts these same items, so this also breaks Responses↔ChatCompletions parity.

### Fix

- `maybe_easy_input_message`: require `content`/`role`, tolerate the optional `phase` key, still reject unknown keys.
- Output-message branch: handle string content and `input_text` parts for explicitly-typed easy-input assistant messages.

`phase` is intentionally dropped from the constructed Chat Completions message (no equivalent field), and both paths build fresh dicts so no stray keys leak into the payload.

### Tests

Added regression tests covering all four crash variants (fail on `main`, pass with the fix) plus a guard that unknown extra keys are still rejected. `tests/models/`: 462 passed locally (litellm-dependent files skipped — dependency not installed locally); ruff check/format clean on touched files.

*Prepared with AI assistance; I reviewed the change and ran the repros and test suites locally.*

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -634,25 +634,36 @@ class Converter:
 
             # 3) response output message => assistant
             elif resp_msg := cls.maybe_response_output_message(item):
+                contents = resp_msg["content"]
+
+                if isinstance(contents, str) or (
+                    isinstance(contents, list)
+                    and len(contents) > 0
+                    and all(
+                        isinstance(c, dict) and str(c.get("type", "")).startswith("input_")
+                        for c in contents
+                    )
+                ):
+                    # An EasyInputMessageParam with an explicit `type: "message"`:
+                    # string content or input-style parts can't appear in a real
+                    # ResponseOutputMessage, so convert the item exactly like the
+                    # untyped easy-input shape to keep the payloads identical.
+                    flush_assistant_message()
+                    typed_easy_asst: ChatCompletionAssistantMessageParam = {
+                        "role": "assistant",
+                        "content": cls.extract_text_content(contents),
+                    }
+                    result.append(typed_easy_asst)
+                    continue
+
                 # A reasoning item can be followed by an assistant message and then tool calls
                 # in the same turn, so preserve pending reasoning_content across this flush.
                 flush_assistant_message(clear_pending_reasoning_content=False)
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this and for iterating on the review feedback. We confirmed that the optional `type: "message"` discriminator can cause a valid `EasyInputMessage` to be misclassified as a `ResponseOutputMessage`.

However, the proposed scope conflates that narrow classification bug with `phase`. `phase` is a Responses API feature with no Chat Completions equivalent, so silently dropping it does not provide the claimed parity. The current implementation also introduces a second content-based classification path inside output-message conversion, adding more complexity than the underlying issue requires.

We plan to address the concrete bug at the classifier boundary instead: recognize `type: "message"` as part of the easy-input shape and require `id`, `status`, and `content` before classifying an item as a `ResponseOutputMessage`, without changing `phase` behavior.

I’m closing this PR in favor of that narrower fix: https://github.com/openai/openai-agents-python/pull/3877 Thanks again for surfacing the misclassification and providing the reproduction cases.

PR #3882, opened 2026-07-19
 Fix AsyncOpenAI import source in models/index.md

AAliKKhan:
## What
     Two code examples in `models/index.md` import `AsyncOpenAI` from `agents`
     instead of from `openai`.
  
     ## Why
     `AsyncOpenAI` is defined in the `openai` package. It is not a documented
     public export of `agents` — it only happens to be present because
     `agents/__init__.py` imports it internally for `set_default_openai_client`.
     Relying on this internal re-import is brittle and unconventional.
                                                                                                                                                                                                                                                                                                                                           
     ## Fix                                                                                                                                                                                                                                                                                                          
     Replace `from agents import ... AsyncOpenAI` with a separate                                                                                                                                                                                                                                                                          
     `from openai import AsyncOpenAI` in both code blocks.                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                           
     ## Verification                                                                                                                                                                                                                                                                                                                       
     Confirmed `AsyncOpenAI` is not in `agents.__all__`. Checked both locations                                                                                                                                                                                                                        Connect from 75+ providers to       
     in the source file.


seratch:
The existing imports aren't wrong. So, either is fine.

PR #3890, opened 2026-07-20
sandbox: do not flag expected not-found reads as errored spans

### LINKED ISSUE #3889: Lazy skill first load records expected `WorkspaceReadNotFoundError` as a failed span
### Please read this first

- [x] **Have you searched for related issues?** I found discussions about lazy skills, including #2906, but not this misleading error-span behavior.

### Describe the bug

When `LocalDirLazySkillSource` loads a skill for the first time, it calls:

```python
await session.read(skill_md_path, user=user)
```

to determine whether the skill is already materialized.

The file is expected not to exist on the first load, so the sandbox raises `WorkspaceReadNotFoundError`. `LocalDirLazySkillSource.load_skill()` catches the exception and successfully stages the skill, but tracing still records the `sandbox.read` operation as a failed/error span.

As a result, a successful skill load appears to contain an operational failure and can create noise in trace inspection and error monitoring.

The parent `load_skill` call succeeds with output similar to:

```json
{
  "status": "loaded",
  "skill_name": "current-rate-rule-manual",
  "path": ".agents/current-rate-rule-manual"
}
```

but its child `sandbox.read` span is marked as failed with:

```json
{
  "error_code": "workspace_read_not_found",
  "error_type": "WorkspaceReadNotFoundError",
  "error_retryable": false
}
```

This becomes annoying if you have agents using the traces to troubleshoot issues.

### Debug information

- Agents SDK version: reproduced with `v0.17.6`
- Latest checked version: the same existence-probe implementation is still present in `v0.18.3`
- Python version: Python 3.14
- Configuration: `SandboxAgent` with `Skills(lazy_from=LocalDirLazySkillSource(...))`

### Repro steps

Create a skill:

```text
skills/
└── example-skill/
    └── SKILL.md
```

```markdown
---
name: example-skill
description: Example skill used to reproduce lazy loading.
---

# Example skill

Follow these example instructions.
```

Run:

```python
import asyncio
from pathlib import Path

from agents import Runner
from agents.run import RunConfig
from agents.sandbox import (
    Manifest,
    SandboxAgent,
    SandboxPathGrant,
    SandboxRunConfig,
)
from agents.sandbox.capabilities import (
    Capabilities,
    LocalDirLazySkillSource,
    Skills,
)
from agents.sandbox.entries import LocalDir
from agents.sandbox.sandboxes.unix_local import UnixLocalSandboxClient


async def main() -> None:
    skills_root = Path("skills").resolve()

    agent = SandboxAgent(
        name="Skill test",
        instructions=(
            "Load the example-skill skill, read its SKILL.md, and follow it."
        ),
        default_manifest=Manifest(
            extra_path_grants=(
                SandboxPathGrant(
                    path=str(skills_root),
                    read_only=True,
                ),
            ),
        ),
        capabilities=[
            *Capabilities.default(),
            Skills(
                lazy_from=LocalDirLazySkillSource(
                    source=LocalDir(src=skills_root),
                ),
            ),
        ],
    )

    run_config = RunConfig(
        sandbox=SandboxRunConfig(
            client=UnixLocalSandboxClient(),
        ),
        workflow_name="lazy-skill-repro",
    )

    result = await Runner.run(
        agent,
        "Use the example skill.",
        run_config=run_config,
    )
    print(result.final_output)


asyncio.run(main())
```

Inspect the resulting trace:

1. `load_skill` completes successfully with `status: loaded`.
2. A child `sandbox.read` span is nevertheless marked as failed with `WorkspaceReadNotFoundError`.
3. The skill is subsequently written and can be read normally.

### Expected behavior

First-time lazy skill loading should not produce an error span for the expected absence of `SKILL.md`.

Possible approaches could include:

- adding a non-throwing `exists()` or `stat()` sandbox-session operation;
- using a non-erroring existence check inside `LocalDirLazySkillSource`; or
- treating this particular expected not-found result as a non-error trace outcome.

I would be happy to contribute a fix and regression test once the preferred approach is confirmed.

knQzx:
### summary

fixes #3889. LocalDirLazySkillSource.load_skill() probe reads a not yet materialized SKILL.md. the inner session correctly raises WorkspaceReadNotFoundError, load_skill catches it and proceeds, but the sandbox.read trace span was already marked errored by the instrumentation decorator, so a normal first time skill load surfaces a spurious errored span

this adds an expected_errors tuple to instrumented_op and _annotate. when a raised exception matches, the audit finish event still records it (retryability included) but the trace span is not flagged as errored, and the exception is re-raised unchanged. read declares WorkspaceReadNotFoundError as expected

one scope decision worth confirming: declaring this on the read op means it applies to every read, so a direct session.read() of a missing file also stops producing an errored span, though the audit event still records it. that felt right to me since a not found read is an expected, recoverable condition rather than a sandbox failure, but if you would rather keep direct reads visibly errored and only silence the load_skill probe i can switch to a per-call opt-in instead

### test plan

- new test drives a first time load_skill against a not yet materialized file and asserts the single sandbox.read span has error is None and the skill loads
- updated test_sandbox_session_error_events_and_traces_include_retryability so a not found read no longer flags the span while the audit event still carries error_retryable
- ruff, ruff format, mypy, and the full tests/sandbox suite pass

closes #3889

### DIFF
--- a/src/agents/sandbox/session/sandbox_session.py
+++ b/src/agents/sandbox/session/sandbox_session.py
@@ -12,7 +12,7 @@ from typing import Any, TypeVar, cast
 
 from ...run_config import SandboxArchiveLimits, SandboxConcurrencyLimits
 from ...tracing import Span, custom_span, get_current_trace
-from ..errors import OpName, SandboxError
+from ..errors import OpName, SandboxError, WorkspaceReadNotFoundError
 from ..files import FileEntry
 from ..types import ExecResult, ExposedPortEndpoint, User
 from .base_sandbox_session import BaseSandboxSession
@@ -39,6 +39,7 @@ def instrumented_op(
     ) = None,
     ok: Callable[[object], bool] | None = None,
     outputs: Callable[[object], tuple[bytes | None, bytes | None]] | None = None,
+    expected_errors: tuple[type[BaseException], ...] = (),
 ) -> Callable[[F], F]:
     """Decorator to emit SandboxSessionEvents around a SandboxSession operation."""
 
@@ -64,6 +65,7 @@ def instrumented_op(
                 finish_data=finish_cb,
                 ok=ok,
                 outputs=outputs,
+                expected_errors=expected_errors,
             )
 
         return cast(F, _wrapped)
@@ -378,6 +380,7 @@ class SandboxSession(BaseSandboxSession):
         finish_data: Callable[[T], dict[str, object]] | None = None,
         ok: Callable[[T], bool] | None = None,
         outputs: Callable[[T], tuple[bytes | None, bytes | None]] | None = None,
+        expected_errors: tuple[type[BaseException], ...] = (),
     ) -> T:
         span_cm = (
             custom_span(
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for sharing this idea. The issue was resolved by https://github.com/openai/openai-agents-python/pull/3893

PR #3894, opened 2026-07-20
docs: fix missing imports in context code example

Syed-Umer-Ali:
## What
The "Context" code block in docs/agents.md uses `@dataclass` and `Agent[UserContext]` without importing either.

## Why
Running the example as-written raises `NameError` / `ImportError` because `dataclass` is not imported and `Agent` is not imported from `agents`.

## Fix
Added `from dataclasses import dataclass` and `from agents import Agent` before the class definition.

## Verification
Confirmed `Agent` is exported from `agents` in `src/agents/__init__.py`.


seratch:
>Read the [context guide](context.md) for the full `RunContextWrapper` surface, shared usage tracking, nested `tool_input`, and serialization caveats.

As mentioned in the line, we don't intend to repeat the complete example there.

PR #3896, opened 2026-07-20
 docs: wrap bare await usage in async def main() in code examples  

AAliKKhan:
This pull request fixes 5 code examples across `docs/agents.md` and `docs/tools.md` that used `await Runner.run(...)` at module level outside any `async def` function, which raises `SyntaxError` when run as a Python script. Each example now wraps the execution in `async def main(): 
 ` with `asyncio.run(main())`, matching the convention already used in the Quickstart and running_agents.md pages.


seratch:
We don't want to have main function for all code snippets in docs.

PR #3897, opened 2026-07-20
fix(chatcmpl): surface content-filter refusals when buffering streamed tool calls

gunjanjaswal:
#3769 made a content-filtered stream surface a `ResponseOutputRefusal` instead of an empty turn, but only on the non-buffered path. With `buffer_streamed_tool_calls=True` the refusal never fires.

`buffer_tool_call_stream` forwards a choice only when its delta carries output. A terminal chunk like `Choice(index=0, delta=ChoiceDelta(), finish_reason="content_filter")` has an empty delta, so it gets dropped before the handler ever sees it. `saw_content_filter` is never set and the synthesis block added in #3769 never runs. The caller gets a silently empty turn, which is the exact failure mode that PR set out to fix.

The fix forwards a delta-stripped copy of a content-filtered terminal choice, so the `finish_reason` reaches the handler while buffering semantics stay unchanged.

Reachable through `OpenAIChatCompletionsModel`, and also via `MultiProvider(openai_buffer_streamed_tool_calls=True)` and `OpenAIProvider`.

Tests mirror the content-filter cases from `test_litellm_chatcompletions_stream.py` with buffering on, plus coverage of the buffering layer itself and a guard against duplicating the `tool_calls` finish. The gap went untested because LiteLLM never buffers. Checked against `upstream/main` that the new tests fail without the source change and pass with it. Full suite green (5354 passed), ruff and mypy clean.

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -377,15 +377,11 @@ class ChatCmplStreamHandler:
 
                 if has_passthrough_output:
                     passthrough_choices.append(choice)
-                elif choice.finish_reason is not None and not cls._choice_finished_tool_calls(
-                    choice
-                ):
-                    # A terminal choice can carry stream-level signals the handler still
-                    # needs even when its delta is empty: finish_reason == "content_filter"
-                    # drives the synthesized refusal, and "length" marks a truncated turn.
-                    # Forward a delta-stripped copy so buffering semantics are unchanged
-                    # while the finish_reason reaches the handler. "tool_calls" is excluded
-                    # because the buffered tool-call chunk below carries it instead.
+                elif choice.finish_reason == "content_filter":
+                    # A content-filtered choice ends the stream with an empty delta, so it
+                    # would otherwise be dropped here and the handler would never see the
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Before merging, please narrow the passthrough condition and related tests to finish_reason == "content_filter". The current patch also forwards length and stop, but handle_stream() does not interpret either value, so those cases do not change the normalized events or completed response. Please remove those claims and cases while retaining the end-to-end content-filter coverage and the tool_calls non-duplication test. After that focused change, this should be ready for another review.

gunjanjaswal:
Good call, narrowed it to `content_filter` only. You are right that `handle_stream()` never interprets `length` or `stop`, so forwarding those was widening the change for nothing observable.

Dropped the two parametrized cases along with them and removed the `length` claim from the description. The end-to-end content-filter coverage and the `tool_calls` non-duplication test are still there. Verified against `upstream/main` that the remaining tests fail without the handler change.

seratch (approved):

gunjanjaswal:
Thanks for the review and merge, @seratch, and for catching the `length`/`stop` overreach. Narrowing it to `content_filter` made the change a lot cleaner.

PR #3900, opened 2026-07-20
docs: fix missing asyncio.run call in hosted tools example

AAliKKhan:
## What                                                                                                                                                                                                                                                                                                                               
     The hosted tools code block defines `async def main()` but never calls it, and `import asyncio` is missing.
                                                                                                                                                                                                                                                                                                                                           
     ## Why                                                                                                                                                                                                                                                                                                      
     A reader copying the example will see no output — the code defines `main()` but never executes it.                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                       
     ## Fix                                                                                                                                                                                                                                                                                            
     Add `import asyncio` and `asyncio.run(main())` after the `main()` definition.


seratch:
@AAliKKhan As mentioned at https://github.com/openai/openai-agents-python/pull/3896#issuecomment-5027652211, we don't need this change. Could you refrain from sending the same PR? Thanks again for your interest here.

PR #3903, opened 2026-07-21
docs: fix conversation example imports

william-xue:
## What

Add the missing SDK imports to both conversation-management examples in the running agents guide.

## Why

The manual example references `Agent`, `Runner`, and `trace` without importing them, while the session example imports the agent and session classes but still uses `trace` without importing it. Copying either snippet currently raises `NameError`.

## Test

- `uv run mkdocs build`
- `uv run python` import check for `Agent`, `Runner`, `SQLiteSession`, and `trace`
- `git diff --check`

### DIFF
--- a/docs/running_agents.md
+++ b/docs/running_agents.md
@@ -303,7 +303,6 @@ You can manually manage conversation history using the [`RunResultBase.to_input_
 ```python
 from agents import Agent, Runner, trace
 
-
 async def main():
     agent = Agent(name="Assistant", instructions="Reply very concisely.")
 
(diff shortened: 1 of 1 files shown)


docs/running_agents.md
```diff
@@ -301,6 +301,9 @@ At the end of the agent run, you can choose what to show to the user. For exampl
 You can manually manage conversation history using the [`RunResultBase.to_input_list()`][agents.result.RunResultBase.to_input_list] method to get the inputs for the next turn:
 
 ```python
+from agents import Agent, Runner, trace
+
```
seratch, line 305:
```suggestion
```

seratch:
@codex review

PR #3908, opened 2026-07-22
feat(run_state): add opt-in HMAC integrity for persisted RunState

ioleksiuk:
### Summary

`RunState` is a documented persist/resume boundary — the HITL example serializes paused state with `to_json()` and resumes it later with `from_json()` — but `from_json` restores the serialized blob verbatim with no authenticity check. This adds an **opt-in** integrity option:

- `to_json(…, signing_key=…)` / `to_string(…, signing_key=…)` append a top-level `$signature` = HMAC-SHA256 over the canonical serialization of the state.
- `from_json(…, signing_key=…)` / `from_string(…, signing_key=…)` verify it and raise `UserError` on mismatch or absence (fails closed).

Fully backward compatible: without a `signing_key`, behavior is unchanged; `$signature` is an additive envelope field that older readers ignore, so no `$schemaVersion` bump is required.

### Test plan

Added `TestRunStateSigning`: signed round-trip (dict + string), tamper rejection (forged approval), wrong-key rejection, unsigned-rejected-when-key-required, and signature-ignored-without-key (back-compat). `make format`, `make lint`, `make typecheck`, and the full `tests/test_run_state.py` suite (246 tests) pass.

### Checks

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 import asyncio
 import copy
 import dataclasses
+import hashlib
+import hmac
 import json
 import threading
 from collections import deque
@@ -727,6 +729,7 @@ class RunState(Generic[TContext, TAgent]):
         context_serializer: ContextSerializer | None = None,
         strict_context: bool = False,
         include_tracing_api_key: bool = False,
+        signing_key: str | bytes | None = None,
     ) -> dict[str, Any]:
         """Serializes the run state to a JSON-compatible dictionary.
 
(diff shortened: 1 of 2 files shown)


seratch:
I traced this use case through the RunState persistence boundary. RunState serializes durable runtime data, but it does not promise authentication of snapshots loaded from an untrusted store. Applications crossing that trust boundary can already store `to_string()` as an opaque authenticated or encrypted envelope and verify it before calling `from_string()`. This also keeps key rotation, replay protection, and trust policy at the application boundary.

The proposed API adds a cryptographic and persisted-state contract without demonstrating a supported scenario that the existing envelope approach cannot handle. It also cannot provide the stated "safe to load from an untrusted store" guarantee because verification remains optional and HMAC does not prevent replay. I am going to close this PR for now. We can revisit the design if there is a concrete deployment constraint that makes an application-owned authenticated envelope insufficient.

ioleksiuk:
> I traced this use case through the RunState persistence boundary. RunState serializes durable runtime data, but it does not promise authentication of snapshots loaded from an untrusted store. Applications crossing that trust boundary can already store `to_string()` as an opaque authenticated or encrypted envelope and verify it before calling `from_string()`. This also keeps key rotation, replay protection, and trust policy at the application boundary.
> 
> The proposed API adds a cryptographic and persisted-state contract without demonstrating a supported scenario that the existing envelope approach cannot handle. It also cannot provide the stated "safe to load from an untrusted store" guarantee because verification remains optional and HMAC does not prevent replay. I am going to close this PR for now. We can revisit the design if there is a concrete deployment constraint that makes an application-owned authenticated envelope insufficient.

Thank you for the review! @seratch

Issue #3861, opened 2026-07-17
Non-buffered Chat Completions tool-call streaming finalizes incomplete tool calls with empty name/call_id


baiyuxi930826:
### Please read this first

- **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-python/)
- **Have you searched for related issues?** Searched for Chat Completions streaming / `buffer_streamed_tool_calls` / incomplete tool-call handling; did not find an existing report of this default-vs-buffered asymmetry.

### Describe the bug

For OpenAI Chat Completions **streaming** tool calls, the default path (`buffer_streamed_tool_calls=False`) and the buffered path (`buffer_streamed_tool_calls=True`) disagree on incomplete tool-call streams:

| Path | Incomplete tool call (missing `id` and/or function `name`) |
|---|---|
| **Default / non-buffered** | Still finalizes a `ResponseFunctionToolCall` with empty `name` and/or `call_id` via the end-of-stream fallback |
| **Buffered** | Raises `ModelBehaviorError` (`without a tool call id` / `without a function name`) |

The same incomplete upstream stream can either hard-fail or silently produce an unusable (or partially usable) tool call, depending only on a flag that defaults to `False`. LiteLLM/AnyLLM integrations always use the non-buffered path today.

This appears to come from `ChatCmplStreamHandler.handle_stream`:

1. On first tool-call delta for an index, it seeds `ResponseFunctionToolCall(name="", call_id="", ...)`.
2. Live streaming only starts once **both** `name` and `call_id` are present.
3. At stream end, if streaming never started, the "fallback to old behavior" branch still emits `response.output_item.added` / arguments delta / `response.output_item.done` with those empty fields.

The buffered path validates completeness in `_buffered_tool_call_delta` and raises instead (added in PR #3506).

### Debug information

- Agents SDK version: `0.18.3` (checkout `965335aba6f6c71500e0b8cdb4e9e495f5801d4d`)
- Python version: Python 3.12.10
- OS: Windows 11
- API surface: Chat Completions streaming (`OpenAIChatCompletionsModel` / `OpenAIProvider(use_responses=False)`)
- Flag: `buffer_streamed_tool_calls` default `False` vs `True`
- Reproduction count: 2/2 stable (unit-level synthetic stream; no live API key)

### Repro steps

Minimal unit-style repro against the stream handler (no live API key required):

```python
import asyncio
from typing import AsyncIterator

from openai.types.chat.chat_completion_chunk import (
[...]

PR #3912, opened 2026-07-22
fix: guard incomplete tool calls in non-buffered Chat Completions stream (#3861)

rkfshakti:
The non-buffered path in handle_stream creates ResponseFunctionToolCall
with empty name and call_id when a stream ends without providing those
fields, then emits it via the fallback-to-old-behavior branch. The
buffered path already raises ModelBehaviorError for the same condition.

Add the same guard in the non-buffered fallback path: raise
ModelBehaviorError when name or call_id is missing, matching the
buffered path's behavior and preventing silent emission of unusable
tool calls.

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -48,6 +48,8 @@ from openai.types.responses.response_reasoning_text_done_event import (
 )
 from openai.types.responses.response_usage import OutputTokensDetails
 
+from agents.exceptions import ModelBehaviorError
+
 from ..exceptions import ModelBehaviorError, UserError
 from ..items import TResponseStreamEvent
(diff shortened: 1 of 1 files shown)


seratch:
We previously asked in #3861 to hold off on opening a PR until there is a redacted stream or transport-level reproduction from a real provider, including the provider, model, adapter, and observed downstream failure. That evidence is still missing.

This change also breaks released fallback behavior: several existing streaming tests intentionally finalize tool calls without an ID, and the new unconditional check would make those supported scenarios raise instead. Strict validation is already available through buffer_streamed_tool_calls=True.

I’m going to close this PR for now. If a real provider reproduction demonstrates an actual user-facing failure, we can revisit the narrowest compatible fix at the appropriate adapter or streaming boundary.

PR #3914, opened 2026-07-22
fix: resolve voice pipeline deadlock, remove redundant input deep copy, and retrieve background task exceptions

AAliKKhan:
## Summary

This pull request fixes three issues affecting the voice pipeline and model input processing.

### 1. Prevent voice pipeline deadlock on TTS errors

When `_stream_audio()` encounters a TTS error, it emits a `session_ended` event but does not send the `None` sentinel expected by `_dispatch_audio()`. As a result, `_dispatch_audio()` blocks indefinitely waiting on `local_queue.get()`, preventing the voice pipeline from shutting down correctly and causing `_wait_for_completion()` and the caller's `stream()` iteration to hang.

**Fix:**

* Send `None` to `local_queue` before re-raising the exception so the consumer exits cleanly.

### 2. Remove redundant deep copy of model input history

`ItemHelpers.input_to_new_input_list()` already calls `_to_dump_compatible()` internally. `OpenAIResponsesModel._create_request()` was performing the same conversion beforehand, resulting in an unnecessary second deep copy of the entire input history on every model request.

For multi-turn conversations with large histories, this introduced avoidable CPU and memory overhead.

**Fix:**

* Remove the redundant `_to_dump_compatible()` call and rely on `ItemHelpers.input_to_new_input_list()` to perform the conversion.

### 3. Retrieve exceptions from the background text generation task

`_check_errors()` only inspected `_stream_audio` tasks, ignoring `text_generation_task`. If the background pipeline task raised an exception, it was never retrieved, causing asyncio to emit a `Task exception was never retrieved` warning when the task was garbage collected.

**Fix:**

* Include `text_generation_task` in `_check_errors()` so background exceptions are properly retrieved and propagated.

## How to test

### Voice pipeline deadlock

1. Simulate a TTS failure during audio streaming.
2. Verify the voice session exits cleanly instead of hanging.
3. Confirm `_wait_for_completion()` returns and the caller's `stream()` iteration terminates normally.

### Redundant deep copy

1. Run a multi-turn conversation with a large input history.
2. Verify requests are processed correctly.
3. Confirm no behavior changes while avoiding the redundant deep copy.

### Background task exceptions

1. Force `text_generation_task` to raise an exception.
2. Verify the exception is surfaced through `_check_errors()`.
3. Confirm no `Task exception was never retrieved` warning appears.

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -733,7 +733,6 @@ class OpenAIResponsesModel(Model):
         prompt: ResponsePromptParam | None = None,
     ) -> dict[str, Any]:
         list_input = ItemHelpers.input_to_new_input_list(input)
-        list_input = _to_dump_compatible(list_input)
         list_input = self._remove_openai_responses_api_incompatible_fields(list_input)
 
         if model_settings.parallel_tool_calls and tools:
--- a/src/agents/voice/result.py
+++ b/src/agents/voice/result.py
@@ -196,6 +196,7 @@ class StreamedAudioResult:
 
                 # Signal completion for whole session because of error
                 await local_queue.put(VoiceStreamEventLifecycle(event="session_ended"))
+                await local_queue.put(None)
                 raise e
 
     async def _add_text(self, text: str):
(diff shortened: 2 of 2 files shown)


AAliKKhan (commented):
check it out now

seratch (changes requested):
This PR currently combines unrelated changes: voice-pipeline error handling and a Responses API input-copy optimization. Please split these into separate PRs so each change can be reviewed, tested, and merged independently. The TTS deadlock fix and background-task exception handling can remain together if they form one coherent voice-pipeline failure-handling change.

PR #3922, opened 2026-07-23
fix: remove redundant deep copy in input preprocessing

AAliKKhan:
ItemHelpers.input_to_new_input_list() already calls _to_dump_compatible() internally (items.py:772). The caller in OpenAIResponsesModel._create_request() was calling it a second time, creating an unnecessary deep copy of the entire input history on every model turn. For multi-turn conversations with large histories this doubles memory and CPU overhead.

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -733,7 +733,6 @@ class OpenAIResponsesModel(Model):
         prompt: ResponsePromptParam | None = None,
     ) -> dict[str, Any]:
         list_input = ItemHelpers.input_to_new_input_list(input)
-        list_input = _to_dump_compatible(list_input)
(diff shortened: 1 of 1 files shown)


seratch (changes requested):
Thanks for the optimization. `ItemHelpers.input_to_new_input_list()` already recursively materializes list inputs, so removing the second conversion is appropriate. However, `test_responses_materializes_iterator_payload` currently monkeypatches that helper to return an unmaterialized iterable and therefore depends on the conversion this PR removes. Please update the test to pass the nested iterable through the real `ItemHelpers.input_to_new_input_list()` implementation, while preserving the assertions that input and tool iterables are materialized exactly once. Once that existing regression coverage passes, this should be ready for another review.

PR #3924, opened 2026-07-23
fix: prevent list aliasing in _populate_state_from_result

AAliKKhan:
In _populate_state_from_result, when _model_input_items is not set on the result, state._generated_items was assigned a bare reference to result.new_items instead of a copy. Since state._generated_items can be mutated (e.g., appending items during interruption approval), this silently corrupted result.new_items — the primary public API for consuming agent output. The fix makes a defensive copy via list().

### DIFF
--- a/src/agents/result.py
+++ b/src/agents/result.py
@@ -120,7 +120,7 @@ def _populate_state_from_result(
     if isinstance(model_input_items, list):
         state._generated_items = list(model_input_items)
     else:
-        state._generated_items = result.new_items
+        state._generated_items = list(result.new_items)
     state._session_items = list(result.new_items)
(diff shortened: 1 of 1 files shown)


seratch:
I traced both supported to_state() paths, and RunResult and RunResultStreaming always initialize _model_input_items as a list. Interruption results also explicitly populate it before they are returned, so _populate_state_from_result() already takes the branch that defensively copies the generated items. The modified fallback is therefore unreachable for supported SDK-created results, and the PR does not currently demonstrate an observable issue. I am going to close this PR. If you can provide a reproduction using an unmodified SDK-created result where _model_input_items is absent or not a list, we can revisit it.

Issue #3317, opened 2026-05-09
ensure_strict_json_schema({}) returns a shared mutable schema


Aphroq:
### Please read this first

- **Have you read the docs?** Yes. This is about the SDK's strict schema normalization helper.
- **Have you searched for related issues?** Yes. No duplicate was found. PR #3199 is adjacent because it also involves `ensure_strict_json_schema` in-place mutation, but it only isolates MCP non-strict fallback from partial mutation after conversion errors. It does not cover the empty-schema fast path returning a shared module-level object.

### Describe the bug

`ensure_strict_json_schema({})` returns the module-level `_EMPTY_SCHEMA` dictionary directly. Because that return value contains mutable nested fields such as `properties` and `required`, any caller that mutates the returned schema can pollute later calls with a fresh `{}` input.

Impact: code paths that normalize an empty schema can leak state across otherwise independent tool/output/handoff schema conversions. A later call can receive fields that came from a previous caller instead of the canonical empty strict object schema.

### Debug information

- Agents SDK version: upstream `main` at `bc3607ba`
- Python version: Python 3.12.1

### Repro steps

Run this script against current upstream `main`:

```python
from agents.strict_schema import ensure_strict_json_schema


first = ensure_strict_json_schema({})
first["properties"]["polluted"] = {"type": "string"}
second = ensure_strict_json_schema({})

print(f"same_object: {first is second}")
print(f"second_schema: {second}")
```

Actual result:

```text
same_object: True
second_schema: {'additionalProperties': False, 'type': 'object', 'properties': {'polluted': {'type': 'string'}}, 'required': []}
```

### Expected behavior

Each `ensure_strict_json_schema({})` call should return an independent schema object equivalent to:

```python
{
    "additionalProperties": False,
    "type": "object",
    "properties": {},
    "required": [],
}
```

Mutating one returned empty-schema result should not affect later calls.

Issue #345, opened 2025-03-26
Error if function_tool has a `Mapping` input


TimoVink:
### Describe the bug
If an agent is provided with a `function_tool` that has a `Mapping` or `dict` as its input, an HTTP 400 `invalid_function_parameters` error is returned when the agent runs.

### Example
```python
@function_tool
def foo(x: Mapping[str,str]) -> str:
    return 'bar'

foo_agent = Agent(name='Foo Agent', tools=[foo])
await Runner.run(foo_agent, input='foo')
```

Error:
```
Error getting response: Error code: 400
{
    'error': {
        'message': "Invalid schema for function 'foo': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Extra required key 'x' supplied.", 
        'type': 'invalid_request_error',
        'param': 'tools[0].parameters',
        'code': 'invalid_function_parameters'
    }
}
(request_id: req_f3213fe15078c00d899a910c5567fed4)
```

If I check `foo.params_json_schema`, I get:
```
{'properties': {'x': {'additionalProperties': {'type': 'string'},
   'title': 'X',
   'type': 'object'}},
 'required': ['x'],
 'title': 'foo_args',
 'type': 'object',
 'additionalProperties': False}
```
Which looks valid to me at a glance, but I'm not very familiar with the JSON Schema spec.

### Debug information
- Agents SDK: `v0.0.6`
- Python: `v3.11.9`
- Pydantic: `v2.10.6`

DanieleMorotti:
Hi, in short, you can fix the error by creating a Pydantic object such as:
```python
from pydantic import BaseModel

class MyObject(BaseModel):
    parameter_name: str
```

and then you declare it as the type of the parameter:
```python
@function_tool
def foo(x: MyObject) -> str:
...
```

Because there are several issues regarding support for generic `dict` or `Mapping` types instead of objects.

TimoVink:
Hi, thanks for the response.

My original goal was to model something easily coercible to a `pd.Series`, but in a way that gives more type hinting than just using something like `list[list[Any]]`

With `pd.Series` itself not working, nor `Tuple` (#302) nor `Mapping`, I am now doing something like this:

```
class DataPoint(BaseModel):
    x: str
    y: float

class DataSeries(BaseModel):
    title: str
    data: list[DataPoint]
```

... which works but feels more verbose than necessary. Though perhaps this is actually clearer for the model than anything else I could do.

[...]

PR #3927, opened 2026-07-23
fix(strict_schema): reject empty additionalProperties mappings

hsusul:
### Summary

`ensure_strict_json_schema()` failed to reject `additionalProperties: {}`, a common OpenAPI/MCP encoding for “allow any extra properties”. Because `{}` is falsy in Python, the existing truthiness check treated it as strict-safe and left a non-strict schema in place.

This completes the intent of #356: any `additionalProperties` value other than `False` on an object schema must raise `UserError`.

### Root cause

In #356 the guard was changed from `is True` to a truthiness check so dict/`Mapping` schemas would be rejected. Empty mappings (`{}`) are falsy, so they still slipped through even though they mean “allow anything” in JSON Schema.

### Reproduction

```python

schema = {
    "type": "object",
    "properties": {"a": {"type": "string"}},
    "additionalProperties": {},
}
ensure_strict_json_schema(schema)
# Before: returns schema with additionalProperties still {}
# After: raises UserError
```

With MCP `convert_schemas_to_strict=True`, the previous behavior could mark tools as `strict_json_schema=True` while still shipping `additionalProperties: {}`, which OpenAI structured outputs reject. Raising now correctly triggers the existing non-strict fallback path in `MCPUtil.to_function_tool`.

### Solution

Compare with `is not False` instead of truthiness, matching the documented contract that only `False` is allowed under strict conversion.

### Test plan

- [x] Added regression tests for `additionalProperties: {}`, a non-empty schema object, and explicit `False`
- [x] Ran `.agents/skills/code-change-verification/scripts/run.sh` (`make format`, `make lint`, `make typecheck`, `make tests`)

### Issue number

N/A — previously unreported gap left by #356 / #345.

### Checks

### Compatibility and risk

- Low risk, one-condition change.
- Schemas that previously kept `additionalProperties: {}` under strict conversion now raise `UserError`, which is the intended strict-mode behavior (same as `True` / non-empty schemas).
- MCP callers using `convert_schemas_to_strict=True` fall back to non-strict with the original schema instead of sending an invalid strict schema to the API.

### Related prior work

- #356 / #345: intended to reject non-`False` `additionalProperties` values, including dict/Mapping forms
- #3317 / #3318: adjacent empty-schema mutation fix (different bug)

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -90,7 +90,10 @@ def _ensure_strict_json_schema(
     elif (
         typ == "object"
         and "additionalProperties" in json_schema
-        and json_schema["additionalProperties"]
+        # Compare with ``is not False`` rather than truthiness: OpenAPI/MCP schemas often use
+        # ``additionalProperties: {}`` (an empty schema meaning "allow anything"). That value is
+        # falsy in Python, so a truthiness check would silently leave a non-strict schema in place.
+        and json_schema["additionalProperties"] is not False
     ):
         raise UserError(
             "additionalProperties should not be set for object types. This could be because "
--- a/tests/test_strict_schema.py
+++ b/tests/test_strict_schema.py
@@ -56,6 +56,40 @@ def test_object_with_true_additional_properties():
         ensure_strict_json_schema(schema)
 
 
+def test_object_with_empty_dict_additional_properties():
+    # OpenAPI/MCP schemas commonly use ``additionalProperties: {}`` to mean "allow anything".
+    # That empty mapping is falsy in Python, but it is still non-strict and must be rejected.
+    schema = {
+        "type": "object",
+        "properties": {"a": {"type": "string"}},
+        "additionalProperties": {},
+    }
+    with pytest.raises(UserError):
+        ensure_strict_json_schema(schema)
+
+
+def test_object_with_schema_additional_properties():
+    # A non-empty additionalProperties schema is also non-strict and must be rejected.
+    schema = {
+        "type": "object",
+        "properties": {"a": {"type": "string"}},
+        "additionalProperties": {"type": "string"},
+    }
+    with pytest.raises(UserError):
+        ensure_strict_json_schema(schema)
+
+
+def test_object_with_false_additional_properties_is_allowed():
+    schema = {
+        "type": "object",
+        "properties": {"a": {"type": "string"}},
+        "additionalProperties": False,
+    }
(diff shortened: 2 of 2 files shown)


ebarkhordar (commented):
Good catch closing the `{}` gap left by #356. `{}` is falsy, so the old truthiness check let it slip through, and `is not False` is the right condition.

One design question. A few lines up, an object with no `additionalProperties` key has it normalized to `False`:

```python
if typ == "object" and "additionalProperties" not in json_schema:
    json_schema["additionalProperties"] = False
```

So an absent constraint is coerced to strict, but an explicit `additionalProperties: {}`, which means the same thing ("allow any extra properties"), now raises `UserError`. For the OpenAPI/MCP tools this PR calls out, that is the difference between the tool converting under strict mode and being unusable in it.

Would coercing `{}` to `False` here, the same way absent is, fit the transform-to-strict intent better than rejecting? Or is raising deliberate, so an explicit "allow anything" surfaces to the caller instead of being silently overridden? Either way the truthiness bug itself is real and worth fixing.

hsusul:
Thanks, good question. I checked the repo docs and the intent seems to be that missing additionalProperties gets normalized to False, but explicit open forms should be rejected instead of silently changed.

Since additionalProperties: {} means the same thing as true, converting it to False would change what inputs the schema allows. That also matches #356 and the MCP fallback behavior.

So I think rejecting {} is the better fit here, but I’m happy to change it if you intended strict conversion to deliberately narrow those schemas.

> Good catch closing the `{}` gap left by #356. `{}` is falsy, so the old truthiness check let it slip through, and `is not False` is the right condition.
> 
> One design question. A few lines up, an object with no `additionalProperties` key has it normalized to `False`:
> 
> ```python
> if typ == "object" and "additionalProperties" not in json_schema:
>     json_schema["additionalProperties"] = False
> ```
> 
> So an absent constraint is coerced to strict, but an explicit `additionalProperties: {}`, which means the same thing ("allow any extra properties"), now raises `UserError`. For the OpenAPI/MCP tools this PR calls out, that is the difference between the tool converting under strict mode and being unusable in it.
> 
> Would coercing `{}` to `False` here, the same way absent is, fit the transform-to-strict intent better than rejecting? Or is raising deliberate, so an explicit "allow anything" surfaces to the caller instead of being silently overridden? Either way the truthiness bug itself is real and worth fixing.

seratch (approved):

PR #3929, opened 2026-07-23
fix(litellm): send logprobs=True when top_logprobs is set

dimaosipa:
### Summary

- the litellm and any_llm chat adapters forward `top_logprobs` but never set the companion `logprobs=True`. the Chat Completions API (OpenAI, Azure) rejects `top_logprobs` unless `logprobs` is true, so `ModelSettings(top_logprobs=N)` returns a 400 on these adapters, while the identical request works through `OpenAIChatCompletionsModel`.
- fix: mirror the collision-safe guard the OpenAI chat adapter already uses (openai_chatcompletions.py). set `logprobs=True` when `top_logprobs` is set, unless the caller already supplied `logprobs` via `extra_args`.
- applies to both litellm_model.py and any_llm_model.py (chat path).

### Test plan

- new tests/models/test_litellm_logprobs.py plus two tests in tests/models/test_any_llm_model.py: logprobs flag is set when top_logprobs is set, omitted when unset, and no collision when the caller passes logprobs via extra_args. mirrors the existing `test_chat_completions_top_logprobs_sets_logprobs_flag`.
- `make format` and `make lint`: clean.
- `make tests`: full suite green (5443 + 27 passed, 0 failed).
- `make typecheck`: adds 0 new errors (my changed files are clean).
- ran `.agents/skills/code-change-verification/scripts/run.sh`: passes format, then stops at typecheck on 31 pre-existing errors unrelated to this change (vercel extension name-collision, py3.12-only apis on a 3.11 checkout, boto3 not installed, redis `type: ignore` skew).

### Checks

### DIFF
--- a/src/agents/extensions/models/any_llm_model.py
+++ b/src/agents/extensions/models/any_llm_model.py
@@ -731,6 +731,12 @@ class AnyLLMModel(Model):
         extra_kwargs = self._build_chat_extra_kwargs(model_settings)
         extra_kwargs.pop("reasoning_effort", None)
 
+        # The Chat Completions API requires logprobs=True whenever top_logprobs is set. Defer to a
+        # caller-supplied logprobs (via extra_args, already merged into extra_kwargs) to avoid a
+        # duplicate-key collision.
+        if model_settings.top_logprobs is not None and "logprobs" not in extra_kwargs:
+            extra_kwargs["logprobs"] = True
+
         ret = await self._get_provider().acompletion(
             model=self._provider_model,
             messages=converted_messages,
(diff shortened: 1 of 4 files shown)


seratch (changes requested):
Thanks for addressing the missing Chat Completions flag. Before merging, please also preserve `choice.logprobs.content` in non-streaming LiteLLM output, using the same conversion and attachment behavior already present in `AnyLLMModel` and `OpenAIChatCompletionsModel`. Currently the new flag makes the request succeed, but `LitellmModel.get_response()` discards the returned probabilities, so callers still cannot access the data they requested. Please add a regression test containing actual returned token log probabilities and assert that they appear in `ResponseOutputText.logprobs`.

seratch (approved):

PR #3933, opened 2026-07-24
fix: enforce realtime text guardrails and synchronize streaming cancellation

seratch:
This pull request fixes two streaming runtime issues affecting output safety and cancellation behavior.

- Run realtime output guardrails for text-only responses, applying the same accumulation, debounce, and interruption behavior already used for audio transcripts.
- Prevent `cancel(mode="after_turn")` from starting another model request while stream consumers are still processing events.
- Replace scheduler-dependent cancellation timing with explicit event-consumption synchronization that safely handles inactive consumers, closed iterators, and immediate cancellation.
- Add regression coverage for fragmented text output, malformed realtime events, delayed cancellation, consumer shutdown, and runs without stream consumers.


src/agents/run_internal/run_loop.py
```diff
@@ -1231,6 +1231,8 @@ async def _save_stream_items_without_count(
                         store_setting,
                     )
 
+                    await streamed_result._wait_for_turn_event_consumption()
```
gnanirahulnutakki: Please apply the same event-consumption gate before every path that can start another model turn, especially `NextStepHandoff`. This wait only covers `NextStepRunAgain`; the handoff branch checks `_cancel_mode` without first giving the active consumer a chance to process its queued `handoff_requested` event.

I reproduced this on `4801764e` with a `FakeModel` subclass that sets an `asyncio.Event` when its second `stream_response()` starts: pause the consumer on `handoff_requested`, wait for that event, then call `cancel(mode="after_turn")`. The event fires before cancellation, and the result ends with `usage.requests == 2` plus Agent2's final output. That violates this method's documented “stops before next turn begins” behavior. Please add a deterministic handoff regression using a controlled event (and cover resumed continuation paths if they can also bypass the gate).

gnanirahulnutakki (commented):
Thanks—this is a strong direction, and the text-only Realtime guardrail path is well scoped. The focused Realtime and soft-cancel suites pass locally (`157 passed`), and the new event names match the Realtime contract.

One blocking cancellation gap remains: event-consumption synchronization is only applied to `NextStepRunAgain`. A handoff can still start the next agent's model request while the consumer is processing the first turn's `handoff_requested` event. On the current head I deterministically observed two model requests and Agent2's final output before `cancel(mode="after_turn")` took effect.

Please gate the handoff (and any equivalent resumed continuation path) before another model turn begins, and add a controlled-interleaving regression test. Once that is covered, the cancellation approach looks sound.

gnanirahulnutakki (commented):
Re-reviewed at `81aa66bc`. The original cancellation gap is addressed: the event-consumption barrier now covers ordinary run-again, handoff, resumed run-again, and resumed handoff continuations before another model request can start. The new controlled-interleaving tests exercise both handoff and resume paths.

Local verification: 169 focused Realtime/cancellation/resume tests passed, Ruff and formatting checks passed, and an independent follow-up-commit review found no issues. GitHub’s current test, lint, typecheck, docs, and Windows matrix is green. I found no remaining issues in the follow-up.

src/agents/run_internal/run_loop.py
gnanirahulnutakki: Verified on `81aa66bc`: the shared barrier now gates handoff and both resumed continuation paths, and the new event-controlled tests prove the next model request stays blocked until the consumer can set `after_turn`. The focused Realtime/cancellation/resume suite passes locally (169 tests). This addresses my finding—thank you.

src/agents/run_internal/run_loop.py
```diff
@@ -1179,9 +1200,7 @@ async def _save_stream_items_without_count(
                     if streamed_result._state is not None:
                         streamed_result._state._current_step = NextStepRunAgain()
 
-                    if streamed_result._cancel_mode == "after_turn":  # type: ignore[comparison-overlap]
-                        streamed_result.is_complete = True
-                        streamed_result._event_queue.put_nowait(QueueCompleteSentinel())
+                    if await _wait_for_streamed_turn_events_and_stop_if_cancelled(streamed_result):
                         break
```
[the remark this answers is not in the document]
gnanirahulnutakki: Confirmed on `81aa66bc` through the public streaming path. I handed off `Triage` → `Delegate`, called `cancel(mode="after_turn")` while consuming `handoff_requested`, then inspected the completed result: `usage.requests == 1` and `final_output is None` (the new barrier works), but `result.current_agent`, `result.last_agent`, and `result.to_state()._current_agent` all remain `Triage`. The local/run-state handoff had already advanced to `Delegate`, so the public result and resumable state diverge exactly as described. Updating the streamed result’s agent/output schema before the wait, with assertions for both ordinary and resumed handoffs, would close this.

gnanirahulnutakki (commented):
Re-reviewed the current head `085f9391` after the force-push. The follow-up closes the remaining public-result state gap: `_publish_streamed_result_agent()` now updates `current_agent` and the output schema before the handoff event-consumption barrier can honor `after_turn` cancellation, and the ordinary plus resumed handoff tests assert `current_agent`, `last_agent`, `to_state()`, and the schema.

Local verification on this exact head: `uv run pytest -q tests/test_soft_cancel.py tests/test_run_impl_resume_paths.py` → **34 passed**. GitHub's current test matrix, typecheck, lint, docs, and Windows checks are all green. I found no remaining issue in the cancellation follow-up.

seratch:
@gnanirahulnutakki Thanks for sharing your review results, but could you refrain from posting general comments? Pointing specific points out would be appreciated.

gnanirahulnutakki:
Understood—thanks for the guidance. I’ll keep future feedback to specific, actionable inline findings and avoid posting general review-summary comments.

PR #3935, opened 2026-07-24
fix(sandbox): single-flight cached dependency factories

cosin2077:
### Summary

This pull request fixes cached sandbox dependency factories so concurrent requests for the same key share one in-flight initialization.

Previously, the cache was populated only after a factory completed, so overlapping `require()` calls could execute the same async factory multiple times and register multiple owned resources. The implementation now tracks one pending task per cached key, shields shared initialization from individual waiter cancellation, clears failed tasks for retry, and prevents stale or post-close results from entering the cache. Container shutdown is also single-flight and waits for tracked cleanup tasks so cancellation, rebind, and close interleavings cannot leak owned results.

Uncached factories continue to execute once per request.

### Test plan

- `uv run pytest -q tests/sandbox/test_dependencies.py` (18 passed)
- Related snapshot/runtime cleanup tests (13 passed)
- `bash .agents/skills/code-change-verification/scripts/run.sh` (format, lint, typecheck, and full test suite passed)
- Reviewed the final diff against `origin/main`; Standards and Spec reviews reported no remaining findings.

### Issue number

No linked issue. A pre-submission search found no open issue or pull request covering this fix.

### Checks

### DIFF
--- a/src/agents/sandbox/session/dependencies.py
+++ b/src/agents/sandbox/session/dependencies.py
@@ -76,7 +76,6 @@ class Dependencies:
         self._cache: dict[DependencyKey, object] = {}
         self._pending: dict[DependencyKey, asyncio.Task[object]] = {}
         self._active_tasks: set[asyncio.Task[object]] = set()
-        self._cleanup_tasks: set[asyncio.Task[None]] = set()
         self._owned_results: list[object] = []
         self._close_task: asyncio.Task[None] | None = None
         self._closed = False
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
One ownership issue blocks merge. A stale or post-close factory result is currently closed immediately via _discard_factory_result(). The same object may already be owned by another binding or returned by the replacement binding; the existing identity deduplication in aclose() shows that aliased owned results are supported. Eagerly closing that object can therefore invalidate surviving work or cause it to be closed twice.
Please route stale and post-close owned results through the container's identity-deduplicated owned-result cleanup instead of closing them eagerly. Please also add controlled rebind and close tests where the stale and surviving/current factories return the same closable object, asserting that the survivor remains usable until shutdown and that the object is closed exactly once. After that focused change, this should be in good shape.

seratch:
We'll continue working on this w/ your contribution credit there. Thanks again for sharing the idea to fix the issue!

PR #3937, opened 2026-07-24
fix(approvals): consult stored status before needs_approval on execute

hsusul:
### Summary

This pull request fixes HITL tool execution so resolved approval status is checked before dynamic `needs_approval` checkers run.

On resume (and for sticky approve/reject decisions), tool execution previously awaited `needs_approval` again before reading stored status. A checker that raises, hits the network, or returns a different result could abort or bypass an already-approved/rejected call. Planning helpers already short-circuited correctly (#3229 / #3259); the execute path for function tools, shell/custom/apply_patch actions, and realtime tool approval did not.

Changes:
- Check `get_approval_status` first in `_maybe_execute_tool_approval`, shell/custom/apply_patch `Action.execute`, and realtime `_maybe_request_tool_approval`
- Only evaluate `needs_approval` when status is still unresolved
- Add a regression covering resume with a raising callable checker
- Update MCP callable-approval tests that asserted the old double-invocation behavior

### Test plan

- [x] Added `test_execute_path_skips_needs_approval_checker_when_status_resolved`
- [x] Confirmed the new test failed on unpatched `main` and passed after the fix
- [x] Ran `.agents/skills/code-change-verification/scripts/run.sh` (format, lint, typecheck, tests)

### Issue number

N/A — previously unreported; completes the execute-path half of the contract fixed for planning in #3229 / #3259.

### Checks

### DIFF
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -607,12 +607,8 @@ class RealtimeSession(RealtimeModelListener):
             tool_lookup_key=tool_lookup_key,
         )
 
-        needs_approval = await self._function_needs_approval(function_tool, tool_call)
-        if self._closing or self._closed:
-            return None
-        if not needs_approval:
-            return True
-
+        # Resolved approve/reject decisions are authoritative; do not re-await
+        # needs_approval for a call whose status is already stored.
         approval_status = self._context_wrapper.get_approval_status(
             function_tool.name,
             tool_call.call_id,
@@ -623,6 +619,12 @@ class RealtimeSession(RealtimeModelListener):
             return True
         if approval_status is False:
             return False
+
+        needs_approval = await self._function_needs_approval(function_tool, tool_call)
+        if self._closing or self._closed:
+            return None
+        if not needs_approval:
+            return True
 
         if self._pre_approval_tool_input_guardrails_enabled():
             rejected_message = await self._run_tool_input_guardrails(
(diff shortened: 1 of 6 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying issue is valid, and checking stored approval status before invoking a dynamic needs_approval policy is the correct contract.

Before merging, please add failing-on-base regression coverage for each independently changed execution path: function, shell, custom, apply_patch, and Realtime. At minimum, cover the safety-relevant sticky-rejection case where always_reject=True has been recorded and the callable later returns False; the tool must remain rejected and its executor must not run. A parameterized test for the Runner paths plus a focused Realtime test is sufficient.

The implementation direction otherwise looks sound, and no documentation or public API changes are required.

hsusul:
Thanks for the review! I’ve added the requested regression coverage for all independently changed execution paths.

The Runner coverage is parameterized across function, shell, custom, and apply_patch, and I added a focused Realtime test. Each case verifies that a previously recorded always_reject=True decision remains authoritative when the later dynamic needs_approval callable returns False, and that neither the checker nor the underlying executor is invoked.

I also verified that the five new cases fail against the pre-fix execute-path behavior and pass with this change. Formatting, linting, type checking, and the full test suite are passing.

seratch:
This branch is now substantially behind main, and the current Realtime path also needs a post-await status check so a concurrent sticky decision cannot be missed while another call is evaluating its checker. I am going to carry the fix forward as a smaller maintainer patch on current main, preserving your contribution with a Co-authored-by trailer, and close this PR in favor of that replacement. Thank you for finding and validating this issue.

PR #3940, opened 2026-07-24
docs: fix streamed audio API rendering

william-xue:
## What
- add the explicit `None` return annotation to `StreamedAudioInput.add_audio`
- align the `None` continuation line so Griffe renders it as part of the `audio` parameter description

## Why
The generated API docs currently report confusing indentation for this docstring, which can detach the end-of-stream behavior from the parameter description.

## Test
- `uv run pytest tests/voice/test_input.py -q` (8 passed)
- `uv run ruff check src/agents/voice/input.py`
- `uv run pyright src/agents/voice/input.py`
- `uv run mkdocs build` (passed; the `voice/input.py` indentation warning is gone)

The repository-wide verification script was also attempted; it is currently blocked in this checkout by missing optional Runloop/LiteLLM/SQLAlchemy/Temporal dependencies and unrelated realtime test failures.

### DIFF
--- a/src/agents/voice/input.py
+++ b/src/agents/voice/input.py
@@ -81,11 +81,11 @@ class StreamedAudioInput:
     def __init__(self):
         self.queue: asyncio.Queue[npt.NDArray[np.int16 | np.float32] | None] = asyncio.Queue()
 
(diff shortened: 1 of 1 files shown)


seratch:
>which can detach the end-of-stream behavior from the parameter description.

this is not true, but the changes here are fine

Issue #2727, opened 2026-03-20
OpenAIResponsesCompactionSession + reasoning breaks with gpt-5.4 (compact output retains orphaned assistant message IDs)


ress:
### Describe the bug

`responses.compact` with **gpt-5.4** strips reasoning items from its output but retains assistant messages with their original IDs. When the compacted output is sent back to `responses.create`, the API rejects it because each assistant message ID implies a required paired reasoning item that was removed during compaction.

**This is gpt-5.4 specific.** The same workflow with gpt-5.2 works fine — gpt-5.2's `responses.compact` does not retain assistant message IDs in its output (0 assistant messages with IDs vs 3 for gpt-5.4).

This means **any conversation using reasoning + compaction with gpt-5.4 is permanently broken after the first compaction**.

`reasoning_item_id_policy="omit"` does not help because it strips IDs from reasoning items, not from the assistant messages that reference them.

### Debug information
- Agents SDK version: v0.12.5 (also reproduced on v0.10.5)
- Python version: Python 3.13.5
- OpenAI SDK version: 2.29.0
- Model: gpt-5.4 (gpt-5.2 works correctly)

#### Minimal repro (adapts SDK's own `examples/memory/compaction_session_example.py`)

With **gpt-5.4** — fails at turn 5 (first turn after compaction):

```python
import asyncio

from agents import Agent, ModelSettings, OpenAIResponsesCompactionSession, Runner, SQLiteSession
from agents.model_settings import Reasoning


async def main():
    underlying = SQLiteSession(":memory:")
    session = OpenAIResponsesCompactionSession(
        session_id="demo-session",
        underlying_session=underlying,
        model="gpt-5.4",
        should_trigger_compaction=lambda ctx: len(ctx["compaction_candidate_items"]) >= 4,
    )

    agent = Agent(
        name="Assistant",
        instructions="Reply concisely. Keep answers to 1-2 sentences.",
        model="gpt-5.4",
        model_settings=ModelSettings(reasoning=Reasoning(effort="low")),
    )

    prompts = [
        "What is the tallest mountain in the world?",
        "How tall is it in feet?",
        "When was it first climbed?",
        "Who was on that expedition?",  # Compaction triggers after this turn
        "What country is the mountain in?",  # This turn fails with 400
    ]

    for i, prompt in enumerate(prompts, 1):
        print(f"Turn {i}: {prompt}")
        result = await Runner.run(agent, prompt, session=session)
[...]

PR #3941, opened 2026-07-24
fix: strip orphaned assistant ids per reasoning pairing, not list-wide

AmirF194:
`_strip_orphaned_assistant_ids` (`src/agents/memory/openai_responses_compaction_session.py`) is supposed to strip the `id` from an assistant message whose paired reasoning item did not survive `responses.compact`, since the Responses API rejects an id whose reasoning item is missing (the #2727 bug). The check it actually ran was `any(item.type == "reasoning" for item in items)`: if any reasoning item is present anywhere in the compacted output, the function returned every item untouched, including an assistant message whose own reasoning item was dropped elsewhere in the same output.

A reasoning item pairs with the next non-reasoning item that follows it, the same convention `_drop_reasoning_items_preceding_dropped_calls` in `run_internal/items.py` uses. So a compacted output can legitimately mix a message whose reasoning survived with one whose reasoning did not, and the list-wide check treats the whole output as paired the moment one reasoning item survives anywhere. The orphaned message keeps its id, and the next `responses.create` call carries the same invalid-id shape #2727 already fixed for the simpler case.

## Fix

Track pairing per adjacent (reasoning, next-item) pair instead of once for the whole list: an id is only kept when the immediately preceding reasoning item survived.

## Verification

Reproduced end to end through `OpenAIResponsesCompactionSession.run_compaction()`, the real public entry point, not the helper directly: a `responses.compact` output shaped `[reasoning(rs_1), assistant(msg_1, paired), assistant(msg_2, orphaned)]` stored `msg_2` with its id intact on main. Added a regression test at that level plus a narrower unit test on `_strip_orphaned_assistant_ids` directly; both fail on main and pass with the fix (differential run, main vs branch, with `module.__file__` provenance checked on each side).

Ran the full `tests/memory/` suite (150 tests, all pass) plus `ruff format --check`, `ruff check`, and `mypy` scoped to the changed file (all clean). I did not make a live call to the Responses API; the existing #2727 test already established that this shape of stored id is what the API rejects, and this PR only widens the same guard's pairing check, it does not touch the API-rejection premise itself.

### DIFF
--- a/src/agents/memory/openai_responses_compaction_session.py
+++ b/src/agents/memory/openai_responses_compaction_session.py
@@ -391,20 +391,32 @@ def _strip_orphaned_assistant_ids(
     because the API expects the paired reasoning item for each assistant message
     ID.  This function detects and removes those orphaned IDs so the compacted
     history can be used safely.
+
+    A reasoning item pairs with the next non-reasoning item that follows it (the
+    same convention ``_drop_reasoning_items_preceding_dropped_calls`` uses), so
+    pairing is tracked per message rather than once for the whole list: a
+    compacted output can legitimately mix a message whose own reasoning item
+    survived with one whose reasoning item was dropped elsewhere in the same
+    output.
     """
     if not items:
         return items
 
-    has_reasoning = any(
-        isinstance(item, dict) and item.get("type") == "reasoning" for item in items
-    )
-    if has_reasoning:
-        return items
-
     cleaned: list[TResponseInputItem] = []
+    has_paired_reasoning = False
     for item in items:
-        if isinstance(item, dict) and item.get("role") == "assistant" and "id" in item:
+        if isinstance(item, dict) and item.get("type") == "reasoning":
+            has_paired_reasoning = True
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. The current helper is indeed list-wide, and the new tests show how the proposed per-item rule behaves for a constructed mixed output. However, the demonstrated #2727 failure contained no reasoning items and is already covered by #2728. This PR does not show that responses.compact actually returns the mixed [reasoning, paired message, orphaned message] shape, or that replaying such output produces the claimed 400.

Because compacted items are documented as opaque, we should not add a new adjacency-based interpretation without real-path evidence. I am going to close this PR for now. We can revisit it with a captured compact output from a supported model and the corresponding failing continuation request.

AmirF194:
Fair, thanks for the close look. We didn't confirm that compact() actually produces the mixed reasoning/orphaned shape on a real path, only constructed it as a test fixture to demonstrate the gap. Appreciate you catching that before it merged.

PR #3949, opened 2026-07-25
fix: support async callable objects as function tools

seratch:
This pull request fixes function tool creation and invocation for callable objects, including instances with async `__call__` methods.

Callable objects now use their class name as a safe fallback tool name and resolve annotations from `__call__` when generating the function schema. Async callable objects are invoked without being misclassified as synchronous, while synchronous callables that return an awaitable have that result awaited exactly once.

Regression coverage includes bare and configured `function_tool` usage, schema generation, timeout compatibility, and single-invocation behavior.


seratch:
@codex review

PR #3951, opened 2026-07-25
fix: include device nodes when parsing ls output

TheSaiEaranti:
### Summary

`parse_ls_la` silently drops every character and block device, so `SandboxSession.ls()` omits them from its results without any error or warning.

`ls -la` prints `major, minor` in place of the size column for device nodes, which shifts every following field by one:

```
-rw-r--r-- 1 root root      123 Jan  1 00:00 regular.txt
crw-rw-rw- 1 root root     1, 3 Jan  1 00:00 null
```

`int(parts[4])` then parses `"1,"`, raises `ValueError`, and the `except` branch `continue`s past the line — the swallowed exception is what makes the loss silent. Listing `/dev` returns only the non-device entries.

This detects the device form and re-splits with the extra field accounted for, reporting `size = 0` to match what `stat` reports for device nodes. Regular files, directories, symlinks, and the existing SELinux/ACL/xattr marker handling are unaffected.

### Test plan

Added `test_parse_ls_la_includes_device_nodes` to `tests/sandbox/test_parse_utils.py`, covering a character device, a block device, and a regular file in one listing, asserting the paths, kind, size, owner and group.

Before the change the new test fails, showing the silent drop:

```
AssertionError: assert ['/dev/regular.txt'] == ['/dev/regula...', '/dev/sda']
  Right contains 2 more items, first extra item: '/dev/null'
```

After the change:

- `uv run pytest tests/sandbox/test_parse_utils.py` — 12 passed
- `uv run pytest tests/sandbox/` — 912 passed, 1 skipped
- `.agents/skills/code-change-verification/scripts/run.sh` — all commands passed (`make tests` in 94s, `make typecheck` in 265s)

### Issue number

None — found while reading `src/agents/sandbox/util/parse_utils.py`.

### Checks

### DIFF
--- a/src/agents/sandbox/util/parse_utils.py
+++ b/src/agents/sandbox/util/parse_utils.py
@@ -20,15 +20,20 @@ def parse_ls_la(output: str, *, base: str) -> list[FileEntry]:
         permissions_str = parts[0]
         owner = parts[2]
         group = parts[3]
-        if parts[4].endswith(","):
-            # Character and block devices print "major, minor" in place of the
-            # size column, which shifts every following field by one. `stat`
-            # reports size 0 for device nodes.
-            parts = line.split(maxsplit=9)
-            if len(parts) < 10:
-                continue
+        # Character and block devices report a device identifier in place of the
+        # size column, in one of two formats. `stat` reports size 0 for them.
+        if permissions_str[:1] in {"c", "b"}:
             size = 0
-            name = parts[9]
+            if parts[4].endswith(","):
+                # GNU coreutils prints "major, minor", which occupies two
+                # fields and shifts every following field by one.
+                parts = line.split(maxsplit=9)
+                if len(parts) < 10:
+                    continue
+                name = parts[9]
+            else:
+                # BSD ls prints a single hexadecimal identifier, e.g. 0x3000002.
+                name = parts[8]
         else:
             try:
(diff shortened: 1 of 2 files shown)


seratch:
Can you continue working until Codex auto review says 👍?

TheSaiEaranti:
Good catch — thanks. Confirmed on macOS: BSD `ls` prints a single hexadecimal device identifier with no field shift, so those entries were still being skipped.

```
crw-rw-rw-  1 root  wheel     0x3000002 Jul 25 00:10 /dev/null
brw-r-----  1 root  operator  0x1000000 Jul  4 22:33 /dev/disk0
```

Fixed in 239f5e0 by keying off the character/block mode instead of the comma, then handling the two size-column formats separately — GNU's `major, minor` shifts the following fields, BSD's hex identifier does not. Added `test_parse_ls_la_includes_bsd_device_nodes` covering the BSD form; it fails on the previous commit and passes now.

Also checked the result against a real `ls -la /dev` listing from macOS: all 342 entries parse, including `/dev/null`, `/dev/zero` and `/dev/disk0`.

Verification re-run on the new commit: `.agents/skills/code-change-verification/scripts/run.sh` — all commands passed.

TheSaiEaranti:
@codex review

seratch (approved):

PR #3952, opened 2026-07-25
fix: prevent duplicate handlers in enable_verbose_stdout_logging

Quratulain-bilal:
## Summary
\enable_verbose_stdout_logging()\ adds a new \StreamHandler\ every time it is called, causing duplicate log messages when invoked multiple times.

## Changes
- Added early return if a stdout handler already exists on the logger.
- Added test to verify no duplicate handlers are added.

## Test Plan
- [x] New test \	est_enable_verbose_stdout_logging_no_duplicate_handlers\ added
- [x] All existing tests pass


seratch:
Thanks for sharing this. The improvement will be done by https://github.com/openai/openai-agents-python/pull/3957 more properly.

PR #3953, opened 2026-07-25
fix: include colon in empty-content transcript summary lines

AmirF194:
`_format_transcript_item_legacy` in `src/agents/handoffs/history.py` formats each transcript item as a `"role: content"` line inside the `<CONVERSATION HISTORY>` block that `nest_handoff_history` builds when summarizing a prior handoff. When an item's content is empty (a role-bearing turn with nothing but tool calls, since the calls themselves are omitted from the summary and leave `content` blank), the function returns a bare `"role"` line with no colon.

`_parse_summary_line` re-parses that same summary text on a later handoff (a third hop, since the second hop's history already contains the first hop's summary nested inside it) by splitting on the first colon: `role_part, sep, remainder = stripped.partition(":")`, returning `None` when there is no colon. A bare `"assistant"` line fails that split and the turn is silently dropped, no error, no warning. Over more chained handoffs the transcript progressively loses every role-bearing turn whose content happened to be empty at format time.

## Fix

Always emit the colon in `_format_transcript_item_legacy`, even when content is empty (`"assistant: "` instead of `"assistant"`). `_parse_summary_line` already reconstructs an empty remainder correctly (no `content` key gets added to the parsed item), so this only restores the missing role marker; formatting and parsing of non-empty content lines is unchanged.

## Verification

Reproduced directly through `_build_summary_message` and `_flatten_nested_history_messages`, the real round-trip functions: a 3-turn transcript with one empty-content assistant turn recovers only 2 turns on main, 3 on the branch. Added `test_nest_handoff_history_preserves_role_bearing_turn_with_empty_content` to `tests/test_extension_filters.py`, driven through the public `nest_handoff_history` entry point across two nesting hops; it fails on main and passes on the branch.

Ran `.agents/skills/code-change-verification/scripts/run.sh` (format, lint, typecheck, tests). `make format`, `make lint`, and `make typecheck` (mypy + pyright, full repo) are all clean. `make tests` has one failure, `test_runner_preserves_unix_local_lifecycle_state_across_pause_and_resume`, from a missing `rclone` binary in this sandboxed container; I confirmed it reproduces identically on unmodified main, so it is a pre-existing environment gap, not something this change touches. Everything else in the suite passes (5689 passed, 4 skipped in the parallel run; 28 passed, 5 skipped in the serial run).

### DIFF
--- a/src/agents/handoffs/history.py
+++ b/src/agents/handoffs/history.py
@@ -429,7 +429,10 @@ def _format_transcript_item_legacy(item: TResponseInputItem) -> str:
         if isinstance(name, str) and name:
             prefix = f"{prefix} ({name})"
         content_str = _stringify_content(item.get("content"))
-        return f"{prefix}: {content_str}" if content_str else prefix
+        # Always emit the colon, even for empty content: _parse_summary_line requires one to
+        # find the role/content split, so a bare "role" line (no colon) fails to parse and the
+        # turn silently vanishes on the next re-nesting hop.
+        return f"{prefix}: {content_str}"
 
     item_type = item.get("type", "item")
     rest = {k: v for k, v in item.items() if k not in ("type", "provider_data")}
--- a/tests/test_extension_filters.py
+++ b/tests/test_extension_filters.py
@@ -1029,6 +1029,32 @@ def test_nest_handoff_history_parse_summary_line_empty_stripped() -> None:
     assert "Hello" in final_summary["content"] or "Reply" in final_summary["content"]
 
 
+def test_nest_handoff_history_preserves_role_bearing_turn_with_empty_content() -> None:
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for investigating this. The PR does not yet demonstrate a user-visible need for this exact empty-message case. The normal tool-call-only conversion does not create an empty assistant message: tool calls remain separate items and are serialized into the nested summary. The regression test instead injects {"role": "assistant", "content": ""} directly.
The current patch also reparses assistant: as {"role": "assistant"} without the required content field, which an identity history mapper can forward as exact model input. 

I am going to close this PR for now. We can revisit it with a real supported provider or gateway reproduction, including the model, API version, a redacted raw payload, and the concrete consequence of losing the empty role-bearing item.

PR #3954, opened 2026-07-25
fix: remove unnecessary list wrapping in serialize_usage

Quratulain-bilal:
## Summary

\serialize_usage()\ was wrapping \input_tokens_details\ and \output_tokens_details\ in lists (\[input_details]\ and \[output_details]\), but these fields are single objects in the \Usage\ dataclass. This caused inconsistent serialization compared to \model_usage_to_span_usage()\.

## Changes
- Removed list wrapping from \serialize_usage\ output
- Updated tests to match the correct (non-list) format

## Test Plan
- [x] Existing tests updated to expect correct format
- [x] No functional changes to deserialization (defensive list handling preserved)

### DIFF
--- a/src/agents/usage.py
+++ b/src/agents/usage.py
@@ -323,9 +323,9 @@ def serialize_usage(usage: Usage) -> dict[str, Any]:
     return {
         "requests": usage.requests,
         "input_tokens": usage.input_tokens,
-        "input_tokens_details": [input_details],
+        "input_tokens_details": input_details,
         "output_tokens": usage.output_tokens,
-        "output_tokens_details": [output_details],
+        "output_tokens_details": output_details,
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for the contribution. I traced `serialize_usage()` through its callers: it is used for `RunState` persistence, and `deserialize_usage()` already accepts both the historical list-wrapped detail objects and direct objects while preserving their values. `model_usage_to_span_usage()` targets a separate tracing payload, so matching its shape does not by itself demonstrate a `RunState` bug. The list-wrapped shape also shipped in v0.18.3.

I am going to close this PR for now. If you can provide a supported scenario where the current `RunState` serialization or restoration produces an incorrect user-visible result, we can revisit the narrowest change from that reproduction.

PR #3958, opened 2026-07-26
fix: detect async callable objects in function_tool

Quratulain-bilal:
inspect.iscoroutinefunction returns False for callable instances with an async __call__ method. This causes function_tool to incorrectly treat async callable objects as sync.

This fix adds detection of async __call__ on callable objects, consistent with the fix in PR #3944 for handoffs and turn_resolution.

Tests added: test_function_tool_with_async_callable_object and test_function_tool_with_sync_callable_object

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -2330,7 +2330,6 @@ def function_tool(
     """
 
     def _create_function_tool(the_func: ToolFunction[...]) -> FunctionTool:
-        is_sync_function_tool = not inspect.iscoroutinefunction(the_func)
         schema = function_schema(
             func=the_func,
             name_override=name_override,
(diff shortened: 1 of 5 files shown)


seratch:
Thanks for investigating this and for adding coverage for both async and sync callable objects. The underlying async callable issue is valid, but it has already been addressed by #3949, which is now merged and normalizes callable instances through the existing function-tool pipeline.

We are following up in #3959 with the narrower compatibility and construction-time validation work needed around callable normalization. This keeps schema generation, naming, and invocation on one consistent path instead of adding a separate dispatch-only fix here.

I am going to close this PR as superseded by #3949 and #3959. Thanks again for the contribution.

PR #3962, opened 2026-07-26
fix: use provider_output in ToolCallOutputItem for function rejection consistency

AAliKKhan:
﻿This pull request fixes a data inconsistency in unction_rejection_item where ToolCallOutputItem.output was set to the plain-text 
ejection_message while 
aw_item contained the JSON-encoded provider_output. When output_json_schema is set for programmatic tool calls, unction_tool_error_output returns a JSON-encoded string ({"error": "..."}) instead of the plain message. Using provider_output for output ensures consistency between output and 
aw_item.

Fixes a latent bug where downstream consumers reading ToolCallOutputItem.output would receive plain text while the model receives JSON-encoded output, causing silent data inconsistency.

### DIFF
--- a/src/agents/run_internal/items.py
+++ b/src/agents/run_internal/items.py
@@ -773,7 +773,7 @@ def function_rejection_item(
         output_json_schema=output_json_schema,
     )
     return ToolCallOutputItem(
-        output=rejection_message,
+        output=provider_output,
         raw_item=ItemHelpers.tool_call_output_item(
             tool_call,
             provider_output,


seratch:
Thanks for the contribution. This separation is intentional: ToolCallOutputItem.output is the SDK-level logical output, while raw_item is the provider-facing and replay representation. PR #3932 encoded structured programmatic errors in raw_item specifically while preserving the public output value.

The current path already sends {"error": "..."} to the provider through raw_item and to_input_item(). Changing .output would instead expose transport JSON through RunResult.new_items and streaming events, making approval rejections inconsistent with FunctionToolResult.output and other SDK-generated error paths.
I am going to close this PR. If there is a supported scenario that cannot obtain the provider representation through raw_item or to_input_item(), please share a minimal reproduction and we can reconsider that narrower gap.

PR #3963, opened 2026-07-26
fix: prevent BaseException cleanup from masking original exception in streaming

AAliKKhan:
﻿This pull request fixes two issues in src/agents/run_internal/run_loop.py:

1. **Prevents exception masking in the streaming cleanup handler**: In the except BaseException block of start_streaming, if streamed_result._event_queue.put_nowait(QueueCompleteSentinel()) fails (e.g., queue full, event loop closed), the original exception is silently lost because 
aise is never reached. Wrapping the put_nowait in a try/except ensures the original exception is always re-raised.

2. **Renames input variable to 	urn_input**: The local variable input shadowed Python's built-in input() function, which is a latent bug waiting to cause a TypeError if any code within the function scope calls the builtin.

### DIFF
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -706,7 +706,10 @@ async def start_streaming(
             streamed_result.trace.finish(reset_current=True)
         if not streamed_result.is_complete:
             streamed_result.is_complete = True
-            streamed_result._event_queue.put_nowait(QueueCompleteSentinel())
+            try:
+                streamed_result._event_queue.put_nowait(QueueCompleteSentinel())
+            except Exception:
+                pass
         raise
 
     try:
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for investigating this. I traced the exact streaming path, and Runner.run_streamed() creates RunResultStreaming with the default unbounded asyncio.Queue; the SDK also never shuts that queue down. The cited queue-full or closed-loop failure is therefore not reachable in supported use.

Additionally, if enqueue did fail while stream_events() was already awaiting get(), swallowing the exception would not wake the consumer, so this change would not guarantee that the original error reaches the caller. The input rename is unrelated cleanup and does not address an observed defect.

I am going to close this PR. If you can provide a minimal public-API reproduction using the SDK-managed queue and the observed traceback, we can revisit the completion-signaling behavior with that narrower evidence.

PR #3964, opened 2026-07-26
fix: use None-checking instead of or-chain for timeout fallback values

AAliKKhan:
﻿This pull request fixes a bug in coerce_shell_call where the or chaining for 	imeout_value silently drops an explicit 	imeout_ms=0 (a valid value meaning no timeout). In Python,  or ... evaluates to the next operand, so the chain would skip 	imeout_ms=0 and fall through to 	imeoutMs and 	imeout, ultimately setting 	imeout_ms to None.

The adjacent code for max_length_value already uses the correct pattern (explicit is None checks). This fix aligns the timeout fallback with that same pattern.

### DIFF
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -706,7 +706,10 @@ async def start_streaming(
             streamed_result.trace.finish(reset_current=True)
         if not streamed_result.is_complete:
             streamed_result.is_complete = True
-            streamed_result._event_queue.put_nowait(QueueCompleteSentinel())
+            try:
+                streamed_result._event_queue.put_nowait(QueueCompleteSentinel())
+            except Exception:
+                pass
         raise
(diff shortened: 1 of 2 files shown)


seratch:
I traced timeout_ms=0 through coerce_shell_call and ShellActionRequest. On current main and v0.18.3, zero is normalized to None, which already represents the stated no-timeout outcome. The platform type documents this only as an optional duration and does not define zero as a sentinel whose literal value must be preserved. The PR provides no real executor reproduction showing that the current behavior is incorrect, while passing literal 0 to custom executors can instead cause an immediate timeout.

This branch also contains the unrelated commit from #3963. I am going to close this PR. If a supported executor has an observable failure under the current normalization, please provide a minimal reproduction and the contract establishing zero semantics, and we can revisit it as a focused PR based on main.

PR #3967, opened 2026-07-26
fix: use last_agent property in RunResultStreaming._create_error_details

AAliKKhan:
﻿This pull request fixes RunResultStreaming._create_error_details() which reads self.current_agent directly instead of self.last_agent. When _release_last_agent_reference() has been called, current_agent is set to None via __dict__, causing RunErrorDetails to be constructed with last_agent=None — a type violation that can cause AttributeError downstream when accessing .name on the error details.

The last_agent property handles the None case by falling back to a stored weakref, matching the type contract of RunErrorDetails.last_agent: Agent[Any].

### DIFF
--- a/src/agents/result.py
+++ b/src/agents/result.py
@@ -866,7 +866,7 @@ class RunResultStreaming(RunResultBase):
             input=self.input,
             new_items=self.new_items,
             raw_responses=self.raw_responses,
-            last_agent=self.current_agent,
+            last_agent=self.last_agent,
             context_wrapper=self.context_wrapper,
             input_guardrail_results=self.input_guardrail_results,
             output_guardrail_results=self.output_guardrail_results,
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying inconsistency is real: after release_agents(), current_agent is None, and using last_agent correctly recovers an agent whose weakref is still alive.

However, a focused base/head probe found that when the weakref has already been collected, this change makes _check_errors() raise AgentsException("Last agent reference is no longer available."), masking the original MaxTurnsExceeded.

Please preserve the original terminal exception when the agent can no longer be resolved, leaving run_data unset instead of constructing invalid details or raising a replacement error. Please also add regression tests for retained and collected weakrefs. Once those cases are covered, this should be ready for another review.

AAliKKhan:
I've updated this branch based on feedback:
- \_create_error_details\ now wraps \last_agent\ in try/except \AgentsException\ and returns \None\ when the weakref is collected (preserving the original terminal exception)
- Added regression tests for both retained and collected weakref scenarios

Would appreciate a re-review when you have a moment.

seratch (approved):

PR #3968, opened 2026-07-26
fix: use Mapping instead of dict in get_mapping_or_attr for broader compatibility

AAliKKhan:
﻿This pull request fixes a subtle type-handling inconsistency in _tool_identity.py's get_mapping_or_attr. The function used isinstance(value, dict) instead of isinstance(value, Mapping), which means any non-plain-dict mapping (e.g., OrderedDict, defaultdict, ChainMap, or a custom Mapping subclass) would fall through to getattr() and return None.

The same function in 	ool_execution.py (line 615) correctly uses isinstance(target, Mapping). This fix aligns the two implementations.

### DIFF
--- a/src/agents/_tool_identity.py
+++ b/src/agents/_tool_identity.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 from typing import Any, Literal, cast
 
 from typing_extensions import Required, TypedDict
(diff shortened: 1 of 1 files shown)


seratch:
The behavior difference for ChainMap and other custom Mapping implementations is real, but I could not find a supported runtime path that produces those values here. ModelResponse.output is typed as OpenAI ResponseOutputItem, the built-in providers produce ResponseFunctionToolCall objects, and RunState JSON restoration produces plain dictionaries.

A custom model adapter can normalize an implementation-specific mapping with dict(...) or construct ResponseFunctionToolCall before returning ModelResponse. Expanding this internal helper based only on symmetry would create a broader input contract without a demonstrated user outcome, linked issue, or real provider scenario.

I am going to close this PR for now. We can reconsider if there is a concrete supported custom-provider case that must preserve a non-dict Mapping and cannot normalize it at the adapter boundary.

PR #3970, opened 2026-07-26
fix: prevent KeyboardInterrupt/SystemExit from being swallowed in MCP retry cleanup

AAliKKhan:
﻿This pull request fixes a bug in MCPServerStreamableHttp._call_tool_with_isolated_retry() where except BaseException: pass silently swallows KeyboardInterrupt and SystemExit. If a user presses Ctrl+C while a cancelled MCP request task finishes its cleanup, the KeyboardInterrupt is discarded, preventing the process from shutting down.

Changing to except Exception: ensures that fatal signals like KeyboardInterrupt and SystemExit still propagate, while non-fatal exceptions from the cancelled child task continue to be suppressed as intended.

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -1641,7 +1641,7 @@ class MCPServerStreamableHttp(_MCPServerWithClientSession):
                 request_task.cancel()
             try:
                 await request_task
-            except BaseException:
+            except Exception:
                 pass
             raise
 


seratch:
Thanks for investigating this. I do not think the stated Ctrl+C failure is demonstrated here. Under asyncio.Runner, SIGINT cancels the main task with CancelledError and raises KeyboardInterrupt after cleanup unwinds, so the user's KeyboardInterrupt does not normally pass through this child-task drain block.

This change also introduces a regression in the common cancellation path: request_task.cancel() makes await request_task raise CancelledError, which is not an Exception, so control never reaches the final bare raise. That replaces the caller's cancellation, including its message, with the child's cancellation.

Since the PR has no reproduction or regression test and changes released cancellation semantics, I am going to close it. We can revisit this with a runnable example on a supported Python version where the base revision actually fails to exit.

PR #3971, opened 2026-07-26
fix: use mode=json in _serialize_raw_item_value to prevent non-serializable types

AAliKKhan:
﻿This pull request fixes _serialize_raw_item_value in 
un_state.py which calls model_dump() without mode="json". This can return non-JSON-serializable Python types (e.g., datetime, Decimal, Path) when the Pydantic model contains such fields, causing TypeError downstream when json.dumps() is called in 	o_string().

Adding mode="json" ensures the model's fields are converted to JSON-safe types before serialization, matching the pattern used elsewhere in the codebase (e.g., 
ealtime/session.py line 114).

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -1292,7 +1292,7 @@ def _transform_field_names(
 def _serialize_raw_item_value(raw_item: Any) -> Any:
     """Return a serializable representation of a raw item."""
     if hasattr(raw_item, "model_dump"):
-        return raw_item.model_dump(exclude_unset=True)
+        return raw_item.model_dump(mode="json", exclude_unset=True)
     if isinstance(raw_item, dict):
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the contribution. The generic failure mechanism is valid: Pydantic's default model_dump() mode can retain Python objects that json.dumps() cannot encode. However, this PR does not show that the problem occurs for a supported RunState raw item. The current Response input/output item contract is JSON-shaped, and there is no linked issue, real provider reproduction, or regression test demonstrating datetime, Decimal, or Path on that path.
I am going to close this PR for now. We can revisit it with a minimal supported-path reproduction, including the provider/model/API version and a redacted raw payload, or a concrete custom Model implementation that returns a valid TResponseOutputItem and fails on the base branch.

PR #3972, opened 2026-07-26
fix: add default=str fallback to RunState.to_string for non-serializable types

AAliKKhan:
﻿This pull request adds default=str to the json.dumps() call in RunState.to_string(). Without this fallback, if any value in the serialized state contains a non-JSON-serializable type (e.g., datetime, Decimal, UUID), 	o_string() would raise TypeError instead of producing a usable string representation.

The _ensure_json_compatible helper at line 1314 already uses default=str as a safety net. This fix brings 	o_string() in line with the same defensive pattern used elsewhere in the codebase.

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -1066,6 +1066,7 @@ class RunState(Generic[TContext, TAgent]):
                 include_tracing_api_key=include_tracing_api_key,
             ),
             indent=2,
+            default=str,
         )
 
     def set_trace(self, trace: Trace | None) -> None:


seratch:
Thanks for exploring the alternative. Applying default=str to the final json.dumps() call is too broad for the durable RunState boundary. It silently converts values in mapping contexts and other state fields while their metadata still says that no deserializer is required, so resume can restore different application types without warning. It also makes to_string() semantics diverge from to_json().

I recommend closing this PR. If a supported non-JSON payload path is demonstrated, the conversion should be scoped to the field or typed Pydantic boundary that owns that value, with round-trip and resume regression coverage.

PR #3973, opened 2026-07-26
fix: propagate tool execution exceptions instead of silently swallowing them

AAliKKhan:
﻿This pull request fixes a bug in _FunctionToolBatchExecutor._await_invoke_task where tool execution exceptions (regular Exception subclasses like ValueError, TypeError, RuntimeError) are silently swallowed when a tool invoke task completes after cancellation.

The condition at line 1978 used 
ot isinstance(invoke_failure, Exception), which only re-raises KeyboardInterrupt/SystemExit/GeneratorExit but silently drops all regular exceptions. The tool failure is replaced by a CancelledError, masking the root cause.

Changing to 
ot isinstance(invoke_failure, asyncio.CancelledError) ensures that:
- Regular tool exceptions (ValueError, TypeError, etc.) are properly propagated
- CancelledError is still not re-raised (falls through to the outer 
aise)
- Fatal signals (KeyboardInterrupt, SystemExit) continue to propagate

### DIFF
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -1976,7 +1976,7 @@ class _FunctionToolBatchExecutor:
                 if not invoke_task.cancelled():
                     invoke_failure = invoke_task.exception()
                     if isinstance(invoke_failure, BaseException) and not isinstance(
-                        invoke_failure, Exception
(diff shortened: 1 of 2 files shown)


seratch:
Please avoid mixing several changes (meaning several unrelated commits) within a single PR.

PR #3975, opened 2026-07-26
fix: propagate tool execution exceptions instead of silently swallowing them

AAliKKhan:
Fix condition in \_await_invoke_task\ to propagate regular \Exception\ subclasses during cancellation instead of silently swallowing them.

### DIFF
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -1976,7 +1976,7 @@ class _FunctionToolBatchExecutor:
                 if not invoke_task.cancelled():
                     invoke_failure = invoke_task.exception()
                     if isinstance(invoke_failure, BaseException) and not isinstance(
-                        invoke_failure, Exception
+                        invoke_failure, asyncio.CancelledError
                     ):
                         raise invoke_failure from cancel_exc
             else:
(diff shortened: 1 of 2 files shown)


AAliKKhan:
This is a resubmission of #3973 with the unrelated \
un_state.py\ change removed. Only the focused \	ool_execution.py\ fix is included, plus a regression test for the condition change.

seratch:
@AAliKKhan By the way, could you please use our pull request template for the description? It should help you organize things.

seratch:
Thanks for narrowing the PR. I traced this branch through `_await_invoke_task`, and the current behavior is intentional rather than a swallowed-exception bug.

Ordinary tool exceptions already propagate when the invocation runs without external cancellation, and the sibling-failure branch immediately above also preserves non-cancellation failures. The branch changed here is specifically the parent-cancellation race: if the parent is cancelled while the shielded invocation completes, the invocation exception is consumed and the original `CancelledError` wins. This behavior is part of v0.18.3 and is covered by `test_parent_cancellation_wins_when_shield_raises_after_tool_finishes`.

The new test only compares two `isinstance` predicates and does not exercise that runtime contract. I am going to close this PR because the change would make external cancellation race-dependent. If you can reproduce an ordinary tool exception being lost without parent cancellation, we can revisit that narrower case.

PR #3976, opened 2026-07-27
fix: guard extract_last_content refusal against None via model_construct

AAliKKhan:
ItemHelpers.extract_last_content() declares -> str but returns last_content.refusal directly, risking None surfacing via model_construct (same LiteLLM bypass path already guarded for ResponseOutputText).


seratch:
@AAliKKhan As you can see in the current comment, this is intentionally kept as-is.

PR #3977, opened 2026-07-27
fix: preserve timeout_ms=0 in coerce_shell_call instead of or-chain dropping it

AAliKKhan:
The or-chain at tool_execution.py:664-668 silently converts timeout_ms=0 (immediate timeout) to None (no timeout) because Python or treats 0 as falsy. Adjacent max_output_length code uses the correct is None pattern at lines 671-673, confirming the or-chain is an oversight.


seratch:
I don't see the necessity to preserve the zero value here.

PR #3981, opened 2026-07-27
fix(chatcmpl): keep thinking-block signatures bound to their own block

gunjanjaswal:
When Anthropic thinking blocks are converted to reasoning items and back, the reasoning text and the signatures were built into two lists filtered by different conditions: a content item was added only for blocks that had thinking text, while a signature was added only for blocks that carried one. The reverse path then re-paired them positionally with `signatures.pop(0)`. Once the blocks are not uniformly signed, the lists are not index-aligned, so a block with text but no signature pulls a later block's signature onto itself and the real one is lost. Anthropic then rejects the mismatched blocks.

This shows up on the heterogeneous-signature path (thinking blocks arriving from LiteLLM or other adapters where some blocks lack a signature), not vanilla single-provider Anthropic where every block is signed.

The forward path now emits one signature slot per `reasoning_text` content item, using an empty placeholder when a text block has no signature, so `encrypted_content` stays aligned with the content items. The reverse path treats an empty slot as "no signature" and guards the pop. Fully-signed round-trips are unchanged.

Added round-trip tests for unsigned-then-signed, signed-then-unsigned, and a signature-only block interleaved. The full suite passes; the new tests fail against `main` without the source change.

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -149,9 +149,14 @@ class Converter:
 
             # Store thinking blocks for Anthropic compatibility
             if hasattr(message, "thinking_blocks") and message.thinking_blocks:
-                # Store thinking text in content and signature in encrypted_content
+                # Store thinking text in content and signature in encrypted_content.
+                # Emit exactly one signature slot per reasoning_text content item so the
+                # two lists stay index-aligned. Blocks with text but no signature get an
+                # empty-string placeholder; that keeps a later block's signature bound to
+                # its own block instead of shifting onto an earlier unsigned one.
                 reasoning_item.content = []
                 signatures: list[str] = []
+                has_signature = False
                 for block in message.thinking_blocks:
                     if isinstance(block, dict):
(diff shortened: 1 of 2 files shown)


seratch:
I could not reproduce that final-message shape through LiteLLM 1.83.0: its stream aggregation combines unsigned thinking deltas and the terminal signature delta into a single signed thinking block, which the base SDK already round-trips correctly.

Please provide a real supported provider or gateway reproduction showing the final message.thinking_blocks payload, including the model, API mode, LiteLLM version, and redacted raw block structure. Until then, the behavior is synthetically reachable, but the practical need remains unproven.

gunjanjaswal:
Fair enough, thanks for digging into the LiteLLM aggregation path. You are right that the current adapters collapse the unsigned deltas and the terminal signature into one signed block, so the mismatched shape is only synthetically reachable right now. I would rather not add code for a case no supported provider actually produces, so I am closing this. If a real gateway ever surfaces heterogeneously-signed thinking blocks, the positional pairing in the converter is the spot to revisit.

PR #3982, opened 2026-07-27
fix(run): cancel the parallel input-guardrail task when the model turn fails

gunjanjaswal:
In the parallel input-guardrail path (`AgentRunner.run`), the guardrail coroutine and the model turn are awaited together with `asyncio.gather` without `return_exceptions=True`, and the only handler catches `InputGuardrailTripwireTriggered`.

If the model turn raises any other exception (a `ModelBehaviorError`, a provider `RuntimeError`) before the guardrail finishes, `gather` propagates the model error but leaves the sibling guardrail task running. It keeps making network calls and side effects after the run has already failed and returned, and an error raised inside that guardrail after `gather` resolves is silently swallowed. Guardrails default to `run_in_parallel=True`, so this is the common path whenever an input guardrail exists.

The fix wraps the guardrail coroutine in an explicit task and adds a symmetric cleanup branch: on any non-tripwire failure, cancel whichever of the guardrail and model tasks is still pending and drain both with `return_exceptions=True` before re-raising. The existing tripwire behavior and the flag-gated model-task cancellation are preserved, and the streamed path is unchanged.

This complements #3239, which cancelled sibling guardrails inside `run_input_guardrails` itself. That fix does not reach the orchestration-level `gather` in `run.py`, so when the model task is the one that raises, the guardrail task is still orphaned. This closes that gap.

Tests cover a model error cancelling the in-flight guardrail task, and a guardrail's non-tripwire error propagating and cancelling the model task; the existing tripwire tests still pass. Both new tests fail on `main` and pass with this change. Full suite: 5783 passed, 7 skipped; ruff, format, and mypy clean.

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1266,14 +1266,17 @@ class AgentRunner:
                             )
 
                             if parallel_guardrails:
+                                guardrail_task = asyncio.create_task(
+                                    run_input_guardrails(
+                                        starting_agent,
+                                        parallel_guardrails,
+                                        copy_input_items(original_input),
+                                        context_wrapper,
+                                    )
+                                )
                                 try:
                                     parallel_results, turn_result = await asyncio.gather(
-                                        run_input_guardrails(
-                                            starting_agent,
-                                            parallel_guardrails,
-                                            copy_input_items(original_input),
-                                            context_wrapper,
-                                        ),
+                                        guardrail_task,
                                         model_task,
                                     )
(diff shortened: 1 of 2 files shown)


seratch (approved):

gunjanjaswal:
Thanks for the review and merge, @seratch. Good to have the parallel path cancel and drain the guardrail task instead of leaving it running after the turn fails.

PR #3984, opened 2026-07-27
fix(memory): prevent close() race and enforce closed state in AsyncSQLiteSession

### LINKED ISSUE #3983: fix(memory): prevent close() race and enforce closed state in AsyncSQLiteSession
### Please read this first

- **Have you searched for related issues?** Yes

### Describe the bug
In `AsyncSQLiteSession`, calling `close()` concurrently or repeatedly causes an `AttributeError: 'NoneType' object has no attribute 'close'`.
This occurs because `close()` checks `if self._connection is None:` outside `self._lock` without re-verifying inside `async with self._lock:`. When two concurrent tasks invoke `close()`, both pass the initial `None` check; the first task acquires `self._lock`, closes `self._connection`, and sets it to `None`. When the second task subsequently acquires `self._lock`, it attempts `await self._connection.close()` on `None`.

In addition, `AsyncSQLiteSession` lacks a `self._closed` state flag (unlike `SQLiteSession`). Consequently, calling `get_items()`, `add_items()`, `pop_item()`, or `clear_session()` after `close()` re-initializes a new database connection instead of raising `RuntimeError("AsyncSQLiteSession is closed")`.

### Debug information
- Agents SDK version: main (0.19.0+)
- Python version: Python 3.13.5
- Operating System: macOS

### Repro steps
```python
import asyncio
from agents.extensions.memory import AsyncSQLiteSession

async def main():
    session = AsyncSQLiteSession("test_close_race")
    await session.add_items([{"role": "user", "content": "hi"}])
    # Concurrent close calls raise AttributeError
    await asyncio.gather(session.close(), session.close())

asyncio.run(main())
```

### Expected behavior
`AsyncSQLiteSession.close()` should be thread-safe, task-safe, and idempotent (subsequent and concurrent calls to `close()` should be safe no-ops).
Operations invoked on a closed `AsyncSQLiteSession` should raise `RuntimeError("AsyncSQLiteSession is closed")`, matching `SQLiteSession` behavior.

I intend to submit a pull request fixing this issue.

hsusul:
### Summary

Fixes a concurrency race in `AsyncSQLiteSession.close()` and enforces closed-state checks across session operations.

- **Problem 1 (Race on close)**: `AsyncSQLiteSession.close()` evaluated `if self._connection is None:` outside `self._lock`. When concurrent tasks (e.g. during application shutdown or teardown) called `close()`, both passed the `None` check. The first task acquired `self._lock`, closed the connection, and set `self._connection = None`. The second task then acquired `self._lock` and executed `await self._connection.close()`, raising `AttributeError: 'NoneType' object has no attribute 'close'`.
- **Problem 2 (Closed state enforcement)**: `AsyncSQLiteSession` lacked a `self._closed` state flag (unlike `SQLiteSession`). Consequently, invoking `get_items()`, `add_items()`, `pop_item()`, or `clear_session()` after `close()` re-initialized a database connection instead of raising `RuntimeError("AsyncSQLiteSession is closed")`.

**Solution**:
1. Added `self._closed` flag to `AsyncSQLiteSession`.
2. Updated `close()` to acquire `self._lock`, check `self._closed`, mark `self._closed = True`, and safely close `self._connection` if active.
3. Updated `_get_connection()` to check `self._closed` both outside and inside `self._init_lock`, raising `RuntimeError("AsyncSQLiteSession is closed")`.

### Test plan

- Added `test_async_sqlite_session_concurrent_close` in `tests/extensions/memory/test_async_sqlite_session.py` to verify that concurrent (`asyncio.gather`) and repeated calls to `close()` succeed cleanly.
- Added `test_async_sqlite_session_closed_operations_raise_runtime_error` to verify that calling `get_items()`, `add_items()`, `pop_item()`, or `clear_session()` on a closed session raises `RuntimeError("AsyncSQLiteSession is closed")`.
- Verified formatting, linting, typechecking, and tests:
  - `make format` (passed, 841 files unchanged)
  - `make lint` (passed, 0 errors)
  - `uv run pyright src/agents/extensions/memory/async_sqlite_session.py tests/extensions/memory/test_async_sqlite_session.py` (passed, 0 errors)
  - `uv run pytest tests/extensions/memory/test_async_sqlite_session.py -v` (20 passed)
  - `git diff --check` (passed, 0 warnings)

### Issue number

Closes #3983

### Checks

### DIFF
--- a/src/agents/extensions/memory/async_sqlite_session.py
+++ b/src/agents/extensions/memory/async_sqlite_session.py
@@ -59,6 +59,7 @@ class AsyncSQLiteSession(SessionABC):
         self._connection: aiosqlite.Connection | None = None
         self._lock = asyncio.Lock()
         self._init_lock = asyncio.Lock()
+        self._closed = False
 
     async def _init_db_for_connection(self, conn: aiosqlite.Connection) -> None:
         """Initialize the database schema for a specific connection."""
@@ -96,10 +97,15 @@ class AsyncSQLiteSession(SessionABC):
 
     async def _get_connection(self) -> aiosqlite.Connection:
         """Get or create a database connection."""
+        if self._closed:
+            raise RuntimeError("AsyncSQLiteSession is closed")
+
         if self._connection is not None:
             return self._connection
 
         async with self._init_lock:
+            if self._closed:
+                raise RuntimeError("AsyncSQLiteSession is closed")
             if self._connection is None:
                 self._connection = await aiosqlite.connect(str(self.db_path))
                 await self._connection.execute("PRAGMA journal_mode=WAL")
@@ -264,8 +270,10 @@ class AsyncSQLiteSession(SessionABC):
 
     async def close(self) -> None:
         """Close the database connection."""
-        if self._connection is None:
-            return
         async with self._lock:
-            await self._connection.close()
-            self._connection = None
+            if self._closed:
+                return
+            self._closed = True
+            if self._connection is not None:
+                await self._connection.close()
+                self._connection = None
--- a/tests/extensions/memory/test_async_sqlite_session.py
+++ b/tests/extensions/memory/test_async_sqlite_session.py
@@ -13,6 +13,7 @@ from typing import Any, cast
 import pytest
 
 pytest.importorskip("aiosqlite")  # Skip tests if aiosqlite is not installed
+import aiosqlite
 
 from agents import Agent, Runner, TResponseInputItem
 from agents.extensions.memory import AsyncSQLiteSession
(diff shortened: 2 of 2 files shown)


seratch (approved):

seratch (changes requested):
Thanks for the contribution. The _closed state is appropriate here: checking _connection only inside the lock fixes concurrent close() calls, but it still allows an operation waiting behind close() to create a new connection immediately after shutdown completes.

Before merging, please add a controlled interleaving regression test where close() acquires the session lock first, an operation starts while close is paused, and close then completes. The test should verify that the waiting operation raises RuntimeError and that no connection is recreated. Please keep the _closed state and the existing post-close checks. This will lock down the terminal lifecycle behavior that the patch introduces.

hsusul:
Added test_async_sqlite_session_close_rejects_operation_waiting_on_lock to lock down the terminal lifecycle. It pauses close() while it holds the session lock (during connection close), starts get_items() so it waits on that lock, then lets close finish. The waiter raises RuntimeError("AsyncSQLiteSession is closed"), _connection stays None, and no new aiosqlite.connect occurs. Kept the _closed flag and existing post-close checks unchanged.

seratch (changes requested):
Thanks for adding the controlled close-versus-operation interleaving test. The scenario and final assertions now cover the requested lifecycle behavior.

One test-lifecycle issue remains: assertions run while close() is intentionally paused, before release_close is set. If any of those assertions fail, the paused close task and the aiosqlite worker thread can remain alive, causing the regression test to hang instead of failing cleanly. Please guarantee release and task cleanup with try/finally, use a test-owned event instead of inspecting asyncio.Lock._waiters, and keep the synchronization waits bounded. The runtime implementation otherwise looks ready.

hsusul:
Updated the interleaving test lifecycle: all mid-pause assertions are inside try/finally so release_close is always set and both tasks are cleaned up; waiting is signaled through a test-owned event on lock.acquire instead of asyncio.Lock._waiters; and the synchronization waits are bounded with asyncio.wait_for.

tonydzi (commented):
Read against `main` at today's HEAD. I went looking for two further races and found both already closed — reporting that too, since "I checked and it holds" is worth as much to you as a finding:

- `_get_connection()` is only reached through `_locked_connection()` (L111–114), which already holds `self._lock`, and `close()` now takes the same lock. So connect-vs-close can't interleave, and a connection can't be established *after* `_closed` is set. The `_init_lock` re-check you added is belt-and-braces rather than load-bearing, which is fine, but it reads as if it were the guard — a one-line comment saying `_lock` is what actually orders this would save the next reader the trip.
- Likewise an in-flight `get_items`/`add_items` can't have the connection closed underneath it, for the same reason.

The concurrency test does reproduce the original failure, which is the part that usually doesn't: with the old code all three tasks pass the pre-lock `is None` check while the first is suspended inside `await self._connection.close()`, so the second wakes to `None.close()`. Good test.

**One real defect, in the new code:**

```python
async with self._lock:
    if self._closed:
        return
    self._closed = True
    if self._connection is not None:
        await self._connection.close()
        self._connection = None
```

`_closed` is set before the close is attempted, and `self._connection = None` only runs if `close()` returns normally. If `aiosqlite.Connection.close()` raises — a wedged worker thread, an already-broken connection — the session is left marked closed with a live connection still referenced, and every subsequent `close()` returns at the `if self._closed` line without retrying. The connection and its thread are then unreachable and unclosable for the process lifetime. Before this change, a caller could retry `close()` and get another attempt.

Keeping `_closed = True` where it is looks right — no new work should start once close has been entered. The fix is just to make the teardown unconditional:

```python
    self._closed = True
    conn, self._connection = self._connection, None
    if conn is not None:
        await conn.close()
```

Now the reference is dropped whether or not the close succeeds, and the exception still propagates to the caller.

**One behaviour change worth confirming is intended:** dropping the pre-lock `if self._connection is None: return` means `close()` now always waits for `self._lock`. Closing a session with a long `add_items` in flight blocks until it finishes, where before an unused session closed instantly. Probably what you want during shutdown — but it's a change in what `close()` costs, and it isn't in the description.

**What I did not do:** run the suite, or check whether `aiosqlite`'s `close()` can realistically raise on a healthy connection — the failure path above is reachable by construction, I haven't produced it.

*Disclosure: I'm an AI agent (Claude) reviewing as Palo Alto AI Research Lab.*

seratch:
Thanks for the contribution and for working through the close-versus-operation interleaving behavior. This was a real lifecycle bug, and your report and tests helped establish the intended terminal-close semantics.

Since this PR was opened, #4109 implemented the same behavior through the shared _locked_connection() path, including close-before-first-use and the add_items([]) fast path. That fix has been merged and released in v0.19.3, so there is no remaining code change to carry from this branch.

I am closing this PR as superseded by #4109. Thank you again for identifying the issue and contributing the original fix and regression coverage.

PR #3985, opened 2026-07-27
fix(tracing): enforce max_batch_size during force_flush and shutdown in BatchTraceProcessor

hsusul:
### Summary

When `force_flush()` or `shutdown()` is called on `BatchTraceProcessor`, `_export_batches(force=True)` was called. In `_export_batches`, the inner loop condition contained `force or len(items_to_export) < self._max_batch_size`. Because `force` evaluated to `True`, the `max_batch_size` limit was completely bypassed during queue draining.

As a result, all queued items (even thousands of spans) were gathered into a single exported batch. When sent to the HTTP exporter (`BackendSpanExporter`), giant single payloads lead to HTTP 413 (Payload Too Large) errors, request timeouts, or connection failures, stranding or dropping queued telemetry.

This PR removes the `force or` condition from the inner batch collector so that `_export_batches` always enforces `max_batch_size` on every batch passed to `exporter.export()`, while the outer loop continues draining the queue until empty.

### Test plan

1. Updated existing `test_batch_trace_processor_force_flush` unit test in `tests/test_trace_processor.py` to assert that exported batches strictly respect `max_batch_size` (e.g. 3 queued items with `max_batch_size=2` are exported as 2 separate batches of sizes `[2, 1]`).
2. Ran complete repository verification:
   - `make format-check`
   - `make lint`
   - `make typecheck`
   - `make tests`

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -668,8 +668,8 @@ class BatchTraceProcessor(TracingProcessor):
         self._export_batches(force=True, deadline=self._shutdown_deadline)
 
     def _export_batches(self, force: bool = False, deadline: float | None = None):
-        """Drains the queue and exports in batches. If force=True, export everything.
-        Otherwise, export up to `max_batch_size` repeatedly until the queue is completely empty.
+        """Drains the queue and exports in batches of up to `max_batch_size` until the queue
+        is completely empty.
         """
         with self._export_lock:
             while True:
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The issue is valid: max_batch_size is documented as the maximum number of items passed to one export, but the current force-flush and shutdown paths bypass it, and the default HTTP exporter can consequently place the entire same-key queue into one request.

Before merging, please remove the now-unused force parameter from _export_batches() and update its callers. Once every invocation drains the queue in bounded batches, force=True and force=False no longer select different behavior, so retaining them would leave a misleading internal contract. The new exact batch-size assertion should remain. After that change and green CI, this should be ready for another review.

hsusul:
Removed the unused force parameter from _export_batches() and updated all callers (force_flush, _run, and the no-worker shutdown path).

seratch (approved):

PR #3999, opened 2026-07-28
feat(usage): add arithmetic operators (+, +=, -) and copy method

amanji0:


seratch:
Thanks for sharing this idea. The usage objects are not intended to manipulate on the developer (sdk user) side, so we don't plan to add these methods.

PR #4000, opened 2026-07-28
fix(mcp): strip URL credentials from server names in raised errors

dimaosipa:
### Summary

- `MCPServerSse` / `MCPServerStreamableHttp` default their name to the connection url (`sse: {url}`), which commonly embeds credentials (`user:password@host` or `?api_key=...`). every raised connection and tool-call error interpolated that raw name, so any transient failure surfaced the credential to the caller.
- unlike the log paths, raised errors are not gated by `DONT_LOG_TOOL_DATA`, so there was no way to suppress this even on the default redact-by-default setting.
- the redaction helper already existed and is unit-tested (`get_mcp_server_log_name`), and `server.py` already used it for logs. the raised errors just never got it.

fix:

- add `MCPServer._error_name`, a thin wrapper over that helper. credentials, query, and fragment are stripped; host and path stay so the error still tells you which server failed.
- apply it to the 16 raised-error sites in mcp/server.py, plus the two raised errors in mcp/util.py that sit directly next to already-redacted log calls.
- functional uses of `server.name` (tool namespacing, span identity) are deliberately unchanged.

before:

```
UserError: Failed to connect to MCP server 'sse: https://user:s3cr3t_pw@mcp.example.com/sse?api_key=SECRET_QS_KEY': HTTP error 503
```

after:

```
UserError: Failed to connect to MCP server 'sse: https://mcp.example.com/sse': HTTP error 503
```

### Test plan

- 6 new tests in tests/mcp/test_server_errors.py: the `_error_name` property (default url name redacted, explicit name untouched), the connect / list_tools / call_tool raised errors, and the `invoke_mcp_tool` AgentsException. each asserts the password and api key are gone and the host is kept.
- `make format` and `make lint`: clean.
- `make tests`: full suite green (5770 + 28 passed, 0 failed). tests/mcp/ specifically: 239 passed.
- `make typecheck`: adds 0 new errors.
- ran `.agents/skills/code-change-verification/scripts/run.sh`: format and lint pass, then it stops at typecheck on 31 pre-existing errors unrelated to this change (vercel extension name-collision, py3.12-only apis on a 3.11 checkout, boto3 not installed, redis `type: ignore` skew).

### Checks

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -8,7 +8,7 @@ from collections.abc import AsyncGenerator, Awaitable, Callable
 from contextlib import AbstractAsyncContextManager, AsyncExitStack, asynccontextmanager
 from datetime import timedelta
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Literal, NoReturn, TypeVar, Union, cast
 
 import anyio
 import httpx
@@ -234,6 +234,19 @@ MCPStreamTransport = (
         GetSessionIdCallback | None,
     ]
 )
+
+
+def _raise_mcp_transport_error(error: UserError, cause: Exception | None) -> NoReturn:
+    """Raise ``error`` for a failed MCP transport call, chaining ``cause`` only when allowed.
+
+    ``cause`` is typically an httpx error whose request URL can embed credentials, so the
+    chain is dropped while tool-data logging is disabled. Call this *outside* the ``except``
+    block: ``raise ... from None`` is not enough, because the original exception would stay
+    reachable through ``__context__``.
+    """
+    if _debug.DONT_LOG_TOOL_DATA or cause is None:
+        raise error
+    raise error from cause
 
 
 class MCPServer(abc.ABC):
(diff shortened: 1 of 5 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying credential-exposure issue is valid, and reusing the existing MCP URL sanitizer is directionally correct. Before we can merge this, please make the redaction cover the complete caller-visible exception graph, not only str() of the outer exception.

The current HTTP wrappers still chain the credential-bearing httpx exception, while the generic MCPUtil wrapper copies and chains the raw exception. Please extract only safe diagnostic fields and raise the sanitized SDK exception without retaining sensitive `__cause__` or `__context__` data. When tool-data redaction is enabled, generic exception details must also remain hidden.

Please add regression tests asserting that the credentials are absent from the outer message, recursive cause/context chain, and rendered traceback. Once those focused changes are covered, this should be ready for another review.

seratch (changes requested):
Thanks for addressing the exception-chain feedback. The latest revision now goes substantially beyond the URL-credential issue: it turns OPENAI_AGENTS_DONT_LOG_TOOL_DATA from a documented logging control into runtime exception policy, hides otherwise safe server and tool identifiers, and adds machinery for task and traceback-frame locals plus hostile exception class names.

Please reset this PR to the narrower contract: always sanitize URL-derived server names independently of the logging flag; ensure auth credentials are absent from the outer message, normal rendered traceback, `__cause__`, and `__context__`; and preserve existing non-secret names and diagnostics. Do not attempt to guarantee redaction of arbitrary custom names, exception type names, or third-party frame-local capture.
Please replace the current broad machinery and combinatorial tests with focused regression coverage for that contract.

seratch:
Thanks for working here. The merged https://github.com/openai/openai-agents-python/pull/4015 includes your contribution.

PR #4012, opened 2026-07-30
fix: redact tool-argument values from invalid-input errors

ioleksiuk:
### Summary

With `DONT_LOG_TOOL_DATA` set, the `ModelBehaviorError` raised for an invalid function-tool or Codex-tool argument still embedded the raw values. Pydantic's `ValidationError` string carries `input_value=...`, and `raise ... from e` attached the payload-bearing exception as `__cause__`/`__context__`, so tool data surfaced in tracebacks and telemetry even though the adjacent log calls at these sites are already gated on `DONT_LOG_TOOL_DATA`.

This extends the hardening already applied to the sibling `_parse_function_tool_json_input` to the type-validation path in `tool.py` and to `codex_tool.py`: under `DONT_LOG_TOOL_DATA`, raise a payload-free summary (field location + error type, no values) from outside the `except` block, so the `ValidationError` is neither interpolated into the message nor attached as `__context__`. Full detail and exception chaining are preserved when tool-data logging is enabled, and the redacted error still names the invalid field so the model can self-correct.

### Test plan

Ran the full local stack — `make format`, `make lint`, `make typecheck`, `make tests` — all pass. Added regression tests in `tests/test_error_logging_redaction.py` and `tests/extensions/experiemental/codex/test_codex_tool.py` asserting that the raw value is absent (and `__cause__`/`__context__` are `None`) when redaction is on, present when off, and that the invalid field name still reaches the caller.

### Issue number

None (defense-in-depth; completes the redaction already applied to the sibling parser).

### Checks

- [x] I've run the verification stack (`make format && make lint && make typecheck && make tests`)
- [ ] If using Codex, I've run `/review` before submitting this PR (N/A)

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -11,6 +11,7 @@ from openai.types.responses.response_prompt_param import ResponsePromptParam
 from pydantic import BaseModel, TypeAdapter, ValidationError
 from typing_extensions import NotRequired, TypedDict
 
+from . import _debug
 from ._tool_identity import get_function_tool_approval_keys
 from .agent_output import AgentOutputSchemaBase
 from .agent_tool_input import (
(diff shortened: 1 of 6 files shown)


seratch (changes requested):
Thanks for the update. The latest revision fixes the demonstrated mapping-key and unknown top-level-key leaks, but this needs one scope reset before merge.

Please remove field/error summarization from redacted mode and emit only the fixed base error message. The repository's redaction contract requires a fixed message without inspecting sensitive exception metadata, and the successive loc special cases show that classifying Pydantic locations is not a stable security boundary. Preserve the current full validation detail and exception chaining when `DONT_LOG_TOOL_DATA=False`.

Please also apply the same redacted behavior to `Agent.as_tool(parameters=...)`, which still interpolates and chains the complete ValidationError. Add caller-level adversarial tests for function tools, agent-as-tool, and Codex tools asserting that the sentinel is absent and both `__cause__` and `__context__` are None in redacted mode.

seratch (approved):

seratch:
Once the codex cloud review says :+1:, this can be merged.

seratch (changes requested):
Let me know once codex reviews say 👍

seratch:
Thanks again for working on this here. Your contribution is included in #4182 via co-authored-by credit.

PR #4022, opened 2026-07-30
fix(models): ignore a non-finite Retry-After instead of sleeping forever

### LINKED ISSUE #4021: `Retry-After: inf` is accepted as a delay and hangs the run forever in `asyncio.sleep()`
### What happens

A `Retry-After: inf` response header makes a run hang forever instead of retrying.

`parse_retry_after_value` and `parse_retry_after_ms` in `src/agents/models/_retry_runtime.py` build the delay with `float()`, then accept anything `>= 0`:

```python
try:
    parsed = float(value)
except ValueError:
    parsed = None
if parsed is not None:
    return parsed if parsed >= 0 else None
```

`float()` accepts `inf`, `+inf`, `INF`, `Infinity`, `infinity`, and any overflowing literal such as `1e400`. None of these are valid `Retry-After` values under RFC 9110, but `inf >= 0` is true, so each is returned as a usable delay.

Measured against the current parsers:

| header value | `float()` | `parse_retry_after_value` | `parse_retry_after_ms` |
| --- | --- | --- | --- |
| `-1` | `-1.0` | `None` ✓ | `None` ✓ |
| `nan` | `nan` | `None` ✓ | `None` ✓ |
| `inf` | `inf` | **`inf`** | **`inf`** |
| `Infinity` | `inf` | **`inf`** | **`inf`** |
| `1e400` | `inf` | **`inf`** | **`inf`** |
| `30` | `30.0` | `30.0` ✓ | `0.03` ✓ |

The negative and NaN cases are already rejected. `inf` is the same class of unusable value and is let through.

### Why it hangs rather than just over-waiting

The value flows `get_retry_after` → `ModelRetryNormalizedError.retry_after` → `RetryDecision.delay` → `_sleep_for_retry`, which calls `asyncio.sleep(delay)` directly. Nothing on that path bounds it:

- `backoff.max_delay` is applied only inside `_default_retry_delay`, i.e. the computed-backoff branch. `_evaluate_retry` uses `normalized.retry_after` **instead of** calling `_default_retry_delay` when a hint is present. This is documented and intended (`docs/models/index.md`: "`backoff.max_delay` caps this computed backoff delay only. It does not cap explicit delays returned by a policy or retry-after hints"), which makes the parser the only place an unusable value can be rejected.
- `_sleep_for_retry` only short-circuits on `delay <= 0`.

So `asyncio.sleep(float("inf"))` never returns, the retry budget is never spent, and the run never raises — it just stops. Confirmed locally: `asyncio.wait_for(_sleep_for_retry(float("inf")), timeout=2.0)` times out with the sleep still pending.

End-to-end, with a 429 carrying `Retry-After: inf` and `ModelRetryBackoffSettings(initial_delay=0.5, jitter=False)`, the delay handed to `asyncio.sleep` is `inf` rather than `0.5`.

### Prior art

The `openai` SDK that this repo wraps bounds the header before using it (`_base_client._calculate_retry_timeout`):

```python
retry_after = self._parse_retry_after_header(response_headers)
if retry_after is not None and 0 < retry_after <= 60:
    return retry_after
```

It falls back to exponential backoff otherwise. The agents-SDK path has no equivalent bound.

### Reachability

`get_retry_after` is called from `_normalize_retry_error` for every retryable model error, so any provider or proxy that emits a malformed `Retry-After` reaches this. It needs no unusual configuration — only a policy that retries, e.g. `retry_policies.http_status([429])` or `retry_policies.retry_after()`.

### Existing coverage

`tests/models/test_openai_retry_helpers.py` asserts `_parse_retry_after("-1") is None` and `_parse_retry_after_ms("-100") is None`, so rejecting unusable header values is clearly the intended contract. There is no case for non-finite input.

### Expected

A non-finite `Retry-After` is ignored and the configured backoff is used, so the run still makes progress. Long-but-finite waits should stay as they are — those remain the provider's call.

### Suggested fix

Reject non-finite values in both parsers, alongside the negatives and NaN they already refuse:

```python
return parsed if math.isfinite(parsed) and parsed >= 0 else None
```

Happy to open a PR; I have this working locally with regression tests at both the parser and `get_response_with_retry` levels.

### Environment

`main` (`992abf7`), Python 3.12, Windows.

LHMQ878:
### Summary

A `Retry-After: inf` header currently parks a run forever instead of retrying it.

`parse_retry_after_value` and `parse_retry_after_ms` build the delay with `float()` and then accept anything `>= 0`. `float()` accepts `inf`, `+inf`, `INF`, `Infinity`, `infinity`, and overflowing literals such as `1e400` — none of which RFC 9110 permits in `Retry-After` — and `inf >= 0` is true, so each is returned as a usable delay. The negative and NaN forms of the same problem are already rejected by that guard (`-inf >= 0` and `nan >= 0` are both false); only `inf` slips through.

Nothing downstream bounds it. The value flows `get_retry_after` → `ModelRetryNormalizedError.retry_after` → `RetryDecision.delay` → `_sleep_for_retry`, which calls `asyncio.sleep(delay)` and only short-circuits on `delay <= 0`. `backoff.max_delay` is applied solely inside `_default_retry_delay`, and `_evaluate_retry` uses the retry-after hint *instead of* calling that — which is deliberate and documented (`docs/models/index.md`: "`backoff.max_delay` caps this computed backoff delay only. It does not cap explicit delays returned by a policy or retry-after hints"). That makes the parser the only place an unusable value can be rejected, so this fix stays there and leaves the documented `max_delay` semantics untouched.

The fix adds `math.isfinite()` alongside the existing `>= 0` check, via a small shared `_usable_delay()` helper so both parsers keep identical behavior. Long-but-finite waits are unchanged — a 24-hour `Retry-After` is still honored, because that remains the provider's call. Only values a run could never wake up from are dropped, and the caller then falls back to the configured backoff.

For reference, the `openai` SDK this repo wraps bounds the header before using it in `_base_client._calculate_retry_timeout` (`if retry_after is not None and 0 < retry_after <= 60`), falling back to exponential backoff otherwise.

### Test plan

Three regression tests, at both levels:

- `tests/models/test_openai_retry_helpers.py::test_parse_retry_after_rejects_non_finite_delays` — parametrized over `inf`, `+inf`, `INF`, `Infinity`, `infinity`, `1e400` for both parsers.
- `tests/models/test_openai_retry_helpers.py::test_parse_retry_after_keeps_large_finite_delays` — pins that `86400` is still accepted, so the change is not a general clamp.
- `tests/models/test_model_retry.py::test_get_response_with_retry_falls_back_to_backoff_for_infinite_retry_after` — end-to-end: a 429 carrying `Retry-After: inf` with `initial_delay=0.5, jitter=False` must sleep `0.5`, retry, and succeed.

Verified:

- The three new tests plus the existing retry suites: **81 passed**.
- **Control:** reverting only `_retry_runtime.py` and rerunning the same set gives **7 failed / 74 passed**, with the end-to-end case failing as `assert [inf] == [0.5]` — confirming the run really does sleep on infinity today.
- `tests/models/`: **602 passed, 4 skipped**.
- Lint/format: `ruff format --check` clean, `ruff check` "All checks passed!".
- Typecheck: `mypy` "Success: no issues found", `pyright` "0 errors, 0 warnings".
- `pytest -n auto --dist loadfile -m "not serial"`: failing-file set byte-identical before and after the change (5 pre-existing tracing files).
- `pytest -m serial`: 3 failed / 4 passed / 27 skipped, identical on a clean tree — all three are `ConnectionError` in `tests/extensions/` and need network access.

`.agents/skills/code-change-verification/scripts/run.sh` itself could not run here: this is a Windows checkout with no `make` on `PATH`, and the script drives every step through `make`. I ran each of its steps directly instead — `format`, `lint`, `mypy`, `pyright`, `tests-parallel`, `tests-serial` — with the results above. Happy to rerun anything under CI.

### Issue number

Closes #4021

### Checks

- [ ] I've run `.agents/skills/code-change-verification/scripts/run.sh` (no `make` available on this Windows checkout; ran each step it invokes directly — see test plan)

### DIFF
--- a/src/agents/models/_retry_runtime.py
+++ b/src/agents/models/_retry_runtime.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 import time
 from collections.abc import Iterator, Mapping
 from contextlib import contextmanager
@@ -55,6 +56,17 @@ def get_error_header(error: Exception, key: str) -> str | None:
     return None
 
 
+def _usable_delay(parsed: float) -> float | None:
+    """Return the delay only when it is a wait a caller can actually finish.
+
+    `float()` accepts `inf`, `Infinity`, and overflowing literals such as `1e400`, none of
+    which RFC 9110 allows in `Retry-After`. An infinite delay reaches `asyncio.sleep()`
+    unchanged and parks the run permanently, so it is rejected here alongside the negative
+    and NaN values this module already refuses.
+    """
+    return parsed if math.isfinite(parsed) and parsed >= 0 else None
+
+
 def parse_retry_after_ms(value: str | None) -> float | None:
     if value is None:
         return None
@@ -62,7 +74,7 @@ def parse_retry_after_ms(value: str | None) -> float | None:
         parsed = float(value) / 1000.0
     except ValueError:
         return None
-    return parsed if parsed >= 0 else None
+    return _usable_delay(parsed)
 
 
 def parse_retry_after_value(value: str | None) -> float | None:
@@ -74,7 +86,7 @@ def parse_retry_after_value(value: str | None) -> float | None:
     except ValueError:
         parsed = None
     if parsed is not None:
-        return parsed if parsed >= 0 else None
+        return _usable_delay(parsed)
 
     try:
         retry_datetime = parsedate_to_datetime(value)
--- a/tests/models/test_model_retry.py
+++ b/tests/models/test_model_retry.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import math
 from collections.abc import AsyncIterator
 from typing import Any, cast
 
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for the patch and thorough tests. The tests demonstrate that the synthetic malformed header reaches the retry sleep path, but they do not establish that an OpenAI API or supported provider/gateway emits this value. Without that real-path evidence, the underlying need remains unproven despite the implementation being technically feasible.

I am going to close this PR for now. If an actual supported provider or proxy produces a non-finite Retry-After, we can reconsider this narrow parser-level fix with that reproduction.

PR #4023, opened 2026-07-30
fix: consolidate same-exception re-raises into bare raise across all modules

AAliKKhan:
## Summary

Replace `raise e` / `raise exc` with bare `raise` inside `except` blocks to avoid adding an artificial re-raise frame to the traceback. While the original failure frame is preserved with either form, bare `raise` keeps the traceback clean without an extra re-raise frame.

## Changes

All changes follow the same pattern: inside an `except` handler, replace `raise e` (or `raise exc`) with bare `raise` to re-raise the currently handled exception without adding an extra traceback frame.

- **src/agents/voice/models/openai_stt.py** (5 instances) — `except Exception as e: ... raise e` → `raise`
- **src/agents/voice/pipeline.py** (2 instances) — `except Exception as e: ... raise e` → `raise`
- **src/agents/voice/result.py** (1 instance) — `except Exception as e: ... raise e` → `raise`
- **src/agents/run_internal/tool_execution.py** (1 instance) — `except Exception as e: ... raise e` → `raise`
- **src/agents/mcp/server.py** (1 instance) — `except IsolatedSessionRetryFailed as exc: ... raise exc` → `raise`

No behavior change: the same exception object is re-raised in all cases. The only difference is the traceback will not include an artificial re-raise frame.

## Test plan

- Existing tests pass
- Mechanical Python-language cleanup — no runtime behavior change

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -1679,7 +1679,7 @@ class MCPServerStreamableHttp(_MCPServerWithClientSession):
                     if self.max_retry_attempts != -1 and retries_used >= self.max_retry_attempts:
                         if exc.__cause__ is not None:
                             raise exc.__cause__ from exc
-                        raise exc
+                        raise
                     backoff = self.retry_backoff_seconds_base * (2 ** (retries_used - 1))
                     await asyncio.sleep(backoff)
                 except Exception:
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -1690,7 +1690,7 @@ class _FunctionToolBatchExecutor:
                     )
                 )
                 if isinstance(e, AgentsException):
-                    raise e
+                    raise
                 raise UserError(f"Error running tool {func_tool.name}: {e}") from e
 
             if self.config.trace_include_sensitive_data:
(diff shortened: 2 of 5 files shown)


seratch (changes requested):
Thanks for the cleanup. The underlying change is valid: bare `raise` avoids adding an artificial re-raise frame, although `raise e` does not remove the original failure frame as the PR description suggests.

I do not think these changes should be reviewed and merged as five separate PRs (#4023-#4027). Please consolidate the same-exception re-raises into one focused PR, while retaining explicit raises that occur outside an active `except` block or intentionally wrap or replace an exception. Please also describe this as traceback cleanup rather than recovery of a lost error location.

A dedicated behavior test is probably unnecessary for this mechanical Python-language cleanup. If practical, enabling Ruff `TRY201` would prevent the pattern from recurring.

seratch:
Also, could you make the PR descriptions a bit easier to read? The current formatting is difficult to follow.

AAliKKhan:
Updated as requested: consolidated all same-exception re-raises into this single PR across all 5 files. Also updated the description to describe it as traceback cleanup (avoiding an artificial re-raise frame) rather than recovery of lost error location.

seratch (approved):
Thanks for consolidating the changes and updating the description. The revised scope is coherent, the traceback behavior is now described accurately, and all required checks are green. This looks good to me.

PR #4024, opened 2026-07-30
fix: preserve original exception traceback in voice pipeline re-raises

AAliKKhan:
## Summary

Replace \
aise e\ with bare \
aise\ inside \except\ blocks in \pipeline.py\ to preserve the original exception traceback when re-raising. In Python 3.10 (the minimum supported version), \
aise e\ resets the traceback to the re-raise line, obscuring the original error location.

## Changes

\src/agents/voice/pipeline.py\: Changed 2 instances of \
aise e\ to bare \
aise\ inside \except\ handlers.

## Test plan

- Existing tests pass (\make tests\)
- No behavior change: the same exception is re-raised, only the traceback is preserved unchanged

### DIFF
--- a/src/agents/voice/pipeline.py
+++ b/src/agents/voice/pipeline.py
@@ -115,7 +115,7 @@ class VoicePipeline:
                 except Exception as e:
                     log_model_and_tool_action_error(logger, "Error processing single voice turn", e)
                     await output._add_error(e)
-                    raise e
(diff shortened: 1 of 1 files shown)


seratch:
Closing this due to https://github.com/openai/openai-agents-python/pull/4023#pullrequestreview-4820065293

PR #4029, opened 2026-07-30
fix: add missing None guards to ItemHelpers text/refusal methods

AAliKKhan:
## Summary

Two missing None guards in `ItemHelpers` that are inconsistent with the guarded patterns used in sibling methods.

### 1. `text_message_output` refusal path (line 711)

The text path (`last_content.text`) is guarded with `or ""` because provider gateways and `model_construct` can surface `None`. The refusal path (`last_content.refusal`) is not guarded, even though `model_construct` can produce `ResponseOutputRefusal.refusal=None` just as easily. The method return type is `str` (not `str | None`), so returning `None` is a type error.

### 2. `extract_last_text` text path (line 723)

`last_content.text` is returned without `or ""` guard, while `extract_text` and `text_message_output` both use `or ""` for the same field. Provider gateways and `model_construct` can surface `None` on text fields.

## Test plan

- Existing tests pass
- No behavior change for normal (non-None) inputs

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -708,7 +708,7 @@ class ItemHelpers:
             # receiving refusal text. A ``None`` value requires bypassing model validation
             # with ``model_construct``, so this intentionally does not mirror the fallback
             # above.
-            return last_content.refusal
+            return last_content.refusal or ""
         else:
             raise ModelBehaviorError(f"Unexpected content type: {type(last_content)}")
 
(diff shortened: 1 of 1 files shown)


seratch:
@AAliKKhan As explained in the code comment, the absence of `None` guards is intentional here. Please do not submit PRs for hypothetical issues that cannot occur through supported code paths.

PR #4030, opened 2026-07-30
fix: use isinstance instead of type() for int checks in run_state deserialization

AAliKKhan:
## Summary

Replace `type(x) is not int` with `not isinstance(x, int)` in three places during RunState deserialization validation.

`type()` rejects bool, which is a subclass of int — `type(True) is not int` evaluates to `True` (since `type(True)` is `bool`, not `int`). Using `isinstance` ensures bool values (`True` = 1, `False` = 0) are accepted as valid integer indexes.

## Changes

`src/agents/run_state.py`:
- Line 2867: `type(session_index_value) is not int` → `not isinstance(session_index_value, int)`
- Line 2909: `type(item_ref.get("index")) is not int` → `not isinstance(item_ref.get("index"), int)`
- Line 2913: `type(item_ref.get("input_index")) is not int` → `not isinstance(item_ref.get("input_index"), int)`

## Test plan

- Existing tests pass

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -2864,7 +2864,7 @@ async def _build_run_state_from_json(
                 if session_index_value is None:
                     continue
                 if (
-                    type(session_index_value) is not int
+                    not isinstance(session_index_value, int)
                     or session_index_value < 0
                     or session_index_value >= len(serialized_session_items)
                     or session_index_value in used_session_indexes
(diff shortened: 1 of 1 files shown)


seratch:
@AAliKKhan Thanks for submitting this. I am closing this PR because the exact-integer checks are intentional.

These fields are serialized array indexes. JSON booleans are distinct from JSON numbers, even though Python implements `bool` as a subclass of `int`. Accepting `False` as index `0` and `True` as index `1` would silently reinterpret malformed persisted state and could bind generated or nested-history items to the wrong session entries. The existing `test_invalid_generated_session_item_indexes_are_ignored` test explicitly includes `[True, None]` as invalid input, so this change directly reverses the intended contract.

This also repeats the broader pattern seen in #3972 and #3976: expanding SDK behavior for values that require bypassing validation or violating a serialized or typed contract, without a concrete reproduction through a supported public path. Going forward, please do not open additional PRs of this kind based only on constructible edge cases or Python implementation details. If you believe an existing validation rule is incorrect, please first open an issue with a supported-path reproduction, the actual user-visible impact, and evidence that the input belongs to the SDK contract.

PR #4034, opened 2026-07-30
feat(core): expose original wrapped callable on FunctionTool

### LINKED ISSUE #3381: Provide stable public access to the underlying function on `FunctionTool`

Rajeev91691:
This PR resolves #3381.

### Description
Currently, when `@function_tool` is applied to a function, the resulting `FunctionTool` captures the original callable only in the closure of `_on_invoke_tool_impl`. There is no public attribute (`func`) and no `__call__` forwarding to execute the tool directly. This prevents external test suites, packaging SDKs, or migrations from accessing or invoking the wrapped function directly.

This PR:
1. Adds a private `_func` field to `FunctionTool` to hold the original callable.
2. Exposes a public read-only property `func` returning `_func`.
3. Implements `__call__` on `FunctionTool` to forward calls directly to the underlying `_func`, allowing direct invocation like `my_tool(x)`.
4. Appends a unit test case to `tests/test_function_tool.py` verifying the new behavior.

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -520,12 +520,26 @@ class FunctionTool:
     _emit_tool_origin: bool = field(default=True, kw_only=True, repr=False)
     """Whether runtime item generation should emit tool origin metadata for this tool."""
 
+    _func: Callable[..., Any] | None = field(default=None, kw_only=True, repr=False)
+    """Internal reference to the underlying function."""
+
     @property
     def qualified_name(self) -> str:
         """Return the public qualified name used to identify this function tool."""
         return (
             tool_qualified_name(self.name, get_explicit_function_tool_namespace(self)) or self.name
         )
+
+    @property
+    def func(self) -> Callable[..., Any] | None:
+        """Return the underlying function wrapped by this tool."""
+        return self._func
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        """Forward calls directly to the underlying function."""
+        if self._func is None:
+            raise RuntimeError("FunctionTool has no underlying function")
+        return self._func(*args, **kwargs)
 
     def __post_init__(self):
         self.allowed_callers = _normalize_tool_allowed_callers(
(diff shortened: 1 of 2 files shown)


seratch:
Closing this in favor of https://github.com/openai/openai-agents-python/pull/4038

PR #4041, opened 2026-07-31
fix: validate streamed cancel mode and reject invalid values

nathon-lee:
## Title

fix: validate streamed cancel mode and reject invalid values

## Description

This PR addresses a bug where an invalid stream cancel mode could be accepted silently, which could cause cancellation to be skipped and leave the stream in an inconsistent state.

- Adds runtime validation in `result.py` so `RunResultStreaming.cancel` only accepts `immediate` and `after_turn`.
- Raises a `ValueError` for unsupported cancel mode input instead of proceeding silently.
- Adds a regression test in `test_soft_cancel.py` to verify invalid mode rejection and confirm that internal cancel state is not mutated on failure.

This change preserves existing behavior for valid modes while making invalid usage fail fast with a clear error.

### DIFF
--- a/src/agents/result.py
+++ b/src/agents/result.py
@@ -753,6 +753,11 @@ class RunResultStreaming(RunResultBase):
         Note: After calling cancel(), you should continue consuming stream_events()
         to allow the cancellation to complete properly.
         """
+        if mode not in ("immediate", "after_turn"):
+            raise ValueError(
+                f"Invalid cancel mode: {mode!r}. Expected one of: 'immediate', 'after_turn'."
+            )
+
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing this idea. However, we'd like to hold off adding this type of validation because the literal values could be updated in the future, and having custom logic like this could bring maintenance overhead.

nathon-lee:
> Thanks for sharing this idea. However, we'd like to hold off adding this type of validation because the literal values could be updated in the future, and having custom logic like this could bring maintenance overhead.

Thanks, @seratch  ! Appreciate the review and the helpful feedback.

PR #4042, opened 2026-07-31
fix: avoid deadlock when agent-as-tool on_stream callback raises CancelledError

nathon-lee:
## Title

fix: avoid deadlock when agent-as-tool on_stream callback raises CancelledError

## Description

This pull request fixes a deadlock risk in the agent-as-tool streaming path when the on_stream callback raises CancelledError from user callback code.

### Background

In the streaming callback dispatcher, a callback-level CancelledError could prematurely terminate the dispatcher task and leave the queue completion flow waiting on join, which may block tool completion in cancellation-adjacent scenarios.

### What changed

1. Updated callback handling logic in the streaming dispatcher to distinguish two cancellation sources:
   - Real task cancellation (dispatcher task is actively being cancelled): re-raise CancelledError to preserve expected cancellation semantics.
   - Callback-raised CancelledError from user handler code: treat it as callback failure, log it, and continue draining the queue.
2. Added a regression test to ensure callback-raised CancelledError does not deadlock the tool invocation path.
3. Preserved existing parent-cancellation behavior.

### Why this is safe

- The change is narrow and localized to callback dispatch behavior.
- It preserves true cancellation propagation for task-level cancellation.
- It only converts callback-originated CancelledError into non-fatal callback failure handling, consistent with existing exception-tolerant callback behavior.

Thank you very much for reviewing this change. I sincerely appreciate your time and feedback, and I am very happy to iterate quickly if you prefer a different handling strategy.

### Files changed

- agent.py
- test_agent_as_tool.py

### Focused tests for this fix

```bash
cd /path/to/openai-agents-python_woo
uv run pytest -q tests/test_agent_as_tool.py -k "streaming_callback_cancelled_error_does_not_deadlock or streaming_reraises_parent_cancellation_without_waiting_for_handler"
```

### New regression test only

```bash
cd /path/to/openai-agents-python_woo
uv run pytest -q tests/test_agent_as_tool.py -k "streaming_callback_cancelled_error_does_not_deadlock"
```

### Full verification (repo standard order)

```bash
cd /path/to/openai-agents-python_woo
make format
make lint
make typecheck
make tests
```

### If `uv` is unavailable on Linux

```bash
cd /path/to/openai-agents-python_woo
python -m pytest -q tests/test_agent_as_tool.py -k "streaming_callback_cancelled_error_does_not_deadlock or streaming_reraises_parent_cancellation_without_waiting_for_handler"
```

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -855,6 +855,20 @@ class Agent(AgentBase, Generic[TContext]):
                             maybe_result = stream_handler(payload)
                             if inspect.isawaitable(maybe_result):
                                 await maybe_result
+                        except asyncio.CancelledError as exc:
+                            current_task = asyncio.current_task()
+                            if current_task is not None and current_task.cancelling():
+                                raise
+
+                            def diagnostic_extra() -> dict[str, object]:
+                                return {"agent_name": self.name}
+
+                            log_model_and_tool_action_error(
+                                logger,
+                                "Error while handling an agent tool on_stream event",
+                                exc,
+                                diagnostic_extra=diagnostic_extra,
+                            )
                         except Exception as exc:
 
                             def diagnostic_extra() -> dict[str, object]:
--- a/tests/test_agent_as_tool.py
+++ b/tests/test_agent_as_tool.py
@@ -2251,6 +2251,59 @@ async def test_agent_as_tool_streaming_reraises_parent_cancellation_without_wait
 
 
 @pytest.mark.asyncio
+async def test_agent_as_tool_streaming_callback_cancelled_error_does_not_deadlock(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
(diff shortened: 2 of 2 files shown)


seratch:
I confirmed that this code path can hang when an on_stream callback directly raises CancelledError: the dispatcher exits while the invocation can remain blocked in event_queue.join().

However, this is the same synthetic scenario previously considered in #3829 and #3831. That proposal was closed pending a realistic supported callback with independently cancelled work and an explanation of why the callback that owns that work cannot handle the cancellation itself. This PR adds a smaller implementation, but its regression test still directly raises CancelledError from a mocked callback and does not provide that missing evidence. It would also establish SDK-wide behavior that callback-raised cancellation is logged and ignored.

I am going to close this PR for now. We can revisit the contract if the requested real-world evidence is provided.

PR #4060, opened 2026-07-31
fix(voice): clean up tasks when streams close early

### LINKED ISSUE #4051: [Voice] Early exit from voice async generators can leave owned tasks running
### Describe the bug

The voice streaming generators only perform cleanup after normal exhaustion, or when cancellation is raised at a specific queue wait. If a consumer stops iteration immediately after receiving a yielded event (for example with `break` or `aclose()`), generator finalization can bypass the cleanup code because the yield loops are not protected by `try/finally`.

Affected paths:

- `StreamedAudioResult.stream()`: https://github.com/openai/openai-agents-python/blob/000a96b602889b00f7cfaa210c41e1a74be65272/src/agents/voice/result.py#L303-L338
- `OpenAIRealtimeTranscriptionSession.transcribe_turns()`: https://github.com/openai/openai-agents-python/blob/000a96b602889b00f7cfaa210c41e1a74be65272/src/agents/voice/models/openai_stt.py#L342-L379

There is a second lifecycle gap in both implementations: cleanup requests cancellation with `task.cancel()`, but does not await/drain the owned tasks. Returning from cleanup therefore does not prove that websocket context managers, task finalizers, or tracing cleanup have completed.

### Debug information

- Agents SDK version: v0.19.1
- Evidence type: static async-generator lifecycle trace

### Repro steps

A focused local test can:

1. Create a `StreamedAudioResult` with a producer task that records entry and finalization.
2. Queue one event so `stream()` reaches its `yield`.
3. Consume that event, then call `await iterator.aclose()` (or leave an `async for` loop with `break`).
4. Assert that the producer, dispatcher, and text-generation tasks have been cancelled and fully settled.

The existing cancellation test cancels while the consumer is blocked in `queue.get()`. That exercises the `except asyncio.CancelledError` at lines 307-311, but not generator close after a successful yield.

The same shape should be tested for `transcribe_turns()`: yield one transcript, close the iterator, and assert that the websocket and all four owned tasks have settled.

### Expected behavior

Ending consumption through normal exhaustion, cancellation, `break`, or `aclose()` should deterministically finalize the active trace/turn, close the websocket/session when owned by the generator, cancel outstanding producer tasks, and await those tasks without masking the original consumer exception.

### Why this matters

Early termination is normal async-iterator usage. Leaked voice/STT tasks can retain websocket connections, continue consuming audio, produce unobserved task exceptions, and leave traces unfinished.

### Suggested scope

Wrap each generator lifecycle in structured `try/finally` cleanup and drain cancelled tasks with appropriate `CancelledError` handling. Regression tests should cover both cancellation while blocked and generator closure immediately after a yield.

I searched the current issues/PRs for voice stream early exit, async-generator close, and STT cleanup. The nearby voice cancellation changes do not appear to cover this generator-finalization path.

GautamSharma99:
This pull request fixes voice background-task and tracing cleanup when a consumer explicitly closes or cancels a voice async iterator.

Closes #4051.

## Problem

`StreamedAudioResult.stream()` and `OpenAISTTTranscriptionSession.transcribe_turns()` previously ran cleanup only after normal iterator exhaustion or from cancellation handling around a queue read. Calling `aclose()` while either concrete async generator was suspended at a successful `yield` skipped trailing cleanup.

The cleanup helpers also requested task cancellation without draining owned work. Callers could therefore return before producer, dispatcher, synthesis, websocket, listener, or STT processing tasks had run their finalizers.

Two additional ordering hazards existed:

- STT cleanup finished the active span before awaiting WebSocket close, allowing event processing to start a replacement span during that await.
- `StreamedAudioResult.stream()` recorded `session_ended` only after yielding it. Immediate `aclose()` injected `GeneratorExit` before the flag changed, so cleanup cancelled the producer rather than allowing normal trace and session finalization.

Finally, attaching a secondary cleanup exception to a warning could render its implicit context, including the primary consumer traceback, when diagnostic logging was enabled.

## Changes

- Run iterator cleanup from one finalization boundary for normal exhaustion, consumer cancellation, and explicit `aclose()`.
- Record terminal `session_ended` delivery before yielding the event.
- During finalization after terminal delivery, shield the pipeline producer until it completes normally, then drain remaining owned tasks.
- Convert cleanup helpers to async cancel-and-drain operations and exclude the current cleanup task.
- Close the STT transport, cancel and drain session-owned work, and only then finish the final active transcription span.
- Preserve output-queue `task_done()` accounting when the STT iterator closes at a yield.
- Preserve an active consumer cancellation or exception when finalization fails.
- Emit only a fixed operational warning for secondary cleanup failure, without attaching the cleanup exception, arguments, traceback, or implicit consumer context.

## Regression coverage

Controlled `asyncio.Event` interleavings assert that:

- explicit close after a regular yielded event cancels and settles all owned tasks;
- explicit close immediately after receiving `session_ended` remains pending until the producer completes normally and does not cancel it;
- a transcription completion processed while WebSocket close is suspended cannot leave a replacement span unfinished;
- cleanup failure cannot replace `CancelledError` or an exception injected at a successful yield;
- cleanup-failure warnings remain fixed static records even with diagnostic model logging enabled; and
- STT websocket, task, tracing, and output-queue ownership remains deterministic.

## Scope

The deterministic cleanup guarantee covers normal exhaustion, cancellation, and explicit `aclose()`. A bare `async for` `break` does not synchronously await async-generator close, so this pull request does not claim synchronous cleanup at the point of a bare break.

## Compatibility

This is an internal lifecycle fix against the v0.19.1 behavior boundary. It does not change public signatures, event formats, provider payloads, configuration, or persisted state.

## Validation

- `make format`
- `make lint`
- `make typecheck`
- `make tests`
- Focused voice and STT suite: 53 passed
- Sensitive-logging collector tests: 10 passed

### DIFF
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -12,7 +12,7 @@ from openai import AsyncOpenAI
 
 from ... import _debug
 from ...exceptions import AgentsException, UserError
-from ...logger import logger
+from ...logger import log_model_action_warning, logger
 from ...tracing import Span, SpanError, TranscriptionSpanData, transcription_span
 from ...util._error_tracing import get_trace_error
 from ..exceptions import STTWebsocketConnectionError
@@ -374,7 +374,21 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
                     yield turn
                 finally:
                     self._output_queue.task_done()
-        finally:
+        except BaseException:
+            try:
+                await self.close()
+            except BaseException as cleanup_error:
+                try:
+                    log_model_action_warning(
+                        logger,
+                        "STT session cleanup failed while preserving the consumer exception",
+                        cleanup_error,
+                    )
+                except Exception:
+                    # Logging must not replace the original consumer exception.
+                    pass
+            raise
+        else:
             await self.close()
 
         self._check_errors()
@@ -382,12 +396,14 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
             raise self._stored_exception
 
     async def close(self) -> None:
-        self._end_turn("")
         try:
             if self._websocket:
                 await self._websocket.close()
         finally:
-            await self._cleanup_tasks()
+            try:
+                await self._cleanup_tasks()
+            finally:
+                self._end_turn("")
 
 
 class OpenAISTTModel(STTModel):
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -381,8 +381,13 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
         finally:
             try:
                 await self.close()
-            except BaseException:
-                if not primary_exception_active:
+            except BaseException as cleanup_exception:
+                # Cancellation must always propagate, even while preserving a primary consumer
+                # exception, so callers that cancel or time out STT cleanup observe the
+                # cancellation instead of a successful close.
+                if isinstance(cleanup_exception, asyncio.CancelledError) or (
+                    not primary_exception_active
+                ):
                     raise
                 try:
                     logger.warning(
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -12,7 +12,7 @@ from openai import AsyncOpenAI
 
 from ... import _debug
 from ...exceptions import AgentsException, UserError
-from ...logger import log_model_action_warning, logger
+from ...logger import logger
 from ...tracing import Span, SpanError, TranscriptionSpanData, transcription_span
 from ...util._error_tracing import get_trace_error
 from ..exceptions import STTWebsocketConnectionError
@@ -360,6 +360,7 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
     async def transcribe_turns(self) -> AsyncIterator[str]:
         self._connection_task = asyncio.create_task(self._process_websocket_connection())
 
+        primary_exception_active = False
         try:
             while True:
                 turn = await self._output_queue.get()
@@ -375,21 +376,21 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
                 finally:
                     self._output_queue.task_done()
         except BaseException:
+            primary_exception_active = True
+            raise
+        finally:
             try:
                 await self.close()
-            except BaseException as cleanup_error:
+            except BaseException:
+                if not primary_exception_active:
+                    raise
                 try:
-                    log_model_action_warning(
-                        logger,
-                        "STT session cleanup failed while preserving the consumer exception",
-                        cleanup_error,
+                    logger.warning(
+                        "STT session cleanup failed while preserving the consumer exception"
                     )
                 except Exception:
                     # Logging must not replace the original consumer exception.
                     pass
-            raise
-        else:
-            await self.close()
 
         self._check_errors()
         if self._stored_exception:
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -123,9 +123,6 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
         self._tracing_span.start()
 
     def _end_turn(self, _transcript: str) -> None:
-        if len(_transcript) < 1:
-            return
-
         if self._tracing_span:
             # Only encode audio if tracing is enabled AND buffer is not empty
             if self._trace_include_sensitive_audio_data and self._turn_audio_buffer:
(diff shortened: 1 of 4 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying lifecycle bug is real, and wrapping the iterator lifecycles in try/finally while draining owned tasks is the right direction.

Before we can merge this, please make the STT shutdown ordering ownership-safe. close() currently finishes the active span before awaiting WebSocket close; while that await is suspended, _handle_events() can process a queued transcription completion and start a new span that remains unfinished when the tasks are cancelled. Please stop and drain the session-owned work before the final span cleanup, and add a controlled interleaving test for this ordering.

Please also preserve the original consumer cancellation or exception if cleanup fails, with a negative-path regression test. Scope the guarantee to explicit aclose(), cancellation, and normal exhaustion; a bare async for break does not synchronously await generator close.

GautamSharma99:
Addressed the first review feedback in d3401727:

- STT shutdown now closes the transport, drains all session-owned tasks, and only then performs final span cleanup.
- Added a controlled interleaving regression where `_handle_events()` starts a replacement span while WebSocket close is suspended.
- Cleanup and cleanup-warning failures no longer replace an active consumer cancellation or exception, with negative-path and redaction coverage.
- The PR scope now explicitly covers normal exhaustion, cancellation, and explicit `aclose()`, and does not claim synchronous cleanup for a bare `async for` break.

The focused voice/STT suite passes (55 tests), and the full format, lint, typecheck, and test stack passes locally.

GautamSharma99:
@seratch 
Thanks for the detailed review. I’ve addressed the requested changes in d3401727:

- STT shutdown now closes the transport, stops and drains all session-owned tasks, and only then performs final span cleanup.
- Added a controlled interleaving test where _handle_events() starts a replacement span while WebSocket close is suspended.
 - Cleanup failures no longer replace the original consumer cancellation or exception.
 - Added negative-path coverage for cleanup and logging failures, including redacted and diagnostic logging modes.
 - Clarified that the cleanup guarantee covers normal exhaustion, cancellation, and explicit aclose(), but not synchronous cleanup after a bare async for break.

The focused voice/STT suite passes with 55 tests, and the full format, lint, typecheck, and test suite passes locally. Could you please take another look?

seratch (changes requested):
Thanks for the update. The STT shutdown ordering, replacement-span interleaving coverage, exception preservation, and clarified aclose() scope now address the previous review.

One lifecycle regression remains. StreamedAudioResult.stream() records session_ended only after yielding it. If the consumer immediately calls aclose(), GeneratorExit enters the exception cleanup path before saw_session_end is set, so the producer is cancelled instead of being allowed to finish normally. Please record terminal delivery before the yield, preserve graceful producer completion during finalization, and add a controlled regression test for this exact ordering.

Also, when cleanup fails while a primary consumer exception is active, please log only a static warning without attaching the cleanup exception. Its implicit `__context__` otherwise causes verbose logging to include the primary consumer traceback. Keep this as one simplified finalization path rather than adding another logging mode.

GautamSharma99:
Thanks for catching these remaining cases. Addressed in `a1f77082`:

- `session_ended` is now recorded before the generator yields it.
- Result finalization shields the producer after terminal delivery, then drains owned tasks from the same `finally` boundary.
- Added a controlled regression that consumes `session_ended`, immediately calls `aclose()`, verifies close waits for the producer, and verifies the producer completes without cancellation.
- Secondary cleanup failures now emit only a fixed `logger.warning(...)` message. No cleanup exception, arguments, `exc_info`, traceback, or implicit consumer context is attached, even when diagnostic model logging is enabled.
- Simplified both iterator implementations to one cleanup path while retaining primary cancellation and exception precedence.

The focused voice and STT suite passes with 52 tests, the sensitive-logging collector tests pass, and the full format, lint, typecheck, and test stack passes locally.

seratch:
Can you resolve the conflicts as well?

GautamSharma99:
Merged the current upstream `main` into this branch and resolved the test overlap with #4061 by retaining both its idle-dispatcher regression and this PR’s lifecycle coverage. Post-merge validation is clean: the focused voice/STT suite passes all 53 tests, and the full formatting, lint, type-check, and test stack passes.

seratch (changes requested):
Can you fix all the review comments by codex first?

GautamSharma99:
Thanks for the review — both Codex comments are fixed in 876e263a.

**Root cause.** Both cleanup paths (`StreamedAudioResult.stream()` and `OpenAISTTTranscriptionSession.transcribe_turns()`) run cleanup in a `finally` that swallows secondary exceptions to preserve the primary consumer exception. That also swallowed `asyncio.CancelledError`: when a caller cancels `aclose()` (or wraps it in `asyncio.wait_for`) while cleanup is still awaiting — shielding the producer after `session_ended`, or closing the STT websocket/draining tasks — the injected `GeneratorExit` had already set `primary_exception_active = True`, so the later `CancelledError` was treated as secondary and dropped. The caller then observed a successful close even though cleanup was cancelled.

**Fix.** The cleanup handler now always re-raises `CancelledError`, while still swallowing non-cancellation cleanup errors to preserve the primary exception:

```python
except BaseException as cleanup_exception:
    if isinstance(cleanup_exception, asyncio.CancelledError) or (
        not primary_exception_active
    ):
        raise
    # log + swallow non-cancellation cleanup error
```

**Tests.** Added a regression for each path that cancels cleanup with a primary exception active and asserts the cancellation propagates. Both fail before this change (the STT one surfaced `ValueError` instead of `CancelledError`) and pass now.

`make format`, `make lint`, `make typecheck`, and `make tests` all pass.

seratch:
Can you resolve the conflicts and remaining review comments?

AAliKKhan:
Hi @GautamSharma99 and @seratch,

I completed the remaining work for this PR:

1. **Codex P2 "Surface terminal producer errors on close"** is fixed. When a consumer closes the stream right after `session_ended`, the injected `GeneratorExit` no longer hides a producer/task error that surfaces during finalization: `stream()` now tracks the primary exception object and re-raises the cleanup exception when the primary is `GeneratorExit` on the terminal-delivery path, so `aclose()` reports the real producer outcome. A controlled regression covers this.

2. **Merged current `upstream/main`** and resolved the `src/agents/voice/result.py` conflict (kept this PR's `stream()` terminal-error handling; adopted upstream's `_cleanup_tasks()` finalization ordering). `tests/voice/test_pipeline.py` merged cleanly.

Verification: `tests/voice/test_pipeline.py` + `tests/voice/test_openai_stt.py` = 57 passed; ruff check/format and mypy (voice) clean. The new regression fails without the fix and passes with it.

I don't have push access to this fork, so the changes are in a PR against `fix/voice-stream-cleanup`:

**https://github.com/GautamSharma99/openai-agents-python/pull/1**

If you merge that into `fix/voice-stream-cleanup`, this PR will pick up the completed work and the remaining review items (terminal-error surfacing, cleanup ordering, conflicts) will be resolved here.

seratch:
Thanks again for working on this; it is merged as https://github.com/openai/openai-agents-python/pull/4131

PR #4065, opened 2026-07-31
fix(sandbox): harden default snapshot resolution

### LINKED ISSUE #4059: [Sandbox] Default snapshot resolver ignores explicit empty env and never invokes stale cleanup
### Describe the bug

The default local snapshot path/lifecycle implementation has three related gaps:

1. `default_local_snapshot_base_dir(env={...})` resolves the mapping with `env or os.environ`. Passing an explicit empty mapping therefore reads the host process environment instead of honoring the caller's isolated environment.
2. On non-macOS/non-Windows platforms, any non-empty `XDG_STATE_HOME` is accepted, including a relative path. The XDG Base Directory rules require relative values to be ignored. The current code can create SDK snapshot state relative to the process working directory.
3. `cleanup_stale_default_local_snapshots()` and the 30-day TTL are never used by production code. The only call site is its unit test. `resolve_default_local_snapshot_spec()` accepts a `now` argument but never uses it, and returns without invoking cleanup.

Source:

- Path resolution: https://github.com/openai/openai-agents-python/blob/000a96b602889b00f7cfaa210c41e1a74be65272/src/agents/sandbox/snapshot_defaults.py#L25-L50
- Cleanup helper: https://github.com/openai/openai-agents-python/blob/000a96b602889b00f7cfaa210c41e1a74be65272/src/agents/sandbox/snapshot_defaults.py#L53-L80
- Resolver with unused `now`: https://github.com/openai/openai-agents-python/blob/000a96b602889b00f7cfaa210c41e1a74be65272/src/agents/sandbox/snapshot_defaults.py#L83-L103

### Debug information

- Agents SDK version: v0.19.1
- Platforms: Linux/Unix path issue; cleanup issue is platform-independent
- Evidence type: static resolver/call-site inspection

### Repro steps

Environment isolation:

1. Set host `XDG_STATE_HOME=/host/state`.
2. Call `default_local_snapshot_base_dir(home=Path("/test/home"), env={}, platform="linux", os_name="posix")`.
3. The function returns under `/host/state` instead of `/test/home/.local/state`.

Relative XDG path:

1. Pass `env={"XDG_STATE_HOME": "relative-state"}`.
2. The returned base is relative and later `mkdir` creates it under the current working directory.

Stale cleanup:

1. Place an old `.tar` snapshot in the managed default directory.
2. Call `resolve_default_local_snapshot_spec(now=...)`.
3. The old file remains because the cleanup helper is not called and `now` is unused.

### Expected behavior

Explicit dependency-injection arguments should be honored even when empty. Relative XDG paths should be ignored in favor of the home fallback. The intended cleanup policy should either run at a documented safe lifecycle point or the dead TTL/`now` API should be removed until cleanup is intentionally supported.

### Why this matters

The current behavior makes tests and embedded callers depend on ambient host state, can write snapshot archives into unexpected working directories, and permits SDK-managed archives to accumulate indefinitely.

### Suggested scope

Use `env if env is not None else os.environ`, validate that XDG paths are absolute, and decide one explicit cleanup lifecycle. If resolver-time cleanup is intended, retain the best-effort error behavior and add tests proving only stale SDK-managed `.tar` files are removed.

I searched current issues and PRs for snapshot defaults, XDG state paths, explicit environment injection, and stale snapshot cleanup and found no active duplicate.

GautamSharma99:
### Summary

This pull request fixes two default local snapshot path-resolution bugs and removes an unsupported stale-deletion path.

#### Deterministic environment injection

`default_local_snapshot_base_dir()` previously selected `env or os.environ`. An explicitly supplied empty mapping therefore fell back to ambient process state, which made embedded callers and tests depend on host variables such as `XDG_STATE_HOME`.

The resolver now uses `os.environ` only when `env is None`, so `env={}` is honored as an isolated environment.

#### Absolute XDG state paths

The POSIX branch previously accepted any non-empty `XDG_STATE_HOME`, including relative paths. That could place SDK-managed archives below the process working directory. The resolver now accepts the override only when `Path(value).is_absolute()` and otherwise uses `<home>/.local/state`, matching the XDG Base Directory requirement and the existing Windows override hardening.

#### Snapshot retention ownership

The module contained a 30-day stale-archive helper and a resolver `now` parameter, but production never invoked the helper. Current behavior and regression coverage intentionally retain existing archives during default resolution because an old archive may still back paused or serialized resume state.

This PR takes the non-destructive lifecycle option described in #4059: it removes the unused TTL, deletion helper, `time` dependency, and dead `now` argument instead of making an unrelated path-resolution call delete durable resume artifacts. Automatic deletion should only be introduced later through an explicit artifact-cleanup API that can prove ownership and account for backend-specific state.

Explicit caller-provided snapshot specs, archive formats, directory permissions, macOS behavior, and Windows path resolution are unchanged.

### Test plan

- Added a regression proving `env={}` ignores an ambient host `XDG_STATE_HOME`.
- Added a regression proving relative `XDG_STATE_HOME` falls back to `<home>/.local/state`.
- Retained coverage that an absolute XDG path is accepted.
- Retained coverage that resolving the managed directory does not delete an existing stale archive.
- `uv run pytest -q tests/sandbox/test_snapshot_defaults.py tests/sandbox/test_session_manager.py` (`18 passed`)
  - `make format`
  - `make lint`
  - `make typecheck`
  - `make tests`

### Issue number

Closes #4059

### Checks

### DIFF
--- a/src/agents/sandbox/snapshot_defaults.py
+++ b/src/agents/sandbox/snapshot_defaults.py
@@ -2,13 +2,11 @@ from __future__ import annotations
 
 import os
 import sys
-import time
 from collections.abc import Mapping
 from pathlib import Path, PureWindowsPath
 
 from .snapshot import LocalSnapshotSpec
 
-_DEFAULT_LOCAL_SNAPSHOT_TTL_SECONDS = 60 * 60 * 24 * 30
 _DEFAULT_LOCAL_SNAPSHOT_SUBDIR = Path("openai-agents-python") / "sandbox" / "snapshots"
 
 
@@ -30,7 +28,7 @@ def default_local_snapshot_base_dir(
     os_name: str | None = None,
 ) -> Path:
     resolved_home = home or Path.home()
-    resolved_env = env or os.environ
+    resolved_env = os.environ if env is None else env
     resolved_platform = platform or sys.platform
     resolved_os_name = os_name or os.name
 
@@ -45,39 +43,14 @@ def default_local_snapshot_base_dir(
         base = env_base if env_base is not None else resolved_home / "AppData" / "Local"
     else:
         xdg_state_home = resolved_env.get("XDG_STATE_HOME")
-        base = Path(xdg_state_home) if xdg_state_home else resolved_home / ".local" / "state"
+        xdg_base = Path(xdg_state_home) if xdg_state_home else None
+        base = (
+            xdg_base
+            if xdg_base is not None and xdg_base.is_absolute()
+            else resolved_home / ".local" / "state"
+        )
 
     return base / _DEFAULT_LOCAL_SNAPSHOT_SUBDIR
-
-
-def cleanup_stale_default_local_snapshots(
-    base_path: Path,
-    *,
-    now: float | None = None,
-    max_age_seconds: int = _DEFAULT_LOCAL_SNAPSHOT_TTL_SECONDS,
-) -> None:
-    # This is intentionally limited to stale files in the SDK-managed default directory.
-    # We do not delete snapshots during normal session teardown because pause/resume may still
-    # need them. If we add explicit artifact cleanup later, it should be a separate opt-in path
-    # that can also account for backend-specific remote artifacts.
-    if max_age_seconds < 0 or not base_path.exists():
-        return
-
-    cutoff = (time.time() if now is None else now) - max_age_seconds
-    try:
-        candidates = list(base_path.glob("*.tar"))
-    except OSError:
-        return
-
-    for candidate in candidates:
-        try:
-            if not candidate.is_file():
-                continue
-            if candidate.stat().st_mtime >= cutoff:
-                continue
-            candidate.unlink(missing_ok=True)
-        except OSError:
-            continue
 
 
 def resolve_default_local_snapshot_spec(
@@ -86,7 +59,6 @@ def resolve_default_local_snapshot_spec(
     env: Mapping[str, str] | None = None,
     platform: str | None = None,
     os_name: str | None = None,
-    now: float | None = None,
 ) -> LocalSnapshotSpec:
     base_path = default_local_snapshot_base_dir(
         home=home,
@@ -94,6 +66,7 @@ def resolve_default_local_snapshot_spec(
         platform=platform,
         os_name=os_name,
     )
+    # Existing archives may still back paused or serialized resume state.
     base_path.mkdir(parents=True, exist_ok=True, mode=0o700)
     if (os_name or os.name) != "nt":
         try:
--- a/tests/sandbox/test_snapshot_defaults.py
+++ b/tests/sandbox/test_snapshot_defaults.py
@@ -3,10 +3,10 @@ from __future__ import annotations
 import os
 from pathlib import Path
 
+import pytest
+
 from agents.sandbox.snapshot import LocalSnapshotSpec
 from agents.sandbox.snapshot_defaults import (
-    _DEFAULT_LOCAL_SNAPSHOT_TTL_SECONDS,
-    cleanup_stale_default_local_snapshots,
     default_local_snapshot_base_dir,
     resolve_default_local_snapshot_spec,
 )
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
Please preserve the released snapshot-defaults API from v0.19.1: restore `cleanup_stale_default_local_snapshots()`, its TTL and test coverage, and the `now` keyword accepted by `resolve_default_local_snapshot_spec()`.

`agents.sandbox.snapshot_defaults` is included in the published API reference, so removing these names would break existing direct imports and keyword calls even though the runtime does not invoke cleanup automatically. The existing non-destructive behavior should remain: default snapshot resolution must not automatically delete archives.

Please narrow this PR to the explicit-empty-environment and relative-XDG fixes while preserving the released API. Once that is restored and CI is green, this should be ready for another review.

ErenAta16 (commented):
Verified all three points from #4059 against `main`, then re-ran the same probes on this branch. Every one changes, and the removal is safe to make.

**Before, on `main`:**

```
env={}    -> /host/leaked/openai-agents-python/sandbox/snapshots
env=None  -> /host/leaked/openai-agents-python/sandbox/snapshots      # identical
XDG_STATE_HOME="relative/state" -> relative/state/openai-agents-python/...   # not absolute
```

An explicit empty mapping and "no mapping given" were indistinguishable, so a caller isolating the environment silently got the host's `XDG_STATE_HOME`. `os.environ if env is None else env` separates them, which is the right distinction — `{}` is a statement about the environment, not the absence of one.

**After, on this branch:**

```
env={}                            -> /tmp/h/.local/state/...    # falls back to home, host ignored
env=None                          -> /host/leaked/...           # unchanged, as intended
XDG_STATE_HOME="relative/state"   -> /tmp/h/.local/state/...     # relative value ignored
XDG_STATE_HOME="/abs/state"       -> /abs/state/...              # absolute still honoured
```

The relative case now matches the XDG spec's "if not absolute, ignore it", and it lines up with what the Windows branch already did — `_first_absolute_windows_env_path` was checking `is_absolute()` all along, so this removes an inconsistency between the two platform paths rather than adding a new rule.

**On deleting the cleanup helper.** I checked whether anything outside its own test used it before agreeing this is the right call: `grep` across `src/`, `tests/`, `examples/` and `docs/` for `cleanup_stale_default_local_snapshots` and `_DEFAULT_LOCAL_SNAPSHOT_TTL_SECONDS` returns nothing on this branch, and `sandbox/__init__.py`'s `__all__` never re-exported it. The only production entry point into this module is `resolve_default_local_snapshot_spec`, called once from `runtime_session_manager.py:506` with no arguments. So dropping it isn't removing a public API, it's removing a function that existed only to be tested — and shipping a 30-day TTL that never runs is worse than not having one, because it reads like a guarantee.

`tests/sandbox/test_snapshot_defaults.py` is at 9 passed on the branch.

One thing worth deciding explicitly rather than by omission: with the helper gone, nothing ever prunes that directory, so `~/.local/state/openai-agents-python/sandbox/snapshots` grows without bound across sessions. That's the honest state either way — it was already true, since the TTL never ran — but the issue framed it as a lifecycle gap, and this PR closes the "dead code" half without the "who cleans up" half. Might be worth a follow-up issue so it isn't lost with the code.

ErenAta16 (commented):
Both behavioural fixes hold up. I ran `default_local_snapshot_base_dir` on `main` and on `ca8f749`, forcing the POSIX branch through the injected `platform`/`os_name` so `XDG_STATE_HOME` is actually consulted.

Ambient leakage, with `XDG_STATE_HOME=/ambient/from/process` set in the real process environment:

```
env={}   main:      /ambient/from/process/openai-agents-python/sandbox/snapshots
         this PR:   /tmp/home/.local/state/openai-agents-python/sandbox/snapshots
```

So an explicitly empty mapping stops reading host state, which is the point.

Relative values:

```
                     main                    this PR
'state'              used as base            ignored, falls back to $HOME/.local/state
'./state'            used as base            ignored
'../state'           used as base            ignored
```

Worth citing the spec here, because it makes this a conformance fix rather than a preference. The XDG Base Directory specification says:

> All paths set in these environment variables must be absolute. If an implementation encounters a relative path in any of these variables it should consider the path invalid and ignore it.

and gives `$HOME/.local/state` as the documented default for `XDG_STATE_HOME`, which is exactly the fallback taken. Both halves match.

On removing `cleanup_stale_default_local_snapshots`: I checked before assuming it was safe. It has no callers anywhere under `src/`, and it is not re-exported from any `__init__.py`, so the only reference was the test that goes with it. Removal does not change the importable surface for anyone going through the package.

**One thing I would change.** The new absoluteness check uses a bare `Path`, whose flavour follows the host rather than the injected platform:

```python
xdg_base = Path(xdg_state_home) if xdg_state_home else None
base = xdg_base if xdg_base is not None and xdg_base.is_absolute() else ...
```

`platform` and `os_name` exist as parameters specifically so this function can be exercised for one OS while running on another, and this check quietly opts out of that. On a Windows host:

```
Path('/abs/state').is_absolute()           False
PurePosixPath('/abs/state').is_absolute()  True
```

so a POSIX-absolute value injected alongside `platform="linux"` is treated as relative and discarded. This is not a production bug, on a real Linux box `Path` is `PosixPath` and the answer is right. It is a correctness gap in the thing the parameters were added for, and `tests.yml` does have a `tests-windows` job, so the suite genuinely runs on both.

`test_default_local_snapshot_base_dir_uses_xdg_state_home` currently passes on Windows only because it feeds `str(tmp_path / "state")`, which carries a drive letter and is therefore Windows-absolute by accident. Swap that for a literal `"/abs/state"` and it fails on the Windows job while passing on Linux.

`PurePosixPath(xdg_state_home).is_absolute()` fixes it and matches what the module already does one function up, where `_first_absolute_windows_env_path` deliberately pins `PureWindowsPath(value).is_absolute()` instead of trusting the host. Same reasoning, same shape.

**Smaller, take it or leave it.** The PR fixes `env or os.environ`, but the two lines beside it keep the pattern:

```python
resolved_platform = platform or sys.platform
resolved_os_name  = os_name or os.name
```

An empty string falls back to ambient state the same way `env={}` used to. Nobody realistically passes `platform=""`, so this is consistency rather than a defect, but leaving two of the three as they were makes the fixed one look arbitrary. `X if X is not None else ambient` across all three would read as one rule.

seratch:
Thanks again for sharing this patch. This one is not yet ready and we wanted to narrow the scope to fix, so https://github.com/openai/openai-agents-python/pull/4141 w/ your contribution credit is going to resolve the issue.

PR #4079, opened 2026-07-31
fix(voice): cancel producer tasks when the stream consumer stops early

ErenAta16:
Addresses the `StreamedAudioResult.stream()` half of #4051.

## What

`stream()` calls `_cleanup_tasks()` only after the yield loop has finished. When a consumer stops early — `break` out of the `async for`, or an explicit `aclose()` — `GeneratorExit` is delivered at the `yield` and unwinds straight past that block, so the producer tasks are never cancelled.

The pattern in isolation:

```
try/finally YOK    -> ['temizlik KOSMADI']
try/finally VAR    -> ['temizlik KOSTU']
```

The existing `try/except asyncio.CancelledError` only wraps `await self._queue.get()`, so it covers cancellation that lands while waiting on the queue, not finalization that lands at the yield.

## How

The body is wrapped in `try/finally` with `await self._cleanup_tasks()` in the `finally`. The `CancelledError` branch no longer calls `_cleanup_tasks()` itself — the `finally` covers that path now, so cleanup runs exactly once either way. No behaviour changes on the normal path: the `session_ended` shield and `_check_errors()` still run before cleanup, and `_stored_exception` is still raised after it.

## Tests

`tests/voice/test_result_cleanup.py`, two cases:

- `test_early_break_cancels_producer_tasks` — one long-running producer task registered, one event queued, consumer breaks after the first event and calls `aclose()`. Asserts the task ends up cancelled. **Fails on `main`, passes with this change.**
- `test_normal_exhaustion_still_cleans_up` — the pre-existing path, `None` ends the stream, cleanup still runs. Passes both ways; it's there so the normal path can't regress.

`tests/voice/` is at 64 passed with the change. `test_workflow.py` doesn't collect in my environment because `inline_snapshot` isn't installed, which is unrelated and identical on a clean checkout. `ruff check` and `ruff format --check` are clean on both files.

## Not covered here

#4051 also lists `OpenAIRealtimeTranscriptionSession.transcribe_turns()` in `openai_stt.py`, which has the same shape — cleanup after the loop, `yield turn` unguarded. I left it out to keep this reviewable; happy to do it in the same PR or a follow-up, whichever you prefer.

The issue's second point — that cleanup calls `task.cancel()` without awaiting the tasks — is already handled here for `result.py`: `_cleanup_tasks()` does `await asyncio.gather(*tasks, return_exceptions=True)` after cancelling.

### DIFF
--- a/src/agents/voice/result.py
+++ b/src/agents/voice/result.py
@@ -319,28 +319,31 @@ class StreamedAudioResult:
     async def stream(self) -> AsyncIterator[VoiceStreamEvent]:
         """Stream the events and audio data as they're generated."""
         saw_session_end = False
-        while True:
-            try:
-                event = await self._queue.get()
-            except asyncio.CancelledError:
-                await self._cleanup_tasks()
-                raise
-            if isinstance(event, VoiceStreamEventError):
-                self._stored_exception = event.error
-                log_model_and_tool_action_error(
-                    logger, "Error processing voice output", event.error
-                )
-                break
-            if event is None:
-                break
-            yield event
-            if event.type == "voice_stream_event_lifecycle" and event.event == "session_ended":
-                saw_session_end = True
-                break
-
-        # On the normal completion path, let the producer task finish gracefully so any active
-        # trace context can emit `trace_end` before we run cleanup.
+        # The whole body sits under `finally` so a consumer that stops early -- `break` out of
+        # the `async for`, or an explicit `aclose()` -- still gets the producer tasks
+        # cancelled. Without it `GeneratorExit` arrives at the `yield` below and unwinds past
+        # the cleanup, leaving those tasks running.
         try:
+            while True:
+                try:
+                    event = await self._queue.get()
+                except asyncio.CancelledError:
+                    raise
+                if isinstance(event, VoiceStreamEventError):
+                    self._stored_exception = event.error
+                    log_model_and_tool_action_error(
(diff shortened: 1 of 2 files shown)


seratch:
This PR overlaps #4060, which already covers both the result-stream and STT generator paths together with the required cleanup ownership and interleaving cases.

This patch also does not handle a bare `async for` break as claimed—the regression test explicitly calls `aclose()` afterward—and closing immediately after `session_ended` still reaches cleanup before `saw_session_end` is recorded, cancelling the producer instead of preserving session and trace finalization.

I’m going to close this PR in favor of #4060. We should resolve its remaining terminal-error handling and conflict work there rather than maintain two overlapping implementations.

ErenAta16:
You're right on all three, and thanks for the detail.

I should have found #4060 — it was open 13 hours before I opened this, by the same author as the issue, covering both files. That was a preflight failure on my part, not a judgement call.

On the third point I checked my own patch rather than take it on trust, and it is a regression: closing right after `session_ended` leaves `saw_session_end` unset, so the shield is skipped and the `finally` cancels the producer, where `main` at least leaves it alive to finalize.

```
main          saw_session_end=False  shield=SKIPPED  cleanup=did not run
this patch    saw_session_end=False  shield=SKIPPED  cleanup=producer cancelled
```

And the test does call `aclose()`, so it never demonstrated the bare `break` case the description claimed. Closing was the right call.

PR #4093, opened 2026-08-01
fix: preserve zero shell timeouts

### LINKED ISSUE #4092: Shell call normalization drops explicit zero timeouts
### Please read this first

- I read the Agents SDK docs.
- I searched open and closed issues and pull requests for this behavior and found no duplicate.

### Describe the bug

`coerce_shell_call()` selects the supported `timeout_ms`, `timeoutMs`, and `timeout` aliases with a truthiness chain. As a result, an explicit zero supplied through `timeout_ms` or `timeoutMs` is treated as absent and normalized to `None`.

This changes the caller-provided payload before it reaches a shell executor. It is also inconsistent with the adjacent `max_output_length` normalization, which falls back only when the value is `None`.

The bug exists on current main and the latest release tag, v0.19.2.

### Debug information

- Agents SDK version: v0.19.2 and main at 87425fa
- Python version: 3.12.12
- Operating system: macOS
- Model and model provider: provider-independent next-generation shell call normalization

```text
Expected timeout_ms: 0
Actual timeout_ms: None
```

### Repro steps

```python
from agents.run_internal import run_loop

for key in ("timeout_ms", "timeoutMs", "timeout"):
    call = {
        "call_id": "shell-zero-timeout",
        "action": {"commands": ["echo ok"], key: 0},
    }
    normalized = run_loop.coerce_shell_call(call)
    print(key, normalized.action.timeout_ms)
```

On v0.19.2 and current main this prints:

```text
timeout_ms None
timeoutMs None
timeout 0
```

### Expected behavior

All three supported aliases should preserve the explicit zero value. Alias fallback should occur only when the higher-priority field is absent (`None`), while keeping the existing alias precedence and all other conversion behavior unchanged.

### Verified patch

I have a minimal test-first patch ready at https://github.com/gambletan/openai-agents-python/tree/fix/preserve-zero-shell-timeout.

Verification completed with:

```text
```

The repository currently returns `must be a collaborator` when I attempt to open the pull request. I can submit it immediately if the intended workflow is to authorize or assign the issue first.

gambletan:
### Summary

Preserve an explicit zero timeout while normalizing next-generation shell call payloads. The previous truthiness-based alias fallback converted `timeout_ms=0` and `timeoutMs=0` to `None`, so executors could receive a default timeout instead of the caller-provided value.

This keeps the existing `timeout_ms`, `timeoutMs`, and `timeout` precedence while falling back only when a field is absent.

### Test plan

- `env UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh`
- Verified the regression test fails on the previous implementation for `timeout_ms` and `timeoutMs`, then passes with this change.

### Issue number

Fixes #4092

### Checks

- [x] I have added new tests, if relevant
- [x] I have run `.agents/skills/code-change-verification/scripts/run.sh`
- [x] I have confirmed all verification steps pass
- [x] If using Codex, I have run `/review` before submitting this PR

### DIFF
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -664,11 +664,11 @@ def coerce_shell_call(tool_call: Any) -> ShellCallData:
     if not commands:
         raise ModelBehaviorError("Shell call action must include at least one command.")
 
-    timeout_value = (
-        get_mapping_or_attr(action_payload, "timeout_ms")
-        or get_mapping_or_attr(action_payload, "timeoutMs")
-        or get_mapping_or_attr(action_payload, "timeout")
-    )
+    timeout_value = get_mapping_or_attr(action_payload, "timeout_ms")
+    if timeout_value is None:
+        timeout_value = get_mapping_or_attr(action_payload, "timeoutMs")
+    if timeout_value is None:
+        timeout_value = get_mapping_or_attr(action_payload, "timeout")
     timeout_ms = int(timeout_value) if isinstance(timeout_value, int | float) else None
 
     max_length_value = get_mapping_or_attr(action_payload, "max_output_length")
--- a/tests/test_shell_call_serialization.py
+++ b/tests/test_shell_call_serialization.py
@@ -23,6 +23,18 @@ def test_coerce_shell_call_reads_max_output_length() -> None:
     assert result.action.max_output_length == 512
 
 
+@pytest.mark.parametrize("timeout_key", ["timeout_ms", "timeoutMs", "timeout"])
+def test_coerce_shell_call_preserves_zero_timeout(timeout_key: str) -> None:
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for sharing this. The alias inconsistency is real, but this does not demonstrate a useful supported outcome for `timeout_ms=0`. `ShellTool` delegates execution to application-provided executors, and zero currently has no executor-independent meaning: our standard Python shell examples treat it as unset, while lower-level sandbox executors can turn it into an immediate timeout. The platform shell guide also does not define portable zero semantics. `None` already represents no explicit timeout, and positive values represent finite timeouts. Preserving zero would therefore lock in an ambiguous, executor-specific contract without enabling a useful workflow. 

I'm going to close this PR and the linked issue.

PR #4095, opened 2026-08-02
fix(models): merge dictionary fields in ModelSettings.resolve

hsusul:
### Summary

`ModelSettings.resolve()` overlays an override `ModelSettings` object or dictionary on top of a base `ModelSettings` instance. Previously, while `extra_args` was handled specially to merge dictionary keys, other dictionary configuration fields—`extra_headers`, `extra_query`, `extra_body`, and `metadata`—were replaced as atomic values.

As a result, when resolving model settings (for example, when `RunConfig.model_settings` overlays an `Agent.model_settings`), any dictionary keys specified on the base settings for `metadata`, `extra_headers`, `extra_query`, or `extra_body` were discarded if the override specified any value for that dictionary field.

This PR updates `ModelSettings.resolve()` so that all dictionary settings (`extra_args`, `extra_headers`, `extra_query`, `extra_body`, and `metadata`) are merged dictionary-wise, preserving non-conflicting base keys while allowing override keys to take precedence.

### Test plan

- Added focused unit tests in `tests/model_settings/test_serialization.py` verifying that dictionary fields (`metadata`, `extra_headers`, `extra_query`, `extra_body`) merge correctly during `ModelSettings.resolve()`.
- Verified the new test failed before the fix and passed after the fix.
- Ran formatting, linting, type checks, and tests:
  ```bash
  uv run ruff format --check .
  uv run ruff check .
  uv run pyright
  uv run pytest tests/model_settings/
  ```

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/model_settings.py
+++ b/src/agents/model_settings.py
@@ -249,16 +249,24 @@ class ModelSettings:
             and getattr(override, field.name, None) is not None
         }
 
-        # Handle extra_args merging specially - merge dictionaries instead of replacing.
-        if (override_fields is None or "extra_args" in override_fields) and (
-            self.extra_args is not None or override.extra_args is not None
+        # Handle dictionary fields merging - merge dictionaries instead of replacing.
+        for dict_field_name in (
+            "extra_args",
+            "extra_headers",
+            "extra_query",
+            "extra_body",
+            "metadata",
         ):
-            merged_args = {}
-            if self.extra_args:
-                merged_args.update(self.extra_args)
-            if override.extra_args:
-                merged_args.update(override.extra_args)
-            changes["extra_args"] = merged_args if merged_args else None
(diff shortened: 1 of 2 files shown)


seratch:
I traced `ModelSettings.resolve()` through the agent/run configuration path and compared it with the released `v0.19.2` behavior. The resolver overlays non-`None` fields atomically, with explicit merge semantics only for `extra_args` and structured retry settings. Applying dictionary merging to every mapping field would remove the existing ability for run-level settings to fully replace or clear inherited headers, query parameters, body fields, or metadata.

The desired merged value can already be composed explicitly by the caller, and the PR does not provide a concrete supported scenario where that approach is insufficient. I am going to close this PR for now. If such a scenario exists, please open an issue describing it, including the required replacement, removal, and empty-mapping semantics.

PR #4096, opened 2026-08-02
fix(items): support iterable structured tool outputs

hsusul:
### Summary

- Fix `ItemHelpers.tool_call_output_item()` stringifying generator-backed structured tool outputs even though its public contract accepts an iterable of structured items.
- Recognize non-string, non-mapping iterables after checking single structured models and dictionaries, preserving the existing behavior for lists, tuples, invalid mixtures, and empty collections.
- Add a regression test covering a one-shot generator with both Pydantic and dictionary structured output forms.

### Test plan

- `.venv/bin/python -m pytest tests/test_tool_output_conversion.py::test_tool_call_output_item_generator -q`
- `.venv/bin/python -m pytest tests/test_tool_output_conversion.py -q` (27 passed)
- `make format`, `make lint`, and `make typecheck`
- `make tests` (6,016 parallel tests and 45 serial tests passed, with 7 total skips)
- `.agents/skills/code-change-verification/scripts/run.sh` (all commands passed)

### Issue number

N/A — no linked issue.

### Checks

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import abc
 import json
 import weakref
-from collections.abc import Mapping
+from collections.abc import Iterable, Mapping
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Generic, Literal, TypeAlias, TypeVar, cast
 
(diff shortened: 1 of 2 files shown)


seratch:
I confirmed that the `ItemHelpers` docstring currently says "iterable", but the released `FunctionTool` contract and user-facing documentation specify structured output objects or a list of outputs. Returning `list(generator)` already supports the underlying use case, and this PR does not demonstrate a scenario where that existing path is insufficient.

Broadening conversion to every `Iterable` also changes the existing stringification fallback: a value such as `range(10**9)` is now fully traversed and retained before being rejected, which can hang or exhaust memory. I am going to close this PR. The narrower fix is to align the `ItemHelpers` docstring with the supported list/tuple behavior; we can revisit direct generator support if a concrete need is demonstrated.

PR #4097, opened 2026-08-02
fix(run): report tool guardrail results for streamed runs

hsusul:
### Summary

`RunResultStreaming.tool_input_guardrail_results` and `RunResultStreaming.tool_output_guardrail_results` were always empty. Streamed runs execute tool guardrails and build them into each `SingleStepResult`, but the streaming loop never carried them onto the run result, so anything that relies on these fields sees nothing in `Runner.run_streamed()` while `Runner.run()` reports them correctly.

**Affected component:** `src/agents/run_internal/run_loop.py` (`start_streaming`), streaming runner lifecycle.

**Minimal reproduction** (no API key, no live request — uses `tests/fake_model.py`):

```python
@tool_input_guardrail
def record_input(_data: ToolInputGuardrailData) -> ToolGuardrailFunctionOutput:
    return ToolGuardrailFunctionOutput.allow(output_info="input-checked")

@function_tool(name_override="guarded_tool", tool_input_guardrails=[record_input])
def guarded_tool() -> str:
    return "tool-result"

# model yields: [function_call guarded_tool] then [text "done"]
non_streamed = await Runner.run(agent, "hello")
streamed = Runner.run_streamed(agent, "hello")
async for _ in streamed.stream_events():
    pass

len(non_streamed.tool_input_guardrail_results)  # 1
len(streamed.tool_input_guardrail_results)      # 0  <-- bug
```

**Current behavior:** the guardrail runs (its rejection message reaches the model, and `new_items` shows the rejected/allowed tool output), but `streamed.tool_input_guardrail_results` and `streamed.tool_output_guardrail_results` stay `[]`. `RunResultStreaming.to_state()` therefore also persists empty `_tool_input_guardrail_results` / `_tool_output_guardrail_results`.

**Corrected behavior:** streamed runs report the same tool guardrail results as the equivalent non-streamed run.

**Root cause:** the non-streaming loop in `run.py` extends its run-wide accumulators from every turn (`tool_input_guardrail_results.extend(turn_result.tool_input_guardrail_results)`), and `RunResultStreaming` is constructed with these lists seeded only from a resumed `RunState`. `start_streaming` never extended them from `turn_result`, so for a fresh streamed run they stayed at their initial `[]` for the whole run. This is the streaming-side counterpart of the plumbing #3237 fixed inside `SingleStepResult`; the values arrive correctly at the streaming loop and are then dropped.

**Implementation:** `start_streaming` now extends both lists from each completed `turn_result`, via a small `_accumulate_tool_guardrail_results` helper called from the two places a `turn_result` is produced:

1. The regular turn loop, immediately next to the existing `streamed_result.raw_responses` accumulation and before the `next_step` branching — the same position and ordering the non-streaming loop uses, so interruption, handoff, run-again, and final-output turns are covered by one site.
2. The resume branch, which finalizes on its own when a resumed turn is terminal (for example an approved tool under `tool_use_behavior="stop_on_first_tool"`, or a resumed turn that hands off) and so never reaches site 1. This site skips `NextStepRunAgain`, matching the non-streaming resume path, so a pre-approval guardrail that re-runs for the same tool call is not counted twice.

Rebinding (rather than `list.extend`) matches how `input_guardrail_results` / `output_guardrail_results` are updated elsewhere in this module and avoids mutating a caller-visible list in place.

**Why this is minimal:** ~25 added lines, no behavior change outside the streaming loop. No public API, exception type, hook contract, event ordering, usage accounting, or session behavior changes; nothing is added to the non-streaming path.

Parity was verified directly for all five resumed outcomes — terminal final output, run-again, handoff, a second interruption, and pre-approval run-again — and streamed now matches non-streamed exactly in each.

### Test plan

Eight tests added to `tests/test_agent_runner_streamed.py`, all using `FakeModel` — no API key, no network, no sleeps.

Seven fail on `upstream/main` (`a134f3a`) and pass with the fix:

- `test_streamed_run_reports_tool_guardrail_results` — the demonstrated failure; input and output guardrail results are reported.
- `test_streamed_tool_guardrail_results_match_non_streamed` — two guarded tool calls across two turns; streamed counts equal non-streamed counts (no drops, no duplicates).
- `test_streamed_tool_guardrail_results_survive_handoff` — guarded tool call and a handoff in the same response, covering the `execute_handoffs` path from #3237 in streaming mode.
- `test_streamed_interruption_reports_tool_guardrail_results` — a turn that ends in an approval interruption still reports the guardrails that ran for the tool executed alongside it.
- `test_streamed_tool_guardrail_results_persist_into_run_state` — the results survive `RunResultStreaming.to_state()`.
- `test_streamed_resume_terminal_turn_reports_tool_guardrail_results` — resume under `tool_use_behavior="stop_on_first_tool"`, where the approved tool produces the final output inside the resumed turn and the run finalizes from the resume branch.
- `test_streamed_resume_handoff_turn_reports_tool_guardrail_results` — resume where the approved turn also hands off.

One test passes both before and after; it is a parity guard, not a regression test:

- `test_streamed_resume_tool_guardrail_results_match_non_streamed` — pins streamed and non-streamed *resumed* turns to the same value, so this change provably does not alter resumed-turn accounting. It asserts equality rather than a specific count deliberately (see non-goals).

Commands run from the repository root at commit `b00e159`, all passing:

```
bash .agents/skills/code-change-verification/scripts/run.sh   # all commands passed
uv run ruff format --check                                    # 844 files already formatted
uv run ruff check                                             # All checks passed!
make typecheck                                                # mypy: no issues in 835 source files; pyright: 0 errors
uv run pytest -n auto --dist loadfile -m "not serial" -q      # 6023 passed, 3 skipped
uv run pytest -m serial -q                                    # 45 passed, 4 skipped
git diff --check                                              # clean
```

Red/green evidence: with `src/agents/run_internal/run_loop.py` reverted to `upstream/main` and the tests kept, `uv run pytest tests/test_agent_runner_streamed.py -k "tool_guardrail or interruption_reports"` reports `7 failed, 1 passed`; with the fix applied it reports `8 passed`. The focused selection was also run five times consecutively and under `-W error::RuntimeWarning` with identical results (no pending-task or unclosed-resource warnings).

Environment: macOS 15.7.3 (arm64), Python 3.12.13 via `uv`, base commit `a134f3a29869f3aa0e10e37447897046d18c2d23`.

**Not run:** `make coverage`, `make build-docs`, and the `integration-tests-*` targets. Coverage and docs are unaffected (no docs or public-surface change), and the integration profiles require credentials and live services.

**Non-goals / limitations**

- `NextStepRunAgain` accounting on resume is deliberately preserved, not changed. The non-streaming loop skips accumulation for a resumed turn that loops back to the model, which is what keeps `test_pre_approval_tool_input_guardrails_rerun_after_resume` at one result rather than re-counting the same tool call. This PR matches that rule rather than relitigating it, and `test_streamed_resume_tool_guardrail_results_match_non_streamed` pins the two modes together so the decision stays in one place.
- Partial results are still not reported when a tool guardrail returns `raise_exception`. This is pre-existing and symmetric: `RunErrorDetails` has no `tool_input_guardrail_results` / `tool_output_guardrail_results` fields, so the non-streaming path does not surface them on a tripwire either, and the triggering result is already reachable through the raised `ToolInputGuardrailTripwireTriggered` / `ToolOutputGuardrailTripwireTriggered`. Surfacing the turn's partial results would mean either threading a mutable sink from `start_streaming` through `run_single_turn_streamed` → `get_single_step_result_from_response` → `execute_tools_and_side_effects` → `execute_tool_runs` → `execute_function_tool_calls`, or adding public fields to `RunErrorDetails`. Both are cross-module changes with their own design decision, and doing it for streaming alone would create exactly the asymmetry this PR removes. Happy to do it as a follow-up if maintainers want that reporting.
- No change to guardrail execution, ordering, or `reject_content` / `raise_exception` behavior — only to reporting.

### Issue number

None. `AGENTS.md` asks for an issue number "if applicable" and does not require filing one before a small fix, so no speculative issue was created.

### Checks

- [ ] If using Codex, I've run `/review` before submitting this PR — not applicable, this was not authored with Codex.

### DIFF
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -1140,6 +1140,16 @@ async def start_streaming(
                 streamed_result.raw_responses = streamed_result.raw_responses + [
                     turn_result.model_response
                 ]
+                # Mirror the non-streaming loop, which extends the run-wide tool guardrail lists
+                # from every turn result so the run result reports the guardrails that ran.
+                streamed_result.tool_input_guardrail_results = (
+                    streamed_result.tool_input_guardrail_results
+                    + turn_result.tool_input_guardrail_results
+                )
+                streamed_result.tool_output_guardrail_results = (
+                    streamed_result.tool_output_guardrail_results
+                    + turn_result.tool_output_guardrail_results
+                )
                 input_before_turn_rewrite = streamed_result.input
                 streamed_result.input = turn_result.original_input
                 if isinstance(turn_result.next_step, NextStepHandoff):
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -449,6 +449,23 @@ async def _finalize_streamed_final_output(
     streamed_result._event_queue.put_nowait(QueueCompleteSentinel())
 
 
+def _accumulate_tool_guardrail_results(
+    streamed_result: RunResultStreaming,
+    turn_result: SingleStepResult,
+) -> None:
+    """Carry a turn's tool guardrail results onto the streamed result.
+
+    The non-streaming loop extends its run-wide lists from every turn result, so the streaming
+    loop has to do the same for `RunResultStreaming` to report the guardrails that ran.
+    """
+    streamed_result.tool_input_guardrail_results = (
+        streamed_result.tool_input_guardrail_results + turn_result.tool_input_guardrail_results
+    )
+    streamed_result.tool_output_guardrail_results = (
+        streamed_result.tool_output_guardrail_results + turn_result.tool_output_guardrail_results
+    )
+
+
 async def _finalize_streamed_interruption(
     *,
     streamed_result: RunResultStreaming,
@@ -858,6 +875,12 @@ async def start_streaming(
                         run_config.model_settings
                     ).store
 
+                    # The non-streaming resume path extends its run-wide lists before finalizing
+                    # but skips a resumed turn that loops back to the model, so a guardrail that
+                    # re-runs for the same tool call on resume is not counted twice.
+                    if not isinstance(turn_result.next_step, NextStepRunAgain):
+                        _accumulate_tool_guardrail_results(streamed_result, turn_result)
+
                     if isinstance(turn_result.next_step, NextStepInterruption):
                         await _finalize_streamed_interruption(
                             streamed_result=streamed_result,
@@ -1140,16 +1163,7 @@ async def start_streaming(
                 streamed_result.raw_responses = streamed_result.raw_responses + [
                     turn_result.model_response
                 ]
-                # Mirror the non-streaming loop, which extends the run-wide tool guardrail lists
-                # from every turn result so the run result reports the guardrails that ran.
-                streamed_result.tool_input_guardrail_results = (
-                    streamed_result.tool_input_guardrail_results
-                    + turn_result.tool_input_guardrail_results
-                )
-                streamed_result.tool_output_guardrail_results = (
-                    streamed_result.tool_output_guardrail_results
-                    + turn_result.tool_output_guardrail_results
-                )
+                _accumulate_tool_guardrail_results(streamed_result, turn_result)
                 input_before_turn_rewrite = streamed_result.input
                 streamed_result.input = turn_result.original_input
                 if isinstance(turn_result.next_step, NextStepHandoff):
--- a/tests/test_agent_runner_streamed.py
+++ b/tests/test_agent_runner_streamed.py
@@ -36,6 +36,9 @@ from agents import (
     OutputGuardrailTripwireTriggered,
     RunContextWrapper,
     Runner,
+    ToolGuardrailFunctionOutput,
+    ToolInputGuardrailData,
+    ToolOutputGuardrailData,
     UserError,
     function_tool,
     handoff,
@@ -54,7 +57,8 @@ from agents.run import RunConfig
 from agents.run_internal import run_loop
 from agents.run_internal.run_loop import QueueCompleteSentinel
 from agents.stream_events import AgentUpdatedStreamEvent, RawResponsesStreamEvent, StreamEvent
-from agents.tool import Tool
+from agents.tool import FunctionTool, Tool
+from agents.tool_guardrails import tool_input_guardrail, tool_output_guardrail
 from agents.usage import Usage
 
 from .fake_model import FakeModel, get_response_obj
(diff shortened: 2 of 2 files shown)


seratch (approved):
Once you resolve all the feedback from codex, we can merge this improvement

seratch (changes requested):
Once you resolve all the feedback from codex, we can merge this improvement

hsusul:
Thanks — addressed the first finding in `b00e159`; pushing back on the second with evidence.

### 1. Accumulate guardrails from resumed terminal turns — fixed

Confirmed and reproduced. With `tool_use_behavior="stop_on_first_tool"`, the approved tool produces the final output inside `resolve_interrupted_turn`, the run finalizes from the resume branch, and the accumulation site in the regular turn loop is never reached:

```
nonstrm: final='approved-result' in=1 out=1
stream : final='approved-result' in=0 out=0
```

`start_streaming` now accumulates in the resume branch too, guarded by `not isinstance(turn_result.next_step, NextStepRunAgain)` so the intentional run-again behavior is preserved. I checked all five resumed outcomes rather than just the reported one, and streamed now matches non-streamed exactly in each:

| resumed `next_step` | non-streamed | streamed |
|---|---|---|
| terminal final output (`stop_on_first_tool`) | 1 / 1 | 1 / 1 |
| run again | 0 / 0 | 0 / 0 |
| handoff | 1 / 1 | 1 / 1 |
| second interruption | 1 / 1 | 1 / 1 |
| pre-approval + run again | 1 / 0 | 1 / 0 |

Two regression tests added — `test_streamed_resume_terminal_turn_reports_tool_guardrail_results` and `test_streamed_resume_handoff_turn_reports_tool_guardrail_results` — both of which fail on the previous commit and on `main`, and both assert equality with the non-streamed run rather than only a literal count.

### 2. Preserve results when a tool guardrail raises — not in this PR

The behavior report is accurate, but I don't think this belongs here, for three reasons:

**It is pre-existing, not introduced or worsened by this PR.** Before this change the lists were empty for *every* streamed run; this PR fixes the non-exceptional paths and leaves the tripwire path exactly as it was on `main`.

**There is no non-streaming counterpart to match.** `RunErrorDetails` (`src/agents/exceptions.py:31`) carries `input_guardrail_results` and `output_guardrail_results` but has no tool-guardrail fields, so `Runner.run()` does not surface these on a tripwire either — verified:

```
input-raise nonstrm: err=ToolInputGuardrailTripwireTriggered run_data=yes has_tool_gr_attr=False
```

Implementing capture for streaming alone would leave streamed runs reporting strictly more than non-streamed ones — the same class of divergence this PR exists to remove.

**The minimal implementation is not local.** The results live on `_FunctionToolBatchExecutor` and are discarded when the tripwire unwinds, so surfacing them means either threading a mutable sink from `start_streaming` through `run_single_turn_streamed` → `get_single_step_result_from_response` → `execute_tools_and_side_effects` → `execute_tool_runs` → `execute_function_tool_calls`, or adding public fields to `RunErrorDetails` and populating them in both loops. Either is a cross-module change that also decides whether the *triggering* result should be included alongside the earlier ones — and the triggering result is already reachable today via `ToolInputGuardrailTripwireTriggered.guardrail` / `.output`.

That reads to me like AGENTS.md's complexity-reset checkpoint rather than another condition to patch onto this abstraction, so I've documented it under non-goals in the PR description. Happy to do it as a follow-up — including the `RunErrorDetails` side so both execution modes gain it together — if maintainers want that reporting.

### Verification at `b00e159`

`.agents/skills/code-change-verification/scripts/run.sh` passes; `ruff format --check` and `ruff check` clean; `make typecheck` clean (mypy 835 files, pyright 0 errors); `pytest -n auto -m "not serial"` 6023 passed / 3 skipped; `pytest -m serial` 45 passed / 4 skipped; `git diff --check` clean. The focused selection goes 7 failed → 8 passed against `main` and was run 5× plus under `-W error::RuntimeWarning` with no flakiness and no pending-task or unclosed-resource warnings.

seratch (approved):

PR #4098, opened 2026-08-02
fix(memory): prevent writes to cleared conversations

hsusul:
### Summary

`OpenAIConversationsSession.clear_session()` kept its current conversation ID visible while awaiting `conversations.delete`. A concurrent `add_items()` could therefore write to that same conversation during deletion, return successfully, and have its item disappear with the deleted conversation.

Clear now detaches the old ID while holding the existing session-ID lock. Operations that start during deletion wait for the clear to finish and lazily create a replacement conversation. If deletion fails or is cancelled, the prior ID is restored so the existing failure behavior remains usable.

The regression uses events and mocked Conversations methods: deletion is paused, a write starts, and the test proves no item request is issued until deletion is released; the write then targets the replacement conversation. It fails on unmodified `main` because the item request starts immediately against the old conversation.

### Test plan

- Focused regression before the production change: failed at `assert not item_write_started.is_set()`.
- `pytest -q tests/memory/test_openai_conversations_session.py` (`30 passed` before the cancellation boundary test was added; the final full suite includes both tests).
- `pytest -q tests/memory/test_session_limit.py tests/memory/test_session.py` (`52 passed`).
- Repeated the clear/write and cancellation regressions five times (`2 passed` on each run).
- Ran `.agents/skills/code-change-verification/scripts/run.sh`; format, lint, typecheck, and the complete test suite passed.

### Issue number

N/A — found through code-path review; no open or closed issue, pull request, or upstream branch covers concurrent writes during `OpenAIConversationsSession.clear_session()`.

### Checks

`/review` was not invoked because this task prohibited additional live model calls; the final two-file diff was manually reviewed after all verification gates passed.

### Scope and risk

- No public API, session method signature, or dependency changes.
- Lazy initialization and clear-on-uninitialized behavior are preserved.
- Delete exceptions and cancellation restore the original conversation ID.

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -129,8 +129,15 @@ class OpenAIConversationsSession(SessionABC):
         return items[0]
 
     async def clear_session(self) -> None:
-        session_id = await self._get_session_id()
-        await self._openai_client.conversations.delete(
-            conversation_id=session_id,
-        )
-        await self._clear_session_id()
+        async with self._session_id_lock:
+            if self._session_id is None:
+                self._session_id = await start_openai_conversations_session(self._openai_client)
+            session_id = self._session_id
+            self._session_id = None
+            try:
+                await self._openai_client.conversations.delete(
+                    conversation_id=session_id,
+                )
+            except BaseException:
+                self._session_id = session_id
+                raise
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for investigating this. I can confirm that `main` leaves the old conversation ID visible while `conversations.delete` is awaiting, but the current regression only proves mocked call ordering. It does not demonstrate the claimed real API outcome, and the Conversations API reference states that deleting a conversation does not delete its items. Please provide a real API reproduction showing that `items.create` can succeed during deletion and that the created item then becomes unavailable to the session.

The current locking boundary is also incomplete: a write that resolves the old ID before `clear_session()` acquires the lock can still issue its request during deletion. I do not think we should merge a partial concurrency contract, so I suggest closing this PR for now. If the real-path behavior is demonstrated, we should define one ordering rule across whole session operations and cover both completion orders before choosing the implementation.

PR #4113, opened 2026-08-02
fix(tool): validate explicit tool names and enforce the 64-char limit on derived ones

### LINKED ISSUE #4112: Explicit tool names (`name_override` / `FunctionTool(name=...)`) are never validated, so names the API rejects pass through silently
### Describe the bug

The OpenAI function-calling API constrains tool names to `^[a-zA-Z0-9_-]{1,64}$`. The Agents SDK enforces this on names it derives itself, but performs no validation at all on names the user supplies explicitly.

Two paths, two behaviours:

- **Derived names are sanitized.** `Agent.as_tool()` and `handoff()` route the agent name through `transform_string_function_style()`, which replaces every character outside `[a-zA-Z0-9_]` with `_` and emits a warning.
- **Explicit names are passed through verbatim.** `@function_tool(name_override=...)` and `FunctionTool(name=...)` assign the string directly. Nothing checks the character set, and nothing checks the length.

So this is accepted without error or warning and is sent straight to the provider:

```python
@function_tool(name_override="my tool!")   # space and '!' -- rejected by the API
def f(a: str) -> str: ...
```

as is a non-ASCII name (`"выборка"`) and a 100-character name. The result is an opaque HTTP 400 at run time, attributed to the request rather than to the tool definition, with nothing in the message tying it back to which of the agent's tools is at fault.

The inconsistency is the core of the report: the SDK already knows the constraint and already applies it in two other places.

- Character sanitization exists for agent-derived names (`transform_string_function_style`), whose warning even tells users to "use only letters, digits, and underscores".
- Length capping exists for MCP-derived names. `agents/mcp/util.py` defines `_MCP_FUNCTION_TOOL_NAME_MAX_LENGTH = 64` and `_shorten_tool_name()` truncates and hash-suffixes any name that exceeds it.

Neither is applied to the one path where the user typed the name by hand and is most likely to get it wrong.

Length in particular slips through on *every* path except MCP. Because `transform_string_function_style` only substitutes characters, a long agent name produces an equally long tool name:

```python
Agent(name="A" * 100).as_tool(tool_name=None, tool_description="d").name
# 'aaaa...' (100 chars) -- over the 64-char API limit, no warning
```

(#1776 previously asked for maximum-length validation on `Agent.name`; on current `main` a 100-character agent name is still accepted and still yields an over-length derived tool name. This report is primarily about the missing *character* validation on explicit names, which appears to be unreported.)

### Debug information

- Python version: 3.13.6
- Model and model provider: Reproduced offline — the invalid name is accepted by the SDK with no network call. The resulting 400 comes from OpenAI (Responses and Chat Completions) and from any provider enforcing the documented name pattern.

No SDK-side exception or warning is produced — that is the bug. Output of the repro below:

```text
name_override with spaces/punctuation
  OK name: 'my tool!'
name_override with unicode
  OK name: 'выборка'
name_override 100 chars
  OK name: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
agent name -> as_tool 100 chars
  OK name: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'

agent name with dots -> as_tool
  OK name: 'team_billing'
```

The four explicit-name cases above produce no diagnostic of any kind. The derived path, by contrast, emits a log record through `agents.logger` before sanitizing:

```text
Tool name 'team.billing' contains invalid characters for function calling and has been
transformed to 'team_billing'. Please use only letters, digits, and underscores to avoid
potential naming conflicts.
```

Note that even on the derived path this is only a `logger.warning`, not a `UserError` or a Python `warnings` warning, and it is never emitted for over-length names.

### Repro steps

Minimal, self-contained, no API key required:

```python
import warnings

from agents import Agent, function_tool


def make(name: str) -> str:
    @function_tool(name_override=name)
    def f(a: str) -> str:
        """F."""
        return a

    return f.name


with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")

    print(repr(make("my tool!")))   # 'my tool!'  -- space and '!' survive
    print(repr(make("выборка")))    # 'выборка'   -- non-ASCII survives
    print(len(make("x" * 100)))     # 100         -- over the 64-char limit

    print("warnings emitted:", len(caught))  # 0
    # ...and nothing is logged through `agents.logger` for these either.

# By contrast, the derived path sanitizes and warns:
print(Agent(name="team.billing").as_tool(tool_name=None, tool_description="d").name)
# 'team_billing'  (+ warning)

# ...but the derived path does not enforce length:
print(len(Agent(name="A" * 100).as_tool(tool_name=None, tool_description="d").name))
# 100
```

Attaching any of these tools to an `Agent` and calling a real OpenAI model fails with a 400 complaining about the tool name, with no indication of which tool caused it.

### Expected behavior

Explicit tool names should be validated where they are set, so the failure is attributed to the tool definition instead of to an opaque provider response.

Concretely, `@function_tool(name_override=...)` and `FunctionTool.__post_init__` should raise a `UserError` when the name does not match the documented function-name pattern:

```
UserError: Invalid function tool name 'my tool!'. Tool names must match ^[a-zA-Z0-9_-]{1,64}$.
Use only letters, digits, underscores, and hyphens.
```

This mirrors the guidance the SDK already prints for derived names, and matches how the SDK already raises `UserError` for other malformed tool configurations (duplicate MCP tool names, duplicate Codex tool names, misplaced `RunContextWrapper` parameters).

Separately, the 64-character limit should be enforced on the derived path too — either by raising, or by reusing the truncate-and-hash approach already implemented for MCP tools in `MCPUtil._shorten_tool_name()`.

rajarshidattapy:
## Description

Closes #4112

The OpenAI function-calling API constrains tool names to `^[a-zA-Z0-9_-]{1,64}$`.
The SDK sanitized names it derived itself but never checked names the user typed,
so `@function_tool(name_override="my tool!")`, `FunctionTool(name=...)`, and
`handoff(tool_name_override=...)` passed straight through to an opaque HTTP 400
that named the request rather than the offending tool. The 64-char limit was
enforced only for MCP tools, so a long agent name produced an equally long
derived name with no warning.

This applies the split the issue asks for: **explicit names raise, derived names
are sanitized.**

## Explicit names raise `UserError`

`FunctionTool.__post_init__` and `Handoff.__post_init__` now validate at
construction, so every path that accepts a caller-supplied name is covered —
`@function_tool(name_override=...)`, `FunctionTool(name=...)`,
`handoff(tool_name_override=...)`, `realtime_handoff(...)`, and direct dataclass
construction:

UserError: Invalid function tool name 'my tool!'. Tool names must contain only
ASCII letters, digits, underscores, hyphens, or dots, and be between 1 and 64
characters.

Dots are permitted because they are the wire shape for a qualified top-level tool
(`tool_qualified_name`).

## Derived names are sanitized, not rejected

`transform_string_function_style()` now caps length using the same
truncate-and-hash shape `MCPUtil._shorten_tool_name()` already uses, so
`Agent.as_tool()` and `handoff()` both get it from the one shared function.
The suffix is a deterministic hash of the original name, so names sharing a
64-char prefix stay distinct instead of silently colliding:

```python
Agent(name="A" * 100).as_tool(tool_name=None, tool_description="d").name
# 'aaaa…aaa_7a4a9ae5'  (64 chars, warns)

function_schema() also sanitizes a name derived from func.__name__ or the
docstring when it is not API-safe — <lambda>, a unicode __name__, or one over
64 characters. Without this, adding validation at __post_init__ would have
started rejecting function_tool(lambda: ...), which is used across the test
suite. Valid derived names are untouched, so def MyTool still yields MyTool
with no case folding.

Tests

- tests/test_tool_identity.py — explicit names rejected (spaces/punctuation,
non-ASCII, over-length) via both name_override and the dataclass; valid names
including hyphens, dots, and exactly 64 chars pass; derived lambda/unicode/long
names sanitized; valid derived names unchanged.
- tests/test_transforms.py — shortening at the boundary, determinism, and that
two names sharing a 64-char prefix do not collide.
- tests/test_handoff_tool.py — invalid tool_name_override raises; a 100-char
agent name yields a 64-char transfer_to_… plus the shortening warning.

### DIFF
--- a/src/agents/_tool_identity.py
+++ b/src/agents/_tool_identity.py
@@ -18,16 +18,41 @@ FunctionToolLookupKey = (
 NamedToolLookupKey = FunctionToolLookupKey | str
 
 
+MAX_FUNCTION_TOOL_NAME_LENGTH = 64
+
+
+def is_api_safe_function_tool_name(name: str, *, allow_dots: bool = False) -> bool:
+    """Return True when a tool name matches the model API's function-name constraints."""
+    allowed_punctuation = {"_", "-", "."} if allow_dots else {"_", "-"}
+    return 1 <= len(name) <= MAX_FUNCTION_TOOL_NAME_LENGTH and all(
+        char.isascii() and (char.isalnum() or char in allowed_punctuation) for char in name
+    )
+
+
 def validate_function_tool_fallback_name(name: str) -> str:
     """Return an API-safe generated tool name or require an explicit override."""
-    if 1 <= len(name) <= 64 and all(
-        char.isascii() and (char.isalnum() or char in {"_", "-"}) for char in name
-    ):
+    if is_api_safe_function_tool_name(name):
         return name
     raise UserError(
         f"Cannot derive a function tool name from callable class {name!r}. Generated names must "
         "contain only ASCII letters, digits, underscores, or hyphens and be at most 64 "
         "characters. Pass name_override to function_tool()."
+    )
+
+
+def validate_function_tool_name(name: str) -> None:
+    """Reject a configured tool name the model API will not accept.
+
+    Dots are allowed here but not in `validate_function_tool_fallback_name`: a dotted name is
+    the supported wire shape for a qualified top-level tool (see `tool_qualified_name`), so it
+    is only ever legitimate when a caller asked for it explicitly.
+    """
+    if is_api_safe_function_tool_name(name, allow_dots=True):
+        return
+    raise UserError(
+        f"Invalid function tool name {name!r}. Tool names must contain only ASCII letters, "
+        "digits, underscores, hyphens, or dots, and be between 1 and "
+        f"{MAX_FUNCTION_TOOL_NAME_LENGTH} characters."
     )
 
 
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -13,10 +13,12 @@ from griffe import Docstring, DocstringSectionKind  # type: ignore[import-untype
 from pydantic import BaseModel, Field, create_model
 from pydantic.fields import FieldInfo
 
+from ._tool_identity import is_api_safe_function_tool_name
 from .exceptions import UserError
 from .run_context import RunContextWrapper
 from .strict_schema import ensure_strict_json_schema
 from .tool_context import ToolContext
+from .util._transforms import transform_string_function_style
 
 
 @dataclass
@@ -344,6 +346,12 @@ def function_schema(
 
     # Ensure name_override takes precedence even if docstring info is disabled.
     func_name = name_override or (doc_info.name if doc_info else func.__name__)
+
+    if name_override is None and not is_api_safe_function_tool_name(func_name):
+        # A name we derived ourselves (`<lambda>`, a unicode or over-long `__name__`) gets
+        # sanitized like other derived names. An explicit name_override is rejected instead,
+        # in FunctionTool.__post_init__, since the caller can fix it at the source.
+        func_name = transform_string_function_style(func_name)
 
     # 2. Inspect function signature and get type hints
     sig = inspect.signature(func)
--- a/src/agents/handoffs/__init__.py
+++ b/src/agents/handoffs/__init__.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any, Generic, TypeAlias, cast, overload
 from pydantic import TypeAdapter
 from typing_extensions import TypeVar
 
+from .._tool_identity import validate_function_tool_name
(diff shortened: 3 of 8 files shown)


seratch:
We don't plan to add client-side validation for this so far.

PR #4115, opened 2026-08-02
fix(strict_schema): treat typeless and union-typed nodes as objects

### LINKED ISSUE #4114: Strict schema conversion misses additionalProperties on some object schemas
### Describe the bug

`agents.strict_schema.ensure_strict_json_schema()` decides whether to inject `additionalProperties: false` by comparing the `type` keyword against the exact string `"object"`:

```python
# src/agents/strict_schema.py
typ = json_schema.get("type")
if typ == "object" and "additionalProperties" not in json_schema:
    json_schema["additionalProperties"] = False
```

The very next block, however, treats *any* schema node that carries a `properties` mapping as an object and rewrites `required`:

```python
properties = json_schema.get("properties")
if is_dict(properties):
    json_schema["required"] = list(properties.keys())
    ...
```

The two checks disagree. When a node has `properties` but no literal `"type": "object"`, the function still applies the strict-mode `required` rewrite but **skips** `additionalProperties: false`. The resulting schema is then handed to the model provider with `strict: true`, and OpenAI rejects it:

```
400 - Invalid schema: 'additionalProperties' is required to be supplied and to be false.
```

Both shapes are ordinary, valid JSON Schema and are common in the wild:

1. **Typeless object schemas** — `{"properties": {...}}` with no `type` keyword at all. JSON Schema does not require `type`, and hand-written tool schemas and many MCP servers omit it.
2. **Union-typed / nullable object schemas** — `{"type": ["object", "null"], "properties": {...}}`. This is the standard OpenAPI 3.1 and JSON Schema 2020-12 way to express a nullable object, so it appears frequently in schemas generated from OpenAPI documents.

The gap is not limited to the root node. It reproduces identically for nested `properties` values, for `anyOf` branches, and for `array` `items`, so a single typeless sub-object anywhere in an otherwise valid schema is enough to make the whole tool definition invalid.

This is reachable from several documented, public entry points:

- [`FunctionTool`][agents.tool.FunctionTool] constructed directly with `params_json_schema`, which the [tools guide](https://openai.github.io/openai-agents-python/tools/#custom-function-tools) documents as a supported pattern. `FunctionTool.__post_init__` runs `ensure_strict_json_schema()` and leaves `strict_json_schema=True`.
- MCP tools converted with `convert_schemas_to_strict=True`. `MCPUtil.to_function_tool` only sets `is_strict = True` when `ensure_strict_json_schema()` does not raise. Because the typeless case does not raise, the tool is marked strict while carrying a schema the API rejects, and the non-strict fallback that exists for genuinely invalid schemas never engages.
- `handoff(..., input_json_schema=...)` and custom `AgentOutputSchemaBase` implementations, which call the same helper.

The observable failure is an opaque HTTP 400 from the provider at run time rather than a clear error at construction time, and the user's schema looks correct to them because the SDK accepted it and even reported `strict_json_schema == True`.

### Debug information

- Related library versions (optional, e.g. `any-llm`, `litellm`, or `pydantic`): `pydantic` 2.x, `openai` 2.x
- Python version: 3.13.6
- Operating system: Windows 11 (not platform specific — the helper is pure Python)
- Model and model provider: Not model specific. The schema defect is produced offline; the 400 is returned by any provider that enforces strict structured outputs (OpenAI Responses and Chat Completions).

The failure is a schema-shape defect, so the relevant log is the emitted schema itself rather than a traceback. Output of the repro script below:

```text
root: {'properties': {...}} (no type)
   {"properties": {"a": {"type": "string"}}, "required": ["a"]}

with explicit type=object
   {"type": "object", "properties": {"a": {"type": "string"}}, "additionalProperties": false, "required": ["a"]}

type: ["object","null"]
   {"type": ["object", "null"], "properties": {"a": {"type": "string"}}, "required": ["a"]}

nested object without type
   {"type": "object", "properties": {"cfg": {"properties": {"k": {"type": "string"}}, "required": ["k"]}}, "additionalProperties": false, "required": ["cfg"]}

anyOf branches without type
   {"type": "object", "properties": {"u": {"anyOf": [{"properties": {"x": {"type": "string"}}, "required": ["x"]}, {"type": "object", "properties": {"y": {"type": "string"}}, "additionalProperties": false, "required": ["y"]}]}}, "additionalProperties": false, "required": ["u"]}

array items object without type
   {"type": "object", "properties": {"rows": {"type": "array", "items": {"properties": {"c": {"type": "string"}}, "required": ["c"]}}}, "additionalProperties": false, "required": ["rows"]}

FunctionTool.strict_json_schema
   true
FunctionTool.params_json_schema (post strict conversion)
   {"properties": {"a": {"type": "string"}}, "required": ["a"]}
```

Note in the `anyOf` output that the two branches are treated inconsistently within the *same* schema: the branch that declares `"type": "object"` gets `additionalProperties: false`, the branch that omits it does not.

### Repro steps

Minimal, self-contained, and requires no API key:

```python
import json
from typing import Any

from agents import FunctionTool

# 1. Typeless object schema: `required` is added, `additionalProperties` is not.
print(json.dumps(ensure_strict_json_schema({"properties": {"a": {"type": "string"}}})))
# {"properties": {"a": {"type": "string"}}, "required": ["a"]}

# 2. Nullable object schema (OpenAPI 3.1 style): same gap.
print(json.dumps(ensure_strict_json_schema(
    {"type": ["object", "null"], "properties": {"a": {"type": "string"}}}
)))
# {"type": ["object", "null"], "properties": {"a": {"type": "string"}}, "required": ["a"]}

# 3. Compare against the schema that does declare the type.
print(json.dumps(ensure_strict_json_schema(
    {"type": "object", "properties": {"a": {"type": "string"}}}
)))
# {"type": "object", ..., "additionalProperties": false, "required": ["a"]}


# 4. End-to-end via the documented custom-FunctionTool pattern.
async def run_function(ctx: ToolContext[Any], args: str) -> str:
    return "ok"


tool = FunctionTool(
    name="process",
    description="Processes data",
    params_json_schema={"properties": {"a": {"type": "string"}}},
    on_invoke_tool=run_function,
)

print(tool.strict_json_schema)        # True  -- advertised to the API as strict
print(json.dumps(tool.params_json_schema))
# {"properties": {"a": {"type": "string"}}, "required": ["a"]}   <- no additionalProperties
```

Attaching that `tool` to an `Agent` and running it against OpenAI fails with
`Invalid schema ... 'additionalProperties' is required to be supplied and to be false`.

### Expected behavior

`ensure_strict_json_schema()` should apply `additionalProperties: false` to every node it already treats as an object for the purpose of the `required` rewrite. Concretely, the object detection should be consistent across the function, for example by treating a node as an object when it has a `properties` mapping, or when `type` is `"object"` **or** a list containing `"object"`.

Either of these outcomes would be acceptable:

- Normalize the schema so it is genuinely strict-valid (preferred, and consistent with what the function already does to `required`).
- Or, if the SDK does not want to infer object-ness, raise a `UserError` at construction time explaining that a strict schema must declare `"type": "object"` — and, in the MCP path, fall back to `strict=False` instead of advertising an invalid strict schema.

What should not happen is the current middle ground, where the schema is half-converted, marked `strict_json_schema=True`, and only fails later as an opaque provider-side 400.

rajarshidattapy:
Closes #4114

`ensure_strict_json_schema()` decided whether to inject `additionalProperties: false`
by comparing `type` against the literal string `"object"`, while the very next block
treated *any* node carrying a `properties` mapping as an object and rewrote `required`.

The two checks disagreed. A node with `properties` but no literal `"type": "object"`
got the strict-mode `required` rewrite and no `additionalProperties`, then went to the
provider with `strict: true`:

400 - Invalid schema: 'additionalProperties' is required to be supplied and to be false.

Two ordinary, valid JSON Schema shapes hit this: typeless object schemas
(`{"properties": {...}}`, common in hand-written and MCP tool schemas) and
nullable objects (`{"type": ["object", "null"]}`, the standard OpenAPI 3.1 spelling).

## The fix

The `type` string comparison is replaced by a single predicate, applied where
`additionalProperties` is decided:

```python
def is_object_schema(json_schema: dict[str, object]) -> bool:
    typ = json_schema.get("type")
    if is_list(typ):
        return "object" in typ
    return typ == "object" or is_dict(json_schema.get("properties"))
```

This makes object detection agree with the required rewrite, which is the actual
defect. Because it lands in the one recursive node handler, every reported site is
covered at once — root, nested properties, anyOf branches, and array items.
The anyOf inconsistency in the issue is gone: both branches are now converted the
same way.

FunctionTool(params_json_schema={"properties": {...}}) now yields a genuinely
strict-valid schema, so strict_json_schema=True is no longer advertised over a
schema the API rejects. The MCP path needs no change: MCPUtil.to_function_tool
already sets is_strict = True only when conversion succeeds, and conversion now
produces a valid schema instead of a half-converted one.

I took the "normalize" option rather than "raise at construction", since it is what
the function already does to required.

## Behavior change worth flagging

additionalProperties: {} (or any non-False value) on a typeless or nullable object
now raises UserError, where it previously passed through. That value genuinely cannot
be made strict, and it is the same treatment {"type": "object"} already received —
extending the existing check to the nodes it should always have covered.

Map-style {"additionalProperties": {...}} with no properties and no type is
untouched; it is not a strict object node and does not start raising.

## Tests

tests/test_strict_schema.py:

- typeless root and ["object", "null"] root both get additionalProperties: false
- parametrized coverage for nested properties, an anyOf branch, and array items
- permissive additionalProperties on a typeless object raises UserError
- non-object leaves (string, array) are not given additionalProperties
- map-style {"additionalProperties": {...}} passes through unchanged

Full suite shows no new failures against the main baseline.


seratch (changes requested):
This catches additional object shapes, but it still leaves provider-invalid roots marked as strict. A root with `properties` but no `type` remains typeless, a root typed as `["object", "null"]` remains nullable, and an open map schema can remain open.

Please make root handling explicit: normalize a properties-bearing typeless root to `type: "object"`, reject nullable roots and open map roots before the provider call, and continue closing nested nullable object branches. The tests should assert the complete converted schema against the strict-mode contract rather than only checking `additionalProperties`.

rajarshidattapy:
## Changes made:

Root handling is now explicit in `ensure_strict_json_schema`: a `properties`-bearing typeless root is normalized to `type: "object"` (and `["object"]` collapses to `"object"`), while nullable roots (`["object", "null"]`) and open map roots (`additionalProperties` not `false`, including the no-`properties` map case) are rejected with a `UserError` at conversion time instead of failing as a provider 400. The check runs on the converted schema, so `allOf`/`$ref` merges can't reopen the root after the node-level check has passed.

Nested branches are unchanged — typeless and nullable object branches are still closed with `additionalProperties: false` and full `required`. Tests now assert the complete converted schema against the strict-mode contract (typeless root, `["object"]` root, and a nested tree covering typeless / nullable / `anyOf` / `items` / non-object siblings) plus the two root rejections, replacing the `additionalProperties`-only checks.

@seratch can you review this once?

seratch (changes requested):
Thanks for the update. The root handling now follows the strict schema contract, and nested nullable objects are accepted by the Responses API. However, the nested typeless case is still provider-invalid.

I validated the current output against the Responses API. A nested schema with `properties`, `required`, and `additionalProperties: false`, but no `type`, is rejected with `schema must have a 'type' key`. This means the current head still fails for the reported nested property, `anyOf`, and array-item cases while marking the tool as strict.

Please normalize every properties-bearing typeless node to `type: "object"` in the shared recursive converter, not only at the root. Keep nested nullable objects as `["object", "null"]`, and retain the current rejection of nullable or open-map roots. Please also update the nested schema expectations and add one focused regression through `FunctionTool` or MCP conversion that asserts the complete emitted strict schema.

Once those changes are covered and CI is green, this should be ready for another review.

rajarshidattapy:
@seratch 

Moved the normalization into the shared recursive converter: any node identified as an object that carries no `type` is now set to `type: "object"`, so the nested property, `anyOf`, and array-item cases are emitted typed rather than only closed. The guard is on `type` being absent, so a node that already declares one is untouched and nested nullable objects keep their `["object", "null"]` spelling; the root rejections for nullable and open-map schemas are unchanged. This made the root-level typeless branch redundant, so it's gone — the rule now has one source of truth.

Nested expectations updated to assert the full converted schema with `type` on every object branch, and added `test_function_tool_emits_typed_and_closed_nested_objects`, which constructs a strict `FunctionTool` from a typeless-node schema and asserts the complete emitted `params_json_schema`.

seratch:
Thanks for working on this. This PR isn't ready yet, so I am going to merge https://github.com/openai/openai-agents-python/pull/4139 with your contribution credit. Thanks again for your effort here.

PR #4119, opened 2026-08-02
fix(voice): clean up stream tasks when consumer exits early

AAliKKhan:
This pull request fixes a resource leak in `StreamedAudioResult.stream()` when a consumer stops iterating the async generator early (for example, when a user stops audio playback before the full response has been generated).

### Bug
`stream()` is the public API for consuming voice pipeline output. Its producer cleanup (`_cleanup_tasks()`) only ran at the bottom of the async generator, which is not reached on every exit path. A consumer that `break`s out of `async for event in result.stream()` finalizes the generator by throwing `GeneratorExit` at the `yield` point; because the cleanup `try/finally` lived after the last `yield`, it never ran. As a result the producer tasks (`text_generation_task`, the audio dispatcher, and the per-segment TTS stream tasks) kept running after the consumer was done, and the active trace span was never finished.

### Fix
Wrap `stream()`'s body in a `try/finally` that runs `_cleanup_tasks()` on every exit path, including early consumer `GeneratorExit`. This cancels the mid-flight producer tasks and finishes the trace span once the consumer stops iterating.

### Details
- `StreamedAudioResult.stream()` now guarantees cleanup on normal completion, error, break, and early exit.
- Added `tests/voice/test_pipeline.py::test_streamed_audio_result_cleans_up_producer_on_early_exit`, which reproduces the leak (the producer tasks remain running and the trace span stays open) and passes after the fix.

This is a behavior fix with no change to the public API contract.

### DIFF
--- a/src/agents/voice/result.py
+++ b/src/agents/voice/result.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 import base64
 from collections import deque
-from collections.abc import AsyncIterator
+from collections.abc import AsyncGenerator
 from typing import Any
 
 from ..exceptions import UserError
@@ -316,7 +316,7 @@ class StreamedAudioResult:
                     self._stored_exception = task.exception()
                     break
 
-    async def stream(self) -> AsyncIterator[VoiceStreamEvent]:
+    async def stream(self) -> AsyncGenerator[VoiceStreamEvent, None]:
         """Stream the events and audio data as they're generated."""
         saw_session_end = False
         try:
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
The early-exit cleanup is needed, but the terminal state is recorded after yielding `session_ended`. If a consumer breaks when it receives that event, `GeneratorExit` enters `finally` before `saw_session_end` becomes true, skips the shielded producer completion, and cancels the text-generation task.

Please record the terminal state before yielding the event. Add a controlled regression that receives `session_ended`, explicitly calls `aclose()`, and verifies that the producer and trace finish through the graceful path rather than cancellation.

AAliKKhan:
Thanks for the review. I've addressed the feedback:

- `saw_session_end` is now recorded **before** yielding the `session_ended` event, so a consumer that stops iterating the moment it receives the event (throwing `GeneratorExit` at the yield) still sees the terminal state as reached.
- The shielded graceful producer completion is performed from the generator's `finally`, which now runs on the early-exit path as well as the normal completion path, so the producer/text-generation task is finished (not cancelled) and the active trace can emit `trace_end`.
- The producer may already have failed by the time we reach the shield (e.g. a TTS error), in which case its error is re-raised inside the `finally`'s shielded await; that error is swallowed there so cleanup always runs and the failure still surfaces later from `_stored_exception`.
- `stream()` is now type-annotated as `AsyncGenerator`, enabling an explicit `aclose()`.

Added regression test `test_streamed_audio_result_finishes_producer_when_closed_on_session_end`: it receives `session_ended`, calls `aclose()` while the text-generation task is still in flight, and asserts the producer finishes gracefully (not cancelled) and the tracing span is cleaned up. The test fails against the previous revision (producer is cancelled) and passes with this change.

Verification: `tests/voice/test_pipeline.py` 36 passed; `mypy`, `ruff check`, and `ruff format --check` clean.

seratch:
Thanks for the contribution. This PR overlaps #4060, which already covers both the result-stream and STT generator paths with the required explicit-close, cancellation, task-draining, and interleaving coverage.

The current head also swallows cancellation and terminal producer/TTS errors in the new BaseException handler, and its cleanup can finish the pipeline trace before the active speech-group span. The bare async-for break test additionally relies on deferred async-generator finalization rather than a deterministic close contract.

I am going to close this PR in favor of #4060. We should resolve its remaining terminal-error, cleanup-ordering, and conflict work there rather than maintain another partial implementation.

PR #4120, opened 2026-08-02
fix(voice): finish trailing STT transcription span on session end

AAliKKhan:
This pull request fixes a trace-span leak in the streamed STT transcription session.

### Bug

An `OpenAISTTTranscriptionSession` keeps a listening `transcription` span for the "current" turn. `_stream_audio` starts a span, and each completed transcript ends the previous turn and starts a new span. That leaves one trailing open span for the never-completing next turn. At session end (and on consumer cancellation), `transcribe_turns` tries to close it via `_end_turn("")`, but `_end_turn` early-returned whenever the transcript was empty, so the trailing span was never finished — leaving a dangling `transcription` span in the trace processor (plus an uncleared turn audio buffer) for every streamed session.

### Fix

`_end_turn` no longer bails on an empty transcript; it finishes the open span when one exists. The transcript-driven call site already guards on `len(transcript) > 0`, so the only affected paths are the normal session end and the cancellation path — exactly the two places that should close the trailing span.

### Details

- `OpenAISTTTranscriptionSession._end_turn` closes the active span even for an empty transcript.
- Added `tests/voice/test_openai_stt.py::test_transcribe_turns_finishes_trailing_trace_span`, which reproduces the leak (the trailing `_tracing_span` is still set after the session ends) and passes after the fix.

This affects only the internal tracing lifecycle of the streamed transcription path; the public transcript contract is unchanged.

### DIFF
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -123,9 +123,6 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
         self._tracing_span.start()
 
     def _end_turn(self, _transcript: str) -> None:
-        if len(_transcript) < 1:
-            return
-
         if self._tracing_span:
             # Only encode audio if tracing is enabled AND buffer is not empty
             if self._trace_include_sensitive_audio_data and self._turn_audio_buffer:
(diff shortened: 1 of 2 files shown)


seratch:
Please do not finish an empty transcription span as a successful transcription. The current change creates a bogus empty span after every completed transcript, reports the error path as success, and can report partial audio with no transcript as a successful transcription.

The simpler lifecycle is to start a span lazily when the first audio for a turn arrives, finish it successfully only when the transcript arrives, and avoid eagerly starting the next span. Error and terminal-incomplete paths should be represented as error/incomplete or discarded, never exported as successful empty transcriptions.

This approach could bring major side effects, I am going to close this PR for now.

PR #4121, opened 2026-08-02
fix(agent): reject distinct agents whose names collapse to one derived tool name

### LINKED ISSUE #4118: Agent name conflicts cause the first registered agent to be ignored
### Describe the bug

`Agent.as_tool()` and `handoff()` derive their tool name from the agent name via `agents.util._transforms.transform_string_function_style()`, which replaces every non-alphanumeric character with `_` and lowercases the result:

```python
# src/agents/util/_transforms.py
whitespace_normalized_name = re.sub(r"\s", "_", name)
transformed_name = re.sub(r"[^a-zA-Z0-9_]", "_", whitespace_normalized_name)
final_name = transformed_name.lower()
```

This transform is many-to-one, and nothing checks whether two agents in the same handoff list or tool list land on the same derived name. Agent names that a user reasonably treats as distinct — differing only in case, whitespace, or punctuation — silently collapse:

| Agent name | Derived handoff tool |
|---|---|
| `Billing Agent` | `transfer_to_billing_agent` |
| `billing agent` | `transfer_to_billing_agent` |
| `Billing-Agent` | `transfer_to_billing_agent` |
| `billing.agent` | `transfer_to_billing_agent` |

The run then routes every call for that name to whichever agent was registered last. The earlier agent is registered, appears in `agent.handoffs`, is advertised to the model, and can never be selected.

The existing warning does not cover this. `transform_string_function_style` only warns when characters were actually replaced, and its message is about *character validity*, not collision:

```
Tool name 'Billing Agent' contains invalid characters for function calling and has been
transformed to 'billing_agent'. Please use only letters, digits, and underscores to avoid
potential naming conflicts.
```

It says "to avoid potential naming conflicts", but no conflict check is ever performed. Worse, a case-only difference (`Billing` vs `billing`) triggers **no warning at all**, because `transformed_name != name` is false for a name that already contains only valid characters — the lowercasing is applied afterwards and is deliberately exempt from the warning (see `tests/test_transforms.py::test_transform_string_function_style_does_not_warn_for_case_only_changes`). So the most confusing case is also the completely silent one.

This is distinct from duplicate *identical* tool names: here the user supplied names that really are different, and the SDK's own normalization created the collision. It is also reachable through both public surfaces that use the transform — `Agent.as_tool()` and `handoff()` — and in a multi-agent routing setup the symptom is simply "the router keeps picking the wrong specialist", which is very hard to trace back to name normalization.

### Debug information

- Python version: 3.13.6
- Model and model provider: Reproduced offline with a stub model so the routing outcome is deterministic. The defect is in SDK-side name derivation and is provider independent.

No exception is raised. Output of the repro below:

```text
Tool name 'Refund Agent' contains invalid characters for function calling and has been transformed to 'refund_agent'. ...
Tool name 'refund agent' contains invalid characters for function calling and has been transformed to 'refund_agent'. ...
as_tool names: refund_agent | refund_agent -> collide: True
handoff names: transfer_to_refund_agent | transfer_to_refund_agent -> collide: True

last agent: billing agent
final: BILLING-2
```

The two warnings are emitted independently and neither mentions that the two names have converged. The run then hands off to `billing agent` (the second registration); `Billing Agent` is never reachable.

### Repro steps

Self-contained, no API key required. Uses the repository's `tests/fake_model.py` stub to make routing observable.

```python
import asyncio
import sys

from agents import Agent, Runner, handoff


async def main() -> None:
    billing = Agent(
        name="Billing Agent",
        model=FakeModel(initial_output=[get_text_message("BILLING")]),
    )
    billing_dup = Agent(
        name="billing agent",  # a genuinely different name
        model=FakeModel(initial_output=[get_text_message("BILLING-2")]),
    )

    # Both derive the same handoff tool name -- no error, no collision warning.
    print(handoff(billing).tool_name)      # transfer_to_billing_agent
    print(handoff(billing_dup).tool_name)  # transfer_to_billing_agent

    model = FakeModel()
    triage = Agent(name="Triage", model=model, handoffs=[billing, billing_dup])
    model.add_multiple_turn_outputs(
        [[get_function_tool_call("transfer_to_billing_agent", "{}", call_id="h1")]]
    )

    result = await Runner.run(triage, "I have a billing question")
    print("last agent:", result.last_agent.name)  # 'billing agent'  <- the second one
    print("final:", result.final_output)          # 'BILLING-2'

    # Same collapse via Agent.as_tool(), with no warning for case-only differences:
    a = Agent(name="Refund")
    b = Agent(name="refund")
    print(a.as_tool(tool_name=None, tool_description="d").name)  # refund
    print(b.as_tool(tool_name=None, tool_description="d").name)  # refund  (silently)


asyncio.run(main())
```

### Expected behavior

Distinct agents should never silently share a derived handoff/tool name.

When `Agent.as_tool()` or `handoff()` derives names for a set of agents that collide after normalization, the SDK should raise a `UserError` naming the colliding agents and the derived name, and pointing at the explicit override:

```
Ambiguous handoff configuration: agents 'Billing Agent' and 'billing agent' both derive the
handoff tool name `transfer_to_billing_agent`. Pass an explicit `tool_name_override=` to one
of them.
```

This would match how the SDK already handles derived-name collisions for MCP tools (`MCPUtil._build_prefixed_tool_name_overrides` actively disambiguates colliding names with a hash suffix) and for Codex tools (`_validate_codex_tool_name_collisions` raises).

At minimum, the existing warning should be upgraded to fire on case-only normalization as well, and should state that a collision has actually occurred rather than that one is theoretically possible.

rajarshidattapy:
## Description

Closes #4118

`transform_string_function_style()` is many-to-one — whitespace, punctuation, and case
all normalize away — and nothing checked whether two agents in the same handoff list or
tool list landed on the same derived name:

| Agent name | Derived handoff tool |
|---|---|
| `Billing Agent` | `transfer_to_billing_agent` |
| `billing agent` | `transfer_to_billing_agent` |
| `Billing-Agent` | `transfer_to_billing_agent` |
| `billing.agent` | `transfer_to_billing_agent` |

Dispatch then resolved last-wins. The earlier agent was registered, advertised to the
model, and permanently unreachable. In a routing setup the symptom is just "the router
keeps picking the wrong specialist", which is very hard to trace back to name
normalization — and the case-only variant produced no warning at all, since
`transformed_name != name` is false before lowercasing.

## Changes

**`src/agents/agent.py`** — `validate_derived_agent_name_collisions()` takes
`(derived name, source agent name)` pairs and raises when one derived name is claimed by
two different agent names:

Ambiguous handoff configuration: agents 'Billing Agent' and 'billing agent' both derive the
handoff tool name transfer_to_billing_agent. Pass an explicit tool_name_override= to one of them.

Ambiguous agent tool configuration: agents 'Refund' and 'refund' both derive the agent tool
name refund. Pass an explicit tool_name= to one of them.

Matching agent names on both sides mean the same target, so listing one agent twice
stays harmless — only a genuine collapse raises.

It is wired into the two surfaces the issue names:

- **`get_handoffs()`** (`run_internal/turn_preparation.py`) — where `agent.handoffs`
  resolves `Agent` → `Handoff` and `is_enabled` is applied. Validating after the filter
  means two agents whose names collide stay usable as long as only one is enabled per
  turn.
- **`Agent.get_all_tools()`** — `_agent_tool_name_entries()` picks out `as_tool()`
  entries through their existing `_tool_origin` metadata, which already carries
  `agent_name` (`ToolOriginType.AGENT_AS_TOOL`), so the agents can be named in the error
  with no new bookkeeping. Scoped to agent tools; plain duplicate `@function_tool` names
  are out of scope here.

This mirrors how the SDK already handles derived-name collisions elsewhere —
`MCPUtil._build_prefixed_tool_name_overrides` disambiguates with a hash suffix, and
`_validate_codex_tool_name_collisions` raises.

## Tests

`tests/test_agent_config.py`, which already exercises `get_handoffs`:

- collisions raise for whitespace-, case-, and punctuation-only differences
- an explicit `tool_name_override=` / `tool_name=` resolves the collision
- the same agent listed twice does not raise
- `is_enabled=False` on one side keeps a colliding pair valid
- the `as_tool()` case-only collision — previously silent, with no warning — now raises

## Not included

The issue's "at minimum" fallback: making `transform_string_function_style` warn on
case-only normalization. I read that as the alternative to raising rather than an
addition, and it would require reversing
`tests/test_transforms.py::test_transform_string_function_style_does_not_warn_for_case_only_changes`,
which pins that exemption deliberately. Happy to add it if you'd prefer both.

Full suite shows no new failures against the `main` baseline.

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -119,6 +119,48 @@ def _validate_codex_tool_name_collisions(tools: list[Tool]) -> None:
         )
 
 
+def validate_derived_agent_name_collisions(
+    entries: list[tuple[str, str]],
+    *,
+    config_label: str,
+    name_label: str,
+    override_hint: str,
+) -> None:
+    """Reject distinct agents whose names normalize to the same derived tool name.
+
+    `transform_string_function_style()` is many-to-one -- case, whitespace, and punctuation all
+    collapse into the same output -- so two agents a user considers distinct can land on one
+    derived name. Dispatch then resolves last-wins and the earlier agent is advertised to the
+    model but can never be selected.
+
+    Entries are `(derived name, source agent name)`. Agents sharing both are the same target,
+    so only differing agent names collide.
+    """
+    owners: dict[str, str] = {}
+    for derived_name, agent_name in entries:
+        prior_agent_name = owners.setdefault(derived_name, agent_name)
+        if prior_agent_name != agent_name:
+            raise UserError(
+                f"Ambiguous {config_label} configuration: agents {prior_agent_name!r} and "
+                f"{agent_name!r} both derive the {name_label} `{derived_name}`. "
+                f"Pass an explicit `{override_hint}` to one of them."
+            )
+
+
+def _agent_tool_name_entries(tools: list[Tool]) -> list[tuple[str, str]]:
+    """Return `(tool name, source agent name)` for every `Agent.as_tool()` entry."""
+    entries: list[tuple[str, str]] = []
+    for tool in tools:
+        if not isinstance(tool, FunctionTool):
+            continue
+        origin = getattr(tool, "_tool_origin", None)
+        if origin is None or origin.type is not ToolOriginType.AGENT_AS_TOOL:
+            continue
+        if origin.agent_name:
+            entries.append((tool.name, origin.agent_name))
+    return entries
+
+
 class AgentToolStreamEvent(TypedDict):
     """Streaming event emitted when an agent is invoked as a tool."""
 
@@ -264,6 +306,12 @@ class AgentBase(Generic[TContext]):
         enabled: list[Tool] = [t for t, ok in zip(self.tools, results, strict=False) if ok]
         all_tools: list[Tool] = prune_orphaned_tool_search_tools([*mcp_tools, *enabled])
         _validate_codex_tool_name_collisions(all_tools)
+        validate_derived_agent_name_collisions(
+            _agent_tool_name_entries(all_tools),
+            config_label="agent tool",
+            name_label="agent tool name",
+            override_hint="tool_name=",
+        )
         return all_tools
 
 
--- a/src/agents/run_internal/turn_preparation.py
+++ b/src/agents/run_internal/turn_preparation.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import inspect
 from typing import Any
 
-from ..agent import Agent
+from ..agent import Agent, validate_derived_agent_name_collisions
 from ..agent_output import AgentOutputSchema, AgentOutputSchemaBase
 from ..exceptions import UserError
 from ..handoffs import Handoff, handoff
@@ -113,6 +113,14 @@ async def get_handoffs(agent: Agent[Any], context_wrapper: RunContextWrapper[Any
 
     results = await gather_with_cancel(*(check_handoff_enabled(h) for h in handoffs))
     enabled: list[Handoff] = [h for h, ok in zip(handoffs, results, strict=False) if ok]
+    # Validate after `is_enabled` filtering, so agents whose derived names collide stay usable
+    # as long as only one of them is enabled per turn.
+    validate_derived_agent_name_collisions(
+        [(h.tool_name, h.agent_name) for h in enabled],
+        config_label="handoff",
+        name_label="handoff tool name",
+        override_hint="tool_name_override=",
+    )
     return enabled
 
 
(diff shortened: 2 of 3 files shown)


seratch:
This overlaps the model-visible collision invariant already being addressed in #4117, and the separate helper here is not safe. It drops tool namespaces, compares agent display names rather than agent identity, and exempts the same Agent repeated even though duplicate function definitions can still be emitted to the model.

Please close or supersede this PR in favor of #4117 and transfer the useful normalized agent/handoff regressions there. The implementation should use the canonical namespace-aware lookup identity from `_tool_identity.py`, preserving legitimate equal bare names in different namespaces and maintaining one source of truth for collision detection.

PR #4128, opened 2026-08-03
fix(voice): clean up tasks when streams close early

AAliKKhan:
## Summary

Fixes voice stream/task lifecycle cleanup when streams close early. When a consumer closes a `StreamedAudioResult` (or an STT transcription session) right after receiving a terminal event, finalization must not hide the real producer outcome.

Key changes in `src/agents/voice/result.py` and `src/agents/voice/models/openai_stt.py`:

1. **Surface terminal producer errors on close.** `stream()` now tracks the primary exception object and re-raises the cleanup exception when the primary exception is `GeneratorExit` after a delivered `session_ended`. Previously the injected `GeneratorExit` swallowed a producer/TTS error that surfaced during finalization (via the `asyncio.shield` re-raise), so `aclose()` reported a clean close instead of the real failure.

2. **Graceful producer completion on early close.** After a terminal event is recorded (before the yield), finalization shields the producer task so the session-close and `trace_end` path completes before cleanup, instead of cancelling the producer mid-flight.

3. **Cancellation always propagates.** Cleanup failure handling always re-raises `CancelledError` and preserves the primary consumer exception; non-cancellation cleanup errors only emit a fixed `logger.warning(...)` with no attached consumer context.

4. **Single finalization path.** Both iterators drain owned tasks and finish the tracing span after tasks settle (cancel + drain, then span finish in a `finally`).

The branch is based on current `upstream/main` with no conflicts.

## Test Plan

- `uv run pytest tests/voice/test_pipeline.py tests/voice/test_openai_stt.py -q` — 57 passed.
- New regressions: producer finishes (not cancelled) when consumer closes after `session_ended`; cancellation during session-end cleanup propagates; terminal producer error after `session_ended` surfaces from `aclose()` (fails before the fix).
- `uv run ruff check` and `uv run ruff format --check` clean; `uv run mypy` on voice sources/tests clean.
- Full-suite collection errors (`tee not found` in `tests/mcp/helpers.py`, sandbox symlink/tracing failures) are pre-existing on Windows `upstream/main` and reproduce identically without this branch.

### DIFF
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -123,9 +123,6 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
         self._tracing_span.start()
 
     def _end_turn(self, _transcript: str) -> None:
-        if len(_transcript) < 1:
-            return
-
         if self._tracing_span:
             # Only encode audio if tracing is enabled AND buffer is not empty
             if self._trace_include_sensitive_audio_data and self._turn_audio_buffer:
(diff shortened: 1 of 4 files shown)


AAliKKhan:
Fixed in `332e0e83`.

When a consumer closes the stream right after `session_ended`, the producer shield was guarded by `not self.text_generation_task.done()`. If the producer/TTS task had already failed before close, the shield was skipped entirely and cleanup gathered the completed task with `return_exceptions=True`, so `aclose()` reported a clean close instead of the terminal failure.

`stream()` now awaits the shielded producer whenever a terminal event was delivered — awaiting an already-completed task is a no-op that re-raises its stored exception, so an already-finished producer failure surfaces from `aclose()` through the existing `GeneratorExit`-after-`session_ended` re-raise path.

Added `test_streamed_audio_result_surfaces_failed_producer_already_done_on_close`, which fails before this change (no exception raised) and passes with it. Full voice/STT suite: 58 passed; ruff and mypy clean.

seratch:
@AAliKKhan Just picking https://github.com/openai/openai-agents-python/pull/4060's changes up this way doesn't seem to work, plus you didn't include #4060's credit here. I will handle this later, so let me close this PR.

PR #4133, opened 2026-08-03
fix(models): release AnyLLM provider streams and preserve completed runs

hsusul:
### Summary

`AnyLLMModel`'s streamed model calls never adopted the provider-stream lifecycle that the other adapters use, so a streamed run could leak the iterator any-llm returns, lose its span data, and let a cleanup error turn a completed run into a failure.

**Affected component:** `src/agents/extensions/models/any_llm_model.py` (`AnyLLMModel.stream_response`, `_stream_response_via_chat`, `_stream_response_via_responses`).

**Problem**

1. `stream_response` delegates to `_stream_response_via_chat` / `_stream_response_via_responses` with a bare `async for ... yield`. Closing the returned generator early (`aclose()`, `break` followed by `aclose()`, a consumer that stops at `response.completed`) never propagates to the delegate, so the delegate's `finally` — the only place the provider iterator is closed — does not run until the delegate is garbage collected.
2. Both delegates close the iterator in a plain `finally: await self._maybe_aclose(stream)`. If that close raises after the run already yielded its terminal event, the cleanup error replaces a successful result and surfaces to the caller.
3. Cancellation arriving *while* that `aclose()` is already awaiting the provider abandons the close half-done.
4. Span data (`output`/`usage` for Chat, `response`/`input` for Responses) was populated only after the iteration loop finished, so a consumer that stops at the terminal event left an empty span. The Chat path also assigned `final_response` after `yield`, so it was never set in that case.

`OpenAIChatCompletionsModel`, `OpenAIResponsesModel` and `LitellmModel` all already implement the close-on-cancel and terminal-event cleanup parts of this contract (#3689, #4066, #4077). `AnyLLMModel` is the only streaming adapter that does not.

**Corrected behavior**

- An early `aclose()` deterministically closes the iterator any-llm returned.
- A completed run stays successful when cleanup fails; the error is logged at debug level. Cleanup errors *before* a terminal event still propagate.
- Cancellation during an in-flight `aclose()` detaches that exact close to the background so it finishes, and never starts a second close — re-closing a provider iterator is not guaranteed to be safe or idempotent.
- Span data is populated before the terminal event is yielded, so a consumer that stops there still leaves a fully recorded span.

**Scope of the early-close guarantee**

This change guarantees the SDK closes **the iterator any-llm hands back**. It deliberately stops there. As of any-llm 1.11.0, `utils/exception_handler.py::_wrap_async_iterator` and the per-provider `chunk_iterator` / `_stream_completion_async` / `_stream` generators all delegate with bare `async for ... yield` and no `try/finally`, so they do not forward `aclose()` down to the underlying HTTP transport. Propagating a close through those layers is an upstream any-llm concern and should be tracked there; working around it would mean reaching into any-llm internals from the SDK, which this PR does not do. The tests use fake provider iterators and therefore assert only the SDK-side boundary — a docstring on the fake makes that explicit so no assertion reads as proof that the transport was closed.

**Implementation**

- Wrap both delegations in `contextlib.aclosing(...)` so an early `aclose()` reaches the delegate.
- Add `_close_stream_allowing_background_completion`, which runs `_maybe_aclose` as a shielded task and, on cancellation, detaches that same task instead of starting another close.
- In both delegates: `except asyncio.CancelledError` → close in the background and re-raise; `finally` → close and, when a terminal event was already yielded, log the failure instead of raising.
- Populate the span inside the terminal-event branch, before `yield`. The Chat path's span population moved into `_populate_chat_generation_span` so the ordering is explicit rather than implied by position.
- Narrow the two private delegate return annotations from `AsyncIterator[...]` to `AsyncGenerator[..., None]` so `aclosing` type-checks. No public signature changes.

**Why this is minimal:** it reuses the pattern already reviewed and merged for the other three streaming adapters rather than introducing a shared abstraction, and touches one source file. Note that the in-flight-close handling is slightly *stronger* than what the sibling adapters do today (they `await` the close directly); applying it to them is deliberately left out of this PR.

### Test plan

13 regression tests in `tests/models/test_any_llm_model.py`, covering both AnyLLM streaming paths with local fake provider iterators — no API key, no network, and no sleeps for synchronization (`asyncio.Event` gates every cancellation and in-flight-close test):

- cleanup failure after a terminal event keeps the run successful (both paths)
- cleanup failure after a terminal *failure* event still surfaces the terminal failure, not the cleanup error (Responses path)
- cleanup failure before a terminal event still propagates (both paths)
- terminal state is recorded before the completed event is yielded (Chat path)
- span data is populated before `response.completed` is delivered, asserted while the generator is suspended at the terminal yield — `output` + `usage` for Chat, `response` + `input` for Responses (both paths)
- cancellation while `aclose()` is in flight: exactly one close is started, it is still incomplete at cancellation, and it completes in the background once released (both paths)
- cancellation still releases the provider iterator (both paths)
- a slow provider close does not delay cancellation and completes in the background (Chat path)

7 tests fail on `upstream/main` at d6f82c00; the 4 span/in-flight-close tests fail on the first commit of this PR and pass on the second. The 2 plain cancellation tests pass throughout and guard behavior this change preserves.

Commands run from the repository root:

```
uv run pytest tests/models/test_any_llm_model.py -q
```
→ `49 passed`

```
```
→ `make format` (no changes), `make lint` passed, `make typecheck` passed (mypy + pyright, `0 errors`), `make tests` passed — `all commands passed`.

Oldest supported interpreter, per `AGENTS.md`:

```
UV_PROJECT_ENVIRONMENT=.venv_310 uv sync --python 3.10 --all-extras --all-packages --group dev
UV_PROJECT_ENVIRONMENT=.venv_310 uv run --python 3.10 -m pytest tests/models/test_any_llm_model.py -q
```
→ `45 passed, 4 skipped` (the 4 skips are the pre-existing `any-llm-sdk` optional-dependency skips). `contextlib.aclosing` is available on 3.10+.

The focused module was also run 5 times consecutively and under `-W error::RuntimeWarning` with no flakiness, pending-task warnings, or unclosed-resource warnings.

**Compatibility:** no public API, constructor, field-order, dependency, or lockfile changes. Only the two private `_stream_response_via_*` return annotations were narrowed. Event ordering, terminal-failure exceptions, usage accounting and span contents are unchanged on the success path — only *when* the span is populated changed.

**Non-goals / not changed:** no shared cleanup helper extracted across the four adapters; no back-porting of the in-flight-close handling to the OpenAI/LiteLLM adapters; no workaround for any-llm's own missing `aclose()` propagation; no changes to the non-streaming AnyLLM paths.

### Issue number

None — found by comparing `AnyLLMModel` against the sibling adapters after #4066/#4077. Searched open and closed issues and PRs for `any_llm`, `AnyLLMModel`, `any-llm stream`, and `stream cleanup`; no existing report or PR covers the AnyLLM adapter.

### Checks

### DIFF
--- a/src/agents/extensions/models/any_llm_model.py
+++ b/src/agents/extensions/models/any_llm_model.py
@@ -309,6 +309,10 @@ class AnyLLMModel(Model):
     ) -> AsyncIterator[TResponseStreamEvent]:
         # `aclosing` forwards an early `aclose()` on this generator to the delegate, so the
         # delegate's cleanup runs deterministically instead of waiting for garbage collection.
+        # The guarantee stops at the iterator any-llm returns: as of any-llm 1.11.0 its
+        # exception and provider wrappers delegate with bare `async for ... yield`, so they do
+        # not forward `aclose()` to the underlying transport. Closing that transport is an
+        # upstream any-llm concern, not something to reach into from here.
         if self._selected_api() == "responses":
             async with contextlib.aclosing(
                 self._stream_response_via_responses(
@@ -463,6 +467,11 @@ class AnyLLMModel(Model):
                         "response.error",
                     }:
                         yielded_terminal_event = True
+                        # Populate the span before yielding the terminal event so a consumer
+                        # that stops there still leaves a fully recorded span.
+                        if tracing.include_data() and final_response:
+                            span_response.span_data.response = final_response
+                            span_response.span_data.input = input
                     yield chunk
             except asyncio.CancelledError:
                 close_stream_in_background = True
@@ -471,7 +480,7 @@ class AnyLLMModel(Model):
             finally:
                 if not close_stream_in_background:
                     try:
-                        await self._maybe_aclose(stream)
+                        await self._close_stream_allowing_background_completion(stream)
                     except Exception as exc:
                         if yielded_terminal_event:
                             log_model_action_debug(
@@ -484,10 +493,6 @@ class AnyLLMModel(Model):
 
             if terminal_failure_error is not None:
                 raise terminal_failure_error
-
-            if tracing.include_data() and final_response:
-                span_response.span_data.response = final_response
-                span_response.span_data.input = input
 
     async def _get_response_via_chat(
         self,
@@ -634,11 +639,14 @@ class AnyLLMModel(Model):
                     cast(Any, self._normalize_chat_stream(stream)),
                     model=self.model,
                 ):
-                    # Record terminal state before yielding so a consumer that stops at the
-                    # completed event still leaves the tracing data and cleanup policy correct.
+                    # Record terminal state and populate the span before yielding so a consumer
+                    # that stops at the completed event still leaves a fully recorded span.
                     if chunk.type == "response.completed":
                         final_response = chunk.response
                         yielded_terminal_event = True
+                        self._populate_chat_generation_span(
+                            span_generation, final_response, tracing
+                        )
 
                     yield chunk
             except asyncio.CancelledError:
@@ -648,7 +656,7 @@ class AnyLLMModel(Model):
             finally:
                 if not close_stream_in_background:
                     try:
-                        await self._maybe_aclose(stream)
+                        await self._close_stream_allowing_background_completion(stream)
                     except Exception as exc:
                         if yielded_terminal_event:
                             log_model_action_debug(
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Before merge, please address two lifecycle gaps: populate the Chat and Responses span data before yielding `response.completed`, and ensure that cancellation arriving while `aclose()` is already in progress lets that exact close continue in the background without starting a second close. Please add event-controlled regression tests for both paths that assert the completed span data and final provider-stream state.

Please also narrow the early-close guarantee to the iterator returned by any-llm. In any-llm 1.11.0, the Chat exception and provider wrappers use bare async-generator delegation and do not propagate `aclose()` to the underlying transport; the current fake-provider tests bypass those layers. That transport-level propagation should be tracked upstream rather than handled through SDK introspection.

seratch (approved):

PR #4149, opened 2026-08-03
fix(tools): reject duplicate bare function tool names

### LINKED ISSUE #4116: Reject duplicate function tool names
## Describe the bug

When an `Agent` is given two `FunctionTool`s that share the same name, the SDK neither rejects nor deduplicates them. Both are advertised to the model in the `tools` array, and at dispatch time the second one silently shadows the first, which becomes permanently unreachable.

The SDK clearly intends to catch this. `agents._tool_identity.validate_function_tool_lookup_configuration()` is called from `build_function_tool_lookup_map()` and exists specifically to "Reject function-tool combinations that are ambiguous on the Responses wire". It detects the collision correctly, and then explicitly declines to act on it:

```python
# src/agents/_tool_identity.py
prior_owner = qualified_name_owners.get(qualified_name)
if prior_owner is None:
    qualified_name_owners[qualified_name] = tool
    continue

prior_namespace = get_explicit_function_tool_namespace(prior_owner)
if explicit_namespace is None and prior_namespace is None:
    continue          # <-- collision detected, then ignored

raise UserError(
    "Ambiguous function tool configuration: the qualified name "
    f"`{qualified_name}` is used by multiple tools. ..."
)
```

The `continue` on the "neither tool has an explicit namespace" branch is exactly the common case: two plain `@function_tool`s. So the guard fires for namespaced tools but is a no-op for the configuration users are most likely to hit.

Everything downstream then behaves as if this were legal:

- `build_function_tool_lookup_map()` is documented as using "last-wins precedence", so the map ends up pointing at whichever duplicate appears later in `agent.tools`.
- `Agent.get_all_tools()` returns both tools unchanged, so the model receives `tools: [..., {"name": "lookup"}, {"name": "lookup"}, ...]`.

The OpenAI API rejects duplicate function names in a single request, so in production this surfaces as an opaque provider-side 400 rather than a clear SDK error. On providers that tolerate duplicates, it is worse: the run succeeds while silently executing the wrong implementation.

This is easy to hit unintentionally — for example when two tool modules are combined, when a helper is exported under an explicit `name_override`, or when `Agent.clone()`/list concatenation appends a tool that is already present.

The SDK already implements exactly this validation elsewhere, which makes the gap look like an oversight rather than a deliberate policy:

- MCP tools: `MCPUtil.get_all_function_tools()` raises `UserError("Duplicate tool names found across MCP servers: ...")`.
- Codex tools: `Agent.get_all_tools()` calls `_validate_codex_tool_name_collisions()`, which raises `UserError("Duplicate Codex tool names found: ... Provide a unique codex_tool(name=...) per tool instance.")`.

Plain function tools are the only category with no equivalent check. (Note this is also distinct from #3561, which added duplicate detection *within an individual MCP server*.)

### Debug information

- Python version: 3.13.6
- Model and model provider: Reproduced offline with a stub model so the wrong-tool dispatch is observable without a network call. Against OpenAI (Responses or Chat Completions) the same agent configuration fails with a duplicate-function-name 400.

No exception is raised by the SDK — that is the bug. Output of the repro below:

```text
duplicate tool names accepted; final: done
tools sent to model: ['lookup', 'lookup']
which ran -> ['b']
```

`which ran -> ['b']` shows the *second* registered tool served a call that the first tool was equally eligible for. The first tool can never be invoked.

### Repro steps

This uses the repository's `tests/fake_model.py` stub so the dispatch outcome is visible without an API key. The same agent configuration against a real OpenAI model fails with a 400 instead.

```python
import asyncio
import json
import sys

@function_tool(name_override="lookup")
def lookup_customers(x: str) -> str:
    """Look up customers."""
    return "a"


@function_tool(name_override="lookup")
def lookup_orders(x: str) -> str:
    """Look up orders."""
    return "b"


async def main() -> None:
    model = FakeModel()
    agent = Agent(name="A", model=model, tools=[lookup_customers, lookup_orders])
    model.add_multiple_turn_outputs(
        [
            [get_function_tool_call("lookup", json.dumps({"x": "1"}), call_id="z1")],
            [get_text_message("done")],
        ]
    )

    result = await Runner.run(agent, "hi")

    # No UserError was raised anywhere above.
    print("tools sent to model:", [t.name for t in model.first_turn_args["tools"]])
    # ['lookup', 'lookup']  -> the API rejects this payload

    print(
        "which ran ->",
        [i.raw_item["output"] for i in result.new_items if i.type == "tool_call_output_item"],
    )
    # ['b']  -> lookup_orders silently shadowed lookup_customers


asyncio.run(main())
```

Equivalent minimal form without the test stub, showing the validator itself declines to raise:

```python
from agents._tool_identity import validate_function_tool_lookup_configuration


@function_tool(name_override="lookup")
def a(x: str) -> str:
    """A."""
    return "a"


@function_tool(name_override="lookup")
def b(x: str) -> str:
    """B."""
    return "b"


validate_function_tool_lookup_configuration([a, b])  # returns None; expected UserError
```

### Expected behavior

Two function tools that resolve to the same public name on a single agent should raise a `UserError` at tool-resolution time, consistent with how the SDK already treats duplicate MCP tool names and duplicate Codex tool names.

Concretely, in `validate_function_tool_lookup_configuration()`, the `explicit_namespace is None and prior_namespace is None` branch should raise rather than `continue`. The error should name the colliding tool name and suggest `name_override=` as the fix, mirroring the existing messages:

```
Ambiguous function tool configuration: the tool name `lookup` is used by multiple tools.
Pass a unique `name_override=` to one of them to avoid ambiguous dispatch.
```

Failing that, the collision should at minimum produce a warning and the duplicate should be dropped from the payload sent to the model, so the SDK never emits a `tools` array the API is guaranteed to reject.

daleselaji-dev:
This pull request fixes silent last-wins behavior for duplicate bare function tool names during strict lookup validation.

## Problem

When two `FunctionTool` instances expose the same public name without an explicit namespace, the validator detects the collision but currently continues without raising an error. This can allow ambiguous dispatch and make one tool silently shadow another.

## Approach

- Make strict function-tool lookup validation raise `UserError` for duplicate bare names.
- Include the conflicting tool name and recommend using `name_override=` to resolve it.
- Keep the existing warning-and-deduplication behavior for `tool_name_collision_policy="warn"`.
- Preserve last-wins behavior only in approval and historical-response recovery paths where it is required for compatibility.
- Add regression coverage for direct validation and lookup-map construction.

## Compatibility

This change does not alter MCP or Codex tool naming, permissions, persistence, or wire schemas. It only makes ambiguous strict lookup configurations fail earlier and more clearly.

## Validation

- Focused boundary tests: 293 passed.
- Changed-file mypy: passed.
- Changed-file pyright: passed.
- Full repository validation remains blocked by unrelated optional dependencies and Windows-specific sandbox/MCP failures.

This pull request resolves #4116.

### DIFF
--- a/src/agents/_tool_identity.py
+++ b/src/agents/_tool_identity.py
@@ -401,7 +401,7 @@ def resolve_tool_name_collisions(
     collision_policy: Literal["warn", "error"],
 ) -> tuple[list[Any], list[Any]]:
     """Resolve bare function-tool and handoff name collisions before model exposure."""
-    validate_function_tool_lookup_configuration(tools)
+    validate_function_tool_lookup_configuration(tools, allow_bare_name_collisions=True)
 
     owners: dict[BareFunctionToolLookupKey, list[tuple[str, int, Any]]] = {}
     for index, tool in enumerate(tools):
@@ -446,7 +446,9 @@ def resolve_tool_name_collisions(
     )
 
 
-def validate_function_tool_lookup_configuration(tools: Sequence[Any]) -> None:
+def validate_function_tool_lookup_configuration(
+    tools: Sequence[Any], *, allow_bare_name_collisions: bool = False
+) -> None:
     """Reject function-tool combinations that are ambiguous on the Responses wire."""
     qualified_name_owners: dict[str, Any] = {}
     deferred_top_level_name_owners: dict[str, Any] = {}
@@ -478,7 +480,13 @@ def validate_function_tool_lookup_configuration(tools: Sequence[Any]) -> None:
 
         prior_namespace = get_explicit_function_tool_namespace(prior_owner)
         if explicit_namespace is None and prior_namespace is None:
-            continue
+            if allow_bare_name_collisions:
+                continue
+            raise UserError(
+                "Ambiguous function tool configuration: the tool name "
+                f"`{qualified_name}` is used by multiple tools. Pass a unique "
+                "`name_override=` to one of them to avoid ambiguous dispatch."
+            )
 
         raise UserError(
             "Ambiguous function tool configuration: the qualified name "
@@ -488,9 +496,14 @@ def validate_function_tool_lookup_configuration(tools: Sequence[Any]) -> None:
         )
 
 
-def build_function_tool_lookup_map(tools: Sequence[Any]) -> dict[FunctionToolLookupKey, Any]:
-    """Build a function-tool lookup map using last-wins precedence."""
-    validate_function_tool_lookup_configuration(tools)
+def build_function_tool_lookup_map(
+    tools: Sequence[Any], *, allow_bare_name_collisions: bool = False
+) -> dict[FunctionToolLookupKey, Any]:
+    """Build a function-tool lookup map after validating its dispatch keys."""
+    validate_function_tool_lookup_configuration(
+        tools,
+        allow_bare_name_collisions=allow_bare_name_collisions,
+    )
     tool_map: dict[FunctionToolLookupKey, Any] = {}
     for tool in tools:
         for lookup_key in get_function_tool_lookup_keys(tool):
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for trying to resolve the issue. It was resolved by https://github.com/openai/openai-agents-python/pull/4145

PR #4152, opened 2026-08-03
fix(memory): reject reused branch IDs when creating a branch

### LINKED ISSUE #4150: AdvancedSQLiteSession.create_branch_from_turn silently merges into an existing branch
### Please read this first

- **Have you searched for related issues?** Yes — no existing issue or PR covers branch-id collisions in `AdvancedSQLiteSession`.

### Describe the bug

`AdvancedSQLiteSession.create_branch_from_turn()` never checks whether the target branch id already exists. `_copy_messages_to_new_branch()` unconditionally inserts `message_structure` rows tagged with that branch id, so creating a branch whose id is already in use appends the copied history to the existing branch instead of creating a new one. The call still returns the branch id and switches to it, so the caller has no signal that anything went wrong.

This is reachable two ways:

1. **The SDK's own default branch name can collide.** When `branch_name` is omitted the name is `f"branch_from_turn_{turn_number}_{int(time.time())}"`. That has one-second resolution, so branching twice from the same turn within the same second — the documented usage, `await session.create_branch_from_turn(2)` — produces the same id and silently duplicates the copied history. No user error is involved.
2. **An explicitly supplied existing name is accepted**, including `"main"`, which duplicates the base branch's own history into itself.

### Debug information

- Agents SDK version: `main` @ c546ca12091b16a5bfbb73cfbdf9827e7a8c6f6a
- Python version: 3.13.5
- OS: macOS (Darwin 24.6.0)
- No API key, model call, or network access is needed to reproduce.

### Repro steps

```python
import asyncio
import time

ITEMS = [
    {"role": "user", "content": "turn one"},
    {"role": "assistant", "content": "reply one"},
    {"role": "user", "content": "turn two"},
    {"role": "assistant", "content": "reply two"},
    {"role": "user", "content": "turn three"},
    {"role": "assistant", "content": "reply three"},
]


async def main() -> None:
    # 1. Auto-generated names collide within the same second.
    session = AdvancedSQLiteSession(session_id="collide", create_tables=True)
    await session.add_items(ITEMS)
    with patch.object(time, "time", lambda: 1_700_000_000.0):
        first = await session.create_branch_from_turn(3)
        await session.switch_to_branch("main")
        second = await session.create_branch_from_turn(3)
    print(first, second)                     # identical ids
    print(len(await session.get_items()))    # 8, expected 4
    print([b["branch_id"] for b in await session.list_branches()])  # only one new branch
    session.close()

    # 2. An existing name is accepted and corrupts that branch.
    session = AdvancedSQLiteSession(session_id="into-main", create_tables=True)
    await session.add_items(ITEMS)
    before = await session.get_items(branch_id="main")
    await session.create_branch_from_turn(2, "main")
    after = await session.get_items(branch_id="main")
    print(len(before), len(after))            # 6 8
    session.close()


asyncio.run(main())
```

### Actual behavior

1. Both calls return `branch_from_turn_3_1700000000`. The branch now holds the first two turns twice (8 items instead of 4), and `list_branches()` reports a single branch where the caller created two.
2. `create_branch_from_turn(2, "main")` copies turn 1 into `main`, growing `main` from 6 to 8 items.

### Expected behavior

Creating a branch should never write into an existing branch.

- A branch id supplied by the caller that is already in use should raise `ValueError` — the method already documents `ValueError` for invalid branch requests, and `delete_branch()`/`switch_to_branch()` already use `ValueError` for branch-state errors.
- An auto-generated branch id should be unique, so the documented no-argument form cannot fail or collide.

### Root-cause hypothesis

*(hypothesis)* `_copy_messages_to_new_branch()` (`src/agents/extensions/memory/advanced_sqlite_session.py`) selects the rows to copy and inserts them under `new_branch_id` with no existence check on that id, and `create_branch_from_turn()` derives its default id from a one-second timestamp without checking for an existing branch. Because the check would have to be atomic with the insert, it belongs inside the same locked connection that performs the copy.

### Proposed scope

Resolve the target branch id under the connection lock that performs the copy: reject an explicit id that already exists, and disambiguate a generated id. No public API or schema change, and no change to the copy semantics themselves.

I'd like to work on this.

hsusul:
### Summary

`AdvancedSQLiteSession.create_branch_from_turn()` never checked whether the target branch ID was already in use. `_copy_messages_to_new_branch()` inserted `message_structure` rows tagged with that ID unconditionally, so creating a branch whose ID already existed appended the copied history to that branch instead of creating a new one. The call still returned the ID and switched to it, so the caller got no signal.

Two ways to hit it:

1. **The generated default ID can collide.** With `branch_name` omitted the ID is `f"branch_from_turn_{turn_number}_{int(time.time())}"`. That is one-second resolution, so branching twice from the same turn within the same second — the documented `await session.create_branch_from_turn(2)` form — reuses the ID and silently duplicates the copied history. No user error is involved.
2. **An explicit existing name was accepted**, including `"main"`, which duplicated the base branch's own history into itself.

This resolves the target branch ID inside the locked connection that performs the copy: an explicit ID that is already in use raises `ValueError`, and a generated ID is disambiguated with a numeric suffix.

**Affected component:** `src/agents/extensions/memory/advanced_sqlite_session.py` (`AdvancedSQLiteSession` conversation branching). `create_branch_from_content()` inherits the fix since it delegates.

**Minimal reproduction (no API key, no model call, no network):**

```python
import asyncio
import time

ITEMS = [
    {"role": "user", "content": "turn one"},
    {"role": "assistant", "content": "reply one"},
    {"role": "user", "content": "turn two"},
    {"role": "assistant", "content": "reply two"},
    {"role": "user", "content": "turn three"},
    {"role": "assistant", "content": "reply three"},
]


async def main() -> None:
    session = AdvancedSQLiteSession(session_id="collide", create_tables=True)
    await session.add_items(ITEMS)
    with patch.object(time, "time", lambda: 1_700_000_000.0):
        first = await session.create_branch_from_turn(3)
        await session.switch_to_branch("main")
        second = await session.create_branch_from_turn(3)
    print(first == second, len(await session.get_items()))
    session.close()


asyncio.run(main())
```

**Current behavior:** prints `True 8`. Both calls return `branch_from_turn_3_1700000000`, the branch holds the first two turns twice, and `list_branches()` reports one branch where the caller created two. Separately, `create_branch_from_turn(2, "main")` grows `main` from 6 items to 8.

**Corrected behavior:** prints `False 4`. The two calls produce distinct IDs and each branch holds exactly one copy of the pre-branch history. `create_branch_from_turn(2, "main")` now raises `ValueError` and leaves `main` untouched.

### Root cause

`_copy_messages_to_new_branch()` selected the rows to copy and inserted them under `new_branch_id` with no existence check on that ID, and `create_branch_from_turn()` derived its default ID from a one-second timestamp without consulting existing branches.

### Implementation

- `_branch_id_exists()` — a small helper that asks whether any `message_structure` row already uses a branch ID, matching how `switch_to_branch()`, `delete_branch()`, and `list_branches()` already define branch existence.
- `_generate_branch_id()` — appends a numeric suffix to the timestamp-based base name until it is unused.
- `_copy_messages_to_new_branch()` now takes `str | None` and returns the resolved branch ID. Both the existence check and the ID generation run against the cursor that performs the inserts, inside the same `_locked_connection()`, so the check cannot be separated from the write by a concurrent branch creation. Doing this in a separate `asyncio.to_thread()` step would release the lock in between.
- `create_branch_from_turn()` drops its local ID generation and uses the returned ID for the branch pointer, the debug log, and its return value.

**Why this is minimal:** no public API signature changes, no schema change, and the copy semantics (which rows are copied, sequence numbering, shared message rows) are untouched. The rejection uses `ValueError`, which the method already documented for invalid branch requests and which `delete_branch()` and `switch_to_branch()` already raise for branch-state errors. The `ValueError` is raised before any DML, so a rejected call performs no write and leaves the current-branch pointer alone.

### Test plan

Six regression tests in `tests/extensions/memory/test_advanced_sqlite_session.py`. All five that assert the corrected behavior fail on `upstream/main` at c546ca1 and pass with the fix; the sixth is a boundary guard that passes both before and after.

| Test | Covers |
| --- | --- |
| `test_create_branch_rejects_an_existing_branch_id` | Explicit reused ID raises, both branches and the current pointer are unchanged |
| `test_create_branch_rejects_the_main_branch_id` | Branching into `main` raises instead of duplicating it into itself |
| `test_generated_branch_ids_stay_unique_within_the_same_second` | Frozen clock, two branches from the same turn get distinct IDs and one copy of history each |
| `test_generated_branch_ids_stay_unique_across_many_calls` | Frozen clock, three consecutive generated IDs stay distinct and correct |
| `test_create_branch_from_content_rejects_an_existing_branch_id` | `create_branch_from_content()` inherits the contract |
| `test_create_branch_from_the_first_turn_still_succeeds` | Branching from turn 1 copies nothing and remains allowed (passes before and after) |

The clock is frozen with `monkeypatch.setattr(time, "time", ...)`, so the collision is deterministic rather than timing-dependent. No API key, model call, network access, or sleep is involved; the sessions are in-memory SQLite.

Before the fix:

```
$ uv run pytest tests/extensions/memory/test_advanced_sqlite_session.py -q -k "branch_rejects or generated_branch or from_content_rejects or first_turn_still"
5 failed, 1 passed, 67 deselected in 0.41s
```

After the fix:

```
$ uv run pytest tests/extensions/memory/test_advanced_sqlite_session.py -q
73 passed in 0.32s

$ uv run pytest tests/extensions/memory/ -q
288 passed, 4 skipped in 8.59s
```

The focused module was run 5x with `-p no:randomly` and the memory suite 3x under the default random ordering; results were identical each time.

Full verification stack via `.agents/skills/code-change-verification/scripts/run.sh` — all commands passed:

```
make format     -> 847 files left unchanged; ruff check --fix: All checks passed!
make lint       -> passed in 0s
make tests      -> passed in 87s
make typecheck  -> passed in 1557s
```

The docs line was added after that run, so `uv run ruff format --check .` (847 files already formatted), `uv run ruff check .` (All checks passed), `uv run pytest tests/test_doc_parsing.py tests/extensions/memory/test_advanced_sqlite_session.py -q` (77 passed), and `make build-docs` (built successfully, no new warnings) were re-run afterwards. `git diff --check` is clean.

### Compatibility

Callers that previously reused a branch ID now get a `ValueError` instead of a silently corrupted branch. That path never produced a usable new branch, so no working behavior is lost. Generated IDs keep their existing `branch_from_turn_{turn}_{timestamp}` shape and only gain a `_2`, `_3`, ... suffix when that exact ID is already taken. The stored schema, the copy semantics, and every public signature are unchanged.

### Non-goals

- No change to which rows are copied or to the exclusive-turn semantics of `from_turn_number`.
- No validation of empty or whitespace branch names — unrelated to the demonstrated defect.
- No change to `switch_to_branch()`, `delete_branch()`, or `clear_session()`.

### Limitations

- The `ValueError` covers a branch that already holds messages. Branching from turn 1 copies no rows, so such a branch leaves no `message_structure` row and is not "existing" by the definition the rest of the class already uses; that behavior is unchanged and is covered by `test_create_branch_from_the_first_turn_still_succeeds`.
- The check is atomic per connection lock, which is the same scope every other write in this class relies on. Cross-process access to the same SQLite file is outside that scope, as it already was.
- Only `docs/sessions/advanced_sqlite_session.md` was updated; the generated `docs/ja`, `docs/ko`, and `docs/zh` pages were intentionally left alone.

### Issue number

Closes #4150

### Checks

### DIFF
--- a/docs/sessions/advanced_sqlite_session.md
+++ b/docs/sessions/advanced_sqlite_session.md
@@ -161,6 +161,8 @@ branch_id = await session.create_branch_from_content(
 )
 ```
 
+Branch IDs must be unique. Passing a `branch_name` that an existing branch already uses raises `ValueError`; use [`switch_to_branch()`][agents.extensions.memory.advanced_sqlite_session.AdvancedSQLiteSession.switch_to_branch] to continue an existing branch instead. Omitting `branch_name` always produces an unused ID.
+
 ### Branch management
 
 ```python
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -4,6 +4,7 @@ import asyncio
 import json
 import logging
 import sqlite3
+import time
 from contextlib import closing
 from pathlib import Path
 from typing import Any, cast
@@ -812,16 +813,16 @@ class AdvancedSQLiteSession(SQLiteSession):
 
         Args:
             turn_number: The branch turn number of the user message to branch from
-            branch_name: Optional name for the branch (auto-generated if None)
+            branch_name: Optional name for the branch. Must not name an existing branch.
+                Auto-generated as a unique id if None.
 
         Returns:
             The branch_id of the newly created branch
 
         Raises:
-            ValueError: If turn doesn't exist or doesn't contain a user message
+            ValueError: If turn doesn't exist, doesn't contain a user message, or
+                `branch_name` names a branch that already exists
         """
-        import time
-
         # Capture the generation before any DB work so a clear that commits
         # while this branch is being created cannot be overwritten by the
         # pointer update below.
@@ -859,13 +860,11 @@ class AdvancedSQLiteSession(SQLiteSession):
 
         turn_content = await asyncio.to_thread(_validate_turn)
 
-        # Generate branch name if not provided
-        if branch_name is None:
-            timestamp = int(time.time())
-            branch_name = f"branch_from_turn_{turn_number}_{timestamp}"
-
-        # Copy messages before the branch point to the new branch
-        await self._copy_messages_to_new_branch(branch_name, turn_number)
+        # Copy messages before the branch point to the new branch. The target branch id is
+        # resolved inside that call, under the same connection lock that performs the copy,
+        # so a generated id cannot collide with an existing branch and an explicit id cannot
+        # silently append the copied history to one.
+        branch_name = await self._copy_messages_to_new_branch(branch_name, turn_number)
 
         # Switch to new branch under the lock; skipped if a clear_session has
         # committed since `generation` was captured (its reset to 'main' wins),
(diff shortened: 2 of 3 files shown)


seratch (changes requested):
Please resolve all the review feedback from codex first.

seratch:
Thanks again for working on this issue here. Your contribution will be included in https://github.com/openai/openai-agents-python/pull/4186

PR #4155, opened 2026-08-03
fix(models): pair Anthropic thinking signatures with their own block

### LINKED ISSUE #4154: Anthropic thinking signatures are replayed on the wrong thinking block
### Please read this first

- **Have you read the docs?** Yes — [Models](https://openai.github.io/openai-agents-python/models/) and the LiteLLM extension docs.
- **Have you searched for related issues?** Yes. Searched open and closed issues and PRs for `thinking_blocks`, `encrypted_content`, `signature thinking`, and `redacted_thinking`. The closest prior work is #1704, which introduced this replay path, and #1797 / #1935, which concern block ordering rather than signature pairing. Nothing covers signature-to-block pairing.

### Describe the bug

`Converter.message_to_output_items()` stores Anthropic thinking blocks as two parallel lists on the reasoning item — the thinking text in `content`, and the signatures newline-joined in `encrypted_content`. `Converter.items_to_messages()` later zips them back together by position to rebuild the `thinking` blocks sent to Claude.

The two lists are built by two independent filters:

```python
# src/agents/models/chatcmpl_converter.py
thinking_text = block.get("thinking", "")
if thinking_text:
    reasoning_item.content.append(Content(text=thinking_text, type="reasoning_text"))
# Store the signature if present
if signature := block.get("signature"):
    signatures.append(signature)
```

A block appends to `content` only when it has thinking text, and to `signatures` only when it has a signature. The positional correspondence therefore holds only when *every* block has both. Any block that satisfies one condition but not the other desyncs the lists, and the reconstruction side (`signatures.pop(0)`) shifts every later signature onto the preceding thinking block, leaving the last block unsigned.

This is reachable two ways:

1. **A block with thinking text but no signature.** LiteLLM types the field as optional — `AnthropicResponseContentBlockThinking.signature: Optional[str]` in `litellm/types/llms/anthropic.py`.
2. **A block with a signature but no thinking text.** Its signature is recorded while it contributes no content, shifting every subsequent pairing by one.

Anthropic verifies a thinking signature against the exact text it was issued for, so a shifted signature is not a cosmetic problem — the next request carries a thinking block whose signature belongs to different text.

### Debug information

- Agents SDK version: `0.19.2` (reproduced on `main` at `c546ca12`)
- Related library versions: `litellm` 1.83.0
- Python version: 3.10.11
- Model and model provider: Anthropic Claude with extended thinking, via `LitellmModel`. Reproduced offline with a stub message so no API key or network access is needed.

No exception is raised by the SDK — the mispaired signature is sent to the provider.

### Repro steps

Self-contained; no API key, no network call, and no optional dependency required.

```python
from openai.types.chat import ChatCompletionMessage

from agents.models.chatcmpl_converter import Converter


class ThinkingMessage(ChatCompletionMessage):
    """Stands in for litellm's InternalChatCompletionMessage."""

    reasoning_content: str = ""
    thinking_blocks: list = []


message = ThinkingMessage(
    role="assistant",
    content="visible answer",
    reasoning_content="step one\nstep two",
    thinking_blocks=[
        {"type": "thinking", "thinking": "step one"},                        # unsigned
        {"type": "thinking", "thinking": "step two", "signature": "SIG-2"},  # signed
    ],
)

output_items = Converter.message_to_output_items(message)
replayed = Converter.items_to_messages(
    [item.model_dump() for item in output_items],
    model="anthropic/claude-4-opus",
    preserve_thinking_blocks=True,
)

for block in replayed[0]["content"]:
    if isinstance(block, dict) and block.get("type") == "thinking":
        print(f"{block['thinking']!r:<12} -> signature={block.get('signature')!r}")
```

### Actual behavior

```text
'step one'   -> signature='SIG-2'
'step two'   -> signature=None
```

`SIG-2` was issued for `"step two"` but is replayed on `"step one"`, and `"step two"` is sent unsigned.

### Expected behavior

```text
'step one'   -> signature=None
'step two'   -> signature='SIG-2'
```

A signature should be replayed on the thinking block it was issued for, and a block without one should stay unsigned.

### Root-cause hypothesis

*(hypothesis)* The `content` and `signatures` lists in `message_to_output_items()` are appended under separate conditions, so their indices only line up when every block is both non-empty and signed. Recording a signature slot for every content entry — an empty string when the block has none — keeps the pairing positional, and reconstructing by index rather than `pop(0)` leaves an unsigned block unsigned instead of consuming a later block's signature.

Note that the streaming path in `chatcmpl_stream_handler.py` is unaffected: it accumulates one concatenated thinking string and one signature, so its pairing is trivially aligned. This makes the two paths disagree today.

### Proposed scope

Keep the two lists in lockstep in `message_to_output_items()` and pair by position in `items_to_messages()`. No public API change and no schema change: fully signed responses — the overwhelmingly common case — serialize byte-identically to today, and history written by earlier versions still pairs correctly.

abhay-codes07:
### Summary

Anthropic thinking blocks are stored on the reasoning item as two parallel lists — the thinking text in `content`, the signatures newline-joined in `encrypted_content` — and `items_to_messages()` zips them back together by position when replaying the conversation to Claude.

`message_to_output_items()` built those lists with two independent filters: `content` was appended only for blocks with thinking text, and `signatures` only for blocks that had a signature. The positional correspondence therefore held only when *every* block had both. A block with text but no signature (LiteLLM types it as `Optional[str]`), or a signature but no text, desynced the lists — `signatures.pop(0)` then reattached every later signature to the preceding thinking block and left the last one unsigned.

Anthropic verifies a signature against the exact thinking text it was issued for, so the next request carries a block signed for different text.

```
before:  'step one' -> signature='SIG-2'      # SIG-2 was issued for 'step two'
         'step two' -> signature=None

after:   'step one' -> signature=None
         'step two' -> signature='SIG-2'
```

The fix records a signature slot for every content entry, using an empty string when the block has none, and pairs by index instead of `pop(0)` when reconstructing. A block with a signature but no thinking text now takes no slot, since there is no text to replay it against.

Scope notes:

- **No public API or schema change.** Fully signed responses — the common case, and what every existing test covers — serialize byte-identically to today.
- **Backward compatible.** History written by earlier versions, where the signature count already matches the `reasoning_text` count, pairs exactly as before; a regression test pins this.
- The streaming path in `chatcmpl_stream_handler.py` needed no change: it accumulates one concatenated thinking string and one signature, so its pairing was already aligned. This change makes the non-streaming path agree with it.

### Test plan

Six regression tests added to `tests/models/test_anthropic_thinking_blocks.py`, covering unsigned-then-signed, signed-then-unsigned, signature-without-text, a `redacted_thinking` block interleaved with a signed one, a fully unsigned response, and legacy `encrypted_content` pairing.

Two of them fail on `main` and pass with the fix; the other four are guards that hold both before and after:

```
FAILED tests/models/test_anthropic_thinking_blocks.py::test_unsigned_thinking_block_does_not_steal_a_later_signature
FAILED tests/models/test_anthropic_thinking_blocks.py::test_signature_without_thinking_text_does_not_shift_later_signatures
2 failed, 10 passed
```

Verification run from the repository root:

| Command | Result |
| --- | --- |
| `make format` | clean, 847 files unchanged |
| `uv run pytest tests/models/` | 632 passed |
| `uv run pytest tests/models/test_anthropic_thinking_blocks.py` | 12 passed |
| `make tests` | 5543 passed |

The full-suite run was done on Windows, where a set of sandbox symlink tests and the `TestSendEventAndConfig` realtime class fail and flake independently of this change. I diffed the failing set against a clean `main` checkout in the same environment: for the affected test families the two sets are identical, and the only whole-suite difference is one realtime interrupt test that alternates between 3, 1 and 3 failures across identical consecutive runs on `main` alone. Nothing in the diff touches realtime, tracing, or sandbox code.

### Issue number

Closes #4154

### Checks

The verification script is a bash script that shells out to `make`; I ran the underlying `make format`, `make lint`, `make mypy`, `make pyright` and `make tests` steps individually instead, with the results above.

---

@seratch one design question if you have a moment. I kept the empty-string placeholder so `encrypted_content` stays a plain newline-joined string and old history keeps parsing, which means a partially signed response now serializes as e.g. `"\nSIG-2"`. The alternative is to keep the signature list dense and store the pairing explicitly (an index map, or a JSON payload) with a version marker. That is cleaner to read but changes the stored shape, so I went with the narrower option. Happy to switch if you would rather fix the format properly while this path is still Chat Completions-only.

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -149,7 +149,12 @@ class Converter:
 
             # Store thinking blocks for Anthropic compatibility
             if hasattr(message, "thinking_blocks") and message.thinking_blocks:
-                # Store thinking text in content and signature in encrypted_content
+                # Store thinking text in content and signature in encrypted_content.
+                # The two lists are replayed positionally, so a signature is recorded for
+                # every content entry -- an empty string when the block has none. Appending
+                # only the signatures that exist would shift each later signature onto the
+                # preceding thinking block, and Anthropic verifies a signature against the
+                # thinking text it was issued for.
                 reasoning_item.content = []
                 signatures: list[str] = []
                 for block in message.thinking_blocks:
@@ -159,12 +164,11 @@ class Converter:
                             reasoning_item.content.append(
                                 Content(text=thinking_text, type="reasoning_text")
                             )
-                        # Store the signature if present
-                        if signature := block.get("signature"):
-                            signatures.append(signature)
+                            # Store the signature if present, keeping the slot either way.
+                            signatures.append(block.get("signature") or "")
 
-                # Store the signatures in encrypted_content with newline delimiter
-                if signatures:
+                # Store the signatures in encrypted_content with newline delimiter.
+                if any(signatures):
                     reasoning_item.encrypted_content = "\n".join(signatures)
 
             items.append(reasoning_item)
@@ -820,8 +824,12 @@ class Converter:
                 ):
                     signatures = encrypted_content.split("\n") if encrypted_content else []
 
-                    # Reconstruct thinking blocks from content and signature
+                    # Reconstruct thinking blocks from content and signature. Signatures are
+                    # matched to reasoning text by position, so a missing trailing entry or an
+                    # empty placeholder leaves that block unsigned instead of consuming the
+                    # signature that belongs to a later block.
                     reconstructed_thinking_blocks = []
+                    reasoning_text_index = 0
                     for content_item in content_items:
                         if (
                             isinstance(content_item, dict)
@@ -831,9 +839,15 @@ class Converter:
                                 "type": "thinking",
                                 "thinking": content_item.get("text", ""),
                             }
+                            signature = (
+                                signatures[reasoning_text_index]
+                                if reasoning_text_index < len(signatures)
+                                else ""
+                            )
+                            reasoning_text_index += 1
                             # Add signatures if available
-                            if signatures:
-                                thinking_block["signature"] = signatures.pop(0)
+                            if signature:
+                                thinking_block["signature"] = signature
                             reconstructed_thinking_blocks.append(thinking_block)
 
                     # Store thinking blocks as pending for the next assistant message
--- a/tests/models/test_anthropic_thinking_blocks.py
+++ b/tests/models/test_anthropic_thinking_blocks.py
@@ -464,3 +464,172 @@ def test_thinking_blocks_do_not_leak_across_an_intervening_user_turn():
     assistant_messages = [msg for msg in messages if msg.get("role") == "assistant"]
     assert len(assistant_messages) == 1
     assert assistant_messages[0].get("content") == "an answer"
+
+
+def _replayed_thinking_blocks(
+    thinking_blocks: list[dict[str, Any]],
+) -> list[dict[str, Any]]:
+    """Round-trip thinking blocks through the converter and return what Claude would receive."""
+    message = InternalChatCompletionMessage(
+        role="assistant",
+        content="visible answer",
+        reasoning_content="reasoning",
+        thinking_blocks=thinking_blocks,
+    )
+    output_items = Converter.message_to_output_items(message)
+    items_as_dicts: list[dict[str, Any]] = [item.model_dump() for item in output_items]
+    messages = Converter.items_to_messages(
+        items_as_dicts,  # type: ignore[arg-type]
+        model="anthropic/claude-4-opus",
+        preserve_thinking_blocks=True,
+    )
+
+    assistant_messages = [msg for msg in messages if msg.get("role") == "assistant"]
+    assert len(assistant_messages) == 1
+    content = assistant_messages[0].get("content")
+    assert isinstance(content, list)
+    blocks = cast(list[dict[str, Any]], content)
+    return [block for block in blocks if block.get("type") == "thinking"]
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
The new `signature_without_thinking_text` test encodes the wrong Anthropic contract. A block with `thinking: ""` and a signature is the documented `display: "omitted"` response shape, not an orphan, and it must be replayed unchanged during tool-use continuation.

I reproduced this through the locked LiteLLM 1.83.0 runtime path. LiteLLM preserves the empty signed block on inbound conversion, but this PR drops it; when it is the only thinking block, `message_to_output_items` creates no reasoning item at all. The parallel `content` / `encrypted_content` representation also cannot preserve `redacted_thinking`.

To improve, it needs to preserve the complete `thinking_blocks` list as one source of truth and replay it through LiteLLM's native assistant `thinking_blocks` field, while retaining legacy decoding for already stored reasoning items. That native field round-trips both empty signed blocks and redacted blocks unchanged in LiteLLM's actual Anthropic request transform. Please also replace this unit expectation with an end-to-end raw Anthropic response-to-outbound-request test. Streaming has the same lossy behavior and can be tracked separately if it remains outside this PR.

seratch:
While reviewing this PR, I found a few more things to fix, so I came up with https://github.com/openai/openai-agents-python/pull/4157 including your contribution credit in it. Thanks again for reporting the issue.

PR #4159, opened 2026-08-04
fix(models): surface content-filter refusals in non-streaming Chat Completions

LeSingh1:
### Summary

`OpenAIChatCompletionsModel.get_response` never inspected `finish_reason`. Some providers — Azure OpenAI content filtering, Bedrock-backed Chat Completions proxies — signal a safety block only via `finish_reason == "content_filter"` with an empty message and no `refusal` field. `Converter.message_to_output_items` then produces zero output items, so the caller sees an indistinguishable "empty turn" and agent loops retry fruitlessly against a block that will never clear.

Both sibling paths already handle this. The streamed handler synthesizes a `ResponseOutputRefusal` (`chatcmpl_stream_handler.py`), and so does `LitellmModel.get_response` (#3769). This brings the non-streaming OpenAI Chat Completions path into agreement with them by setting `message.refusal` before conversion, and only when the turn is genuinely empty — no content, no provider-supplied refusal, and no tool calls — so no existing successful shape changes.

### Test plan

New `tests/models/test_openai_chatcompletions_content_filter.py`, mirroring the existing `test_litellm_content_filter.py`:

- a refusal is synthesized for an empty content-filtered turn
- real content, a provider-supplied refusal, and tool calls are each left untouched
- `finish_reason="stop"` is unaffected in both the text and empty cases

Fails on `main` (`AssertionError: expected a refusal item, got: []`), passes with the fix. Full `tests/models/` suite: 646 passed. `.agents/skills/code-change-verification/scripts/run.sh` passed.

### Issue number

N/A — found while auditing provider parity between the streaming and non-streaming Chat Completions paths.

### Checks

### DIFF
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -306,6 +306,23 @@ class OpenAIChatCompletionsModel(Model):
                 "output_tokens_details": usage.output_tokens_details.model_dump(),
             }
 
+            # Surface content-filter refusals explicitly. Some providers (for example Azure
+            # OpenAI content filtering, or Bedrock-backed Chat Completions proxies) signal a
+            # safety block only via ``finish_reason == "content_filter"`` with an empty message
+            # and no ``refusal`` field. Without this the message converts to zero output items
+            # and the caller sees an indistinguishable "empty turn", which drives agent loops
+            # into fruitless retries. The streamed path already synthesizes the same refusal,
+            # so keep both paths aligned.
+            if (
+                message is not None
+                and first_choice is not None
+                and first_choice.finish_reason == "content_filter"
+                and not message.content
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The behavior and fix are valid, but the impact description is slightly too broad: an empty content-filtered response does not always drive the agent loop into retries. For plain-text agents, the current runner completes successfully with an empty string; structured-output agents retry until the turn limit.

Please update the source comment, test docstring, and PR description to state that the lost refusal signal produces an empty final output for plain-text runs and repeated retries for structured-output runs. No runtime implementation change is needed.

seratch:
https://github.com/openai/openai-agents-python/pull/4188 includes your contribution; thanks again for your efforts here.

PR #4161, opened 2026-08-04
fix(realtime): release the websocket when the initial session config fails

LeSingh1:
### Summary

`OpenAIRealtimeWebSocketModel.connect()` opens the websocket and starts the listener task, then sends the initial session config. If that send fails, neither the socket nor the task is released.

This is reachable from ordinary agent configuration, not just transport errors: `_tools_to_session_tools` raises `UserError("Tool ... is unsupported. Must be a function tool.")` for any non-`FunctionTool`, and `RealtimeSession._get_updated_model_settings_from_agent` does not reject hosted tools beforehand. So a realtime agent configured with, say, a `WebSearchTool` leaks a live connection. `ensure_tool_choice_supports_backend` and a `ConnectionClosed` on send reach the same point.

The caller cannot clean up after it either. `RealtimeSession.__aenter__` only calls `remove_listener`, and Python does not invoke `__aexit__` when `__aenter__` raises, so the usual `async with await runner.run() as session:` form gives no cleanup hook. On top of the leaked socket and background task, `connect()` opens with `assert self._websocket is None`, so the model instance can never be reconnected.

Fix: wrap the initial `_update_session_config` call and `await self.close()` before re-raising, restoring the same clean state that `test_connect_websocket_failure_propagates` already asserts for a failure one step earlier. `.agents/references/realtime-session-lifecycle.md` states this ownership rule; this is the untested half of it.

### Test plan

Added `test_connect_session_config_failure_releases_websocket`: connects with a `WebSearchTool` in `initial_model_settings`, asserts the `UserError` propagates, and asserts `_websocket is None`, `_websocket_task is None`, and that `close()` was awaited on the socket.

Fails on `main` (`assert <AsyncMock> is None`), passes with the fix. Full `tests/realtime/` suite: 378 passed. `.agents/skills/code-change-verification/scripts/run.sh` passed.

### Issue number

N/A — found while auditing connection ownership on realtime failure paths.

### Checks

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -588,7 +588,14 @@ class OpenAIRealtimeWebSocketModel(RealtimeModel):
             transport_config=self._transport_config,
         )
         self._websocket_task = asyncio.create_task(self._listen_for_messages())
-        await self._update_session_config(model_settings)
+        try:
+            await self._update_session_config(model_settings)
+        except BaseException:
+            # The websocket and its listener task are owned by this method, so release them
+            # when the initial session config fails. Otherwise the connection stays open and
+            # the model can never be reconnected because of the asserts above.
+            await self.close()
+            raise
 
     async def _create_websocket_connection(
         self,
--- a/tests/realtime/test_openai_realtime.py
+++ b/tests/realtime/test_openai_realtime.py
@@ -10,7 +10,7 @@ import pytest
 import websockets
 from pydantic import TypeAdapter
 
-from agents import Agent, function_tool
+from agents import Agent, WebSearchTool, function_tool
 from agents.exceptions import UserError
 from agents.handoffs import handoff
 from agents.realtime.model import RealtimeModelConfig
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
Thanks for the focused fix. The leak is real, and cleaning up when the initial session update fails is the correct direction.

Before we merge this, please make the existing single-connection invariant effective across the whole connection attempt. Two `connect()` calls can currently pass the assertions before either assigns `_websocket`; if A's initial send is pending, B installs newer connection state, and A then fails, A's unconditional `self.close()` can close B's socket and listener task. Please mark an attempt active before the first `await`, reject a second attempt before it acquires resources, and clear that guard on every failure path.

Please also add a controlled interleaving test that pauses A's initial send, verifies B is rejected without opening another socket, then fails A and confirms its original exception and resource cleanup.

seratch:
Closing this PR in favor of https://github.com/openai/openai-agents-python/pull/4189; your contribution is included in the commit.

PR #4162, opened 2026-08-04
fix(memory): skip conversation creation on empty add_items

LeSingh1:
### Summary

`OpenAIConversationsSession.add_items` called `_get_session_id()` before its `if not items: return` guard. On an uninitialized session that made a no-op write provision a real remote conversation through `conversations.create`, leaving a stray conversation on the OpenAI platform and a session object that now reports as initialized — all as a side effect of writing nothing.

This is the same ordering bug #4111 fixed for `clear_session`. Moving the empty-list guard ahead of session-ID resolution means nothing touches the network when there is nothing to write.

### Test plan

Added `test_add_items_empty_does_not_create_session` and `test_add_items_empty_keeps_existing_session_id` to `tests/memory/test_openai_conversations_session.py`. The first fails on `main` with `Expected 'create' to not have been called. Called 1 times.` and passes with the change. Full file: 35 passed. `.agents/skills/code-change-verification/scripts/run.sh` passed.

### Issue number

N/A — found while auditing session backends for network side effects on no-op writes.

### Checks

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -107,10 +107,10 @@ class OpenAIConversationsSession(SessionABC):
         return all_items  # type: ignore
 
     async def add_items(self, items: list[TResponseInputItem]) -> None:
-        session_id = await self._get_session_id()
         if not items:
             return
 
+        session_id = await self._get_session_id()
         await self._openai_client.conversations.items.create(
             conversation_id=session_id,
             items=items,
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
The empty-list guard is at the right layer: `add_items([])` should not create remote state, and moving the guard before session-ID resolution makes this backend consistent with the other session implementations.

Before merging, please update the `session_id` docstring and `ValueError` guidance. They currently say that calling `add_items()` triggers lazy initialization, but this PR intentionally leaves the session uninitialized after `add_items([])`. Please also assert that public behavior in the regression test by checking that `session.session_id` still raises after the empty add, rather than relying only on the private `_session_id` field. No broader session refactor is needed.

seratch:
https://github.com/openai/openai-agents-python/pull/4190 includes your contribution credit in the commit. Thanks again for sharing this.

PR #4164, opened 2026-08-04
fix(extensions): serialize structured tool outputs as JSON when trimming

LeSingh1:
### Summary

A `function_call_output` may carry a list of content parts rather than a string — the SDK produces that shape itself whenever a tool returns `ToolOutputText`, `ToolOutputImage`, or `ToolOutputFileContent` (`ResponseFunctionCallOutputItemListParam`, via `_convert_tool_output_as_structured` in `items.py`). Those items land in session history and flow straight into the trimmer.

`ToolOutputTrimmer._trim_function_call_output` stringified them with `str()`, which caused two problems:

1. The preview handed back to the model was a Python repr (`[{'type': 'input_text', ...}]`) — single-quoted and not valid JSON.
2. The `max_output_chars` check measured the repr instead of the payload. Because repr leaves embedded double quotes unescaped, an output that exceeds the threshold on the wire can measure under it and escape trimming entirely — in the added test, 196 JSON chars measured as 156 against a 170 limit.

Reuse `_serialize_json_like()`, which the sibling `tool_search_output` path already uses and which falls back to `str()` for values JSON cannot encode, so behavior is preserved exactly for non-serializable values. The `isinstance(output, str)` branch is untouched, so all existing string outputs behave identically.

### Test plan

Added `test_trims_structured_output_as_json` (exact-match assertion on the JSON preview) and `test_structured_output_threshold_uses_serialized_size` (the mis-sizing case). Both fail on `main` — the first on a repr-vs-JSON diff, the second with `AttributeError: 'list' object has no attribute 'startswith'` because the item was never trimmed — and pass with the change. Existing string-output tests are unchanged.

Full file: 43 passed. `.agents/skills/code-change-verification/scripts/run.sh` passed.

### Issue number

N/A — found while auditing serialization consistency within `ToolOutputTrimmer`.

### Checks

### DIFF
--- a/src/agents/extensions/tool_output_trimmer.py
+++ b/src/agents/extensions/tool_output_trimmer.py
@@ -223,7 +223,10 @@ class ToolOutputTrimmer:
     ) -> tuple[dict[str, Any] | None, int]:
         """Trim a function_call_output item when its serialized output is too large."""
         output = item.get("output", "")
-        output_str = output if isinstance(output, str) else str(output)
+        # A function_call_output may carry a list of content parts instead of a string.
+        # Serialize those as JSON so the preview stays readable to the model and the size
+        # check measures the payload rather than its Python repr.
+        output_str = output if isinstance(output, str) else self._serialize_json_like(output)
         output_len = len(output_str)
         if output_len <= self.max_output_chars:
             return None, 0
--- a/tests/extensions/test_tool_output_trimmer.py
+++ b/tests/extensions/test_tool_output_trimmer.py
@@ -203,6 +203,51 @@ class TestTrimming:
         assert "1000 chars" in trimmed
         assert len(trimmed) < len(large)
 
+    def test_trims_structured_output_as_json(self) -> None:
+        """List-shaped tool outputs should preview as JSON, not as a Python repr."""
+        structured = [{"type": "input_text", "text": "x" * 400}]
+        items = [
+            _user("q1"),
+            _func_call("c1", "search"),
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for the contribution. I traced this through the SDK-generated structured tool output path and `ToolOutputTrimmer`. The tests show that `json.dumps()` changes the representation and threshold result, but they do not establish that JSON wire length is the correct meaning of `max_output_chars`. Structured outputs contain typed text, image, and file parts, so transport escaping and metadata are not a uniform proxy for model context. The preview is also still not valid JSON because it is sliced and followed by `...`.

I am going to close this PR for now. If there is a concrete supported scenario where the current behavior materially increases context usage or harms the model result, we can revisit it with content-type-aware sizing for `input_text` and an explicit policy for image and file parts.

LeSingh1:
That's fair, and the point about the preview is a good catch — it gets sliced and suffixed with `...`, so it was never going to be valid JSON either way. The readability half of my rationale doesn't hold up.

Agreed too that JSON wire length isn't self-evidently the right meaning for `max_output_chars`. I was treating `_serialize_json_like` on the `tool_search_output` path as the established convention and making the sibling path match it, but consistency between the two isn't an argument that either is measuring the right thing for structured parts — escaping and metadata aren't uniform across `input_text`, image, and file parts, as you say.

No need to revisit on my account without evidence. If I hit a concrete case where an image or file part pushes a turn over the threshold in a way that measurably changes the model result, I'll bring the repro rather than the diff.

Thanks for tracing it through properly rather than just declining it.

PR #4165, opened 2026-08-04
test: make timing-sensitive tool and guardrail tests deterministic

LeSingh1:
### Summary

Eight async tests decided their assertions by racing two `asyncio.sleep` durations against each other, or against a bounded wait inside the runtime. The cushions were 20-50ms, so on a loaded machine the scheduler — not the code under test — picked the winner.

The runtime genuinely waits `_FUNCTION_TOOL_POST_INVOKE_WAIT_SECONDS = 0.1` for post-invoke siblings before propagating a tool failure, and up to `_FUNCTION_TOOL_CANCELLED_DRAIN_SECONDS = 0.25` for cancelled ones. Tests raced 0.05s and 0.3s sleeps against those budgets and guarded the calls with 0.1-0.2s timeouts, which is below the runtime's own worst case.

Under a 24-process CPU load with `pytest -n 8`, these flipped at roughly 5-20% per file run on `main`. This PR replaces each race with explicit happens-before synchronization:

- **Post-invoke drain tests** gate the failing sibling on an event set by the guardrail, so the guardrail is provably in flight when the error propagates. `...do_not_wait_indefinitely...` now blocks its guardrail on an event the test releases only after the error surfaced, so the sibling outlasts any finite drain budget by construction rather than by a 0.3s sleep. Its counterpart widens the drain budget so a scheduling hiccup cannot decide which failure wins.
- **Parallel input guardrails** trip on an event set by the tool they must not block. A regression that gates the turn now deadlocks and fails loudly instead of quietly losing a 50ms race.
- **Guardrails and model calls that exist to be cancelled** block forever instead of sleeping, so only cancellation can end them. This also removes the `cancellation_delay < 0.2` timestamp assertions that were standing in for that guarantee.

Timeouts kept in these tests are hang guards on causally guaranteed conditions, not margins. Two guards shorter than the runtime's 0.25s cancelled-sibling drain are widened for the same reason: they can fire on a healthy run.

No product code changes. `test_mixed_blocking_and_parallel_guardrails` and its streaming twin are intentionally untouched — they are handled in #4158. The two branches touch disjoint regions of `tests/test_guardrails.py` and merge cleanly in either order.

### Test plan

- Reproduced on `main` with 24 CPU-burner processes and `uv run pytest <file> -q -n 8`: `tests/test_run_step_execution.py` failed 2/35 runs; `tests/test_guardrails.py` failed 3/15 runs across 6 distinct tests.
- Same harness on this branch, 25 runs of both files: zero failures among the changed tests.
- Each changed test was verified to still catch regressions by temporarily removing the behavior it covers and confirming the test fails — the post-invoke drain, its bound, parallel-guardrail non-blocking, sibling cancellation on tripwire, sibling cancellation on non-tripwire failure, late-cleanup reporting, and late-failure merging. All reverted afterwards.
- `.agents/skills/code-change-verification/scripts/run.sh` → `code-change-verification: all commands passed.`

A few tests with similar shapes were examined and deliberately left alone because their margins are causally enforced or are themselves the thing under test — for example the sequential blocking-guardrail ordering asserts (zero margin by construction), and the `elapsed < 0.3` latency budget in `test_cancel_streaming.py`, where the discriminator is 0.1s vs 0.5s.

### Issue number

N/A — found while investigating CI flakiness.

### Checks

### DIFF
--- a/tests/test_guardrails.py
+++ b/tests/test_guardrails.py
@@ -493,8 +493,12 @@ async def test_parallel_guardrail_may_not_prevent_tool_execution():
     guardrail_executed = False
     tool_executed = asyncio.Event()
 
+    # Async so the tool body runs on the event-loop thread. A sync tool is dispatched
+    # through asyncio.to_thread, and asyncio.Event is not thread-safe, so setting it
+    # from the worker thread would be the new race. This test covers guardrail/tool
+    # ordering, not sync-tool offloading.
     @function_tool
-    def fast_tool() -> str:
+    async def fast_tool() -> str:
         nonlocal tool_was_executed
         tool_was_executed = True
         tool_executed.set()
@@ -850,8 +854,12 @@ async def test_parallel_guardrail_may_not_prevent_tool_execution_streaming():
     guardrail_executed = False
     tool_executed = asyncio.Event()
 
+    # Async so the tool body runs on the event-loop thread. A sync tool is dispatched
+    # through asyncio.to_thread, and asyncio.Event is not thread-safe, so setting it
+    # from the worker thread would be the new race. This test covers guardrail/tool
+    # ordering, not sync-tool offloading.
     @function_tool
-    def fast_tool() -> str:
+    async def fast_tool() -> str:
         nonlocal tool_was_executed
         tool_was_executed = True
         tool_executed.set()
--- a/tests/test_guardrails.py
+++ b/tests/test_guardrails.py
@@ -491,11 +491,13 @@ async def test_blocking_guardrail_prevents_agent_execution_streaming():
 async def test_parallel_guardrail_may_not_prevent_tool_execution():
     tool_was_executed = False
     guardrail_executed = False
+    tool_executed = asyncio.Event()
 
     @function_tool
     def fast_tool() -> str:
         nonlocal tool_was_executed
         tool_was_executed = True
+        tool_executed.set()
         return "tool_executed"
 
     @input_guardrail(run_in_parallel=True)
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
In both `test_parallel_guardrail_may_not_prevent_tool_execution` variants, `fast_tool` is synchronous, so the SDK invokes it through `asyncio.to_thread()`. Calling `tool_executed.set()` there mutates an `asyncio.Event` from a worker thread while the guardrail is awaiting it. Under asyncio debug mode this can raise `RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one`.

Please make `fast_tool` async in both variants so the event is set on the event-loop thread. These tests validate guardrail/tool ordering rather than sync-tool offloading. After that change and green checks, this should be ready for another review.

LeSingh1:
Thanks — you're right, and this was a defect I introduced. Both `fast_tool` variants are now `async def`, so the event is set on the loop thread.

Worth recording the failure mode, because it's sharper than "may raise under debug mode": `Event.set()` resolves each waiter future *before* `loop.call_soon` raises, so the future ends up `FINISHED` with its callbacks never scheduled — and the surrounding `wait_for` then can't cancel it either, since `Future.cancel()` returns `False` on a finished future. With `PYTHONASYNCIODEBUG=1` both tests deadlock outright before this change, and pass after it.

I audited the whole PR rather than just the two flagged sites: an AST sweep of both files for every sync callable reachable as a `FunctionTool` (decorator form and `function_tool(fn, ...)` form), checked against asyncio-primitive contact. Exactly two instances, both the ones you identified. The other sync tools (`dangerous_tool` ×4, `lookup_secret`, `record_side_effect`, `synthetic_tool`, `_echo`) only assign plain `nonlocal` values that are read after the run completes. Neither file dispatches to a thread by any other route, and sync guardrails run inline on the loop thread.

Teeth re-proved per test with a targeted break: forcing `run_in_parallel=True` guardrails to gate the turn in `run.py` and `run_internal/run_loop.py` makes each variant fail on its own `wait_for`. Load re-run (24 CPU-burner processes, `-n 8`, 30 iterations of both files): no failures in any test this PR touches, before or after.

One thing I found while re-running and deliberately did **not** change, to keep this PR to what was asked: `test_mixed_blocking_and_parallel_guardrails` and its `_streaming` twin still decide `timestamps["model_called"] <= timestamps["parallel_end"]` by racing a 30ms sleep, and one of them flipped once in 30 loaded runs. They aren't touched by this PR. Happy to do them as a separate follow-up, or leave them.

seratch:
Closing this PR in favor of https://github.com/openai/openai-agents-python/pull/4192

PR #4166, opened 2026-08-04
test: make parent-cancellation tool tests deterministic

LeSingh1:
### Summary

Follow-up to #4165. Two parent-cancellation tests still decided pass/fail from wall-clock deadlines smaller than the runtime's own bounds, so a loaded machine could fail them even when the runtime behaved correctly.

`test_parent_cancellation_does_not_wait_for_tool_cleanup` asserted the parent settles within 100ms, but `_FUNCTION_TOOL_CANCELLED_DRAIN_SECONDS` alone allows the runtime 250ms — the deadline was below the worst case it was supposed to tolerate, and it carried the whole assertion. The ordering is now established with events: wait until the tool has parked inside its cleanup, then assert the parent surfaces cancellation while that cleanup is still blocked. The remaining timeouts only detect a hang.

`test_parent_cancellation_does_not_report_tool_failure_as_background_error` ended with two bare `asyncio.sleep(0)` calls and a negative assertion, so it could pass simply because nothing had run yet. It now waits for the tool to actually fail — proving the intended interleaving happened — and then drains the loop's callback queue, since the report is delivered from a done callback.

Tests only; no runtime behavior changes.

### Test plan

- `uv run pytest tests/test_run_step_execution.py -k parent_cancellation` passes.
- Teeth confirmed with a targeted break per test, reverted afterwards:
  - awaiting the pending tool tasks in the parent-cancellation branch of `execute()` makes `..._does_not_wait_for_tool_cleanup` fail with `TimeoutError`;
  - making `_parent_cancelled_task_exception_message` report `Exception` with the cancellation-cleanup message makes `..._does_not_report_tool_failure_as_background_error` fail immediately.
  Each break fails only its own test, so the two are independently pinned.
- 40 invocations of both tests under 32 competing CPU-bound processes: 2 failures before this change, 0 after.
- `bash .agents/skills/code-change-verification/scripts/run.sh` → all commands passed.

### Issue number

N/A — follow-up to #4165.

### Checks

### DIFF
--- a/tests/test_run_step_execution.py
+++ b/tests/test_run_step_execution.py
@@ -52,7 +52,7 @@ from agents import (
     trace,
 )
 from agents._public_agent import set_public_agent
-from agents.run_internal import run_loop, turn_resolution
+from agents.run_internal import run_loop, tool_execution, turn_resolution
 from agents.run_internal.agent_bindings import bind_execution_agent, bind_public_agent
 from agents.run_internal.run_loop import (
     NextStepFinalOutput,
@@ -98,15 +98,8 @@ from .utils.hitl import (
 # Deadlock detector for the parent-cancellation tests below. It is deliberately far larger
 # than any bound the runtime itself applies (see `_FUNCTION_TOOL_CANCELLED_DRAIN_SECONDS`)
 # so a slow machine can never turn it into a behavioral deadline; ordering is asserted with
-# events instead.
+# events and runtime instrumentation instead.
 _CANCELLATION_HANG_GUARD_SECONDS = 5.0
-
-
-async def _flush_loop_callbacks(loop: asyncio.AbstractEventLoop) -> None:
-    """Yield until every callback already queued on the loop has run."""
-    marker = loop.create_future()
-    loop.call_soon(marker.set_result, None)
-    await marker
 
 
 def _function_spans() -> list[dict[str, Any]]:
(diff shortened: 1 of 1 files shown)


seratch:
@LeSingh1 Can you resolve the requested changes before sending further PRs? Thanks for your understanding!

LeSingh1:
Understood, and apologies for the volume — I've stopped opening new PRs here and will work through the outstanding review feedback before sending anything else.

Working the Codex findings now. Two of them are correct and worth stating plainly:

On #4165, the `asyncio.Event` thread-safety point is a real defect in my own change. The synchronous `function_tool` runs via `asyncio.to_thread`, so `tool_executed.set()` fires on a worker thread while the guardrail awaits the same event on the loop thread. A PR whose entire purpose was removing flakiness introduced a different flakiness. I'll make those tools async or signal via `loop.call_soon_threadsafe`, and fix the streaming copy with the same pattern.

On this PR, both P2 comments land. The hang guard being the only check does mean a bounded cleanup drain that finishes under it would still pass, and the failure ordering doesn't guarantee the background done-callback path is the one being exercised. I'll pin both to state transitions rather than to timing.

I'll push fixes to the existing PRs and won't open anything new until you've had a chance to look.

LeSingh1:
Both findings land. Fixed in the follow-up commit.

**On the hang guard carrying the assertion:** I tried the state-transition option first and convinced myself it can't work here — the detached tool task's wakeup and the parent's wakeup are adjacent callbacks in the same loop round, so nothing causally observable separates "did not wait" from "waited briefly"; the only difference is elapsed time. So I took the instrumentation route instead. The test now records every call to `_settle_pending_function_tool_tasks` — the single bounded settle primitive that both `_drain_cancelled_function_tool_tasks` and `_wait_pending_function_tool_tasks_for_timeout` funnel through — and requires parent cancellation never to reach it. The parked cleanup and hang guard stay, so an unbounded wait is still caught.

Verified against your exact scenario: adding a bounded `_drain_cancelled_tasks` before the re-raise leaves the old test green and fails the new one.

A caveat I'd rather state than hide — an ad-hoc inline `asyncio.wait(..., timeout=...)` that bypasses the module's drain machinery would still get through. Catching that would mean instrumenting `loop.call_later`, which the test's own guards use.

**On the failure ordering:** the `ValueError` now comes out of the tool's cancellation cleanup, so it cannot exist before the parent detaches the task, and the test rendezvouses on `_consume_function_tool_task_result` — the callback that actually makes the reporting decision — rather than on "the tool raised". `_flush_loop_callbacks` is gone; the rendezvous replaces its only caller.

One correction for the record, since I checked before changing it. I couldn't reproduce the specific "reports detached `Exception`s and still passes" case: the old test caught that break 40/40 idle and 40/40 under load, because the done callback attached by `_cancel_pending_tasks_for_parent_cancellation` still fires on the outer task even when the invoke task finished early. The window you describe is real — if cancellation lands inside `_raise_failure_after_draining_siblings`, `pending_tasks` is empty and nothing is attached — it just never hit on my machine.

But the underlying point is right and, if anything, understated: the old test also passed when I removed the detached-task reporting policy *entirely*, because it only ever had a negative assertion and never proved the callback ran at all. The new version fails that too.

Load re-run (24 burners, `-n 8`, 30 iterations): neither parent-cancellation test failed in 60 loaded runs. Full verification stack green.

seratch:
Closing this PR in favor of https://github.com/openai/openai-agents-python/pull/4193

PR #4173, opened 2026-08-04
fix(trimmer): preview structured tool outputs from their text parts

tonydzi:
> written by mycroft, an AI agent at Palo Alto AI Research Lab, working with Anton Dziatkovskii. Every command and output below was run locally and pasted verbatim.

### Summary

`function_call_output.output` can be a list of content parts instead of a string, and the SDK produces that shape itself: `ItemHelpers._convert_tool_output_as_structured` emits `ResponseFunctionCallOutputItemListParam` whenever a tool returns `ToolOutputText`, `ToolOutputImage` or `ToolOutputFileContent`. Those items stay in session history and reach `ToolOutputTrimmer` on every later turn.

`_trim_function_call_output` stringified that list and sliced the result, so the preview was the serialization rather than the content. With the config from the module's own docstring (`max_output_chars=500, preview_chars=200`) and a tool that returns a chart plus its caption, the model receives this for the rest of the conversation:

```
[Trimmed: plot output — 3169 chars → 200 char preview]
[{'type': 'input_image', 'image_url': 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAQQQQQQQQQ…
```

The caption — the only part of that output a later turn can act on — is gone, and the whole preview budget is spent on a base64 fragment that is re-sent on every subsequent call. Part order decides the outcome: the same tool with the caption first leaks less base64 but still spends ~40 characters of the budget on `[{'type': 'input_text', 'text': '`.

This is the concrete case asked for when #4164 was closed — current behavior both harms the model result and spends context, and the fix is content-type-aware. That PR proposed JSON sizing; this one deliberately leaves sizing alone. The trim threshold stays `len(str(output))`, exactly as today, so the *when* of trimming does not change — only *what the model is shown*.

After:

```
[Trimmed: plot output — 3169 chars → 200 char preview; dropped 1 input_image]
Revenue chart: Q3 up 12% YoY, driven by EMEA....
```

Policy for non-text parts, stated explicitly since that was the open question: text parts (`input_text` / `output_text` / `text`) are concatenated and previewed; image and file parts are **dropped and named** (`dropped 1 input_image, 1 input_file`). Dropping matches what the current code already does — collapsing the list into one string removes them either way — and naming them tells the model that the tool did return an image, which the old preview could not. If you would rather keep image/file parts and trim only the text alongside them, that is a small change to the same method; happy to switch it.

Two details worth calling out:

- When an output has **no** text part at all, there is nothing to preview, so the summary omits the `→ N char preview` clause instead of promising a preview and appending a bare `...`.
- Part types are only checked after confirming they are strings. A history item can hold anything a caller once put there, and `{"type": []}` would otherwise raise `TypeError: unhashable type: 'list'` from the set membership test — a call-model-input filter should not be able to fail a run on odd history.

String outputs go through the untouched code path; every existing test passes unmodified.

### Test plan

Three new tests in `tests/extensions/test_tool_output_trimmer.py`, all of which fail on `main`:

- `test_structured_output_preview_keeps_text_and_names_dropped_parts` — image part first, caption second: asserts the caption survives, `base64,` does not appear, and the dropped image is named. On `main`: `AssertionError: assert 'base64,' not in '[Trimmed: p...QQQQQQQQQ...'`
- `test_structured_output_without_text_is_named_not_sliced` — image + file, no text: asserts the parts are named and no base64 leaks.
- `test_structured_output_with_malformed_parts_does_not_raise` — unhashable `type`, a bare string part and one usable text part: on `main` this raises `TypeError` once the guard is removed.

Plus `test_small_structured_output_is_preserved` as a regression guard (passes both before and after).

```
uv run pytest tests/extensions/test_tool_output_trimmer.py -q   →  45 passed
uv run pytest tests/extensions tests/test_call_model_input_filter.py \
              tests/test_call_model_input_filter_unit.py -q     →  1217 passed, 6 skipped
uv run ruff check / ruff format --check                          →  clean
uv run mypy src/agents/extensions/tool_output_trimmer.py         →  clean
.agents/skills/code-change-verification/scripts/run.sh           →  1 failed, 6349 passed
```

The one failure is `tests/sandbox/test_session_utils.py::test_read_path_probe_resolves_symlinks_before_classifying_missing`, and it is not mine. Control run of the same command on unmodified `19e364c` (`perf: speed up the test suite (#4171)`, no patch applied) on this machine:

```
FAILED tests/mcp/test_mcp_pagination_integration.py::test_stdio_server_auto_paginates_tools_and_prompts
FAILED tests/mcp/test_mcp_pagination_integration.py::test_agent_calls_tool_from_second_stdio_page
FAILED tests/sandbox/test_session_utils.py::test_read_path_probe_resolves_symlinks_before_classifying_missing
3 failed, 6343 passed, 4 skipped in 104.51s
```

Same symlink test, plus two more, with none of my code loaded — a local macOS environment effect under `-n auto`, not a regression. Both sandbox symlink tests pass on their own with the patch applied (`1 passed` each), and the checkbox below stays unchecked because the full stack did not come back green here.

Credit to @LeSingh1, whose #4164 identified this code path; this PR answers the evidence question that closed it rather than reopening the sizing argument.

### Issue number

N/A — follow-up to the closed #4164.

### Checks

- [ ] I've confirmed all verification steps pass — one pre-existing local failure, reproduced on unmodified `main`, documented above

### DIFF
--- a/src/agents/extensions/tool_output_trimmer.py
+++ b/src/agents/extensions/tool_output_trimmer.py
@@ -62,6 +62,11 @@ _NAME_KEYED_SCHEMA_MAPS = frozenset(
 # schema keyword, so they are copied through untouched.
 _DATA_SCHEMA_KEYWORDS = frozenset({"default", "const", "enum"})
 
+# Content parts of a structured tool output whose payload is model-readable text. Every other
+# part (image, file) carries an opaque payload — a base64 data URL or a file reference — that
+# says nothing about what the tool returned when it is sliced by character count.
+_TEXT_CONTENT_PART_TYPES = frozenset({"input_text", "output_text", "text"})
+
 
 @dataclass
 class ToolOutputTrimmer:
@@ -223,6 +228,9 @@ class ToolOutputTrimmer:
     ) -> tuple[dict[str, Any] | None, int]:
         """Trim a function_call_output item when its serialized output is too large."""
         output = item.get("output", "")
+        if isinstance(output, list):
+            return self._trim_structured_function_call_output(item, output, tool_names)
+
         output_str = output if isinstance(output, str) else str(output)
         output_len = len(output_str)
         if output_len <= self.max_output_chars:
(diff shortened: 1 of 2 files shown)


luisangelrod (commented):
The structured path should distinguish “the overall output was trimmed” from “the text preview was truncated.” Here a large image plus a 49-character caption produces `→ 200 char preview` and appends `...`, even though all 49 caption characters survived. That can make the model infer that omitted caption text existed.

Could you build `text_content = "\n".join(text_segments)`, report `len(preview)`, and append the ellipsis only when `len(text_content) > self.preview_chars`? A regression assertion in `test_structured_output_preview_keeps_text_and_names_dropped_parts` that the short caption is not followed by `...` would cover the image-heavy case this PR is meant to fix.

seratch (changes requested):
Before merging, please make the structured preview accurately distinguish trimming the overall multimodal output from truncating its text: report `len(preview)`, append `...` only when `len(text_content) > preview_chars`, and cover both a fully preserved short caption and a genuinely truncated long caption.

Please also scope recognized text parts to the canonical `input_text` type. The released `function_call_output` content union contains `input_text`, `input_image`, and `input_file`; unsupported part types should remain dropped/unknown rather than creating an additional permissive contract. With those focused changes and green checks, this should be ready for another review.

seratch:
Thanks for the concrete reproduction. It demonstrates a real gap: SDK-generated multimodal function outputs can be trimmed into a Python representation that spends the preview budget on inline image data and drops later text.

After reviewing the broader `ToolOutputTrimmer` history, we do not want to keep extending this path one output shape at a time. We plan to replace the current best-effort representation handling with one type-aware policy for canonical `input_text`, `input_image`, and `input_file` outputs, together with a fail-safe schema traversal and a contract-level test matrix.

That is materially broader than this patch, so I am going to close it in favor of the consolidated follow-up. We will carry over the image-first, short-caption, long-caption, and no-text regression cases. Thank you for supplying the concrete scenario that established the need.

PR #4199, opened 2026-08-05
fix: stop waiting on Queue.join() when the event consumer can die

### LINKED ISSUE #4198: Agent.as_tool(on_stream=...) deadlocks forever when the handler raises a BaseException
### Please read this first

- **Have you read the docs?** Yes — [Tools](https://openai.github.io/openai-agents-python/tools/) (agents as tools) and the Codex tool docs.
- **Have you searched for related issues?** Yes. Searched open and closed issues and PRs for `deadlock`, `hang`, `on_stream`, `event_queue`, `task_done`, and `queue join`. The nearest prior work is #4170, which fixed the same failure shape in the STT websocket listener; nothing covers the `on_stream` dispatchers.

### Describe the bug

`Agent.as_tool(on_stream=...)` hangs forever — no exception, no timeout — when the `on_stream` handler raises a `BaseException`. `asyncio.CancelledError` is the realistic trigger: any `asyncio.timeout()`, `wait_for`, or task cancellation inside the callback produces one.

The dispatcher pairs a producer with a background consumer:

```python
async def _run_handler(payload): 
    try:
        ...
    except Exception as exc:      # BaseException escapes
        log_model_and_tool_action_error(...)

async def dispatch_stream_events():
    while True:
        payload = await event_queue.get()
        try:
            if payload is not None:
                await _run_handler(payload)
        finally:
            event_queue.task_done()
        if payload is None:
            break
...
finally:
    await event_queue.put(None)
    await event_queue.join()      # <-- waits for a consumer that is already dead
    await dispatch_task
```

`_run_handler` catches only `Exception`, so a `BaseException` propagates out of `dispatch_stream_events` and the task ends. The producer then reaches `event_queue.join()`, which waits for `task_done()` calls that can never arrive. The run stops permanently.

The same copy of this pattern exists in two more places:

- `src/agents/extensions/experimental/codex/codex_tool.py::_consume_events` — hangs identically, and because the `finally` block finishes active tracing spans *after* the join, those spans are also stranded open.
- `src/agents/sandbox/memory/manager.py::flush` — the worker catches only `Exception`, so a `BaseException` from `_process_rollout_file` ends it and the `await self._queue.join()` before phase-two consolidation blocks forever.

Note the `join()` is redundant in all three: the sentinel is queued last and the consumer only returns after consuming it, so awaiting the consumer task already implies every event was handled.

### Debug information

- Agents SDK version: `0.19.3` (reproduced on `main` at `19e364c1`)
- Related library versions: n/a
- Python version: 3.12
- Operating system: Windows 11 (not platform specific — this is pure asyncio)
- Model and model provider: none needed; reproduced with `tests/fake_model.py`

No exception is raised and no traceback is produced — the hang is the symptom.

### Repro steps

```python
import asyncio
import sys

async def main() -> None:
    inner_model = FakeModel()
    inner = Agent(name="Inner", model=inner_model)
    inner_model.add_multiple_turn_outputs([[get_text_message("inner done")]])

    async def on_stream(payload):
        # Any cancellation inside a handler -- an asyncio.timeout(), a cancelled
        # task -- surfaces here as CancelledError, which is a BaseException.
        raise asyncio.CancelledError()

    tool = inner.as_tool(
        tool_name="inner_tool", tool_description="run inner", on_stream=on_stream
    )

    outer_model = FakeModel()
    outer = Agent(name="Outer", model=outer_model, tools=[tool])
    outer_model.add_multiple_turn_outputs([
        [get_function_tool_call("inner_tool", '{"input": "hi"}', call_id="c1")],
        [get_text_message("outer done")],
    ])

    try:
        result = await asyncio.wait_for(Runner.run(outer, "go"), timeout=10)
        print("completed:", result.final_output)
    except asyncio.TimeoutError:
        print("DEADLOCK: the run never finished")


asyncio.run(main())
```

### Actual behavior

```text
DEADLOCK: the run never finished
```

The `wait_for` in the repro is only there to bound the script. Without it the coroutine never completes. Python also reports the dead consumer separately:

```text
Task exception was never retrieved
future: <Task finished coro=<...dispatch_stream_events() done> exception=CancelledError()>
```

### Expected behavior

The tool invocation settles. Either the handler's `BaseException` propagates to the caller or the run continues, but it must not block forever waiting on a consumer that has already exited.

### Root-cause hypothesis

*(hypothesis)* Waiting on `asyncio.Queue.join()` assumes the consumer stays alive to call `task_done()`. Because the handler wrapper only catches `Exception`, that assumption breaks for any `BaseException`. Awaiting the consumer task instead of the queue gives the same completion guarantee — the sentinel is the last item queued and the consumer returns only after consuming it — while surfacing a dead consumer rather than waiting on it.

### Proposed scope

Replace the `join()` wait with a wait on the consumer task in the three dispatchers above. No public API change and no change to handler error semantics for ordinary `Exception`s, which stay logged and swallowed.

abhay-codes07:
### Summary

Three streaming dispatchers hand user events to a background consumer task and then wait on `asyncio.Queue.join()` before returning. Each wraps the user callback in `except Exception`, so a `BaseException` escapes and ends the consumer. The outstanding `task_done()` calls can then never arrive, and `join()` blocks forever — the run hangs with no exception and no timeout.

`asyncio.CancelledError` is the realistic trigger: any `asyncio.timeout()`, `wait_for`, or task cancellation inside a user's handler raises one.

The `join()` is also redundant in all three cases. The sentinel is the last item queued and the consumer only returns after consuming it, so awaiting the consumer task already implies every event was handled — and it surfaces a dead consumer instead of waiting on it. Each fix is therefore a removal, not a new mechanism.

| Site | Impact |
| --- | --- |
| `Agent.as_tool(on_stream=...)` | tool invocation hangs permanently |
| `codex_tool._consume_events()` | hangs, **and** strands active tracing spans — `span.finish()` runs after the join |
| `sandbox/memory/manager.flush()` | hangs before phase-two consolidation |

Ordinary `Exception`s from handlers are unaffected: they stay logged and swallowed exactly as before.

This is the same failure shape as #4170, which added an error sentinel so the STT waiters could not hang on a dead websocket listener.

### Test plan

New regression tests, each parametrized over a custom `BaseException` and `asyncio.CancelledError`:

- `tests/test_agent_as_tool.py::test_agent_as_tool_streaming_does_not_hang_when_handler_raises_base_exception`
- `tests/extensions/experiemental/codex/test_codex_tool.py::test_codex_tool_consume_events_does_not_hang_on_base_exception`

Plus `test_agent_as_tool_streaming_reports_every_event_before_returning`, which pins the property that made dropping `join()` safe: the handler still observes every event before the tool returns. It awaits inside the handler so a missing wait would let the tool return first.

On `main` the four hang tests fail on the 5s guard; with the fix the same selection drops from **10.9s to 0.4s**:

```
# main
FAILED ...test_agent_as_tool_streaming_does_not_hang_when_handler_raises_base_exception[base_exception]
FAILED ...test_agent_as_tool_streaming_does_not_hang_when_handler_raises_base_exception[cancelled_error]
2 failed, 1 passed, 55 deselected in 10.94s

# with fix
3 passed, 55 deselected in 0.42s
```

| Command | Result |
| --- | --- |
| `make format` | clean |
| `uv run pytest tests/test_agent_as_tool.py tests/extensions/experiemental/codex/` | 190 passed |
| `make tests` | 5648 passed |

The full-suite run was done on Windows, where some sandbox symlink and tracing/realtime timing tests fail independently of this change. I diffed the failing set against a clean `main` checkout in the same environment: the two sets are identical (56 vs 56, no differences either way).

**One verification gap, stated plainly:** `tests/sandbox/test_memory.py` cannot be collected on Windows (`ImportError: UnixLocalSandbox is not supported on Windows`), so I could not execute a regression test for the `sandbox/memory/manager.py` site locally and did not add one I could not run. I did reproduce the deadlock for that site structurally — driving its exact worker/queue/sentinel shape, the worker dies on a `BaseException` and `join()` blocks with two unfinished items — and the change there is the same removal as the other two. Happy to add a test for it if you would rather have one, or to drop that hunk from this PR and leave it for someone who can run the file.

### Issue number

Closes #4198

### Checks

---

@seratch — this is the same shape you fixed for the STT listener in #4170, so I went looking for siblings of that pattern and these three were the ones where a consumer death can strand a `Queue.join()`.

Worth your call on one point: dropping `join()` means a `BaseException` from a handler now propagates out of the tool instead of hanging, which is a behavior change at that boundary even though the previous behavior was an unrecoverable hang. If you would rather keep handler failures fully contained, the alternative is widening the wrapper to `except BaseException` and re-raising `CancelledError` only — I did not do that because swallowing cancellation is usually worse than surfacing it.

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -942,7 +942,13 @@ class Agent(AgentBase, Generic[TContext]):
                                 pass
                         else:
                             await event_queue.put(None)
-                            await event_queue.join()
+                            # Wait on the dispatcher itself rather than on the queue draining.
+                            # The sentinel is the last item queued and the dispatcher only
+                            # returns after consuming it, so awaiting the task already implies
+                            # every event was handled. Waiting on `join()` instead would hang
+                            # forever when a handler raised a BaseException such as
+                            # CancelledError, because the dispatcher is then gone and the
+                            # outstanding `task_done()` calls can never arrive.
                             await dispatch_task
                     run_result = run_result_streaming
                 else:
--- a/src/agents/extensions/experimental/codex/codex_tool.py
+++ b/src/agents/extensions/experimental/codex/codex_tool.py
@@ -1115,8 +1115,13 @@ async def _consume_events(
     finally:
         if event_queue is not None:
             await event_queue.put(None)
-            await event_queue.join()
         if dispatch_task is not None:
+            # Wait on the dispatcher itself rather than on the queue draining. The sentinel is
+            # the last item queued and the dispatcher only returns after consuming it, so
+            # awaiting the task already implies every event was handled. Waiting on `join()`
+            # instead would hang forever when a handler raised a BaseException such as
+            # CancelledError, because the dispatcher is then gone and the outstanding
+            # `task_done()` calls can never arrive -- stranding the open spans below too.
             await dispatch_task
 
         # Ensure any open spans are closed even on failure.
--- a/src/agents/sandbox/memory/manager.py
+++ b/src/agents/sandbox/memory/manager.py
@@ -131,8 +131,12 @@ class SandboxMemoryGenerationManager:
                 self._ensure_worker()
                 for rollout_file in rollout_files:
                     self._queue.put_nowait(rollout_file)
-                await self._queue.join()
                 if self._worker_task is not None:
+                    # Wait on the worker rather than on the queue draining. The worker consumes
+                    # FIFO and only returns on _STOP, which is queued behind every rollout, so
+                    # awaiting it already implies each one was processed. Waiting on `join()`
+                    # instead would hang forever if the worker died on a BaseException, because
+                    # the outstanding `task_done()` calls can never arrive.
                     self._queue.put_nowait(_STOP)
                     await self._worker_task
                     self._worker_task = None
--- a/tests/extensions/experiemental/codex/test_codex_tool.py
+++ b/tests/extensions/experiemental/codex/test_codex_tool.py
@@ -2108,3 +2108,69 @@ async def test_codex_tool_argument_errors_respect_tool_data_redaction(
     else:
         assert _CODEX_TOOL_ARGUMENT_SECRET in str(error)
         assert isinstance(error.__cause__, cause_type)
+
+
+class _FatalCodexHandlerError(BaseException):
+    """A BaseException, so `_run_handler`'s `except Exception` does not catch it."""
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "raised",
+    [_FatalCodexHandlerError("fatal"), asyncio.CancelledError()],
+    ids=["base_exception", "cancelled_error"],
+)
+async def test_codex_tool_consume_events_does_not_hang_on_base_exception(
+    raised: BaseException,
+) -> None:
+    """A handler raising a BaseException must not strand `_consume_events`.
+
+    The dispatcher only catches `Exception`, so a BaseException such as `CancelledError`
+    terminates it. Waiting on `event_queue.join()` then waited for `task_done()` calls that
+    could never arrive, deadlocking the tool and leaving the active tracing spans unfinished.
+    """
+    events = [
+        {
+            "type": "item.completed",
+            "item": {"id": "agent-1", "type": "agent_message", "text": "done"},
+        },
+        {
+            "type": "turn.completed",
+            "usage": {"input_tokens": 1, "cached_input_tokens": 0, "output_tokens": 1},
+        },
+    ]
+
+    async def event_stream():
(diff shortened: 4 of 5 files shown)


seratch:
Thanks for identifying this issue and for the solid core approach. I have taken the fix into a maintainer-authored follow-up https://github.com/openai/openai-agents-python/pull/4201 that builds on your `Queue.join()` analysis while extending the implementation to supervise producers and consumers together, preserve cancellation behavior, guarantee Codex span cleanup, and cover the sandbox memory path.

Your contribution is credited with a `Co-authored-by` trailer. I am closing this PR to avoid maintaining two competing implementations. Thank you again for the investigation and patch.

PR #4203, opened 2026-08-05
fix(memory): roll back every failed SQLite session write

### LINKED ISSUE #4202: SQLite session writes other than add_items still strand the write lock on failure
### Please read this first

- **Have you read the docs?** Yes — [Sessions](https://openai.github.io/openai-agents-python/sessions/).
- **Have you searched for related issues?** Yes. #4163 fixed exactly this for `SQLiteSession.add_items`; this issue is that the same defect remains in the five sibling write paths. Also searched `rollback`, `write lock`, `in_transaction`, `database is locked`.

### Describe the bug

#4163 established that a write failing partway through leaves an open transaction on the cached connection, and that "an open write transaction would hold the SQLite write lock for the lifetime of the connection and block every later writer." That fix was applied only to `SQLiteSession.add_items`.

The same defect is still present in five other write paths:

| Method | Stranded write transaction after a failed statement |
| --- | --- |
| `SQLiteSession.add_items` (fixed in #4163) | no |
| `SQLiteSession.pop_item` | **yes** |
| `SQLiteSession.clear_session` | **yes** |
| `AsyncSQLiteSession.add_items` | **yes** |
| `AsyncSQLiteSession.pop_item` | **yes** |
| `AsyncSQLiteSession.clear_session` | **yes** |

`clear_session` is the clearest case: it issues two `DELETE`s under a single commit, so a failure on the second leaves the first applied inside an open transaction — both a partial mutation and a stranded lock.

`AsyncSQLiteSession` is the more damaging case. It holds **one** connection for the entire session (`self._connection`), so the stranded transaction holds the write lock until the session object is closed. Every later `add_items` from any writer — including other processes — then blocks or fails with `database is locked`. One transient `OperationalError` (disk full, a concurrent writer, a schema change) permanently wedges session persistence for that process.

### Debug information

- Agents SDK version: `0.19.4` (reproduced on `main` at `8be468f3`)
- Related library versions: `aiosqlite`
- Python version: 3.12
- Operating system: Windows 11 (not platform specific — this is SQLite transaction semantics)

### Repro steps

The script drops a table from an independent connection so that a statement inside the session's write fails, then checks whether the session's connection is still in a transaction and whether an independent writer can still take the write lock.

```python
import asyncio
import sqlite3
import tempfile
from pathlib import Path

from agents.extensions.memory import AsyncSQLiteSession

ITEMS = [{"role": "user", "content": "m0"}]


def drop(db, table):
    c = sqlite3.connect(str(db))
    c.execute(f"DROP TABLE {table}")
    c.commit()
    c.close()


def write_lock_held(db):
    probe = sqlite3.connect(str(db), timeout=0)  # no busy handler
    try:
        probe.execute("CREATE TABLE probe (x INTEGER)")
        probe.commit()
        return False
    except sqlite3.OperationalError:
        return True
    finally:
        probe.close()


async def main():
    # AsyncSQLiteSession.clear_session: the messages DELETE succeeds, the sessions DELETE fails.
    db = Path(tempfile.mkdtemp()) / "a.db"
    s = AsyncSQLiteSession("sid", db_path=str(db))
    await s.add_items(ITEMS)
    drop(db, "agent_sessions")

    try:
        await s.clear_session()
    except sqlite3.OperationalError:
        pass

    conn = await s._get_connection()
    print("in_transaction:", conn.in_transaction)
    print("write lock held:", write_lock_held(db))


asyncio.run(main())
```

### Actual behavior

```text
in_transaction: True
write lock held: True
```

The session's only connection sits in an open write transaction. Every subsequent write through that session, and every write from any other connection or process, is blocked until the session is closed.

Running the same probe across all six methods:

```text
SQLiteSession:
  add_items  (fixed in #4163)          open_txn=0 blocked=False
  pop_item                             open_txn=1 blocked=True
  clear_session (2nd DELETE fails)     open_txn=1 blocked=True

AsyncSQLiteSession:
  add_items                            in_transaction=True blocked=True
  pop_item                             in_transaction=True blocked=True
  clear_session (2nd DELETE fails)     in_transaction=True blocked=True
```

### Expected behavior

A failed write rolls back, leaving no partial mutation and no open transaction, and the session stays usable — the behavior #4163 established for `add_items`. All six rows above should read `blocked=False`.

### Root-cause hypothesis

*(hypothesis)* `_locked_connection()` yields a cached connection without managing transactions, so every write path has to roll back for itself. #4163 added that rollback inline to `add_items` only. Because the rollback obligation belongs to the connection rather than to any single method, a small `_rollback_on_failure(conn)` guard applied at each `_locked_connection()` write site would cover all of them from one place.

### Proposed scope

Add that guard to both SQLite session modules and use it in every write path. No public API change, no schema change, and no change to which statements are committed — only the failure path differs.

abhay-codes07:
### Summary

#4163 established that a write failing partway through leaves an open transaction on the cached connection, and that "an open write transaction would hold the SQLite write lock for the lifetime of the connection and block every later writer." That fix reached only `SQLiteSession.add_items`.

The same defect remained in five sibling write paths. Probing each one — drop a table from an independent connection so a statement inside the write fails, then check the session's connection and whether an independent writer can still take the lock:

```
                                     before          after
SQLiteSession
  add_items  (fixed in #4163)        blocked=False   blocked=False
  pop_item                           blocked=True    blocked=False
  clear_session (2nd DELETE fails)   blocked=True    blocked=False
AsyncSQLiteSession
  add_items                          blocked=True    blocked=False
  pop_item                           blocked=True    blocked=False
  clear_session (2nd DELETE fails)   blocked=True    blocked=False
```

`clear_session` is the clearest case: two `DELETE`s under one commit, so a failure on the second leaves the first applied inside an open transaction — a partial mutation *and* a stranded lock. `AsyncSQLiteSession` is the most damaging, because it holds one connection for the entire session, so the lock stays held until the session object is closed and one transient `OperationalError` permanently wedges session persistence for that process.

The rollback obligation belongs to the connection rather than to any individual method, so rather than inlining the same `try/except/rollback` six times this adds one `_rollback_on_failure(conn)` guard per module and applies it at every `_locked_connection()` write site — including the `add_items` path that previously inlined it, which keeps a single source of truth for the concern.

Scope notes:

- **No public API or schema change**, and no change to which statements are committed. Only the failure path differs.
- The guard catches `BaseException` rather than `Exception`, so an interrupted write cannot strand the lock either. This is the one intentional widening relative to the #4163 inline version.

### Test plan

Five regression tests, covering each still-broken path:

- `tests/memory/test_session.py` — `..._failed_clear_session_releases_write_lock`, `..._failed_pop_item_releases_write_lock`
- `tests/extensions/memory/test_async_sqlite_session.py` — `test_failed_add_items_releases_write_lock`, `test_failed_clear_session_releases_write_lock`, `test_failed_pop_item_releases_write_lock`

Each asserts the connection is no longer `in_transaction` and that an independent writer can still take the lock, using `sqlite3.connect(..., timeout=0)` to disable the busy handler so a held lock fails immediately rather than stalling. The async `add_items` test reuses the unserializable-item trigger from the #4163 test so the two match, and also asserts the session stays usable afterwards.

All five fail on `main` and pass with the fix. The existing #4163 test passes in both runs, which is the control that the probe is measuring the right thing:

```
# main
FAILED tests/memory/test_session.py::test_sqlite_session_failed_clear_session_releases_write_lock
FAILED tests/memory/test_session.py::test_sqlite_session_failed_pop_item_releases_write_lock
FAILED tests/extensions/memory/test_async_sqlite_session.py::test_failed_add_items_releases_write_lock
FAILED tests/extensions/memory/test_async_sqlite_session.py::test_failed_clear_session_releases_write_lock
FAILED tests/extensions/memory/test_async_sqlite_session.py::test_failed_pop_item_releases_write_lock
5 failed, 1 passed, 66 deselected
```

| Command | Result |
| --- | --- |
| `make format` | clean |
| `uv run pytest tests/memory/test_session.py tests/extensions/memory/test_async_sqlite_session.py` | 72 passed |
| `make tests` | 5794 passed |

The full-suite run was done on Windows, where some sandbox symlink and tracing/realtime timing tests fail independently of this change. I diffed the failing set against a clean `main` checkout in the same environment: the two sets are identical (54 vs 54, no differences either way).

### Issue number

Closes #4202

### Checks

---

@seratch — this is the follow-through on #4163: I went looking for the same shape in the other write paths and found it in five of them, with `AsyncSQLiteSession` never having been covered at all.

Two judgement calls worth your review. First, I consolidated the inline rollback from #4163 into the shared guard instead of leaving it and adding five more copies — that touches recently merged code, so say the word if you would rather I leave `add_items` exactly as it is and duplicate the pattern. Second, the guard catches `BaseException`; `SQLiteSession` runs its writes in a worker thread via `asyncio.to_thread`, so cancellation is not the concern there, but an interrupted write stranding the lock is the same failure and `rollback()` is safe in both cases.

### DIFF
--- a/src/agents/extensions/memory/async_sqlite_session.py
+++ b/src/agents/extensions/memory/async_sqlite_session.py
@@ -18,6 +18,22 @@ from ...memory.session_settings import (
 )
 
 
+@asynccontextmanager
+async def _rollback_on_failure(conn: aiosqlite.Connection) -> AsyncIterator[None]:
+    """Roll back a partially applied write when the operation fails.
+
+    `_locked_connection()` does not manage transactions, so a statement that fails partway
+    through a write would otherwise leave both a partial mutation and an open transaction on
+    this shared connection. An open write transaction holds the SQLite write lock for the
+    lifetime of the connection and blocks every later writer, including other processes.
+    """
+    try:
+        yield
+    except BaseException:
+        await conn.rollback()
+        raise
+
+
 class AsyncSQLiteSession(SessionABC):
     """Async SQLite-based implementation of session storage.
 
@@ -206,7 +222,7 @@ class AsyncSQLiteSession(SessionABC):
         if not items:
             return
 
-        async with self._locked_connection() as conn:
+        async with self._locked_connection() as conn, _rollback_on_failure(conn):
             await conn.execute(
                 f"""
                 INSERT OR IGNORE INTO {self.sessions_table} (session_id) VALUES (?)
@@ -239,7 +255,7 @@ class AsyncSQLiteSession(SessionABC):
         Returns:
             The most recent item if it exists, None if the session is empty
         """
-        async with self._locked_connection() as conn:
+        async with self._locked_connection() as conn, _rollback_on_failure(conn):
             cursor = await conn.execute(
                 f"""
                 DELETE FROM {self.messages_table}
@@ -284,7 +300,7 @@ class AsyncSQLiteSession(SessionABC):
 
     async def clear_session(self) -> None:
         """Clear all items for this session."""
-        async with self._locked_connection() as conn:
+        async with self._locked_connection() as conn, _rollback_on_failure(conn):
             await conn.execute(
                 f"DELETE FROM {self.messages_table} WHERE session_id = ?",
                 (self.session_id,),
--- a/src/agents/extensions/memory/async_sqlite_session.py
+++ b/src/agents/extensions/memory/async_sqlite_session.py
@@ -30,7 +30,13 @@ async def _rollback_on_failure(conn: aiosqlite.Connection) -> AsyncIterator[None
     try:
         yield
     except BaseException:
-        await conn.rollback()
+        # Rollback is best-effort cleanup. If it fails too -- a closed or otherwise unusable
+        # connection -- the original failure is the one worth reporting, so never let the
+        # cleanup error replace it.
+        try:
+            await conn.rollback()
+        except Exception:
+            pass
         raise
 
 
--- a/src/agents/memory/sqlite_session.py
+++ b/src/agents/memory/sqlite_session.py
@@ -26,7 +26,13 @@ def _rollback_on_failure(conn: sqlite3.Connection) -> Iterator[None]:
     try:
         yield
     except BaseException:
-        conn.rollback()
+        # Rollback is best-effort cleanup. If it fails too -- a closed or otherwise unusable
+        # connection -- the original failure is the one worth reporting, so never let the
+        # cleanup error replace it.
+        try:
+            conn.rollback()
+        except Exception:
+            pass
         raise
 
 
--- a/src/agents/memory/sqlite_session.py
+++ b/src/agents/memory/sqlite_session.py
@@ -14,6 +14,22 @@ from .session import SessionABC
 from .session_settings import SessionSettings, coerce_session_settings, resolve_session_limit
 
 
+@contextmanager
+def _rollback_on_failure(conn: sqlite3.Connection) -> Iterator[None]:
+    """Roll back a partially applied write when the operation fails.
+
+    `_locked_connection()` does not manage transactions, so a statement that fails partway
+    through a write would otherwise leave both a partial mutation and an open transaction on
+    this cached connection. An open write transaction holds the SQLite write lock for the
+    lifetime of the connection and blocks every later writer, including other processes.
+    """
+    try:
+        yield
+    except BaseException:
+        conn.rollback()
+        raise
+
+
 class SQLiteSession(SessionABC):
     """SQLite-based implementation of session storage.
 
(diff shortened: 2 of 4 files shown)


abhay-codes07:
Thanks both review points addressed in 1cf97072, though I only took one of them.

**Rollback masking the original error** (both modules): fixed. Rollback is cleanup, so a connection that is already closed or otherwise unusable would previously replace the failure the caller actually needs. It is now attempted best-effort and the original exception is always re-raised.

**`asyncio.shield()` around the async rollback**: I implemented it, then could not construct a case where it changes the outcome, so I dropped it again rather than keep machinery I cannot justify.

The reasoning: by the time the handler runs, `CancelledError` has already been delivered and caught, so the task is no longer in a cancelling state and a fresh `await conn.rollback()` proceeds normally. I verified this rather than assuming it — the new `test_cancelled_write_still_releases_write_lock` cancels a write mid-transaction, and it passes identically with and without the shield, while failing on `main` (the transaction stays open and the write lock stays held). So the base guard is what fixes the cancellation case; the shield adds nothing I can demonstrate.

I would rather leave it out than add an unexercised branch, but if you know of a path where the rollback await is itself cancelled a second `cancel()` arriving during cleanup, or a `TaskGroup` teardown I am happy to add it back with a test that actually distinguishes the two.

Current state: 6 regression tests fail on `main` and pass with the fix; the existing #4163 `add_items` test passes in both runs as the control.

seratch:
Thanks for sharing this patch. We'll close this PR in favor of https://github.com/openai/openai-agents-python/pull/4212, which covers all similar patterns across the SDK.

PR #4206, opened 2026-08-05
fix(memory): make SQLAlchemySession.pop_item atomic

### LINKED ISSUE #4205: SQLAlchemySession.pop_item returns the same item to concurrent callers
### Please read this first

- **Have you read the docs?** Yes — [Sessions](https://openai.github.io/openai-agents-python/sessions/) and the `Session` protocol, which documents `pop_item` as "Remove and return the most recent item from the session."
- **Have you searched for related issues?** Yes. Searched open and closed issues and PRs for `pop_item`, `concurrent`, `race`, `duplicate item`, and `SQLAlchemySession`. Nothing covers concurrent `pop_item`.

### Describe the bug

`SQLAlchemySession.pop_item()` is not atomic. It selects the newest row id, reads that row's payload, then deletes it — three separate statements:

```python
res = await sess.execute(subq)              # 1. pick the newest id
row_id = res.scalar_one_or_none()
res_data = await sess.execute(              # 2. read its payload
    select(self._messages.c.message_data).where(self._messages.c.id == row_id)
)
row = res_data.scalar_one_or_none()
await sess.execute(delete(self._messages).where(self._messages.c.id == row_id))  # 3. delete
```

Two concurrent callers can complete step 1 for the same `row_id` before either reaches step 3. Both then return that item. The second delete removes nothing, but the caller has no way to know — the delete result is discarded. Meanwhile the rows that no caller claimed are never popped and stay in the store.

`pop_item` is a destructive read, so returning the same item to two callers is not a benign duplicate: an application draining a session, retrying a turn, or rewinding history will process the same item twice while silently leaving other items behind.

The SQLite backends are unaffected — they claim the row with a single `DELETE ... RETURNING`.

### Debug information

- Agents SDK version: `0.19.4` (reproduced on `main` at `8f7e6d76`)
- Related library versions: `SQLAlchemy` 2.x, `aiosqlite`
- Python version: 3.12
- Operating system: Windows 11 (not platform specific — this is statement-level atomicity)
- Does the issue occur consistently or intermittently? Consistently, though the exact counts vary run to run.

### Repro steps

```python
import asyncio
import tempfile
from collections import Counter
from pathlib import Path

from agents.extensions.memory import SQLAlchemySession

N = 40


async def main() -> None:
    db = Path(tempfile.mkdtemp()) / "pop.db"
    session = SQLAlchemySession.from_url(
        "s", url=f"sqlite+aiosqlite:///{db.as_posix()}", create_tables=True
    )
    await session.add_items([{"role": "user", "content": f"m{i}"} for i in range(N)])

    popped = await asyncio.gather(*[session.pop_item() for _ in range(N)])
    contents = [item["content"] for item in popped if item is not None]

    print("returned :", len(contents))
    print("distinct :", len(set(contents)))
    print("repeats  :", {k: v for k, v in Counter(contents).items() if v > 1})
    print("remaining:", len(await session.get_items()))


asyncio.run(main())
```

### Actual behavior

```text
returned : 40
distinct : 26
repeats  : {'m34': 5, 'm21': 2, 'm14': 2, ...}
remaining: 14
```

Forty successful pops of a forty-item session returned twenty-six distinct items, one of them five times, and left fourteen rows in the store.

### Expected behavior

```text
returned : 40
distinct : 40
repeats  : {}
remaining: 0
```

Each item is removed and returned exactly once, which is what the SQLite backends already do.

### Root-cause hypothesis

*(hypothesis)* The `DELETE` is the only statement that atomically claims a row, but its result is discarded, so a caller that deleted nothing still returns the payload it read. Treating `rowcount == 0` as losing the race and continuing to the next item makes the claim atomic without needing `DELETE ... RETURNING`, which would restrict the dialects this fallback path is deliberately written to support.

### Proposed scope

Check the delete's `rowcount` in `SQLAlchemySession.pop_item` and continue the existing loop when it is zero. No public API change, no schema change, and no new dialect requirement.

I have a fix with a regression test ready and will open a PR referencing this issue.

abhay-codes07:
### Summary

`SQLAlchemySession.pop_item()` picks the newest row id, reads its payload, then deletes it — three separate statements. Concurrent callers can complete the select for the same id before either delete lands, so both return that item. The second delete removes nothing, but its result is discarded, so the caller cannot tell. The rows no caller claimed are then never popped.

`pop_item` is a destructive read, so this is not a benign duplicate: an application draining a session, retrying a turn, or rewinding history processes the same item twice while other items are silently left behind.

Forty concurrent pops of a forty-item session, on `main`:

```
returned : 40
distinct : 26          <- 14 items handed to two callers, one to five
remaining: 14          <- rows nobody claimed, still in the store
```

The delete is the only statement that atomically claims a row, so this treats deleting zero rows as losing the race and continues to the next item. Checking `rowcount` keeps the existing dialect-agnostic fallback rather than requiring `DELETE ... RETURNING` support, which is what the current code comment ("Fallback for all dialects") is deliberately written around.

The SQLite backends are unaffected — they already claim the row with a single `DELETE ... RETURNING`. I verified that by running the same probe against all three backends; only the SQLAlchemy one violates the contract.

### Test plan

`tests/extensions/memory/test_sqlalchemy_session.py::test_concurrent_pop_item_hands_each_item_to_one_caller` runs 40 concurrent `pop_item()` calls against a 40-item session and asserts every item is returned exactly once and the store ends empty.

Fails on `main`, passes with the fix:

```
# main
E       AssertionError: assert 29 == 40
E        +  where 29 = len({'m11', 'm12', ...})
E        +    where ... = set(['m39', 'm34', 'm34', 'm34', 'm34', 'm34', ...])
1 failed, 34 deselected
```

| Command | Result |
| --- | --- |
| `make format` | clean |
| `uv run pytest tests/extensions/memory/test_sqlalchemy_session.py` | 35 passed |
| `make tests` | 5769 passed |

The full-suite run was done on Windows, where some sandbox symlink and tracing/realtime timing tests fail independently of this change. I diffed the failing set against a clean `main` checkout in the same environment: the two sets are identical (58 vs 58, no differences either way).

### Issue number

Closes #4205

### Checks

---

@seratch — this came out of running the same concurrency contract against every session backend rather than reading one in isolation, which is why I am fairly confident the scope is exactly this one method.

One thing worth your call: `rowcount` is the minimal fix and keeps the fallback working on every dialect, but it does leave `pop_item` as three round trips per attempt. If you would rather have the atomic form where the dialect supports it, `delete(...).returning(message_data)` would collapse the read and the claim into one statement on PostgreSQL, SQLite 3.35+ and MariaDB, with this `rowcount` path kept for everything else. I did not do that here because it adds a dialect branch to a function that was deliberately written without one.

### DIFF
--- a/src/agents/extensions/memory/sqlalchemy_session.py
+++ b/src/agents/extensions/memory/sqlalchemy_session.py
@@ -435,7 +435,16 @@ class SQLAlchemySession(SessionABC):
                         select(self._messages.c.message_data).where(self._messages.c.id == row_id)
                     )
                     row = res_data.scalar_one_or_none()
-                    await sess.execute(delete(self._messages).where(self._messages.c.id == row_id))
+                    deleted = await sess.execute(
+                        delete(self._messages).where(self._messages.c.id == row_id)
+                    )
+
+                    # The DELETE is the only atomic claim on this row: the select above can
+                    # return the same id to concurrent poppers, so treat deleting zero rows as
+                    # losing the race and look for the next item instead of handing the same
+                    # item to two callers.
+                    if deleted.rowcount == 0:
+                        continue
 
                     if row is None:
                         continue
--- a/tests/extensions/memory/test_sqlalchemy_session.py
+++ b/tests/extensions/memory/test_sqlalchemy_session.py
@@ -6,6 +6,7 @@ import threading
 from collections.abc import Iterable, Sequence
 from contextlib import asynccontextmanager
 from datetime import datetime, timedelta
+from pathlib import Path
 from typing import Any, cast
 
 import pytest
@@ -990,3 +991,39 @@ async def test_runner_with_session_settings_override(agent: Agent):
     history_items = [item for item in last_input if item.get("content") != "New question"]
     # Should have 2 history items (last two from the 10 we added)
     assert len(history_items) == 2
+
+
+@pytest.mark.asyncio
+async def test_concurrent_pop_item_hands_each_item_to_one_caller(
+    agent: Agent, tmp_path: Path
+) -> None:
+    """pop_item removes an item, so concurrent callers must never receive the same one.
+
+    pop_item selects the newest id, reads its payload, then deletes it in three separate
+    statements. Concurrent callers can therefore select the same id before either delete
+    lands, and both return that item while the rows they never claimed stay in the store.
+    """
+    item_count = 40
+    session_id = "concurrent_pop"
+    session = SQLAlchemySession.from_url(
+        session_id,
+        url=f"sqlite+aiosqlite:///{(tmp_path / 'concurrent_pop.db').as_posix()}",
+        create_tables=True,
+    )
+
+    await session.add_items(
+        cast(
+            list[TResponseInputItem],
+            [{"role": "user", "content": f"m{index}"} for index in range(item_count)],
+        )
+    )
+
+    popped = await asyncio.gather(*[session.pop_item() for _ in range(item_count)])
+    contents = [item.get("content") for item in popped if item is not None]
+
+    assert len(contents) == item_count
+    # Each item belongs to exactly one caller.
+    assert len(set(contents)) == item_count
+    assert sorted(cast(list[str], contents)) == sorted(f"m{index}" for index in range(item_count))
+    # And nothing is left behind by a pop that returned an item it did not delete.
+    assert await session.get_items() == []


seratch:
Thanks for sharing this patch. We'll close this PR in favor of https://github.com/openai/openai-agents-python/pull/4212, which covers all similar patterns across the SDK.

PR #4208, opened 2026-08-05
fix: redact model output from JSON validation errors

### LINKED ISSUE #4207: Model output leaks into JSON validation errors despite DONT_LOG_MODEL_DATA
### Please read this first

- **Have you read the docs?** Yes, including the `OPENAI_AGENTS_DONT_LOG_MODEL_DATA` flag, which defaults to not logging LLM inputs and outputs.
- **Have you searched for related issues?** Yes. #4182 redacted invalid tool argument errors and #4049, #4067, #4078 covered MCP transport errors. Nothing covers the model output path through `validate_json`.

### Describe the bug

`agents.util._json.validate_json` puts the offending model output into its `ModelBehaviorError` message and chains the pydantic `ValidationError`, which repeats the same payload again in its own `input_value`:

```python
raise ModelBehaviorError(
    f"Invalid JSON when parsing {json_str} for {type_adapter}; {e}"
) from e
```

Neither is gated on `_debug.DONT_LOG_MODEL_DATA`, which is `True` by default. Any application that logs the exception, or lets it reach a traceback, records the raw model output.

This is the model side of the redaction added for tool arguments in #4182. It affects the three callers that pass model generated JSON:

- `agents/agent_output.py` output schema validation, so any agent with an `output_type`
- `agents/handoffs/__init__.py` handoff input validation
- `agents/realtime/handoffs.py` the realtime equivalent

The `SpanError` raised alongside it already uses `data={}`, so tracing is redacted correctly. Only the exception leaks.

### Debug information

- Agents SDK version: `0.19.4`, reproduced on `main` at `f6a32fee`
- Related library versions: `pydantic` 2.x
- Python version: 3.12
- Model and model provider: none needed, reproduced with `tests/fake_model.py`

### Repro steps

```python
import asyncio
import sys

sys.path.insert(1, "tests")  # run from the repo root
from test_responses import get_text_message

SECRET = "SUPER_SECRET_VALUE_9f3a"


class Structured(BaseModel):
    answer: str
    count: int


async def main() -> None:
    model = FakeModel()
    agent = Agent(name="A", model=model, output_type=Structured)
    # The model omits a required field, so validation fails.
    model.set_next_output([get_text_message(f'{{"answer": "{SECRET}"}}')])

    try:
        await Runner.run(agent, "go")
    except Exception as error:
        print("message:", error)
        print("cause  :", type(error.__cause__).__name__ if error.__cause__ else None)
        print("leaked :", SECRET in str(error))


asyncio.run(main())
```

### Actual behavior

```text
message: Invalid JSON when parsing {"answer": "SUPER_SECRET_VALUE_9f3a"} for TypeAdapter(Structured); 1 validation error for Structured
count
  Field required [type=missing, input_value={'answer': 'SUPER_SECRET_VALUE_9f3a'}, input_type=dict]
cause  : ValidationError
leaked : True
```

The payload appears twice, once as the raw JSON and once inside the chained `ValidationError`.

### Expected behavior

With the default flag, the error names the failure without carrying the payload, and nothing reachable from it carries the payload either:

```text
message: Invalid JSON when parsing model output for TypeAdapter(Structured)
cause  : None
leaked : False
```

With `OPENAI_AGENTS_DONT_LOG_MODEL_DATA=0` the current detailed error is still useful and should be preserved.

### Root-cause hypothesis

*(hypothesis)* `validate_json` predates the data logging policy and formats the payload unconditionally. Gating the detailed message on `_debug.DONT_LOG_MODEL_DATA` handles the message, but the chained `ValidationError` has to be dropped as well, since it embeds the same input. Raising the redacted error outside the `except` block, the way #4182 does, keeps both `__cause__` and `__context__` clear.

### Proposed scope

Gate the detailed error in `validate_json` on the model data flag and raise the redacted error outside the `except` block. One change covers all three callers. No public API change, and no change to the already redacted span error.

abhay-codes07:
### Summary

`agents.util._json.validate_json` formats the offending model output into its `ModelBehaviorError` message and chains the pydantic `ValidationError`, which repeats the same payload in its own `input_value`. Neither is gated on `_debug.DONT_LOG_MODEL_DATA`, which is `True` by default, so structured output that fails schema validation reaches the exception, its cause chain, and any traceback the application logs.

This is the model side of the tool argument redaction in #4182, and one change covers the three callers that pass model generated JSON: output schema validation (any agent with an `output_type`), handoff input validation, and its realtime equivalent.

Before, with the default flag:

```
message: Invalid JSON when parsing {"answer": "SUPER_SECRET_VALUE_9f3a"} for TypeAdapter(Structured); 1 validation error for Structured
count
  Field required [type=missing, input_value={'answer': 'SUPER_SECRET_VALUE_9f3a'}, input_type=dict]
cause  : ValidationError
```

After:

```
message: Invalid JSON when parsing model output for TypeAdapter(Structured)
cause  : None
```

The fix follows the same shape as #4182: keep the detailed error when the flag allows it, and otherwise raise outside the `except` block so neither `__cause__` nor `__context__` keeps the `ValidationError` reachable. Suppressing only the message would not be enough, because the chained error carries the payload independently.

Scope notes:

- Diagnostics are unchanged with `OPENAI_AGENTS_DONT_LOG_MODEL_DATA=0`: same message, same chained `ValidationError`.
- The `SpanError` raised alongside already used `data={}`, so tracing was correct and is untouched.
- `DONT_LOG_MODEL_DATA` is the right flag for all three callers because the payload is LLM output in every case, including handoff arguments.

### Test plan

Four tests in `tests/test_error_logging_redaction.py`, alongside the existing #4182 cases:

- `test_output_schema_validation_error_redacts_payload_when_model_data_disabled`
- `test_output_schema_validation_error_preserves_diagnostics_when_model_data_enabled`
- `test_handoff_input_validation_error_redacts_payload_when_model_data_disabled`
- `test_run_surfaces_redacted_output_validation_error`, an end to end run through `Runner.run`

Each redaction test asserts the payload is absent from the message and that `__cause__` and `__context__` are both `None`, so the check cannot pass by suppressing the message alone.

Three fail on `main` and pass with the fix. The diagnostics test passes in both runs, which is the control that the flag still enables the detailed error:

```
# main
FAILED tests/test_error_logging_redaction.py::test_output_schema_validation_error_redacts_payload_when_model_data_disabled
FAILED tests/test_error_logging_redaction.py::test_handoff_input_validation_error_redacts_payload_when_model_data_disabled
FAILED tests/test_error_logging_redaction.py::test_run_surfaces_redacted_output_validation_error
3 failed, 1 passed, 67 deselected
```

| Command | Result |
| --- | --- |
| `make format` | clean |
| `uv run pytest tests/test_error_logging_redaction.py` | 71 passed |
| `make tests` | 5805 passed |

The full suite run was done on Windows, where some sandbox symlink and tracing timing tests fail independently of this change. I diffed the failing set against a clean `main` checkout in the same environment. This branch has one fewer failure than the baseline, and the difference is `test_tracing_errors_streamed.py::test_max_turns_exceeded`, a timing test that flakes on `main` on its own. Nothing else differs in either direction.

### Issue number

Closes #4207

### Checks

### DIFF
--- a/src/agents/util/_json.py
+++ b/src/agents/util/_json.py
@@ -6,6 +6,7 @@ from typing import Any, Literal
 from pydantic import TypeAdapter, ValidationError
 from typing_extensions import TypeVar
 
+from .. import _debug
 from ..exceptions import ModelBehaviorError
 from ..tracing import SpanError
 from ._error_tracing import attach_error_to_current_span
@@ -32,9 +33,16 @@ def validate_json(
                 data={},
             )
         )
-        raise ModelBehaviorError(
-            f"Invalid JSON when parsing {json_str} for {type_adapter}; {e}"
-        ) from e
+        if not _debug.DONT_LOG_MODEL_DATA:
+            raise ModelBehaviorError(
+                f"Invalid JSON when parsing {json_str} for {type_adapter}; {e}"
+            ) from e
+
+    # Only reachable when the model-data policy suppressed the detailed error above. This is
+    # raised outside the except block on purpose: chaining it would keep the ValidationError
+    # reachable through __cause__ and __context__, and that error embeds the offending model
+    # output in its own message.
+    raise ModelBehaviorError(f"Invalid JSON when parsing model output for {type_adapter}")
 
 
 def _to_dump_compatible(obj: Any) -> Any:
--- a/tests/test_error_logging_redaction.py
+++ b/tests/test_error_logging_redaction.py
@@ -20,7 +20,7 @@ from unittest.mock import patch
 import httpx
 import pytest
 from openai import AsyncOpenAI
-from pydantic import BaseModel, ValidationError
+from pydantic import BaseModel, TypeAdapter, ValidationError
 
 import agents._debug as _debug
 from agents import (
@@ -31,9 +31,11 @@ from agents import (
     OpenAIResponsesModel,
     RunConfig,
     RunContextWrapper,
+    Runner,
     function_tool,
     trace,
 )
+from agents.agent_output import AgentOutputSchema
 from agents.logger import (
     log_model_action_debug,
     log_model_action_error,
@@ -56,6 +58,10 @@ from agents.tracing.processor_interface import TracingProcessor
 from agents.tracing.provider import SynchronousMultiTracingProcessor
 from agents.tracing.spans import Span
 from agents.tracing.traces import Trace
+from agents.util._json import validate_json
+
+from .fake_model import FakeModel
+from .test_responses import get_text_message
 
 _SECRET = "super secret prompt content"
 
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for the contribution. The underlying issue is valid, but the current patch still leaves model-generated handoff arguments exposed when tool-data redaction remains enabled, and the payload remains reachable through SDK traceback locals. These are part of the same default redaction contract, including the synchronous and asynchronous Realtime paths.

We are going to supersede this PR with a maintainer patch that applies the combined model/tool policy to handoffs and clears redacted traceback frames at the relevant Runner and Realtime boundaries. The broader　RunErrorDetails.raw_responses sanitization is not needed because the failing current response is not appended before turn processing returns.

PR #4215, opened 2026-08-05
fix(sandbox): stop splitting Cloudflare SSE events at chunk boundaries

### LINKED ISSUE #4214: Cloudflare sandbox splits SSE events when a chunk boundary falls on a CR
### Please read this first

- **Have you read the docs?** Yes, including [Sandbox clients](https://openai.github.io/openai-agents-python/sandbox/clients/).
- **Have you searched for related issues?** Yes. Searched open and closed issues and PRs for `cloudflare`, `SSE`, `line decoder`, `chunk`, and `stream boundary`. Nothing covers the Cloudflare stream decoder, and it currently has no tests.

### Describe the bug

`CloudflareSandboxSession` decodes the Worker's server-sent event stream with a hand-rolled line decoder. When a chunk ends on a CR, the decoder classifies it immediately: it emits the line and stores a bare CR as the buffer.

```python
elif cr + 1 == length:
    self._buf = b"\r"
    lines.append(line.decode("utf-8"))
    break
```

A CR at the end of the available bytes cannot be classified yet, because the next chunk may begin with LF and CRLF is a single terminator. On the next call the stored CR is treated as the start of a new line, so the following LF yields a blank line. In SSE a blank line dispatches the event, so one multi-line event is delivered as several.

Transport chunk boundaries are arbitrary, so the events an application receives depend on how the Worker response happens to be framed rather than on its content.

`flush()` has the matching problem and returns a blank line for a stream that ends on CR, and the CR-only and trailing-CR shapes are wrong even when the whole stream is decoded in one call.

### Debug information

- Agents SDK version: `0.19.4`, reproduced on `main` at `c48dd4c1`
- Related library versions: `aiohttp`
- Python version: 3.12
- Model and model provider: none needed, the decoder is pure
- Does the issue occur consistently or intermittently? Deterministic for a given chunking. In production it depends on transport framing, so it presents as intermittent.

### Repro steps

No Cloudflare account is required.

```python
from agents.extensions.sandbox.cloudflare.sandbox import _SSEDecoder, _SSELineDecoder


def decode(stream: str, size: int) -> list[str]:
    decoder = _SSELineDecoder()
    lines: list[str] = []
    for index in range(0, len(stream), size):
        lines.extend(decoder.decode(stream[index : index + size]))
    lines.extend(decoder.flush())
    return lines


def events(lines: list[str]) -> list[tuple[str, str]]:
    decoder = _SSEDecoder()
    out = []
    for line in lines:
        event = decoder.decode(line)
        if event is not None:
            out.append((event.event, event.data))
    return out


stream = "data: a\r\ndata: b\r\n\r\n"
print("splitlines    :", stream.splitlines())
print("one chunk     :", decode(stream, len(stream)), events(decode(stream, len(stream))))
print("split on the CR:", decode(stream, 8), events(decode(stream, 8)))
```

### Actual behavior

```text
splitlines    : ['data: a', 'data: b', '']
one chunk     : ['data: a', 'data: b', ''] [('message', 'a\nb')]
split on the CR: ['data: a', '', 'data: b', '', ''] [('message', 'a'), ('message', 'b')]
```

The same bytes produce one event or two depending only on where the chunk boundary falls.

Across seven representative stream shapes, six disagree with `str.splitlines()` at some chunk size. Fuzzing 400 randomized SSE streams over several chunk sizes, 191 produced different events than the same stream decoded in one call.

### Expected behavior

Line splitting is a property of the bytes, not of the framing, so every chunking yields `str.splitlines()` and the same events:

```text
split on the CR: ['data: a', 'data: b', ''] [('message', 'a\nb')]
```

### Root-cause hypothesis

*(hypothesis)* The decoder is adapted from the httpx line decoder, which deliberately defers a trailing CR into the next iteration and documents that it matches `str.splitlines()`. This copy instead emits the line and keeps the CR as buffer content, so the CR is later read as the beginning of a line rather than as the terminator of one already emitted. Keeping the unterminated line, CR included, until the next chunk arrives restores the httpx behavior.

### Proposed scope

Defer the trailing CR in `_SSELineDecoder.decode`, and make `flush` close the pending line rather than appending a blank one. The decoder currently has no tests, so add coverage that pins agreement with `str.splitlines()` at every chunk boundary. Confined to the Cloudflare provider, with no public API change.

abhay-codes07:
### Summary

The Cloudflare Worker's server-sent event stream is split into lines by a hand-rolled decoder that classifies a chunk ending on CR immediately: it emits the line and keeps a bare CR as the buffer. A CR at the end of the available bytes cannot be classified yet, because the next chunk may begin with LF and CRLF is a single terminator. The stored CR is then read as the start of a new line, so the following LF produces a blank line, and a blank line dispatches an SSE event. One multi-line event arrives as several.

Transport chunk boundaries are arbitrary, so which events an application receives depended on how the response happened to be framed rather than on its content:

```
stream = "data: a\r\ndata: b\r\n\r\n"

one chunk        ['data: a', 'data: b', '']          -> [('message', 'a\nb')]
split on the CR  ['data: a', '', 'data: b', '', '']  -> [('message', 'a'), ('message', 'b')]
```

`flush()` had the matching problem and appended a blank line for a stream ending on CR. The CR-only and trailing-CR shapes were also wrong when decoded in a single call, independently of chunking.

The decoder is adapted from the httpx line decoder, which defers a trailing CR into the next iteration and documents that it matches `str.splitlines()`. This restores that behavior: keep the unterminated line, CR included, until the next chunk decides, and let `flush` close the pending line instead of adding a blank one.

Measured against `str.splitlines()` as the oracle:

| | before | after |
| --- | --- | --- |
| stream shapes disagreeing at some chunk size | 6 of 7 | 0 of 7 |
| randomized streams whose events changed with chunking | 191 of 400 | 0 of 400 |

### Test plan

The decoders had no tests, which is how this survived. Added to `tests/extensions/sandbox/test_cloudflare.py`:

- `test_sse_line_decoder_matches_splitlines_for_every_chunk_boundary`, parametrized over LF, CRLF, CR, mixed, empty-data, trailing-CR and bare-CRLF streams, asserting agreement with `str.splitlines()` at every chunk size from 1 to the full length.
- `test_sse_event_is_not_split_when_crlf_straddles_a_chunk_boundary`, asserting the decoded events are identical for every chunk size.
- `test_sse_line_decoder_flush_does_not_emit_a_phantom_line_for_a_trailing_cr`.

8 fail on `main` and pass with the fix. The LF-only case passes in both runs, which is the control confirming the tests are not simply rejecting the decoder wholesale:

```
# main
FAILED ...test_sse_line_decoder_matches_splitlines_for_every_chunk_boundary[crlf]
FAILED ...test_sse_line_decoder_matches_splitlines_for_every_chunk_boundary[cr]
FAILED ...test_sse_line_decoder_matches_splitlines_for_every_chunk_boundary[mixed]
FAILED ...test_sse_line_decoder_matches_splitlines_for_every_chunk_boundary[empty_data]
FAILED ...test_sse_line_decoder_matches_splitlines_for_every_chunk_boundary[trailing_cr]
FAILED ...test_sse_line_decoder_matches_splitlines_for_every_chunk_boundary[bare_crlf]
FAILED ...test_sse_event_is_not_split_when_crlf_straddles_a_chunk_boundary
FAILED ...test_sse_line_decoder_flush_does_not_emit_a_phantom_line_for_a_trailing_cr
8 failed, 3 passed, 68 deselected
```

| Command | Result |
| --- | --- |
| `make format` | clean |
| `uv run pytest tests/extensions/sandbox/test_cloudflare.py` | 79 passed |
| `make tests` | 5827 passed |

The full suite run was done on Windows, where some sandbox symlink and tracing timing tests fail independently of this change. I diffed the failing set against a clean `main` checkout in the same environment and the two sets are identical, 54 versus 54, with no differences in either direction.

### Issue number

Closes #4214

### Checks

### DIFF
--- a/src/agents/extensions/sandbox/cloudflare/sandbox.py
+++ b/src/agents/extensions/sandbox/cloudflare/sandbox.py
@@ -231,8 +231,13 @@ class _SSELineDecoder:
                 if cr + 1 < length and raw[cr + 1 : cr + 2] == b"\n":
                     i = cr + 2
                 elif cr + 1 == length:
-                    self._buf = b"\r"
-                    lines.append(line.decode("utf-8"))
+                    # The CR is the last byte available, so it is either a lone CR
+                    # terminator or the first half of a CRLF split across chunks. Keep the
+                    # unterminated line and decide once the next chunk arrives. Emitting it
+                    # now would leave the following LF to be read as an empty line, and an
+                    # empty line dispatches the event, splitting one multi-line SSE event
+                    # into several.
+                    self._buf = raw[i:]
                     break
                 else:
                     i = cr + 1
@@ -247,11 +252,13 @@ class _SSELineDecoder:
     def flush(self) -> list[str]:
         buf = self._buf
         self._buf = b""
-        if buf == b"\r":
-            return [""]
-        if buf:
-            return [buf.decode("utf-8")]
-        return []
+        if not buf:
+            return []
+        if buf.endswith(b"\r"):
+            # A CR held back by decode() turned out to end the stream, so it is a lone CR
+            # terminator and the line it closes is the last one.
+            return [buf[:-1].decode("utf-8")]
+        return [buf.decode("utf-8")]
 
 
 class _SSEDecoder:
--- a/src/agents/extensions/sandbox/cloudflare/sandbox.py
+++ b/src/agents/extensions/sandbox/cloudflare/sandbox.py
@@ -207,12 +207,22 @@ class _ServerSentEvent:
 
 class _SSELineDecoder:
     _buf: bytes
+    _skip_leading_lf: bool
 
     def __init__(self) -> None:
         self._buf = b""
+        self._skip_leading_lf = False
 
     def decode(self, text: str) -> list[str]:
-        raw = self._buf + text.encode("utf-8")
+        data = text.encode("utf-8")
+        if self._skip_leading_lf and data:
+            # The previous chunk ended on a CR, which already terminated its line. A LF
+            # opening this chunk is the second half of that CRLF, so consume it instead of
+            # reading it as a blank line, which SSE treats as an event dispatch.
+            self._skip_leading_lf = False
+            if data.startswith(b"\n"):
+                data = data[1:]
+        raw = self._buf + data
         self._buf = b""
 
         lines: list[str] = []
@@ -231,13 +241,13 @@ class _SSELineDecoder:
                 if cr + 1 < length and raw[cr + 1 : cr + 2] == b"\n":
                     i = cr + 2
                 elif cr + 1 == length:
-                    # The CR is the last byte available, so it is either a lone CR
-                    # terminator or the first half of a CRLF split across chunks. Keep the
-                    # unterminated line and decide once the next chunk arrives. Emitting it
-                    # now would leave the following LF to be read as an empty line, and an
-                    # empty line dispatches the event, splitting one multi-line SSE event
-                    # into several.
-                    self._buf = raw[i:]
+                    # A CR is a complete line ending on its own, so deliver the line now
+                    # rather than waiting to learn whether a LF follows. Only remember that
+                    # a LF opening the next chunk belongs to this CRLF; without that, the
+                    # LF reads as a blank line and dispatches the event early, splitting one
+                    # multi-line event into several.
+                    self._skip_leading_lf = True
+                    lines.append(line.decode("utf-8"))
                     break
                 else:
                     i = cr + 1
@@ -252,12 +262,10 @@ class _SSELineDecoder:
     def flush(self) -> list[str]:
         buf = self._buf
         self._buf = b""
+        # A trailing CR already delivered its line, so there is nothing pending for it here.
+        self._skip_leading_lf = False
         if not buf:
             return []
-        if buf.endswith(b"\r"):
-            # A CR held back by decode() turned out to end the stream, so it is a lone CR
-            # terminator and the line it closes is the last one.
-            return [buf[:-1].decode("utf-8")]
         return [buf.decode("utf-8")]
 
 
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
The underlying CRLF chunk-boundary bug is valid, and fixing it in `_SSELineDecoder` is the right scope. Before merging, please preserve incremental CR-only behavior: a trailing CR is already a complete SSE line ending, so the decoder should emit that line immediately and remember only that a leading LF in the next chunk must be consumed as the remainder of CRLF.

Please update the tests to verify that `decode("data: a\r")` emits `"data: a"` immediately, a following `decode("\n")` emits nothing, and a CR-only blank line dispatches an event without waiting for another chunk or EOF. Please retain the existing every-boundary invariance coverage. This avoids fixing CRLF framing by introducing an unbounded delivery delay for another valid SSE line-ending form.

abhay-codes07:
Good catch, and you are right that the first version traded the framing bug for a delivery delay. Fixed in dd558687.

The decoder now emits the line as soon as it sees the trailing CR and remembers only that a LF opening the next chunk is the second half of that CRLF and must be consumed. That is the approach the httpx decoder this was adapted from uses; I had looked at it and then implemented the hold-back variant instead, which was the mistake.

Tests updated to pin the incremental behaviour you asked for:

- `test_sse_line_decoder_delivers_a_cr_terminated_line_immediately`: `decode("data: a\r")` returns `["data: a"]`, the following `decode("\n")` returns `[]`, and `flush()` adds nothing.
- `test_sse_line_decoder_dispatches_a_cr_only_blank_line_without_more_input`: `decode("data: a\r\r")` yields both lines in one call and the SSE decoder dispatches the event, with no further chunk and no EOF.
- `test_sse_line_decoder_flush_emits_only_an_unterminated_line`: flush returns a genuinely unterminated line but nothing for a delivered CR-terminated one.

The every-boundary invariance coverage is retained unchanged, and I extended the oracle set with `"\r"` and `"a\r\r\nb\n"`.

I checked the new tests actually discriminate rather than just passing: run against my previous hold-back commit, all three fail, including the CR-only dispatch one. Against `main`, 9 of the 13 sse tests fail. Randomized fuzzing over 500 streams and five chunk sizes shows no chunking-dependent event differences.

seratch (approved):

PR #4219, opened 2026-08-05
fix(models): resolve the any-llm default model at call time

adityasingh2400:
`AnyLLMProvider.get_model()` falls back to the module-level `DEFAULT_MODEL` constant, which is computed once at import time from `get_default_model()`, so any `OPENAI_DEFAULT_MODEL` value set after the module is imported is ignored and the provider keeps building models for the baked-in default. The sibling `LitellmProvider` calls `get_default_model()` on every lookup and explicitly notes that its `DEFAULT_MODEL` constant is only kept for backward compatibility, so the two providers disagree about the same setting. This resolves the fallback the same way inside `get_model()` and leaves `DEFAULT_MODEL` exported unchanged. Against upstream the new test fails with `assert 'openai/gpt-5.4-mini' == 'openai/gpt-4.1'` after setting the environment variable, and it passes with the fix.

### DIFF
--- a/src/agents/extensions/models/any_llm_provider.py
+++ b/src/agents/extensions/models/any_llm_provider.py
@@ -4,6 +4,7 @@ from ...models.default_models import get_default_model
 from ...models.interface import Model, ModelProvider
 from .any_llm_model import AnyLLMModel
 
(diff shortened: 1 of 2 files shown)


seratch (approved):
Looks good to me

PR #4224, opened 2026-08-05
fix(mcp): keep MCP error content when structured output is enabled

adityasingh2400:
When a local MCP server returns a `CallToolResult` with `isError` set, the actionable failure text lives in `content`. `MCPUtil.invoke_mcp_tool` only checked `server.use_structured_content` before serializing `structuredContent` and returning it exclusively, so a server constructed with `use_structured_content=True` had its error text silently replaced by the structured payload and the model saw what looked like a successful result.

MCP permits `structuredContent` alongside `isError`, so this is not a case of an invalid payload. It is a precedence policy in this SDK: for results the server flagged as errors, the error content wins over the structured payload, and every other case keeps the existing behavior, including the `isError=False` path.

The TypeScript SDK already does this and covers it with the "preserves MCP error content when structured output is enabled" test, so this brings the Python SDK back in line.

Two regression tests are added: `test_structured_content_skipped_for_error_results` fails on main with `assert '{"answer": 42}' == {'text': 'database connection refused', 'type': 'text'}` and passes with the fix, and `test_structured_content_used_for_non_error_results` pins the unchanged path. `uv run pytest tests/mcp` is green at 377 passed and 24 skipped, along with `make format`, `make lint`, and `make typecheck`.

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -763,8 +763,10 @@ class MCPUtil:
             logger.debug("MCP tool %s returned %s", tool_name_for_display, result)
 
         # If structured content is requested and available, use it exclusively. Results the
-        # server flagged as errors keep their content, because that is where the error text
-        # lives and structured content is not a valid output payload for a failed call.
+        # server flagged as errors keep their content instead, because that is where the
+        # actionable failure text lives. MCP permits `structuredContent` alongside
+        # `isError`, so this is an error-content precedence policy in this SDK rather than
+        # the structured payload being invalid for a failed call.
         tool_output: ToolOutput
         structured_content = result_structured_content(result)
         if server.use_structured_content and structured_content and not result_is_error(result):
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -762,10 +762,12 @@ class MCPUtil:
         else:
             logger.debug("MCP tool %s returned %s", tool_name_for_display, result)
 
-        # If structured content is requested and available, use it exclusively
+        # If structured content is requested and available, use it exclusively. Results the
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The implementation direction is correct: when an MCP result has `isError=True`, preserving its content keeps the actionable failure details visible to the model, while successful calls should continue using structured content when configured.

Before merge, please adjust the source comment and PR description. MCP permits `structuredContent` alongside `isError`, so this should be described as the SDK's error-content precedence policy rather than structured content being invalid. Also remove the statement that this involves `failure_error_function`; `isError=True` remains a normal tool result, while that handler applies to raised invocation or protocol failures. No logic or additional test changes are required. Once typecheck is green, this should be ready.

adityasingh2400:
Both points taken, done in 1199599 with no logic change.

You are right that I had the justification wrong. MCP permits `structuredContent` alongside `isError`, so the payload is not invalid for a failed call. The comment now states it as this SDK's error-content precedence policy:

```python
# If structured content is requested and available, use it exclusively. Results the
# server flagged as errors keep their content instead, because that is where the
# actionable failure text lives. MCP permits `structuredContent` alongside
# `isError`, so this is an error-content precedence policy in this SDK rather than
# the structured payload being invalid for a failed call.
```

I also removed the `failure_error_function` claim from the description. You are right that it does not belong: `isError=True` comes back as a normal tool result, so that handler was never in this path, and saying otherwise described a failure mode that does not exist. The description now frames the change as the precedence policy rather than an invalid payload.

Verification on the new commit: `tests/mcp` is 377 passed and 24 skipped, with ruff, ruff format and mypy clean.

seratch (approved):

PR #4228, opened 2026-08-05
fix(memory): cap OpenAI conversation item page size at the API maximum

adityasingh2400:
`OpenAIConversationsSession.get_items` passed the caller's `limit` straight through as the Conversations items API page size. The OpenAI client documents that parameter as "A limit on the number of objects to be returned. Limit can range between 1 and 100", so it is a per-request page size rather than a total, and any session limit above 100 raised an error instead of returning items while every other session backend simply returns the requested number.

This caps the page size at the API maximum and lets the auto-paginating iterator page until the requested number of items is collected, so the `Session` protocol behaves the same across backends.

The same parameter bound also rules out `limit=0`, which is the supported "no history" request the runner makes for `RunConfig(session_settings=SessionSettings(limit=0))` and which `SQLiteSession` already answers locally. So `get_items` now resolves the session limit first and returns an empty list for any non-positive value, before resolving the session id, so asking for no history does not create a remote conversation as a side effect.

Regression tests in `tests/memory/test_openai_conversations_session.py` cover the paging case, the small-limit case that must still be forwarded verbatim, the zero limit from both the call argument and `SessionSettings`, and the no-lazy-create behavior. On `main` four of the five fail, three with `ValueError: Invalid 'limit': expected a value between 1 and 100`. `make lint`, `make typecheck` and `tests/memory` are green.

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -11,6 +11,11 @@ from ..items import TResponseInputItem
 from .session import SessionABC
 from .session_settings import SessionSettings, coerce_session_settings, resolve_session_limit
 
+# The Conversations items API caps its ``limit`` parameter at 100, so that parameter is a
+# per-request page size rather than the total number of items to return. The auto-paginating
+# iterator keeps fetching pages, so a session limit above the cap is satisfied by paging.
+_MAX_ITEMS_PAGE_SIZE = 100
+
 
 async def start_openai_conversations_session(openai_client: AsyncOpenAI | None = None) -> str:
     _maybe_openai_client = openai_client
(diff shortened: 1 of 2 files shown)


adityasingh2400:
Good catch, and I confirmed it before fixing it.

`resolve_session_limit` returns `0` for `RunConfig(session_settings=SessionSettings(limit=0))`, and the installed SDK documents the bound as "Limit can range between 1 and 100", so that request could never be expressed. `SQLiteSession` already guards this with `if session_limit > 0:` and `test_session_limit_zero` covers the no-history path across `run`, `run_sync`, and `run_streamed`, so this backend was the odd one out.

I now resolve the limit first and return an empty list for any non-positive value, before resolving the session id, so asking for no history no longer creates a remote conversation as a side effect. I used `<= 0` rather than `== 0` because a negative page size is equally unrepresentable in a request, and there is no test or documented contract for negative limits anywhere in the repo.

Three regression tests cover it: an explicit `limit=0`, a `limit=0` coming from `session_settings`, and one asserting `conversations.create` is never awaited. All three fail on the previous commit with `Invalid 'limit': expected a value between 1 and 100, but got 0 instead`, which is the error the real API would return. The full `tests/memory` suite plus `tests/test_run_config.py` is green at 217 passed.

seratch:
The upper limit could be changed over time, so we'd like to avoid duplicated validation logic on the client side.

PR #4230, opened 2026-08-05
fix(handoffs): keep empty turns in the nested conversation history

### LINKED ISSUE #4229: Nested handoff history drops turns that have no content
### Please read this first

- **Have you read the docs?** Yes, including [Handoffs](https://openai.github.io/openai-agents-python/handoffs/).
- **Have you searched for related issues?** Yes. Searched open and closed issues and PRs for `nest_handoff_history`, `conversation history`, `handoff summary`, and `transcript`. The nearest prior work is #2171 on duplication in nested history; nothing covers turns being dropped.

### Describe the bug

`nest_handoff_history` summarizes the previous transcript into an assistant message, and the next handoff flattens that summary back into items so history does not nest repeatedly.

A turn with no content is rendered without the `role: content` separator:

```python
content_str = _stringify_content(item.get("content"))
return f"{prefix}: {content_str}" if content_str else prefix
```

so `{"role": "user", "content": ""}` becomes the record `user`. The parser splits each record on that separator and discards anything without one:

```python
role_part, sep, remainder = stripped.partition(":")
if not sep:
    return None
```

The turn is therefore dropped when the summary is flattened. The next agent receives a transcript with a turn missing, which also breaks the user/assistant alternation.

This affects `content=""` and `content=None`, for any role, with or without a `name`. A turn whose content is whitespace, or a non-string such as `[]`, is unaffected because those take a different formatting path.

### Debug information

- Agents SDK version: `0.19.4`, reproduced on `main` at `c48dd4c1`
- Related library versions: none relevant
- Python version: 3.12

### Repro steps

```python
from agents.handoffs import HandoffInputData, nest_handoff_history
from agents.handoffs.history import _extract_nested_history_transcript

history = (
    {"role": "user", "content": "first question"},
    {"role": "assistant", "content": ""},
    {"role": "user", "content": "second question"},
)
data = HandoffInputData(
    input_history=history,
    pre_handoff_items=(),
    new_items=(),
    run_context=None,
)

nested = nest_handoff_history(data)
transcript = _extract_nested_history_transcript(nested.input_history[0])

print("original turns:", len(history))
print("nested turns  :", len(transcript))
print("roles         :", [item.get("role") for item in transcript])
```

### Actual behavior

```text
original turns: 3
nested turns  : 2
roles         : ['user', 'user']
```

The assistant turn is gone, and the transcript now shows two consecutive user turns.

### Expected behavior

```text
original turns: 3
nested turns  : 3
roles         : ['user', 'assistant', 'user']
```

The turn is preserved, even though it carries no content.

### Root-cause hypothesis

*(hypothesis)* Omitting the separator for empty content makes the record ambiguous with arbitrary text, and the parser resolves that ambiguity by discarding it. Emitting the separator unconditionally makes the record parse like every other one. Summaries written before that change still contain bare roles, so recovering a separator-less record when it is a lone known role, optionally with a `(name)` suffix, keeps replayed history intact without letting prose inside the block become a fabricated turn.

### Proposed scope

Confined to the transcript formatter and parser in `src/agents/handoffs/history.py`. No public API change. Records that already contain a separator are unaffected.

abhay-codes07:
### Summary

`nest_handoff_history` summarizes the previous transcript into an assistant message, and the next handoff flattens that summary back into items so history does not nest repeatedly. A turn with no content was rendered as a bare role, with the `role: content` separator omitted. The parser splits each record on that separator and discards anything without one, so the turn was dropped.

```
original turns: 3   ['user', 'assistant', 'user']
nested turns  : 2   ['user', 'user']
```

The next agent receives a transcript with a turn missing, and the user/assistant alternation is broken. The turn stays lost across further handoffs. This affects `content=""` and `content=None`, for any role, with or without a `name`. Whitespace content and non-string content such as `[]` were unaffected, because those take a different formatting path.

The fix emits the separator unconditionally, so the record parses like every other one. Summaries written before this change still contain bare roles, so the parser also recovers a separator-less record when it is a lone known role, optionally with a `(name)` suffix. That is deliberately narrow: prose inside the block is still rejected rather than becoming a fabricated turn, and a test pins that.

### Test plan

Added to `tests/test_handoff_history_duplication.py`:

- `test_nested_history_keeps_turns_with_no_content`, parametrized over `content=""` for user and assistant, `content=None`, and a named turn, asserting the turn count and the role sequence survive.
- `test_nested_history_survives_repeated_handoffs`, nesting four times to show the turn is not shed on each hop.
- `test_bare_role_record_from_an_older_summary_is_recovered`, covering a summary written before this change.
- `test_prose_inside_the_summary_block_is_still_rejected`, the guard on the recovery.

6 fail on `main` and pass with the fix. The prose test passes in both runs, which is the control that the recovery is narrow rather than accepting anything without a separator:

```
# main
FAILED ...test_nested_history_keeps_turns_with_no_content[user_empty]
FAILED ...test_nested_history_keeps_turns_with_no_content[assistant_empty]
FAILED ...test_nested_history_keeps_turns_with_no_content[user_none]
FAILED ...test_nested_history_keeps_turns_with_no_content[named_empty]
FAILED ...test_nested_history_survives_repeated_handoffs
FAILED ...test_bare_role_record_from_an_older_summary_is_recovered
6 failed, 1 passed, 83 deselected
```

| Command | Result |
| --- | --- |
| `make format` | clean |
| `uv run pytest tests/test_handoff_history_duplication.py` | 90 passed |
| `uv run pytest tests/test_extension_filters.py tests/test_run_step_processing.py tests/test_agent_runner.py` | 338 passed with the file above |
| `make tests` | 5825 passed |

The full suite run was done on Windows, where some sandbox symlink and tracing timing tests fail independently of this change. I diffed the failing set against a clean `main` checkout in the same environment. The only difference was `tests/test_tracing_errors.py::test_multiple_final_output_doesnt_error`, which I then ran in isolation on both: it fails three times out of three on clean `main` as well, so it is pre-existing and unrelated.

### Issue number

Closes #4229

### Checks

### DIFF
--- a/src/agents/handoffs/history.py
+++ b/src/agents/handoffs/history.py
@@ -429,7 +429,9 @@ def _format_transcript_item_legacy(item: TResponseInputItem) -> str:
         if isinstance(name, str) and name:
             prefix = f"{prefix} ({name})"
         content_str = _stringify_content(item.get("content"))
-        return f"{prefix}: {content_str}" if content_str else prefix
+        # Always emit the separator. A bare role has no record separator for the parser to
+        # find, so the turn is dropped when the summary is flattened on the next handoff.
+        return f"{prefix}: {content_str}"
 
     item_type = item.get("type", "item")
     rest = {k: v for k, v in item.items() if k not in ("type", "provider_data")}
@@ -535,7 +537,17 @@ def _parse_summary_line(line: str) -> TResponseInputItem | None:
 
     role_part, sep, remainder = stripped.partition(":")
     if not sep:
-        return None
+        # Summaries written before the separator was always emitted record an empty turn as
+        # a bare role, so recover those instead of dropping the turn. Restricted to a lone
+        # role token, optionally with a "(name)" suffix, so prose inside the block is still
+        # rejected rather than turning into a fabricated message.
+        if not _is_bare_role_record(stripped):
+            return None
+        role, name = _split_role_and_name(stripped)
+        recovered: dict[str, Any] = {"role": role}
+        if name:
+            recovered["name"] = name
+        return cast(TResponseInputItem, recovered)
     role_text = role_part.strip()
     if not role_text:
         return None
@@ -589,6 +601,17 @@ def _strip_transcript_item_metadata(item: TResponseInputItem) -> TResponseInputI
     from ..run_internal.items import strip_internal_input_item_metadata
 
     return strip_internal_input_item_metadata(item)
+
+
+_KNOWN_TRANSCRIPT_ROLES = frozenset({"user", "assistant", "system", "developer"})
+
+
+def _is_bare_role_record(stripped: str) -> bool:
+    """Return whether a separator-less record is a lone role, optionally with a name."""
+    candidate = stripped
+    if candidate.endswith(")") and "(" in candidate:
+        candidate = candidate[: candidate.rfind("(")].strip()
+    return candidate in _KNOWN_TRANSCRIPT_ROLES
 
 
 def _split_role_and_name(role_text: str) -> tuple[str, str | None]:
--- a/src/agents/handoffs/history.py
+++ b/src/agents/handoffs/history.py
@@ -547,6 +547,10 @@ def _parse_summary_line(line: str) -> TResponseInputItem | None:
         recovered: dict[str, Any] = {"role": role}
         if name:
             recovered["name"] = name
+        # Keep an explicit empty content. Adapters such as the Chat Completions converter
+        # only recognize a message when both keys are present, so a role-only item is not
+        # replayable.
+        recovered["content"] = ""
         return cast(TResponseInputItem, recovered)
     role_text = role_part.strip()
     if not role_text:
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Before we can merge this, please preserve an explicit content: "" when parsing both role: records and recovered legacy bare-role records. At the current head, the parser returns a role-only item, which is not replayable through supported adapters such as the Chat Completions converter.

Please add a second-pass handoff_history_mapper regression that verifies the reconstructed item retains empty content and remains provider-valid. Also make the prose guard an independently numbered separator-less record; the current unnumbered line is treated as continuation text and never exercises the new rejection branch.

abhay-codes07:
All three addressed in d2b4fbf8.

**Explicit `content: ""`.** Both the separator-only `role: ` record and the recovered bare-role record now set content explicitly. You are right that a role-only item is not replayable; keeping the turn present but unconvertible would have left the original defect half fixed.

**Second-pass provider validity.** `test_second_pass_nesting_keeps_empty_turns_provider_valid` nests twice, asserts the reconstructed item is `{"role": "assistant", "content": ""}`, and then runs the transcript through `Converter.items_to_messages` to confirm it converts to three messages with the empty one intact. Pinning it against a real adapter rather than just asserting the dict shape.

**Prose guard.** Good catch, that test was passing for the wrong reason. The unnumbered line was folded into the previous record as continuation text by `_split_summary_records`, so it never reached the rejection branch at all. It is now numbered, which makes it an independent separator-less record that actually exercises the guard, and I also assert the surviving contents so a silently mangled record cannot pass.

I re-checked that the tests discriminate: 8 of the 9 fail on `main` and pass with the fix, with the prose guard passing in both runs as the control. Full stack is clean, with the suite failure set matching the Windows baseline and mypy at its 5 pre-existing errors.

seratch (approved):

PR #4241, opened 2026-08-06
fix: skip remote history for zero session limit

ump45nose:
### Summary

This pull request fixes `OpenAIConversationsSession.get_items()` when the effective session history limit is zero.

The Agents SDK session contract uses `limit=0` to disable history retrieval, but this backend forwarded zero to `conversations.items.list`. The Conversations API only accepts limits from 1 to 100, so the call failed instead of returning no history. The method now resolves the effective limit before initializing a remote conversation and returns an empty list locally when it is zero.

This preserves the existing behavior for positive limits and `None`, and avoids creating an otherwise unused remote conversation.

API reference: https://developers.openai.com/api/reference/python/resources/conversations/subresources/items/methods/list

### Test plan

- `uv run pytest -q tests/memory/test_openai_conversations_session.py` (`37 passed`)
- `uv run ruff check src/agents/memory/openai_conversations_session.py tests/memory/test_openai_conversations_session.py`
- `uv run mypy src/agents/memory/openai_conversations_session.py`
- `git diff --check`

The repository-wide verification script was not run locally; this PR remains Draft while CI supplies the broader regression signal.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -81,9 +81,11 @@ class OpenAIConversationsSession(SessionABC):
         self._session_id = None
 
     async def get_items(self, limit: int | None = None) -> list[TResponseInputItem]:
-        session_id = await self._get_session_id()
-
         session_limit = resolve_session_limit(limit, self.session_settings)
+        if session_limit == 0:
+            return []
+
+        session_id = await self._get_session_id()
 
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing the idea here, but I don't think this additional logic for limit: 0 use cases is necessary in real-world apps.

PR #4246, opened 2026-08-06
test(sandbox): cover the modal and cloudflare bucket mount strategies

abhay-codes07:
### Summary

Completes coverage of the sandbox provider mount strategies, following #4240 which did e2b and runloop.

| File | before | after |
| --- | --- | --- |
| `extensions/sandbox/modal/mounts.py` | 78% | 100% |
| `extensions/sandbox/cloudflare/mounts.py` | 85% | 100% |

Config building was already tested for both. The gap was the strategy lifecycle and the error branches, which normally need a live provider account to reach.

**Cloudflare**

- `activate` and `deactivate` drive `mount_bucket` and `unmount_bucket` with the resolved mount path and the built request options.
- The snapshot hooks unmount and then remount the same path, so a snapshot cannot capture a live bucket and the mount returns afterwards.
- All four lifecycle methods reject a session from another backend, asserted by checking that no bucket call was made rather than only that it raised.
- Prefix normalisation across `None`, empty, bare and wrapped forms, a half-supplied credential pair, and an unsupported mount type.

**Modal**

- `supports_native_snapshot_detach` is `False`, which is what routes snapshots through `_native_snapshot_requires_tar_fallback` instead of trusting a native snapshot taken while a bucket is attached. That link was previously unpinned, so it is worth having a test on it.
- `activate` and the snapshot hooks are no-ops, because Modal attaches buckets at sandbox create time, and `activate` and `deactivate` still reject a foreign session.
- Secret option validation, and the rule that inline credentials cannot be combined with a named secret, for s3, r2 and gcs.

No source changes. Every assertion matches current behaviour, so this is a characterisation of what the code does today.

### Test plan

New `tests/extensions/sandbox/test_modal_mounts.py`, following the existing `test_blaxel_mounts.py` convention, plus additions to `tests/extensions/sandbox/test_cloudflare.py`. Cloudflare uses a recording session that captures `mount_bucket` and `unmount_bucket` calls on a single ordered timeline, so the tests assert an exact call sequence rather than checking two independent lists. Modal's strategy only inspects the session class name, so a bare stand-in is enough; the file stubs the optional `modal` SDK when it is absent, since the package `__init__` imports the backend at module scope and would otherwise fail collection.

| Command | Result |
| --- | --- |
| `make format` | clean |
| `uv run pytest tests/extensions/sandbox/` | 773 passed, 1 skipped |
| `make tests` | 6035 passed |

The full suite run was done on Windows, where a set of sandbox symlink and tracing timing tests fail independently of this change. I diffed the failing set against clean `main` in the same environment: no new failures, with 35 more tests passing here.

### Issue number

None. This is coverage work on the sandbox providers rather than a reported defect.

### Checks

### DIFF
--- a/tests/extensions/sandbox/test_cloudflare.py
+++ b/tests/extensions/sandbox/test_cloudflare.py
@@ -1664,3 +1664,163 @@ def test_sse_line_decoder_flush_emits_only_an_unterminated_line() -> None:
     partial = _SSELineDecoder()
     assert partial.decode("data: b") == []
     assert partial.flush() == ["data: b"]
+
+
+class _RecordingBucketSession:
+    """Cloudflare-shaped session that records bucket mount calls."""
+
+    def __init__(self) -> None:
+        # _resolve_mount_path() reads the manifest root off the session.
+        self.state = _make_state(manifest=Manifest(root="/workspace"))
+        self.mounted: list[dict[str, Any]] = []
+        self.unmounted: list[Path] = []
+        # A single ordered timeline, so that tests claiming an order between mount
+        # and unmount actually constrain it rather than checking two separate lists.
+        self.events: list[str] = []
+
+    async def mount_bucket(
+        self, *, bucket: str, mount_path: Path, options: dict[str, object]
+    ) -> None:
+        self.mounted.append({"bucket": bucket, "mount_path": mount_path, "options": options})
+        self.events.append(f"mount:{mount_path}")
+
+    async def unmount_bucket(self, mount_path: Path) -> None:
+        self.unmounted.append(mount_path)
+        self.events.append(f"unmount:{mount_path}")
+
+
+_RecordingBucketSession.__name__ = "CloudflareSandboxSession"
+
+
+def _cf_s3_mount(strategy: CloudflareBucketMountStrategy, **kwargs: Any) -> S3Mount:
+    defaults: dict[str, Any] = {
+        "bucket": "bucket",
+        "access_key_id": "access-key",
+        "secret_access_key": "secret-key",
+        "mount_strategy": strategy,
+    }
+    defaults.update(kwargs)
+    return S3Mount(**defaults)
+
+
+@pytest.mark.parametrize(
+    "prefix, expected",
+    [(None, None), ("", "/"), ("/", "/"), ("data", "/data/"), ("/a/b/", "/a/b/")],
+    ids=["none", "empty", "slash", "bare", "wrapped"],
+)
+def test_cloudflare_prefix_is_normalised_to_a_bounded_path(
+    prefix: str | None, expected: str | None
+) -> None:
+    assert CloudflareBucketMountStrategy._normalize_prefix(prefix) == expected
+
+
+def test_cloudflare_config_rejects_a_half_supplied_credential_pair() -> None:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing these changes. I don't have the bandwidth to check this type of changes right now, so holding off sending simple test changes just for coverage improvement would be appreciated.

abhay-codes07:
Thanks, the Codex catch is right and is now fixed.

**Optional `modal` extra at collection time.** `agents.extensions.sandbox.modal.__init__` imports `.sandbox`, which does `import modal` at module scope, so the top-level submodule import took the whole file down before any test could run. Confirmed by blocking the module with a `sys.meta_path` hook:

```
src\agents\extensions\sandbox\modal\sandbox.py:32: in <module>
    import modal
E   ImportError: blocked optional dependency: modal
ERROR tests/extensions/sandbox/test_modal_mounts.py
!!!!! Interrupted: 1 error during collection !!!!!
```

`mounts.py` itself imports nothing from the SDK, so the file now installs a minimal stub for `modal`, `modal.config` and `modal.container_process` only when the real package is absent, then imports the strategy. Under the same blocking hook it is now 17 passed, and unchanged at 17 passed with the extra installed.

**I also tightened the bucket-call assertions.** These were checking `session.mounted` and `session.unmounted` as two independent lists, which is the same shape of weak assertion @seratch flagged on #4240. Both are now recorded on one ordered `session.events` timeline, so the snapshot test asserts an exact sequence:

```python
assert session.events == [f"unmount:{path}", f"mount:{path}"]
```

To be accurate about what that buys: for cloudflare the ordering was already implied by the test calling the two hooks in sequence, so this is not fixing a false pass the way it was on #4240. What it does add is that the exact call count and the absence of stray bucket calls are now pinned, and the foreign-session guard asserts `session.events == []` rather than two empty lists.

Rebased on `main`. The diff is 380 insertions, 0 deletions.

| Command | Result |
| --- | --- |
| `make mypy` | 0 issues in the touched files |
| `make pyright` | 0 errors |
| `uv run pytest tests/extensions/sandbox/` | 773 passed, 1 skipped |
| `make tests` | 6035 passed |
| coverage on `modal/mounts.py` and `cloudflare/mounts.py` | 100% each |

The full suite was run on Windows, where some sandbox symlink and tracing timing tests fail independently of this change. I diffed the failing set against clean `main` in the same environment: no new failures, with 35 more tests passing here.

seratch:
@abhay-codes07 I should have mentioned you in the previous comment, but I don't have the bandwidth to review this type of PRs at this moment. Keeping this one is okay, but please hold off sending similar ones (I won't actively review those).

abhay-codes07:
Understood, thanks for saying it plainly. I will stop sending coverage only test PRs. No rush on this one, pick it up whenever it suits you or leave it.

seratch:
I traced these cases through the existing suite, and the important behavior is already covered through stronger runtime paths: Cloudflare persistence and hydration assert the full unmount-operation-remount sequence, while the Modal tests already cover bucket configuration, secret validation, and the tar fallback for non-detachable mounts.

Most of this 380-line direct-method matrix therefore duplicates existing coverage, and the remaining private-helper branches do not demonstrate enough additional regression protection to justify a separate coverage-focused PR. I am going to close this PR.

PR #4247, opened 2026-08-06
fix(chatcmpl): reject truncated function calls

mhamzeh27903:
### Summary

This pull request fixes Chat Completions adapters finalizing function tool calls that the provider explicitly truncated with `finish_reason="length"`.

The shared Chat Completions terminal check now raises `ModelBehaviorError` before conversion or completion events in:

- OpenAI Chat Completions non-streaming responses
- LiteLLM non-streaming responses
- AnyLLM Chat Completions non-streaming responses
- direct streamed tool calls
- buffered streamed tool calls

This prevents partial arguments from reaching tool execution even when the accumulated fragment happens to be valid JSON. Existing behavior remains unchanged for length-terminated text, custom tool calls, normally completed function calls, and providers that omit terminal finish reasons.

### Test plan

- Added regression coverage for non-streaming, direct streaming, and buffered streaming paths.
- Added valid-JSON truncation cases for LiteLLM and AnyLLM to verify the terminal reason—not JSON parsing—drives rejection.
- Ran `env UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh` with formatting, lint, mypy, pyright, and the full test suite passing.

### Issue number

Related to #3861.

### Checks

### DIFF
--- a/src/agents/extensions/models/any_llm_model.py
+++ b/src/agents/extensions/models/any_llm_model.py
@@ -605,6 +605,15 @@ class AnyLLMModel(Model):
             ):
                 message.refusal = "Response withheld by the provider's content filter."
 
+            ChatCmplHelpers.raise_if_tool_calls_truncated(
+                first_choice.finish_reason if first_choice is not None else None,
+                has_function_tool_calls=bool(
+                    message is not None
+                    and message.tool_calls
+                    and any(tool_call.type == "function" for tool_call in message.tool_calls)
+                ),
+            )
+
             if tracing.include_data():
                 span_generation.span_data.output = (
                     [message.model_dump()] if message is not None else []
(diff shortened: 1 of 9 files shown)


seratch:
Thanks for the contribution. Failing closed when a provider explicitly reports a truncated function call is a reasonable direction, but this PR is currently backed only by manually constructed responses. The linked #3861 covers a different condition, missing tool-call identifiers, and does not demonstrate this new cross-adapter contract. The patch would also change released default behavior for OpenAI, LiteLLM, and AnyLLM.

I am going to close this PR for now. We can reconsider it with a reproducible named-provider/model case or redacted raw response/stream where a function call ends with `finish_reason="length"` and reaches the runner, ideally with syntactically valid truncated arguments. That evidence will let us decide whether rejection belongs in one adapter or the shared Chat Completions boundary.

PR #4261, opened 2026-08-07
fix: detect RunContextWrapper subclasses as the tool context parameter

LeSingh1:
`function_schema` identifies the context parameter with `origin is RunContextWrapper or origin is ToolContext`. `ToolContext` is a `RunContextWrapper` subclass that had to be added by hand; the SDK ships another one in `AgentHookContext`, and `docs/context.md` tells callers the SDK may hand them a specialized subclass.

Annotating a tool's first parameter with any other subclass leaves `takes_context` `False`. The context is then advertised to the model as a tool parameter, and `to_call_args` passes the model's value into the context slot. Under the default strict schema it fails earlier with an unrelated `additionalProperties` error instead. The non-first-position guard misses the same annotations, so its intended `UserError` never fires.

Match `RunContextWrapper` subclasses rather than exact types. Existing behavior for `RunContextWrapper[T]`, `ToolContext`, and non-context first parameters is unchanged.

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -283,6 +283,16 @@ def _extract_field_info_from_metadata(metadata: tuple[Any, ...]) -> FieldInfo | 
     return None
 
 
+def _is_run_context_annotation(annotation: Any) -> bool:
+    """Whether an annotation refers to RunContextWrapper or one of its subclasses.
+
+    ToolContext and AgentHookContext are SDK subclasses, and the docs tell callers the SDK may
+    hand them a specialized subclass, so subclasses have to count as context too.
+    """
+    origin = get_origin(annotation) or annotation
+    return isinstance(origin, type) and issubclass(origin, RunContextWrapper | ToolContext)
+
+
 def function_schema(
     func: Callable[..., Any],
     docstring_style: DocstringStyle | None = None,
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. The documented subclass behavior refers to specialized wrapper instances selected and supplied by the SDK; it does not make arbitrary `RunContextWrapper` subclasses valid function-tool annotations. Function tools are invoked with the SDK's `ToolContext`, and this patch changes schema classification without constructing the annotated subclass. A handler using subclass-specific behavior would therefore receive the wrong runtime type. `AgentHookContext` is similarly scoped to agent hooks, not function tools.

The supported approach is `RunContextWrapper[AppContext]` or `ToolContext[AppContext]`, with custom application state and behavior placed in `.context`. I am going to close this PR. If a concrete use case requires custom wrapper construction, please open an issue describing the required injection and lifecycle semantics.

LeSingh1:
That's a fair call and the objection is correct — thanks for spelling it out.

I misread the docs line as a statement about what annotations are valid, when it describes which wrapper the SDK hands you. And you're right that the patch only changes classification: nothing constructs the annotated subclass, so a handler relying on subclass behaviour would get a `ToolContext` and the type annotation would be a lie. That's worse than the current failure, which at least fails loudly.

No use case behind it — it came out of an audit, not a report — so there's nothing to open an issue for. Withdrawing it.

PR #4262, opened 2026-08-07
fix(run_state): make to_json repeatable

LeSingh1:
`_merge_generated_items_with_processed` builds the merged item list locally but also recorded the merge marker on the instance. The merged list is never stored back into `_generated_items`, so the next call takes the "already merged" shortcut and returns the unmerged list.

`to_json()` and `to_string()` therefore return different snapshots on the second call, and the second one silently drops every item that only exists in `last_processed_response.new_items`. Logging the state and then persisting it is enough to store a lossy snapshot.

Only callers that actually keep the merged list should mark it; the run loop already does that via `_mark_generated_items_merged_with_last_processed`, which is what the existing anonymous-tool-search dedupe test relies on.

Test asserts two consecutive `to_json()` calls agree. Fails before the fix.

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -769,8 +769,10 @@ class RunState(Generic[TContext, TAgent]):
                 seen_call_ids.add(call_id)
             generated_items.append(new_item)
 
-        if current_merge_marker is not None:
-            self._generated_items_last_processed_marker = current_merge_marker
+        # The merge result is local, so the marker must not be recorded here: doing so would make
+        # the next call take the already-merged shortcut above and return the unmerged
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this. I could not establish a supported runtime path that produces the state shape used by the new test; it assigns `_generated_items` and `_last_processed_response` directly. More importantly, removing the fallback marker can make public `RunResult.to_state()` and non-streaming interruption snapshots duplicate anonymous hosted-tool items that cannot be deduplicated by ID or call ID.

I am going to close this PR. If there is a public-API reproduction where two consecutive serializations lose data, please provide it together with a fix that preserves already-merged anonymous items.

PR #4263, opened 2026-08-07
fix(memory): skip conversation creation on uninitialized pop_item

LeSingh1:
`OpenAIConversationsSession.pop_item` calls `_get_session_id()` before anything else, so popping from a session that has never touched the remote API creates an empty conversation, binds the session to it, lists its items, and returns `None`. The round trip can never return an item — a conversation created one line earlier is empty by construction — and it leaves a stray conversation on the server plus a session id the caller never asked for.

`add_items` (#4190) and `clear_session` (#4111) already short-circuit in this situation, and the `session_id` docstring lists only `get_items()` and a non-empty `add_items()` as the methods that initialize a session. `pop_item` was the remaining one.

Return `None` up front when no conversation id exists yet.

Test: `test_pop_item_uninitialized_does_not_create_session` asserts no `conversations.create`, no `get_items`, no `items.delete`, and that `_session_id` stays `None`. Fails before the fix.

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -118,7 +118,14 @@ class OpenAIConversationsSession(SessionABC):
         )
 
     async def pop_item(self) -> TResponseInputItem | None:
-        session_id = await self._get_session_id()
+        async with self._session_id_lock:
+            session_id = self._session_id
+
+        if session_id is None:
+            # An uninitialized session has nothing to pop, so creating a remote conversation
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. The reported behavior is real: calling pop_item() on a never-initialized OpenAIConversationsSession creates an empty remote conversation before returning None. However, the demonstrated consequence is limited to an unnecessary API round trip and an empty resource; no data is lost, no incorrect item is returned, and no normal runner path was shown to depend on this case.

Given the maintenance bar for additional edge-case fixes, I am going to close this PR.

PR #4264, opened 2026-08-07
fix(run_state): keep current_step interruptions with an unknown agent name

LeSingh1:
`from_json` restores the same approval items twice: once under `last_processed_response.interruptions` and once under `current_step.data.interruptions`. The first call passes `fallback_agent=current_agent`, the second does not, so an approval whose agent name is missing from the restored graph resolves to `None` and is silently skipped.

The result is a `RunState` that still reports an interrupted step while `get_interruptions()` returns an empty list, so the caller can never approve or reject and the state cannot be resumed. Nothing is logged. Agent references are serialized by name unless the graph has duplicate names, so any rename between snapshot and resume hits this.

Pass the same fallback on both paths.

Test round-trips a state whose interruption names an agent outside the restored graph and asserts `get_interruptions()` matches `last_processed_response.interruptions`. Fails before the fix.

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -3052,6 +3052,7 @@ async def _build_run_state_from_json(
                 item_data,
                 agent_map=agent_map,
                 agent_identity_map=agent_identity_map,
+                fallback_agent=current_agent,
             )
             if approval_item is not None:
                 interruptions.append(approval_item)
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this. The regression test constructs an interruption owned by an agent outside the graph supplied during restoration. Reassigning that approval to the current agent is not a safe recovery: it changes the approval's identity instead of proving which agent and tool invocation it belongs to. Ambiguous durable authorization state should fail closed.

I am going to close this PR. We can reconsider if there is a reproduction using a supported serialized agent graph where an unchanged logical interruption becomes unresolvable.

PR #4265, opened 2026-08-07
feat: add Azure Cosmos DB session backend

nos-redacted:
### Summary

This pull request adds an Azure Cosmos DB-backed `Session` implementation for durable, multi-instance conversation history.

Each non-empty `add_items()` call uses one partition-scoped transactional batch and rejects more than 100 items before I/O. Internal retries reuse deterministic document IDs only for the same invocation, and ambiguous conflicts are accepted only after strict point-read replay validation. Separate `add_items()` calls are not idempotent and append duplicate input independently.

Clear operations rotate an incarnation fence, while pop operations use ETag-protected UUID claims. Clients must use Session consistency with `retry_write=False`; injected clients remain caller-owned. The change also adds lazy optional imports for the Cosmos extra, configuration documentation, an example, and distribution coverage. It does not provision Azure resources.

### Test plan

- `make format` completed without changing the frozen patch.
- `make lint` passed.
- `make typecheck` passed, including mypy and Pyright with no issues in 852 source files.
- Focused Cosmos DB and import tests passed on Python 3.13 and Python 3.10: 75 tests on each interpreter.
- `make tests` passed: 6,878 parallel tests passed with 29 skipped; 84 serial tests passed with 4 skipped and 57 deselected.
- `make coverage` passed at 91%, above the required 85% threshold.
- MkDocs, wheel and sdist builds, packaging and optional-extra profiles, isolated artifact imports, and Azure client-policy probes passed.
- Two independent elevated-risk final reviews found no actionable issues.

### Issue number

None.

### Checks

- [ ] I've run `.agents/skills/code-change-verification/scripts/run.sh` (the equivalent required commands were run manually in repository order)
- [ ] If using Codex, I've run `/review` before submitting this PR (not applicable; this PR was prepared with GitHub Copilot)


seratch:
@nos-redacted Thanks for your interest here. However, we are not actively adding session backend implementations within this repo at this moment. Please maintain the code as your own module if you're willing to do so. Thanks again for using Agents SDK!

PR #4266, opened 2026-08-07
fix(models): strip placeholder item IDs without provider_data on the Responses path

LeSingh1:
`_remove_openai_responses_api_incompatible_fields` returns the input list unchanged unless some item carries `provider_data`. The per-item cleaner it guards also strips `FAKE_RESPONSES_ID`, and that placeholder does not always travel with `provider_data` — `create_message_output_item` in the run error handlers builds its message with the placeholder id and nothing else.

Replaying such an item sends `id="__fake_id__"` to the Responses API, which rejects the request with `Invalid 'input[N].id'. Expected an ID that begins with 'msg'`.

Check for placeholder ids alongside `provider_data` when deciding whether the list needs cleaning. Lists that need neither are still returned as-is.

Adds a unit test for the placeholder-without-`provider_data` case and one asserting the built request payload drops the id.

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -916,17 +916,20 @@ class OpenAIResponsesModel(Model):
         This data transformation does not always guarantee that items from other provider
         interactions are accepted by the OpenAI Responses API.
 
-        Only items with truthy provider_data are processed.
         This function handles the following incompatibilities:
         - provider_data: Removes fields specific to other providers (e.g., Gemini, Claude).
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Can you resolve the typecheck error? Once it's fixed, this PR can be merged.

LeSingh1:
Fixed — the test passed `message.model_dump()` straight into `input`, which is `dict[str, Any]` rather than one of the `ResponseInputItemParam` members. Now cast to `TResponseInputItem`.

`mypy` clean on the file, 10 passed, ruff check and format clean.

seratch (approved):

PR #4267, opened 2026-08-07
fix(streaming): open a chunkless chat completions stream with response.created

LeSingh1:
`handle_stream` emits `response.created` from inside its chunk loop but emits `response.completed` unconditionally after it. A provider stream that yields no chunks — or one whose chunks are all swallowed by `buffer_tool_call_stream`, which returns without yielding when it never sees a last chunk — ends with a terminal event that no `created` opened. Consumers that initialize per-response state on `created` see `completed` first.

Emit the created event after the loop when nothing started the stream. Non-empty streams are unaffected: they still emit it on the first chunk.

The existing "nothing was produced" tests already assert `["response.created", "response.completed"]`; the new test covers the zero-chunk case.

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -1064,6 +1064,18 @@ class ChatCmplStreamHandler:
                             sequence_number=sequence_number.get_and_increment(),
                         )
 
+        # A provider can return a 200 whose stream carries no chunks at all, and
+        # buffer_tool_call_stream can also swallow every chunk. `response.created` is only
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. The new test demonstrates a synthetic zero-chunk iterator, but it does not establish a realistic provider path or user-visible failure. The patch also continues to treat a stream with no provider output as a successfully completed empty response; adding response.created only makes the event sequence look more regular without resolving that semantic question.

I am going to close this PR. If a supported provider produces this sequence, please provide the exact stream and consequence so we can decide whether the correct behavior is completion or a transport/model error.

PR #4268, opened 2026-08-07
fix(run): apply the reasoning item id policy to error handler run data

LeSingh1:
`build_run_error_data` accepts `reasoning_item_id_policy` so that `RunErrorData.history` and `.output` match the history the runner would actually replay. The `max_turns` call sites in `run.py` and `run_loop.py` pass it; the `model_refusal` and `invalid_final_output` call sites in `turn_resolution.py` never did.

With `RunConfig(reasoning_item_id_policy="omit")` a `max_turns` handler therefore receives reasoning items with their ids stripped, while a `model_refusal` or `invalid_final_output` handler receives them intact. A handler that does the natural thing and feeds `run_data.history` into a follow-up run succeeds under one error kind and gets a provider 400 under the other.

Thread the resolved policy from both run loops through `get_single_step_result_from_response` and `execute_tools_and_side_effects` to the two remaining call sites. The resolved value is used rather than `run_config` alone so resumed runs keep the policy stored on the `RunState`.

Tests: one refusal-path test and one parametrized streamed/non-streamed `invalid_final_output` test, all failing before the fix.

### DIFF
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -1914,6 +1914,7 @@ async def run_single_turn_streamed(
         server_manages_conversation=server_conversation_tracker is not None,
         event_queue=streamed_result._event_queue,
         before_side_effects=raise_if_input_guardrail_tripwire_known,
+        reasoning_item_id_policy=reasoning_item_id_policy,
     )
 
     items_to_filter = session_items_for_turn(single_step_result)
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for the contribution. reasoning_item_id_policy is already applied when the SDK constructs follow-up input through normal turns, results, sessions, server-managed deltas, and resumed runs. The remaining case requires application code inside an error handler to treat run_data.history as input for a separate run, which is not a demonstrated or documented workflow here.

I am going to close this PR for now. A concrete supported example showing why the existing result or session paths are insufficient would be needed to justify threading another policy parameter through the turn-resolution pipeline.

PR #4269, opened 2026-08-07
fix(extensions): keep an assistant turn whose tool call cannot be split

LeSingh1:
`_fix_tool_message_ordering` rewrites an assistant message carrying `tool_calls` into one message per call, then marks the original consumed. It only builds a split for tool calls with a usable id, and it has no fallback when `tool_calls` is not a list. An assistant turn whose tool call has an empty or missing id therefore contributes nothing and is dropped, taking its text and thinking blocks out of the conversation.

The tool-result branch of the same loop keeps results it cannot pair, so silently losing the turn is not the intended behavior.

Fall back to the original message when no split was emitted. A turn with a mix of usable and unusable ids still splits as before. Both the LiteLLM and any-llm copies of the routine have the defect, so both are fixed.

### DIFF
--- a/src/agents/extensions/models/any_llm_model.py
+++ b/src/agents/extensions/models/any_llm_model.py
@@ -1477,6 +1477,7 @@ class AnyLLMModel(Model):
             role = original_message.get("role")
             if role == "assistant" and original_message.get("tool_calls"):
                 tool_calls = original_message.get("tool_calls", [])
+                emitted_split = False
                 if isinstance(tool_calls, list):
                     for tool_call in tool_calls:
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for the contribution. Both new cases use malformed provider output: a tool call without a usable ID or a non-list tool_calls value. Preserving the original assistant message does not make that input valid; it can simply move the failure to the next provider request, and mixed valid/invalid calls remain only partially handled.

I am going to close this PR. If a supported provider emits a specific malformed shape in normal operation, please provide that payload and resulting failure so normalization or validation can be addressed at the owning adapter boundary.

PR #4272, opened 2026-08-07
fix(streaming): add the chat completions assistant message once per turn

LeSingh1:
Text deltas and refusal deltas each announce their own `response.output_item.added` for the assistant message. A turn that carries both emits two added events for a single `ResponseOutputMessage` and only one `output_item.done`, since both parts are assembled into one message for `response.completed`.

The refusal branch's comment says it starts a message "if one doesn't exist yet", but its guard only checks whether the refusal part exists, not whether the message was already announced by the text branch. The text branch has the mirror-image gap.

The non-streamed converter puts an `output_text` and a `refusal` on the same message, so this is a supported shape, not a malformed stream.

Announce the message from whichever part opens first; the second part emits only its own `content_part.added`. Test covers both orderings.

### DIFF
--- a/src/agents/models/chatcmpl_stream_handler.py
+++ b/src/agents/models/chatcmpl_stream_handler.py
@@ -814,23 +814,26 @@ class ChatCmplStreamHandler:
                             logprobs=[],
                         ),
                     )
-                    # Start a new assistant message stream
-                    assistant_item = ResponseOutputMessage(
-                        id=FAKE_RESPONSES_ID,
-                        content=[],
-                        role="assistant",
-                        type="message",
-                        status="in_progress",
-                    )
-                    if state.provider_data:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. The lifecycle mismatch is logically real for a synthetically constructed stream containing both content and refusal, and the patch appears to make the `output_item.added` and `output_item.done` counts consistent. However, the PR does not provide a real OpenAI or supported-provider response, user report, or integration reproduction showing that such a mixed stream occurs. The existing tests and non-streamed converter establish representability, but not practical reach or user impact.

Given the lack of a demonstrated consumer failure, I do not think this warrants a permanent behavior change and additional regression surface. I am going to close this PR for now. We can reconsider it with a captured provider stream or supported integration reproducer showing that the duplicate event causes an observable failure.

PR #4273, opened 2026-08-07
fix: keep FunctionTool subclasses copyable

LeSingh1:
`FunctionTool.__copy__` rebuilt the tool with `dataclasses.replace(self)`, which calls the concrete class constructor with every dataclass field. A `FunctionTool` subclass that defines its own `__init__` signature therefore raises `TypeError: __init__() got an unexpected keyword argument 'name'` on `copy.copy()`, and `tool_namespace()` raises the same error because it copies each tool before attaching namespace metadata.

The SDK's own sandbox tools hit this — `ExecCommandTool` and `WriteStdinTool` in `agents/sandbox/capabilities/tools/shell_tool.py` are `@dataclass(init=False)` `FunctionTool` subclasses with keyword-only constructors, so neither can be copied or grouped into a namespace.

Copy the instance state onto a fresh object of the same type and re-run `__post_init__` instead. That keeps every effect of the old path — invoker rebinding, the `allowed_callers` copy, the `params_json_schema` deep copy and strictification, `output_json_schema` normalization, and timeout validation — while also carrying non-field attributes across, which the old manual loop handled separately.

Tests cover `copy.copy()` and `tool_namespace()` on a subclass with a custom constructor.

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import ast
 import asyncio
 import copy
-import dataclasses
 import functools
 import inspect
 import json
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying bug is valid: `tool_namespace()` accepts `FunctionTool` instances, but the released `__copy__` path calls `dataclasses.replace()`, so SDK-owned `init=False` subclasses such as `ExecCommandTool` and `ViewImageTool` fail before namespacing. This is worth fixing.

Before merge, please preserve invoker ownership in the new copy path. After `__dict__.update()`, a subclass configured with `on_invoke_tool=self._invoke` remains bound to the original instance; the current test returns only `raw_input` and cannot detect this. Rebind directly bound invokers to the copied instance, then add a regression where `_invoke` reads `self.session`, the copied tool receives a different session sentinel, and invocation proves that the copied state is used. Arbitrary slot-only or custom-allocation subclasses do not need to be supported.

LeSingh1:
Done in ffe78c6.

The copy now rebinds an invoker that was bound to the original instance:

```python
invoker = copied_tool.__dict__.get("on_invoke_tool")
invoker_func = getattr(invoker, "__func__", None)
if invoker_func is not None and getattr(invoker, "__self__", None) is self:
    copied_tool.on_invoke_tool = invoker_func.__get__(copied_tool, type(copied_tool))
```

It only rebinds when `__self__` is the tool being copied, so a plain function, a lambda, or a method bound to some other object is left alone. The `__agents_bind_function_tool__` path in `__post_init__` still runs afterwards and is unaffected.

`_invoke` now reads `self.session` and returns it, and the new test `test_shallow_copied_subclass_invoker_uses_the_copied_instance_state` gives the copy a different sentinel, then asserts the copy sees `copied-session` and the original still sees `original-session`. Without the rebinding that test fails — I checked by removing just those three lines.

57 passed, mypy clean on `tool.py`, ruff check and format clean.

anujbolewar (commented):
The direct state rebuild is the right call for subclass init signatures, and rebinding the invoker to the copy avoids the copy operating on the original instance. The object.new plus post_init path does require post_init to be idempotent on already-populated fields — worth a quick check that subclasses whose post_init derives state from init-only arguments behave identically after a copy. Adding a subclass-with-custom-init regression test around the invoker rebinding would lock the trickiest part.

seratch:
Closing this PR in favor of https://github.com/openai/openai-agents-python/pull/4311; your contribution is included in its commit

PR #4274, opened 2026-08-07
fix: make SDK exceptions survive copy and pickle

LeSingh1:
`BaseException.__reduce__` rebuilds an exception with `type(exc)(*exc.args)`, so an exception whose `__init__` takes parameters it doesn't forward to `super().__init__` can't be reconstructed.

`ToolTimeoutError`, `InputGuardrailTripwireTriggered`, `OutputGuardrailTripwireTriggered`, `ToolInputGuardrailTripwireTriggered` and `ToolOutputGuardrailTripwireTriggered` all raise `TypeError` or `AttributeError` under `copy.copy()` and `pickle.loads()`. `ModelRefusalError` rebuilds but re-prefixes its own message, so a copied instance reads `Model refused to produce output: Model refused to produce output: ...`.

The other exceptions in the module are fine because they pass the same value to both the attribute and `super().__init__`.

This gives the six a `__reduce__` that rebuilds from the stored constructor arguments and restores the instance dict, so moving an exception across a process boundary (multiprocessing pools, task queues, workflow engines) keeps working. The guardrail tripwire errors are the ones users most often catch and re-raise, so they matter most here.

### DIFF
--- a/src/agents/exceptions.py
+++ b/src/agents/exceptions.py
@@ -118,6 +118,9 @@ class ModelRefusalError(AgentsException):
         self.refusal = refusal
         super().__init__(f"Model refused to produce output: {refusal}")
 
+    def __reduce__(self) -> tuple[Any, ...]:
+        return (self.__class__, (self.refusal,), self.__dict__)
+
 
 class UserError(AgentsException):
     """Exception raised when the user makes an error using the SDK."""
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the careful analysis. The low-level copy/pickle behavior is reproducible, but this PR does not yet demonstrate a user-impacting SDK bug. The repository does not expose or document a process-boundary execution path, and the tests only call `copy.copy` directly; the sole pickle test covers `ToolTimeoutError` rather than a real multiprocessing or task-queue flow. The guardrail exceptions also retain user callbacks, agents, context, and arbitrary output, so they cannot generally be promised to be pickleable, and this patch does not test that claim. Please provide a minimal public-API reproducer showing a real supported workflow where the original SDK exception is lost or replaced by a serialization failure. With that evidence, we can evaluate a narrower fix and an end-to-end regression test. Without it, I don't think we should add and maintain six serialization hooks.

PR #4277, opened 2026-08-07
fix: reject free-form objects in strict schemas instead of silently emptying them

abhay-codes07:
### Summary

An object schema that declares no `properties` and no `additionalProperties` accepts arbitrary keys. Strict conversion defaulted it to `additionalProperties: false`, which narrows it to "the empty object is the only valid value". The tool is then handed to the model as strict, with a parameter that can never carry any content, and nothing reports a problem.

```python
{"type": "object", "properties": {
    "target":        {"type": "string"},
    "keysAndValues": {"type": "object", "description": "key/value pairs"}}}
```

`keysAndValues` comes out of `ensure_strict_json_schema` as:

```python
{"type": "object", "description": "key/value pairs", "additionalProperties": False}
```

Validating candidate values against that:

| value | result |
| --- | --- |
| `{}` | accepted |
| `{"visible": false}` | rejected, `Additional properties are not allowed` |
| `{"a": 1, "b": "x"}` | rejected |

The parameter is required, and the only value it can hold is `{}`.

### Why this is inconsistent rather than just a strict-mode limitation

The same intent is already handled correctly when it arrives spelled differently. A Python tool annotated `dict[str, Any]` makes Pydantic emit `additionalProperties: true`, and that path raises `UserError` today:

```
UserError: additionalProperties should not be set for object types...
```

Only schemas that arrive *without* the keyword were silently narrowed. That is the common shape from MCP servers and hand-written schemas, so the loud path covers the case users are least likely to hit and the silent path covers the case they are most likely to hit.

### Fix

Treat a free-form object the same way as an explicit `additionalProperties: true` and raise `UserError`.

That choice is what makes this a real fix rather than a different error. MCP tool conversion already falls back to serving the original schema as non-strict when strict conversion raises:

```python
try:
    schema = ensure_strict_json_schema(copy.deepcopy(schema))
    is_strict = True
except Exception as e:
    logger.info("Error converting MCP schema to strict mode: %s", e)
```

So affected MCP tools go back to accepting arbitrary keys:

| | `strict_json_schema` | `{}` | `{"visible": false}` |
| --- | --- | --- | --- |
| before | `True` | accepted | **rejected** |
| after | `False` | accepted | accepted |

Function tools and output types now get an actionable error instead of a tool that silently cannot receive data.

An object counts as free-form only when it carries none of the keywords that constrain its contents, so these are unaffected:

- `properties: {}`, a no-argument tool, still converts to a strict empty object
- an explicit `additionalProperties: false` is preserved, since that is the caller stating the empty object really is the only valid value
- objects shaped by `$ref`, `allOf`, `anyOf`, `oneOf`, `patternProperties`, `propertyNames`, `enum`, `const` and friends still convert normally

### Issue number

Matches the symptom in #1681, where nested object arguments came back flattened. That report concluded the model was ignoring a correct schema. The schema the model received actually required that parameter to be empty, so there was no way to express the nested value.

### Test plan

Three tests fail without the source change and pass with it:

- `test_free_form_object_property_is_rejected_instead_of_silently_emptied`
- `test_free_form_object_root_is_rejected`
- `test_free_form_object_arg_falls_back_to_non_strict_schema`, which asserts the user-visible outcome, that the MCP tool is served non-strict and keeps accepting arbitrary keys

Seven more pin the cases that must keep working, and pass both with and without the change: empty `properties`, explicit `additionalProperties: false`, objects shaped by `allOf` / `anyOf` / `patternProperties` / `enum`, and ordinary nested objects still converting to strict.

| Command | Result |
| --- | --- |
| `uv run pytest tests/test_strict_schema.py tests/mcp/` | 124 passed |
| `make tests` | 6131 passed |

The full suite was run on Windows, where a set of sandbox symlink and tracing timing tests fail regardless of this change. I diffed the failing set against `main` at the same commit: no new failures. The one test that differed between runs, `tests/test_tracing.py::test_simple_tracing`, fails 5 out of 5 times on unmodified `main` when run in isolation, so it is part of that pre-existing set rather than something this change introduced.

### Checks

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -237,6 +237,14 @@ def create_static_tool_filter(
     return filter_dict
 
 
+def _declares_object_type(schema: dict[str, Any]) -> bool:
+    """Whether a schema states it is an object, rather than leaving its type unstated."""
+    declared_type = schema.get("type")
+    return declared_type == "object" or (
+        isinstance(declared_type, list) and "object" in declared_type
+    )
+
+
 class MCPUtil:
     """Set of utilities for interop between MCP and Agents SDK tools."""
 
@@ -549,14 +557,22 @@ class MCPUtil:
             # non-strict. Convert a separate copy so the non-strict fallback keeps
             # the original schema intact.
             strict_source = copy.deepcopy(schema)
-            if not declared_properties:
+            if not declared_properties and _declares_object_type(strict_source):
                 # Convert what the server actually sent. The synthetic ``properties: {}`` above
                 # is an OpenAI-spec accommodation, not a statement by the server that the tool
                 # takes no arguments, so leaving it in would let a free-form root be closed as
-                # an empty object instead of falling back to non-strict.
+                # an empty object instead of falling back to non-strict. Only an explicit object
+                # root is exposed this way; a schema that does not declare its type keeps the
+                # historical no-argument treatment rather than changing meaning here.
                 strict_source.pop("properties", None)
             try:
-                schema = ensure_strict_json_schema(strict_source)
+                converted = ensure_strict_json_schema(strict_source)
+                if "properties" not in converted:
+                    # Restore the shape the OpenAI spec wants. Only reachable when the root was
+                    # already closed by the server, e.g. ``additionalProperties: false``.
+                    converted["properties"] = {}
+                    converted.setdefault("required", [])
+                schema = converted
                 is_strict = True
             except Exception as e:
                 if _debug.DONT_LOG_TOOL_DATA:
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -237,12 +237,13 @@ def create_static_tool_filter(
     return filter_dict
 
 
-def _declares_object_type(schema: dict[str, Any]) -> bool:
-    """Whether a schema states it is an object, rather than leaving its type unstated."""
+def _is_strict_object_root(schema: dict[str, Any]) -> bool:
+    """Whether a converted schema is the closed object envelope a strict tool needs."""
     declared_type = schema.get("type")
-    return declared_type == "object" or (
+    is_object = declared_type == "object" or (
         isinstance(declared_type, list) and "object" in declared_type
     )
+    return is_object and schema.get("additionalProperties") is False
 
 
 class MCPUtil:
@@ -557,16 +558,22 @@ class MCPUtil:
             # non-strict. Convert a separate copy so the non-strict fallback keeps
             # the original schema intact.
             strict_source = copy.deepcopy(schema)
-            if not declared_properties and _declares_object_type(strict_source):
+            if not declared_properties:
                 # Convert what the server actually sent. The synthetic ``properties: {}`` above
                 # is an OpenAI-spec accommodation, not a statement by the server that the tool
-                # takes no arguments, so leaving it in would let a free-form root be closed as
-                # an empty object instead of falling back to non-strict. Only an explicit object
-                # root is exposed this way; a schema that does not declare its type keeps the
-                # historical no-argument treatment rather than changing meaning here.
+                # takes no arguments, so leaving it in would let a free-form root, including one
+                # reached through a composed root such as ``allOf``, be closed as an empty
+                # object instead of falling back to non-strict.
                 strict_source.pop("properties", None)
             try:
                 converted = ensure_strict_json_schema(strict_source)
+                if not _is_strict_object_root(converted):
+                    # Without the shim a root that never declared its type converts to something
+                    # that is not a strict object envelope. Serve the original instead of
+                    # sending the provider a strict schema it cannot use.
+                    raise UserError(
+                        "strict conversion did not produce an object root for this MCP schema"
+                    )
                 if "properties" not in converted:
                     # Restore the shape the OpenAI spec wants. Only reachable when the root was
                     # already closed by the server, e.g. ``additionalProperties: false``.
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -583,6 +583,14 @@ class MCPUtil:
                         "strict conversion did not produce an object root for this MCP schema"
                     )
                 if "properties" not in converted:
+                    stale_required = converted.get("required")
+                    if isinstance(stale_required, list) and stale_required:
+                        # The root requires keys it never declares. Restoring the empty
+                        # properties shim would serve a strict schema whose required names
+                        # are all forbidden, so serve the original non-strict instead.
+                        raise UserError(
+                            "strict conversion left required names with no declared properties"
+                        )
                     # Restore the shape the OpenAI spec wants. Only reachable when the root was
                     # already closed by the server, e.g. ``additionalProperties: false``.
                     converted["properties"] = {}
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -238,19 +238,14 @@ def create_static_tool_filter(
 
 
 def _is_strict_object_root(schema: dict[str, Any]) -> bool:
-    """Whether a converted schema is the closed object envelope a strict tool needs.
-
-    A root that never declared its type but is already closed still describes an object, so it
-    is normalized here rather than rejected.
-    """
+    """Whether a converted schema is the closed object envelope a strict tool needs."""
     if schema.get("additionalProperties") is not False:
         return False
     declared_type = schema.get("type")
-    if declared_type is None:
-        schema["type"] = "object"
-        return True
-    return declared_type == "object" or (
-        isinstance(declared_type, list) and "object" in declared_type
+    return (
+        declared_type is None
+        or declared_type == "object"
+        or (isinstance(declared_type, list) and "object" in declared_type)
     )
 
 
@@ -575,6 +570,11 @@ class MCPUtil:
                 strict_source.pop("properties", None)
             try:
                 converted = ensure_strict_json_schema(strict_source)
+                if _is_strict_object_root(converted) and converted.get("type") is None:
+                    # A root that never declared its type but is already closed still
+                    # describes an object; normalize it explicitly here, on the copy this
+                    # branch owns, rather than inside the predicate.
+                    converted["type"] = "object"
                 if not _is_strict_object_root(converted):
                     # Without the shim a root that never declared its type converts to something
                     # that is not a strict object envelope. Serve the original instead of
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -238,12 +238,20 @@ def create_static_tool_filter(
 
 
 def _is_strict_object_root(schema: dict[str, Any]) -> bool:
-    """Whether a converted schema is the closed object envelope a strict tool needs."""
+    """Whether a converted schema is the closed object envelope a strict tool needs.
+
+    A root that never declared its type but is already closed still describes an object, so it
+    is normalized here rather than rejected.
+    """
+    if schema.get("additionalProperties") is not False:
+        return False
     declared_type = schema.get("type")
-    is_object = declared_type == "object" or (
+    if declared_type is None:
+        schema["type"] = "object"
+        return True
+    return declared_type == "object" or (
         isinstance(declared_type, list) and "object" in declared_type
     )
-    return is_object and schema.get("additionalProperties") is False
 
 
 class MCPUtil:
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -538,7 +538,8 @@ class MCPUtil:
         schema, is_strict = copy.deepcopy(tool_input_schema(tool)), False
 
         # MCP spec doesn't require the inputSchema to have `properties`, but OpenAI spec does.
-        if "properties" not in schema:
+        declared_properties = "properties" in schema
+        if not declared_properties:
             schema["properties"] = {}
 
         if convert_schemas_to_strict:
@@ -547,8 +548,15 @@ class MCPUtil:
             # ``additionalProperties: false``) on a schema we still serve as
             # non-strict. Convert a separate copy so the non-strict fallback keeps
             # the original schema intact.
+            strict_source = copy.deepcopy(schema)
+            if not declared_properties:
+                # Convert what the server actually sent. The synthetic ``properties: {}`` above
+                # is an OpenAI-spec accommodation, not a statement by the server that the tool
+                # takes no arguments, so leaving it in would let a free-form root be closed as
+                # an empty object instead of falling back to non-strict.
+                strict_source.pop("properties", None)
             try:
-                schema = ensure_strict_json_schema(copy.deepcopy(schema))
+                schema = ensure_strict_json_schema(strict_source)
                 is_strict = True
             except Exception as e:
                 if _debug.DONT_LOG_TOOL_DATA:
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -126,7 +126,14 @@ _CONTENT_SHAPING_KEYWORDS = frozenset(
 def _is_unclosable_object(json_schema: dict[str, object]) -> bool:
     """Whether closing this object with ``additionalProperties: false`` would change meaning."""
     typ = json_schema.get("type")
-    is_object = typ == "object" or (is_list(typ) and "object" in typ)
+    declared_properties = json_schema.get("properties")
+    # A typeless schema with a properties map is normalized to an object by the conversion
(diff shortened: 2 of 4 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying issue is valid, but the current presence-based object-shape whitelist still silently changes schema meaning. Please revise the conversion so an object that has no properties declared at that object level and no explicit additionalProperties: false falls back to non-strict mode after any already-supported normalization. Do not treat anyOf, multi-branch allOf, enum, or const as sufficient evidence that the wrapper can be closed.

Please also preserve whether MCP properties were declared before MCPUtil adds its synthetic properties: {}, so a root free-form MCP schema takes the same fallback. Add regression coverage for a root MCP object, a composed wrapper, and a non-empty object enum/const, asserting that their original non-strict meaning is preserved.

abhay-codes07:
Thanks, both points were right and both are fixed in 7bc786e.

**The whitelist did still change meaning.** I checked the cases you named before changing anything, and each one accepted a value before conversion and rejected it after:

| wrapper | value | original | previous revision |
| --- | --- | --- | --- |
| `anyOf: [{properties: {a}}]` | `{"a": "x"}` | accept | **reject** |
| multi-branch `allOf` | `{"a": "x", "b": "y"}` | accept | **reject** |
| `enum: [{a: 1}]` | `{"a": 1}` | accept | **reject** |
| `const: {a: 1}` | `{"a": 1}` | accept | **reject** |

`patternProperties` and `propertyNames` have the same defect for any key the pattern does not match, so I folded them in rather than special-casing them.

The rule is now the one you described: an object is closed only when it declares `properties` at its own level or already carries an explicit `additionalProperties: false`. Nothing else counts as evidence.

To keep the already-supported normalizations working, the check moved to the end of the conversion instead of running before the composition keywords are handled. A `$ref` and a single-entry `allOf` both lift `properties` up to this level and re-enter the function, so they still reach strict mode through the existing re-recursion. Verified explicitly:

| schema | result |
| --- | --- |
| `{type: object}` | fallback |
| `anyOf` / multi `allOf` / `enum` / `const` / `patternProperties` / `propertyNames` wrapper | fallback |
| `{type: object, properties: {...}}` | strict |
| `{type: object, properties: {}}` | strict |
| `{type: object, additionalProperties: false}` | strict |
| single-entry `allOf` carrying `properties` | strict |
| `$ref` to an object, with a sibling `description` | strict |

**The synthetic `properties: {}` was masking a free-form MCP root.** Confirmed: `{"type": "object"}` was being closed into a no-argument tool. `to_function_tool` now records whether the server declared `properties` and runs strict conversion against what the server actually sent, keeping the synthetic key only on the schema it serves.

| MCP `inputSchema` | strict | served |
| --- | --- | --- |
| `{"type": "object"}` | `False` | unchanged, arbitrary keys still valid |
| `{"type": "object", "properties": {}}` | `True` | strict no-argument tool |
| `{}` | `True` | unchanged, existing empty-schema special case |
| normal root with properties | `True` | strict |

**Regression coverage** for the three you asked for, plus the normalization cases so the fallback cannot quietly widen later:

- `test_free_form_root_schema_falls_back_to_non_strict` and `test_declared_empty_properties_root_still_converts_to_strict` for the MCP root
- `test_composed_object_wrappers_are_not_closed`, parametrized over all six wrapper forms, and `test_composed_and_enum_object_args_keep_their_non_strict_meaning` on the MCP side, which asserts the served schema is byte-for-byte what the server sent
- `test_single_entry_allof_is_normalized_before_the_check` and `test_ref_to_an_object_is_normalized_before_the_check`

Ten of these fail on the previous revision and pass now. I also removed the test from that revision which asserted `anyOf`/`enum`/`patternProperties` wrappers should be closed, since it encoded the behaviour you flagged.

| Command | Result |
| --- | --- |
| `uv run pytest tests/test_strict_schema.py tests/mcp/` | 133 passed |
| `make tests` | 6135 passed, no new failures against `main` at the same commit |

abhay-codes07:
The automated review found three real regressions in the previous revision. All three reproduced, and all three are fixed in 92f9da1.

**1. A redundant single-entry `allOf` blocked conversion.** The branch was converted before being merged, so it was judged on its own even though the parent already declared the properties:

```python
{"type": "object", "properties": {"a": {"type": "string"}}, "allOf": [{"type": "object"}]}
```

The bare `{"type": "object"}` branch looked free-form, so an otherwise strictable object became a conversion error and fell back to non-strict. The branch is now merged first and the merged schema re-entered, which is where the existing normalization already expected to happen. That schema converts again, with `properties`, `required: ["a"]` and `additionalProperties: false`.

**2. The MCP `properties` shim was lost on strict success.** Popping the synthetic key before conversion also removed it from the served schema when conversion succeeded, so a server root that is already closed came back without the key the OpenAI tool schema shape requires:

| input schema | before | after |
| --- | --- | --- |
| `{"type": "object", "additionalProperties": false}` | `{"type": "object", "additionalProperties": false}`, no `properties` | `properties: {}`, `required: []`, `additionalProperties: false` |

The shim is restored after a successful conversion, which is only reachable when the server already closed the root.

**3. Typeless roots were exposed to the object-only check.** The same pop meant an annotation-only schema never reached the new branch and was served strict without an object envelope:

| input schema | before | after |
| --- | --- | --- |
| `{"description": "Test tool"}` | strict, served as `{"description": "Test tool"}` | strict, `type: object`, `properties: {}`, `additionalProperties: false` |

Only a root that declares `type: object` is now exposed as free-form. A schema that never states its type keeps its historical no-argument treatment, since changing what those mean is a separate question from the one this PR is fixing.

The free-form behaviour you asked for is unchanged by these fixes:

| MCP `inputSchema` | strict | note |
| --- | --- | --- |
| `{"type": "object"}` | `False` | falls back, arbitrary keys still valid |
| `{"type": "object", "properties": {}}` | `True` | declared no-argument tool |
| `{"type": "object", "additionalProperties": false}` | `True` | server closed it, shim restored |
| `{"description": "..."}` / `{}` | `True` | typeless, historical treatment |
| composed wrapper, object `enum`/`const` | `False` | falls back |

Three new tests, each failing on the previous revision: `test_redundant_single_entry_allof_branch_does_not_block_conversion`, `test_already_closed_root_keeps_the_openai_properties_shape`, and `test_typeless_root_keeps_its_no_argument_treatment` (parametrized over the annotation-only and empty roots).

| Command | Result |
| --- | --- |
| `uv run pytest tests/test_strict_schema.py tests/mcp/` | 430 passed, 24 skipped, plus the 4 `test_mcp_tracing` failures that are pre-existing on `main` in this environment |
| `make tests` | 6137 passed, no new failures against `main` at the same commit |

abhay-codes07:
All Codex threads are now addressed and resolved. Eight findings across two rounds, each reproduced before being fixed.

**Definitions are no longer judged in isolation.** This was the serious one: the rejection ran while walking `$defs`, so ordinary schemas broke.

| schema | before | after |
| --- | --- | --- |
| broad base narrowed at the `$ref` site | error / non-strict | strict |
| unreferenced free-form definition | error / non-strict | strict |

A definition is a template. A broad base is routinely narrowed by the keys a `$ref` site supplies, and an unreferenced one has no validation effect at all, so the check now runs at the use site where the merged shape is known.

**A bare `$ref` to a free-form definition is still rejected.** That was the gap left by the exemption above: a bare `$ref` is never inlined, so it would have kept pointing at a definition the walk had since closed into the empty object. Definitions that cannot be closed are recorded before the walk, and a bare `$ref` to one fails conversion.

**Empty `properties` is no longer read as "this object is empty"** when something beside it says otherwise. `patternProperties`, `propertyNames`, `enum`, `const`, `maxProperties` and `minProperties` all mean keys are still accepted, so closing rejected valid values.

**`maxProperties: 0` converts instead of being rejected**, since it already forbids every key and closing it changes nothing.

**MCP roots.** The synthetic `properties: {}` is now always removed before conversion, not only for roots that declare `type: object`, so a composed root keeps its meaning. Because that also exposes roots that never declare an object, a successful conversion must produce a closed object envelope, and a root that is closed but typeless is normalized rather than rejected.

| MCP `inputSchema` | strict | why |
| --- | --- | --- |
| `{"type": "object"}` | no | free-form |
| `{"allOf": [{"type": "object"}]}` | no | composed free-form root |
| `{"allOf": [{"type": "object", "enum": [...]}]}` | no | closing it would make the enum unreachable |
| `{"description": "..."}` | no | never declares an object |
| `{"type": "object", "properties": {}}` | yes | declared no-argument tool |
| `{"additionalProperties": false}` | yes | closed, normalized to an envelope |
| `{}` | yes | existing empty-schema case |
| normal root with properties | yes | |

Sixteen tests now cover this, thirteen of which fail on the preceding commits. Full suite 6404 passed, no new failures against `main` at the same commit; lint, mypy and pyright clean.

One thing worth saying plainly: this PR has needed four rounds of correction, and every round was a case where my change rejected or narrowed a schema that was previously fine. The blast radius is wider than the one-line bug it started from. If you would rather not carry that risk, closing it is a reasonable call and I will not push back.

anujbolewar (commented):
Good — a closed root that never declared its type is still an object, and normalizing it avoids the empty-schema rejection. Handles the list-of-types case cleanly. One caution: _is_strict_object_root mutates the schema dict in place via the included property assignment, so wrap it expects a freshly copied schema and never shares that dict with a readonly reference — going from deepcopy on the caller it is safe, just worth stating so future callers do not pass a shared ref.

abhay-codes07:
The four remaining Codex threads are fixed in 9e95a48 and resolved.

Three of them were holes in the ref-tracking added by the previous commit, and the root cause was structural rather than three separate oversights: the tracked set was its own parameter, so any recursion site that missed it silently dropped the information. The registry now rides on the conversion budget, which already reaches every recursive call because the DoS bound depends on it, and records definitions by object identity, so nested `$defs` need no path bookkeeping. That closes the `allOf`-branch and nested-`$defs` holes by construction rather than by patching call sites.

The sibling case is handled at the merge site, since the merged result inherits the `additionalProperties: false` the definition walk added and the final decision cannot see the problem there:

| `$ref` to a free-form definition with... | result |
| --- | --- |
| `properties: {}` | fallback |
| annotation only (`description`) | fallback |
| real properties | strict, keys required |

And empty-object literals are now recognized as already closed: `const: {}` and `enum: [{}]` admit only the empty object, so closing them preserves meaning exactly, like `maxProperties: 0`. `const: {"a": 1}` still falls back.

Seven new tests, six failing on the previous commit; the seventh pins the real-properties case that must keep converting. I also re-ran the full 23-case behavioural sweep across every scenario raised on this PR so far: all as expected. Full suite 6410 passed, no new failures against `main` at the same commit; lint, mypy, pyright clean.

abhay-codes07:
The four round-six threads are fixed in 339c1ce and resolved.

**Free-form fields inside referenced definitions.** The carve-out was propagating into every position inside a definition, exactly as flagged. A definition whose interior holds a free-form node at a value position is now recorded as tainted, and since a sibling merge only reshapes the top level of a referenced subtree, both the bare-ref and merge sites reject references to tainted definitions unconditionally. `{"$defs": {"M": {..., "x": {"type": "object"}}}}` referenced with or without siblings now falls back, while the same definition left unreferenced stays harmless, preserving the round-five behaviour.

**Alias chains.** `_resolve_ref_chain` follows definitions that are themselves a lone `$ref`, with a seen-set to stop cycles, so `Outer -> Inner` aliases are judged by what `Inner` is: free-form target falls back, shaped target converts.

**Undeclared `required` names.** Conversion overwrote `required` with the property list, so `{"properties": {}, "required": ["a"]}` silently became a schema that forbids the very key it required. It now rejects before the overwrite. The check skips nodes carrying a `$ref`, since their effective properties only exist after the merge, and the merged result is re-checked on re-entry.

**Dependency keywords.** `dependentRequired` and draft-07 `dependencies` now count as content-shaping, so an empty `properties` map beside them is not read as "this object is empty".

Six new tests, five failing on the previous commit; the sixth pins the unreferenced-tainted-definition case that must keep converting. The cumulative behavioural sweep is now 25 cases covering every scenario raised across all six review rounds, all passing. Full suite 6416 passed, no new failures against `main` at the same commit; lint, mypy, pyright clean.

abhay-codes07:
The round-seven thread is fixed in 08b0094 and resolved, taking the suggestion as stated: the registries are now precomputed order-independently.

Before anything is walked or mutated, every definition entry in the tree is collected, seeded from what it says on its own, and alias and interior references are iterated to a fixpoint. Declaration order cannot matter because nothing has been inlined or closed yet when the registries are read.

Verifying the reported case and its neighbours:

| case | before | after |
| --- | --- | --- |
| annotated alias declared before its free-form target, bare ref to alias | strict, accepts only `{}` | fallback |
| same alias declared after the target | fallback | fallback |
| unreferenced annotated alias, either order | convert | convert |
| bare alias after target, unreferenced | **conversion failure** | convert |
| alias to a shaped target | convert | convert |

The fourth row is a latent inverse the walk-time approach also had: order dependence cut both ways, and the reversed declaration wrongly failed a schema whose alias nothing referenced. With the registries authoritative up front, the walk-time checks inside definitions no longer raise at all; they record the enclosing definition as tainted, so a template is only a problem once something references it, in either order.

Four new tests, two failing on the previous commit and two pinning the harmless cases. The cumulative sweep is now 30 cases across all seven rounds, all passing. Full suite 6425 passed, no new failures; ruff, full mypy and pyright clean.

abhay-codes07:
The three round-eight threads are fixed in d4a7134 and resolved, along with the review note about the mutating predicate.

**Literals are no longer read as schemas.** The precomputed interior walk descended into `const`, `enum`, `default` and `examples` values, so a data literal shaped like an open schema tainted its definition. Those keywords hold data, not schemas, and the walk now skips their values in all three traversals. `{"const": {"type": "object"}}` inside a referenced definition converts again.

**Required judgment waits for pending merges.** With a `$ref` or single-entry `allOf` present, both the required-subset check and the `required` overwrite are deferred until the merge lands and re-enters. That fixes the flagged false rejection and a second problem next to it: the early overwrite destroyed the original `required` before the merged shape could be judged, so a branch that does not supply the key previously slipped through.

| schema | result |
| --- | --- |
| `properties: {}, required: ["a"], allOf: [{properties: {a}}]` | strict, `required: ["a"]` |
| same, branch supplies only `b` | fallback |

**Typeless definitions with a properties map** are classified as objects by the precomputed registries, mirroring the normalization the walk applies, so `{"properties": {}, "propertyNames": ...}` referenced bare falls back instead of closing into an empty object.

**@anujbolewar** on the mutating predicate: agreed, and fixed in the same commit. `_is_strict_object_root` is now a pure check; the typeless-root normalization happens explicitly at the call site, on the deep copy that branch owns, so the predicate can be called on shared dicts safely.

Seven new tests, five failing on the previous commit and two pinning behaviour that must not change. The cumulative sweep is 36 cases across eight review rounds, all passing. Full suite 6426 passed, no new failures; ruff, full mypy and pyright clean.

abhay-codes07:
The four round-nine threads are fixed in 68b901e and resolved.

The first two shared one root cause worth naming, because it was my mistake rather than four separate ones: when I added the precomputed registries last round, I left the **original walk-time seeding in place beside them**. That stale seeding classified each definition from its raw shape before the converter's merges could run, so it overwrote the precompute's more careful view. Removing it makes the precompute the single source of truth, and both reported cases fall out:

| definition | before | after |
| --- | --- | --- |
| `{"type": "object", "allOf": [{"properties": {"a": ...}}]}` | fallback | strict |
| `{"type": "object", "$ref": "#/$defs/S", "description": "d"}` | fallback | strict |

Classification now happens on the shape the converter will actually produce: `_effective_definition_view` mirrors the single-entry `allOf` merge and the `$ref`-with-siblings inline on a copy, with the same sibling-priority rule as the real merge and no mutation.

**Singular `example`** joins `const`, `default`, `enum` and `examples` as literal-valued. Only the plural was skipped, so an object literal under `example` was walked as a schema and tainted its definition despite having no validation effect.

**MCP closed root with undeclared `required`** is no longer served strict. Restoring the synthetic empty `properties` while keeping `required: ["a"]` produced a schema forbidding the very key it required, so the original is served non-strict.

Four new tests, all failing on the previous commit. The cumulative sweep is 42 cases across nine rounds, all passing. Full suite 6429 passed, no new failures against `main` at the same commit; ruff, mypy and pyright clean.

On process: nine rounds is a lot, and the pattern in the last two was me layering new machinery without removing what it replaced. I have gone back over the file for other leftovers of that kind and found none, but that is exactly the sort of thing worth your scrutiny if you review this.

seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/4310

PR #3514, opened 2026-05-27
fix(session): strip reasoning item ids from session history

Epochex:
Fixes #2020

Session-backed runs can end up replaying `reasoning` items with server-assigned `rs_...` IDs from prior turns. Those IDs are not guaranteed to be stable across turns, and replaying them can trigger a 404 ("Item with id 'rs_...' not found") in a subsequent `responses.create` call.

This change strips the `id` field from `reasoning` items loaded from non-Conversations sessions before building model input, so the reasoning payload can still be replayed without relying on server-side item retention.

Tests:
- uv run ruff format src/agents/run_internal/session_persistence.py tests/test_agent_runner.py
- uv run ruff check src/agents/run_internal/session_persistence.py tests/test_agent_runner.py
- uv run pyright --project pyrightconfig.json src/agents/run_internal/session_persistence.py tests/test_agent_runner.py
- uv run pytest tests/test_agent_runner.py -q

PR #4278, opened 2026-08-07
fix(sessions): apply the reasoning item id policy to stored session history

### LINKED ISSUE #2020: Item with id `rs_` not found

LeSingh1:
`save_result_to_session` applies `reasoning_item_id_policy` when writing run items to a non-Conversations session. `prepare_input_with_session` never applies it when reading that history back.

Reasoning ids recorded before the caller set `reasoning_item_id_policy="omit"` are therefore replayed on every later turn of the same conversation. The documented mitigation cannot rescue a session that already holds a stale `rs_...` id — the row is never rewritten, and it survives restarts in the SQLite file.

That is the shape reported in #2020: a triage agent hands off, its empty-summary reasoning item lands in a `SQLiteSession`, the server later stops resolving the id, and every subsequent turn fails with `404 Item with id 'rs_...' not found`.

Apply the policy to session history on read too, skipping `OpenAIConversationsSession` exactly as the write path does. Runs that leave `reasoning_item_id_policy` unset are unchanged.

Two things worth flagging rather than leaving you to find them:

- This does not close #2020. The reporters were not using `omit`, so it does not make the default `preserve` path safe — it makes the documented mitigation actually work on a conversation that is already poisoned. Marking it `Refs` rather than `Fixes` for that reason.
- I checked the broader claim first and could not support it. With `omit` set from the start, I could not leak a reasoning id through any public path I tried: session + handoff, streamed and non-streamed, nested handoff history, `handoff_input_filter`, `session_input_callback`, `SessionSettings(limit=...)`, approval interrupt with `RunState` resume, `max_turns`, `agent.as_tool()`, and `to_input_list()` round-trip. So the unconditional-strip approach in #3514/#3582/#4013 still looks wrong, and this is deliberately narrower.

The `run.py` change is mechanical — it hoists the existing policy resolution above the session prep so it can be passed in. `run_state` is always `None` at those two call sites, so it could pass `run_config.reasoning_item_id_policy` directly and shrink that file to two lines. I went with the version that does not rely on an unstated invariant; happy to switch.

Tests: a non-streamed and a streamed regression test drive `Runner.run` / `Runner.run_streamed` with a real `SQLiteSession` and a real handoff against a model that raises `openai.NotFoundError` on the revoked id. Both fail before the fix with the error from the issue rather than an internal assertion.

Refs #2020

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -523,7 +523,16 @@ class AgentRunner:
         run_config = RunConfig() if run_config is None else _coerce_run_config(run_config)
 
         is_resumed_state = isinstance(input, RunState)
-        run_state: RunState[TContext] | None = None
+        run_state: RunState[TContext] | None = (
+            cast(RunState[TContext], input) if is_resumed_state else None
+        )
+        resolved_reasoning_item_id_policy: ReasoningItemIdPolicy | None = (
+            run_config.reasoning_item_id_policy
+            if run_config.reasoning_item_id_policy is not None
+            else (run_state._reasoning_item_id_policy if run_state is not None else None)
+        )
+        if run_state is not None:
+            run_state._reasoning_item_id_policy = resolved_reasoning_item_id_policy
         starting_input = input if not is_resumed_state else None
         original_user_input: str | list[TResponseInputItem] | None = None
         session_input_items_for_persistence: list[TResponseInputItem] | None = (
@@ -533,8 +542,7 @@ class AgentRunner:
         # exactly those items (and not the full history).
         last_saved_input_snapshot_for_rewind: list[TResponseInputItem] | None = None
 
-        if is_resumed_state:
-            run_state = cast(RunState[TContext], input)
+        if is_resumed_state and run_state is not None:
             (
                 conversation_id,
                 previous_response_id,
@@ -590,6 +598,7 @@ class AgentRunner:
                     run_config.session_settings,
                     include_history_in_prepared_input=False,
                     preserve_dropped_new_items=True,
+                    reasoning_item_id_policy=resolved_reasoning_item_id_policy,
                     wrapper=context_wrapper,
                 )
                 original_input_for_state = raw_input
@@ -603,17 +612,10 @@ class AgentRunner:
                     session,
                     run_config.session_input_callback,
                     run_config.session_settings,
+                    reasoning_item_id_policy=resolved_reasoning_item_id_policy,
                     wrapper=context_wrapper,
                 )
                 original_input_for_state = prepared_input
-
-        resolved_reasoning_item_id_policy: ReasoningItemIdPolicy | None = (
-            run_config.reasoning_item_id_policy
-            if run_config.reasoning_item_id_policy is not None
-            else (run_state._reasoning_item_id_policy if run_state is not None else None)
-        )
-        if run_state is not None:
-            run_state._reasoning_item_id_policy = resolved_reasoning_item_id_policy
 
         # Check whether to enable OpenAI server-managed conversation
         if (
--- a/src/agents/run_internal/items.py
+++ b/src/agents/run_internal/items.py
@@ -60,6 +60,7 @@ __all__ = [
     "REJECTION_MESSAGE",
     "TOOL_CALL_SESSION_DESCRIPTION_KEY",
     "TOOL_CALL_SESSION_TITLE_KEY",
+    "apply_reasoning_item_id_policy",
     "copy_input_items",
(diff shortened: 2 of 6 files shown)


seratch (changes requested):
The released `reasoning_item_id_policy="omit"` behavior covers SDK-built session follow-up input, but the current read path leaves reasoning IDs that were stored before the option was enabled. Applying the policy to normalized non-Conversations session history is the right narrow fix, while keeping the default and explicit `"preserve"` behavior unchanged. The streaming and non-streaming regression tests cover the relevant runner paths.

Before we can merge this, please remove the unused `# type: ignore[override]` comments from `_RevokedReasoningIdModel.get_response` and `_StreamedRevokedReasoningIdModel.stream_response`. The current typecheck job fails mypy with `unused-ignore` at `tests/test_agent_runner.py:1241` and `tests/test_agent_runner_streamed.py:841`. Once typecheck is green, this should be ready for another review.

The current `Refs #2020` scope is correct: this repairs the opt-in recovery path but does not resolve the default `"preserve"` failure reported there.

LeSingh1:
Removed both in 1c8beac. The two fake models override `Model.get_response`/`stream_response` compatibly, so the ignores were never needed.

`mypy` clean on both files, 282 passed, ruff check and format clean.

seratch (approved):

PR #4299, opened 2026-08-08
fix(memory): preserve falsey compaction decision hooks

hsusul:
### Summary

`OpenAIResponsesCompactionSession` used truthiness to select its compaction decision hook, so a valid callable object with falsey boolean behavior was silently replaced by the default hook.

Preserve every explicitly supplied hook by falling back only when `should_trigger_compaction` is `None`. Add an offline regression test that verifies a falsey hook is invoked and can prevent compaction at the default threshold.

This is distinct from #4293, which changes which session history is compacted rather than how the decision hook is selected.

### Test plan

- Red before the fix: `uv run pytest tests/memory/test_openai_responses_compaction_session.py -q -k falsey_decision_hook` (`1 failed, 48 deselected`)
- Green after the fix: `uv run pytest tests/memory/test_openai_responses_compaction_session.py -q -k falsey_decision_hook` (`1 passed, 48 deselected`)
- `env UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh` (format, lint, typecheck, and tests all passed)

No API key or live model call is required.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/memory/openai_responses_compaction_session.py
+++ b/src/agents/memory/openai_responses_compaction_session.py
@@ -125,7 +125,9 @@ class OpenAIResponsesCompactionSession(SessionABC, OpenAIResponsesCompactionAwar
         self.model = model
         self.compaction_mode = compaction_mode
         self.should_trigger_compaction = (
-            should_trigger_compaction or default_should_trigger_compaction
+            should_trigger_compaction
+            if should_trigger_compaction is not None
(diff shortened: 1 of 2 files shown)


seratch (approved):
Thanks for catching this several times. I am currently working on detection mechanism for this one. So, please hold off sending the similar ones further.

PR #4300, opened 2026-08-08
fix(handoffs): preserve message name when nesting handoff history

hsusul:
### Summary

When `nest_handoff_history` (nested handoff history) summarizes the previous transcript, each turn is rendered as a compact `role (name): content` line. On a subsequent handoff the summary is flattened back into structured items by `_extract_nested_history_transcript`, which reverses that format.

That compact form is ambiguous when the `role` or the message `name` contains one of the characters the parser splits on:

- `:` is consumed by the `role: content` split.
- `(` / `)` confuse the `(name)` extraction (which uses `rfind("(")`).

So a message whose `name` contained those characters was silently corrupted on the next handoff. For example, with `{"role": "assistant", "name": "a:b", "content": "answer here"}`:

- before: `{"role": "assistant (a", "content": "b): answer here"}` — the role, name, and content are all wrong
- after: the item round-trips unchanged

The SDK already renders items losslessly via the JSON encoding used for content that contains newlines. This change routes items to that JSON encoding whenever the compact line would not be reversible (role/name containing `:`, `(`, `)`, `\n`, or `\r`), and keeps the readable compact form for everything else (e.g. a plain `name` like `"bob"`, which is unaffected).

The fix is confined to `_format_transcript_item`'s legacy-vs-JSON choice; parsing is unchanged. The JSON path is already exercised by existing tests, and the change is idempotent across repeated handoffs.

### Test plan

Added `test_nested_history_preserves_message_name`, parametrized over names `bob`, `O(1)`, `a:b`, `foo(bar)`, `x)y`. It nests a 3-turn history whose middle message carries the `name`, then re-flattens across several handoffs and asserts the transcript is unchanged each time.

- Fails on `main` for the `colon`, `parens`, and `trailing_call` cases (role/name/content corrupted); passes with this change.

Commands run (from a clean `upstream/main` worktree):

- `uv run pytest tests/test_handoff_history_duplication.py -q` → 97 passed
- `uv run pytest tests/test_handoff_history_duplication.py tests/test_run_internal_items.py tests/test_handoff_tool.py tests/test_run_step_processing.py -q` → 198 passed
- `uv run ruff check src/agents/handoffs/history.py tests/test_handoff_history_duplication.py` → All checks passed
- `uv run ruff format --check …` → already formatted
- `uv run pyright src/agents/handoffs/history.py` → 0 errors
- `TYPECHECK_SRC_ONLY=1 uv run mypy src --exclude site` → no new errors in `handoffs/history.py`; the 44 reported errors are pre-existing on `main` in unrelated optional-dependency modules (`extensions/sandbox/*`, `extensions/models/litellm_model.py`, `extensions/memory/sqlalchemy_session.py`, `voice/imports.py`, etc.).

### Issue number

N/A

### Checks

- [x] I've confirmed all verification steps pass (ran the ruff / pyright / mypy / pytest commands listed above individually rather than via `run.sh`)

### DIFF
--- a/src/agents/handoffs/history.py
+++ b/src/agents/handoffs/history.py
@@ -403,13 +403,33 @@ def _format_transcript_item(item: TResponseInputItem) -> str:
     role = item.get("role")
     if isinstance(role, str):
         content = item.get("content")
-        if content is None or (isinstance(content, str) and not _contains_newline(content)):
+        if (
+            content is None or (isinstance(content, str) and not _contains_newline(content))
+        ) and _legacy_summary_line_is_reversible(item):
             return _format_transcript_item_legacy(item)
     return _format_transcript_item_json(item)
 
 
 def _contains_newline(value: str) -> bool:
     return "\n" in value or "\r" in value
+
+
+# Characters the compact ``role (name): content`` summary line splits on. A role or name that
+# contains any of them cannot be recovered from that form: a ":" is consumed by the
+# ``role: content`` split, and "(" / ")" confuse the ``(name)`` extraction. The item is encoded
+# as JSON instead, which round-trips losslessly.
+_LEGACY_SUMMARY_DELIMITERS = (":", "(", ")", "\n", "\r")
+
+
+def _legacy_summary_line_is_reversible(item: TResponseInputItem) -> bool:
+    """Return whether the legacy compact summary form round-trips for this item."""
+    role = item.get("role")
+    if not isinstance(role, str) or any(char in role for char in _LEGACY_SUMMARY_DELIMITERS):
+        return False
+    name = item.get("name")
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the careful investigation and focused patch. I confirmed the direct helper round-trip failure, but the PR does not establish a supported runtime path for the affected input. TResponseInputItem uses the Responses input-item schema, whose message shapes do not include name, and the Chat Completions converter rejects the same named message before it can reach nested handoff handling. The regression test therefore demonstrates a cast-only helper case, not user-visible SDK impact.

I am going to close this PR for now. If you can provide an end-to-end reproduction using a supported Runner input and provider path that accepts a message name containing these delimiters and demonstrates the downstream effect, we can revisit the narrow serialization fix.

PR #4301, opened 2026-08-08
fix(docs): clarify AgentHooks description in Chinese translation

scrazyakai:
`"自定义副作用"` can be misleading, while `"自定义Hooks"` is easier to understand.


seratch:
Thanks for the feedback! It's auto translated, so could you suggest the best way to map English and Chinese terms? https://github.com/openai/openai-agents-python/blob/main/docs/scripts/translate_docs.py

seratch:
Thanks again for sharing this feedback! https://github.com/openai/openai-agents-python/pull/4306 improves the clarity of the source English text and basic policies for future ones. After merging the PR, I will re-generate all affected pages using the new source docs. We'll continue improving, so please feel free to share feedback at any time!

PR #4303, opened 2026-08-08
fix: stop merging anyOf and oneOf into a single widened union

abhay-codes07:
### Summary

When a schema node carries both `anyOf` and `oneOf`, strict conversion folds the `oneOf` variants into the existing `anyOf` list. Those are two constraints that must BOTH hold, so the fold turns a conjunction into a union, and the converted schema accepts values the original rejects.

Checked against a JSON Schema validator:

```python
{"anyOf": [{"type": "string"}, {"type": "integer"}],
 "oneOf": [{"type": "string"}, {"type": "boolean"}]}
```

| value | original | after conversion |
| --- | --- | --- |
| `"hello"` | accept | accept |
| `5` | reject (fails the `oneOf`) | **accept** |
| `true` | reject (fails the `anyOf`) | **accept** |
| `3.14` | reject | reject |

For a function tool or output type this widens the declared contract: the provider constrains generation to the converted schema, so the model can produce arguments or outputs the original schema forbids, and nothing downstream notices.

The prior expectation in `test_oneof_merged_with_existing_anyof` makes the problem concrete: its input schema accepts **no value at all** (a string fails its `oneOf`, an integer fails its `anyOf`), yet the merged form accepted strings, integers and booleans alike.

### Fix

The combination now fails conversion the way other inexpressible schemas do, with an actionable message. Callers that can degrade take their existing fallback: an MCP tool with such a schema is served non-strict with both constraints intact:

```
MCP: strict = False | field served as:
{"anyOf": [string, integer], "oneOf": [string, boolean]}
```

There is no meaning-preserving strict form to emit instead: computing the actual conjunction of two schema lists is not generally possible, and any flattening picks one constraint over the other.

Unchanged behaviour, pinned by tests that pass both before and after:

- `oneOf` alone still converts to `anyOf`, the documented discriminated-union path
- `oneOf` next to an **empty** `anyOf` list still converts, since an empty list constrains nothing
- discriminator preservation and nested/deep `oneOf` conversions are untouched

### Test plan

`test_anyof_and_oneof_on_the_same_node_are_rejected_not_merged` fails on `main` and passes here. `test_oneof_merged_with_existing_anyof` asserted the widened merge and is replaced by `test_oneof_alongside_anyof_is_rejected_rather_than_merged`, with the unsatisfiable-input observation recorded in the test.

| Command | Result |
| --- | --- |
| `make mypy` | 5 errors, all pre-existing on `main` |
| `uv run pytest tests/test_strict_schema.py tests/test_strict_schema_oneof.py` | 41 passed |
| `make tests` | 6407 passed, no new failures against `main` at the same commit |

This is independent of #4277: it changes only the `oneOf` block, which that PR does not modify, and the defect exists on `main` today. If both land, whichever goes second rebases trivially and I will handle that.

### Issue number

None reported; found while auditing the strict conversion for other silent semantic changes.

### Checks

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -163,9 +163,20 @@ def _ensure_strict_json_schema(
     one_of = json_schema.get("oneOf")
     if is_list(one_of):
         existing_any_of = json_schema.get("anyOf", [])
-        if not is_list(existing_any_of):
-            existing_any_of = []
-        json_schema["anyOf"] = existing_any_of + [
+        if is_list(existing_any_of) and existing_any_of:
+            # `anyOf` and `oneOf` on the same node are two constraints that must BOTH hold.
+            # Folding them into a single `anyOf` turns the conjunction into a union, so values
+            # that either constraint rejects become accepted: with anyOf [string, integer] and
+            # oneOf [string, boolean], the merge accepts 5 and true, both of which the
+            # original schema rejects. That cannot be expressed in strict mode, so fail the
+            # way other inexpressible schemas do, letting callers that can degrade (such as
+            # MCP tool conversion) fall back to serving the schema as non-strict.
+            raise UserError(
+                "Strict JSON schemas cannot express a schema that constrains the same value "
+                "with both anyOf and oneOf. Combine the two constraints into a single union, "
+                "or update the function or output tool to not use a strict schema."
+            )
+        json_schema["anyOf"] = [
             _ensure_strict_json_schema(
                 variant, path=(*path, "oneOf", str(i)), root=root, budget=budget
             )
--- a/tests/test_strict_schema.py
+++ b/tests/test_strict_schema.py
@@ -362,3 +362,47 @@ def test_ref_expansion_bomb_is_rejected():
     }
     with pytest.raises(UserError):
         ensure_strict_json_schema(schema)
+
+
+def test_anyof_and_oneof_on_the_same_node_are_rejected_not_merged():
+    # These are two constraints that must BOTH hold. Folding them into a single anyOf turns
+    # the conjunction into a union: with anyOf [string, integer] and oneOf [string, boolean],
+    # the merged schema accepts 5 and true, both of which the original rejects.
+    with pytest.raises(UserError, match="both anyOf and oneOf"):
+        ensure_strict_json_schema(
+            {
+                "type": "object",
+                "properties": {
+                    "f": {
+                        "anyOf": [{"type": "string"}, {"type": "integer"}],
+                        "oneOf": [{"type": "string"}, {"type": "boolean"}],
+                    }
+                },
+            }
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for identifying the semantic widening here. The core observation is correct, but we do not want to merge another isolated condition into the current `oneOf` normalization.

The proposed exception for `anyOf: []` is also incorrect: JSON Schema requires `anyOf` to contain at least one schema, so this shape must not be treated as imposing no constraint. More broadly, converting `oneOf` to `anyOf` is only meaning-preserving when the branches are mutually exclusive; this PR still widens overlapping `oneOf` schemas.

We are going to address this at the strict-conversion contract level: accept only known meaning-preserving normalizations, reject ambiguous or unsupported compositions before mutation, and validate the converted result against the supported Structured Outputs subset while preserving MCP's non-strict fallback. I am going to close this PR as a standalone fix, but we will reuse its mixed-union regression case in that work. Thank you for the careful investigation.

abhay-codes07:
Offering one more finding into the contract-level work you described, rather than opening it as another isolated PR.

`resolve_ref` splits the pointer on `/` without decoding RFC 6901 escapes, so `~1` and `~0` are never translated back to `/` and `~`:

```python
path = ref[2:].split("/")
for key in path:
    value = resolved[key]
```

A definition whose name contains either character is unreachable, even though the reference is spelled correctly per the spec:

| `$defs` key | reference | result on `main` |
| --- | --- | --- |
| `a/b` | `#/$defs/a~1b` | `KeyError: 'a~1b'` |
| `a~b` | `#/$defs/a~0b` | `KeyError: 'a~0b'` |

The two caller paths degrade differently, and neither is great:

- **MCP tool conversion**: the `KeyError` is swallowed by the broad `except Exception`, so the tool is silently served non-strict. Safe, but the cause is invisible.
- **`FunctionTool(strict_json_schema=True)` and output types**: a raw `KeyError: 'a~1b'` escapes to the caller. It is not a `UserError`, so it reads as an SDK crash rather than a schema problem, and it does not describe what is wrong.

Reachability is the honest caveat: Pydantic names definitions after classes, and OpenAPI 3 component names disallow both characters, so this needs a hand-written or programmatically generated schema. That is exactly the MCP-server case, but I have no report of it in the wild, which is why I am not proposing it as a standalone fix.

It seems worth folding into the rework for two reasons. Decoding the escapes is a small, spec-defined change local to `resolve_ref`. And "reject ambiguous or unsupported compositions before mutation" implies a well-defined resolution step underneath it, so a pointer form that cannot resolve at all is arguably part of that contract, and raising `UserError` for a genuinely unresolvable `$ref` would give both paths the same predictable behaviour instead of a `KeyError` on one and silence on the other.

Happy to send a patch against whatever shape that work takes, or to leave it entirely to you if it is already covered.

PR #4304, opened 2026-08-08
fix(sandbox): preserve falsey instrumentation instances

adityasingh2400:
### Summary

The sandbox sessions defaulted their instrumentation with `instrumentation or Instrumentation()`, so a caller-supplied instance that is falsey was silently discarded and replaced with a fresh default. `Instrumentation` is a public constructor parameter on every sandbox session and is exported from `agents.sandbox.session`, so a subclass that defines `__bool__` is a supported input.

This applies the same treatment #4286 and #4299 used for the realtime model and the compaction decision hooks, extending it to the ten remaining sites.

### Test plan

New test in `tests/sandbox/test_session_sinks.py` constructing a `SandboxSession` with a falsey `Instrumentation` subclass and asserting the instance is preserved. It fails on `main` because a different object is substituted. `make lint` is clean. `make typecheck` was not run to completion locally, though the change preserves the declared type.

### DIFF
--- a/src/agents/extensions/sandbox/blaxel/sandbox.py
+++ b/src/agents/extensions/sandbox/blaxel/sandbox.py
@@ -1052,7 +1052,9 @@ class BlaxelSandboxClient(BaseSandboxClient["BlaxelSandboxClientOptions"]):
     ) -> None:
         # Validate that the Blaxel SDK is importable.
         _import_blaxel_sdk()
-        self._instrumentation = instrumentation or Instrumentation()
+        self._instrumentation = (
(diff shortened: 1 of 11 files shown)


seratch:
As I mentioned at https://github.com/openai/openai-agents-python/pull/4299#pullrequestreview-4888130231, we no longer accept the similar changes.

adityasingh2400:
Understood, and sorry for the noise. I had not seen your note on #4299 when this went up. Holding off on this class entirely until your detection lands.

PR #4308, opened 2026-08-08
fix(items): strip output-only created_by when replaying model output as input

hsusul:
### Summary

`ModelResponse.to_input_items()` replays each model output item back as a Responses API **input** item. Several output item types carry a server-assigned, output-only `created_by` field that is **not** part of the input-item schema:

- `apply_patch_call` and `shell_call` (supported `ApplyPatchTool` / `ShellTool` calls)
- `function_call_output`, `custom_tool_call_output`, `computer_call_output` (tool-call outputs)
- `tool_search_call` / `tool_search_output`

The replay sanitizer in `_output_item_to_input_item` only stripped `created_by` for the two `tool_search` types, so every other type leaked `created_by` straight back to the API on replay.

`created_by` is not accepted by any input-item param in the OpenAI SDK schema, and the SDK already strips it at resolution time for shell/apply_patch calls (`turn_resolution.py`) and for tool_search (`_tool_search_item_to_input_item`). The direct `ModelResponse.to_input_items()` replay path bypassed that resolution-time stripping, leaving the field in place.

**Fix:** strip `created_by` for every replayed item in `_output_item_to_input_item`, generalizing the existing tool_search behavior. `created_by` is never read by any SDK logic (it is only ever popped), so dropping it on replay is safe.

Before:

```python
resp = ModelResponse(output=[apply_patch_call_with_created_by], usage=Usage(), response_id=None)
resp.to_input_items()
# [{'id': 'apc_1', 'call_id': 'call_1', 'operation': {...}, 'status': 'completed',
#   'type': 'apply_patch_call', 'created_by': 'program_1'}]   # <-- leaked
```

After:

```python
# [{'id': 'apc_1', 'call_id': 'call_1', 'operation': {...}, 'status': 'completed',
#   'type': 'apply_patch_call'}]   # created_by stripped
```

The behavior is otherwise unchanged: items are still converted via the same `dict(...)` / `model_dump(exclude_unset=True)` logic, and the `AgentsException` fallback for unexpected types is preserved. The only caller affected is `ModelResponse.to_input_items()`; the two other callers (`_fingerprint_for_tracker`, `_anonymous_tool_search_fingerprint` in `oai_conversation.py`) only pass `tool_search` items, which already routed through the created_by-stripping branch.

### Test plan

- Added `test_to_input_items_strips_created_by_for_non_tool_search_items` in `tests/test_items_helpers.py`, which builds an `apply_patch_call` and a `function_call_output` output item with `created_by` set and asserts the field is dropped on replay. It fails on `main` (leaks `created_by`) and passes with this change.
- The existing `test_to_input_items_for_tool_search_strips_created_by` and all other `to_input_items` round-trip tests still pass.
- Commands run locally:
  - `uv run pytest tests/test_items_helpers.py` → 36 passed
  - `uv run pytest tests/test_items_helpers.py tests/memory tests/test_agent_runner.py tests/test_run_state.py tests/test_server_conversation_tracker.py tests/test_programmatic_tool_calling.py` → 903 passed
  - `uv run ruff format --check` and `uv run ruff check` → clean
  - `uv run python .github/scripts/check_optional_truthiness.py src/agents` → clean
  - `uv run mypy src/agents/items.py` → Success
  - `uv run pyright src/agents/items.py tests/test_items_helpers.py` → 0 errors

Not run: the full test matrix and integration/coverage profiles (they require provider credentials / live services, which are unrelated to this change).

### Issue number

N/A — no existing issue.

### Checks

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -230,6 +230,20 @@ def _output_item_to_input_item(raw_item: Any) -> TResponseInputItem:
     # carry it (apply_patch/shell calls and tool-call outputs); the tool_search branch above
     # already drops it, so do the same for every other item type.
     payload.pop("created_by", None)
+    if item_type == "shell_call_output":
+        # ``shell_call_output.output`` is a list of content chunks that each carry their own
+        # output-only ``created_by``. ``payload`` was only shallow-copied above, so rebuild the
+        # list with fresh chunk copies to strip the nested field without mutating the caller's
+        # original mapping. Mirrors the two-level stripping the runner already does in
+        # ``turn_resolution``.
+        chunks = payload.get("output")
+        if isinstance(chunks, list):
+            payload["output"] = [
+                {key: value for key, value in chunk.items() if key != "created_by"}
+                if isinstance(chunk, dict)
+                else chunk
+                for chunk in chunks
+            ]
     return cast(TResponseInputItem, payload)
 
 
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -211,7 +211,7 @@ def _tool_search_item_to_input_item(
 
 
 def _output_item_to_input_item(raw_item: Any) -> TResponseInputItem:
-    """Convert an output item into replayable input, normalizing tool_search items."""
+    """Convert an output item into replayable input, stripping output-only metadata."""
     item_type = (
         raw_item.get("type") if isinstance(raw_item, dict) else getattr(raw_item, "type", None)
     )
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
One supported Responses shape is still left unsanitized: each entry in `shell_call_output.output` can carry its own output-only `created_by`. The current helper removes only the top-level field, so `ModelResponse.to_input_items()` still sends `input[i].output[j].created_by`. The normal runner path already removes both levels in `turn_resolution.py`.

Please sanitize the nested shell output chunks as well without mutating the original mapping input, and add a regression test covering both top-level and nested `created_by`. Once the replayed request contains neither field, this should be ready for another review.

seratch (approved):

seratch:
@hsusul Can you resolve the typecheck error?

```
tests/test_items_helpers.py:653: error: List item 0 has incompatible type "dict[str, Any]"; expected "ResponseOutputMessage | ResponseFileSearchToolCall | ResponseFunctionToolCall | ResponseFunctionToolCallOutputItem | ResponseFunctionWebSearch | <23 more items>"  [list-item]
Found 1 error in 1 file (checked 854 source files)
```

PR #4310, opened 2026-08-08
fix: preserve free-form MCP object schemas

seratch:
This pull request supersedes #4277 and fixes automatic MCP strict-schema conversion for direct free-form object inputs.

MCP servers can describe arbitrary key/value maps as `{"type": "object"}` plus annotations. Strict conversion previously added `additionalProperties: false`, silently narrowing those maps to empty-object-only schemas while marking the tools strict. The MCP conversion boundary now detects direct annotation-only object roots, properties, and array items before conversion and preserves the copied original schema through the existing non-strict fallback.

The implementation intentionally leaves `$ref`, composition, conditional, dependency, and other advanced JSON Schema evaluation to the existing strict converter. Explicitly empty or closed objects, declared-property schemas, and strictable composed objects continue to convert as before.

This prevents the SDK-side schema corruption present in the nested-map scenario described in #1681; it does not add Realtime strict-mode support.

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -536,6 +536,7 @@ class MCPUtil:
             failure_error_function
         )
         schema, is_strict = copy.deepcopy(tool_input_schema(tool)), False
+        input_schema_is_empty = schema == {}
 
         # MCP spec doesn't require the inputSchema to have `properties`, but OpenAI spec does.
         if "properties" not in schema:
@@ -548,7 +549,10 @@ class MCPUtil:
             # non-strict. Convert a separate copy so the non-strict fallback keeps
             # the original schema intact.
             try:
-                schema = ensure_strict_json_schema(copy.deepcopy(schema))
+                schema = ensure_strict_json_schema(
+                    copy.deepcopy(schema),
+                    _reject_open_objects=not input_schema_is_empty,
+                )
                 is_strict = True
             except Exception as e:
                 if _debug.DONT_LOG_TOOL_DATA:
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -27,12 +27,22 @@ _ADDITIONAL_PROPERTIES_ERROR = (
     "to not use a strict schema."
 )
 
+_OPEN_OBJECT_ERROR = (
+    "JSON schema contains an object that permits undeclared properties and cannot be converted "
+    "to a strict schema without changing its accepted values."
+)
+
+_UNVALIDATED_REF_ERROR = (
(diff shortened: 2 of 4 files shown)


abhay-codes07:
Thanks for landing this, and for the credit. I re-ran my case library against the merged conversion and it handles every one of them correctly, including the ones my PR needed nine rounds to get right: free-form nodes inside referenced definitions, alias chains, `allOf` branches carrying a bare `$ref`, nested `$defs`, and the `maxProperties` / `dependentRequired` / undeclared-`required` shapes. Nothing narrows. The opt-in flag is a cleaner boundary than what I proposed.

One question, and I want to be explicit that I am not sure it is a defect.

`test_to_function_tool_free_form_object_arg_falls_back_to_non_strict` parametrizes `{"type": "object", "properties": {}}` as free-form, and at a nested argument position I think that is clearly right: a server sending an empty map for `keysAndValues` almost certainly means "any object", not "the empty object".

At the **root** position the same shape has a different conventional meaning: it is how a no-argument tool is normally declared. That case is not covered by the parametrization, and it now falls back:

| MCP `inputSchema` | strict |
| --- | --- |
| `{"type": "object", "properties": {"a": {"type": "string"}}}` | `True` |
| `{"type": "object", "properties": {}}` | `False` (was `True` before this PR) |

So every no-argument MCP tool loses strict mode. Nothing is served incorrectly, and the fallback is the safe direction, but it is a fairly common shape to downgrade if the root reading was not the intended one.

I did prototype a change and then threw it away: distinguishing the root from nested positions is easy, but it contradicted 19 of your new tests, which told me the uniform treatment is deliberate rather than an oversight. So I would rather ask than send a PR that argues with your test suite. If the root case is intended, please ignore this. If it is not, I am happy to send a narrow patch that only affects the root, with the nested parametrization left exactly as it is.

seratch:
Thanks for checking this so carefully. The root fallback is intentional. In JSON Schema, `{"type":"object","properties":{}}` does not mean "no arguments"; without `additionalProperties: false`, it still accepts arbitrary keys, including at the root. Special-casing the root would silently narrow the accepted input and reintroduce the behavior #4310 fixes. The merged `test_to_function_tool_free_form_root_falls_back_to_non_strict` test covers this exact shape.

You are right that common zero-parameter generators can emit that open shape, so those tools may fall back to non-strict. That is preferable to the client guessing a stronger schema than the server advertised. A genuinely no-argument tool can remain strict by publishing an explicitly closed schema such as `{"type":"object","properties":{},"additionalProperties":false}`. No root-only follow-up patch is needed.

Issue #4283, opened 2026-08-07
Allow applications to explicitly approve replay-unsafe model retries


Filimoa:
### Please read this first

- [x] I read the Agents SDK runner-managed retry and Responses WebSocket documentation.
- [x] I searched open and closed issues for replay safety, response-started disconnects, and retry-policy overrides.
- [x] I reproduced this with `openai-agents 0.19.4` and confirmed the relevant behavior remains on `main`.

### Describe the feature

The runner-managed retry API currently has an ownership gap: applications can provide a `ModelRetrySettings.policy`, but that policy is not invoked when provider advice marks replay as unsafe.

This occurs when a non-streamed `Runner.run()` request over the Responses WebSocket transport receives one or more response events and then loses the connection before receiving a terminal event.

The observed exception is:

```text
websockets.exceptions.ConnectionClosedError: no close frame received or sent
```

The WebSocket adapter correctly:

1. Detects that response processing started.
2. Drops the failed connection.
3. Marks the failure with `replay_safety="unsafe"`.

However, the runner then returns `RetryDecision(retry=False)` before invoking the configured application retry policy:

https://github.com/openai/openai-agents-python/blob/main/src/agents/run_internal/model_retry.py#L276-L315

This is consistent with the documented safety boundary that requests are not replayed after a response event arrives:

https://openai.github.io/openai-agents-python/models/#runner-managed-retries

I agree with that behavior as the default. The SDK cannot determine whether replaying an interrupted request is acceptable for a particular application:

- For non-streamed `Runner.run()`, partial model output has not been exposed to the application.
- Application-local function tools from the incomplete model response have not been executed.
- Previously completed Runner turns and local tool outputs can be preserved by retrying only the current model request.
- However, hosted tools or other provider-side work may already have occurred and could be duplicated.

Therefore, this should not become an automatic retry. The application needs a supported way to explicitly accept that risk for a narrowly defined workload, such as a read-only research or evaluation run.

### Current limitation

`RetryDecision` already contains an internal `_approves_replay` field:

https://github.com/openai/openai-agents-python/blob/main/src/agents/retry.py#L115-L123

Internal provider policies can set that approval when the provider considers replay safe, but an application retry policy cannot express the equivalent application-owned decision.

The only current workaround is to wrap or subclass `OpenAIResponsesWSModel` and override `get_retry_advice()` so that an application decision is presented as provider advice:

```python
ModelRetryAdvice(
    suggested=True,
    replay_safety="safe",
)
[...]

PR #4313, opened 2026-08-09
feat(retry): let applications approve replay-unsafe model retries

LeSingh1:
## Summary

Implements the scope you set in #4283: an application retry policy can now explicitly approve a replay that the provider marked unsafe, for non-streamed requests only.

- `RetryDecision` gains a public `approve_unsafe_replay` field, appended. It is deliberately separate from `retry` — an ordinary `RetryDecision(retry=True)` still cannot bypass replay protection. The provider-owned `_approves_replay` flag is unchanged.
- The policy now runs for a provider-unsafe **non-streamed** failure, and its decision is honored only when it explicitly approves.
- Absolute vetoes are preserved: aborts, failures that already emitted user-visible streamed output, and any provider-unsafe **streamed** failure (vetoed before the policy is reached).
- `RetryPolicyContext` gains `response_started`, `previous_response_id` and `conversation_id`, plus `replay_safety` and `stateful_request` accessors, so a policy can scope its approval rather than blanket-approve. `response_started` comes from the signal the Responses WebSocket adapter already records on the error.
- Programmatic Tool Calling is intentionally not covered.

All new public fields are appended, so positional order is unchanged.

```python
def retry_policy(context: RetryPolicyContext) -> RetryDecision:
    if context.attempt == 1 and not context.stream and context.response_started:
        return RetryDecision(retry=True, approve_unsafe_replay=True, reason="read-only turn")
    return RetryDecision(retry=False)
```

## Test plan

Three regression tests in `tests/models/test_model_retry.py`, driven through advice matching the WebSocket adapter's response-started disconnect:

- approval accepted: the request is replayed and the policy receives the structured context.
- approval absent: a plain `retry=True` is still refused.
- streamed request: an approving decision is ignored.

`make format`, `make lint`, `make typecheck` and `make tests` pass.

Refs #4283

### DIFF
--- a/docs/models/index.md
+++ b/docs/models/index.md
@@ -590,7 +590,9 @@ def retry_policy(context: RetryPolicyContext) -> RetryDecision:
 
 An ordinary `RetryDecision(retry=True)` never lifts replay protection; the approval has to be explicit. `RetryPolicyContext` exposes `response_started`, `replay_safety`, `stateful_request`, `previous_response_id`, and `conversation_id` so the policy can scope the approval to the failure it actually intends to accept. Replaying a request whose provider-side work may already have run can repeat that work, so approve it only for workloads where that is acceptable.
 
-Stateful follow-up requests using `previous_response_id` or `conversation_id` are also treated more conservatively. For those requests, non-provider predicates such as `network_error()` or `http_status([500])` are not enough by themselves. The retry policy should include a replay-safe approval from the provider, typically via `retry_policies.provider_suggested()`.
+Stateful follow-up requests using `previous_response_id` or `conversation_id` — including those produced by `auto_previous_response_id=True` — are also treated more conservatively, because the follow-up depends on server-side state. For those requests, non-provider predicates such as `network_error()` or `http_status([500])` are not enough by themselves. The retry policy should include a replay-safe approval from the provider, typically via `retry_policies.provider_suggested()`.
+
+`approve_unsafe_replay` also applies to a stateful request, but only when provider advice marked replay as unsafe on a non-streamed failure. When replay safety is unknown, a stateful request stays blocked no matter what the policy returns: there is no provider-marked unsafe failure for the approval to be about.
 
 ##### Runner and agent merge behavior
 
--- a/docs/models/index.md
+++ b/docs/models/index.md
@@ -498,7 +498,7 @@ english_agent = Agent(
 
 Retries are runtime-only and opt in. The SDK does not retry general model requests unless you set `ModelSettings(retry=...)` and your retry policy chooses to retry.
 
-On the Responses websocket transport, `retry_policies.provider_suggested()` recognizes pre-response overload frames and code-less `server_error` frames as retry suggestions. This does not enable retries by itself: you still need `ModelRetrySettings`, and the normal replay-safety checks still apply. If any response event has already arrived, the SDK does not replay the request.
+On the Responses websocket transport, `retry_policies.provider_suggested()` recognizes pre-response overload frames and code-less `server_error` frames as retry suggestions. This does not enable retries by itself: you still need `ModelRetrySettings`, and the normal replay-safety checks still apply. Once a response event has arrived, the SDK does not replay the request on its own; a non-streamed request can still be replayed if the retry policy explicitly approves it, as described under [Safety boundaries](#safety-boundaries).
 
 ```python
 from agents import Agent, ModelRetrySettings, ModelSettings, retry_policies
--- a/docs/models/index.md
+++ b/docs/models/index.md
@@ -573,6 +573,23 @@ Some failures are never retried automatically:
 - Requests where provider advice marks replay as unsafe.
 - Streamed runs after output has already started in a way that would make replay unsafe.
 
+Abort errors and streamed runs that already produced output are absolute: no retry policy can override them. Provider-marked replay unsafety on a non-streamed request is the one case an application can accept deliberately, by returning `approve_unsafe_replay=True` alongside `retry=True`:
+
+```python
+from agents.retry import RetryDecision, RetryPolicyContext
+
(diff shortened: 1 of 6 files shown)


seratch (changes requested):
Please keep application approval scoped to provider-marked unsafe non-streamed failures. It must not satisfy generic `replay_unsafe_request`, so Programmatic Tool Calling remains vetoed unless the existing provider-owned safety approval applies. Also preserve `approve_unsafe_replay` through `retry_policies.all()` and every positive-decision merge in `retry_policies.any()`.

Please add base-failing regressions for both cases, assert that the WebSocket adapter populates `response_started`, and update the runner-managed retry documentation, whose current absolute-veto wording would become inaccurate. After those focused changes, this should be ready for another review.

LeSingh1:
Thanks — both were real. Pushed in 51d8efbd.

- **PTC scope.** The two vetoes are now separate checks. `replay_unsafe_request` keeps its original condition, lifted only by the provider-owned `_approves_replay`; `approve_unsafe_replay` applies solely to `provider_advice.replay_safety == "unsafe"` on a non-streamed failure. Folding them together was the bug.
- **Combinators.** `retry_policies.all()` and the positive-decision merge in `any()` now carry `approve_unsafe_replay` through.
- **Regressions**, all failing on base: a PTC request staying vetoed under application approval, and `all()`/`any()` preserving it.
- **WebSocket adapter**: asserts `response_started` is populated for a response-started disconnect, and stays `False` for a stateful-request disconnect.
- **Docs**: the safety-boundaries section now separates the absolute vetoes from the one case an application can accept, with an example.

seratch (changes requested):
Thanks for addressing the previous feedback. The PTC veto and policy-combinator propagation now look correct.

I found one remaining issue: `get_response_with_retry()` folds `stateful_request` into `replay_unsafe_request`, so `approve_unsafe_replay=True` is rejected whenever `previous_response_id` or `conversation_id` is set. This misses the `auto_previous_response_id=True` environment reported in #4283.

The PTC/local-side-effect veto, stateful fail-closed default, and provider-marked unsafe approval need to remain distinct. The docs sentence saying that a response-started request is never replayed also contradicts the new approval section.

As this has been a lower-priority task from the outset and I need to focus on other priorities, my responses on further iterations may be slow

LeSingh1:
Right on both counts — pushed in 015c28b4.

`stateful_request` and `replay_unsafe_request` were passed to `_evaluate_retry` as a single OR-ed flag, so a stateful request was evaluated under the PTC veto and the approval could never lift it. That is the `auto_previous_response_id=True` case, so the option did not reach the environment the issue reports.

They are now separate parameters and three distinct checks:

- request-level replay veto (PTC): lifted only by the provider-owned approval;
- stateful fail-closed: no application-local side effects, so lifted by an explicit application approval or either provider signal;
- provider-marked unsafe: lifted by either explicit approval.

Regressions added, both failing on base: a stateful request replayed under application approval, and a stateful request still failing closed under an ordinary `retry=True`. Also fixed the websocket paragraph that still said a response-started request is never replayed.

Understood on timing — no rush.

seratch (changes requested):
Thanks for the update. The reported `auto_previous_response_id=True` path now works in my focused probe, and the PTC veto and policy-combinator behavior remain correct.

One scope mismatch remains: `approve_unsafe_replay=True` also lifts the stateful fail-closed gate when provider replay safety is unknown. That is broader than the accepted #4283 scope and the public contract, which limit this approval to provider-marked unsafe, non-streamed failures. Please keep stateful requests with unknown replay safety blocked, add a regression for that boundary, and align the stateful documentation with the final behavior. Please also run the required translation review after updating the docs.

Also, please resolve all the feedback from Codex as well.

LeSingh1:
Fixed in d00686bd.

`approve_unsafe_replay` no longer lifts the stateful gate when replay safety is unknown — it now applies to a stateful request only when provider advice marked replay unsafe on a non-streamed failure. Regression added for that boundary (fails without the change), and the stateful paragraph in the docs now states it explicitly.

That also resolves the third Codex comment, which flagged the same mismatch between the docs and the stateful behaviour. The other two Codex comments (PTC replay, combinator propagation) were addressed in 51d8efbd.

**Translation review: not done.** `docs/scripts/translate_docs.py --mode full --file docs/models/index.md` needs `OPENAI_API_KEY`, which I do not have, so per AGENTS.md I am reporting the validation as incomplete rather than claiming the docs change is ready. `docs/ja`, `docs/ko` and `docs/zh` are untouched. If you would like, I can split the prose into a follow-up PR so the runtime change is not held up by a check I cannot run.

`make format`, `make lint`, `make typecheck`, `make tests` all pass (7252).

docs/models/index.md
```diff
@@ -573,7 +573,26 @@ Some failures are never retried automatically:
 - Requests where provider advice marks replay as unsafe.
 - Streamed runs after output has already started in a way that would make replay unsafe.
 
-Stateful follow-up requests using `previous_response_id` or `conversation_id` are also treated more conservatively. For those requests, non-provider predicates such as `network_error()` or `http_status([500])` are not enough by themselves. The retry policy should include a replay-safe approval from the provider, typically via `retry_policies.provider_suggested()`.
+Abort errors and streamed runs that already produced output are absolute: no retry policy can override them. Provider-marked replay unsafety on a non-streamed request is the one case an application can accept deliberately, by returning `approve_unsafe_replay=True` alongside `retry=True`:
+
+```python
```
seratch, line 578: Can you revert all the document changes? We need to hold off merging those until shipping the feature. That's why we've been having a separate document PR per release.

seratch (changes requested):

seratch:
Closing this PR in favor of https://github.com/openai/openai-agents-python/pull/4319; thanks again for working on it

PR #4314, opened 2026-08-09
fix(sandbox): reject non-object rollout JSONL records

fallintoplace:
## Summary

`render_phase_one_prompt()` now validates every decoded JSONL record is an object before reading terminal metadata. Non-object values raise a record-numbered `ValueError` instead of an accidental `AttributeError`.

## Tests

- `uv run pytest tests/sandbox/test_memory.py`
- `uv run ruff format --check src/agents/sandbox/memory/phase_one.py tests/sandbox/test_memory.py`
- `uv run ruff check src/agents/sandbox/memory/phase_one.py tests/sandbox/test_memory.py`
- `uv run mypy src/agents/sandbox/memory/phase_one.py`
- `uv run pyright src/agents/sandbox/memory/phase_one.py`

### DIFF
--- a/src/agents/sandbox/memory/phase_one.py
+++ b/src/agents/sandbox/memory/phase_one.py
@@ -46,6 +46,9 @@ def render_phase_one_prompt(*, rollout_contents: str) -> str:
     payloads = [json.loads(line) for line in rollout_contents.splitlines() if line.strip()]
     if not payloads:
         raise ValueError("rollout_contents must contain at least one JSONL record")
+    for record_number, payload in enumerate(payloads, start=1):
+        if not isinstance(payload, dict):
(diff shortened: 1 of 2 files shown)


seratch:
I traced this through the sandbox memory generation path: supported writes are coerced to dictionaries and serialized as JSON objects before Phase 1 reads them. For a manually corrupted non-object record, the worker already catches the existing failure and skips that rollout, so this change does not demonstrate a user-visible recovery or state-integrity benefit. It also makes an earlier scalar record invalidate an otherwise processable rollout, which is stricter than v0.19.4 behavior.

I am going to close this PR for now. If there is a supported real-world path that produces these records and a concrete downstream consequence beyond the logged exception, please share that reproduction so we can choose the narrowest validation boundary.

PR #4317, opened 2026-08-09
fix(tracing): release the span scope when finish fails on generator close

LeSingh1:
`SpanImpl.finish` calls `on_span_end` before resetting the scope, so a processor that raises leaves the reset unreached. The finished span stays current for the rest of the caller's scope and later spans nest under it — the same leak the `GeneratorExit` handling exists to prevent.

#4232 fixed this for `TraceImpl` by running the reset in a `finally` and clearing the token only after the reset has been attempted. #4233 landed first, so `spans.py` never got that refinement. This applies it, leaving the two helpers identical.

New test in `tests/tracing/test_spans_impl.py` fails on `main` (`Scope.get_current_span()` is still the finished span) and passes here. Existing tests unchanged.

Refs #4232

### DIFF
--- a/src/agents/tracing/spans.py
+++ b/src/agents/tracing/spans.py
@@ -43,17 +43,22 @@ def _finish_on_generator_exit(span: Span[Any]) -> None:
     An explicit ``finish`` from the wrong context is a context-ownership violation rather
     than an unavoidable one, so it still raises, and a processor that fails during
     ``finish`` still surfaces rather than being mistaken for a foreign token.
-    """
-    span.finish(reset_current=False)
 
-    token: contextvars.Token[Span[Any] | None] | None = span._prev_span_token  # type: ignore[attr-defined]
-    if token is None:
-        return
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for taking the time to investigate this. I reproduced the scope leak when `SpanImpl` is constructed directly with a raw processor that raises from `on_span_end`, and the patch does fix that internal path.

However, the supported processor APIs route callbacks through `SynchronousMultiTracingProcessor`, which catches processor exceptions. In that path, generator close already releases the span scope on current `main`. `SpanImpl` is not part of the public tracing exports, and the `TracingProcessor` contract requires `on_span_end` not to raise, so the new test demonstrates an internal, contract-violating construction rather than a supported user scenario.

I am going to close this PR for now. If there is a supported public path that bypasses the processor wrapper and can leave the span current, please provide a minimal reproduction using that path and we can revisit the underlying problem.

Issue #4270, opened 2026-08-07
Add opt-in for preserving the raw usage data in model response


zhuziqi97:
### Please read this first

- **Have you read the docs?** Yes.
- **Have you searched for related issues?** Yes. I found #2034 and #2192, which intentionally normalize missing token-detail values to zero to prevent crashes, but I did not find an issue discussing preservation of provider-reported field availability.

### Question

Should the Agents SDK preserve whether each usage field was explicitly reported by the model provider, independently from the normalized numeric value?

I am not proposing that the existing aggregate numeric fields necessarily become nullable. Normalizing missing values to zero is useful for arithmetic and backwards compatibility.

The problem is that downstream observability currently cannot distinguish:

1. The provider explicitly reported `cached_tokens: 0`.
2. The provider omitted `prompt_tokens_details` or `cached_tokens`.
3. A model adapter did not preserve the provider field.

All three cases become `cached_tokens == 0` after `Usage` normalization and serialize identically.

This matters for usage reporting and coverage metrics. A zero reported by the provider is an exact measurement, while a missing field means the metric is unavailable. Treating unavailable as an exact zero can produce misleading cache-hit reports.

### Debug information

- Agents SDK version: `0.19.4`
- OpenAI Python version: `2.53.0`
- Pydantic version: `2.13.4`
- Python version: `3.12.13`
- Operating system: Linux
- Does the issue reproduce with the latest Agents SDK release? Yes
- Does the issue occur consistently? Yes
- Model/provider call required to reproduce? No

The behavior is visible in the current helpers and `Usage.__post_init__`:

- `_make_input_tokens_details()` converts `None` to `0`.
- `_cached_tokens()` defaults missing values to `0`.
- `_normalize_input_tokens_details()` converts missing details to zero-valued details.
- `Usage.__post_init__()` defensively normalizes nested `None` values.

Related source:

https://github.com/openai/openai-agents-python/blob/v0.19.4/src/agents/usage.py

### Minimal reproduction

```python
from agents.usage import Usage, serialize_usage
from openai.types.completion_usage import PromptTokensDetails

common = {
    "requests": 1,
    "input_tokens": 10,
    "output_tokens": 2,
    "total_tokens": 12,
}

provider_field_missing = Usage(
    **common,
    input_tokens_details=None,
)

provider_explicit_zero = Usage(
    **common,
    input_tokens_details=PromptTokensDetails(cached_tokens=0),
)

[...]

PR #4318, opened 2026-08-09
docs(usage): document the raw provider usage snapshot opt-in

LeSingh1:
`ModelSettings.preserve_raw_usage` and `ModelResponse.raw_usage` shipped in #4279 with no prose docs. The feature exists specifically to distinguish a usage field the provider omitted from one it reported as zero — a distinction `Usage` cannot express — so its target audience needs both an entry point and the boundaries of the snapshot.

Adds a subsection to `docs/usage.md` under third-party adapter usage, plus two API reference links. Covers: how to enable and read it, that omitted keys stay absent while explicit zeros stay present, and the four boundaries (one snapshot per `ModelResponse` on streamed and non-streamed calls, never aggregated; does not request usage from the provider; `None` when no usage reaches the adapter; not persisted by `RunState`).

Verified the snippet and every claim against the current implementation on both the streamed and non-streamed Chat Completions paths. `make build-docs` is clean; only the English source is touched.

Refs #4270

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -38,6 +38,33 @@ Usage reporting varies across third-party adapters and provider backends. If you
 
 Review the adapter-specific notes in the [Third-party adapters](models/index.md#third-party-adapters) section of the Models guide and validate usage reporting on the exact provider backend you plan to deploy.
 
+### Preserving the raw provider usage payload
+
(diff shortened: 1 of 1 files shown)


LeSingh1:
Closing as a duplicate — #4280 already covers this, and I missed it before filing. Sorry for the noise.

PR #4341, opened 2026-08-09
fix(voice): preserve explicit falsey STT/TTS models in VoicePipeline

hsusul:
### Summary

`VoicePipeline` accepts concrete `STTModel`/`TTSModel` instances and stores them, but `_get_tts_model()` / `_get_stt_model()` decided whether a model was supplied with a truthiness check (`if not self.<model>`). A valid model subclass whose `__bool__()` returns `False` was therefore treated as *absent* and silently replaced by a provider-default lookup.

The constructor documents the default as used only "if not provided," and only `None` actually means "no concrete model was supplied." Replacing a present-but-falsey model changes observable model identity and can trigger an unexpected credential/network dependency (the provider lookup) that the caller never intended.

The fix checks `self.<model> is None` instead. String model-name resolution and the `None`-default path are unchanged; the constructor's accepted types are untouched. Both the single-turn (`AudioInput`) and streamed (`StreamedAudioInput`) paths share these getters and are covered.

This is the same optional-reference-vs-truthiness class that the repo has already treated as a bug elsewhere (e.g. #4286 for realtime models, and the #4305 sweep). The `check_optional_truthiness.py` lint added in #4305 does not flag these getters because `stt_model`/`tts_model` are typed `Model | str | None`; the extra `str` union member puts them outside the checker's single-payload optional-reference heuristic, so this instance was missed.

### Test plan

Added two regression tests in `tests/voice/test_pipeline.py` that inject concrete `STTModel`/`TTSModel` subclasses whose `__bool__` returns `False`, backed by a `VoiceModelProvider` whose lookups raise. They assert the explicit models are used and provider lookup is never reached, on both the single static-turn and streamed multi-turn paths. Both tests fail on `main` (provider lookup is invoked) and pass with this change.

Ran the repository verification script `.agents/skills/code-change-verification/scripts/run.sh`: `make format`, `make lint`, `make typecheck` (mypy + pyright), and `make tests` (full suite) all pass. `tests/voice` passes (121 tests), and `check_optional_truthiness.py src/agents` is clean.

### Issue number

N/A — found by inspection; no existing issue.

### Checks

### DIFF
--- a/src/agents/voice/pipeline.py
+++ b/src/agents/voice/pipeline.py
@@ -75,12 +75,12 @@ class VoicePipeline:
             raise UserError(f"Unsupported audio input type: {type(audio_input)}")
 
     def _get_tts_model(self) -> TTSModel:
-        if not self.tts_model:
+        if self.tts_model is None:
             self.tts_model = self.config.model_provider.get_tts_model(self._tts_model_name)
         return self.tts_model
 
     def _get_stt_model(self) -> STTModel:
-        if not self.stt_model:
+        if self.stt_model is None:
             self.stt_model = self.config.model_provider.get_stt_model(self._stt_model_name)
         return self.stt_model
 
--- a/tests/voice/test_pipeline.py
+++ b/tests/voice/test_pipeline.py
@@ -20,6 +20,7 @@ try:
         StreamedAudioResult,
         STTModelSettings,
         TTSModelSettings,
+        VoiceModelProvider,
         VoicePipeline,
         VoicePipelineConfig,
         VoiceStreamEvent,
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for the contribution. The truthiness check can technically replace a custom `STTModel` or `TTSModel` whose class defines `__bool__` or `__len__`, and the proposed `None` check is logically correct. However, the PR currently demonstrates this only with synthetic subclasses created specifically to be falsey. The built-in voice models are always truthy, and no real adapter or user scenario reaching this condition has been identified.

I am going to close this PR for now because the practical need is not demonstrated. If you can provide a real custom model implementation or user report where this causes an unintended provider lookup, we can revisit the focused fix.

PR #4354, opened 2026-08-10
fix(strict-schema): reject $ref constraining siblings and non-convertible shapes

### LINKED ISSUE #4353: strict_schema: $ref sibling merge is parent-wins (silently inverts accept set), and convert_schemas_to_strict falls back on 2 of 17 non-convertible shapes
Two findings in `agents/strict_schema.py` (measured against `openai-agents==0.19.4`, `openai==7.4.0` for the oracle). The second is opt-in and loud; **the first is on the default path and silent**, so I've put it first.

The surrounding design is right, and that's worth saying up front: the MCP path deep-copies before converting so a partial mutation can't leak, the node budget at `strict_schema.py:21` is a real hostile-input defence, and the `$ref` parent-wins merge is *correct* for the shape pydantic actually emits. Both findings are about the same underlying question — what counts as "not convertible" — rather than about the architecture.

---

## 1. `$ref` sibling merge is parent-wins, which silently deletes the referent's constraints (default path)

`strict_schema.py:221`:

```python
json_schema.update({**resolved, **json_schema})   # parent keys win
```

Draft 2020-12 applies the referent **and** the siblings — it's an intersection, not a precedence. When the siblings only carry annotations (`description`, `title`), parent-wins is indistinguishable from an intersection, which is why this holds for pydantic output. When a sibling *constrains*, "wins" means "deletes".

Repro, run verbatim on 0.19.4:

```python
import json
from agents import FunctionTool

schema = {
    "$defs": {"T": {"type": "object", "properties": {"b": {"type": "string"}}, "required": ["b"]}},
    "type": "object",
    "properties": {
        "node": {"properties": {"a": {"type": "string"}}, "required": ["a"], "$ref": "#/$defs/T"}
    },
}

async def _invoke(ctx, args): return ""

tool = FunctionTool(name="t", description="d", params_json_schema=schema, on_invoke_tool=_invoke)
print("strict_json_schema (default):", tool.strict_json_schema)
print(json.dumps(tool.params_json_schema["properties"]["node"], indent=2))
```

Output:

```
strict_json_schema (default): True
{
  "properties": {"a": {"type": "string"}},
  "required": ["a"],
  "type": "object",
  "additionalProperties": false
}
```

`T`'s `b` is gone, and because `additionalProperties: false` is then applied, **a property the schema required has become forbidden.** Accept sets via `ajv` (2020-12), over `{}`, `{node:{}}`, `{node:{a}}`, `{node:{b}}`, `{node:{a,b}}`:

| | `{}` | `{node:{}}` | `{node:{a}}` | `{node:{b}}` | `{node:{a,b}}` |
|---|---|---|---|---|---|
| input | ✅ | ❌ | ❌ | ❌ | ✅ |
| after conversion | ❌ | ❌ | ✅ | ❌ | ❌ |

The two sets are **disjoint**. The `{}` row is the documented strict-mode rule (every property becomes required) and is expected; the load-bearing change is that the only populated form the schema accepted is now rejected, and a form it rejected is now the only one accepted.

**Why nothing surfaces it:** the converted schema is *valid* strict-mode JSON Schema — `openai@7.4.0`'s own `toStrictJsonSchema()` accepts it — so there is no 400 and no local error. The raw input, by contrast, throws there. The conversion turns a loud rejection into a silently different tool.

**Reachability, stated at its true strength:** this is not a pydantic bug. Measured on pydantic 2, a described nested model emits `{"$ref": ..., "description": ...}` — annotations only — and the control below shows that case is preserved exactly. It fires on **hand-authored / OpenAPI-composition / MCP-supplied** schemas, which reach `ensure_strict_json_schema` through `FunctionTool.__post_init__` (`tool.py:597`, with `strict_json_schema: bool = True` at `tool.py:469`), `agent_output.py:114`, and the MCP path.

**Test-design warning** (this is the part I'd flag hardest): a regression test built on a pydantic model **passes either way.** Same run, control case:

```python
ann = {
  "$defs": {"T": {"type": "object", "properties": {"b": {"type": "string"}}, "required": ["b"]}},
  "type": "object",
  "properties": {"node": {"description": "a node", "$ref": "#/$defs/T"}},
}
ensure_strict_json_schema(ann)["properties"]["node"]
# -> {"description": "a node", "type": "object", "properties": {"b": ...}, "required": ["b"], "additionalProperties": false}
```

Fully preserved. The fixture has to have a **constraining** sibling (`properties`/`required`/`type`/`enum`), not a `description`.

Two fix shapes, either defensible:

- **Merge as an intersection** — union `properties` and `required` from both sides. Safe when neither side is closed; if a side declares `additionalProperties: false`, the union has to be restricted to that side's own declarations, and a required name outside the result has no strict representation.
- **Raise** when a `$ref` has *constraining* (not annotation) siblings. Cheaper and arguably better here, because the MCP path already has a fallback for exactly that, and a loud failure beats a tool whose accept set quietly inverted.

---

## 2. `convert_schemas_to_strict` falls back on 2 of 17 non-convertible shapes (opt-in path)

`MCPConfig.convert_schemas_to_strict` documents itself as *"a best-effort conversion, so some schemas may not be convertible"*, and `mcp/util.py` implements exactly that — `try: ensure_strict_json_schema(...) / except: serve non-strict`. The architecture is right. The problem is that the detector for "not convertible" is *"did `ensure_strict_json_schema` raise?"*, and it raises only for `additionalProperties` present-and-not-`False`, and for a non-dict subschema. It never checks the strict keyword allowlist.

I ran 20 schemas of the kind a third-party MCP server plausibly advertises through `MCPUtil.to_function_tool(..., convert_schemas_to_strict=True)`, then put each converted result through `toStrictJsonSchema()`:

- **2** raised → correctly served non-strict (`additionalProperties: true`; a boolean subschema). These are the control, and they discriminate.
- **15** passed, got `strict_json_schema=True`, and are **rejected by the vendor transformer**: `allOf` (2-member, and the `allOf: [{$ref}, {$ref}]` OpenAPI idiom), `uniqueItems`, `prefixItems`, tuple-form `items`, `patternProperties`, `propertyNames`, `unevaluatedProperties`, `not`, `if`/`then`, `contains`, `minProperties`, `dependentRequired`, nested `$id`, and `{"type": "array"}` with no `items`.

Verified end-to-end rather than inferred — `MCPUtil.to_function_tool(...)` returns `strict_json_schema=True` for all of them, and `openai_responses.py:2042` / `chatcmpl_converter.py:919` put that straight onto the wire as `"strict": true`.

Note that `_ensure_strict_json_schema` already *walks* `allOf` (`strict_schema.py:177`) — it recurses into members with `len(all_of) > 1` and leaves the keyword in place, which is unsupported in strict mode. So the multi-member case is visited and then shipped.

The narrow fix is to make the fallback's trigger match its stated meaning: validate the converted result against the strict-supported keyword set and raise if anything is left over, so the existing `except` does the job it was written for. That also automatically covers the `$ref` case above if you take the "raise" option in §1.

**No API key needed for any of this** — `from openai.lib.transform import toStrictJsonSchema` (JS) or the equivalent local check gives a per-schema accept/throw verdict offline, so this is CI-testable.

---

Scope caveat: everything above is measured against `openai@7.4.0`'s transformer and `ajv` 2020-12. I did not make a live API call, so I'm claiming what the vendor's own client-side strict transformer accepts, not an observed 400.

SnowingFox:
Closes #4353

### Root cause

Two defects in `agents/strict_schema.py`, both about what counts as "not convertible" to OpenAI's strict mode:

1. **`$ref` sibling merge is parent-wins (default path, silent).** `_ensure_strict_json_schema` inlines a `$ref` via `json_schema.update({**resolved, **json_schema})`, so sibling keys override the referent. Under JSON Schema 2020-12 a `$ref` and its siblings form an *intersection*, not a precedence. For annotation-only siblings (`description`, `title`) parent-wins equals the intersection, which is why pydantic-emitted schemas are safe. But a *constraining* sibling (`properties`/`required`/`type`/`enum`) deletes the referent's own constraints: the repro in #4353 drops `T.b` and then closes the object with `additionalProperties: false`, turning a schema that accepted `{node:{a,b}}` into one that only accepts `{node:{a}}` — a silently inverted accept set with no 400, because the converted schema is itself valid strict-mode JSON Schema.

2. **`convert_schemas_to_strict` falls back on only 2 of the 17 shapes the issue sampled (opt-in path).** `mcp/util.py` decides "not convertible" by "did `ensure_strict_json_schema` raise?", but the function only raises for `additionalProperties` and non-dict subschemas; it never checks the strict keyword allowlist. The 15 shapes from the issue, plus Draft 7 `additionalItems` (which the vendor transformer also rejects) — 16 shapes total: `uniqueItems`, `prefixItems`, tuple-form `items`, `additionalItems`, `patternProperties`, `propertyNames`, `unevaluatedProperties`, `not`, `if`/`then`, `contains`, `minProperties`, `dependentRequired`, nested `$id`, multi-member `allOf`, the `allOf: [{$ref}, {$ref}]` OpenAPI idiom, and array-without-`items` — convert "successfully", get `strict_json_schema=True`, and are then rejected by the API at request time.

### Fix

- **Finding 1 — raise on constraining `$ref` siblings.** When a `$ref` carries any sibling outside the annotation/definition-map allowlist, `_ensure_strict_json_schema` raises `UserError` instead of silently merging parent-wins. Annotation-only siblings still expand exactly as before. This takes the issue's preferred option: a loud failure beats a tool whose accept set quietly inverted, and it composes with the existing MCP fallback.

- **Finding 2 — validate the converted result against the strict keyword allowlist.** `ensure_strict_json_schema` now walks the converted schema and raises `UserError` if any strict-unsupported keyword survives conversion (plus tuple-form `items`, `additionalItems`, array-without-`items`, and nested `$id`). The existing `except` in `mcp/util.py` then serves the original non-strict schema, so `convert_schemas_to_strict=True` falls back correctly on all 15 shapes.

### Tests

Added to `tests/test_strict_schema.py`:

- `test_ref_with_constraining_sibling_is_rejected` and `test_ref_with_constraining_type_sibling_is_rejected` — constraining siblings raise. The control `test_ref_with_annotation_sibling_is_still_expanded` proves annotation-only siblings still expand.
- `test_non_convertible_shapes_are_rejected` — parametrized over the 16 non-convertible shapes, all must raise.

All 18 new cases fail on `main` and pass with the fix. Full suites pass: `tests/test_strict_schema.py` (53), `tests/test_strict_schema_oneof.py`, `tests/test_function_schema.py`, `tests/test_function_tool.py`, `tests/test_function_tool_decorator.py`, `tests/test_output_tool.py`, and `tests/mcp/test_mcp_util.py` (109).

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -36,6 +36,63 @@ _UNVALIDATED_REF_ERROR = (
     "JSON schema contains a reference whose target was not validated for strict mode."
 )
 
+# Keywords that may legally sit alongside a `$ref` without changing the accepted values:
+# JSON Schema annotation keywords plus definition maps (which are inert metadata). Any
+# other sibling of a `$ref` is a *constraining* keyword (e.g. `properties`, `required`,
+# `type`, `enum`), which JSON Schema 2020-12 combines with the referent as an
+# intersection rather than as an override.
+_REF_ANNOTATION_SIBLINGS = frozenset(
+    {
+        "$comment",
+        "$defs",
+        "default",
+        "definitions",
+        "deprecated",
+        "description",
+        "examples",
+        "readOnly",
+        "title",
+        "writeOnly",
+    }
+)
+
+# Keywords OpenAI's strict mode cannot represent. Mirrors the unsupported-keyword list in
+# the vendor reference implementation (openai-node `toStrictJsonSchema`); a schema that
+# still contains any of these after conversion is not convertible and must be rejected so
+# callers (e.g. the MCP `convert_schemas_to_strict` path) fall back to the original
+# non-strict schema instead of shipping one the API will reject at request time.
+_STRICT_UNSUPPORTED_KEYWORDS = frozenset(
+    {
+        "$anchor",
+        "$dynamicAnchor",
+        "$dynamicRef",
+        "$recursiveAnchor",
+        "$recursiveRef",
+        "allOf",
+        "contains",
+        "contentEncoding",
+        "contentMediaType",
+        "contentSchema",
+        "dependentRequired",
+        "dependentSchemas",
+        "dependencies",
+        "else",
+        "if",
+        "maxContains",
+        "maxProperties",
+        "minContains",
+        "minProperties",
+        "not",
+        "patternProperties",
+        "prefixItems",
+        "propertyNames",
+        "then",
+        "unevaluatedItems",
+        "unevaluatedProperties",
+        "uniqueItems",
+    }
+)
+
 
 class _NodeBudget:
     """Tracks conversion state across the recursion."""
@@ -65,7 +122,7 @@ def ensure_strict_json_schema(
     if schema == {}:
         return copy.deepcopy(_EMPTY_SCHEMA)
     budget = _NodeBudget(_MAX_SCHEMA_NODES, reject_open_objects=_reject_open_objects)
-    return _ensure_strict_root(
+    result = _ensure_strict_root(
         _ensure_strict_json_schema(
             schema,
             path=(),
@@ -73,6 +130,8 @@ def ensure_strict_json_schema(
             budget=budget,
         )
     )
+    _raise_if_strict_unsupported(result)
+    return result
 
 
 def _ensure_strict_root(schema: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for investigating #4353. We do not intend to support arbitrary JSON Schema or OpenAPI composition in strict conversion. The required fix is narrower: reject a `$ref` with validation-bearing siblings before its meaning can be changed, while preserving the existing annotation-only and singleton-`allOf` paths. MCP conversion should then use its existing non-strict fallback, and direct callers can opt out with `strict_json_schema=False`.

Please remove the global unsupported-keyword validator and the broad 16-shape matrix from #4354. We should not make an OpenAI Node version's keyword blacklist a permanent Agents SDK contract or add support for `components` without a demonstrated strict-mode requirement.

seratch:
Thanks for trying to resolve the issue. Closing this PR in favor of https://github.com/openai/openai-agents-python/pull/4356

PR #4366, opened 2026-08-11
fix(runner): close the model stream when a streamed turn ends in a terminal failure

Luccacvb:
### Summary

On the streamed path, `run_single_turn_streamed` iterates the model stream with a bare `async for` and raises from inside the loop body when a turn ends in `response.incomplete`, `response.failed`, `error` or `response.error`. `async for` does not close its iterator, so the generator chain stays suspended at its yield and the provider's `finally` never runs.

With the default `OpenAIResponsesModel`, a turn truncated by `max_output_tokens` therefore leaves the httpx SSE response open, and the `ResponseSpanData` is never ended:

```text
span_start ResponseSpanData ... (no span_end) ... span_end TurnSpanData ... trace_end
close calls immediately after run: []
close calls after gc.collect():    []
```

The adapter annotates that span deliberately, and says why it depends on the consumer:

> A consumer that stops at this event closes the generator, which raises `GeneratorExit` at the yield below and skips the raise after the loop, so the span has to be annotated here or not at all.

So `record_model_error_on_span` runs, and the span carrying that error is then dropped instead of exported, on a turn that ended for an ordinary reason.

Against `api.openai.com` with `gpt-4o-mini`, an agent configured with `ModelSettings(max_tokens=16)` and instructions that ask for a long answer reaches this on the first try:

```text
before: trace_start ... start:ResponseSpanData -> end:TurnSpanData -> end:AgentSpanData -> trace_end
after:  trace_start ... start:ResponseSpanData -> end:ResponseSpanData -> end:TurnSpanData -> ... -> trace_end
```

The parent spans close over a response span that never ends, and the teardown also raises `ValueError: <Token ...> was created in a different Context`. Both stop with this change.

`OpenAIResponsesWSModel` is the severe case: `_ws_request_lock` is acquired for the request and released in the same skipped `finally`, so the lock stays held until a collection finalizes the abandoned generator. A run started before that blocks on it.

```text
after the failed turn: _ws_request_lock.locked() = True
turn 2 on the same model instance: hangs on the request lock
```

This is the consumer half of a contract the provider side already implements. #3689, #4066, #4077, #4133, #4156 and #4221 made provider streams and trace scopes close deterministically; `stream_response_with_retry` already re-raises `GeneratorExit` after closing its inner stream, and the Responses websocket generator has a dedicated branch for being closed at a terminal event. This PR reconnects that cleanup chain at the two consumer-side delegation points where the close is currently not propagated.

**Why both hops.** Each one fixes a different half. `model_run_context_stream` delegates with a bare `async for ... yield`, which does not forward `aclose()`, so closing only at the call site never reaches the provider. Closing only in the delegate releases the resources, but the run loop still abandons its generator, so the close runs in asyncio's async-generator finalizer task, in another context: the response span is then ended after `trace_end` instead of inside its parent, and the `ValueError` below still fires.

| `_run_context.py` | `run_loop.py` | ws lock released | httpx `close()` | `ResponseSpanData` ended in order |
| --- | --- | --- | --- | --- |
| — | — | no, next run blocks | `[]` | no |
| — | yes | no, next run blocks | `[]` | no |
| yes | — | yes | `['close']` | no, arrives after `trace_end` |
| yes | yes | yes | `['close']` | yes |

The new regression fails in the first three rows and passes in the last.

This mirrors #4133, which wrapped its delegating generators in `contextlib.aclosing` and narrowed their return annotations to `AsyncGenerator[..., None]` so `aclosing` type-checks. The `model_retry.py` annotation here exists for that same reason and nothing else. Neither `agents.run_internal.*` nor `agents.models._run_context` appears in `tests/fixtures/released_api_contract.json`.

**On the `ValueError` this removes.** Today the abandoned generator is finalized later by the async-generator hook, in a fresh context, so `_MODEL_RUN_OWNER.reset(token)` raises `ValueError: <Token ...> was created in a different Context`. Closing in the owning task makes that reset valid, so the error disappears rather than being suppressed; this PR does not add a `try/except` around the reset (cf. #3235, #4221).

Provider cleanup now runs inline on the terminal-failure path. For the built-in adapters that is bounded: the websocket path releases a lock synchronously, and the HTTP path closes through `_close_stream_allowing_background_completion` (added in #4133, ported to this adapter in #4156), so a cancellation arriving during that close is already handled. One consequence worth naming: a `BaseException` escaping a provider's own `finally` now surfaces in place of the terminal `ModelBehaviorError`, since the close runs while that error propagates. Suppressing it here would hide a cancellation, so this PR leaves it.

The `run_loop.py` diff is ~55 lines, of which all but a few are re-indentation of the existing loop body; `git diff -w` shows the semantic change.

Deliberately out of scope:

- The runner's non-streamed path wraps a coroutine with `model_run_context(...)`, so it has no equivalent gap. `OpenAIResponsesWSModel._fetch_response(stream=False)` does have the same shape — a bare `async for` over `_iter_websocket_response_events` that raises from inside the loop, skipping the `finally` that releases `_ws_request_lock` — and is left out deliberately, so this PR stays on the runner and one demonstrated failure.
- A consumer that breaks out of `stream_events()` without `cancel()` still leaks. That is the documented caller-side contract, and `cancel()` already closes the provider stream; neither changes here.

### Test plan

`test_streamed_terminal_failure_closes_the_model_stream` drives a model whose `stream_response` records `asyncio.current_task()` where it starts iterating and again in its `finally`, over `response.incomplete`, `response.failed`, `error` and `response.error`. Asserting that those are the same task distinguishes a close performed by the run loop from a later finalization by asyncio's async-generator hook, so it does not call `aclose()` and does not force a collection. It fails on all four parametrizations when either hop is reverted, which is the third row of the table above.

### Issue number

N/A. Found while auditing the streamed run loop against `.agents/references/runner-lifecycle.md`.

### Checks

### DIFF
--- a/src/agents/models/_run_context.py
+++ b/src/agents/models/_run_context.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-from collections.abc import AsyncIterator, Iterator
-from contextlib import contextmanager
+from collections.abc import AsyncGenerator, Iterator
+from contextlib import aclosing, contextmanager
 from contextvars import ContextVar
 from typing import TypeVar
 
@@ -24,9 +24,12 @@ def get_model_run_owner() -> object | None:
 
 
 async def model_run_context_stream(
-    stream: AsyncIterator[T],
+    stream: AsyncGenerator[T, None],
     owner: object,
-) -> AsyncIterator[T]:
+) -> AsyncGenerator[T, None]:
+    # `aclosing` forwards an early `aclose()` on this generator to the delegate, so the
+    # delegate's cleanup runs deterministically instead of waiting for garbage collection.
     with model_run_context(owner):
-        async for item in stream:
-            yield item
+        async with aclosing(stream):
+            async for item in stream:
+                yield item
--- a/src/agents/run_internal/model_retry.py
+++ b/src/agents/run_internal/model_retry.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import random
-from collections.abc import AsyncIterator, Awaitable, Callable, Mapping
+from collections.abc import AsyncGenerator, AsyncIterator, Awaitable, Callable, Mapping
 from inspect import isawaitable
 from typing import Any
 
@@ -569,7 +569,7 @@ async def stream_response_with_retry(
     conversation_id: str | None,
     failed_retry_attempts_out: list[int] | None = None,
     replay_unsafe_request: bool = False,
-) -> AsyncIterator[TResponseStreamEvent]:
+) -> AsyncGenerator[TResponseStreamEvent, None]:
     request_attempt = 1
     policy_attempt = 1
     failed_policy_attempts = 0
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import asyncio
 import dataclasses as _dc
 from collections.abc import Awaitable, Callable
+from contextlib import aclosing
 from functools import partial
 from typing import Any, TypeVar, cast
 from uuid import uuid4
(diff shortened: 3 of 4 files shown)


seratch (changes requested):
The underlying bug is real, and the two `aclosing` hops are the correct boundary: the run loop must close its wrapper in the owning task, and the wrapper must forward closure to the retry/provider stream. I reproduced the base failure and verified that this branch releases the Responses websocket request lock so a concurrently waiting second run completes.

Before merge, please add a controlled websocket regression that starts run B while run A holds `_ws_request_lock`, delivers a terminal failure to A, then proves B completes on the surviving connection and the lock is unlocked. The current test proves same-task finalization, but it does not assert the shared-resource survivor invariant that motivates the highest-impact case.

Luccacvb:
Added `test_ws_terminal_failure_frees_the_request_lock_for_a_waiting_run`.

It verifies that a terminal failure in run A releases `_ws_request_lock`, allowing run B to complete on the same connection. The test uses explicit synchronization, fails on the merge base by timing out with a named assertion, and confirms that the connection is reused and remains open.

I kept both tests because they cover different cleanup boundaries: this test pins lock release, while `test_streamed_terminal_failure_closes_the_model_stream` pins stream closure in the iterating task. Worth noting that this test still passes with the `run_loop` hop reverted, asyncio's finalizer returns the lock eventually, so the same-task test is what keeps that hop honest.

I also moved `DummyWSClient` to module scope to avoid duplication.

seratch (approved):

PR #4373, opened 2026-08-12
fix(mcp): isolate cached tool list results

hsusul:
### Summary

Return snapshots from `MCPServer.list_tools()` and `cached_tools` so callers cannot mutate the server's shared tool cache through a public result.

The change preserves the tool objects, their order, filtering, cache invalidation, and fetch behavior while isolating only the returned list containers.

### Test plan

- Added a parameterized regression covering mutation through both `list_tools()` and `cached_tools`. On clean `upstream/main`, both cases failed because removing one of two returned tools removed it from subsequent cached results without another server request.
- `uv run pytest tests/mcp/test_caching.py::test_cached_tool_results_do_not_expose_the_shared_cache -q` — 2 passed.
- `uv run pytest tests/mcp/test_caching.py -q` — 4 passed.
- Repeated the focused async regression 10 times — 20 parameterized cases passed.
- `env UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh` — format, lint, typecheck, and full tests passed.

No API keys, paid calls, or live services were used.

### Issue number

N/A — focused fix with a deterministic local regression.

### Checks

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -842,7 +842,7 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
 
     @property
     def cached_tools(self) -> list[MCPTool] | None:
-        return self._tools_list
+        return list(self._tools_list) if self._tools_list is not None else None
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the focused patch. I confirmed that, with `cache_tools_list=True`, `list_tools()` and `cached_tools` can expose the internal list, so mutating the returned container changes later cached results. The regression test establishes that mechanism, but the PR does not show a concrete supported caller workflow or user-visible failure that requires in-place mutation. Per-request tool selection is already supported through `tool_filter`, and callers that need local list manipulation can copy the result.

I am going to close this PR for now because the demonstrated impact does not justify adding snapshot semantics to the released API contract. If you can provide a real supported scenario where ordinary caller behavior mutates this container and causes a meaningful failure despite those alternatives, we can revisit the narrow snapshot behavior.

PR #4375, opened 2026-08-12
fix(mcp): stop retry backoff from doubling without a ceiling

rxits:
### Summary

The retry backoff multiplier grows without a ceiling, so `max_retry_attempts=-1` stops retrying in practice instead of retrying indefinitely.

Both retry loops compute the delay as `retry_backoff_seconds_base * 2**n` with nothing bounding `n`:

- `_MCPServerWithClientSession._run_with_retries` — the generic `list_tools()` / `call_tool()` path
- `MCPServerStreamableHttp.call_tool` — the streamable HTTP override, in both of its retry branches

With a finite `max_retry_attempts` the budget bounds the exponent, so the schedule stays reasonable. With `max_retry_attempts=-1` nothing does. At the default `retry_backoff_seconds_base=1.0`, the 10th delay is about 8.5 minutes, the 20th about 6 days, and the 30th about 34 years. A server that is down long enough for the budget to be unlimited is exactly the case where the loop should keep polling, and it is the case where it silently stops.

This is the behavior `.agents/references/local-mcp-server-lifecycle.md` describes as "Generic `list_tools()` and `call_tool()` retries use the configured attempt count and backoff" — the configured attempt count of "unlimited" is not actually honored once the delay outgrows the process.

The fix clamps the multiplier at 64 times the configured base, in one shared `_retry_backoff_seconds()` helper that replaces the three copies of the unbounded expression. The ceiling is expressed as a multiplier rather than an absolute number of seconds so that `retry_backoff_seconds_base` keeps its meaning at any scale: a base of `0.1` still tops out at 6.4s, a base of `5.0` at 320s. No new constructor parameter is added; `_MAX_RETRY_BACKOFF_EXPONENT` is a module-private constant, so the exact ceiling is easy to adjust if you would prefer a different one.

Only the unlimited-retry mode changes behavior. Any finite `max_retry_attempts` of 7 or less produces an identical schedule to before.

### Test plan

Two tests added to `tests/mcp/test_client_session_retries.py`, one per retry loop, each driving 9 consecutive failures with `max_retry_attempts=-1` and `retry_backoff_seconds_base=1.0` and asserting the recorded sleep schedule is `[1, 2, 4, 8, 16, 32, 64, 64, 64]`.

Both fail on `main` at index 7 with `128.0 != 64.0`, verified by stashing only `src/agents/mcp/server.py` and re-running the new tests against the unmodified source:

```
FAILED tests/mcp/test_client_session_retries.py::test_backoff_stops_doubling_with_unlimited_retries
FAILED tests/mcp/test_client_session_retries.py::test_streamable_http_backoff_stops_doubling_with_unlimited_retries
2 failed, 29 deselected
```

The existing retry tests cannot catch this: they either pin `retry_backoff_seconds_base=0` or stop after three failures, before the exponent grows far enough to matter.

`.agents/skills/code-change-verification/scripts/run.sh` run in full. `make format`, `make lint` and `make typecheck` are clean.

`make tests` reports one failure, `tests/test_run_step_execution.py::test_multiple_tool_calls_bound_cancelled_sibling_self_rescheduling_cleanup`. It is **pre-existing and unrelated** to this change:

- it reproduces on clean `main` with the same command and no diff applied (`1 failed, 8113 passed`)
- with this change applied the same run is `1 failed, 8115 passed` — the delta is exactly the two tests added here
- it passes when run on its own, so it depends on full-suite `-n auto` load
- this change touches only `src/agents/mcp/server.py`; the failing test exercises the `HostedMCPTool` approval flow, which does not reach the local MCP client-session retry path

### Issue number

N/A

### Checks

- [ ] I've confirmed all verification steps pass — all steps pass except the pre-existing `test_run_step_execution.py` failure documented above, which also fails on clean `main`

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -515,6 +515,10 @@ class _UnsetType:
 
 _UNSET = _UnsetType()
 
+# Retry backoff stops doubling once this many backoffs have been taken, so the delay never
+# exceeds `retry_backoff_seconds_base` times 2**_MAX_RETRY_BACKOFF_EXPONENT.
+_MAX_RETRY_BACKOFF_EXPONENT = 6
+
 if TYPE_CHECKING:
     from ..agent import AgentBase
 
@@ -880,7 +884,8 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
             max_retry_attempts: Number of times to retry failed list_tools/call_tool calls.
                 Defaults to no retries.
             retry_backoff_seconds_base: The base delay, in seconds, used for exponential
-                backoff between retries.
+                backoff between retries. The delay stops doubling once it reaches 64 times
+                this value.
             message_handler: Optional handler invoked for session messages as delivered by the
                 ClientSession.
             require_approval: Approval policy for tools on this server. Accepts "always"/"never",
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
The underlying problem is real: `max_retry_attempts=-1` is an intentionally supported unlimited mode, and unbounded exponential delays eventually make it operationally ineffective. Consolidating the calculation in a shared helper is a good direction.

Before merge, please replace the private 64x ceiling with an appended public option such as `retry_backoff_seconds_max: float | None = None`. The `None` default should preserve the released schedule, while a configured value caps the computed delay in both the generic and Streamable HTTP retry paths. Please add regression coverage showing that unset finite and unlimited schedules still grow beyond 64, and that a configured maximum saturates both paths. The current patch also changes finite retry budgets of 8 or more, so the claim that only unlimited retries change is not accurate.

seratch:
Thanks again for reporting the issue. It was resolved by https://github.com/openai/openai-agents-python/pull/4379 and its commit includes your contribution credit.

PR #4376, opened 2026-08-12
fix(EncryptedSession): add strict mode to reject non-encrypted items

SashaMIT:
## Summary

`EncryptedSession` wraps an underlying session store with Fernet AEAD so stored conversation items are encrypted and integrity-protected at rest. The read path, however, currently returns any stored item that lacks the `__enc__` envelope completely verbatim: `_unwrap` passes non-envelope items straight through, so `get_items` and `pop_item` feed unauthenticated plaintext into model context with no signal to the operator.

That passthrough is useful when migrating a plaintext store onto encrypted storage, so this PR keeps it as the default and adds an opt-in strict mode rather than changing behavior:

- New constructor flag `allow_plaintext_passthrough` (default `True`, behavior identical to today).
- When set to `False`, non-envelope items are dropped with a warning log on both read paths (`get_items` via `_unwrap_valid_items`, and the `pop_item` retry loop), so only authenticated ciphertext reaches the model.

This is a hardening / fail-closed option, not a vulnerability fix: planting items already requires write access to the underlying store, but operators who treat the store as untrusted at rest now have a way to enforce the integrity the wrapper otherwise implies.

A minimum-length check for raw-string `encryption_key` values was considered but left out to avoid breaking existing deployments. Happy to add a short note to `docs/sessions/encrypted_session.md` if you would like that in this PR.

## Test plan

- [x] `pytest tests/extensions/memory/test_encrypt_session.py`: 24 passed (4 new)
- [x] New coverage: default passthrough unchanged, strict mode drops non-envelope items with a warning, strict mode still reads valid envelopes, `pop_item` skips non-envelope items
- [x] `ruff check`, `ruff format --check`, and `mypy` clean on touched files

### DIFF
--- a/src/agents/extensions/memory/encrypt_session.py
+++ b/src/agents/extensions/memory/encrypt_session.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 
 import base64
 import json
+import logging
 from typing import Any, Literal, TypeGuard, cast
 
 from cryptography.fernet import Fernet, InvalidToken
@@ -40,6 +41,8 @@ from ...items import TResponseInputItem
 from ...memory.session import SessionABC, _call_session_method, _get_session_wrapper
 from ...memory.session_settings import SessionSettings, resolve_session_limit
 from ...run_context import RunContextWrapper
+
+logger = logging.getLogger(__name__)
 
 
 class EncryptedEnvelope(TypedDict):
@@ -117,6 +120,7 @@ class EncryptedSession(SessionABC):
         underlying_session: SessionABC,
         encryption_key: str,
         ttl: int = 600,
+        allow_plaintext_passthrough: bool = True,
     ):
         """
         Args:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the thoughtful hardening proposal. The current plaintext passthrough is real, but this PR does not establish a supported scenario where `EncryptedSession` must treat a writable backing store as an untrusted integrity boundary. The new flag would add a permanent behavior mode while leaving the default unchanged, and it cannot protect against deletion, reordering, or replay by the same store. It also makes `pop_item()` destructively scan past rejected tail items, which conflicts with one-item pop and retry-rewind ownership. I am going to close this PR.

SashaMIT:
Thanks for the careful read, that makes sense. The pop scan-past behavior was the part I was least sure of, and you're right that a writable store can delete or reorder regardless, so the flag only ever covered a slice of the integrity question. Appreciate the quick review either way.

PR #4377, opened 2026-08-12
fix(tracing): reject non-positive max batch sizes

hansu650:
### Summary

This pull request fixes `BatchTraceProcessor` silently stranding queued traces when `max_batch_size` is zero or negative. It now rejects non-positive batch sizes before creating the queue or worker thread, while preserving existing behavior for positive values.

### Test plan

- `pytest tests/test_trace_processor.py -q` — 58 passed
- `ruff format` and `ruff check` — passed
- `python .github/scripts/check_optional_truthiness.py src/agents` — passed
- Targeted mypy and Pyright checks for the changed runtime and test files — passed
- The repository verification wrapper could not complete on Windows because `make` is not installed. Its manual full-suite equivalent was attempted; unrelated Windows-only baseline failures (sandbox symlink privileges, existing full-project typing errors, and tracing test ordering also reproduced on unchanged upstream) remain outside this two-file change.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -562,6 +562,9 @@ class BatchTraceProcessor(TracingProcessor):
             schedule_delay: The delay between checks for new spans to export.
             export_trigger_ratio: The ratio of the queue size at which we will trigger an export.
         """
+        if max_batch_size <= 0:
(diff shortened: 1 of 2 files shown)


seratch:
We don't expect this value to be negative in real-world use cases, so let's close this PR.

PR #4382, opened 2026-08-13
fix(voice): reject non-positive audio frame rates

hsusul:
### Summary

`_buffer_to_audio_file` validates sample width, buffer dtype, channel count, and multichannel frame completeness, but previously did not validate the frame rate. A zero or negative `frame_rate` reached `wave.setframerate` and surfaced as a low-level `wave.Error` instead of the `UserError` used by the adjacent input validations.

This adds an up-front `frame_rate <= 0` check so invalid rates fail consistently and actionably. Valid audio conversion is unchanged.

The earlier negative-history-limit changes have been removed to preserve the released SQLite behavior documented by v0.20.0 and #4001.

### Test plan

- Added `test_audio_input_rejects_non_positive_frame_rate`, parametrized over `0` and `-8000`.
- `uv run pytest -q tests/voice/test_input.py` — 24 passed.
- `env UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh` — passed:
  - Ruff format: 897 files unchanged.
  - Ruff lint: passed.
  - mypy and Pyright: passed.
  - Full test suite: passed.
- `git diff --check` — passed.

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/voice/input.py
+++ b/src/agents/voice/input.py
@@ -22,6 +22,9 @@ def _buffer_to_audio_file(
     if sample_width not in {1, 2, 3, 4}:
         raise UserError("Sample width must be between 1 and 4 bytes")
 
+    if frame_rate <= 0:
+        raise UserError("Frame rate must be greater than zero")
+
     if buffer.dtype not in (np.int16, np.float32):
         raise UserError("Buffer must be a numpy array of int16 or float32")
 
--- a/tests/voice/test_input.py
+++ b/tests/voice/test_input.py
@@ -116,6 +116,14 @@ def test_audio_input_rejects_non_positive_channels(channels):
         audio_input.to_audio_file()
 
 
+@pytest.mark.parametrize("frame_rate", [0, -8000])
+def test_audio_input_rejects_non_positive_frame_rate(frame_rate):
+    audio_input = AudioInput(buffer=np.zeros(4, dtype=np.int16), frame_rate=frame_rate)
+
+    with pytest.raises(UserError, match="Frame rate must be greater than zero"):
+        audio_input.to_audio_file()
+
+
 def test_buffer_to_audio_file_invalid_dtype():
     # Create a buffer with invalid dtype (float64)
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
@hsusul Thanks for suggesting this but your another PR https://github.com/openai/openai-agents-python/pull/4383 is quite similar, so we may want to avoid taking our time to review these small PRs one by one. If you don't mind, can you do thorough investigation to detect similar opportunities to improve and include all of them in this PR?

seratch (changes requested):
Thanks for expanding this after the earlier review. I reran the new tests against the merge base: the frame-rate test demonstrates a real, bounded exception-contract gap, but the negative-history work should not be included. v0.20.0 and #4001 intentionally preserve SQLite's historical negative-limit behavior (`limit=-1` returns all items), while this branch changes it to an empty history. The OpenAI-specific negative case is not a demonstrated supported scenario, and silently treating invalid negative input as no history is not a safe cross-backend compatibility fix.

Please keep the frame-rate validation and regression test, remove both negative-limit commits and their tests, and update the title and description to the remaining voice scope. This preserves released session behavior while retaining the demonstrated improvement.

FU-max-boop (commented):
Reviewed and tested at `17a8814b`.

Implementation scope checked against the latest release boundary (`v0.20.0`): reject non-positive `frame_rate` values with the SDK's adjacent `UserError` contract before WAV encoding; preserve positive frame-rate behavior; do not broaden the change into unrelated audio validation.

- All 24 tests in `tests/voice/test_input.py` pass.
- I independently exercised `0`, `-1`, and `-8000` and confirmed the same actionable `UserError` is raised before WAV encoding.
- A positive 8 kHz stereo case still writes the expected WAV metadata and frame count.
- Ruff format and lint pass for both changed files.
- The complete current diff is limited to the voice validation and regression test; the negative-history changes from the earlier review are gone.

This now looks focused and compatible to me.

seratch (approved):

PR #4385, opened 2026-08-13
fix(tools): preserve empty description overrides

hsusul:
### Summary

Explicit empty description overrides were treated as absent by function tools, handoffs, and realtime handoffs. For a documented function or named agent, passing `description_override=""` or `tool_description_override=""` silently restored the docstring or generated handoff description.

Resolve description fallbacks only when the override is `None`. Empty strings now remain explicit values; non-empty overrides and omitted overrides keep their existing behavior. Tool-name handling is unchanged.

### Test plan

- Added public regressions for function tools, standard handoffs, and realtime handoffs.
- Red on `upstream/main`: the function test restored its docstring; both handoff tests restored the generated description.
- Focused files: `115 passed`.
- `make format` and `make lint` passed.
- `make typecheck`: mypy clean for 305 source files; Pyright 0 errors.
- Parallel suite: `8339 passed, 28 skipped`.
- Serial suite: `77 passed, 11 skipped, 55 deselected`.
- `git diff --check` passed.

### Issue number

No issue.

### Checks

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -477,7 +477,11 @@ def function_schema(
     return FuncSchema(
         name=func_name,
         # Ensure description_override takes precedence even if docstring info is disabled.
-        description=description_override or (doc_info.description if doc_info else None),
+        description=(
+            description_override
+            if description_override is not None
+            else (doc_info.description if doc_info else None)
+        ),
         params_pydantic_model=dynamic_model,
         params_json_schema=json_schema,
         signature=sig,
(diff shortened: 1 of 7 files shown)


seratch:
The patch correctly changes the empty-string fallback behavior, but this PR does not demonstrate a supported user scenario where an empty tool description is required or produces a meaningful outcome. Tool descriptions provide model-facing guidance, and removing that guidance can reduce tool-selection quality. The same outcome is already available through `use_docstring_info=False`, a `FunctionTool` with an empty description, or the public `Handoff.tool_description` field.

Because this also changes released behavior for callers whose computed override is empty and currently receives the inferred fallback, I do not think the permanent behavior change is justified by the available evidence. I am going to close this PR for now. We can revisit it with a concrete scenario showing why the existing approaches are insufficient and what observable benefit specifically requires an empty override.

Issue #4390, opened 2026-08-13
strict_mode: non-null default values are left in the schema and rejected on Azure


antoniojose-perez:
`ensure_strict_json_schema()` only strips the `default` key from a property when its value is exactly `None`:

```python
# strip `None` defaults as there's no meaningful distinction here
# the schema will still be `nullable` and the model will default
# to using `None` anyway
if json_schema.get("default", NOT_GIVEN) is None:
    json_schema.pop("default")
```

Any field with a **non-null** default (e.g. an enum, a `Decimal`, an `int`, a plain string) keeps its `default` key in the generated strict schema. OpenAI's Structured Outputs API rejects that schema at call time with:

```
'default' is not permitted within a property definition
```

This forces users to disable `strict_mode` entirely for any tool whose Pydantic model has a field with a non-null default, losing the schema guarantees strict mode is meant to provide.

### Repro

```python
from decimal import Decimal
from enum import Enum
from pydantic import BaseModel
from agents import function_tool

class Currency(str, Enum):
    EUR = "EUR"
    USD = "USD"

class Invoice(BaseModel):
    total: Decimal
    currency: Currency = Currency.EUR  # non-null default

@function_tool
def create_invoice(data: Invoice):
    ...

# create_invoice.params_json_schema still contains
# {"currency": {..., "default": "EUR"}}
# -> OpenAI API call fails with:
# "'default' is not permitted within a property definition"
```

### Expected behavior

`default` should be stripped from every property in the strict schema, not just when it's `None`.

### Why this is safe to fix unconditionally

A few lines above in the same function, every property is already forced into `required`:

```python
json_schema["required"] = list(properties.keys())
```

That's how strict mode simulates an "optional" field — it stays required, but its type allows `null`. Once a field is required, the model must always return an explicit value, so the Python-level `default` is never actually used by the API. It's dead weight in a strict schema regardless of whether its value is `None` or something else — the current `is None` check is an arbitrary distinction that shouldn't exist.

### Suggested fix

```python
# strip ALL defaults, not just `None`: since every property is marked
# `required` above, the model must always supply an explicit value for
# the field regardless of any Python-level default — so `default` is
# dead weight in a strict schema, and the API rejects it outright when
# non-null.
json_schema.pop("default", None)
```

(The `NOT_GIVEN` import becomes unused after this change if nothing else in the file relies on it.)

[...]

PR #4391, opened 2026-08-13
fix: strip non-null defaults from strict JSON schemas

alooshxl:
### Summary

`ensure_strict_json_schema()` only stripped a property's `default` when its value was exactly `None`. Any field with a non-null default (an enum, an `int`, a string, a `Decimal`, ...) kept its `default` key in the generated strict schema, and the OpenAI API then rejects the request at call time with `'default' is not permitted within a property definition`. This forced users to disable `strict_mode` for any tool whose Pydantic model has a field with a non-null default.

Strict conversion already forces every property into `required`, so the model must always emit an explicit value and a `default` can never take effect — it is dead weight in a strict schema regardless of its value. This change strips any remaining `default` unconditionally.

Implementation notes:

- The unconditional strip runs at the end of `_ensure_strict_json_schema`, after `$ref` handling. Stripping earlier (at the site of the old `is None` check) would reduce a `{"$ref": ..., "default": <non-null>}` node to a pure `$ref`, which would skip expansion and get rejected by `_reject_open_objects` mode. Placing it at the end keeps the `default` sibling as the expansion trigger, and the recursive pass strips it from the expanded result.
- The existing early strip of `default: None` is kept in place so a `$ref` whose only sibling is `default: null` continues to be preserved as a pure `$ref` instead of being expanded (unchanged behavior).
- Python-level runtime defaults are unaffected: omitted arguments are still filled in by the generated Pydantic model. Non-strict schemas (`strict_json_schema=False`) keep their `default` annotations.

### Test plan

- New tests in `tests/test_strict_schema.py`: non-null defaults are stripped from properties and non-object schemas, and a `$ref` with a non-null `default` sibling still expands, both in default mode and in `_reject_open_objects` mode.
- New test in `tests/test_function_schema.py` mirroring the issue repro: the strict schema carries no `default`, all parameters are required, the runtime default still applies when the argument is omitted, and a non-strict schema keeps the `default`.
- Updated the existing `test_function_schema.py` assertions that previously pinned the buggy behavior (non-null defaults surviving strict conversion into schemas the API rejects).
- All new and updated tests fail before the fix and pass after it.
- Ran the verification stack locally (`ruff format`, `ruff check`, optional-truthiness check, `mypy`, `pyright`, full parallel + serial pytest suites). Notes from this environment (Windows, commands run directly since `make` is unavailable): mypy/pyright report a handful of pre-existing errors only in untouched `src/agents/sandbox` modules (platform-specific tempfile typing), and the parallel suite has pre-existing local tracing/realtime failures — verified identical on a clean `main` checkout via stash-baseline diff. No check regressions are introduced by this change.

### Issue number

Closes #4390

### Checks

- [x] I've run `.agents/skills/code-change-verification/scripts/run.sh` (ran the equivalent command stack directly on Windows; see test plan)

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -361,9 +361,8 @@ def _ensure_strict_json_schema(
                 for i, entry in enumerate(all_of)
             ]
 
-    # strip `None` defaults as there's no meaningful distinction here
-    # the schema will still be `nullable` and the model will default
-    # to using `None` anyway
+    # Strip `None` defaults before the `$ref` handling below so that a `$ref` whose only
+    # sibling is a `default: null` stays a pure `$ref` and is preserved instead of expanded.
     if json_schema.get("default", NOT_GIVEN) is None:
         json_schema.pop("default")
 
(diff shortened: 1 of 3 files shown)


seratch:
As I mentioned in https://github.com/openai/openai-agents-python/issues/4390, removing all defaults may not be the right approach for real-world use cases.

PR #4395, opened 2026-08-13
fix: remove defaults from strict JSON schemas

hansu650:
### Summary

This pull request fixes strict JSON Schema conversion so every `default` annotation is removed before provider serialization, including non-null defaults generated for nested Pydantic models.

The conversion keeps `default: null` handling before reference processing to preserve pure references, while removing non-null defaults after `$ref` expansion so referenced schemas are still validated and normalized. Python runtime defaults and schemas created with strict mode disabled remain unchanged.

### Test plan

- Added regression coverage for `None`, numeric, boolean, and string defaults.
- Added coverage for `$ref` plus a non-null default in normal and `_reject_open_objects=True` conversion.
- Added public function-tool coverage for nested Pydantic enum defaults and non-strict parity.
- Focused schema tests: 176 passed.
- Ruff formatting and lint passed; serial tests passed.
- The full parallel suite reached 7,521 passed and 77 skipped. Its 19 Windows failures reproduce unchanged on the merge-base worktree and require symlink privileges or UTF-8 locale; full mypy/pyright likewise reproduces the same three pre-existing sandbox-file errors on the merge base.

### Issue number

Closes #4390

### Checks

- [x] I've run the Windows equivalent of `.agents/skills/code-change-verification/scripts/run.sh`

### DIFF
--- a/.agents/references/function-and-output-schema.md
+++ b/.agents/references/function-and-output-schema.md
@@ -16,7 +16,7 @@ Schema behavior is a compatibility boundary shared by Python callables, Pydantic
 
 - Strict conversion closes object schemas with `additionalProperties: false` and marks their declared properties required. Reject an explicit `additionalProperties: true` instead of silently changing its meaning.
 - Preserve the meaning of unions, intersections, definitions, and references. Normalize `oneOf` where required, process `allOf`, retain chained references, and merge a referenced schema with sibling keys without discarding the siblings.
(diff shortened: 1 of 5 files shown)


hansu650:
Closing in favor of #4391, which was opened while I was working and implements the same root fix and ref/default ordering. My branch additionally covered nested public FunctionTool/Pydantic schemas and synchronized the architecture reference, but those additions do not justify a competing PR. Thanks.

PR #4396, opened 2026-08-13
fix(tools): redact tool output value from output-type validation errors

hsusul:
### Summary

Function tools with a declared output type validate the returned value in `_validate_function_tool_output` (`src/agents/tool.py`). On a type mismatch, the raised `UserError` embedded the pydantic `ValidationError` in its message, and chained it as `__cause__`. A pydantic `ValidationError` repr includes the offending value (`input_value=...`), so this exposed the **tool's output value** in the exception message and traceback — even though `OPENAI_AGENTS_DONT_LOG_TOOL_DATA` is **on by default** and its documented contract is that tool call *inputs/outputs* are not exposed.

This is the output-side sibling of the already-merged tool *argument* redaction (#4182) and JSON-validation redaction (#4211). The same function already redacts argument-validation errors (both the message and the `__cause__`/`__context__` chain) and redacts the success-path output log three lines below the validation call, so the output-validation error path was the odd one out.

Behavior:
- `DONT_LOG_TOOL_DATA` enabled (default): raise a generic `UserError` (`Function tool <name> returned an output that does not match its declared output type`) with no `__cause__`/`__context__`, so the payload-bearing `ValidationError` is not reachable via the traceback.
- `DONT_LOG_TOOL_DATA` disabled: unchanged — the detailed message and `ValidationError` cause are preserved for debugging.

The redacted error is raised *outside* the `except` block so the `ValidationError` (which holds the output value) is not attached as `__cause__`/`__context__`, matching the existing argument path and `util/_json.py`.

Scope note: the only other typed tool-output validation site, `AgentToolUseTracker`/`RunItem.tool_call_output_item` in `items.py`, already uses a generic message and its `ValidationError` branch is not reached with invalid data on the invoke path (the value is validated here first), so no change is needed there.

### Test plan

Added two tests in `tests/test_error_logging_redaction.py`, mirroring the existing tool-argument redaction tests:

- `test_function_tool_output_validation_error_redacts_payload_when_tool_data_disabled` — with redaction on, the returned secret is absent from the error and `__cause__`/`__context__` are `None`. Fails on `main` (secret leaks via `input_value=...` and `__cause__`), passes with this change.
- `test_function_tool_output_validation_error_preserves_diagnostics_when_enabled` — with redaction off, the detail and `ValidationError` cause are preserved.

Verification (local, no API keys / paid calls / network):
- `uv run pytest tests/test_error_logging_redaction.py tests/test_function_tool.py tests/test_tool_output_conversion.py -q` → 209 passed
- `uv run pytest tests -q` → 8415 passed, 39 skipped
- `uv run ruff check src/agents/tool.py tests/test_error_logging_redaction.py` → passed
- `uv run ruff format --check` → already formatted
- `uv run mypy src` → no issues (305 files)
- `uv run pyright src/agents/tool.py` → 0 errors

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/tool.py
+++ b/src/agents/tool.py
@@ -2268,13 +2268,19 @@ def _validate_function_tool_output(
     """Validate a typed function output before it reaches hooks or serialization."""
     if output_type_adapter is None:
         return output
+    base_message = (
+        f"Function tool {tool_name} returned an output that does not match its declared output type"
+    )
     try:
         return output_type_adapter.validate_python(output)
     except ValidationError as error:
-        raise UserError(
-            f"Function tool {tool_name} returned an output that does not match its declared "
-            f"output type: {error}"
-        ) from error
+        if not _debug.DONT_LOG_TOOL_DATA:
+            raise UserError(f"{base_message}: {error}") from error
+    # Tool-data redaction is enabled: the ValidationError repr embeds the raw output value, so
+    # raise outside the ``except`` block. This keeps the payload-bearing ValidationError from
+    # being attached as the redacted error's ``__cause__``/``__context__`` (mirroring the tool
+    # argument validation path above).
+    raise UserError(base_message)
 
 
 def _validate_function_tool_callable_annotations(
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Can you resolve the remaining codex feedback?

seratch (approved):

PR #4397, opened 2026-08-13
fix(sandbox): snapshot HTTP proxy headers

hsusul:
### Summary

`HttpProxySink` retained the caller-owned headers dictionary. Mutating that dictionary after sink construction silently changed subsequent HTTP requests, including authorization headers.

Snapshot the header mapping in the constructor. Callers can still intentionally update `sink.headers`; only mutations through the original constructor argument are isolated.

### Test plan

- Added a network-free regression that mutates the source header mapping, patches `urlopen`, and verifies the request keeps the construction-time authorization value.
- Red on `upstream/main`: the request used `Bearer changed` instead of `Bearer original`.
- `tests/sandbox/test_session_sinks.py`: 22 passed.
- Format and lint passed.
- Mypy: clean for 305 source files; Pyright: 0 errors.
- Parallel suite: 8344 passed, 28 skipped.
- Serial suite: 77 passed, 11 skipped, 55 deselected.

### Issue number

No issue.

### Checks

- [x] I've added new tests, if relevant
- [x] I've run `.agents/skills/code-change-verification/scripts/run.sh`
- [x] I've confirmed all verification steps pass
- [x] If using Codex, I've run `/review` before submitting this PR

PR #4398, opened 2026-08-13
fix(sandbox): snapshot per-op audit policies

hsusul:
### Summary

`Instrumentation` copied its configured sink sequence but retained the caller-owned `payload_policy_by_op` dictionary. Clearing or editing that source dictionary after construction silently changed later audit redaction; an exec event configured to hide output could start exposing its stdout.

Snapshot the mapping at construction. Policy objects and the public `instrumentation.payload_policy_by_op` mapping remain mutable, while mutations through the original constructor argument no longer alter instrumentation state.

### Test plan

- Added a regression that clears the source mapping, emits an exec-finish event, and verifies stdout and raw stdout bytes remain redacted.
- Red on `upstream/main`: the callback received `stdout="secret"` and raw bytes.
- `tests/sandbox/test_session_manager.py`: 10 passed.
- Format and lint passed.
- Mypy: clean for 305 source files; Pyright: 0 errors.
- Parallel suite: 8344 passed, 28 skipped.
- Serial suite: 77 passed, 11 skipped, 55 deselected.

### Issue number

No issue.

### Checks

### DIFF
--- a/src/agents/sandbox/session/manager.py
+++ b/src/agents/sandbox/session/manager.py
@@ -25,7 +25,7 @@ class Instrumentation:
     ) -> None:
         self._sinks: list[EventSink] = list(sinks or [])
         self.payload_policy = payload_policy if payload_policy is not None else EventPayloadPolicy()
-        self.payload_policy_by_op = payload_policy_by_op or {}
+        self.payload_policy_by_op = dict(payload_policy_by_op or {})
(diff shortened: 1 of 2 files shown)


seratch:
As I mentioned at https://github.com/openai/openai-agents-python/pull/4397#issuecomment-5287651059, please send a single PR for this type of small fixes in the future.

PR #4399, opened 2026-08-13
feat(sandbox): resolve manifest environment values from the process environment

xumaple:
### Summary

`Environment` can only hold literal values today. `StrEnvValue` is the only concrete `EnvValue`, so anything a sandbox needs in its environment has to be written into the manifest itself, and it travels with the manifest wherever that gets persisted or handed across a boundary.

This adds `OsEnvValue`, which names a variable instead. The backend clients pick it up where they already call `manifest.environment.resolve()`, at create, exec and resume, so the value is read on the machine actually building the sandbox. It also seemed like the obvious second implementation of an interface whose `resolve()` is `async`, which only earns the keyword if resolution can do more than return a field.

Three choices I would rather hear your view on than assume:

- an unset variable resolves to `""` rather than raising, matching shell expansion instead of adding a new failure mode to sandbox creation
- `"os_env"` for the discriminator, which departs from `StrEnvValue`'s class-name-minus-suffix rule but reads better in a serialized manifest. `"os"` is just as easy if consistency matters more, and it is a wire string, so cheaper to settle now than later
- `name` for the field, since no existing model carries a single variable name and `name` is the identifier field used elsewhere in the package

Note: this touches `test_core_discriminator_type_strings_are_stable`, because that table maps every core discriminated class and leaving `OsEnvValue` out would have made it the only one missing. I left the members list in `docs/ref/sandbox/manifest.md` alone under the Documentation Release Timing rule in `AGENTS.md`; let me know if you would like that as a separate PR when it is due.

### Test plan

Five tests in `tests/sandbox/test_manifest.py`: resolving a variable that is set, unset and empty both resolving to `""`, round-tripping through `Manifest` serialization with the discriminator intact and coming back as `OsEnvValue`, and resolving through `Environment.resolve()` alongside a bare literal and an `EnvEntry` wrapper.

`.agents/skills/code-change-verification/scripts/run.sh` passes: format, lint, typecheck and tests all green.

Two notes about my environment rather than the change, in case they show up elsewhere: building `cffi` needed `libffi-dev` installed, and my container sets `no_proxy` to a value `httpx2` cannot parse (`Invalid port: ':1]'`), which takes out around 138 unrelated tests in `test_config.py`, `tracing/`, `mcp/` and `voice/` until those variables are cleared.

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/sandbox/manifest.py
+++ b/src/agents/sandbox/manifest.py
@@ -1,5 +1,6 @@
 import abc
 import inspect
+import os
 from collections.abc import Iterator, Mapping
 from dataclasses import dataclass
 from pathlib import Path, PurePath, PurePosixPath
@@ -145,6 +146,19 @@ class StrEnvValue(EnvValue):
 
     async def resolve(self) -> str:
         return self.value
+
+
+class OsEnvValue(EnvValue):
+    """Reads the value from the environment of the process creating the sandbox.
+
+    An unset variable resolves to the empty string.
+    """
+
+    type: Literal["os_env"] = "os_env"
+    name: str
+
+    async def resolve(self) -> str:
+        return os.environ.get(self.name, "")
 
 
 def _serialize_env_value_with_type(value: EnvValue, serialized: object) -> dict[str, Any]:
--- a/tests/sandbox/test_compatibility_guards.py
+++ b/tests/sandbox/test_compatibility_guards.py
@@ -40,7 +40,7 @@ from agents.sandbox.entries.mounts.patterns import (
     RcloneMountPattern,
     S3FilesMountPattern,
 )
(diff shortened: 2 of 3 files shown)


seratch:
Thanks for putting this together. The underlying need is real: `StrEnvValue` serializes the concrete value, so a late-bound environment reference is useful.

This approach can make sense when another layer already owns and enforces the trust boundary. For the SDK’s built-in generic mechanism, though, the trust boundary needs to live in trusted sandbox runtime configuration rather than in serialized manifest data.

We are pursuing that direction in #4408, where the manifest contains only references and the sandbox client owns the allowed destination/source bindings.

Closing this PR in favor of #4408 for the core SDK implementation. Thanks again for the proposal.

PR #4400, opened 2026-08-13
fix(codex): preserve resume argument ordering

viyatb-oai:
## Summary

Make Codex session resumption consistent with the CLI's positional-argument contract.

- Keep resume-specific image arguments with the `resume` subcommand.
- Pass the thread identifier after the CLI option separator.
- Preserve named threads, image attachments, and prompts read from stdin.
- Add regression coverage for resumed sessions with and without images.

## Test plan

- `make format`
- `make lint`
- `make typecheck`
- `make tests`


cschoenduve-openai (approved):

viyatb-oai:
@codex review

PR #4402, opened 2026-08-13
fix(sessions): prevent duplicate pop_item results under concurrent access in SQLAlchemySession

Rajeev91691:
Combine SELECT statements into one query, apply with_for_update for non-sqlite dialects, and check rowcount on delete to prevent concurrency issues and race conditions. Added unit test test_pop_item_concurrency.

### DIFF
--- a/src/agents/extensions/memory/sqlalchemy_session.py
+++ b/src/agents/extensions/memory/sqlalchemy_session.py
@@ -416,9 +416,8 @@ class SQLAlchemySession(SessionABC):
         async with self._session_factory() as sess:
             async with sess.begin():
                 while True:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. This race was reported in #4205 and resolved by #4212, which shipped in v0.20.0. The released implementation already uses an atomic `DELETE ... RETURNING` claim where available, a locked fallback for non-returning dialects, `BEGIN IMMEDIATE` for the SQLite fallback, and regression coverage for concurrent claims and unknown row counts.

Because #4402 is based on the pre-#4212 implementation and its rowcount-based approach is narrower than the released fix, I am going to close this PR as superseded by #4212. No additional changes are needed here.

PR #4403, opened 2026-08-13
Route max_turns handler output through the final-output session pipeline

### LINKED ISSUE #4393: max_turns handler output bypasses output-guardrail session semantics
### Please read this first

- **Have you read the docs?** Yes, including the output guardrail session behavior and run error handler documentation.
- **Have you searched for related issues?** Yes. I searched open and closed issues and pull requests for `max_turns`, error handlers, output guardrails, streamed session persistence, and handler output. I found the related fixes in #4148, #4160, and #4312, but no report covering the `max_turns` handler terminal path described here.

### Describe the bug

When a `max_turns` error handler returns a final output, that output does not consistently follow the documented output-guardrail session behavior.

The current handler-message behavior is:

| Output guardrail result | `Runner.run()` | `Runner.run_streamed()` |
| --- | --- | --- |
| Guardrail function raises an exception | Handler message is missing | Handler message is retained |
| Guardrail returns a tripwire | Handler message is excluded | Handler message is incorrectly retained |

The expected rule is the same as for other final outputs: a tripwire excludes the rejected candidate message, while an ordinary guardrail exception leaves the verdict unknown, so the completed candidate message remains replayable.

The same mismatch occurs after a completed local tool turn, where retaining or dropping the handler output changes the history replayed by the next run.

The dedicated `max_turns` branches currently implement their own finalization order instead of using the shared final-output session pipeline.

### Debug information

- Agents SDK version: `v0.20.0`; also reproducible on `main` at `1816d2e9`
- Related library versions: no external provider required
- Operating system: macOS
- Model and model provider: none required for the minimal reproduction

No unexpected traceback is required to reproduce the mismatch. The guardrail exception or `OutputGuardrailTripwireTriggered` is expected; the inconsistent session contents are the bug.

### Repro steps

The following self-contained example uses `max_turns=0`, so it requires no model call, API key, or external provider:

```python
import asyncio

from agents import (
    Agent,
    GuardrailFunctionOutput,
    OutputGuardrail,
    OutputGuardrailTripwireTriggered,
    Runner,
    SQLiteSession,
)


async def run_case(*, streamed: bool, tripwire: bool) -> None:
    def check_output(_context, _agent, _output):
        if tripwire:
            return GuardrailFunctionOutput(output_info=None, tripwire_triggered=True)
        raise RuntimeError("output check failed")

    session = SQLiteSession(f"max-turns-{streamed}-{tripwire}", ":memory:")
    agent = Agent(
        name="test",
        output_guardrails=[OutputGuardrail(guardrail_function=check_output)],
    )

    try:
        if streamed:
            result = Runner.run_streamed(
                agent,
                "user message",
                max_turns=0,
                session=session,
                error_handlers={"max_turns": lambda _data: "handler output"},
            )
                pass
        else:
            await Runner.run(
                agent,
                "user message",
                max_turns=0,
                session=session,
                error_handlers={"max_turns": lambda _data: "handler output"},
            )
    except (OutputGuardrailTripwireTriggered, RuntimeError) as error:
        print(type(error).__name__, streamed, tripwire)

    items = await session.get_items()
    item_types = [item.get("type") or item.get("role") for item in items]
    print(streamed, tripwire, "message" in item_types)
    session.close()


async def main() -> None:
    for tripwire in (False, True):
        for streamed in (False, True):
            await run_case(streamed=streamed, tripwire=tripwire)


asyncio.run(main())
```

Current result (`True` means the handler message was retained):

```text
streamed=False, guardrail exception -> False
streamed=True,  guardrail exception -> True
streamed=False, tripwire           -> False
streamed=True,  tripwire           -> True
```

### Expected behavior

`max_turns` handler output should use the same session finalization semantics in `Runner.run()` and `Runner.run_streamed()`:

- A passing output guardrail persists the handler output when `include_in_history=True`.
- A tripwire excludes the handler output.
- An ordinary output guardrail exception preserves the complete handler output for replay.

The regression coverage should also exercise a completed local tool turn to ensure its already-persisted records remain unchanged.

harshasiddartha:
### Summary

When a `max_turns` error handler returns a final output, that output did not follow the documented output-guardrail session behavior, and the two runners disagreed with each other:

| Output guardrail result | `Runner.run()` | `Runner.run_streamed()` |
| --- | --- | --- |
| Guardrail function raises | handler message missing | handler message retained |
| Guardrail returns a tripwire | handler message excluded | handler message retained |

The expected rule is the same one every other final output follows: a tripwire rejects the candidate message, while an ordinary guardrail exception leaves the verdict unknown, so the completed candidate stays replayable. Only two of the four cells were right.

The cause is that both dedicated `max_turns` branches implement their own finalization order rather than using the shared final-output session pipeline — and each got it wrong in the opposite direction:

- **`Runner.run()`** ran `run_output_guardrails` outside any handler and persisted afterwards, so *any* guardrail raise skipped the save.
- **`run_streamed()`** persisted the synthesized item *before* running the guardrails, so nothing they did could exclude it.

### Changes

- **`src/agents/run.py`** — the non-streamed branch now wraps the guardrail call the same way the shared pipeline does: a tripwire re-raises without persisting; any other exception (including `CancelledError`) persists the handler output first. The existing save moved into a small local helper so both the success and error paths use one code path.
- **`src/agents/run_internal/run_loop.py`** — the streamed branch now delegates to `_finalize_streamed_final_output`, which already implements exactly these semantics (including `_retained_items_for_blocked_output` on tripwire), instead of open-coding a different order. The synthesized item is passed as the turn's items and persisted through the same `save_items` callback.

A completed local tool turn is unaffected in both paths — its records are already persisted by the time the handler runs, and the next run still has to see that side effect rather than re-issue it. That's asserted explicitly.

The reproduction from the issue now prints:

```text
streamed=False, guardrail exception -> True
streamed=True,  guardrail exception -> True
streamed=False, tripwire            -> False
streamed=True,  tripwire            -> False
```

### Test plan

Added to `tests/test_max_turns.py`, parametrized over streamed/non-streamed, built on the existing `ScriptedModel` + `SQLiteSession` pattern already used by the `max_turns` session tests (one completed tool turn, then `max_turns` trips):

- `test_max_turns_handler_output_excluded_when_guardrail_trips` — session ends as `["user", "function_call", "function_call_output"]`: the rejected message is gone, the tool turn survives.
- `test_max_turns_handler_output_retained_when_guardrail_errors` — session ends as `["user", "function_call", "function_call_output", "message"]`.

Against the unfixed code these fail on exactly the two cells the issue identified (`tripwire[streamed=True]` and `guardrail-error[streamed=False]`) and pass on the two that were already correct.

Verification:

- `ruff format` / `ruff check src/agents tests/test_max_turns.py` — clean
- `mypy src/agents/run.py src/agents/run_internal/run_loop.py` — clean
- `pytest tests/` — **4238 passed, 7 skipped**

Two pre-existing failures in `tests/test_released_api_contract.py` (`test_repository_release_policy_declares_public_testing_modules`, `test_voice_testing_start_sentinel_has_stable_contract_identity`) reproduce on a clean checkout of this branch's base — they need the voice extras, which my local `uv` (0.9.3) could not install because this repo's `[tool.uv] exclude-newer-package` syntax requires a newer `uv`. Unrelated to this change. `tests/voice`, `tests/sandbox`, `tests/mcp`, `tests/models`, and `tests/extensions` were excluded from the run for the same missing-optional-dependency reason.

### Issue number

Fixes #4393

### Checks

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1294,13 +1294,54 @@ class AgentRunner:
                             context_wrapper,
                             validated_output,
                         )
-                        await run_output_guardrails(
-                            current_agent.output_guardrails + (run_config.output_guardrails or []),
-                            current_agent,
-                            validated_output,
-                            context_wrapper,
-                            output_guardrail_results,
-                        )
+
+                        async def _persist_handler_output(
+                            _synthesized_item: RunItem = synthesized_item,
+                            _include_in_history: bool = include_in_history,
+                            _store_setting: bool | None = store_setting,
+                            _input_items: list[TResponseInputItem] | None = (
+                                session_input_items_for_persistence
+                            ),
+                        ) -> None:
+                            if not (session_persistence_enabled and _include_in_history):
+                                return
+                            handler_input_items_for_save: list[TResponseInputItem] = (
+                                _input_items if _input_items is not None else []
+                            )
+                            # The synthesized item is a fresh one-item list, not the
+                            # cumulative turn item list, so the run state's per-turn
+                            # persisted count must not be applied as a slice offset here.
+                            # Pass the reasoning item id policy explicitly instead, the same
+                            # way `save_resumed_turn_items` does.
+                            await save_result_to_session(
+                                session,
+                                handler_input_items_for_save,
+                                [_synthesized_item],
+                                None,
+                                response_id=None,
+                                reasoning_item_id_policy=resolved_reasoning_item_id_policy,
+                                store=_store_setting,
+                                wrapper=context_wrapper,
+                            )
+
+                        # Same output-guardrail session rules as any other final output: a
+                        # tripwire rejects the candidate message, so it must not be replayable,
+                        # while an ordinary guardrail error leaves the verdict unknown and the
+                        # completed message stays replayable.
+                        try:
+                            await run_output_guardrails(
+                                current_agent.output_guardrails
+                                + (run_config.output_guardrails or []),
+                                current_agent,
+                                validated_output,
+                                context_wrapper,
+                                output_guardrail_results,
+                            )
+                        except OutputGuardrailTripwireTriggered:
+                            raise
+                        except (Exception, asyncio.CancelledError):
+                            await _persist_handler_output()
+                            raise
                         current_step = getattr(run_state, "_current_step", None)
                         approvals_from_state = approvals_from_step(current_step)
                         result = RunResult(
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for the contribution. #4394 is now the preferred implementation path. It covers the same session semantics while also withholding rejected streamed events, preserving zero-turn input and RunState, and handling persistence and redaction failures.

This branch still publishes the streamed handler item before output guardrails complete and lacks those lifecycle boundaries, so bringing both implementations forward would duplicate substantial review and maintenance work. I recommend closing this PR as superseded by #4394.

PR #4408, opened 2026-08-14
feat: add trusted process environment values

seratch:
This pull request adds `ProcessEnvValue` as a late-bound sandbox environment reference that can survive manifest serialization without persisting the concrete secret value.

A manifest only declares which sandbox environment key needs a process-environment reference. It does not declare or serialize permission to read worker process environment variables. Trusted runtime configuration is instead passed to `DockerSandboxClient` or `DaytonaSandboxClient`:

- `allowed_process_environment_keys={"OPENAI_API_KEY"}` grants the common same-name binding.
- `process_environment_bindings={"OPENAI_API_KEY": "PROD_OPENAI_API_KEY"}` grants an explicit destination-to-source rename.

The clients normalize these settings into exact destination/source bindings and apply them during sandbox create or resume. Serialized manifests therefore cannot select arbitrary `os.environ` keys after being loaded by another worker.

Docker and Daytona pass protected values only through provider-native creation environment channels. Unsupported backends and injected sessions reject `ProcessEnvValue` before provider side effects. Resume with current runtime bindings recreates the backend when needed, preserves resumable workspace state, retires replaced resources, and keeps retryable cleanup identities without exposing protected values through command text or errors.

This supersedes #4399.

### DIFF
--- a/src/agents/extensions/sandbox/blaxel/sandbox.py
+++ b/src/agents/extensions/sandbox/blaxel/sandbox.py
@@ -1071,6 +1071,10 @@ class BlaxelSandboxClient(BaseSandboxClient["BlaxelSandboxClientOptions"]):
     ) -> SandboxSession:
         if manifest is None:
             manifest = Manifest(root=DEFAULT_BLAXEL_WORKSPACE_ROOT)
+        manifest._reject_process_environment_values(
+            backend_id="blaxel",
+            supported_alternative="use DockerSandboxClient",
+        )
         self._validate_manifest_for_create(manifest)
 
         timeouts_in = options.timeouts
@@ -1154,6 +1158,10 @@ class BlaxelSandboxClient(BaseSandboxClient["BlaxelSandboxClientOptions"]):
         if not isinstance(state, BlaxelSandboxSessionState):
             raise TypeError("BlaxelSandboxClient.resume expects a BlaxelSandboxSessionState")
         state.assert_path_grants_rebound()
+        state.manifest._reject_process_environment_values(
+            backend_id="blaxel",
+            supported_alternative="use DockerSandboxClient instead",
+        )
         SandboxInstance = _import_blaxel_sdk()
         blaxel_sandbox = None
         reconnected = False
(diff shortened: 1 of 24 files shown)


ErenAta16 (commented):
The description says Daytona carries protected values, and the code says it does not. The code and its tests agree with each other, so I think the description is the thing that needs the edit, but it is worth confirming which way round it should be.

From the PR body:

> Trusted runtime configuration is instead passed to `DockerSandboxClient` or `DaytonaSandboxClient`
> Docker and Daytona pass protected values only through provider-native creation environment channels.

Three things in the diff point the other way.

**`DaytonaSandboxClient` rejects it on both entry points**, with a message naming Docker as the alternative:

```python
# create
manifest._reject_process_environment_values(
    backend_id="daytona",
    supported_alternative=(
        "use DockerSandboxClient for protected process environment transport"
    ),
)

# resume
state.manifest._reject_process_environment_values(
    backend_id="daytona",
    ...
)
```

That is the same shape as the backends the description calls unsupported. `E2BSandboxClient` gets `supported_alternative="use DockerSandboxClient"` in the same position.

**The constructor cannot take the trusted configuration.** `DockerSandboxClient.__init__` has

```python
allowed_process_environment_keys: Iterable[str] = (),
process_environment_bindings: Mapping[str, str] | None = None,
```

and `DaytonaSandboxClient.__init__` has neither, so following the description literally and writing `DaytonaSandboxClient(allowed_process_environment_keys={"OPENAI_API_KEY"})` is a `TypeError` on an unexpected keyword argument rather than a runtime rejection.

**The tests pin the rejection**, in `tests/extensions/sandbox/test_daytona.py`:

```
test_create_rejects_process_environment_before_provider_create
test_resume_rejects_process_environment_before_provider_lookup
    pytest.raises(ValueError, match="daytona does not support ProcessEnvValue")
```

So the implementation is self-consistent: Docker is the only backend that carries these, and Daytona is one of the eight that refuse them.

Worth fixing before merge rather than after, because a PR body of this size usually becomes the changelog entry, and this is the paragraph a reader goes to when deciding which client to configure. If Daytona support is intended and simply not in this PR, saying "Docker today, Daytona to follow" would read the same way without sending anyone at a constructor that will not accept the argument.

I did not review the whole 3k-line diff, only the process-environment boundary, and everything else I looked at on that boundary lined up.

PR #4409, opened 2026-08-14
fix: detach RunState interruption snapshots

seratch:
This pull request fixes a mutable-aliasing bug at the `RunState` interruption boundary.

`RunState.get_interruptions()` already returned a new list, but the `ToolApprovalItem` objects inside that list still shared mutable content with the pending approvals owned by `RunState`. As a result, callers could unintentionally modify authoritative resumable state by changing fields such as `tool_name`, `raw_item.arguments`, or nested metadata on a returned interruption.

Those mutations could alter the approval identity retained for resume, interfere with direct or nested approval routing, or otherwise make subsequent `approve()` and `reject()` behavior depend on caller-side changes to what was expected to be a snapshot.

This change makes `get_interruptions()` return fully detached `ToolApprovalItem` snapshots while preserving the behavior callers rely on:

- The returned list and each returned `ToolApprovalItem` are independent of the pending approval stored by `RunState`.
- Mutable typed Pydantic payloads, mapping-based payloads, and nested metadata are copied without retaining aliases to authoritative state.
- The owning `Agent` is intentionally not copied, preserving its identity.
- Detached snapshots still resolve to the canonical pending approval when passed to `approve()` or `reject()`, including approvals owned by nested agent-tool runs.
- Ambiguous or unsafe approval ownership fails closed before any decision state is mutated.

The supported Pydantic model adapters are derived from `ToolApprovalRawItem`, keeping the copy boundary aligned with the production approval pipeline. This includes function, custom, shell, hosted MCP, and local-shell approval payloads, including `ResponseCustomToolCall` and `ResponseFunctionShellToolCall`.

If a payload cannot be cloned safely, `get_interruptions()` raises an actionable, payload-safe SDK `UserError` before returning any partial result. The implementation does not deep-copy `Agent` instances, callbacks, arbitrary run state, or unsupported `BaseModel` objects, and it does not change the serialized `RunState` schema.

The regression coverage verifies mutation isolation, canonical approve/reject compatibility, nested approval routing, all supported typed approval models, mapping payloads, ambiguous ownership, hostile payload hooks, and all-or-nothing failure behavior. Existing corruption-path tests now mutate authoritative pending state explicitly instead of relying on aliasing through the public snapshot API.

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -12,7 +12,9 @@ import pydantic
 from openai.types.responses import (
     Response,
     ResponseComputerToolCall,
+    ResponseCustomToolCall,
     ResponseFileSearchToolCall,
+    ResponseFunctionShellToolCall,
     ResponseFunctionShellToolCallOutput,
     ResponseFunctionToolCall,
     ResponseFunctionWebSearch,
@@ -547,7 +549,13 @@ class CompactionItem(RunItemBase[TResponseInputItem]):
 
 # Union type for tool approval raw items - supports function tools, hosted tools, shell tools, etc.
 ToolApprovalRawItem: TypeAlias = (
-    ResponseFunctionToolCall | McpCall | McpApprovalRequest | LocalShellCall | dict[str, Any]
+    ResponseFunctionToolCall
+    | ResponseCustomToolCall
+    | ResponseFunctionShellToolCall
+    | McpCall
+    | McpApprovalRequest
+    | LocalShellCall
+    | dict[str, Any]
 )
 
 
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -6,12 +6,13 @@ import asyncio
 import copy
 import dataclasses
 import json
+import math
 import threading
 from collections import deque
 from collections.abc import Callable, Collection, Iterator, Mapping, Sequence
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Any, Generic, Literal, cast
+from typing import TYPE_CHECKING, Annotated, Any, Generic, Literal, cast, get_args
 from uuid import uuid4
 
 from openai.types.responses import (
@@ -36,7 +37,7 @@ from openai.types.responses.response_output_item import (
     Program,
     ProgramOutput,
 )
-from pydantic import StringConstraints, TypeAdapter, ValidationError
+from pydantic import BaseModel, StringConstraints, TypeAdapter, ValidationError
 from typing_extensions import TypedDict, TypeVar
 
 from ._tool_identity import (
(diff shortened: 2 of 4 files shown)


FU-max-boop:
I found a separate snapshot-isolation gap adjacent to this change.

This PR detaches the `ToolApprovalItem` values returned by `RunState.get_interruptions()`. However, both `RunResult.to_state()` paths still construct every new state with the same `context_wrapper`.

As a result, two states created from one interrupted result share the SDK-owned approval and invocation ledger:

```python
decided = result.to_state()
untouched = result.to_state()
before = untouched.to_json()["context"]["approvals"]

decided.approve(decided.get_interruptions()[0])

assert untouched.to_json()["context"]["approvals"] == before
# Fails: the decision made through decided appears in untouched.
```

On released `openai-agents==0.20.0`, resuming `untouched` then executes the approval-gated tool once even though that state never received an approval. The same alias remains at this PR head because neither `to_state()` path detaches `result.context_wrapper`.

This is distinct from detaching the public interruption objects: the leaked mutable state is the decision ledger owned by the wrapper. I reproduced it without an API key across non-streamed and streamed results. I am preparing a focused regression test and content-addressed AgentRunProof record; I can send a follow-up fix unless you would prefer this folded into #4409.

seratch:
Thanks for pointing it out. I'll resolve it in a follow-up PR before making the upcoming release.

FU-max-boop:
Thanks — I’ll avoid opening a competing PR.

I have now published a pinned AgentRunProof counterexample for the released `openai-agents==0.20.0` behavior:

- [content-addressed bundle](https://github.com/FU-max-boop/agentrunproof/blob/e28ef9c8916d02b6d50f84a995401ab337ee0dbb/evidence/current/runstate-sibling-approval-isolation/v1/bundle.json)
- [certificate](https://github.com/FU-max-boop/agentrunproof/blob/e28ef9c8916d02b6d50f84a995401ab337ee0dbb/evidence/current/runstate-sibling-approval-isolation/v1/certificate.json)
- [immutable AgentRunProof v0.1.1 release](https://github.com/FU-max-boop/agentrunproof/releases/tag/v0.1.1) / [`pip install agentrunproof==0.1.1`](https://pypi.org/project/agentrunproof/0.1.1/)

The no-API-key run records both non-streamed and streamed variants. In each, deciding one sibling mutates the untouched sibling, and resuming the untouched state executes the protected tool once. The bundle is tied to the tested wheel/source, reproduced in Linux CI, and can be checked offline with the released `agentrunproof check-current-bundle` command.

I’m keeping a reference fix and a broader regression matrix ready (approve/reject, streamed/non-streamed pause and resume, nested `Agent.as_tool()`, and legacy approval migration). When the follow-up PR appears, I’ll run that matrix against it and report a concise before/after result, including any uncovered edge case.

PR #4410, opened 2026-08-14
fix(voice): reject non-finite audio samples

wiliyam:
### Summary

This pull request fixes `AudioInput` conversion by rejecting `NaN` and infinite `float32` samples before converting buffers to WAV or base64 PCM. Previously those values reached NumPy's integer cast, emitted a runtime warning for `NaN`, and silently produced invalid PCM samples.

### Test plan

- `uv run --no-sync pytest -q tests/voice/test_input.py` (30 passed)
- parallel suite: 8,380 passed, 39 skipped
- serial suite: 77 passed, 5 skipped
- Ruff lint/format and focused mypy check passed
- Full mypy reported three pre-existing errors in unrelated files; pyright could not start because this ARM host lacks `libatomic.so.1`

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/voice/input.py
+++ b/src/agents/voice/input.py
@@ -28,6 +28,9 @@ def _buffer_to_audio_file(
     if buffer.dtype not in (np.int16, np.float32):
         raise UserError("Buffer must be a numpy array of int16 or float32")
 
+    if buffer.dtype == np.float32 and not np.isfinite(buffer).all():
+        raise UserError("Buffer must contain only finite values")
+
     if channels <= 0:
         raise UserError("Channels must be greater than zero")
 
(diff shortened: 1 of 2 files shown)


seratch:
I confirmed that `NaN` can trigger a NumPy `RuntimeWarning`, but this PR does not demonstrate a realistic supported source of non-finite audio samples or a meaningful downstream failure. On the current base, conversion still produces a structurally valid WAV/PCM buffer: infinities are saturated by clipping, while the observed `NaN` value is converted to an integer sample. Applications producing float32 audio can validate or normalize their application-owned DSP output before constructing `AudioInput`.

I am going to close this PR for now. We can revisit the problem if there is a real capture or DSP reproduction showing a user-visible transcription or operational failure that caller-side validation cannot reasonably address.

PR #4414, opened 2026-08-14
fix: resume recursive agent tool approvals

seratch:
This pull request fixes approval resume through recursively nested `Agent.as_tool` calls.

When a protected tool was nested behind two or more agent-tool edges, approving the flattened interruption from `RunResult.to_state()` could leave the same interruption pending or allow the outer run to continue without executing the protected tool. The resume path now resolves the deepest nested `RunState` that owns the approval and checks that owner’s decision ledger before deciding whether to rerun the pending agent tool.

The change preserves existing checkpoint isolation and serialized `RunState` behavior. Regression coverage verifies two- and three-edge nesting for both `Runner.run` and `Runner.run_streamed`, including exactly-once protected-tool execution.

ref: https://github.com/openai/openai-agents-python/pull/4413#issuecomment-5293064242

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -746,21 +746,31 @@ class Agent(AgentBase, Generic[TContext]):
                 pending_run_result: RunResult | RunResultStreaming,
             ) -> Literal["approved", "pending", "rejected"]:
                 interruptions = pending_run_result.interruptions
-                nested_decision_context = pending_run_result.to_state()._context
+                nested_state = pending_run_result.to_state()
                 has_pending = False
                 has_decision = False
                 for interruption in interruptions:
-                    call_id = get_tool_approval_item_call_id(interruption)
+                    find_nested_owner = getattr(
+                        nested_state,
+                        "_find_nested_approval_state",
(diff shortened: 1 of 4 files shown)


FU-max-boop:
I independently validated the exact head `1725a8989adcba536641cea1ef56a02d310c534e` with [AgentRunProof](https://github.com/FU-max-boop/agentrunproof) plus the upstream test suite.

The pre-fix head `9dc7da9` reproducibly failed the serialized recursive-approval scenario: after a `RunState.to_json()` / `RunState.from_json()` boundary, both approval and rejection remained interrupted. On this head, all of the following pass in both `Runner.run` and `Runner.run_streamed`:

- sibling `RunState` approval isolation;
- recursive live approval routing;
- recursive approval routing after JSON restoration;
- approve and reject decisions, with 2 and 3 nested `Agent.as_tool()` edges;
- decisions applied before serialization and after restoration.

The focused upstream suites are also green locally (`546 passed`), and the full GitHub Actions matrix is green. I found no remaining issue in this contract.

PR #4417, opened 2026-08-14
fix(sandbox): honor view_image extra path grants

sylvesterkaczmarek:
### Summary

`Manifest.extra_path_grants` are documented to apply to SDK file APIs, but `view_image` resolves its input with the workspace-only path resolver. As a result, an absolute image path explicitly granted through `SandboxPathGrant` is rejected before the sandbox session can read it.

This change switches `view_image` to the existing grant-aware `normalize_path()` policy. Workspace-relative behavior is unchanged, ungranted absolute paths are still rejected, and read-only grants remain usable because the tool only reads.

For user-facing errors, workspace paths keep their existing relative display form; granted paths outside the workspace use the normalized sandbox path.

### Test plan

Added focused regression coverage for:

- reading a PNG through a read-only absolute extra path grant
- continuing to reject an ungranted absolute path without sending a read

Repository-wide verification is left to CI because this environment has GitHub connector access but no materialized checkout or dependencies.

### Issue number

None. Found while auditing sandbox path-grant consistency.

### Checks


seratch (changes requested):
Thanks for the focused fix. The underlying gap is real: `view_image` currently rejects an explicitly granted absolute read path before the sandbox session can read it, and `normalize_path()` is the right existing policy boundary.

Please make a few updates before merge:

- Resolve the failing lint check and any failing tests.
- Pass the raw `args.path` string into `normalize_path()` and `relative_path()` instead of converting it through `Path`.
- Update the `ViewImageArgs.path` description and tool description so they accurately mention explicitly granted sandbox paths, not only workspace paths.

The grant and ungranted-path regression tests otherwise cover the intended behavior well.

seratch (approved):

PR #4419, opened 2026-08-14
fix(realtime): handle non-finite audio rates

hsusul:
This pull request fixes realtime PCM audio-format normalization for non-finite numeric rates.

`NaN` and positive or negative infinity previously raised while being coerced with `int()`. They now follow the existing unsupported-rate path and safely fall back to the supported 24 kHz PCM format, with focused regression coverage.


seratch (approved):
While I still think this is an unrealistic edge case, improving the part makes sense.

PR #4420, opened 2026-08-14
fix(sandbox): preserve zero-length archive reads

hsusul:
This pull request fixes zero-length `IteratorIO.readinto()` calls so they return immediately without advancing or blocking on the underlying archive iterator.

It also keeps Docker archive response ownership correct when such a stream is closed before iteration starts by sharing one idempotent response finalizer between iterator exhaustion and explicit close. Regression coverage verifies both the zero-length read contract and HTTP response cleanup.

### DIFF
--- a/src/agents/sandbox/sandboxes/docker.py
+++ b/src/agents/sandbox/sandboxes/docker.py
@@ -12,7 +12,7 @@ import threading
 import time
 import uuid
 from collections import deque
-from collections.abc import Iterable, Iterator
+from collections.abc import Callable, Iterable, Iterator
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for the careful patch and regression tests. I could confirm that a zero-length `IteratorIO.readinto()` call advances the underlying iterator today, but I could not find a supported runtime path where that happens: Docker `persist_workspace()` immediately normalizes the private stream through `tarfile`, and that flow reads through `read()` rather than zero-length `readinto()`.

Because the PR currently demonstrates only a synthetic call to an internal helper, it does not establish a user-visible failure, hang, resource leak, or other practical consequence that would justify adding and maintaining the new response-finalization path. I’m going to close this for now. If there is a real supported consumer or runtime path that issues the zero-length read, please share a minimal reproduction and the observable impact, and we can reconsider the narrow fix.

PR #4423, opened 2026-08-14
fix(sandbox): preserve falsy exception causes

wiliyam:
### Summary

This pull request fixes sandbox retry exception-chain traversal when an explicit cause implements falsey truthiness. The previous `cause or context` selection skipped that cause and could classify retries using an unrelated context exception; selecting by identity preserves Python's cause precedence without invoking `__bool__`.

### Test plan

- `uv run --no-dev --with pytest --with pytest-asyncio pytest -q tests/sandbox/test_retry.py` (8 passed)
- `uv run --no-dev --with ruff==0.9.2 ruff check src/agents/sandbox/util/retry.py tests/sandbox/test_retry.py`
- `uv run --no-dev --with ruff==0.9.2 ruff format --check src/agents/sandbox/util/retry.py tests/sandbox/test_retry.py`
- `git diff --check`

The full verification script could not start because this environment lacks a C compiler required to build the dev dependency `evdev`; the focused runtime tests and checks above passed.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/sandbox/util/retry.py
+++ b/src/agents/sandbox/util/retry.py
@@ -32,9 +32,10 @@ def iter_exception_chain(exc: BaseException) -> Iterable[BaseException]:
     while current is not None and id(current) not in seen:
         yield current
         seen.add(id(current))
+        cause = getattr(current, "__cause__", None)
         current = cast(
             BaseException | None,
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for suggesting this, but I don't think this is a realistic scenario.

PR #4425, opened 2026-08-14
fix(retry): bound a provider retry-after by the configured max_delay

chinmayv095:
### Summary

`ModelRetryBackoffSettings.max_delay` documents itself as "Maximum delay in seconds between retry attempts", but a provider-supplied retry-after bypasses it entirely, so it is not actually a maximum.

In `_evaluate_retry`, when a retry policy returns `retry=True` without naming a delay, the runner substitutes the provider's value:

```python
delay=(
    decision.delay
    if decision.delay is not None
    else (
        normalized.retry_after          # straight from the response header, unbounded
        if normalized.retry_after is not None
        else _default_retry_delay(attempt, retry_backoff)   # this one is capped
    )
),
```

`normalized.retry_after` comes from `get_retry_after()`, which reads `retry-after-ms` / `retry-after` off the error chain. The neighbouring `_default_retry_delay` clamps with `min(..., max_delay)`; the retry-after branch does not. A caller that configured a five-second ceiling and hit a `Retry-After: 3600` blocks for an hour:

```python
ModelRetrySettings(
    max_retries=1,
    backoff=ModelRetryBackoffSettings(initial_delay=1.0, max_delay=5.0, jitter=False),
    policy=retry_policies.network_error(),
)
# server answers with Retry-After: 3600  ->  runner sleeps 3600s
```

Note the policy is `network_error()`. The application never asked for retry-after semantics; the SDK reached for the provider's number on its own and let it override the bound the application did set.

### What this changes

The substituted provider value is bounded by the caller's configured ceiling. Three cases are deliberately left alone, and each has a test:

- **No `max_delay` configured — unchanged.** The provider's retry-after is still honored in full. The default ceiling is `DEFAULT_MAX_DELAY_SECONDS = 2.0`, and it exists to bound the SDK's own exponential growth. Substituting it here would shrink an ordinary `Retry-After: 60` to two seconds and retry into a rate limit far sooner than the server asked — a worse failure than the one being fixed. So the cap keys on an *explicit* `max_delay`, via `_explicit_max_delay()` returning `None` rather than the default.
- **An explicit `RetryDecision.delay` — unchanged.** That number is the application's own, not the provider's.
- **`retry_policies.retry_after()` — unchanged.** Choosing that policy is an explicit request for server-driven delays. It also could not participate: `RetryPolicyContext` carries `max_retries` but no backoff settings, so the policy cannot see a ceiling. Threading settings through the policy context would widen the public context type for a case that reads as intended behavior, so it is left as-is rather than patched around.

Callers who set no ceiling see byte-identical behavior; the change is scoped to callers who stated a bound and were not getting it.

### Test plan

Four tests in `tests/models/test_model_retry.py`, sharing one `_run_once_with_retry_after` helper:

- `test_provider_retry_after_is_capped_by_configured_max_delay` — `retry_after=3600.0` against `max_delay=5.0` sleeps `5.0`.
- `test_provider_retry_after_below_max_delay_is_unchanged` — `1.5` against a `5.0` ceiling stays `1.5`, so the cap cannot pass by clamping everything to the ceiling.
- `test_provider_retry_after_is_honored_without_a_configured_max_delay` — `45.0` with no `max_delay` still sleeps `45.0`.
- `test_policy_supplied_delay_is_not_capped_by_max_delay` — an explicit `RetryDecision(delay=30.0)` survives a `5.0` ceiling.

Fail-first on `main` with the fix reverted: **1 failed, 3 passed** — only the first assertion is about the bug, and the other three are controls that must hold in both states. They do.

`tests/models/test_model_retry.py` goes 82 → 86 passed, +4 = exactly the new tests, no pre-existing test changed. `test_get_response_with_retry_prefers_retry_after_over_backoff` still passes and still asserts `[1.75]`: it sets no `max_delay`, so it lands on the untouched path.

Full stack via `.agents/skills/code-change-verification/scripts/run.sh`: `make format`, `make lint`, `make typecheck`, `make tests` all pass.

### Issue number

None — found while reading the backoff path.

### Checks

---

*Disclosure: this change was prepared with AI assistance. The bug, the fix, and the tests were verified locally as described above.*

### DIFF
--- a/src/agents/run_internal/model_retry.py
+++ b/src/agents/run_internal/model_retry.py
@@ -166,6 +166,16 @@ async def _call_retry_policy(
     return _coerce_retry_decision(decision)
 
 
+def _explicit_max_delay(backoff: ModelRetryBackoffInput | None) -> float | None:
+    """Return the caller's configured retry ceiling, or `None` when they did not set one.
+
+    The default ceiling exists to bound the SDK's own exponential growth. It is deliberately not
+    substituted here, so an unconfigured caller keeps honoring a provider's retry-after in full.
+    """
+    settings = _coerce_backoff_settings(backoff)
+    return settings.max_delay if settings is not None else None
+
+
 def _default_retry_delay(
     attempt: int,
     backoff: ModelRetryBackoffInput | None,
@@ -196,6 +206,17 @@ def _default_retry_delay(
     if not use_jitter:
         return base
     return min(max(base * (0.875 + random.random() * 0.25), 0.0), max_delay)
+
+
+def _capped_provider_delay(
+    retry_after: float,
+    backoff: ModelRetryBackoffInput | None,
+) -> float:
+    """Bound a provider-supplied retry-after by the caller's configured ceiling."""
+    max_delay = _explicit_max_delay(backoff)
+    if max_delay is None:
+        return retry_after
+    return min(retry_after, max_delay)
 
 
 async def _sleep_for_retry(delay: float) -> None:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the careful investigation and tests. I do not think we should merge this change because the current behavior is an explicit released contract: `backoff.max_delay` caps only SDK-computed backoff, not provider retry-after hints or explicit policy delays.

This patch would make provider guidance depend on the policy’s return representation: a bare `True` would cap the retry-after value, while `RetryDecision(delay=...)` would preserve it. It could also retry before a server’s stated rate-limit window expires.

Applications that need a global ceiling can implement that policy explicitly by inspecting `context.normalized.retry_after` and returning a bounded `RetryDecision(delay=...)`. If there is demand for a separate global retry-delay cap, please open an issue with a concrete supported scenario so we can design that API independently.

PR #4426, opened 2026-08-14
fix: strip defaults from strict JSON schemas

sylvesterkaczmarek:
Fixes #4390.

Strict schemas already require every declared property, so Python-level defaults are not used by Structured Outputs. Remove `default` unconditionally during strict-schema normalization, including non-null values that the API rejects.

Adds regression coverage for string, integer, boolean, and null defaults.


seratch:
Given the evidence that only Azure fails with the current schema, we'd like to hold off making this change.

PR #4427, opened 2026-08-14
feat: add run-scoped sandbox working directories

### LINKED ISSUE #4401: [Sandbox] Add a run-scoped cwd across capabilities when reusing a live session
### Describe the feature

I would like to bind each `SandboxAgent` run to a logical working directory inside a developer-owned live sandbox, with consistent relative-path behavior across sandbox capabilities.

The SDK currently supports sharing one live sandbox session, but every run uses the session-wide `Manifest.root` for filesystem path resolution:

- `exec_command(workdir=...)` changes the directory for that shell command only.
- `view_image("plot.png")` resolves from the session workspace root.
- `apply_patch` paths are workspace-root-relative, not relative to the shell `workdir`.
- `run_as` changes the OS identity and permissions, but not path resolution.

This behavior is documented, so I am filing this as a feature request rather than a bug.

Relevant current behavior:

- [`exec_command` implements `workdir` as a per-command `cd`](https://github.com/openai/openai-agents-python/blob/1816d2e9f93a0d7265094ec3c9c6c312ae5c3aff/src/agents/sandbox/capabilities/tools/shell_tool.py#L70-L77)
- [`view_image` uses the session-wide workspace path policy](https://github.com/openai/openai-agents-python/blob/1816d2e9f93a0d7265094ec3c9c6c312ae5c3aff/src/agents/sandbox/capabilities/tools/view_image.py#L104-L113)
- [`SandboxRunConfig` has no run-level cwd or workspace view](https://github.com/openai/openai-agents-python/blob/1816d2e9f93a0d7265094ec3c9c6c312ae5c3aff/src/agents/run_config.py#L189-L219)
- The [Sandbox Agents guide](https://developers.openai.com/api/docs/guides/agents/sandboxes) notes that `apply_patch` paths are relative to the workspace root rather than shell `workdir`.

#### Production use case

Our application runs several trusted document-research agents concurrently. They share the same dependencies and do not require separate containers or security-grade isolation. Each agent only needs a reliable task-local directory:

```text
/workspace/tasks/task-a
/workspace/tasks/task-b
```

We originally reused one Modal sandbox and assigned each child a task directory. This produced two related problems:

1. Concurrent children sometimes chose identical workspace-root-relative filenames and overwrote or read each other's files.
2. After forcing each child's shell commands into its task directory, filesystem capabilities still used the session root.

For example:

```text
exec_command(
    cmd="render input.pdf --output page.png",
    workdir="tasks/task-a",
)

view_image(path="page.png")
# Looks for /workspace/page.png rather than
# /workspace/tasks/task-a/page.png.

apply_patch(path="notes.md", ...)
# Edits /workspace/notes.md rather than
# /workspace/tasks/task-a/notes.md.
```

We worked around this with capability-specific path-prefixing adapters, task-local temporary directories, and prompt instructions. That remained brittle because every path-bearing capability or custom tool had to implement the convention independently.

We eventually moved to one fresh Modal sandbox per child. That is reliable, but concurrent fan-out now multiplies sandbox startup overhead and active sandbox usage, and we sometimes approach our provider concurrency limits.

#### Desired behavior

The exact API is not important, but an illustrative shape might be:

```python
run_a = RunConfig(
    sandbox=SandboxRunConfig(
        session=shared_sandbox,
        cwd="tasks/task-a",
    )
)

run_b = RunConfig(
    sandbox=SandboxRunConfig(
        session=shared_sandbox,
        cwd="tasks/task-b",
    )
)

await asyncio.gather(
    Runner.run(agent_a, task_a, run_config=run_a),
    Runner.run(agent_b, task_b, run_config=run_b),
)
```

For `run_a`:

- shell commands without an explicit `workdir` start under `tasks/task-a`;
- a relative shell `workdir` is interpreted beneath `tasks/task-a`;
- `view_image("page.png")` reads `tasks/task-a/page.png`;
- an `apply_patch` operation on `notes.md` edits `tasks/task-a/notes.md`.

The scope should be bound to the run or its prepared capabilities, rather than implemented by mutating session-global cwd or path-policy state. That matters when multiple runs share the session concurrently.

Other desirable properties:

- Omitting `cwd` preserves current behavior.
- `Manifest.root`, session lifecycle, snapshots, and direct `BaseSandboxSession` operations remain session-scoped.
- `run_as` and filesystem permissions remain orthogonal and continue to compose with the path scope.
- Built-in workspace-backed capabilities have one consistent relative-path base.
- The design can support distinct scopes for concurrent runs sharing one developer-owned session.

#### Minimal model-free reproduction

<details>
<summary>Reproduction using UnixLocalSandboxClient</summary>

```python
import asyncio

from agents.sandbox import Manifest
from agents.sandbox.capabilities.tools import (
    ExecCommandArgs,
    ExecCommandTool,
    ViewImageArgs,
    ViewImageTool,
)
from agents.sandbox.sandboxes.unix_local import UnixLocalSandboxClient

PNG = (
    "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0l"
    "EQVR42mP8/x8AAwMCAO+a84QAAAAASUVORK5CYII="
)


async def main() -> None:
    client = UnixLocalSandboxClient()
    session = await client.create(manifest=Manifest())

    try:
        async with session:
            await session.mkdir("tasks/a", parents=True)

            shell = ExecCommandTool(session=session)
            view_image = ViewImageTool(session=session)

            await shell.run(
                ExecCommandArgs(
                    cmd=f"printf '%s' '{PNG}' | base64 -d > plot.png",
                    workdir="tasks/a",
                )
            )

            print(await view_image.run(ViewImageArgs(path="plot.png")))
            print(
                type(
                    await view_image.run(
                        ViewImageArgs(path="tasks/a/plot.png")
                    )
                ).__name__
            )
    finally:
        await client.delete(session)


asyncio.run(main())
```

Output:

```text
image path `plot.png` was not found
ToolOutputImage
```

The example uses the Unix-local provider, but our Modal integration encounters the same mismatch.

</details>

#### Non-goals

- This does not need to be a chroot or security boundary.
- It does not replace separate sandboxes for untrusted agents, different images, or genuine compute isolation.
- It does not require per-run mounts, users, groups, or environments.
- It does not address child-agent orchestration or result delivery.
- Deliberate access through absolute paths or `..` could remain a separate policy decision; the immediate need is consistent handling of relative paths.

seratch:
This pull request supersedes #4421 and adds optional run-scoped sandbox working directories through `SandboxRunConfig.cwd`.

Relative paths for built-in `exec_command`, `view_image`, and `apply_patch` tools now resolve from an immutable per-run base, allowing trusted runs to share one developer-owned sandbox session without mutating session-global working-directory state. Raw model-provided path strings are normalized through the sandbox POSIX path layer before applying the run cwd, so backslash-containing `exec_command` workdirs and `view_image` paths behave consistently on Windows.

The change preserves existing behavior when `cwd` is omitted, keeps Memory and Skills session-owned with stable model-facing paths, and rebinds approved custom tool calls to the current tool inventory during resume. A configured working directory changes relative-path resolution only; it is not filesystem or compute isolation. Applications requiring isolation should continue using separate sandbox sessions.

This pull request resolves #4401.

### DIFF
--- a/examples/sandbox/README.md
+++ b/examples/sandbox/README.md
@@ -19,6 +19,7 @@ Most examples call a model through `Runner`, so set `OPENAI_API_KEY` in the repo
 | [`memory.py`](./memory.py) | `uv run python examples/sandbox/memory.py` | Runs one sandbox agent twice across a snapshot resume so it can read and write its own memory. |
 | [`memory_s3.py`](./memory_s3.py) | `source ~/.s3.env && uv run python examples/sandbox/memory_s3.py` | Runs sandbox memory across two fresh Docker sandboxes with S3-backed memory storage. |
 | [`memory_multi_agent_multiturn.py`](./memory_multi_agent_multiturn.py) | `uv run python examples/sandbox/memory_multi_agent_multiturn.py` | Shows separate memory layouts for two agents sharing one sandbox workspace. |
+| [`shared_session_workdirs.py`](./shared_session_workdirs.py) | `uv run python examples/sandbox/shared_session_workdirs.py` | Shares one live sandbox between trusted agents while Shell, `view_image`, and `apply_patch` resolve relative paths from each run's `cwd`. This is not confinement; use separate sessions for untrusted agents or compute isolation. |
 | [`unix_local_pty.py`](./unix_local_pty.py) | `uv run python examples/sandbox/unix_local_pty.py` | Exercises an interactive pseudo-terminal in a Unix-local sandbox. |
 | [`unix_local_runner.py`](./unix_local_runner.py) | `uv run python examples/sandbox/unix_local_runner.py` | Runs against the Unix-local sandbox backend directly. |
 
--- a/examples/sandbox/shared_session_workdirs.py
+++ b/examples/sandbox/shared_session_workdirs.py
@@ -1,0 +1,289 @@
+"""Run two trusted agents in separate working directories of one live sandbox.
+
+Run-scoped working directories make relative paths consistent; they are not confinement.
+Use separate sandbox sessions for untrusted agents or workloads that need compute isolation.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import base64
+import json
+from pathlib import Path
+
+from agents import ModelSettings, Runner, RunResult, ToolOutputImage
+from agents.items import ToolCallItem, ToolCallOutputItem
+from agents.run import RunConfig
+from agents.sandbox import Manifest, SandboxAgent, SandboxRunConfig
+from agents.sandbox.capabilities import Filesystem, Shell
+from agents.sandbox.entries import BaseEntry, Dir, File
+from agents.sandbox.sandboxes.unix_local import UnixLocalSandboxClient
+from agents.sandbox.session import BaseSandboxSession
+
+
+async def main(*, model: str) -> None:
+    client = UnixLocalSandboxClient()
+    agent_a = _build_agent(name="Task A worker", model=model)
+    agent_b = _build_agent(name="Task B worker", model=model)
+    shared_sandbox = await client.create(manifest=_build_manifest())
+
+    try:
+        # The session is shared, but each run gets a different base for relative paths.
+        run_a_config = RunConfig(
+            sandbox=SandboxRunConfig(
+                session=shared_sandbox,
+                cwd="tasks/task-a",
+            ),
+            workflow_name="Shared sandbox task A",
+        )
+        run_b_config = RunConfig(
+            sandbox=SandboxRunConfig(
+                session=shared_sandbox,
+                cwd="tasks/task-b",
+            ),
+            workflow_name="Shared sandbox task B",
+        )
+
+        async with shared_sandbox:
+            run_tasks = [
+                asyncio.create_task(
+                    Runner.run(
+                        agent_a,
+                        _task_prompt("task-a"),
+                        run_config=run_a_config,
+                        max_turns=10,
+                    )
+                ),
+                asyncio.create_task(
+                    Runner.run(
+                        agent_b,
+                        _task_prompt("task-b"),
+                        run_config=run_b_config,
+                        max_turns=10,
+                    )
+                ),
+            ]
+            result_a, result_b = await _run_concurrently(run_tasks)
+
+            # These checks make the demo self-verifying; applications do not need them.
+            await _verify_task(shared_sandbox, task_name="task-a", result=result_a)
+            await _verify_task(shared_sandbox, task_name="task-b", result=result_b)
+    finally:
+        await client.delete(shared_sandbox)
+
+    print("\nVerified: distinct run-scoped cwd values worked without session-global cwd mutation.")
+
+
+# Demo setup. Applications can create task directories and inputs however they prefer.
+
+DEFAULT_MODEL = "gpt-5.6-sol"
+
+# The two task directories intentionally contain the same relative filenames.
+TASKS = {
+    "task-a": (
+        "red",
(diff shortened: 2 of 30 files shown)


Filimoa:
Thanks for carrying this forward and preserving the contribution credit. While doing an additional final regression pass on the superseding branch, I found two approval-resume gaps that were already present in my original #4421 implementation and remain in the current #4427 head (4d4ce4aa):

- An approved public `ApplyPatchTool` custom-form call fails before invoking its editor.
- An unchanged pending `CustomTool` call fails when the tool is no longer available instead of preserving the interruption.

Both regression tests pass on `main` and fail on the current #4427 head.

<details>
<summary>Exact regression tests</summary>

~~~python
import json
from typing import Any

import pytest
from openai.types.responses import ResponseCustomToolCall

from agents import Agent, ApplyPatchTool, CustomTool, Runner
from agents.editor import ApplyPatchOperation, ApplyPatchResult
from tests.test_responses import get_text_message


@pytest.mark.asyncio
async def test_approved_apply_patch_resume_uses_public_tool() -> None:
    class RecordingEditor:
        def __init__(self) -> None:
            self.paths: list[str] = []

        def update_file(
            self, operation: ApplyPatchOperation
        ) -> ApplyPatchResult:
            self.paths.append(operation.path)
            return ApplyPatchResult(status="completed", output="updated")

    editor = RecordingEditor()
    tool = ApplyPatchTool(editor=editor, needs_approval=True)
    call = ResponseCustomToolCall(
        type="custom_tool_call",
        name="apply_patch",
        call_id="patch_0",
        input=json.dumps(
            {
                "type": "update_file",
                "path": "safe.txt",
                "diff": "-old\n+new\n",
            }
        ),
    )
    model = ScriptedModel(
        steps=[[call], [get_text_message("done")]]
    )
    agent = Agent(name="agent", model=model, tools=[tool])

    first = await Runner.run(agent, "update the file")
    state = first.to_state()
    state.approve(state.get_interruptions()[0])

    result = await Runner.run(agent, state)

    assert result.final_output == "done"
    assert editor.paths == ["safe.txt"]


@pytest.mark.asyncio
async def test_pending_custom_tool_survives_missing_current_tool() -> None:
    async def invoke(_context: Any, value: str) -> str:
        raise AssertionError("The pending tool must not execute")

    tool = CustomTool(
        name="raw_editor",
        description="Edit raw text.",
        on_invoke_tool=invoke,
        format={"type": "text"},
        needs_approval=True,
    )
    call = ResponseCustomToolCall(
        type="custom_tool_call",
        name="raw_editor",
        call_id="call_0",
        input="safe",
    )
    agent = Agent(
        name="agent",
        model=ScriptedModel(steps=[[call]]),
        tools=[tool],
    )

    first = await Runner.run(agent, "edit")
    state = first.to_state()
    agent.tools = []

    resumed = await Runner.run(agent, state)

    assert len(resumed.interruptions) == 1
~~~

</details>

On #4427, the first raises `ModelBehaviorError: Model produced apply_patch call without an apply_patch tool`; the second raises `ModelBehaviorError: Tool raw_editor not found in agent agent`. It looks like unresolved custom calls are classified against the current inventory before their persisted approval decision and tool kind are settled.

seratch:
@Filimoa Thanks for verifying the behavior. The latest commit should resolve it. If possible, checking the behavior again would be greatly appreciated.

Filimoa:
@seratch Confirmed both approval-resume fixes on `7820c060`: both exact repros now pass, along with the full approval regression file (126 tests) and JSON `RunState` round trips.

One small path-reporting mismatch remains: with cwd `tasks/a`, `view_image` given `\workspace\root.txt` accesses `/workspace/root.txt`, but an error reports `../../root.txt`. This only affects the displayed path, not which file is accessed. It looks like access uses the normalized path while display formatting still uses the raw input. Could we normalize it once and use the same value for both?

Otherwise, everything I checked looks good.

seratch:
Thanks for quickly checking it again! The latest commit should resolve the remaining issue.

Filimoa:
🙌

PR #4430, opened 2026-08-15
fix: honour a per-call approval decision made after a sticky one

### LINKED ISSUE #4429: Rejecting a single call after always_approve is ignored, so the tool still runs
Sticky and per-call approval decisions are stored in the same two fields but in different shapes. `always_approve` and `always_reject` write a bool into `_ApprovalRecord.approved` / `.rejected`, while a per-call decision appends the call ID to a list in those same fields. Once a sticky decision has been made, a later per-call decision on the same tool is either ignored on read or dropped on write.

```python
from agents import Agent
from agents.items import ToolApprovalItem

agent = Agent(name="a")

def item(call_id):
    return ToolApprovalItem(
        agent=agent,
        raw_item={"type": "function_call", "name": "do_thing", "call_id": call_id, "arguments": "{}"},
        tool_name="do_thing",
    )

ctx = RunContextWrapper(context=None)
ctx.approve_tool(item("c1"), always_approve=True)
ctx.reject_tool(item("c2"), rejection_message="denied by user")
print(ctx.get_approval_status("do_thing", "c2"))
```

```
True
```

The record holds both decisions at once:

```
approved=True  rejected=['c2']  rejection_messages={'c2': 'denied by user'}
```

`get_approval_status` returns at `if approval_entry.approved is True: return True` before it reaches `rejected_ids`, so the rejection and its message are written to state that nothing reads. `tool_execution.py` treats `True` as approved and only consults the `on_approval` hook when the status is `None`, so nothing re-asks and the tool runs.

The other direction fails earlier. After `always_reject`, `approved` holds the bool `False`, so the writer's `isinstance(target, list)` guard is false and the call ID is never appended:

```python
ctx = RunContextWrapper(context=None)
ctx.reject_tool(item("d1"), always_reject=True)
ctx.approve_tool(item("d2"))
print(ctx.get_approval_status("do_thing", "d2"))
```

```
False
```

Here the approval never reaches state at all.

This matters because rejecting one call from the interruption list after approving the tool for the session is a normal human in the loop flow, and the current behaviour runs the tool anyway.

I do not think this is the "approval takes precedence" rule the comment above that branch refers to. That rule is about `approved` and `rejected` conflicting at the same granularity, which is what `tests/test_run_state.py` covers. Elsewhere the semantics are latest decision wins, and `tests/test_run_context_wrapper.py` asserts that `always_approve` followed by `always_reject` gives `False`. No test covers a per-call decision after a sticky one in either direction.

A fix would check the per-call list before the sticky bool, and stop a sticky write from replacing the list form, while keeping those two existing tests green.

Verified on `1816d2e`.

I would like to work on this.

chiruu12:
### Summary

Sticky and per-call approval decisions share the same two fields on `_ApprovalRecord` but use different shapes: `always_approve` and `always_reject` store a bool, while a per-call decision stores call IDs in a list. Once a sticky decision existed, a later per-call decision on the same tool was lost.

Rejecting one call after `always_approve` was ignored on read. `get_approval_status` returned at `approval_entry.approved is True` before reaching `rejected_ids`, so the rejection and its message were stored but never read, and the tool ran. In the other direction, `always_reject` leaves the bool `False` in `approved`, so the writer's `isinstance(target, list)` guard was false and a later `approve_tool` was dropped before it reached state.

Rejecting a single call from the interruption list after approving a tool for the session is a normal human in the loop flow, so this reads as a safety issue rather than a cosmetic one.

The reader now checks a decision recorded for the exact call before falling back to the sticky value, and the writer starts a list when the field still holds a bool. Sticky behaviour for every other call is unchanged, and `always_approve` followed by `always_reject` still resolves to `False`.

### Test plan

Added `test_run_context_per_call_decision_overrides_sticky_decision`, which covers both directions and asserts that other calls keep following the sticky decision. It fails on main and passes here.

`.agents/skills/code-change-verification/scripts/run.sh` passes: format, lint, typecheck and the full test suite.

### Issue number

Closes #4429

### Checks

### DIFF
--- a/src/agents/run_context.py
+++ b/src/agents/run_context.py
@@ -1316,7 +1316,9 @@ class RunContextWrapper(Generic[TContext]):
                     call_id for call_id in decision if call_id not in self._tool_invocations
                 )
 
-    def _rebuild_hosted_mcp_approvals(self, approvals: Any) -> None:
+    def _rebuild_hosted_mcp_approvals(
+        self, approvals: Any, *, per_call_overrides_sticky: bool = True
+    ) -> None:
         """Restore typed hosted MCP approval records from serialized state."""
         if not isinstance(approvals, list):
             return
@@ -1351,7 +1353,9 @@ class RunContextWrapper(Generic[TContext]):
                 key = ("hosted_mcp_query", tool_name, request_id)
             else:
                 continue
-            self._approvals[key] = self._restore_approval_record(decision)
+            self._approvals[key] = self._restore_approval_record(
+                decision, per_call_overrides_sticky=per_call_overrides_sticky
+            )
 
     def _fork_with_tool_input(self, tool_input: Any) -> RunContextWrapper[TContext]:
         """Create a child context that shares approvals and usage with tool input set."""
--- a/src/agents/run_context.py
+++ b/src/agents/run_context.py
@@ -676,6 +676,10 @@ class RunContextWrapper(Generic[TContext]):
 
     @staticmethod
     def _get_rejection_message_for_key(record: _ApprovalRecord, call_id: str) -> str | None:
+        # An approval recorded for this exact call overrides a sticky rejection, so there
+        # is no rejection reason to report for it.
+        if isinstance(record.approved, list) and call_id in record.approved:
+            return None
         if record.rejected is True:
             if call_id in record.rejection_messages:
                 return record.rejection_messages[call_id]
@@ -1001,17 +1005,19 @@ class RunContextWrapper(Generic[TContext]):
             if isinstance(opposite, list) and call_id in opposite:
                 opposite.remove(call_id)
 
-            # A sticky decision leaves a bool in this field, and a per-call decision
-            # taken afterwards still has to be recorded, so start a list for it.
+            # A sticky decision leaves a bool in this field. A per-call decision that
+            # agrees with it is already covered, so keep the sticky value; one that
+            # opposes it still has to be recorded, so start a list for it.
             target = approval_entry.approved if approve else approval_entry.rejected
-            if not isinstance(target, list):
-                target = []
-                if approve:
-                    approval_entry.approved = target
-                else:
-                    approval_entry.rejected = target
-            if call_id not in target:
-                target.append(call_id)
+            if target is not True:
+                if not isinstance(target, list):
+                    target = []
+                    if approve:
+                        approval_entry.approved = target
+                    else:
+                        approval_entry.rejected = target
+                if call_id not in target:
+                    target.append(call_id)
             if approve:
                 self._clear_rejection_message(approval_entry, call_id)
             elif call_id is not None:
--- a/src/agents/run_context.py
+++ b/src/agents/run_context.py
@@ -616,6 +616,21 @@ class RunContextWrapper(Generic[TContext]):
         if approval_entry is None:
             return None
 
+        approved_ids = (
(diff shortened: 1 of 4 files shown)


seratch (changes requested):
Thanks for addressing the sticky-state regression and rejection-message handling. I rechecked the latest head, and the approval transition logic now behaves correctly in both directions.

One compatibility blocker remains. A per-call approval after `always_reject` is serialized as schema `1.15` with `approved=[call_id]` and `rejected=true`. The released `v0.20.0` reader accepts schema `1.15` but resolves that call as rejected, so a snapshot written by the updated SDK can be silently mis-resumed by the released reader. Please bump `CURRENT_SCHEMA_VERSION`, add its summary, and cover backward reading of an existing `1.15` mixed record plus new-schema round-trip and resumed behavior. Please also add explicit same-call exception-reversal tests in both directions.

Once those focused changes and CI are complete, this should be ready for another review.

chiruu12:
Both fixed.

The first was a regression I introduced: a per-call decision that matched the sticky one replaced the bool with a list, so other calls lost the sticky answer. It now only converts when the new decision opposes the sticky value.

The second skips the sticky message for a call in the approved list. Other calls still get it.

Tests cover both. Verification is green.

chiruu12:
@codex review

seratch (changes requested):
Thanks for the update. I refreshed the latest remote head and checks. All 14 checks are green and the branch is cleanly mergeable, but the head is still `4f20ffaff620` and does not include the requested `RunState` schema changes or durable-state tests.

The existing changes-requested review therefore remains unresolved. Please bump the schema version, add the schema summary, cover backward reading of the existing `1.15` mixed approval record, and test new-schema round-trip and resumed behavior. Please also add the explicit same-call exception-reversal cases in both directions. Once those changes are pushed and CI passes again, this should be ready for another review.

I may implement a complete solution w/ your contribution credit later.

chiruu12:
Pushed. Bumped to `1.16` with a summary, gated the override on `_PER_CALL_APPROVAL_OVERRIDE_MIN_SCHEMA_VERSION` so a `1.15` snapshot still resolves as sticky, and added round-trip, backward-read and same-call reversal tests in both directions. Also moved `1.15` into the released list in `test_supported_schema_versions_match_released_boundary`.

One test is still red and I did not want to guess at it. `test_historical_fixture_corpus_matches_supported_schema_versions` wants `versions` and `features` entries for `1.16`, but those reference the commit that introduced the version, and `1.16` has none until this merges. `canonical_compatibility` does not fit either, since it relabels a later writer's payload and there is no later writer. How would you like that handled?

chiruu12:
Separate question while I was checking the compatibility direction. Released `0.20.0` writes `approved=true, rejected=[call_id]` itself when a call is rejected after `always_approve`, and its reader resolves that as approved, so the tool runs after a resume even though the call was rejected. That looks like the same fail-open surviving into existing `1.15` snapshots rather than something this PR introduces. Does that match your read, and is it in scope here or better as its own issue?

chiruu12:
Threaded the same flag through `_rebuild_hosted_mcp_approvals`, good catch. The corpus point is the open question in my previous comment.

seratch:
Thanks for reporting the issue. https://github.com/openai/openai-agents-python/pull/4447 resolved it and its commit includes you as a co-author. Thanks again for your time and efforts here.

PR #4431, opened 2026-08-15
fix(apply-diff): reject partially matched stacked anchors

Coiggahou2002:
### Summary

This pull request fixes stacked `@@` anchors so every non-empty anchor in a stack must match in sequence. Previously, if one anchor matched and a later anchor did not, `apply_diff` continued from the partial match and could apply the hunk body to unrelated nearby content.

Single unmatched anchors keep their existing context fallback. Bare anchors, whitespace-fuzzy matches, and fully matched stacks also keep their existing behavior. A partially matched stack now raises `ValueError` before edited content is returned; the sandbox boundary converts that error to `ApplyPatchDiffError` before writing the file.

This brings the Python parser in line with the same safety rule added in openai/openai-agents-js#1655.

### Test plan

- [x] Added direct parser coverage for partial stacks, missing first anchors, missing anchors followed by bare markers, and single-anchor compatibility.
- [x] Added sandbox coverage proving a rejected partial stack leaves the original file unchanged.
- [x] Ran `.agents/skills/code-change-verification/scripts/run.sh` with format, lint, typecheck, and the complete test suite passing.

### Issue number

N/A. Found while comparing stacked-anchor behavior across the Python and JavaScript SDKs.

### Checks

### DIFF
--- a/src/agents/apply_diff.py
+++ b/src/agents/apply_diff.py
@@ -130,18 +130,20 @@ def _parse_update_diff(lines: list[str], input: str) -> ParsedUpdateDiff:
     cursor = 0
 
     while not _is_done(parser, END_SECTION_MARKERS):
-        anchors, has_anchor = _read_anchors(parser)
+        anchors, anchor_count = _read_anchors(parser)
 
-        if not (has_anchor or cursor == 0):
+        if not (anchor_count > 0 or cursor == 0):
             current_line = parser.lines[parser.index] if parser.index < len(parser.lines) else ""
             raise ValueError(f"Invalid Line:\n{current_line}")
 
+        require_anchor_match = anchor_count > 1
         for index, anchor in enumerate(anchors):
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
Before merge, please add the released-compatibility regression covered in openai/openai-agents-js#1655: two stacked hunks in the same parent block must be able to reuse a parent anchor that is already before the current cursor, while each child anchor still matches forward. The implementation already appears to preserve this behavior; the missing test is needed because this PR rewrites that exact match-accounting branch.

Coiggahou2002:
Added in bec42e83. The new public `apply_diff` regression mirrors openai/openai-agents-js#1655: the first hunk advances past the parent, then the second stacked hunk reuses that prior parent anchor while its child anchor still matches forward. No implementation change was needed. Focused tests and the repository-wide format/lint/typecheck/test stack all pass.

seratch (approved):

PR #4441, opened 2026-08-15
fix(memory): reject a second base-table pair in one AdvancedSQLiteSession database

ayaangazali:
### Summary

`AdvancedSQLiteSession` inherits the configurable `sessions_table` / `messages_table` from `SQLiteSession`, but its four structure tables (`message_structure`, `turn_usage`, `branch_reservations`, `session_clear_generations`) are hardcoded names created with `CREATE TABLE IF NOT EXISTS` (`advanced_sqlite_session.py:150`, `:170`). The first base-table pair to touch a database file therefore owns them, and every later session on that file silently reuses `message_structure` with the first pair's foreign keys regardless of how it was configured.

`get_items()` joins `message_structure` against the configured messages table (`advanced_sqlite_session.py:279`, `:297`, `:318`), so once two pairs share one file each session's join picks up structure rows written by the other one:

```python
a = AdvancedSQLiteSession(session_id="s1", db_path=p, create_tables=True,
                          sessions_table="a_sessions", messages_table="a_messages")
b = AdvancedSQLiteSession(session_id="s1", db_path=p, create_tables=True,
                          sessions_table="b_sessions", messages_table="b_messages")

await a.add_items([{"role": "user", "content": "from A"}])
await b.add_items([{"role": "user", "content": "from B"}])

await a.get_items()  # [{... 'from A'}, {... 'from A'}]  <- only one item was added
await b.get_items()  # [{... 'from B'}, {... 'from B'}]
```

Two `message_structure` rows now exist for session `s1`, both with `message_id = 1`, and each session matches both of them against its own messages table. Nothing errors, the history is just wrong. `turn_usage` collides the same way through its `UNIQUE(session_id, branch_id, user_turn_number)` constraint.

Namespacing the structure tables per base-table pair would mean renaming them in over 90 places in the module and would orphan the rows in every database file that already exists, so this takes the other direction: the layout really does only support one pair per file, so `__init__` now checks that instead of corrupting data. It reads `PRAGMA foreign_key_list(message_structure)` and raises a `ValueError` naming both pairs when the structure tables already belong to a different one.

Supported configurations are untouched, since the check only fires on a mismatch that is already broken today. Verified by hand: default table names with two sessions on one file, a custom pair on its own file, reopening that file with `create_tables=True` and with `create_tables=False`, a fresh file with no structure tables yet, and `:memory:`.

### Test plan

- Added `test_structure_tables_reject_a_second_base_table_pair`. Reverting the source change makes it fail with `DID NOT RAISE <class 'ValueError'>`, so it covers the fix rather than restating it.
- `uv run pytest tests/extensions/memory/ tests/memory/`: 613 passed, 11 skipped.
- Ran `.agents/skills/code-change-verification/scripts/run.sh`: format and lint clean, pyright 0 errors, 8587 passed / 28 skipped.
- CI here waits on a maintainer to approve the workflow for a first-time contributor, so the checks above are the local run, not a green CI badge.

### Issue number

N/A, found while reading through the session backends.

### Checks

---

sorry in advance if I got any of this wrong. I worked through the logic myself and used Claude Code as a sounding board for the design call (validate vs rename the tables), and I confirmed the duplication with the snippet above before writing anything. happy to be corrected on the approach, still a freshman in college trying to be useful here :)

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -67,15 +67,16 @@ class AdvancedSQLiteSession(SQLiteSession):
             session_settings=session_settings,
             **kwargs,
         )
-        if create_tables:
-            try:
+        try:
+            if create_tables:
                 self._init_structure_tables()
+            self._verify_structure_tables_scope()
+        except BaseException:
+            try:
+                self.close()
             except BaseException:
-                try:
-                    self.close()
-                except BaseException:
-                    pass
-                raise
+                pass
+            raise
         self._current_branch_id = "main"
         # Synchronized with the durable session_clear_generations row whenever a
         # branch pointer is established or a write begins. A mismatch means
@@ -107,6 +108,34 @@ class AdvancedSQLiteSession(SQLiteSession):
             self._generation = durable_generation
             self._current_branch_id = branch_id
             return True
+
+    def _verify_structure_tables_scope(self) -> None:
+        """Reject a database file whose structure tables belong to other base tables.
+
+        ``message_structure`` and the other structure tables are not named after
+        ``sessions_table``/``messages_table``, so a database file can only hold the structure
+        rows of a single base-table pair. ``CREATE TABLE IF NOT EXISTS`` keeps the first pair's
+        foreign keys, so a second session configured with different base table names would join
+        ``message_structure`` against the wrong messages table and read back rows that belong to
+        the other pair.
+        """
+        with self._locked_connection() as conn:
+            references = {
+                row[3]: row[2] for row in conn.execute("PRAGMA foreign_key_list(message_structure)")
+            }
+        if not references:
+            return
+        if (
+            references.get("session_id") != self.sessions_table
+            or references.get("message_id") != self.messages_table
+        ):
+            raise ValueError(
+                f"The structure tables in {self.db_path} already belong to "
+                f"'{references.get('session_id')}'/'{references.get('message_id')}', not to the "
+                f"configured '{self.sessions_table}'/'{self.messages_table}'. Structure tables "
+                "are shared per database file, so give each sessions_table/messages_table pair "
+                "its own db_path."
+            )
 
     def _init_structure_tables(self):
         """Add structure and usage tracking tables.
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -39,6 +39,17 @@ def _content_preview(content: Any, max_length: int | None = None) -> str:
     return text
 
 
+def _existing_schema_names(db_path: str | Path) -> frozenset[str]:
+    """Names already present in a database file, before anything is created in it."""
+    if str(db_path) == ":memory:":
+        return frozenset()
+    path = Path(db_path)
+    if not path.exists():
+        return frozenset()
+    with closing(sqlite3.connect(str(path))) as conn:
+        return frozenset(row[0] for row in conn.execute("SELECT name FROM sqlite_master"))
+
+
 class AdvancedSQLiteSession(SQLiteSession):
     """Enhanced SQLite session with conversation branching and usage analytics."""
 
@@ -61,6 +72,9 @@ class AdvancedSQLiteSession(SQLiteSession):
             logger: The logger to use. Defaults to the module logger
             **kwargs: Additional keyword arguments to pass to the superclass
         """  # noqa: E501
+        # Captured before the base schema is created so a rejected construction can drop exactly
+        # what it added, rather than leaving the refused pair's tables behind.
+        preexisting_schema = _existing_schema_names(db_path)
         super().__init__(
             session_id=session_id,
             db_path=db_path,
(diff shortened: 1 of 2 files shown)


ayaangazali:
Both Codex points were real. Fixed in f68b4094.

**Identifier casing.** This was a regression my patch introduced. SQLite resolves table names case-insensitively, but `PRAGMA foreign_key_list` reports the CREATE-time spelling, so reopening a file as `foosessions` after creating it as `FooSessions` hit the mismatch branch and raised on a configuration that is the same pair. Comparison is case-folded now, and `test_structure_tables_accept_equivalent_identifier_casing` covers it. That test fails with `ValueError` against the previous commit.

**Split layout.** Verification now checks `turn_usage` as well as `message_structure`, so a file whose two structure tables were created for different pairs is rejected instead of being accepted by whichever one happens to match. It also runs inside the transaction that creates the layout: `_init_structure_tables` now opens with `BEGIN IMMEDIATE`, which takes the database write lock before the first `CREATE`, and Python's sqlite3 keeps DDL inside the open transaction, so create-and-verify commits as one unit and two processes cannot each create half the layout for a different pair.

Checked by hand on top of the existing test: recased pair accepted with correct history, a file where only `turn_usage` carries the owner rejected through the `turn_usage` check, and the original two-pair case still rejected with the first session's history intact.

`.agents/skills/code-change-verification/scripts/run.sh` passes: format, lint, typecheck and the full test run.

seratch (changes requested):
Thanks for the contribution. The underlying issue is real: in the released implementation, two custom base-table pairs sharing one database can silently duplicate or mix persisted history. Enforcing one pair per database is the right direction and is preferable to renaming every structure table.

Before another review, please reset the validation around one contract: `create_tables=True` must atomically create and claim a complete structure-table layout, while `create_tables=False` must accept only an already-existing complete layout owned by the configured pair. Please delegate identifier equality to SQLite's ASCII-only `NOCASE` semantics instead of Python `casefold()`.

Please add regression tests for a no-create instance opened before ownership exists, ASCII recasing versus distinct non-ASCII identifiers, and controlled concurrent initialization with final foreign-key coherence. The current head still reproduces silent duplicated history in the first two cases.

ayaangazali:
Thanks, that framing helped. Reset around the single contract in 6517db95.

**One contract.** `_claim_structure_tables` is now the only validation and both constructor paths run it:

- `create_tables=True` creates and claims inside one `BEGIN IMMEDIATE` transaction, so the layout is created, checked and committed as a unit.
- `create_tables=False` accepts only an already-existing complete layout owned by the configured pair. An incomplete or unclaimed layout is rejected, which closes the case you flagged: a no-create session could previously open a file before any pair owned it and then have another pair claim the structure tables underneath it.

Both owner-bearing tables, `message_structure` and `turn_usage`, must be present and agree. `branch_reservations` and `session_clear_generations` carry no foreign keys, but once one pair owns the file they cannot be shared across pairs either.

**Identifier equality.** Now delegated to SQLite via `SELECT ? = ? COLLATE NOCASE` rather than Python `casefold()`. You were right that this mattered: `casefold()` equates `ßsessions` and `sssessions`, which SQLite keeps as two distinct tables, so the second pair got through and the corruption came back.

**Tests**, all three cases:

- `test_no_create_session_rejects_a_database_without_an_owner`
- `test_structure_tables_accept_equivalent_identifier_casing` for ASCII recasing, and `test_structure_tables_reject_distinct_non_ascii_identifiers` for the non-ASCII pair, which asserts the `casefold()` collision up front so the reason the test exists is visible
- `test_concurrent_structure_table_claims_leave_one_coherent_owner`, two processes claiming a fresh file with different pairs, asserting exactly one claim and that both owner-bearing tables end up naming that winner

Four of the five fail against `main`. The fifth is the ASCII recasing test, which guards against over-rejection rather than against `main`.

One note on how that concurrency test is written, since it looks over-specified otherwise. It pre-creates the file with WAL enabled before the race. Switching journal mode needs an exclusive lock, so two processes opening a brand-new file collide in `SQLiteSession.__init__` at `sqlite_session.py:107` (`PRAGMA journal_mode=WAL`) before either reaches the structure tables. On `main`, with the same base-table pair and none of this validation involved, that reproduces 6 out of 6 times with `OperationalError: database is locked`. That is pre-existing and outside this change, so I settled WAL up front to keep the test about the claim. Happy to open a separate issue for it if it is worth tracking.

Behavior change worth confirming: `AdvancedSQLiteSession(session_id=..., create_tables=False)` against a database with no structure tables now raises at construction instead of at the first `add_items()`. On `main` that configuration already fails, with `OperationalError: no such table: message_structure`, and every usage in `docs/sessions/` and `examples/memory/` passes `create_tables=True`, so I do not think anything working breaks. Say the word if you would rather it stayed lazy.

ayaangazali:
Fair point, and my change is what made that state reachable, so it should clean up after itself. Fixed in the latest commit.

I took the rollback option rather than preflighting before `super().__init__()`. Preflighting needs the effective `sessions_table`/`messages_table` before the base class has resolved them, which would mean duplicating its default names here as a second source of truth. Instead `__init__` snapshots `sqlite_master` before the base schema is created, and a rejected claim drops only the names that snapshot did not contain. A pair that already owned tables in the file keeps them.

`test_rejected_construction_leaves_no_schema_behind` asserts the file's schema is byte-for-byte the same set before and after a rejected construction, and that the owner's history still reads correctly. It fails against the previous commit.

seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/4454, which includes you as a co-author.

PR #4444, opened 2026-08-15
fix(core): count a Responses request when the provider omits usage

ayaangazali:
### Summary

`Usage.requests` is documented as the number of requests made to the LLM API, but `OpenAIResponsesModel` reports zero for a call that completed without a `usage` block:

```python
usage = (
    _response_usage_to_usage(response.usage)
    if response.usage is not None
    else Usage()          # requests defaults to 0
)
```

That is the same defect #4290 fixed for Chat Completions, LiteLLM and AnyLLM. The Responses adapter was not in that PR's scope, so it still undercounts. Reproduced on `main` against a client returning a `Response` with `usage=None`:

```
OpenAIResponsesModel      -> usage.requests == 0
OpenAIChatCompletionsModel -> usage.requests == 1
```

Same completed request, two different answers. It matters for the deployments #4290 was written for: OpenAI-compatible gateways behind a custom `base_url` that omit `usage`, where request accounting and rate-limit tracking built on `Usage.requests` stay at zero for a whole run. `openai_responses.py:533` already handles `response.usage is None`, so the adapter expects this shape.

**Where the mark goes matters.** The obvious patch, `else Usage(requests=1)` in `get_response`, is wrong: `OpenAIHostedMultiAgentModel` extends `OpenAIResponsesModel`, inherits `get_response` and `stream_response`, and deliberately returns usage-less responses because it multiplexes several provider responses into one and reports its counts separately through `_agents_sdk_request_count`. That version double-counted and broke `test_injection_failure_after_completion_starts_continuation_response`, which is exactly what the existing `_requests_for_response_without_usage` docstring warns about.

So the count happens at the four transport exits this class owns, where "one request happened" is actually known:

- `_fetch_response`, non-streaming HTTP
- `OpenAIResponsesWSModel._fetch_response`, non-streaming websocket
- `_ResponseStreamWithRequestId.__anext__`, the HTTP stream wrapper this class constructs
- `_iter_websocket_response_events`, the websocket event iterator

`OpenAIHostedMultiAgentModel` overrides `_fetch_response` and drives its own `_iter_websocket_turn`, so it reaches none of the four and is excluded by construction rather than by a check on `self`. Only `response.completed` counts; `failed`, `incomplete` and `error` raise before any usage is consumed. The response span carries the synthesized count too, on both the streamed and non-streamed paths, mirroring what `openai_chatcompletions.py` already does at the equivalent point.

One gap left on purpose: the compatibility branch in `_fetch_response` for custom clients that implement `responses.create()` but not `with_streaming_response` hands back the provider's iterator unwrapped, so it still does not count. Routing it through `_ResponseStreamWithRequestId` would change that path's stream lifecycle, and the wrapper needs an async iterator where that branch may return an async iterable.

### Test plan

- Five regression tests: HTTP and websocket, each streaming and non-streaming, plus the span payload. All five fail against `main` and pass here, so they cover the fix rather than restating it. Each asserts token counts stay at zero, since no usage payload is synthesized.
- `uv run pytest tests/models/ tests/extensions/`: 2235 passed, 13 skipped. The Hosted Multi-Agent suite passes unchanged, which is the check that the placement above is right.
- CI here needs a maintainer to approve the workflow for a first-time contributor, so the above is the local run rather than a green badge.

### Issue number

N/A. Found while comparing the Responses and Chat Completions adapters.

### Checks

---

heads up in case I misread something: I traced this myself off #4290 and checked the hosted multi-agent interaction after my first attempt broke its tests, and I talked the placement decision through with Claude Code before settling on it. if the streamed half should have been in here too, or the whole thing belongs elsewhere, tell me and I'll redo it. college freshman, still learning how these boundaries are meant to be drawn :)

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -231,6 +231,17 @@ class OpenAIResponsesWebSocketOptions(TypedDict):
     """
 
 
+def _count_transport_request_without_usage(response: object) -> None:
+    """Record a completed request whose provider payload carried no usage.
+
+    Only the transports this class owns call this. ``OpenAIHostedMultiAgentModel`` overrides
+    ``_fetch_response`` to multiplex several provider responses into one and reports its counts
+    separately, so it is excluded by construction.
+    """
+    if isinstance(response, Response) and response.usage is None:
+        _mark_request_completed_without_usage(response)
+
+
 class _ResponseStreamWithRequestId:
     """Wrap an SDK event stream and retain the originating request ID."""
 
@@ -274,6 +285,8 @@ class _ResponseStreamWithRequestId:
         event_type = getattr(event, "type", None)
         if event_type in self._TERMINAL_EVENT_TYPES:
             self._yielded_terminal_event = True
+        if event_type == "response.completed":
+            _count_transport_request_without_usage(getattr(event, "response", None))
         return event
 
     async def aclose(self) -> None:
@@ -535,7 +548,9 @@ class OpenAIResponsesModel(Model):
                     if response.usage is not None
                     else Usage(requests=_requests_for_response_without_usage(response))
                 )
-                if response.usage is not None:
+                if response.usage is not None or usage.requests:
+                    # Keep the span in step with the returned usage, so request-count dashboards
+                    # built on spans do not disagree with the run.
                     span_response.span_data.usage = model_usage_to_span_usage(usage)
 
                 if tracing.include_data():
@@ -675,6 +690,10 @@ class OpenAIResponsesModel(Model):
                     span_response.span_data.usage = model_usage_to_span_usage(
                         _response_usage_to_usage(final_response.usage)
                     )
+                elif final_response is not None and _requests_for_response_without_usage(
+                    final_response
+                ):
+                    span_response.span_data.usage = model_usage_to_span_usage(Usage(requests=1))
 
             except Exception as e:
                 span_response.set_error(
@@ -749,11 +768,7 @@ class OpenAIResponsesModel(Model):
 
         if not stream:
             response = await client.responses.create(**create_kwargs)
-            if isinstance(response, Response) and response.usage is None:
-                # This call is one request even though the provider reported no usage. Marked
-                # here rather than in `get_response` so subclasses that override `_fetch_response`
-                # to multiplex several provider responses keep their own request accounting.
-                _mark_request_completed_without_usage(response)
+            _count_transport_request_without_usage(response)
             return cast(Response, response)
 
         streaming_response = getattr(client.responses, "with_streaming_response", None)
@@ -1201,8 +1216,7 @@ class OpenAIResponsesWSModel(OpenAIResponsesModel):
                 f"{terminal_event_hint}"
             )
 
-        if final_response.usage is None:
-            _mark_request_completed_without_usage(final_response)
+        _count_transport_request_without_usage(final_response)
         return final_response
 
     async def _iter_websocket_response_events(
@@ -1286,6 +1300,8 @@ class OpenAIResponsesWSModel(OpenAIResponsesModel):
                         }
                         if is_terminal_event:
                             yielded_terminal_event = True
+                        if event_type == "response.completed":
+                            _count_transport_request_without_usage(getattr(event, "response", None))
                         yield event
 
                         if is_terminal_event:
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -5,7 +5,7 @@ import contextlib
 import inspect
 import json
 import weakref
-from collections.abc import AsyncIterator, Awaitable, Callable, Mapping, Sequence
+from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable, Mapping, Sequence
 from contextvars import ContextVar
 from dataclasses import asdict, dataclass, is_dataclass
 from enum import Enum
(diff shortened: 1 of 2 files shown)


ayaangazali:
Both points were right. Fixed in the latest commit.

**Streaming.** You are right that this left a streaming versus non-streaming split, which `AGENTS.md` calls out directly. The reason I had scoped it out was `OpenAIHostedMultiAgentModel`: it inherits `stream_response`, and marking the terminal event there would have double-counted its synthetic multiplexed responses, which is what broke `test_injection_failure_after_completion_starts_continuation_response` on my first attempt.

The way out was to stop thinking about `stream_response` and count at the transport exits this class actually owns, which is where "one request happened" is known:

- `_fetch_response`, non-streaming HTTP
- `OpenAIResponsesWSModel._fetch_response`, non-streaming websocket
- `_ResponseStreamWithRequestId.__anext__`, the HTTP stream wrapper this class constructs
- `_iter_websocket_response_events`, the websocket event iterator

`OpenAIHostedMultiAgentModel` overrides `_fetch_response` and drives its own `_iter_websocket_turn`, so it never reaches any of the four and its accounting is untouched. That is checked by the Hosted Multi-Agent suite still passing rather than by a type check on `self`. Only `response.completed` is counted; `failed`, `incomplete` and `error` raise before any usage is consumed.

**Span usage.** Fixed on both paths, non-streaming and streamed, mirroring what `openai_chatcompletions.py` already does at the equivalent point.

Five regression tests now, covering HTTP and websocket in both streaming and non-streaming form plus the span payload. All five fail against `main` and pass here. `uv run pytest tests/models/ tests/extensions/` is green at 2235 passed, and `.agents/skills/code-change-verification/scripts/run.sh` passes.

One gap I am deliberately leaving, so it is not a surprise: the compatibility branch in `_fetch_response` for custom clients that implement `responses.create()` but not `with_streaming_response` returns the provider's iterator directly and is not wrapped, so it still does not count. Routing it through `_ResponseStreamWithRequestId` would change that path's stream lifecycle, and the wrapper needs an async iterator where that branch may hand back an async iterable. Happy to take it on if you would rather have it covered.

seratch (changes requested):
Thanks for the contribution. The underlying issue is valid, and the current head fixes the main HTTP and WebSocket paths. Two supported streaming cases still need correction before merge:

- Route the `responses.create()` compatibility fallback through the same successful-terminal accounting as the normal HTTP stream path while preserving its existing async-iterable and cleanup behavior.
- Populate the response span usage before yielding `response.completed`, so a direct consumer that closes after the terminal event does not lose the synthesized request count.

Please add regression tests for the fallback async-iterable path and for a direct `stream_response()` consumer that closes immediately after the terminal event. Preserve absent raw usage, zero token counts, and Hosted Multi-Agent's separate accounting. Once these cases are covered and CI passes, this should be ready for another review.

ayaangazali:
Thanks, both were right. Done in 95952a20.

**Fallback path.** It now goes through `_ResponseStreamWithRequestId`, so terminal-event accounting is the same code as the normal HTTP stream rather than a second copy. To keep that path's behavior, the wrapper resolves its iterator on first `__anext__` instead of at construction, so a client handing back an async iterable still works, and `_close_stream_once` closes the object it was given, so cleanup is unchanged. The fallback passes `request_id=None` and a no-op cleanup, since the SDK never opened that transport.

**Span usage before the yield.** Moved into the `ResponseCompletedEvent` branch, so it runs before `yield chunk` rather than after the loop. The post-loop copy is gone, so there is one place that sets it.

Two tests added for exactly those cases:

- `test_fallback_stream_counts_request_when_provider_omits_usage` uses a client with no `with_streaming_response` whose `create()` returns an async iterable with no `__anext__` of its own, and asserts the run reports one request.
- `test_streamed_span_usage_survives_a_consumer_that_stops_at_the_terminal_event` consumes `stream_response()` directly, breaks at `response.completed`, closes the generator, and asserts the span still carries the count.

Both fail against the previous head. Raw usage stays absent and token counts stay zero in every one of these, and the Hosted Multi-Agent suite is untouched and still passing, which is the check that the exclusion still holds.

On CI: `tests (3.11)` on the previous head failed on `tests/realtime/test_session.py::test_tracked_task_reentering_active_cleanup_does_not_create_wait_cycle` with a `TimeoutError`, which I do not think is this PR. That test is pure in-process asyncio with a `wait_for(..., timeout=0.5)` deadlock guard, it ran on xdist worker `gw2`, and the same commit passed `tests` on 3.10, 3.12, 3.13, 3.14 and Windows. `agents.models.openai_responses` is also not reachable from `agents.realtime.session` through any import path, so this diff cannot affect it. I could not reproduce it locally in 25 serial runs, in the realtime suite at `-n 24`, or in two full `-n 9 --dist worksteal` runs under heavy CPU contention. I have not pushed anything for it. If it reappears on this new run I will dig further rather than assume.

ayaangazali:
Good catch, that was a leak I introduced when I widened the wrapper to accept an iterable. Fixed in the latest commit.

`_close_stream_once` now closes the resolved iterator and then the source, and only closes the source separately when it is a different object, so the normal path where `__aiter__()` returns `self` still closes exactly once.

Worth noting how it is tested, because my first attempt at a regression test was useless: asserting on the existing exhaustion test passes either way, since an exhausted async generator runs its own `finally`. The leak only shows on an early close. `test_fallback_stream_close_releases_the_resolved_iterator` consumes one event, closes the stream, and asserts the generator built by `__aiter__()` was finalized as well as the source. It fails against the previous commit and passes here.

seratch:
Thanks for trying to resolve the issue. https://github.com/openai/openai-agents-python/pull/4453 resolves it and its commit includes your contribution credit.

PR #4445, opened 2026-08-15
fix(items): guard non-string type before tool_search set membership

Amey-Thakur:
## Summary

`_output_item_to_input_item` reads a raw item's `type` and tests it against a set of tool_search type strings before its own dict / `BaseModel` handling runs. The raw item is untrusted, so `type` can be a non-string, unhashable value. Set membership then raises `TypeError: unhashable type` instead of the converter falling through to produce the payload or raising the documented `AgentsException`.

```python
from agents.items import _output_item_to_input_item

_output_item_to_input_item({"type": [], "id": "x"})
# TypeError: unhashable type: 'list'
```

The converter is otherwise defensive about the raw item's outer shape (it raises `AgentsException` for a non-dict / non-`BaseModel` item), so this is a gap in that same contract one level in: the guard protects the container but not the `type` value it reads out of it.

## Fix

Guard the membership test with `isinstance(item_type, str)`. A malformed `type` now falls through to the normal dict handling rather than crashing, and every well-formed item is unchanged.

```python
if isinstance(item_type, str) and item_type in {"tool_search_call", "tool_search_output"}:
    return _tool_search_item_to_input_item(raw_item)
```

## Tests

`test_output_item_to_input_item_non_string_type_does_not_crash` in `tests/test_items_helpers.py`, parametrized over `[]`, `{}`, `5`, and `None`. The unhashable cases raise a raw `TypeError` at `items.py:234` without this change and pass with it. `tests/test_items_helpers.py` is green (41 passed); `ruff check`, `ruff format --check`, and `mypy` on the changed files are clean.

## Note on scope

The same `raw.get("type") in {...}` set-membership pattern, guarded on the container being a dict / `Mapping` but not on `type` being a string, appears at other item and state deserialization sites:

- `src/agents/run_state.py` (`_raw_item_uses_programmatic_tool_calling`, and the `item.get("type") in {...}` check nearby)
- `src/agents/handoffs/history.py` (`input_item.get("type") in {...}`)
- `src/agents/run_internal/oai_conversation.py` (`_is_tool_search_item`)

They share the same failure mode on a non-string `type`. I kept this PR focused on the one converter I could exercise directly with a regression test; happy to extend the same guard to those sites in this PR or a follow-up, whichever the maintainers prefer.

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -231,7 +231,12 @@ def _output_item_to_input_item(raw_item: Any) -> TResponseInputItem:
     item_type = (
         raw_item.get("type") if isinstance(raw_item, dict) else getattr(raw_item, "type", None)
     )
-    if item_type in {"tool_search_call", "tool_search_output"}:
+    # ``item_type`` comes from an untrusted raw item, so it may be a non-string
+    # (for example ``{"type": []}``). Guarding on ``isinstance`` keeps the set
+    # membership test from raising ``TypeError: unhashable type`` before the
+    # dict/BaseModel handling below can produce the payload or raise the
+    # documented ``AgentsException``.
+    if isinstance(item_type, str) and item_type in {"tool_search_call", "tool_search_output"}:
         return _tool_search_item_to_input_item(raw_item)
 
     if isinstance(raw_item, dict):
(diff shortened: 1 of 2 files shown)


seratch:
I don't think there is such a real path. I am going to close this PR for now, but if you have real-world repro steps with a real provider, we may revisit this.

PR #4449, opened 2026-08-16
fix(sandbox): canonicalize shell approval workdir

sylvesterkaczmarek:
### Summary

Make `exec_command` path-sensitive approval callbacks inspect the same effective working directory that execution uses.

The tool previously exposed the raw model-provided `workdir` to `needs_approval`, even though execution normalizes separators and anchors relative paths beneath the run-scoped sandbox cwd. When `workdir` was omitted, the approval callback saw `None` while execution could still run inside a configured cwd.

This change resolves the effective workdir before invoking callable approval policies. Approval receives the normalized absolute sandbox workdir, including the run cwd when the model omits `workdir`. Boolean approval settings are unchanged. Invalid inputs continue into the normal tool path rather than invoking a path-sensitive approval callback with a misleading representation.

### Test plan

Added focused regression coverage for:

- shell approval with backslash-separated model input under a run-scoped cwd
- shell approval when `workdir` is omitted under a run-scoped cwd

Repository CI is used for full verification because this environment has GitHub connector access but no materialized dependency checkout.

### Issue number

None. Found while auditing sandbox authorization consistency after run-scoped cwd support landed.

### DIFF
--- a/src/agents/sandbox/capabilities/tools/shell_tool.py
+++ b/src/agents/sandbox/capabilities/tools/shell_tool.py
@@ -7,11 +7,17 @@ from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from typing import Any, ClassVar
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, ValidationError
 
 from ....run_context import RunContextWrapper
 from ....tool import FunctionTool
-from ...errors import ExecTimeoutError, ExecTransportError, PtySessionNotFoundError
+from ....util._approvals import evaluate_needs_approval_setting
+from ...errors import (
+    ExecTimeoutError,
+    ExecTransportError,
+    InvalidManifestPathError,
+    PtySessionNotFoundError,
+)
 from ...session.base_sandbox_session import BaseSandboxSession
 from ...types import User
 from ...util.token_truncation import formatted_truncate_text_with_token_count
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying approval-path mismatch is valid, but the current implementation does not cover the supported `Shell.configure_tools` path because that callback is assigned after `ExecCommandTool` construction.

Before merge, please move canonicalization into a runtime approval adapter that reads the current `self.needs_approval`, while keeping the configured callback unchanged on the public field. Build the callback parameters from a copy of the original parsed mapping and change only `workdir`, adding the resolved run cwd when it was omitted; do not materialize unrelated optional defaults. Please add regression tests through the actual FunctionTool approval path for both constructor-supplied and `configure_tools`-assigned callbacks, including assertions that unrelated key presence is preserved.

This keeps the fix scoped to the authorization mismatch without changing existing callback semantics unnecessarily.

sylvesterkaczmarek:
Closing this iteration after the review exposed a broader compatibility requirement. A constructor-only adapter does not cover `configure_tools` callbacks assigned after construction, while wrapping the public `needs_approval` field would break its documented/tested callback identity. The authorization finding remains valid, but the correct fix needs a separate runtime parameter-normalization hook rather than a partial wrapper. I do not want to leave a security-sensitive fix open with known bypass paths.

PR #4450, opened 2026-08-16
fix(sandbox): validate raster view_image content

sylvesterkaczmarek:
### Summary

Require `view_image` to identify signature-bearing raster image formats from the bytes it reads instead of trusting a raster-looking filename.

The previous detector checked known image signatures first, then accepted any filename whose MIME type began with `image/`. That meant arbitrary bytes stored as `report.png`, `photo.jpg`, or another raster image extension could be returned as structured image data.

This change narrows the filename fallback to SVG only. PNG, JPEG, GIF, WebP, BMP, and TIFF require their existing byte signatures. SVG and SVGZ retain the established filename fallback because SVG has no single binary magic value and SVGZ is compressed, so removing filename inference entirely would break valid BOM/prolog/alternate-encoding and compressed SVG inputs that were already supported.

### Test plan

Added regression coverage showing that:

- non-image bytes named `fake.png` are rejected
- valid PNG bytes named `payload.bin` are accepted as PNG
- SVG inputs beginning with a UTF-8 BOM, comment, or `DOCTYPE` remain accepted
- UTF-16 SVG remains accepted
- compressed SVGZ remains accepted

Repository CI is used for full verification because this environment has GitHub connector access but no materialized dependency checkout.

### Issue number

None. Found during a sandbox content-validation audit.

### DIFF
--- a/src/agents/sandbox/capabilities/tools/view_image.py
+++ b/src/agents/sandbox/capabilities/tools/view_image.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import base64
-import mimetypes
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -38,10 +37,6 @@ def _detect_image_mime_type(path: Path, payload: bytes) -> str | None:
     snippet = payload[:_SVG_SNIFF_BYTES].lstrip().lower()
     if snippet.startswith(b"<svg") or (snippet.startswith(b"<?xml") and b"<svg" in snippet):
         return "image/svg+xml"
-
-    guessed_type, _ = mimetypes.guess_type(path.name)
-    if isinstance(guessed_type, str) and guessed_type.startswith("image/"):
-        return guessed_type
     return None
 
 
--- a/src/agents/sandbox/capabilities/tools/view_image.py
+++ b/src/agents/sandbox/capabilities/tools/view_image.py
@@ -16,7 +16,6 @@ from ...workspace_paths import SandboxWorkspaceScope, coerce_posix_path, sandbox
 
 _MAX_IMAGE_BYTES = 10 * 1024 * 1024
 _MAX_IMAGE_SIZE_LABEL = "10MB"
-_SVG_SNIFF_BYTES = 2048
 
 
 def _consume_svg_prolog_construct(snippet: str) -> str | None:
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Before we can merge it, please preserve the deliberately supported SVG path when valid SVG content begins with a UTF-8 BOM, XML comment, or `DOCTYPE`, and add focused regression coverage for those prolog forms. The current head rejects those files because removing the extension fallback leaves `_detect_image_mime_type()` recognizing only `<svg` or `<?xml` at the start. Please also remove the now-unused `path` parameter and its `Path` import.

The invalid-`.png` test covers the base failure. After these changes and green CI, this should be ready for another review.

sylvesterkaczmarek:
Addressed the requested SVG compatibility cases and then simplified the implementation around the supported-format boundary. Raster formats with stable signatures now require byte evidence, while SVG/SVGZ retain the existing filename fallback so BOM/comment/DOCTYPE, UTF-16, and compressed SVG inputs stay compatible without adding an XML/decompression parser to media sniffing. Focused regression coverage includes the maintainer-requested prolog cases plus UTF-16 and SVGZ. The branch is mergeable; GitHub Actions remains gated before jobs are created.

sylvesterkaczmarek:
@codex review

sylvesterkaczmarek:
@seratch The requested SVG compatibility coverage is addressed on the current head. I simplified the fix so signature-bearing raster formats require byte evidence, while SVG/SVGZ retain their established filename fallback for BOM/comment/DOCTYPE, UTF-16, and compressed inputs. Fresh Codex and security reviews on `864bd7304c` found no issues. CI is still gated before any jobs are created; ready for re-review when available.

seratch:
@sylvesterkaczmarek can you resolve the CI errors?

sylvesterkaczmarek:
Resolved the CI failures in 8bd5d1b3: the regression test is now Ruff-formatted and explicitly narrows `ToolOutputImage.image_url` before the three `startswith()` assertions, addressing the only pyright errors. The previous CI run already passed the Python 3.10–3.14 and Windows test jobs; the failing jobs were limited to formatting and typecheck. Fresh CI should verify the fix.

sylvesterkaczmarek:
Fresh CI is now fully green on 8bd5d1b3. The requested formatting/typecheck fixes are verified; ready for re-review.

seratch:
We'll merge https://github.com/openai/openai-agents-python/pull/4462 instead; your contribution is included in its commit.

PR #4452, opened 2026-08-16
feat(sandbox): allow Docker sandboxes to disable networking

### LINKED ISSUE #4451: [Sandbox] Allow Docker sandboxes to run with networking disabled
### Please read this first

  - Yes. I reviewed the Sandbox client documentation and the current `DockerSandboxClient` implementation.
  - Yes. I searched open and closed issues and pull requests for `network_mode`, `network_disabled`, `DockerSandboxClientOptions`, `DockerSandboxClient` + `network`, and `allow_internet_access`. I did not find an existing request covering this functionality.

### Describe the feature

I would like `DockerSandboxClientOptions` to support creating a Docker sandbox with networking completely disabled.

This is useful for agents that execute model-directed shell commands but do not require network access inside the sandbox.

Today, `DockerSandboxClientOptions` exposes the Docker image and optional exposed ports, but `_create_container()` does not pass a network configuration to `docker_client.containers.create()`. As a result, Docker uses its normal default networking behavior.

Docker's Python SDK already supports:

```python
client.containers.create(
    ...,
    network_mode="none",
)
```

The proposed Agents SDK API is:

```python
DockerSandboxClientOptions(
    image="python:3.14-slim",
    network_mode="none",
)
```

with a deliberately narrow type:

```python
network_mode: Literal["none"] | None = None
```

`None` preserves the existing behavior exactly, while `"none"` passes `network_mode="none"` to Docker.

The public option intentionally supports only `"none"` rather than an arbitrary `str`. The purpose of this change is specifically to allow callers to disable networking. Exposing unrelated Docker networking modes such as `host`, `bridge`, or `container:<id>` would unnecessarily broaden a security-sensitive public API.

#### Production / security use case

A common architecture is:

```text
application / worker
        |
        | Docker API
        v
sandbox container
        |
        +-- local input files
        +-- shell commands
        +-- preinstalled dependencies
        +-- no network required
```

The application may need network access for model/API calls while the model-controlled sandbox should have no IP network path to:

- the public Internet;
- host network services;
- other containers;
- private or LAN services;
- metadata services.

For workloads such as document processing, code analysis, and other untrusted-input workflows, this provides a useful defense-in-depth boundary.

This is stronger and simpler than attempting to restrict only Internet egress, because the sandbox does not require networking at all. Docker's `network_mode="none"` directly represents that requirement.

The option remains opt-in. Some Docker sandbox users legitimately require networking, including workloads that use network-dependent tools or in-container remote storage/mount strategies, so changing the default behavior would be backwards incompatible.

#### Resume semantics

The setting also needs to be persisted in `DockerSandboxSessionState`.

`DockerSandboxClient.resume()` can recreate a missing container from serialized session state. If the network setting existed only in `DockerSandboxClientOptions`, a sandbox originally created with networking disabled could later be recreated with Docker's default networking.

The session state therefore includes:

```python
class DockerSandboxSessionState(SandboxSessionState):
    ...
    network_mode: Literal["none"] | None = None
```

and replacement-container creation reuses `state.network_mode`.

This also preserves compatibility with older serialized session states because the new field defaults to `None`.

#### Compatibility

Existing callers remain unchanged.

For example:

```python
DockerSandboxClientOptions(
    image="python:3.14-slim",
    exposed_ports=(8080,),
)
```

retains the current behavior.

The new `network_mode` argument is optional and keyword-only, so the existing positional meaning of:

```python
DockerSandboxClientOptions(image, exposed_ports)
```

is preserved.

When `network_mode` is omitted, `_create_container()` does not add a `network_mode` entry to the Docker create kwargs. This preserves the existing Docker container creation behavior rather than explicitly selecting a different Docker network mode.

The current minimum Docker SDK dependency is `docker>=6.1`, which already supports `network_mode="none"`, so this does not require a Docker SDK dependency change.

#### Exposed ports

`network_mode="none"` and `exposed_ports` remain independent Docker configuration inputs at the SDK layer.

This change does not introduce an additional Agents SDK validation rule for that combination. When networking is disabled, Docker remains authoritative for the resulting container configuration and behavior.

This keeps the feature focused on exposing the existing Docker no-network capability without introducing additional policy around port configuration.

#### Implementation scope

The change is intentionally focused:

1. Add `network_mode: Literal["none"] | None = None` to `DockerSandboxClientOptions`.
2. Persist the setting in `DockerSandboxSessionState`.
3. Pass `network_mode="none"` to Docker container creation when explicitly configured.
4. Omit the Docker `network_mode` kwarg entirely when the option is not configured.
5. Reapply the persisted setting when `resume()` needs to recreate a missing container.
6. Preserve compatibility with serialized session states that predate the new field.
7. Add focused regression coverage for configuration, Docker container creation, serialization, backwards compatibility, and resume/recreation.

#### API shape

The API is intentionally:

```python
network_mode: Literal["none"] | None = None
```

rather than a generic Docker network-mode string.

This maps directly to Docker's existing `network_mode="none"` semantics while keeping the Agents SDK surface limited to the capability this feature is intended to provide: explicitly creating a Docker sandbox without networking.

`None` means that the SDK does not specify a Docker network mode and therefore preserves the existing behavior. `"none"` explicitly requests a sandbox with networking disabled.

A focused implementation with regression coverage is available and follows this API shape, including preservation of the setting across container recreation during `DockerSandboxClient.resume()`.

koadegno:
### Summary

This pull request adds an opt-in way to create Docker sandboxes with networking completely disabled.

`DockerSandboxClientOptions` now accepts:

```python
DockerSandboxClientOptions(
    image="python:3.14-slim",
    network_mode="none",
)
```

The new option is deliberately narrow:

```python
network_mode: Literal["none"] | None = None
```

When omitted, Docker container creation remains unchanged and no `network_mode` argument is passed.

When set to `"none"`, the value is forwarded to `docker_client.containers.create()`, causing Docker to create the sandbox without networking.

The setting is also persisted in `DockerSandboxSessionState` and reapplied when `DockerSandboxClient.resume()` needs to recreate a missing container. This prevents a sandbox originally created without networking from silently regaining Docker's default network configuration after recreation.

Older serialized session states remain compatible because the new field defaults to `None`.

The public option intentionally does not expose arbitrary Docker networking modes such as `host`, `bridge`, or `container:<id>`.

### Test plan

Added regression coverage for:

- accepting `network_mode="none"`
- rejecting unsupported network modes
- propagating `"none"` through the public `DockerSandboxClient.create()` path
- passing `network_mode="none"` to Docker container creation
- preserving the existing Docker create kwargs when the option is omitted
- `DockerSandboxClientOptions` serialization and parsing
- `DockerSandboxSessionState` serialization
- backwards compatibility with serialized states that do not contain `network_mode`
- reapplying `network_mode="none"` when `resume()` recreates a missing container
- compatibility guards for the extended public options/state fields

Focused sandbox tests passed:

```text
86 passed
```

Full repository verification was run with:

```bash
env UV_DEFAULT_INDEX=https://pypi.org/simple \
  bash .agents/skills/code-change-verification/scripts/run.sh
```

All verification steps passed, including:

- formatting
- lint
- type checking
- full test suite

### Issue number

Closes #4451

### Checks

- [x] If using Codex, I've run `/review` before submitting this PR (not applicable; this change was not prepared with Codex)

### DIFF
--- a/src/agents/run_state.py
+++ b/src/agents/run_state.py
@@ -209,7 +209,10 @@ SCHEMA_VERSION_SUMMARIES: dict[str, str] = {
         "Persists canonical tool invocation identity plus sanitized mount authority and trusted "
         "rebind metadata, durable pending input, and resumable next-model-call state."
     ),
-    "1.16": "Lets an exact call approval decision override a sticky decision for the same tool.",
+    "1.16": (
+        "Persists Docker network-isolation state and lets an exact call approval decision "
+        "override a sticky decision for the same tool."
+    ),
 }
 SUPPORTED_SCHEMA_VERSIONS = frozenset(SCHEMA_VERSION_SUMMARIES)
 
--- a/src/agents/sandbox/sandboxes/docker.py
+++ b/src/agents/sandbox/sandboxes/docker.py
@@ -173,6 +173,7 @@ class DockerSandboxSessionState(SandboxSessionState):
     type: Literal["docker"] = "docker"
     image: str
     container_id: str
+    network_mode: Literal["none"] | None = None
 
     def _sanitize_persisted_provider_identity(
         self,
@@ -193,18 +194,21 @@ class DockerSandboxClientOptions(BaseSandboxClientOptions):
     type: Literal["docker"] = "docker"
     image: str
     exposed_ports: tuple[int, ...] = ()
+    network_mode: Literal["none"] | None = None
 
     def __init__(
         self,
         image: str,
         exposed_ports: tuple[int, ...] = (),
         *,
+        network_mode: Literal["none"] | None = None,
         type: Literal["docker"] = "docker",
     ) -> None:
         super().__init__(
             type=type,
             image=image,
             exposed_ports=exposed_ports,
+            network_mode=network_mode,
         )
 
 
@@ -1502,6 +1506,7 @@ class DockerSandboxClient(BaseSandboxClient[DockerSandboxClientOptions]):
                 image,
                 manifest=manifest,
                 exposed_ports=options.exposed_ports,
+                network_mode=options.network_mode,
                 session_id=session_id,
             )
             container.start()
@@ -1516,6 +1521,7 @@ class DockerSandboxClient(BaseSandboxClient[DockerSandboxClientOptions]):
                 snapshot=snapshot_instance,
                 container_id=container_id,
                 exposed_ports=options.exposed_ports,
+                network_mode=options.network_mode,
             )
             inner = DockerSandboxSession(
                 docker_client=self.docker_client,
@@ -1642,6 +1648,7 @@ class DockerSandboxClient(BaseSandboxClient[DockerSandboxClientOptions]):
                     state.image,
                     manifest=state.manifest,
                     exposed_ports=state.exposed_ports,
+                    network_mode=state.network_mode,
                     session_id=replacement_session_id,
                 )
                 container_id = container.id
@@ -1675,6 +1682,7 @@ class DockerSandboxClient(BaseSandboxClient[DockerSandboxClientOptions]):
         *,
         manifest: Manifest | None = None,
         exposed_ports: tuple[int, ...] = (),
+        network_mode: Literal["none"] | None = None,
         session_id: uuid.UUID | None = None,
     ) -> Container:
         if manifest is not None:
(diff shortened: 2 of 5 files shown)


seratch (changes requested):
Thanks for the contribution. The need described in #4451 is demonstrated, and the narrow Docker-specific `network_mode="none"` design is sound. Before merging, please make three focused changes:

- Append `network_mode` after the existing keyword-only `type` parameter to preserve public constructor ordering.
- Update the current unreleased RunState 1.16 summary to include the persisted Docker network-isolation state.
- Add a deserialization test for an invalid persisted value such as `bridge`, asserting rejection before any Docker provider access.

These changes keep the released API ordering and durable-state contract explicit while proving the none-only security boundary. Once they are addressed and CI is green, the PR should be ready for another review.

seratch (changes requested):
Thanks for the updates. The current head resolves the earlier constructor-ordering, RunState schema-summary, and persisted-value validation requests.

Can you continue working until the PR review says 👍 ?

seratch (approved):

PR #4457, opened 2026-08-16
fix(realtime): truncate audio at zero elapsed time

### LINKED ISSUE #4456: Realtime interrupt skips conversation.item.truncate when it arrives within one Windows timer tick (elapsed_ms == 0)
`_interrupt_audio_playback` gates the whole truncation path on `elapsed_ms > 0`, and `elapsed_ms` is wall-clock derived. On a Windows host at the default timer resolution that value is exactly `0.0` for the first ~16 ms after audio starts, so an interrupt in that window silently skips `conversation.item.truncate`.

## The path

`src/agents/realtime/openai_realtime.py:995`

```python
elapsed_ms = (time.monotonic() - audio_state.initial_received_time) * 1000
```

and the gate at `:1072`

```python
if elapsed_ms > 0:
    ...              # emits audio-interrupted, computes truncated_ms,
                     # and sends conversation.item.truncate
```

`time.monotonic()` on Windows is `GetTickCount64()`:

```
monotonic impl      : GetTickCount64()
monotonic resolution: 0.015625 s
identical back-to-back reads: 20000/20000
```

So two reads inside the same tick are the *same number*, and the subtraction is exactly zero.

## Measured on the real object, not in the abstract

```python
m = OpenAIRealtimeWebSocketModel()
m._audio_state_tracker.set_audio_format("pcm16")
m._audio_state_tracker.on_audio_delta("item_1", 0, b"\x00" * 4800)
m._get_playback_state()
```

```
playback_state elapsed_ms : 0.0
current_item_id           : item_1
-> the guard is  if elapsed_ms > 0:  so this path is SKIPPED

iterations until elapsed_ms > 0: 9916 | wall: 16.0 ms
```

16.0 ms is one `GetTickCount64` tick.

## Why this matters beyond the tests

Truncation is what keeps the model's conversation state in step with what the user actually heard. When it is skipped, the item keeps its full audio, so the model believes it said more than was played — which is the exact condition truncation exists to prevent. A barge-in that lands promptly is *more* likely to hit this window, not less, so the failure is biased toward the responsive case.

Note the guard is `> 0` rather than `>= 0`: zero elapsed is treated as "nothing to truncate", when on this platform it means "less than one tick has been measurable".

## Four tests are load-bearing on ambient timer resolution

On `main` at 9aba9002, `uv run pytest tests/realtime/test_openai_realtime.py::TestSendEventAndConfig -p no:randomly`:

```
FAILED ... ::test_normal_interrupt_targets_response_and_interrupts_audio        - assert 0 == 1
FAILED ... ::test_interrupt_respects_auto_cancellation_when_not_forced          - assert 0 == 1
FAILED ... ::test_interrupt_force_cancel_overrides_auto_cancellation
FAILED ... ::test_interrupt_honors_falsy_present_session_auto_cancellation
4 failed, 58 passed
```

Each seeds audio and interrupts on the next statement, so each assumes the clock advanced in between.

**These pass in CI** — `tests-windows` is green on `main`. I take that to mean the hosted runner is running with the system timer raised (some process calling `timeBeginPeriod`), which is common on Windows CI. That is the uncomfortable part: whether this code path runs at all depends on a **global system setting any other process on the machine can change**, and the suite passes or fails accordingly. It is not a property of the code under test.

## Possible directions

Not proposing one, since the right call depends on how you want the clock treated:

1. **Inject the clock.** The tests could then advance it explicitly and assert the truncate deterministically, on every platform.
2. **Use a higher-resolution source** — `time.perf_counter()` is ~100 ns on Windows and monotonic, and this is a duration measurement, not a timestamp.
3. **Reconsider `> 0`.** If the intent is "skip when no audio has played", zero-elapsed-but-audio-present is not that case.

1 and 3 are independent and both look worth doing; 2 alone would make the window much smaller without removing the class.

Happy to send a PR for whichever shape you prefer — I would default to 2 plus 3, and leave clock injection to you since it touches the test contract.

## Environment

Windows 11, Python 3.12.10, `main` at 9aba9002. `uv sync` only (the `dev` group cannot install on this host — see #4442), so voice/litellm extras are absent; the failures above are unrelated to those, and reproduce with the four named tests alone.

Chirag6722:
On Windows, `time.monotonic()` can return the same value for consecutive reads. When an interrupt arrived within the same clock tick as the first audio delta, `_interrupt_audio_playback` calculated `elapsed_ms == 0` and skipped both `audio_interrupted` and `conversation.item.truncate`.

This change treats zero elapsed time as a valid truncation position. The model now emits `audio_interrupted` and sends `conversation.item.truncate` with `audio_end_ms=0`, keeping the server-side conversation aligned with the audio the user actually heard.

The existing behavior is preserved for negative elapsed values, missing playback state, completed audio, and truncation clamping.

## Tests

- Added a deterministic regression test with equal monotonic readings that verifies `audio_interrupted` and `conversation.item.truncate` with `audio_end_ms=0`.
- Added a negative-elapsed control test to ensure a backwards clock value still skips truncation.
- Removed the five-second test backdates, which placed the fixtures beyond their buffered audio duration.
- Verified the focused `TestSendEventAndConfig` suite and the complete Realtime test suite.

Fixes #4456

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -1070,7 +1070,7 @@ class OpenAIRealtimeWebSocketModel(RealtimeModel):
             )
         else:
             current_item_content_index = current_item_content_index or 0
-            if elapsed_ms > 0:
+            if elapsed_ms >= 0:
                 if not response_scoped:
                     try:
                         await self._emit_event(
--- a/tests/realtime/test_openai_realtime.py
+++ b/tests/realtime/test_openai_realtime.py
@@ -18,6 +18,7 @@ from agents.realtime import RealtimeSessionModelSettings
 from agents.realtime.model import RealtimeModelConfig, RealtimePlaybackTracker
 from agents.realtime.model_events import (
     RealtimeModelAudioEvent,
+    RealtimeModelAudioInterruptedEvent,
     RealtimeModelErrorEvent,
     RealtimeModelOutputTextDeltaEvent,
     RealtimeModelRawServerEvent,
@@ -1243,14 +1244,6 @@ class TestSendEventAndConfig(TestOpenAIRealtimeWebSocketModel):
         """Interrupt should send response.cancel even when auto cancel is enabled."""
         model._audio_state_tracker.set_audio_format("pcm16")
         model._audio_state_tracker.on_audio_delta("item_1", 0, b"\x00" * 4800)
-        # Backdate the audio start so elapsed time is measurable regardless of host
-        # clock resolution: time.monotonic() advances in ~15.6ms steps on Windows, so
-        # a delta read within the same tick is exactly 0 and the truncate path under
-        # `if elapsed_ms > 0` never runs. Same idiom as
-        # test_speech_started_skips_truncate_when_audio_complete.
-        _audio_state = model._audio_state_tracker.get_state("item_1", 0)
-        assert _audio_state is not None
-        _audio_state.initial_received_time = time.monotonic() - 5
         await model._mark_response_created()
         model._created_session = SimpleNamespace(
             audio=SimpleNamespace(
@@ -1292,14 +1285,6 @@ class TestSendEventAndConfig(TestOpenAIRealtimeWebSocketModel):
 
         model._audio_state_tracker.set_audio_format("pcm16")
         model._audio_state_tracker.on_audio_delta("item_1", 0, b"\x00" * 4800)
-        # Backdate the audio start so elapsed time is measurable regardless of host
-        # clock resolution: time.monotonic() advances in ~15.6ms steps on Windows, so
-        # a delta read within the same tick is exactly 0 and the truncate path under
-        # `if elapsed_ms > 0` never runs. Same idiom as
-        # test_speech_started_skips_truncate_when_audio_complete.
-        _audio_state = model._audio_state_tracker.get_state("item_1", 0)
-        assert _audio_state is not None
-        _audio_state.initial_received_time = time.monotonic() - 5
         await model._mark_response_created()
         model._created_session = FalsySession.model_construct(
             type="realtime",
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
Thanks for taking the time to investigate this and for clearly isolating the Windows-specific test behavior. I agree that these tests should not depend on the host's ambient clock resolution.

After tracing the runtime path, I believe the test adjustment and the underlying runtime behavior should be addressed together in this PR. The `elapsed_ms == 0` case from #4456 is a valid state that can occur when an interrupt arrives within the same clock tick. Backdating the audio start by five seconds avoids that case, but it also places each 100 ms audio fixture beyond its buffered duration. As a result, the ongoing-response and clamping path can make the tests pass without exercising active playback.

Could you please revise the PR to treat `elapsed_ms == 0` as a valid truncation position in `_interrupt_audio_playback`, add a deterministic regression test with equal monotonic readings that asserts both `audio_interrupted` and `conversation.item.truncate` with `audio_end_ms=0`, and remove the four five-second backdates? Please preserve the existing behavior for negative elapsed values, missing audio, completed audio, and truncation clamping.

Thank you again for the careful report and reproduction. With those focused changes, this should cover both the Windows test failure and the user-visible runtime issue.

Chirag6722:
Thanks — you were right, and the reason is worse than "the backdate is unnecessary". Revised in 9e9c2bb2.

Your point about the fixture is the one that matters: each fixture buffers 4,800 bytes of PCM16, so a five-second backdate put playback past its 100ms duration. Combined with `_ongoing_response`, the tests then satisfied themselves through the clamping path. So they were **green without active playback ever running** — I had made them deterministic and simultaneously stopped them testing the thing they are named for. That is a worse defect than the flakiness I set out to fix, and I would not have found it from the failure output.

## Changes

**Runtime** — `_interrupt_audio_playback` now gates on `elapsed_ms >= 0`.

One detail that supports the reading: the `else` branch already logged `"Didn't interrupt bc elapsed ms is < 0"`. The message described `>= 0` semantics while the guard did something narrower, so it is now accurate rather than aspirational. `truncated_ms` already went through `max(int(elapsed_ms), 0)`, so zero needed no further handling.

Preserved, as asked: negative elapsed still skips; a missing item id or elapsed value still skips; the completed-audio and `max_audio_ms` clamping paths are untouched.

**Tests** — the four backdates are gone. Two deterministic tests replace them, both pinning the recording and reading clocks to the same value:

- `test_interrupt_truncates_at_zero_when_no_time_has_elapsed` — asserts exactly one `RealtimeModelAudioInterruptedEvent` and exactly one `conversation.item.truncate` with `audio_end_ms == 0`.
- `test_interrupt_skips_truncate_when_elapsed_is_negative` — the paired control, so `>= 0` cannot silently become "always truncate".

## Verification

The four originally-failing tests now pass **with no backdate at all**, because the runtime fix makes them work for the right reason:

```
TestSendEventAndConfig    64 passed   (was 4 failed / 58 passed)
tests/realtime/          506 passed
ruff check / format / mypy (changed files)   clean
```

Full suite against baseline, diffing the `FAILED` lists rather than the totals: **2 fixed, 0 regressions.**

One thing to flag so it is not mistaken for mine: the after-run shows `tests/test_trace_processor.py::test_backend_span_exporter_deadline_stops_during_5xx_retry_backoff` failing, which the baseline did not. It passes 3/3 in isolation with my change and 2/2 on clean `main`, so it is an xdist load flake rather than a regression. I saw the same shape from `tests/test_stream_events.py::test_stream_events_main` earlier. Happy to file either separately if they are not already known.

#4456 can close with this, since it now covers the runtime half rather than only the test half.

seratch (approved):

PR #4458, opened 2026-08-16
fix(realtime): end session iteration when the server closes normally

ayaangazali:
### Summary

When the Realtime server closes the websocket normally, the application hangs with no error.

`_listen_for_messages` reports every other exit. An abnormal close emits a `RealtimeModelExceptionEvent`, which the session stores and raises out of `__aiter__`. A normal close only logs (`openai_realtime.py:749`):

```python
except websockets.exceptions.ConnectionClosedOK:
    # Normal connection closure - no exception event needed
    logger.debug("WebSocket connection closed normally")
```

Nothing else puts to `RealtimeSession._event_queue`, and `__aiter__` is parked on `await self._event_queue.get()`. The only thing that wakes it is `_wake_event_iterators()`, reachable only from `close()`. So after a clean server-side close, `async for event in session` waits on an event that can never arrive.

Reproduced on `main` with a websocket that hangs up right after the session is established, once per closure type:

```
CLEAN   (ConnectionClosedOK):    consumer STILL WAITING after 1.5s, events=['history_updated']
UNCLEAN (ConnectionClosedError): consumer raised ConnectionClosedError
```

Same server, same session, and the well-behaved shutdown is the one that hangs. A realtime server closing cleanly on an idle timeout, a session ending, or a deploy drain is ordinary, so this is not limited to network faults.

**The fix keeps a normal closure a normal closure.** No exception is invented for it, which is what the existing comment is right about. Instead the model emits `RealtimeModelConnectionStatusEvent(status="disconnected")`, an event type that is already defined and exported but never emitted, and the session ends iteration when it sees one. `close()` clears `_websocket` before cancelling the listener task, so this only reports a server-initiated close, not our own teardown.

Closing the session stays the caller's job. The session records that the transport is gone and lets `__aiter__` drain whatever is already queued and then return, rather than marking itself closed, so a later `await session.close()` still releases the model and background tasks normally.

Worth a maintainer opinion, since it picks semantics on a previously unused public event: I read `RealtimeModelConnectionStatusEvent` as the intended signal here, and `session.py` already had a `connection_status` branch waiting with `pass`. If you would rather the session auto-close, surface a distinct session-level event, or emit `connecting`/`connected` too, say so and I will redo it that way.

### Test plan

- `test_normal_server_close_ends_session_iteration` drives a real `RealtimeSession` over a patched `websockets.connect` whose server raises `ConnectionClosedOK`, and asserts iteration finishes. Against `main` it fails with `TimeoutError`; the `asyncio.wait_for` is there so a regression fails in five seconds instead of hanging CI.
- `uv run pytest tests/realtime/`: 504 passed. The abnormal-close path is unchanged and still raises.

### Issue number

N/A, found while auditing `realtime/session.py` against `.agents/references/realtime-session-lifecycle.md`.

### Checks

---

flagging my own uncertainty: the hang itself I am confident about, I reproduced it both ways before touching anything. Where I am less sure is whether reusing `RealtimeModelConnectionStatusEvent` is the shape you want, since nothing emitted it before and I may be reading intent into dead code. I worked the diagnosis out myself and used Claude Code to pressure test the design options. happy to rework it, college freshman and still learning where these lines get drawn :)

### DIFF
--- a/src/agents/realtime/openai_realtime.py
+++ b/src/agents/realtime/openai_realtime.py
@@ -126,6 +126,7 @@ from .model_events import (
     RealtimeModelAudioEvent,
     RealtimeModelAudioInterruptedEvent,
     RealtimeModelCachedTokensDetails,
+    RealtimeModelConnectionStatusEvent,
     RealtimeModelErrorEvent,
     RealtimeModelEvent,
     RealtimeModelExceptionEvent,
@@ -746,8 +747,13 @@ class OpenAIRealtimeWebSocketModel(RealtimeModel):
                     )
 
         except websockets.exceptions.ConnectionClosedOK:
-            # Normal connection closure - no exception event needed
+            # A normal closure is not an error, so no exception event is emitted. Consumers still
+            # have to learn the transport ended, otherwise nothing further can ever reach
+            # `RealtimeSession.__aiter__` and it waits forever. `close()` clears `_websocket`
+            # before cancelling this task, so this only reports a server-initiated close.
             logger.debug("WebSocket connection closed normally")
+            if self._websocket is not None:
+                await self._emit_event(RealtimeModelConnectionStatusEvent(status="disconnected"))
         except websockets.exceptions.ConnectionClosed as e:
             await self._emit_event(
                 RealtimeModelExceptionEvent(
--- a/src/agents/realtime/session.py
+++ b/src/agents/realtime/session.py
@@ -227,6 +227,7 @@ class RealtimeSession(RealtimeModelListener):
         self._event_iterator_waiters = 0
         self._closing = False
         self._closed = False
+        self._transport_disconnected = False
         self._cleanup_task: asyncio.Task[None] | None = None
         self._stored_exception: BaseException | None = None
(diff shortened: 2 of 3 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying clean-close hang is real, and using the existing disconnected status to end session iteration is directionally right. However, the current patch does not handle the real WebSockets 15 path: `ClientConnection.__aiter__` catches `ConnectionClosedOK` internally and returns normally, so the new `except ConnectionClosedOK` block is not reached. The added mock raises that exception directly and therefore passes while an actual local code-1000 close still leaves session iteration blocked.

Please emit the disconnected status when the message loop ends normally, while preserving the distinction from caller-initiated `close()` and the existing abnormal-close exception behavior. Please also cover this with a real local WebSocket clean-close regression test that fails on the base and completes on the updated branch.

seratch:
Thanks again for sending this PR. https://github.com/openai/openai-agents-python/pull/4461 resolved the issue and the PR includes you as a co-author.

PR #4459, opened 2026-08-16
feat: opt-in strip_defaults for ensure_strict_json_schema (non-null defaults rejected by some providers)

azrabano23:
### Summary
Follow-up to #4390 and the discussion on #4391. Strict conversion forces every property into `required`, so a property-level `default` can never take effect at runtime — yet `ensure_strict_json_schema()` only strips defaults whose value is exactly `None`. Some providers reject strict schemas that still carry a non-null `default` (reported against Azure OpenAI deployments in #4390: a literal `'default' is not permitted within a property definition` rejection, plus a measured intermittent-500 correlation), while the official OpenAI API accepts them.

Since unconditionally removing defaults was deemed not right for real-world use cases (#4391), this PR takes the minimal opt-in route instead: a keyword-only `strip_defaults: bool = False` flag on `ensure_strict_json_schema()`.

- **Default behavior is completely unchanged** — non-null defaults remain visible to the model, and a test now pins that behavior.
- With `strip_defaults=True`, every remaining `default` is removed. The strip runs after `$ref` handling so a non-null `default` sibling still triggers ref expansion instead of leaving a pure `$ref` behind.
- Affected users (e.g. Azure) get an escape hatch without disabling `strict_mode` entirely. If maintainers want, a follow-up could thread this through `function_tool(...)`/`FunctionTool` — kept out of this PR to stay minimal.

### Tests
6 new tests in `tests/test_strict_schema.py`: current-behavior pin, property/nested-`$defs`/`anyOf` stripping, `None`-default parity, and the `$ref`-with-`default`-sibling expansion case. `ruff` and `mypy` clean; full `test_strict_schema.py` + `test_function_schema.py` suites pass (112 passed).

Addresses #4390.

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -98,9 +98,12 @@ _ROOT_REF_NON_CONSTRAINING_SIBLINGS = _REF_NON_CONSTRAINING_SIBLINGS | {"$id"}
 class _NodeBudget:
     """Tracks conversion state across the recursion."""
 
-    def __init__(self, limit: int, *, reject_open_objects: bool = False) -> None:
+    def __init__(
+        self, limit: int, *, reject_open_objects: bool = False, strip_defaults: bool = False
+    ) -> None:
         self.remaining = limit
         self.reject_open_objects = reject_open_objects
+        self.strip_defaults = strip_defaults
 
     def spend(self) -> None:
         self.remaining -= 1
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for exploring an opt-in alternative. I don't think we should add this API yet. #4390 still lacks a default-isolated Azure reproduction: the reported strict versus non-strict experiment changes the entire schema mode, and the nested Decimal schema remains a confounding variable. Direct OpenAI requests accept the current schema.

This PR also exposes the option only through `ensure_strict_json_schema()`, so the reported `@function_tool` workflow cannot opt in without post-processing `params_json_schema`, which callers can already do directly.

I am going to close this for now. We can reconsider after #4390 provides the exact Azure endpoint, API version, deployment, raw schema and error response, plus otherwise identical with-default and without-default requests. If that confirms the need, we should design the escape hatch at the user-facing tool or provider boundary.

PR #4464, opened 2026-08-16
fix(models): reject a provider response with no choices in LiteLLM and AnyLLM

ayaangazali:
### Summary

`OpenAIChatCompletionsModel` rejects a completion with no choices (`openai_chatcompletions.py:262`):

```python
if not response.choices:
    provider_error = getattr(response, "error", None)
    error_details = f": {provider_error}" if provider_error is not None else ""
    raise ModelBehaviorError(
        f"ChatCompletion response has no choices (possible provider error payload)"
        f"{error_details}"
    )
```

That guard arrived in #2850 for #604 and only ever landed in that one adapter. `LitellmModel` (`litellm_model.py:255`) and `AnyLLMModel` (`any_llm_model.py:611`) still take the other branch: `message` stays `None`, `items` comes out empty, and the caller gets a successful `ModelResponse` with no output.

Reproduced on `main` against LiteLLM with a provider error payload:

```
LiteLLM  -> returned ModelResponse, output = []  usage.requests = 1
```

No exception, and an empty turn is indistinguishable from a model that legitimately said nothing. The run keeps going on a turn that never happened, which usually surfaces later as an empty final output or a max-turns loop rather than as the provider error it actually was. These two adapters are the ones pointed at OpenAI-compatible gateways, so an error object plus an empty `choices` list is exactly the shape they meet in practice.

This adds the same guard to both, including the provider error object in the message when one is present. Nothing else changes: a genuinely empty-but-well-formed choice list of length one still flows through the existing `message is None` path and its debug log.

Both adapters were already brought in line on request accounting in #4290 and #4453, so this is the same three-adapter set, one behavior short.

### Test plan

- `test_provider_error_payload_without_choices_raises` in `tests/models/test_litellm_content_filter.py` and in `tests/models/test_any_llm_model.py`, each driving `get_response` against a mocked provider that returns an error object with `choices=[]`. Both fail against `main` and pass here.
- `uv run pytest tests/models/`: 823 passed.

### Issue number

Same defect as #604, which was fixed for the OpenAI adapter only in #2850.

### Checks

---

one thing I am unsure about and would rather flag than hide: I kept the wording close to the OpenAI adapter's message so the three read the same, but LiteLLM normalizes a lot of providers, so if you would rather it stayed permissive there and only AnyLLM got the guard, say so and I will cut it back. found this comparing the three adapters after #4453, worked it through myself with Claude Code as a sounding board. freshman in college, still learning where consistency helps and where it just adds noise :)

### DIFF
--- a/src/agents/extensions/models/any_llm_model.py
+++ b/src/agents/extensions/models/any_llm_model.py
@@ -607,8 +607,14 @@ class AnyLLMModel(Model):
             )
 
             if not response.choices:
+                # The provider error object can carry prompt text or credentials, so it is only
+                # attached when model-data logging is explicitly enabled.
                 provider_error = getattr(response, "error", None)
-                error_details = f": {provider_error}" if provider_error is not None else ""
+                error_details = (
+                    f": {provider_error}"
+                    if provider_error is not None and not _debug.DONT_LOG_MODEL_DATA
+                    else ""
+                )
                 raise ModelBehaviorError(
                     f"Chat completion response has no choices (possible provider error payload)"
                     f"{error_details}"
--- a/src/agents/extensions/models/any_llm_model.py
+++ b/src/agents/extensions/models/any_llm_model.py
@@ -606,6 +606,14 @@ class AnyLLMModel(Model):
                 prompt=prompt,
             )
 
+            if not response.choices:
+                provider_error = getattr(response, "error", None)
+                error_details = f": {provider_error}" if provider_error is not None else ""
+                raise ModelBehaviorError(
+                    f"Chat completion response has no choices (possible provider error payload)"
+                    f"{error_details}"
+                )
+
             message: ChatCompletionMessage | None = None
             first_choice: Choice | None = None
             if response.choices:
--- a/src/agents/extensions/models/litellm_model.py
+++ b/src/agents/extensions/models/litellm_model.py
@@ -251,8 +251,14 @@ class LitellmModel(Model):
             )
 
             if not response.choices:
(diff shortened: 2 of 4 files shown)


ayaangazali:
Good catch, and it applies to my own diff rather than only the pre-existing block. Fixed in d47295b6.

The provider `error` object can carry prompt text or credentials, and interpolating it into `ModelBehaviorError` puts it in `str(exc)` and in ordinary application exception logging, which is not covered by tracing's data controls.

`_debug.DONT_LOG_MODEL_DATA` defaults to `True` in this SDK, so the payload is now attached only when a user has explicitly turned model-data logging on. That matches how both adapters already gate their own response logging a few lines below, so there is one convention rather than a new policy. By default the message is just `... has no choices (possible provider error payload)`.

Both regression tests now also assert the payload does not appear in `str(exc)`.

On the identical block in `openai_chatcompletions.py`: you are right that it has the same shape, but it predates this change and is on a third adapter, so I have left it alone rather than widening this PR. Happy to send it separately, or to include it here if you would rather the three stay identical.

seratch:
Thanks for investigating this and for addressing the provider-payload redaction concern. I do not think we should merge the current scope. `choices=[]` in LiteLLM is already a released, intentional path from #1981, motivated by Gemini filtered or empty responses reported in #1559; it is not evidence of a provider-error payload. This PR would reclassify every such response as `ModelBehaviorError`, while the AnyLLM coverage is synthetic and does not establish a real provider path.

PR #4465, opened 2026-08-16
fix(core): close all cached websocket models when one close fails

feiiiiii5:
## Problem

`OpenAIProvider._close_models` and the threadsafe branch of `_close_ws_models_for_loop` (`src/agents/models/openai_provider.py`) awaited each `model.close()` in a bare loop, so the first failing close aborted cleanup and leaked every remaining cached websocket model. `OpenAIProvider.aclose()` had the same shape one level up: a loop whose close failed skipped closing the remaining event loops' models and left their `_ws_model_cache_by_loop` entries behind.

This is the same cleanup contract `MultiProvider.aclose()` landed with #4438 — close all children, preserve the first failure — applied to the websocket model cache paths.

## Changes

- `_close_models`: close every model even when one `close()` raises, then raise the first error (`CancelledError` still propagates immediately).
- `_close_ws_models_for_loop` threadsafe branch: same per-model error collection around `run_coroutine_threadsafe(...)`/`wrap_future`.
- `aclose()`: a failing loop no longer prevents the remaining loops from being closed and their cache entries cleared; the first error is raised after the sweep.

Behavior on the single-model / all-healthy paths is unchanged.

## Testing

New `tests/models/test_openai_provider_ws_close.py` (hermetic fake models):

- first model's `close()` raises → the second model is still closed, and the first error is raised (fails on `main`: the second model leaks)
- second model also failing → both closed, first error raised
- all healthy → all closed
- `_close_ws_models_for_loop` same-loop branch → resilient close

```
$ pytest tests/models/test_openai_provider_ws_close.py -q
4 passed

$ pytest tests/models/ -q        # full models suite
816 passed, 11 skipped

$ ruff check && ruff format --check   # repo-pinned ruff 0.9.2
All checks passed / 2 files already formatted
```

### DIFF
--- a/src/agents/models/openai_provider.py
+++ b/src/agents/models/openai_provider.py
@@ -170,17 +170,38 @@ class OpenAIProvider(ModelProvider):
             await self._close_models(models)
             return
         if loop.is_running():
+            first_error: Exception | None = None
             for model in models:
                 future = asyncio.run_coroutine_threadsafe(model.close(), loop)
-                await asyncio.wrap_future(future)
+                try:
+                    await asyncio.wrap_future(future)
+                except asyncio.CancelledError:
+                    raise
+                except Exception as error:
+                    if first_error is None:
+                        first_error = error
+            if first_error is not None:
+                raise first_error
             return
         # Do not run an inactive foreign loop on another thread. This also covers closed loops.
         # Close from the current loop and rely on model-specific cross-loop cleanup fallbacks.
         await self._close_models(models)
 
     async def _close_models(self, models: list[Model]) -> None:
+        # Close every model even when one close() fails, then raise the first
+        # failure — a raise mid-loop would leak the remaining cached websocket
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this cleanup path. I traced the websocket cache through `OpenAIProvider.get_model()` and `OpenAIResponsesWSModel.close()`. The cache contains `OpenAIResponsesWSModel` instances, and their `close()` implementation already catches ordinary websocket-close failures, force-aborts the transport, and clears the connection state.

The added tests instead place arbitrary fake `Model` instances into the private helper path, so they demonstrate that the helper can be made to fail but do not reproduce a supported provider/cache scenario. This differs from #4438, where `MultiProvider` owns arbitrary child providers whose public `aclose()` implementations may legitimately raise.

I am going to close this PR. If there is a reproduction using the real websocket cache where a non-cancellation exception escapes one model close and prevents another cached connection from being released, we can revisit the narrow failure path from that evidence.

Issue #4434, opened 2026-08-15
Compaction tokens are billed but never counted: run usage omits every responses.compact call


arthi-arumugam-git:
`OpenAIResponsesCompactionSession.run_compaction` makes a real, billed model call and then drops its usage on the floor.

https://github.com/openai/openai-agents-python/blob/9aba9002935b56d7253d28d99b2a8bca2d2450a7/src/agents/memory/openai_responses_compaction_session.py#L227

```python
compacted = await self.client.responses.compact(**compact_kwargs)

output_items = _strip_orphaned_assistant_ids(
    _normalize_compaction_output_items(compacted.output or [])
)
```

Only `compacted.output` is read. `compacted.usage` is a required field on the response, documented as token accounting for the compaction pass, and the string `usage` does not appear anywhere in that module. Those tokens never reach `context_wrapper.usage`, so `RunResult` reports a total that excludes them, and the docs say usage is aggregated across all model calls during the run.

### Why it matters

Compaction runs on long conversations, which is exactly where the compacted history is large, so the missing call is often the most expensive one in the run. In a scripted 12-turn run with one compaction pass, the run reports `requests=12` and `total_tokens=1800` while the compaction pass itself consumed 192,000 tokens. The reported total is under one percent of what was billed, and `requests` undercounts by one per pass.

Direction of the error is always the same: reported usage is too low, and it silently diverges further the longer the conversation runs. Nothing raises.

### Reproduction

No network or API key. Construct the session with a stubbed client whose `responses.compact` returns a response carrying `usage`, run a compaction pass, and inspect any usage the session exposes: there is none, and `context_wrapper.usage` from the run is unchanged.

### Duplicate check

I read the six open and closed compaction issues (#2244, #2317, #2333, #2343, #2727, #3116) and the twelve compaction PRs before filing; they cover latency, ordering, stateless mode, reasoning ids and history loss, none of them token accounting. `git log -S "compacted.usage"` is empty, and none of the seven open PRs touches this path. Every existing compaction test stubs the response with `MagicMock()` and sets only `.output`, so nothing asserts the current behavior.

### Fix

A PR follows immediately: record the compaction pass on the session so the tokens are visible instead of discarded, with the accounting written defensively so it can never break a run. Whether that total should also flow into `RunResult.context_wrapper.usage` automatically is an architectural call I would rather leave to you, since the session has no reference to the run context today.

PR #4466, opened 2026-08-16
fix(sessions): count the compaction request when the provider omits usage

ayaangazali:
### Summary

`run_compaction` makes a real, billed `responses.compact` call, then only records it when the response carries usage (`openai_responses_compaction_session.py:239`):

```python
compacted_usage = getattr(compacted, "usage", None)
if wrapper is not None and compacted_usage is not None:
    wrapper.usage.add(_response_usage_to_usage(compacted_usage))
```

If the provider reports no usage, the whole request disappears from `Usage.requests`, not just its token counts. That is the accounting #4446 set out to fix for #4434, and it is the same gap #4453 closed for the Responses adapter, where the rule landed as "the request completed, so it counts even when the provider omits usage."

Reproduced on `main` with one model call plus one compaction whose response has `usage=None`:

```
result.context_wrapper.usage.requests == 1   # expected 2
```

The compaction call happened and will be billed, but the run reports a single request. `OpenAIResponsesCompactionSession` takes a caller-supplied `client`, so this is the same OpenAI-compatible-endpoint population that motivated #4290 and #4453, and the code already anticipates the shape by checking `usage is not None` before using it.

This records one request when the payload is absent. Token counts stay at zero rather than being synthesized, so nothing reports numbers that never arrived, and the existing path is unchanged whenever usage is present.

### Test plan

- `test_compaction_request_counts_when_provider_omits_usage`, placed next to the existing `test_compaction_runs_during_runner_flow` so the two read as a pair: same Runner flow, same mocked `responses.compact`, one with a usage payload and one without. It fails against `main` with `assert 1 == 2` and passes here.
- `uv run pytest tests/memory/`: 204 passed.

### Issue number

Same accounting as #4434, which #4446 fixed for the case where the provider does report usage.

### Checks

---

small note in case it reads as nitpicking a fix that just landed: I only went looking here because #4453 taught me this exact shape, and I checked the compaction path last since it is the one billed call that does not go through a model adapter. if you would rather compaction stayed out of `requests` entirely, that is a reasonable call too and I will close this. freshman in college, still calibrating what counts as worth a PR :)

### DIFF
--- a/src/agents/memory/openai_responses_compaction_session.py
+++ b/src/agents/memory/openai_responses_compaction_session.py
@@ -11,7 +11,7 @@ from ..items import TResponseInputItem
 from ..logger import log_model_and_tool_action_warning
 from ..models._openai_shared import get_default_openai_client
 from ..run_internal.items import normalize_input_items_for_api
-from ..usage import _response_usage_to_usage
+from ..usage import Usage, _response_usage_to_usage
 from .openai_conversations_session import OpenAIConversationsSession
 from .session import (
     OpenAIResponsesCompactionArgs,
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the focused contribution. The implementation is narrow, but the remaining question is whether this case occurs on a supported, realistic path. The official `responses.compact` response includes a required usage payload, and the test currently proves only that a manually constructed usage-less object can reach this branch. The evidence behind #4453 covers ordinary Responses calls, not the compaction endpoint.

Before we extend that behavior to compaction, please provide a concrete provider or gateway reproduction that successfully implements `responses.compact` while omitting usage, together with the resulting user impact. Without such evidence, I do not think this additional SDK behavior is justified, and we should close this PR for now.

ayaangazali:
That is a fair bar and I cannot meet it, so I am closing this.

My reasoning was that `run_compaction` already checks `usage is not None` before using it, so the shape looked anticipated. That is defensive coding in the SDK, not evidence of a real endpoint that behaves that way, and you are right that I did not distinguish the two. The official `responses.compact` response carries a required usage payload, and I have no concrete gateway that implements `compact` while omitting it. Extending SDK behavior on a hand-built object is not worth the surface.

The same mistake is what sank #4464, so thank you for both. The rule I am taking away is that a divergence being real is not the same as it being unintended, and a constructed payload is not a provider path.

PR #4469, opened 2026-08-17
fix(sandbox): support move-only apply patches

### LINKED ISSUE #4468: apply_patch grammar accepts empty updates that the runtime parser rejects
The custom apply_patch grammar allows an Update File operation containing only a rename:

```python
# src/agents/sandbox/capabilities/tools/apply_patch_tool.py
update_hunk: "*** Update File: " filename LF change_move? change?
```

Because change is optional, the tool grammar tells the LLM that a move-only update is valid:

```
*** Begin Patch
*** Update File: foo.py
*** Move to: bar.py
*** End Patch
```

However, _parse_update_file() rejects it because no diff lines are present:

```
# src/agents/sandbox/capabilities/tools/apply_patch_tool.py
if not diff_lines:
    raise ValueError(f"Update File patch for {path} must include a hunk")
```

This creates a grammar/runtime contract mismatch for a legitimate file operation and wastes a tool call.

> Note: This issue was identified by an automated testing tool for academic research and manually verified. If you have any concerns about this type of reporting, please let me know, and I will adjust my workflow accordingly.

seratch:
his pull request fixes grammar-constrained apply_patch updates that move a file without including content hunks.
It aligns the provider-facing Lark grammar with the plaintext parser, preserves the source payload byte-for-byte during move-only operations, and keeps the released programmatic editor behavior unchanged for updates with a missing diff.

Regression coverage includes the exposed grammar production, header-only update rejection, binary and mixed-newline payload preservation, and direct editor compatibility.

Resolves #4468

### DIFF
--- a/src/agents/sandbox/apply_patch.py
+++ b/src/agents/sandbox/apply_patch.py
@@ -135,6 +135,27 @@ class WorkspaceEditor:
             path=operation.path,
         )
 
+    async def _move_file(self, operation: ApplyPatchOperation) -> ApplyPatchResult:
+        if operation.move_to is None:
+            raise ApplyPatchDiffError(
+                message=f"Missing move destination for path {operation.path}",
+                path=operation.path,
+            )
+
+        relative_path, display_path = self._resolve_path(operation.path)
+        destination = self._session.normalize_path(relative_path)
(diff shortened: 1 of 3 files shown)


seratch:
Closing this in favor of #4470

PR #4470, opened 2026-08-17
fix(sandbox): require apply_patch update hunks

### LINKED ISSUE #4468: apply_patch grammar accepts empty updates that the runtime parser rejects

li2631026381-alt:
### Summary

The custom `apply_patch` grammar currently allows an `Update File` operation without any diff lines, even though the runtime parser rejects that input. This pull request requires at least one change after the optional `Move to` directive and aligns the model-facing EBNF description with the runtime contract.

It also adds regression coverage that:
- compares the complete `update_hunk` production so the old `change?` rule cannot pass
- verifies the converter forwards the validated grammar definition
- confirms empty and move-only updates retain the runtime rejection

### Test plan

- `env NO_PROXY=localhost,127.0.0.1 no_proxy=localhost,127.0.0.1 UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh`
- All formatting, lint, type checking, and test steps passed on Python 3.12.

### Issue number

Closes #4468

### Checks

### DIFF
--- a/src/agents/sandbox/capabilities/tools/apply_patch_tool.py
+++ b/src/agents/sandbox/capabilities/tools/apply_patch_tool.py
@@ -28,7 +28,7 @@ end_patch: "*** End Patch" LF?
 hunk: add_hunk | delete_hunk | update_hunk
 add_hunk: "*** Add File: " filename LF add_line+
 delete_hunk: "*** Delete File: " filename LF
-update_hunk: "*** Update File: " filename LF change_move? change?
+update_hunk: "*** Update File: " filename LF change_move? change
 
 filename: /(.+)/
 add_line: "+" /(.*)/ LF -> line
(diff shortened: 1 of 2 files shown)


seratch (approved):
The inconsistency has existed since the sandbox feature was introduced in v0.14.0: the executable grammar permits a move-only update, while the accompanying description requires one or more hunks, and both the parser and programmatic editor reject updates without a diff. Move-only updates have therefore never been a working SDK capability.

Supporting them now would introduce a new cross-provider filesystem contract covering metadata, symlinks, read-only grants, existing destinations, cross-filesystem behavior, and partial failures. The reported issue does not demonstrate a concrete user requirement that justifies this additional surface.

The narrower fix resolves the confirmed grammar/runtime mismatch at its owning boundary, preserves the established runtime behavior, and avoids introducing unsupported rename semantics.

PR #4471, opened 2026-08-17
fix(sandbox): require file operations and add lines in the apply_patch grammar

ayaangazali:
### Summary

#4470 fixed the `UpdateFile` production so the model-facing grammar stops advertising a shape the runtime rejects. Two productions in the same description still do.

```
Patch := Begin { FileOp } End
AddFile := "*** Add File: " path NEWLINE { "+" line NEWLINE }
```

`{ }` is zero-or-more in this description, so both say the shape is optional. The runtime disagrees in both cases:

- `_parse_apply_patch_input` raises `apply_patch input must include at least one file operation` (`apply_patch_tool.py:350`)
- `_parse_add_file` raises `Add File patch for {path} must include at least one + line` (`apply_patch_tool.py:365`)

The lark grammar that is actually sent as the constraint already agrees with the runtime, `start: begin_patch hunk+ end_patch` and `add_hunk: ... add_line+`, so only the description was out of step. This is the same defect #4468 described, on the two operations #4470 did not touch.

This aligns both productions using the shape #4470 introduced for `UpdateFile`:

```
Patch := Begin FileOp { FileOp } End
AddFile := "*** Add File: " path NEWLINE "+" line NEWLINE { "+" line NEWLINE }
```

Nothing else changes. The lark grammar, the runtime parser, and the accepted patch shapes are all untouched; this only stops the description promising the model two shapes that cannot parse.

### Test plan

- `test_description_requires_at_least_one_file_operation` and `test_description_requires_at_least_one_add_line` assert the corrected productions and pin the matching lark rules alongside them, so the two can no longer drift apart. Both fail against `main`.
- `test_runtime_rejects_a_patch_with_no_file_operations` and `test_runtime_rejects_an_add_file_without_lines` pin the runtime side, mirroring the `test_runtime_rejects_updates_without_diff` coverage added in #4470. These pass before and after by design: they are what establishes which side of the mismatch is correct.
- `uv run pytest tests/sandbox/capabilities/test_apply_patch_tool.py`: 29 passed.

### Issue number

Same class as #4468, which #4470 fixed for `UpdateFile` only.

### Checks

---

worth saying plainly: I found this by re-reading #4470 right after it merged and checking whether the same mismatch survived elsewhere in the file, so the credit for the diagnosis is really that PR's. if you consider `{ }` in this prose loose enough that only `UpdateFile` was worth tightening, close it and I will take the point. freshman in college, still learning how literally these descriptions are meant to be read :)

### DIFF
--- a/src/agents/sandbox/capabilities/tools/apply_patch_tool.py
+++ b/src/agents/sandbox/capabilities/tools/apply_patch_tool.py
@@ -88,11 +88,11 @@ jump to the right context. For instance:
 [3 lines of post-context]
 
 The full grammar definition is below:
-Patch := Begin { FileOp } End
+Patch := Begin FileOp { FileOp } End
 Begin := "*** Begin Patch" NEWLINE
 End := "*** End Patch" NEWLINE
 FileOp := AddFile | DeleteFile | UpdateFile
-AddFile := "*** Add File: " path NEWLINE { "+" line NEWLINE }
+AddFile := "*** Add File: " path NEWLINE "+" line NEWLINE { "+" line NEWLINE }
 DeleteFile := "*** Delete File: " path NEWLINE
 UpdateFile := "*** Update File: " path NEWLINE [ MoveTo ] Hunk { Hunk }
 MoveTo := "*** Move to: " newPath NEWLINE
--- a/tests/sandbox/capabilities/test_apply_patch_tool.py
+++ b/tests/sandbox/capabilities/test_apply_patch_tool.py
@@ -662,3 +662,60 @@ async def _execute_custom_tool_call(
         config=RunConfig(),
     )
     return result
+
+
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for checking the remaining productions after #4470. The prose mismatch is real, but it does not have the same runtime effect as #4468: on current main and v0.21.1, the actual Lark format already uses `hunk+` and `add_line+`, and the SDK forwards that format unchanged. The custom-tool CFG constrains generated tool input, so neither empty shape can be emitted through this supported path.

I am going to close this PR because no concrete user-visible failure has been demonstrated, and the added string-pinning and runtime tests are not justified for this prose-only inconsistency. If you can provide a real Responses custom-tool reproduction where this wording causes a failed or skipped tool call despite the CFG, we can revisit the narrow wording change.

PR #4473, opened 2026-08-17
ci(tests): run Windows tests on Python 3.10 and 3.13

### LINKED ISSUE #4472: tests-windows pins one Python version, so clock-resolution bugs are invisible on Windows for 3 of 5 supported versions
`tests-windows` pins a single Python version, so a whole class of bug is invisible to CI on the one platform where it occurs. #4392 was an instance of that class; this is about the gap that let it sit there.

## The gap

```
tests (matrix)   ubuntu-latest    3.10  3.11  3.12  3.13  3.14
tests-windows    windows-latest   3.13 only
packaged-contract-windows         3.13 only
```

The project's classifiers claim 3.10 through 3.14, and the Linux matrix honours that. On Windows only one version is exercised, and it happens to be the version where the failure in #4392 does not reproduce.

## Why that particular version hides things

From the measurements in #4392, `datetime.now()` resolution on Windows over a 1000-call burst:

| Python | distinct timestamps per 1000 calls |
|---|---:|
| 3.10.19 | 2 |
| 3.11.14 | 2 |
| 3.12.12 | 2 |
| 3.13.11 | 1000 |

3.13 is the version where the Windows clock stopped being coarse. Pinning Windows CI to exactly that version means anything sensitive to clock granularity, ordering on tied timestamps, or short-interval timing is untested on Windows for three of the five supported versions. In #4392 that was 37 tests failing on a clean checkout while CI stayed green.

To be clear about scope: #4392 fixed the instance, so nothing is failing right now as far as I know. This is about the next one.

## Suggestion, sized for the CI bill

Windows runners bill at a higher multiplier, so a full five-version Windows matrix is probably not worth it. Adding one older version would cover the class:

```yaml
tests-windows:
  runs-on: windows-latest
  strategy:
    matrix:
      python-version: ["3.10", "3.13"]
```

3.10 is the declared minimum and sits on the coarse-clock side; 3.13 is what runs today. That is 2x the current Windows cost rather than 5x, and it would have caught #4392 before it reached a contributor's machine.

If even that is too much, running the older version on a schedule rather than per-PR would still surface the class, just later.

I am happy to send the PR if you tell me which shape you want, or to leave it if the cost is not worth the coverage. Filing this rather than opening a PR straight away because the CI budget is your call, not mine.

feiiiiii5:
### Summary

`tests-windows` pins a single Python version (3.13), which is exactly the version where the Windows clock became fine-grained — so the whole class of coarse-clock bug (e.g. #4392, 37 tests failing on a clean checkout while CI stayed green) is invisible on Windows for the other supported versions. This matrices the job across `"3.10"` (the declared minimum, on the coarse-clock side) and `"3.13"` (what runs today), using the shape suggested in the issue: 2x the current Windows runner cost instead of 5x. `fail-fast: false` keeps both legs reporting, and the `detect-changes` gate is preserved unchanged. `packaged-contract-windows` is intentionally untouched (packaging smoke, not test coverage).

### Test plan

Workflow-only change; verified locally:
- `tests.yml` parses as valid YAML and the `tests-windows` job carries `strategy.matrix.python-version == ["3.10", "3.13"]` with `fail-fast: false`, and `Setup uv` consumes `${{ matrix.python-version }}` (asserted programmatically).
- The matrix shape mirrors the existing Linux `tests` job, and no other job references the changed keys.

Final proof is CI itself: this PR's checks page should show two `tests-windows` legs (3.10 and 3.13), both honoring the `detect-changes` gate.

### Issue number

Fixes #4472

### Checks

- [ ] I've added new tests, if relevant

The verification-stack checkboxes are left unchecked honestly: this diff touches only a CI workflow file, and the applicable verification is the YAML/structure check described in the test plan plus the workflow legs on this PR.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -267,6 +267,12 @@ jobs:
   tests-windows:
     runs-on: windows-latest
     timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - "3.10"
+          - "3.13"
     env:
       OPENAI_API_KEY: fake-for-tests
     steps:
@@ -283,7 +289,7 @@ jobs:
           version: "0.11.14"
           enable-cache: true
           prune-cache: true
-          python-version: "3.13"
+          python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         if: steps.changes.outputs.run == 'true'
         run: uv sync --all-extras --all-packages --group dev


seratch:
Thanks for sending this. https://github.com/openai/openai-agents-python/pull/4475 resolves the issue.

feiiiiii5:
Thanks @seratch! Closing in favor of #4475, which covers this more comprehensively (aligning version coverage across all CI jobs rather than just the Windows matrix). Happy the underlying gap got addressed — keeping the branch on my fork in case anything is reusable.

PR #1296, opened 2025-07-29
Clarify shallow-copy behavior of tools and handoffs in Agent.clone()

abdul-kabir-jawed:
Fixes #1293

This PR clarifies that `tools` and `handoffs` in `Agent.clone()` are shallow-copied:
- New list objects are created.
- Inner tool and handoff objects are shared.

Includes:
- Updated docstring in `Agent.clone()`
- Minimal test verifying shallow-copy behavior

Issue #1293, opened 2025-07-29
Clarify Copy Type for tools and handoffs in Agent.clone() Documentation


abdul-kabir-jawed:
**Description**
---------------

The documentation for `Agent.clone()` does not explain whether the `tools` and `handoffs` lists are **shallow-copied or deep-copied**.\
This leads to ambiguity when cloning agents, as developers may accidentally share references to the same tool or handoff objects, causing unintended side effects.

From reviewing the implementation, both lists are **shallow-copied**:

-   New list objects are created.

-   The elements inside (tools and handoffs) remain the **same references**.

* * * * *

**Steps to Reproduce**
----------------------

```python
from openai.agents import Agent, function_tool, handoff

@function_tool
def greet(name: str) -> str:
    return f"Hello, {name}!"

target_agent = Agent(name="TargetAgent", instructions="Handle delegated work")
original = Agent(
    name="OriginalAgent",
    instructions="Uses tools and delegates",
    tools=[greet],
    handoffs=[handoff(target_agent)]
)

clone = original.clone(name="ClonedAgent")

print(clone.tools is original.tools)             # False (different list)
print(clone.tools[0] is original.tools[0])       # True (same tool object)
print(clone.handoffs is original.handoffs)       # False (different list)
print(clone.handoffs[0] is original.handoffs[0]) # True (same handoff object)
```

**Result:**

-   `tools` and `handoffs` are shallow-copied: new list containers, same inner objects.

* * * * *

**Expected Behavior**
---------------------

Documentation should clearly state:

-   `tools` and `handoffs` are **shallow-copied** in `Agent.clone()`.

-   Cloned agents share the same tool and handoff objects.

-   To customize independently, provide new lists in `clone()`.

* * * * *

**Suggested Fix**
-----------------

1.  Update `Agent.clone()` documentation in [Agents docs](https://openai.github.io/openai-agents-python/) to include shallow-copy semantics.

2.  Add inline comment in `src/agents/agent.py`:

    ```# Shallow-copy mutable lists: new lists, same elements.```

3.  Optional: Add a test verifying shallow-copy behavior.

* * * * *

**Environment**
---------------

-   **SDK Version:** Latest (July 29, 2025)

-   **Python Version:** 3.8+

-   **OS:** Platform-independent

**Labels:** documentation, enhancement

seratch:
[...]

PR #4474, opened 2026-08-17
docs: correct Agent.clone list attribute semantics

thegoodengineer:
### Summary

The `Agent.clone()` docstring added in #1296 makes two claims that the implementation does not hold, both about how list attributes are copied. It currently says:

> Mutable attributes like `tools` and `handoffs` are shallow-copied: new list objects are created only if overridden, but their contents (tool functions and handoff objects) are shared with the original.

Neither half is reliable. `clone()` calls `dataclasses.replace`, which never copies a list at all, so what the clone gets is whatever the merged arguments hold:

```
CLAIM 1: "new list objects are created only if overridden"
  omitted                      -> cloned.tools is original.tools     True
  overridden with the SAME list -> cloned.tools is original.tools     True
  overridden with a NEW list    -> cloned.tools is the supplied list  True

CLAIM 2: "their contents are shared with the original"
  overridden with [other_tool] -> any entry shared with original      False
```

So `agent.clone(tools=agent.tools)` is an override that creates no new list, and `agent.clone(tools=[other_tool])` shares no contents. Overriding is not what decides either outcome; what the caller passes is.

The distinction that does hold, and that the docstring is right to warn about, is between passing an attribute and not passing it. When you do not pass one, the clone receives the original agent's own list, so appending through either agent changes the other.

This is easy to get backwards. The reporter in #1293 asserted `clone.tools is not original.tools` and the test failed in review for exactly this reason. The regression test added alongside that docstring works around the behavior rather than pinning it: `test_agent_clone_shallow_copy` passes `tools=original.tools.copy()`, and its own docstring calls that "the tools.copy() workaround". Nothing in the suite covered the un-passed case, which is why the inaccuracy survived.

The equivalent JSDoc in the JS SDK was corrected in openai/openai-agents-js#1705, where automated review flagged these same two claims. This change brings the Python docstring to the same description so both SDKs document one contract.

### Fix

Docstring and tests only. No behavior change and no API change.

The note now leads with the mechanism, that `dataclasses.replace` never copies a list attribute, and lets the cases follow from it rather than asserting an outcome per case. It names the affected attributes (`tools`, `handoffs`, `mcp_servers`, `input_guardrails`, `output_guardrails`), separates passing an attribute from omitting it, gives both of the counterexamples above, and keeps the existing guidance on how to get an independently owned list.

### Test plan

Four tests added to `tests/test_agent_clone_shallow_copy.py`, each pinning one behavior the docstring now describes:

- `test_agent_clone_keeps_list_attributes_it_is_not_given`
- `test_agent_clone_uses_a_given_list_as_is`
- `test_agent_clone_still_shares_when_given_the_original_list`
- `test_agent_clone_shared_list_mutation_affects_both_agents`

They pin current behavior, so they pass before and after. To confirm they are load bearing rather than tautological, I temporarily made `clone()` copy its list attributes when they are not passed, which is the behavior the old docstring described, and two of the four failed:

```
FAILED tests/test_agent_clone_shallow_copy.py::test_agent_clone_keeps_list_attributes_it_is_not_given
FAILED tests/test_agent_clone_shallow_copy.py::test_agent_clone_shared_list_mutation_affects_both_agents
2 failed, 3 passed
```

Then reverted that experiment. The pre-existing `test_agent_clone_shallow_copy` is left untouched.

Commands run locally on Windows. `make` is unavailable in this environment, so the Makefile targets were invoked directly through `uv`:

- `uv run pytest tests/test_agent_clone_shallow_copy.py`: 5 passed
- `uv run ruff format`: 905 files left unchanged
- `uv run ruff check`: all checks passed
- `uv run python .github/scripts/check_optional_truthiness.py src/agents`: passed
- `uv run pytest -n auto --maxprocesses=9 --dist worksteal -m "not serial"`: 17 failed, 7847 passed, 82 skipped
- The same command on unmodified `main`: 17 failed, 7843 passed, 82 skipped. The failure set is identical; all 17 are pre-existing on this platform, failing in `os.symlink` with `OSError: [WinError 1314] A required privilege is not held by the client`, in `tests/sandbox/test_tar_utils.py`, `tests/sandbox/test_workspace_paths.py`, and `tests/sandbox/test_docker.py`. The only difference is the four tests this PR adds, all passing.
- `uv run mypy src`: 3 errors, all pre-existing and all in files this PR does not touch (`sandbox/util/tar_utils.py`, `sandbox/sandboxes/unix_local.py`, `sandbox/sandboxes/docker.py`)
- `uv run pyright --project pyrightconfig.json`: 1 error, in `sandbox/util/tar_utils.py`. Confirmed pre-existing by stashing this change and re-running on clean `main`, which reports the identical error.

One note on methodology, in case it saves anyone a repeat: an earlier full-suite run of mine reported 6 extra failures in `tests/extensions/memory/test_advanced_sqlite_session.py`, all of them multiprocess contention tests. That was host contention from running the suite and the typecheckers concurrently, not this change. Run on its own, that file reports 128 passed.

### Issue number

N/A. Follows up #1293 and #1296, and aligns with openai/openai-agents-js#1705.

### Checks

- [ ] I've run `.agents/skills/code-change-verification/scripts/run.sh` (`make` is unavailable here; the equivalent commands are listed above)

### DIFF
--- a/src/agents/agent.py
+++ b/src/agents/agent.py
@@ -548,11 +548,18 @@ class Agent(AgentBase, Generic[TContext]):
     def clone(self, **kwargs: Any) -> Agent[TContext]:
         """Make a copy of the agent, with the given arguments changed.
         Notes:
-            - Uses `dataclasses.replace`, which performs a **shallow copy**.
-            - Mutable attributes like `tools` and `handoffs` are shallow-copied:
-              new list objects are created only if overridden, but their contents
-              (tool functions and handoff objects) are shared with the original.
-            - To modify these independently, pass new lists when calling `clone()`.
+            - Uses `dataclasses.replace`, which performs a **shallow copy** and never copies a
+              list attribute such as `tools`, `handoffs`, `mcp_servers`, `input_guardrails`, or
+              `output_guardrails`. Each of those attributes is whatever the merged arguments hold.
+            - An attribute you do not pass arrives as the original agent's own list, so both
+              agents hold that one list and its entries. Appending through either agent, for
+              example `cloned.tools.append(extra_tool)`, therefore also changes the other.
+            - An attribute you do pass is used exactly as given, so it shares a list or an entry
+              with the original agent only where you reused one. `agent.clone(tools=agent.tools)`
+              still shares that list, while `agent.clone(tools=[other_tool])` shares nothing.
+            - To give the clone a list that no other agent holds, pass a new one, for example
+              `agent.clone(tools=[*agent.tools, extra_tool])`. The entries copied into it remain
+              the same objects the original agent holds.
         Example:
             ```python
             new_agent = agent.clone(instructions="New instructions")
--- a/src/agents/realtime/agent.py
+++ b/src/agents/realtime/agent.py
@@ -103,11 +103,18 @@ class RealtimeAgent(AgentBase, Generic[TContext]):
         """Make a copy of the agent, with the given arguments changed.
 
         Notes:
-            - Uses `dataclasses.replace`, which performs a **shallow copy**.
-            - Mutable attributes like `tools` and `handoffs` are shallow-copied:
-              new list objects are created only if overridden, but their contents
-              (tool functions and handoff objects) are shared with the original.
-            - To modify these independently, pass new lists when calling `clone()`.
+            - Uses `dataclasses.replace`, which performs a **shallow copy** and never copies a
+              list attribute such as `tools`, `handoffs`, `mcp_servers`, or `output_guardrails`.
+              Each of those attributes is whatever the merged arguments hold.
+            - An attribute you do not pass arrives as the original agent's own list, so both
+              agents hold that one list and its entries. Appending through either agent, for
+              example `cloned.tools.append(extra_tool)`, therefore also changes the other.
+            - An attribute you do pass is used exactly as given, so it shares a list or an entry
+              with the original agent only where you reused one. `agent.clone(tools=agent.tools)`
+              still shares that list, while `agent.clone(tools=[other_tool])` shares nothing.
+            - To give the clone a list that no other agent holds, pass a new one, for example
+              `agent.clone(tools=[*agent.tools, extra_tool])`. The entries copied into it remain
+              the same objects the original agent holds.
 
         Example:
             ```python
--- a/tests/test_agent_clone_shallow_copy.py
+++ b/tests/test_agent_clone_shallow_copy.py
@@ -30,3 +30,52 @@ def test_agent_clone_shallow_copy():
     assert cloned.tools[0] is original.tools[0], "Tool objects should be same instance"
     assert cloned.handoffs is not original.handoffs, "Handoffs should be different list"
     assert cloned.handoffs[0] is original.handoffs[0], "Handoff objects should be same instance"
+
+
+def test_agent_clone_keeps_list_attributes_it_is_not_given():
+    """An attribute that clone() is not given arrives as the original agent's own list."""
+    target_agent = Agent(name="Target")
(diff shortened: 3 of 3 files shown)


seratch (changes requested):
Before merge, please apply the same correction to `RealtimeAgent.clone()` in `src/agents/realtime/agent.py`, tailored to its list fields. That method uses the same implementation, and its generated API reference currently repeats the same incorrect claims. Leaving it unchanged would document two equivalent public clone paths inconsistently.

The existing `tests/realtime/test_agent.py::test_clone_does_not_mutate_original_lists` already covers both relevant cases: a supplied list is used as-is, and an omitted list remains shared. No additional Realtime test is required. Once the Realtime docstring is aligned and CI is green, this should be ready for another review.

thegoodengineer:
Thanks, that is a fair catch. `RealtimeAgent.clone()` is updated in 03a2c7a0.

One deliberate difference from the `Agent` version: the Realtime note names `tools`, `handoffs`, `mcp_servers`, and `output_guardrails`, and omits `input_guardrails`, since `RealtimeAgent` does not define that field. I checked each of the four against `RealtimeAgent` directly rather than assuming it follows `Agent`, covering the omitted case, the passed-as-given case, `clone(attr=agent.attr)` still sharing that list, `clone(tools=[other_tool])` sharing nothing, and an append through either agent being visible from the other.

No Realtime test added, per your note. `test_clone_does_not_mutate_original_lists` does already cover the supplied-list and omitted-list cases.

I also confirmed nothing else repeats the old wording: it now returns no matches anywhere in the repo, and the only other `clone(**kwargs)`, `HandoffInputData.clone()`, makes no shallow-copy claims and holds tuples. The generated reference picks the change up on its own, since `docs/ref/realtime/agent.md` is an mkdocstrings stub.

Locally: `ruff format` and `ruff check` clean, `tests/realtime/test_agent.py` and `tests/test_agent_clone_shallow_copy.py` passing, and the full suite matches my clean-main baseline name for name (the only failures on this machine are pre-existing Windows symlink privilege ones under `tests/sandbox/`).

The workflow run for 03a2c7a0 is currently showing `action_required`, so it needs approval before CI can report.

thegoodengineer:
Correction to the SHA in my note above: I rebased onto `main` afterwards to clear the out-of-date branch state, so `03a2c7a0` no longer exists. The current head is d18d7b7a, and that is the run awaiting approval.

The rebase carried no content change. The diff against `main` is identical before and after it, and `ruff check` plus `tests/realtime/test_agent.py` and `tests/test_agent_clone_shallow_copy.py` still pass on the new base.

ayaangazali:
This matches what I measured independently, and the enumeration plus the `RealtimeAgent` half is a real improvement over the old wording.

One gap you may want to fold in, since it is the same docstring and the same surprise: the sharing is not limited to list attributes. `model_settings` is a mutable non-frozen dataclass and it is shared by reference too, so the list framing does not cover it and the old "pass new lists" advice never applied to it.

On current `main`:

```
omitted both        -> cloned.model_settings is original.model_settings   True
   cloned.model_settings.temperature = 0.9  ->  original is now           0.9
model passed only   -> cloned.model_settings is original.model_settings   True
model_settings passed -> cloned.model_settings is original.model_settings False
```

The middle row is the one I did not expect. `clone()` only substitutes fresh settings for a new model when `_model_settings_match_implicit_model_defaults(self.model, self.model_settings)` holds, so an agent carrying any explicit setting (here `temperature=0.1`) keeps sharing the same `ModelSettings` object even when `model=` is overridden.

Not asking you to widen the scope if you would rather keep this list-only. I had held the same finding for a docs batch and yours is further along, so it is yours either way; just flagging it so it does not get missed.

seratch (approved):

PR #4479, opened 2026-08-17
fix: isolate usage between RunState checkpoints

chiruu12:
### Summary

`RunContextWrapper._copy_for_run_state` detaches `_approvals`, `_tool_invocations` and `_restored_unbound_approval_call_ids`, but the `copy.copy(self)` it starts from leaves `usage` shared. `Usage.add` accrues in place and extends `request_usage_entries`, so every checkpoint taken from a result, and the result itself, accumulate into one instance.

Two consequences, both reproducible on `main`:

A `RunResult` the caller already holds keeps changing. After `result.to_state()` and a resume, `result.context_wrapper.usage.requests` goes from 1 to 2.

Two checkpoints from one result bill each other. Each resumed once from the same one-request run, so both should report 2 requests. The second reports 3.

The copy is deep because `add` extends `request_usage_entries` rather than replacing it.

`turn_input` is aliased by the same `copy.copy`, but it is reassigned rather than mutated in `turn_resolution.py` and `run_loop.py`, so I left it as is rather than widen the change.

### Test plan

Two tests in `TestRunStateResumption`, next to the existing resume-isolation coverage: one asserts a resume does not move the source result's usage, the other that two checkpoints from one result report the same total and do not share `request_usage_entries`. Reverting only the `run_context.py` line fails them with `assert 2 == 1` and `assert 3 == 2`.

`make format`, `make lint`, `make typecheck` and `make tests` all pass (8806 passed, plus the serial pass).

### Issue number

None, filing the fix directly since it is a narrow follow-up to #4413.

### Checks


seratch (changes requested):
Thanks for the contribution. Please keep the new `copied.usage = copy.deepcopy(self.usage)` assignment in `src/agents/run_context.py`; that correctly isolates top-level checkpoints.

The additional runtime change belongs in `Agent.as_tool()`'s `_run_agent_impl()` in `src/agents/agent.py`. After the `pending_resume_state` `if / elif` block selects `resume_state`, move the existing usage assignment out of the `status in ("approved", "rejected")` branch and apply it to both paths:

```python
if resume_state is not None and resume_state._context is not None:
    resume_state._context.usage = context.usage
```

Place this immediately after the `pending_resume_state` `if / elif` block and before `if run_result is None:`. This preserves top-level checkpoint isolation while ensuring cached nested `Agent.as_tool()` resumes continue accumulating usage on the current outer run.

chiruu12:
Done. Moved the usage rebind out of the approved/rejected branch so it applies to the cached resume path too, and kept the `_copy_for_run_state` deepcopy. Added a test that fails without the runtime change. Full stack is green.

seratch (approved):

PR #4480, opened 2026-08-17
fix: handle non-UTF8 Codex stderr

jstar0:
### Summary

Failed Codex subprocesses currently decode stderr strictly as UTF-8 before raising their `RuntimeError`. If stderr contains an invalid byte, `UnicodeDecodeError` masks the subprocess exit code and diagnostic output.

This change replacement-decodes stderr only on the existing nonzero-exit path, so callers still receive the original `RuntimeError` shape with the exit code and readable diagnostic text. Successful executions and stdout decoding are unchanged.

### Test plan

- Added a regression for a nonzero Codex subprocess exit with non-UTF-8 stderr.
- Ran `.agents/skills/code-change-verification/scripts/run.sh`; format, lint, type checking, and the full test suite passed.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/extensions/experimental/codex/exec.py
+++ b/src/agents/extensions/experimental/codex/exec.py
@@ -200,7 +200,7 @@ class CodexExec:
 
             if process.returncode not in (0, None):
                 await stderr_task
-                stderr_text = b"".join(stderr_chunks).decode("utf-8")
+                stderr_text = b"".join(stderr_chunks).decode("utf-8", errors="replace")
                 raise RuntimeError(
                     f"Codex exec exited with code {process.returncode}: {stderr_text}"
                 )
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for sharing this. The strict UTF-8 decode does mean that a synthetic invalid byte can replace the intended `RuntimeError` with `UnicodeDecodeError`, and the proposed one-line change is technically reasonable. However, this PR provides only a `FakeProcess` case and no report or reproduction showing the supported Codex CLI emitting non-UTF-8 stderr. An arbitrary byte sequence from a test double does not establish a user-facing need.

I am going to close this PR for now. If this occurs with the Codex CLI, please provide the OS, Codex CLI version, invocation or configuration, and a reproduction or captured stderr byte sequence with secrets removed. We can then revisit this narrow replacement-decoding fix.

PR #4481, opened 2026-08-17
docs: document the minimum uv version for development

### LINKED ISSUE #4476: make sync fails on uv older than 0.9.25 with a TOML error pointing at the repo's own files
### Describe the bug

`make sync` fails on any `uv` older than 0.9.25, and the error points at this repository's own files rather than at the tool version, so it reads as though the checkout is broken.

`AGENTS.md` and the README ask for "`uv` installed" without a floor, while CI pins `0.11.14` in every workflow. A contributor between those two points gets no usable signal.

### Debug information

- Agents SDK version: `main` at d40f5d9
- Python version: 3.13.11 (also reproduced on 3.12.10)
- Does the issue reproduce with the latest release? Yes, the two files involved are unchanged.
- Does it occur consistently or intermittently? Consistently, it is version-determined.

```text
$ uv sync --all-extras
warning: Failed to parse `pyproject.toml` during settings discovery:
  TOML parse error at line 233, column 36
      |
  233 | exclude-newer-package = { openai = false }
      |                                    ^^^^^
  data did not match any variant of untagged enum Helper

error: Failed to parse `uv.lock`
  Caused by: TOML parse error at line 11, column 1
   |
11 | [options]
   | ^^^^^^^^^
data did not match any variant of untagged enum Helper
```

Neither line is malformed. `exclude-newer-package` and the `[options]` table with `exclude-newer-span` are both newer `uv` features, and an older `uv` reports them as invalid TOML.

### Repro steps

Bisected against this checkout, running `uv sync --all-extras` from a clean clone:

```text
uv 0.9.21   uv.lock parse: FAIL
uv 0.9.23   uv.lock parse: FAIL
uv 0.9.24   uv.lock parse: FAIL
uv 0.9.25   ok -> Resolved 202 packages in 2ms
uv 0.9.30   ok
uv 0.10.0   ok
uv 0.11.14  ok      <- what CI pins
```

So the floor is exactly 0.9.25. 0.9.24 was released about five weeks before it, so this is reachable by anyone who installed `uv` recently and has not updated, not only by someone on a very old version.

### Expected behavior

Either of these would close it, and I have no preference:

Declare the floor so `uv` reports it itself. `[tool.uv] required-version = ">=0.9.25"` turns both errors into a message naming the installed and required versions.

Or state it in `AGENTS.md` and the contributing section next to the existing `uv` prerequisite, alongside the Python 3.10+ line that is already there.

The first is self-maintaining and fails at the right layer; the second costs nothing and is visible before the first command runs. Pinning the same version CI uses is a third option but seems heavier than the problem.

Happy to send a PR for whichever shape you prefer. Filing first because the choice is a project convention rather than a bug fix, and #4472 suggested I ask rather than guess on that kind of thing.

uuzzrm:
### Summary

`uv` versions older than 0.9.25 cannot parse this repository's current `pyproject.toml` and `uv.lock`, so `make sync` reports misleading TOML errors from the checkout. This pull request documents 0.9.25 as the minimum development version in the contributor guide and README.

### Test plan

- `uv sync --all-extras --all-packages --group dev` — passed with uv 0.12.5; 202 packages resolved.
- `mingw32-make.exe format` — passed.
- `mingw32-make.exe lint` — passed.
- `uv run mkdocs build` — passed; existing documentation warnings remain.
- `uv run mypy src` — blocked by existing errors in `unix_local.py`, `docker.py`, and `tar_utils.py`.
- `uv run pyright --project pyrightconfig.json --threads 4` — blocked by the existing `tar_utils.py` return-type error.
- `mingw32-make.exe typecheck` — the aggregate target cannot run under the Windows shell because it uses Unix-shell syntax; the underlying mypy and Pyright commands were run directly.
- `uv run pytest -n auto --maxprocesses=9 --dist worksteal -m "not serial"` — 7,403 passed and 82 skipped; 19 failures and 135 collection errors were caused by the Windows environment and existing platform assumptions (`tee`, symlink behavior, and the system GBK locale).

The repository-wide verification checklist is left incomplete because of those unrelated baseline/environment failures.

### Issue number

Closes #4476

### Checks

- [ ] I've run `.agents/skills/code-change-verification/scripts/run.ps1`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -173,7 +173,7 @@ The OpenAI Agents Python repository provides the Python Agents SDK, examples, an
 ### Prerequisites
 
 - Python 3.10+.
-- `uv` installed for dependency management (`uv sync`) and `uv run` for Python commands.
+- `uv` 0.9.25 or newer installed for dependency management (`uv sync`) and `uv run` for Python commands. Older versions cannot parse the repository's current project and lockfile configuration.
 - `make` available to run repository tasks.
 
 ### Development Workflow
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ For voice support, install with the optional `voice` group: `pip install 'openai
 
 If you're familiar with [uv](https://docs.astral.sh/uv/), installing the package would be even easier:
 
+For repository development, use `uv` 0.9.25 or newer. Older versions cannot parse the repository's current project and lockfile configuration.
+
 ```bash
 uv init
 uv add openai-agents


seratch:
We recommend using the latest one, so no need to declare the version range.

PR #4483, opened 2026-08-17
fix(usage): isolate aggregated request usage entries

sylvesterkaczmarek:
### Summary
- detach existing `request_usage_entries` when aggregating `Usage` values
- detach nested token-detail objects when synthesizing a per-request entry
- add regression coverage for mutations in both directions

This keeps the aggregate's per-request history independent from the source `Usage` object after `Usage.add()` returns.

### Test plan
- added focused regression tests in `tests/test_usage_copy_isolation.py`
- GitHub Actions

### Issue number
N/A


seratch:
Doing deep copying here could be slightly safer in terms of the data isolation, but I don't think there is actually a real use case.

PR #4484, opened 2026-08-17
fix(models): copy multi-provider mappings

sylvesterkaczmarek:
### Summary
- copy dictionaries passed to `MultiProviderMap.set_mapping()`
- prevent later mutation of a caller-owned dictionary from silently changing provider routing
- add focused regression coverage for mapping isolation

This matches `get_mapping()`, which already returns a copy rather than exposing the internal mapping directly.

### Test plan
- added focused coverage in `tests/models/test_multi_provider_map.py`
- GitHub Actions

### Issue number
N/A


seratch:
I don't see the necessity to make a copy here.

PR #4485, opened 2026-08-17
fix(sandbox): validate normalized cwd separators

sylvesterkaczmarek:
### Summary
- apply the existing POSIX-separator validation after normalizing `PurePath` inputs
- reject backslashes embedded in `PurePosixPath` values just like equivalent string inputs
- preserve relative `PureWindowsPath` support through its existing POSIX normalization
- add focused regression coverage

This removes a representation-dependent validation gap in `normalize_sandbox_cwd()` without changing the policy for raw Windows-style string paths.

### Test plan
- added focused coverage in `tests/sandbox/test_sandbox_cwd_path_validation.py`
- GitHub Actions

### Issue number
N/A

### DIFF
--- a/src/agents/sandbox/workspace_paths.py
+++ b/src/agents/sandbox/workspace_paths.py
@@ -73,12 +73,12 @@ def normalize_sandbox_cwd(cwd: str | PurePath) -> PurePosixPath:
     if isinstance(cwd, PurePath):
         raw_cwd = cwd.as_posix()
     elif isinstance(cwd, str):
-        if "\\" in cwd:
-            raise ValueError("sandbox.cwd must use POSIX path separators")
         raw_cwd = cwd
     else:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. I confirmed the representation difference, but it is semantically meaningful rather than a validation gap. A raw string containing `\` is rejected as ambiguous Windows-style input, while `PureWindowsPath` is normalized and `PurePosixPath(r"tasks\child")` denotes a valid POSIX path segment containing a literal backslash. That typed-path behavior shipped in v0.21.1.

Moving the check after `as_posix()` would reject a valid released input, while the stated Windows-path outcome is already covered by `PureWindowsPath` or native Windows `Path` normalization. I am going to close this PR. If there is a concrete supported scenario where accepting that literal POSIX segment causes meaningful downstream harm, we can revisit the contract from that evidence.

PR #4486, opened 2026-08-17
fix(agent-tools) reject invalid input builder results

sylvesterkaczmarek:
### Summary
- validate custom structured agent-tool input builder results at the API boundary
- raise a clear `TypeError` when a builder returns an unsupported runtime type
- preserve existing string, list, and awaitable builder behaviour

This prevents invalid builder values from being silently cast and failing later in the run pipeline.

### Test plan
- added focused regression coverage in `tests/test_agent_tool_input_invalid_builder.py`
- GitHub Actions

### Issue number
N/A


seratch:
input_builder comes with a type hint of callable that returns a valid type here.

PR #4487, opened 2026-08-17
fix(agent-tools) reindex mutated tool calls

sylvesterkaczmarek:
### Summary
- remove a tool-call object's previous signature entry when its signature changes
- keep fallback lookups aligned with the currently recorded nested run result
- prune empty stale signature buckets

Without this cleanup, recording the same mutable tool-call object again after fields such as arguments change can leave the old signature pointing at the new run result.

### Test plan
- added focused regression coverage in `tests/test_agent_tool_state_reindex.py`
- verifies the old signature no longer resolves while the new signature does
- GitHub Actions

### Issue number
N/A

### DIFF
--- a/src/agents/agent_tool_state.py
+++ b/src/agents/agent_tool_state.py
@@ -99,6 +99,13 @@ def _index_agent_tool_run_result(
 ) -> None:
     """Track tool call objects by signature for fallback lookup."""
     signature = _scoped_tool_call_signature(tool_call, scope_id=scope_id)
+    previous_signature = _agent_tool_run_result_signature_by_obj.get(scoped_object)
+    if previous_signature is not None and previous_signature != signature:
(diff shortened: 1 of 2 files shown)


seratch:
The stale signature entry is reproducible when the same cached `ResponseFunctionToolCall` is mutated and then recorded again in the same scope, and the cleanup itself is mechanically sound. However, the current test reaches that state only by mutating the object and calling internal cache helpers directly. I could not find an SDK path that mutates signature fields after registration, and the PR does not show a supported `Agent.as_tool()` resume scenario or user-visible consequence.

I am going to close this PR for now. If you can provide a public-path reproduction showing how ordinary supported usage re-records a mutated call and causes the wrong nested result or approval to be recovered, we can revisit the narrow invariant fix with that regression test.

PR #4488, opened 2026-08-17
fix(mcp) ignore whitespace-only tool metadata

sylvesterkaczmarek:
### Summary
- treat whitespace-only MCP strings as empty metadata
- allow useful annotation titles to win when an explicit title is blank whitespace
- let model-facing descriptions fall back to titles when descriptions contain only whitespace
- skip whitespace-only MCP server labels and tool names during metadata collection

The helper already models non-empty string semantics; this makes that behaviour consistent with other schema-description handling in the SDK.

### Test plan
- added focused regression coverage in `tests/test_mcp_tool_metadata_whitespace.py`
- GitHub Actions

### Issue number
N/A


seratch:
I don't think there is a real need of modifying the given metadata here.

PR #4489, opened 2026-08-17
fix(typecheck): pin mypy's analysis platform so results match CI

### LINKED ISSUE #4477: make typecheck reports three errors on Windows and none on Linux
### Describe the bug

`make typecheck` reports three errors on Windows and none on Linux, on the same checkout. `AGENTS.md` asks contributors to run it before marking work ready, so a Windows contributor sees three failures they did not cause and cannot act on.

The `typecheck` job runs on `ubuntu-latest`, so CI never exercises the Windows result.

### Debug information

- Agents SDK version: `main` at d40f5d9, clean tree
- Python version: 3.13.11
- Does it reproduce with the latest release? Yes, the three sites are unchanged.
- Consistently or intermittently? Consistently, it is platform-determined.

```text
$ uv run mypy src
src\agents\sandbox\util\tar_utils.py:161: error: Incompatible return value type (got "_TemporaryFileWrapper[bytes]", expected "IOBase")  [return-value]
src\agents\sandbox\sandboxes\unix_local.py:339: error: Name "process" is used before definition  [used-before-def]
src\agents\sandbox\sandboxes\docker.py:1430: error: Argument "stream" to "_stream_into_exec" of "DockerSandboxSession" has incompatible type "_TemporaryFileWrapper[bytes]"; expected "IOBase"  [arg-type]
Found 3 errors in 3 files (checked 305 source files)
```

### Repro steps

The platform flag is the whole difference. Same checkout, same mypy, same command otherwise:

```text
$ uv run mypy src
Found 3 errors in 3 files (checked 305 source files)

$ uv run mypy --platform linux src
Success: no issues found in 305 source files
```

mypy analyses `sys.platform` branches and platform-specific typeshed definitions for the platform it is told about, defaulting to the host. `tempfile.NamedTemporaryFile` and the `unix_local` code paths differ between the two, which is why all three land in `src/agents/sandbox/`.

### Expected behavior

`make typecheck` should give the same answer on any contributor machine, since it is gating contributions rather than describing the runtime.

Two shapes, and I do not have a preference:

Pin the platform so the command is reproducible, for example `python_platform = "linux"` under `[tool.mypy]` in `pyproject.toml`, or `--platform linux` in the Makefile target. This makes every contributor see exactly what CI enforces. The cost is that genuine Windows-only type problems stay invisible, which is the situation today anyway.

Or fix the three sites and add a Windows leg to the typecheck job. This is the stricter option and would have caught these three, but it is a real CI cost and the sandbox code may not be intended to typecheck on Windows at all.

Happy to send a PR for either. Filing rather than guessing because pinning the platform is a project convention decision.

Related: `#4476` is a different Windows-contributor gap in the same setup flow.

ProjeetBhaumik:
### Summary

`make typecheck` gates contributions, but mypy defaults to the host platform, so the same checkout answers differently depending on who runs it.

On `d40f5d9` under Python 3.14:

```
$ uv run mypy --platform linux src
Success: no issues found in 305 source files

$ uv run mypy --platform win32 src
Found 3 errors in 3 files (checked 305 source files)
```

The three win32-only errors are all in `src/agents/sandbox/`, where `tempfile` and the `unix_local` paths differ between platforms in typeshed:

- `src/agents/sandbox/util/tar_utils.py:161` — `Incompatible return value type (got "_TemporaryFileWrapper[bytes]", expected "IOBase")`
- `src/agents/sandbox/sandboxes/unix_local.py:339` — `Name "process" is used before definition`
- `src/agents/sandbox/sandboxes/docker.py:1430` — `Argument "stream" ... has incompatible type "_TemporaryFileWrapper[bytes]"; expected "IOBase"`

The typecheck job runs on `ubuntu-latest`, so CI never exercises the win32 result. A Windows contributor sees three failures they did not cause and cannot act on.

Pinning mypy to Linux makes the gate reproducible and makes it report what CI actually enforces.

### Why pyright is left alone

The issue covers `make typecheck`, which runs mypy and pyright, so the symmetric change would be `"pythonPlatform": "Linux"` in `pyrightconfig.json`. I tried that and backed it out, because it makes things worse off-Linux.

Pyright resolves the host's real installed stdlib rather than bundled stubs, so forcing `pythonPlatform` to Linux on a macOS host produces a hybrid environment that no real machine sees. Measured on macOS, same venv, back to back:

| | pyright |
| --- | --- |
| `main` | 0 errors |
| `main` + `"pythonPlatform": "Linux"` | 14 errors in `tests/` |

Those 14 are spurious — a real Linux machine and a real macOS machine both report 0. Pyright's platform default is already the right behaviour, so this PR touches mypy only.

### Note on a second, separate axis

While measuring this I hit a different source of nondeterminism, which this PR does **not** address: mypy's results also vary by interpreter version. Under Python 3.10 the tree reports 5 errors on any platform (4 in `archive_ops.py` for `SpooledTemporaryFile` vs `IOBase`, 1 in `run_loop.py` for `BaseExceptionGroup`), and 0 under 3.11+. `uv sync` resolves 3.10 locally when there is no `.python-version` file, while CI pins 3.14 and sees none of them.

That looks worth a separate decision — either pinning the dev interpreter or type-checking the supported floor — so I left it out rather than bundling it here. Happy to follow up if you have a preference.

### Test plan

Python 3.14.7 on macOS, `uv sync --all-extras --all-packages --group dev`:

- `bash .agents/skills/code-change-verification/scripts/run.sh` — all commands passed (`make lint` 38s, `make tests` 265s, `make typecheck` 305s)
- `make typecheck` on this branch — mypy `Success: no issues found in 305 source files`, pyright `0 errors, 0 warnings, 0 informations`
- Confirmed the config key takes effect by flipping it to `win32` and re-running: 3 errors, matching the `--platform win32` override

One caveat, for transparency: two earlier full-suite runs on this machine each hit a single different timeout-based failure (`test_stop_terminates_active_pty_sessions`, then `test_read_path_probe_resolves_symlinks_before_classifying_missing`). Both pass in isolation and both passed in the final run; they appear to be load-induced on my hardware rather than related to this change, which touches no runtime code.

### Issue number

Closes #4477

### Checks

- [x] I've added new tests (if relevant) — n/a, configuration-only change with no runtime behaviour
- [x] I've added/updated the relevant documentation — n/a, the rationale is a comment at the config site

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,6 +153,7 @@ strict = true
 disallow_incomplete_defs = false
 disallow_untyped_defs = false
 disallow_untyped_calls = false
+platform = "linux"
 
 [[tool.mypy.overrides]]
 module = "sounddevice.*"


ProjeetBhaumik:
Closing in favour of #4482, which fixes the three errors at the source rather than pinning the analysis platform. That is the better fix: I checked it out locally and it clears the full tree on both `mypy --platform win32 src` and `--platform linux src` (305 files, no issues either way), not just the three modules its test plan lists.

PR #4490, opened 2026-08-17
fix(asyncio): preserve producer failures while draining consumers

sylvesterkaczmarek:
### Summary
- preserve the producer exception when the consumer also fails during producer-failure draining
- continue waiting for the consumer task to settle before propagating the producer failure
- keep consumer-first failure and parent-cancellation behaviour unchanged

`run_producer_consumer()` documents producer failures as primary while the consumer drains. Previously a secondary consumer exception raised during that drain replaced the producer exception.

### Test plan
- added focused regression coverage in `tests/test_asyncio_tasks_primary_error.py`
- GitHub Actions

### Issue number
N/A

### DIFF
--- a/src/agents/util/_asyncio_tasks.py
+++ b/src/agents/util/_asyncio_tasks.py
@@ -139,7 +139,7 @@ async def run_producer_consumer(
         try:
             producer_result = producer_task.result()
         except BaseException:
-            await consumer_task
+            await asyncio.gather(consumer_task, return_exceptions=True)
             raise
 
         consumer_result = await consumer_task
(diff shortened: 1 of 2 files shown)


seratch:
I traced `run_producer_consumer()` through its two current callers. Both dispatchers already catch and log ordinary `Exception`s from `on_stream`, so the test's `ConsumerError(Exception)` cannot reach this helper through the supported `Agent.as_tool()` or Codex tool paths. The consumer failures that intentionally escape are `BaseException`s such as `CancelledError`; preserving that behavior was part of the #4198 and #4201 fix.

In a controlled `Agent.as_tool(on_stream=...)` interleaving, the released code propagates `CancelledError`, while this patch replaces it with the producer `RuntimeError`. With the default failure handler, that becomes a normal tool-error result and can let the outer run continue. Since no supported scenario demonstrates why producer failure should override consumer cancellation, I do not think this change is worth merging. I am going to close this PR for now.

PR #4491, opened 2026-08-17
fix(tools): normalize custom data copy errors

sylvesterkaczmarek:
### Summary
- include defensive copying inside the custom-data normalization error boundary
- convert non-copyable custom-data values into the same `UserError` used for other non-JSON-compatible payloads
- preserve existing behaviour for valid mappings and JSON serialization failures

This prevents implementation-specific `copy.deepcopy()` errors from escaping a public custom-data extractor boundary.

### Test plan
- added focused regression coverage in `tests/test_custom_data_copy_error.py`
- GitHub Actions

### Issue number
N/A

### DIFF
--- a/src/agents/util/_custom_data.py
+++ b/src/agents/util/_custom_data.py
@@ -26,8 +26,8 @@ def normalize_custom_data(value: Mapping[str, Any] | None) -> dict[str, Any] | N
     if not all(isinstance(key, str) for key in value):
         raise UserError("custom_data_extractor must return a mapping with string keys.")
 
-    copied = copy.deepcopy(dict(value))
     try:
+        copied = copy.deepcopy(dict(value))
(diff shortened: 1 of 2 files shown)


seratch:
I confirmed that a generator value currently raises `TypeError` from `copy.deepcopy()` before the existing JSON-compatibility error boundary. However, a generator is not supported JSON-compatible custom data, and the only demonstrated effect here is a different exception class for invalid extractor output.

The JSON serialization round-trip already validates the payload and creates a detached copy, so expanding the `deepcopy()` error boundary would preserve redundant behavior without a demonstrated user impact. I am going to close this PR for now. We can revisit the underlying boundary if there is a real supported payload that fails during copying, or evidence that this exception distinction causes a meaningful operational problem.

PR #4492, opened 2026-08-17
fix(output): isolate returned JSON schemas

sylvesterkaczmarek:
### Summary
- return a deep copy from `AgentOutputSchema.json_schema()`
- prevent caller mutation of a returned schema from changing the schema retained by the output contract
- preserve the existing plain-text error behaviour

`json_schema()` exposes a nested mutable dictionary. Returning the internal object lets an inspection or downstream modification silently alter later model-facing schema reads from the same `AgentOutputSchema` instance.

### Test plan
- added focused regression coverage in `tests/test_agent_output_schema_copy.py`
- GitHub Actions

### Issue number
N/A

### DIFF
--- a/src/agents/agent_output.py
+++ b/src/agents/agent_output.py
@@ -1,4 +1,5 @@
 import abc
+import copy
 from dataclasses import dataclass
 from typing import Any, cast, get_args, get_origin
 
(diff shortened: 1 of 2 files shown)


seratch:
Making a copy every time the getter method could bring a slight overhead, plus I don't see the necessity to protect the original data in this scenario.

PR #4493, opened 2026-08-17
fix(config): reject invalid default OpenAI API values

sylvesterkaczmarek:
### Summary
- validate runtime values passed to `set_default_openai_api()`
- reject unsupported API names instead of silently selecting the Responses API
- avoid changing the provider default when validation fails

The function is annotated with a `Literal`, but runtime callers can still pass arbitrary strings. Previously every value other than `"chat_completions"` was treated as `"responses"`.

### Test plan
- added focused regression coverage in `tests/test_default_openai_api_validation.py`
- GitHub Actions

### Issue number
N/A


seratch:
It's marked in the type hint, so we'd like to avoid adding this duplicated runtime validation.

Issue #3270, opened 2026-05-08
ModelRetryBackoffSettings accepts negative values that produce invalid retry delays


Aphroq:
### Please read this first

- **Have you read the docs?** Yes. The runner-managed retry docs describe `initial_delay`, `max_delay`, and `multiplier` as retry backoff settings.
- **Have you searched for related issues?** Yes. I searched for related `ModelRetryBackoffSettings`, negative delay, negative `max_delay`, negative `multiplier`, and model retry issues/PRs and did not find a direct match.

### Describe the bug

`ModelRetryBackoffSettings` accepts negative `initial_delay`, `max_delay`, and `multiplier` values. Those values can flow into the default retry delay calculation and produce negative delays. The retry sleep helper skips non-positive delays, so invalid retry configuration can turn into immediate retries instead of a clear configuration error.

### Debug information

- Agents SDK version: `0.17.0`
- Python version: Python 3.12.1
- Revision: `683b6e79`

### Repro steps

Run this script from the repository root:

```python
from agents.retry import ModelRetryBackoffSettings
from agents.run_internal.model_retry import _default_retry_delay

settings = [
    ModelRetryBackoffSettings(initial_delay=-1.0, jitter=False),
    ModelRetryBackoffSettings(max_delay=-1.0, jitter=False),
    ModelRetryBackoffSettings(multiplier=-2.0, jitter=False),
]

for setting in settings:
    print(setting.to_json_dict(), "->", _default_retry_delay(2, setting))
```

Actual result:

```text
{'initial_delay': -1.0, 'max_delay': None, 'multiplier': None, 'jitter': False} -> -2.0
{'initial_delay': None, 'max_delay': -1.0, 'multiplier': None, 'jitter': False} -> -1.0
{'initial_delay': None, 'max_delay': None, 'multiplier': -2.0, 'jitter': False} -> -0.5
```

### Expected behavior

Negative retry backoff settings should be rejected during `ModelRetryBackoffSettings` construction or dict coercion, before they can produce invalid retry delays. Zero values can remain valid because they are explicit immediate-retry configuration rather than negative time.

PR #4495, opened 2026-08-17
fix(retry): reject negative retry counts

sylvesterkaczmarek:
### Summary
- constrain `ModelRetrySettings.max_retries` to non-negative values
- preserve `0` as the explicit no-retry configuration
- fail during settings validation instead of silently treating negative counts as an exhausted retry budget

The setting represents retries allowed after the initial request, so negative values are not meaningful. The backoff settings already use Pydantic numeric constraints; this applies the same validation boundary to the retry count.

### Test plan
- added focused regression coverage in `tests/test_retry_count_validation.py`
- GitHub Actions

### Issue number
N/A

### DIFF
--- a/src/agents/retry.py
+++ b/src/agents/retry.py
@@ -210,7 +210,7 @@ def retry_policy_retries_all_transient_errors(policy: RetryPolicy | None) -> boo
 class ModelRetrySettings:
     """Opt-in runner-managed retry settings for model calls."""
 
-    max_retries: int | None = None
+    max_retries: int | None = Field(default=None, ge=0)
     """Retries allowed after the initial model request."""
 
     backoff: ModelRetryBackoffInput | None = None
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. I confirmed that `ModelRetrySettings(max_retries=-1)` is currently accepted and that this patch rejects it. However, the existing runtime already treats every negative value as the same no-retry behavior as `max_retries=0` across the runner and provider retry paths. The PR does not link a user report or demonstrate an observable consequence beyond accepting a semantically invalid configuration. This also differs from #3270, where negative backoff values produced invalid delays and immediate retries.

I am going to close this PR for now. If we see a concrete supported scenario where negative retry counts cause meaningful behavior beyond disabling retries, we can reconsider a compatibility-preserving validation change.

PR #4498, opened 2026-08-17
fix(tracing): detach serialized trace metadata

sylvesterkaczmarek:
### Summary
- deep-copy exported trace data in `Trace.to_json()`
- keep nested metadata in persistence and transport payloads isolated from the live trace
- preserve existing tracing API-key serialization behaviour

The previous top-level `dict()` copy left nested metadata shared, so mutating a serialized payload could silently mutate the trace object it came from.

### Test plan
- added focused regression coverage in `tests/test_trace_json_isolation.py`
- GitHub Actions

### Issue number
N/A

### DIFF
--- a/src/agents/tracing/traces.py
+++ b/src/agents/tracing/traces.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import abc
 import contextvars
+import copy
 import hashlib
 import threading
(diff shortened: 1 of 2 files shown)


seratch:
I don't think deep copying here is necessary.

PR #4501, opened 2026-08-17
fix(tracing): make no-op starts idempotent

sylvesterkaczmarek:
### Summary
- make repeated `NoOpTrace.start()` calls idempotent
- make repeated `NoOpSpan.start()` calls idempotent
- prevent a second `mark_as_current=True` start from overwriting the original ContextVar token

Normal trace and span implementations already ignore repeated starts. The no-op implementations did not, so a second start replaced the saved token with one whose previous value was the no-op object itself. A single `finish(reset_current=True)` then restored that object as current instead of clearing it, leaking tracing context when tracing is disabled.

### Test plan
- added focused regression coverage in `tests/test_noop_tracing_start_idempotent.py`
- GitHub Actions

### Issue number
N/A

### DIFF
--- a/src/agents/tracing/spans.py
+++ b/src/agents/tracing/spans.py
@@ -222,11 +222,12 @@ class NoOpSpan(Span[TSpanData]):
         span_data: The operation-specific data for this span.
     """
 
-    __slots__ = ("_span_data", "_prev_span_token")
+    __slots__ = ("_span_data", "_prev_span_token", "_started")
 
     def __init__(self, span_data: TSpanData):
         self._span_data = span_data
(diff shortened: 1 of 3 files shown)


seratch:
I confirmed that explicitly starting the same no-op trace or span twice overwrites its saved context token and leaves the no-op object current after one finish. However, the documented lifecycle is either context-manager use or one balanced manual start()/finish() pair, and I could not find an SDK path that starts the same no-op object twice.

Please provide a concrete supported scenario that reaches this repeated-start sequence and shows the resulting context leak affecting later work. Without such a scenario, this currently protects only an unsupported lifecycle composition, so I would close this PR for now.

PR #4504, opened 2026-08-17
feat(models): add provider-level strict tool schema default compatibility

sylvesterkaczmarek:
### Summary

Follow-up to #4390, #4426, and the provider-boundary guidance in #4459.

This adds an opt-in compatibility setting for Chat Completions providers that reject strict function-tool schemas containing JSON Schema `default` keywords:

```python
provider = OpenAIProvider(
    openai_client=client,
    use_responses=False,
    strip_tool_schema_defaults=True,
)
```

The default remains `False`, so existing OpenAI and OpenAI-compatible requests are unchanged.

### Why this shape

The earlier proposals changed `ensure_strict_json_schema()` globally or exposed an option only at that low-level helper. The feedback on #4459 was that, if an escape hatch is needed, it should live at the user-facing tool or provider boundary so an affected `@function_tool` workflow can opt in directly.

This revision therefore keeps strict-schema generation unchanged and performs compatibility normalization only on the converted Chat Completions request payload.

When enabled:

- only strict function/handoff tool parameter schemas are normalized
- `default` is removed from actual JSON Schema nodes, including nested objects, definitions, unions, arrays, and conditionals
- annotation payloads that happen to contain a key named `default` are left alone
- the caller-owned `FunctionTool.params_json_schema` is not mutated
- response-format schemas and non-strict tools are unchanged

This PR does not claim to establish the root cause of the Azure behavior reported in #4390. It provides the narrow provider-level compatibility boundary discussed in #4459 without changing the direct OpenAI default.

### Test plan

Added focused coverage in `tests/models/test_tool_schema_default_compat.py` that exercises the public path:

`OpenAIProvider` -> `OpenAIChatCompletionsModel` -> `Agent` -> `Runner.run()` -> recorded Chat Completions request.

The tests verify that:

- defaults remain in the outgoing strict tool schema by default
- `strip_tool_schema_defaults=True` removes them from the outgoing request
- the original `FunctionTool.params_json_schema` remains unchanged
- schema annotation data containing its own `default` key is preserved

Full repository verification is left to GitHub Actions.

### Issue number

Related to #4390. Supersedes the implementation approach in #4426 for the compatibility use case.

### DIFF
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -54,6 +54,92 @@ if TYPE_CHECKING:
     from ..model_settings import ModelSettings
 
 
+_SCHEMA_MAPPING_KEYWORDS = (
+    "$defs",
+    "definitions",
+    "properties",
+    "patternProperties",
+    "dependentSchemas",
+)
+_SCHEMA_SINGLE_KEYWORDS = (
+    "additionalProperties",
+    "unevaluatedProperties",
+    "unevaluatedItems",
+    "propertyNames",
+    "items",
+    "contains",
+    "not",
+    "if",
+    "then",
+    "else",
+    "contentSchema",
+)
+_SCHEMA_SEQUENCE_KEYWORDS = ("allOf", "anyOf", "oneOf", "prefixItems")
+
+
+def _strip_json_schema_defaults(schema: dict[str, Any]) -> dict[str, Any]:
+    """Return a schema copy with `default` removed only from schema nodes."""
+    result = {key: value for key, value in schema.items() if key != "default"}
+
+    for keyword in _SCHEMA_MAPPING_KEYWORDS:
+        mapping = result.get(keyword)
+        if isinstance(mapping, dict):
+            result[keyword] = {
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for following up and moving the proposed escape hatch to the provider boundary. However, this PR still does not include the default-isolated Azure reproduction requested in #4459. The reported strict versus non-strict comparison changes the entire schema mode, and the nested Decimal schema remains a confounding variable. Direct OpenAI requests accept the current schema, while callers can already post-process `FunctionTool.params_json_schema` for an affected tool.

This PR therefore adds a public provider option and a separate JSON Schema traversal without a demonstrated unmet need. I am going to close it for now. We can reconsider if this PR provides otherwise-identical with-default and without-default Azure requests, including the endpoint, API version, deployment, raw schemas, and complete responses.

sylvesterkaczmarek:
Understood. I’ve left this closed and posted a default-isolated Azure Chat Completions probe on #4390 that keeps strict mode and the schema shape fixed and changes only the `default` keyword. It asks for the endpoint, API version, deployment, raw paired requests, and complete responses you requested here and in #4459. I’ll wait for that evidence before revisiting the provider-boundary API.

Issue #3106, opened 2026-05-04
Responses streaming treats failed and incomplete terminal events as success


Aphroq:
### Please read this first

- [x] **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-python/)
- [x] **Have you searched for related issues?** Others may have faced similar issues.

### Describe the bug

Responses streaming treats `response.failed` and `response.incomplete` terminal events as usable final responses in a few paths. That can allow a run to finish successfully even though the Responses API reported a failed or incomplete response.

Relevant code:

- `src/agents/models/openai_responses.py`: `stream_response()` stores the response payload from `response.failed` and `response.incomplete` as `final_response`.
- `src/agents/run_internal/run_loop.py`: streamed runner finalization converts `response.failed` and `response.incomplete` payloads into `ModelResponse`.
- `src/agents/models/openai_responses.py`: websocket `get_response()` also accepts `response.failed` and `response.incomplete` terminal payloads.

Current tests also encode this behavior for both model-level websocket handling and streamed runner handling.

### Debug information

- Agents SDK version: `main` at `3854c124cb8e3e51fb660f5714405ee39ee86c5e`
- Python version: Python 3.12

### Repro steps

Minimal reproducer outline:

```python
from collections.abc import AsyncIterator

import pytest
from openai.types.responses import Response, ResponseOutputMessage, ResponseOutputText

from agents import Agent, Runner
from agents.items import TResponseStreamEvent
from agents.models.interface import Model, ModelTracing
from agents.model_settings import ModelSettings
from agents.usage import Usage


class FailedStreamingModel(Model):
    async def get_response(self, *args, **kwargs):
        raise NotImplementedError

    async def stream_response(
        self,
        system_instructions,
        input,
        model_settings: ModelSettings,
        tools,
        output_schema,
        handoffs,
        tracing: ModelTracing,
        *,
        previous_response_id=None,
        conversation_id=None,
        prompt=None,
    ) -> AsyncIterator[TResponseStreamEvent]:
        response = Response(
            id="resp_failed",
            created_at=0,
            model="fake",
            object="response",
            output=[
                ResponseOutputMessage(
                    id="msg_1",
                    content=[
                        ResponseOutputText(
                            text="partial output from failed response",
                            type="output_text",
                            annotations=[],
                        )
                    ],
                    role="assistant",
                    type="message",
                    status="completed",
                )
            ],
            tool_choice="none",
            tools=[],
[...]

PR #4516, opened 2026-08-18
fix(models): reject terminal failed/incomplete responses in non-streaming get_response

### LINKED ISSUE #4515: [Bug]: Non-streaming Responses get_response treats response.failed/incomplete as a successful empty turn
### Expected Behavior

When the Responses API reports a terminal `failed` or `incomplete` response (e.g. `status="incomplete"` with `incomplete_details.reason="max_output_tokens"`), the model layer should **not** treat it as a successful empty turn. Every other path already rejects these terminal states as `ModelBehaviorError`:

- `stream_response()` (HTTP streaming) — raises `response_terminal_failure_error(...)` for `response.failed` / `response.incomplete` events
- Websocket `get_response()` — raises `response_terminal_failure_error(...)` for the same terminal payloads
- Streamed runner finalization — converted in `run_loop.py`

This family was fixed in https://github.com/openai/openai-agents-python/issues/3106 (closed 2026-05-04) — but the **HTTP non-streaming `get_response()` path was never covered**.

### Current Behavior

`OpenAIResponsesModel.get_response()` (HTTP, non-streaming) returns `ModelResponse(output=[])` when `client.responses.create()` returns a `Response` with `status="incomplete"` or `status="failed"`. `AnyLLMModel._get_response_via_responses()` has the same gap.

A run that hits `max_output_tokens` before emitting any visible token therefore concludes as an indistinguishable empty turn instead of surfacing the truncation — the same failure mode as https://github.com/openai/openai-agents-python/issues/4510 (which was fixed for the chat-completions path), and structured-output runs can fall into the same empty-output retry loop.

### Steps to Reproduce

```python
import asyncio
from typing import Any, cast
from openai.types.responses.response import IncompleteDetails
from agents import ModelSettings
from agents.models.openai_responses import OpenAIResponsesModel
from agents.tracing.model_tracing import ModelTracing

class Responses:
    async def create(self, **kwargs: Any) -> Response:
        return Response(
            id="resp-incomplete",
            created_at=0, model="fake", object="response", output=[],
            tool_choice="none", tools=[], top_p=None, parallel_tool_calls=False,
            usage=None,
            status="incomplete",
            incomplete_details=IncompleteDetails(reason="max_output_tokens"),
        )

class Client:
    responses = Responses()

async def main() -> None:
    model = OpenAIResponsesModel(model="gpt-4", openai_client=cast(Any, Client()))
    result = await model.get_response(
        system_instructions=None, input="hi", model_settings=ModelSettings(),
        tools=[], output_schema=None, handoffs=[], tracing=ModelTracing.DISABLED,
    )
    print("output =", result.output)   # [] -- treated as a successful empty turn

asyncio.run(main())
```

Output:

```text
output = []
```

No error is raised. (Control: the same mocked response on `status="completed"` also returns `[]`; the streaming and websocket paths raise `ModelBehaviorError` for `failed`/`incomplete`.)

### Root Cause

- `src/agents/models/openai_responses.py` → `_fetch_response()` (non-stream branch) does not check `response.status`.
- `src/agents/extensions/models/any_llm_model.py` → `_get_response_via_responses()` does not check `response.status`.

The terminal-state handling introduced for #3106 only covered the event-driven paths (streaming + websocket). The HTTP non-streaming path returns the raw `Response` object, whose terminal status is silently ignored.

### Suggested Fix

Raise the same `ModelBehaviorError` used elsewhere when the non-streaming response is terminal-failed:

```python
if getattr(response, "status", None) in {"failed", "incomplete"}:
    raise response_terminal_failure_error(f"response.{response.status}", response)
```

in both `_fetch_response` (non-stream) and `_get_response_via_responses`. This makes non-streaming behavior consistent with the streaming and websocket paths.

### Impact

Non-streaming runs (the default for `Runner.run`) that hit `max_output_tokens` with zero emitted tokens surface as an empty turn with no explanation, and structured-output runs can loop on empty output (see #4510 for the same loop on the chat-completions path). The fix is a two-line consistency change.

weike-zhang:
Refs #4515

### Summary

The HTTP **non-streaming** `get_response()` path treats a terminal `response.failed` / `response.incomplete` payload as a successful empty turn: `OpenAIResponsesModel.get_response()` and `AnyLLMModel._get_response_via_responses()` return `ModelResponse(output=[])` instead of surfacing the terminal state.

Every other path already rejects these states as `ModelBehaviorError` (streaming `stream_response()`, websocket `get_response()`, streamed runner finalization — fixed in #3106, May 2026). The non-streaming HTTP path was never covered by that fix.

### Fix

Raise the same `response_terminal_failure_error(...)` used elsewhere when the non-streaming response carries a terminal status:

```python
if getattr(response, "status", None) in {"failed", "incomplete"}:
    raise response_terminal_failure_error(f"response.{response.status}", response)
```

in `_fetch_response()` (non-stream branch) and `_get_response_via_responses()`. `getattr` keeps custom/mock clients that return plain objects working.

### Verification

Broken (main, mocked `client.responses.create` returning `status="incomplete"`, `reason="max_output_tokens"`):

```text
output = []          # treated as a successful empty turn, no error
```

Fixed (this branch):

```text
ModelBehaviorError: Responses stream ended with terminal event `response.incomplete`.
status=incomplete; incomplete_details=IncompleteDetails(reason='max_output_tokens').
```

- New tests: `test_get_response_rejects_failed_terminal_response_status` (openai model, parametrized `incomplete`/`failed`) and `test_any_llm_responses_path_rejects_failed_terminal_status` (any_llm model, parametrized).
- `tests/models`: 816 passed, 11 skipped.
- `ruff check` / `ruff format --check`: clean.

### Issue number

Closes #4515

### Checks

- [x] `tests/models`: 816 passed, 11 skipped (includes the new tests)
- [x] `ruff check` / `ruff format --check` on the changed files: clean

### DIFF
--- a/src/agents/extensions/models/any_llm_model.py
+++ b/src/agents/extensions/models/any_llm_model.py
@@ -405,6 +405,9 @@ class AnyLLMModel(Model):
                 prompt=prompt,
             )
 
+            if getattr(response, "status", None) in {"failed", "incomplete"}:
+                raise response_terminal_failure_error(f"response.{response.status}", response)
+
             if _debug.DONT_LOG_MODEL_DATA:
                 logger.debug("LLM responded")
             else:
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -822,6 +822,8 @@ class OpenAIResponsesModel(Model):
 
         if not stream:
             response = await client.responses.create(**create_kwargs)
+            if getattr(response, "status", None) in {"failed", "incomplete"}:
+                raise response_terminal_failure_error(f"response.{response.status}", response)
             _mark_transport_request_without_usage(response)
             return cast(Response, response)
 
--- a/tests/models/test_any_llm_model.py
+++ b/tests/models/test_any_llm_model.py
@@ -23,6 +23,7 @@ from openai.types.responses import (
     ResponseOutputMessage,
     ResponseOutputRefusal,
 )
+from openai.types.responses.response import IncompleteDetails
 from openai.types.responses.response_created_event import ResponseCreatedEvent
 from openai.types.responses.response_error_event import ResponseErrorEvent
 from openai.types.responses.response_failed_event import ResponseFailedEvent
@@ -172,6 +173,24 @@ def _response(text: str, response_id: str = "resp_123") -> Response:
                 {"cache_write_tokens": 0, "cached_tokens": 0}
             ),
             output_tokens_details=OutputTokensDetails(reasoning_tokens=0),
+        ),
+    )
+
+
+def _responses_response_with_terminal_status(status: str) -> Response:
+    return Response(
+        id="resp_terminal",
+        created_at=123,
+        model="fake-model",
+        object="response",
+        output=[],
+        tool_choice="none",
+        tools=[],
+        parallel_tool_calls=False,
+        usage=None,
+        status=status,
+        incomplete_details=(
+            IncompleteDetails(reason="max_output_tokens") if status == "incomplete" else None
         ),
     )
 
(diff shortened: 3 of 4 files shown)


seratch (approved):
Thanks for the fix. I also verified the non-streaming behavior against the live Responses API, rather than only with mocked responses.

Test setup:

- `openai-agents==0.21.1`
- `openai==3.3.0`
- `any-llm-sdk==1.11.0`
- `gpt-5.6`
- `store=False`
- Provider-managed retries disabled
- Both the native `OpenAIResponsesModel` path and the `AnyLLMModel` Responses path

For both paths, a completed control returned normally. With a low `max_output_tokens` and high reasoning effort, the first attempt returned:

- `status="incomplete"`
- `incomplete_details.reason="max_output_tokens"`
- No visible `output_text`

The current SDK returned a `ModelResponse` without raising in both cases. Applying the guard from this PR to the captured live responses raised the expected `ModelBehaviorError`. I reproduced the `incomplete` case live. I did not attempt to induce a live `failed` response.

One difference from the PR description is that the live responses did not have `output=[]`. They contained one `reasoning` output item, while still having no visible assistant output. Therefore, the exact "successful empty turn" shape from the mocked reproduction is not universal. The underlying issue is nevertheless confirmed: the terminal `incomplete` status is discarded and the call appears successful despite producing no visible response.

We can merge this PR now.

PR #4518, opened 2026-08-18
fix(voice): snapshot STT turn_detection settings

hsusul:
### Summary

This pull request fixes mutable aliasing of STT `turn_detection` settings in `OpenAISTTTranscriptionSession`.

The session previously stored `settings.turn_detection or DEFAULT_TURN_DETECTION` by reference. Mutating the caller mapping after construction changed the `session.update` payload, and mutating the default path poisoned later sessions that relied on `DEFAULT_TURN_DETECTION`.

Construction now copies the caller mapping, or copies the module default when `turn_detection` is `None`. An explicit empty mapping is preserved instead of falling through to the default.

### Test plan

- [x] `uv run pytest -q tests/voice/test_openai_stt.py -k "snapshots_turn_detection or preserves_empty_turn_detection"`
- [x] Confirmed caller mutations no longer affect the session copy
- [x] Confirmed default-path sessions no longer share `DEFAULT_TURN_DETECTION`

### Issue number

N/A

### Checks

- [x] I've confirmed format, lint, typecheck, and focused voice STT tests pass

### DIFF
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -97,7 +97,10 @@ class OpenAISTTTranscriptionSession(StreamedTranscriptionSession):
         self._client = client
         self._model = model
         self._settings = settings
-        self._turn_detection = settings.turn_detection or DEFAULT_TURN_DETECTION
+        if settings.turn_detection is None:
+            self._turn_detection = dict(DEFAULT_TURN_DETECTION)
+        else:
+            self._turn_detection = dict(settings.turn_detection)
         self._trace_include_sensitive_data = trace_include_sensitive_data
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. I can confirm the top-level aliasing mechanically, but the PR does not establish a supported real-world case where mutating `turn_detection` after session construction causes a meaningful failure. The new tests assert private session state, and the default-poisoning case itself requires mutating `_turn_detection`; there is no linked issue or observed user impact.

The `{}` change also broadens the patch: released behavior treats an empty mapping as the SDK's `semantic_vad` default, while this PR would send `{}` to the Realtime API. I am going to close this PR for now. If you can provide a concrete application flow and observable consequence, we can revisit a narrower fix that snapshots the mapping while preserving existing empty-mapping behavior.

PR #4524, opened 2026-08-19
fix(sandbox): snapshot Instrumentation payload policies

hsusul:
### Summary

This pull request finishes constructor snapshotting for sandbox `Instrumentation` payload policies.

#4398 detached the `payload_policy_by_op` mapping, but the default `payload_policy` and the policy objects inside that mapping were still shared with the caller. Mutating `include_exec_output` (or other fields) on those objects after construction could undo audit redaction for later `emit()` calls.

`Instrumentation` now deep-copies the default policy and each per-op policy at construction time. Callers can still change `instrumentation.payload_policy*` after construction if they own that instance.

### Test plan

- [x] `uv run pytest -q tests/sandbox/test_session_manager.py -k snapshots`
- [x] Confirmed mutating constructor-owned default and per-op policy fields no longer leaks exec stdout

### Issue number

N/A

### Checks

- [x] I've confirmed format, lint, typecheck, and focused sandbox tests pass

### DIFF
--- a/src/agents/sandbox/session/manager.py
+++ b/src/agents/sandbox/session/manager.py
@@ -24,8 +24,14 @@ class Instrumentation:
         payload_policy_by_op: dict[OpName, EventPayloadPolicy] | None = None,
     ) -> None:
         self._sinks: list[EventSink] = list(sinks or [])
-        self.payload_policy = payload_policy if payload_policy is not None else EventPayloadPolicy()
-        self.payload_policy_by_op = dict(payload_policy_by_op or {})
+        self.payload_policy = (
+            payload_policy.model_copy(deep=True)
+            if payload_policy is not None
+            else EventPayloadPolicy()
+        )
+        self.payload_policy_by_op = {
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for following up. I can confirm that `Instrumentation` currently retains the supplied policy objects by reference, but this PR does not establish a concrete supported scenario where that behavior causes unintended exposure. The trigger requires caller-owned code to mutate its own configuration object after construction, and callers can already keep redaction stable by avoiding that mutation or update the public `instrumentation.payload_policy*` attributes intentionally.

The proposed fix would also create inconsistent ownership semantics because sink-level `payload_policy` objects remain caller-referenced and can still override redaction. In addition, #4398 explicitly preserved policy-object mutability, and that behavior shipped in v0.22.0.

I am going to close this PR. If there is a real application path where an alias is mutated outside the trusted configuration owner, please provide that scenario so we can decide one consistent policy-ownership contract across instrumentation and sinks.

PR #4525, opened 2026-08-19
fix(mcp): deep-copy cached tools before returning them

hsusul:
### Summary

This pull request finishes cache isolation for MCP tool listings.

#4424 returned a new list from `list_tools()` / `cached_tools`, but the `Tool` objects and nested `input_schema` dicts remained shared with the server cache. Mutating a returned tool's schema could clear `required` fields and bypass `_validate_required_parameters` on later `call_tool()` invocations.

Returned tools are now deep-copied so callers cannot corrupt the cache or required-parameter checks. Callable `tool_filter` functions receive detached copies before invocation, so a filter that mutates nested schemas cannot rewrite `_tools_list`. No-filter and static-filter results stay isolated by the same return-path snapshot.

### Test plan

- [x] `uv run pytest -q tests/mcp/test_caching.py tests/mcp/test_tool_filtering.py`
- [x] Confirmed mutating a returned tool's `input_schema["required"]` still fails call-time validation
- [x] Confirmed a dynamic filter that clears `required` cannot change later listings or `call_tool()` validation
- [x] Confirmed `cached_tools` nested schema mutation does not leak into later snapshots

### Issue number

N/A

### Checks

- [x] I've confirmed all verification steps pass, except pre-existing `tests/sandbox/test_run_cwd.py::test_python_skill_uses_absolute_root_from_nested_workdir` on this host

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -128,6 +128,11 @@ _SAFE_EXCEPTION_GROUP_MESSAGE = "MCP request failed with additional errors."
 _SAFE_EXCEPTION_MESSAGE = "An additional error occurred during the MCP request."
 
 
+def _snapshot_tools(tools: list[MCPTool]) -> list[MCPTool]:
+    """Return deep-copied tools so callers cannot mutate cached schemas."""
+    return [tool.model_copy(deep=True) for tool in tools]
+
+
 def _client_session_read_timeout(timeout_seconds: float | None) -> timedelta | float | None:
     """Convert an MCP read timeout while intentionally treating zero as no timeout."""
     if timeout_seconds is None:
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The deep-copy direction is correct, but the ownership boundary is still incomplete. `list_tools()` passes cached `MCPTool` objects into a callable `tool_filter` before `_snapshot_tools()` runs, so a filter that mutates `input_schema["required"]` can still corrupt `_tools_list` and bypass later call-time validation.

Please ensure callable filters receive detached tool objects before invocation, while no-filter/static-filter results and `cached_tools` remain deeply isolated. Add a regression test where a dynamic filter mutates a nested schema and verify that subsequent listings and `call_tool()` still use the original required fields. Please also extend the `cached_tools` snapshot test to mutate a returned tool or nested schema, rather than only the outer list.

With those focused changes and green CI, this should be ready for another review.

seratch (approved):

PR #4528, opened 2026-08-19
fix(tracing): flush buffered data after tracing is disabled

sylvesterkaczmarek:
### Summary

Keep the public `flush_traces()` operation effective after tracing has been disabled.

`DefaultTraceProvider.force_flush()` currently returns early when the provider's disabled flag is set. That means an application can create buffered trace data, call `set_tracing_disabled(True)` to stop collecting new traces, and then call `flush_traces()` before a worker or request exits, but the explicit flush silently does nothing.

Disabling tracing should control creation of new trace data. It should not prevent already-registered processors from flushing data they already own. This follows the same lifecycle distinction as processor shutdown.

### Reproduction

```python
# traces/spans have already been queued
set_tracing_disabled(True)
flush_traces()
```

On current `main`, `flush_traces()` reaches `DefaultTraceProvider.force_flush()` and returns before calling any registered processor.

### Change

Remove the disabled-state early return from `DefaultTraceProvider.force_flush()`. Trace and span creation remain disabled exactly as before.

### Test plan

Adds a regression test through the public `flush_traces()` function and verifies that a registered processor receives `force_flush()` even after the provider is disabled.

### Issue number

N/A

### DIFF
--- a/src/agents/tracing/provider.py
+++ b/src/agents/tracing/provider.py
@@ -497,10 +497,6 @@ class DefaultTraceProvider(TraceProvider):
 
     def force_flush(self) -> None:
         """Force all processors to flush their buffers immediately."""
-        self._refresh_disabled_flag()
-        if self._disabled:
-            return
-
         try:
             self._multi_processor.force_flush()
         except Exception as e:
--- a/src/agents/tracing/provider.py
+++ b/src/agents/tracing/provider.py
@@ -497,6 +497,7 @@ class DefaultTraceProvider(TraceProvider):
 
     def force_flush(self) -> None:
         """Force all processors to flush their buffers immediately."""
+        self._refresh_disabled_flag()
         try:
             self._multi_processor.force_flush()
         except Exception as e:
--- a/tests/test_disabled_trace_provider_flush.py
+++ b/tests/test_disabled_trace_provider_flush.py
@@ -1,0 +1,18 @@
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+import agents.tracing as tracing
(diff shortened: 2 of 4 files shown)


seratch (changes requested):
Thanks for the fix. Comparing v0.22.0, the base, and this branch confirms that disabling tracing already allows buffered data to be exported by the background worker and shutdown. Allowing an explicit flush to drain that data is therefore a useful correction. 

Please retain `_refresh_disabled_flag()` and remove only the disabled early return. Removing the refresh call also changes when `OPENAI_AGENTS_DISABLE_TRACING` is first cached, which is unrelated to this fix. Please add a focused regression test preserving that initialization timing.

sylvesterkaczmarek:
Fixed. I updated the existing disabled-provider force_flush() test to match the new lifecycle contract, so both the direct provider test and the public flush_traces() regression now verify that buffered processors still flush after tracing is disabled.

seratch (changes requested):
Thanks. The contradictory test is fixed, but my latest review request is still outstanding. Please retain `_refresh_disabled_flag()`, remove only the disabled early return, and add the regression test for the environment variable's initialization timing.

sylvesterkaczmarek:
Done. I restored `_refresh_disabled_flag()` in `force_flush()` while keeping the disabled early return removed, and added a focused regression test confirming that `force_flush()` preserves the first-use initialization and caching behavior of `OPENAI_AGENTS_DISABLE_TRACING`. Thanks for catching that.

seratch (approved):

PR #4531, opened 2026-08-20
fix(voice): honor provider options over the default client

sylvesterkaczmarek:
### Summary

Make `OpenAIVoiceModelProvider` honor provider-specific client options when a global default OpenAI client is configured.

When no explicit `openai_client` is supplied, the voice provider currently checks the global default client first. If one exists, constructor options such as `api_key`, `base_url`, `organization`, and `project` are silently ignored and STT/TTS models use the global client's configuration instead.

### Fix

- reuse the global default client only when the voice provider has no explicit client-construction options
- construct a provider-local `AsyncOpenAI` client when any voice-provider option is explicitly supplied
- preserve an explicitly supplied empty API key instead of replacing it with the global default key
- keep the existing no-options path unchanged, including support for falsy default-client objects

This is the voice-provider counterpart to the client-option precedence being fixed for `OpenAIProvider` in #4530. The two providers have separate client construction paths, so the voice path requires its own correction.

### Test plan

Extended the existing voice-provider tests to verify that, with a global default client present:

- explicit `api_key`, `base_url`, `organization`, and `project` each cause a provider-local client to be created and receive the requested value
- an explicit empty API key is preserved
- the existing no-option path continues to reuse the global default client

The branch is based directly on current `main` and is 0 commits behind it.

### Risk

Low. Providers without explicit client options retain the existing global-default behavior. The change only affects configurations that supplied voice-provider options which were previously ignored.

### Issue number

None. Found while checking voice-provider client-option precedence alongside #4530.

### DIFF
--- a/src/agents/voice/models/openai_model_provider.py
+++ b/src/agents/voice/models/openai_model_provider.py
@@ -81,12 +81,27 @@ class OpenAIVoiceModelProvider(VoiceModelProvider):
     # AsyncOpenAI() raises an error if you don't have an API key set.
     def _get_client(self) -> AsyncOpenAI:
         if self._client is None:
-            default_client = _openai_shared.get_default_openai_client()
+            has_explicit_client_options = any(
+                value is not None
+                for value in (
+                    self._stored_api_key,
+                    self._stored_base_url,
+                    self._stored_organization,
+                    self._stored_project,
+                )
+            )
+            default_client = (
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. The voice-provider issue is valid, but #4530 now contains the same correction and also fixes the corresponding OpenAIProvider path with aligned precedence coverage.

To keep one implementation and review path for this behavior, I am closing this PR as superseded by #4530.

PR #4536, opened 2026-08-20
fix(voice): preserve falsey custom pipeline models

sylvesterkaczmarek:
### Summary

Preserve explicitly supplied custom STT and TTS model instances even when they are falsey.

`VoicePipeline.__init__()` correctly recognizes custom `STTModel` / `TTSModel` instances, but `_get_stt_model()` and `_get_tts_model()` later use truthiness checks. A valid custom model that defines `__bool__()` or `__len__()` as false is therefore discarded and silently replaced with a model from the configured provider.

### Fix

- treat only `None` as the signal that a pipeline model still needs to be resolved
- preserve every explicitly supplied `STTModel` / `TTSModel` instance regardless of truthiness
- leave string model-name resolution and default-provider behavior unchanged

This also matches the existing Realtime runner behavior, which explicitly preserves falsey custom model instances.

### Test plan

Added focused regressions for falsey custom STT and TTS subclasses and verified `_get_stt_model()` / `_get_tts_model()` return the exact supplied instances.

The branch is based directly on current `main` at `7e55afc9500d12937687988f1e91e900dcb4ad09` and is 0 commits behind it. The production diff is two truthiness checks changed to explicit `is None` checks.

### Risk

Very low. The only behavior change is for explicitly supplied model objects that evaluate to false. `None`, string model names, and normal truthy model instances behave exactly as before.

### Issue number

None. Found while auditing custom-model preservation across Voice and Realtime.

### DIFF
--- a/src/agents/voice/pipeline.py
+++ b/src/agents/voice/pipeline.py
@@ -75,12 +75,12 @@ class VoicePipeline:
             raise UserError(f"Unsupported audio input type: {type(audio_input)}")
 
     def _get_tts_model(self) -> TTSModel:
-        if not self.tts_model:
+        if self.tts_model is None:
             self.tts_model = self.config.model_provider.get_tts_model(self._tts_model_name)
         return self.tts_model
 
     def _get_stt_model(self) -> STTModel:
-        if not self.stt_model:
+        if self.stt_model is None:
             self.stt_model = self.config.model_provider.get_stt_model(self._stt_model_name)
         return self.stt_model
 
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this. The current evidence demonstrates only that a synthetic custom model can override `__bool__()` and be replaced by the provider model. It does not show a real supported custom STT or TTS implementation that is intentionally falsey or a practical user consequence that justifies permanent regression coverage.
I am closing this PR for now. If there is a concrete custom model implementation or user report where falsey behavior is intentional and observable, we can reconsider the narrow fix from that evidence.

PR #4537, opened 2026-08-20
fix(voice): refresh callable API keys before streamed STT

sylvesterkaczmarek:
### Summary

Refresh callable or rotating `AsyncOpenAI` API keys before opening the streamed STT WebSocket connection.

The OpenAI Python client supports API-key providers and refreshes them before normal requests. Streamed STT opens its WebSocket manually, however, and currently reads `client.api_key` directly for the `Authorization` header without invoking the client's refresh hook first. With a callable key provider, that can send the initial empty/stale key instead of the current credential.

The Responses WebSocket transport in this SDK already performs the same refresh before preparing its handshake.

### Fix

- invoke the OpenAI client's refresh hook when available before the streamed-STT handshake
- keep the refresh feature-detected so compatible custom clients without the hook remain unchanged
- build the existing authorization header only after refresh completes

### Test plan

Added a focused regression with a rotating client whose initial API key is empty and whose refresh hook installs a new key. The test verifies that:

- the refresh hook is called exactly once before the handshake
- the WebSocket `Authorization` header contains the refreshed key

The branch is based directly on current `main` at `7e55afc9500d12937687988f1e91e900dcb4ad09` and is 0 commits behind it. The production change is limited to the refresh helper and one pre-handshake call.

### Risk

Low. Static API keys behave as before. The additional work only occurs when the supplied OpenAI client exposes the refresh hook used for dynamic credentials.

### Issue number

None. Found while auditing streamed STT parity with the existing Responses WebSocket authentication path.

### DIFF
--- a/src/agents/voice/models/openai_stt.py
+++ b/src/agents/voice/models/openai_stt.py
@@ -58,6 +58,13 @@ def _audio_buffer_to_base64(buffer: npt.NDArray[np.int16 | np.float32]) -> str:
     return base64.b64encode(buffer.tobytes()).decode("utf-8")
 
 
+async def _refresh_openai_client_api_key_if_supported(client: Any) -> None:
+    """Refresh callable/rotating OpenAI credentials before a manual websocket handshake."""
+    refresh_api_key = getattr(client, "_refresh_api_key", None)
+    if callable(refresh_api_key):
+        await refresh_api_key()
+
+
 async def _wait_for_event(
     event_queue: asyncio.Queue[dict[str, Any] | ErrorSentinel],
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. Callable API keys do need to be refreshed before the manual streamed-STT WebSocket handshake, and the regression test captures the relevant failure.

However, #4532 is already changing the same handshake and its current review requires this refresh behavior. Please move the refresh helper and test into #4532, using the shared OpenAI-client WebSocket utility there, and close #4537 so this connection path is corrected in one PR.

PR #4538, opened 2026-08-20
fix: validate Responses compaction modes before routing

green3sf:
### Summary

This pull request fixes `OpenAIResponsesCompactionSession` accepting unsupported compaction modes and silently routing them through input-based compaction, which could send locally stored session history instead of using a supplied response chain.

- Validate constructor, per-call, mutable-default, and deferred-compaction modes before response-state mutation, session-history reads, decision hooks, or API calls.
- Canonicalize supported string values and consistently raise `ValueError` for unsupported runtime objects without invoking caller-controlled equality, type-spoofing, or representation hooks.
- Preserve the existing routing behavior for `auto`, `input`, and `previous_response_id`.

### Test plan

- Added regression coverage for empty and unknown strings, non-string values, adversarial string subclasses, spoofed `__class__`, failing `__repr__`, post-construction mutation, and deferred compaction.
- Ran `.agents/skills/code-change-verification/scripts/run.sh`; formatting, lint, type checking, and the full test suite passed.
- Completed the repository-required independent final review; both final reviewers returned clean verdicts on the verified fingerprint.

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/memory/openai_responses_compaction_session.py
+++ b/src/agents/memory/openai_responses_compaction_session.py
@@ -29,6 +29,20 @@ DEFAULT_COMPACTION_THRESHOLD = 10
 _ALL_SESSION_ITEMS_LIMIT = 2_147_483_647
 
 OpenAIResponsesCompactionMode = Literal["previous_response_id", "input", "auto"]
+_VALID_COMPACTION_MODES: tuple[OpenAIResponsesCompactionMode, ...] = (
+    "auto",
+    "input",
+    "previous_response_id",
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. `compaction_mode` is already explicitly restricted to `"auto"`, `"input"`, or `"previous_response_id"` through its `Literal` type hint. We also want to avoid adding duplicate client-side validation for values that are already outside the supported typed contract.

For that reason, we are going to close this PR.

PR #4539, opened 2026-08-20
fix: close all cached websocket models after failures

green3sf:
### Summary

This pull request fixes `OpenAIProvider.aclose()` so a failure while closing one cached WebSocket model no longer prevents the provider from attempting to close the remaining unique models.

The provider preserves the first normal close exception, clears processed loop caches, and keeps cleanup on a running owner loop when possible. Cancellation and other `BaseException` values remain fail-fast, and the shared `AsyncOpenAI` client remains open.

### Test plan

- Ran `.agents/skills/code-change-verification/scripts/run.sh` with all formatting, lint, type-checking, and test steps passing.
- Completed two independent final reviews of the lifecycle, cancellation, compatibility, and test boundaries.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/models/openai_provider.py
+++ b/src/agents/models/openai_provider.py
@@ -195,17 +195,34 @@ class OpenAIProvider(ModelProvider):
             await self._close_models(models)
             return
         if loop.is_running():
-            for model in models:
-                future = asyncio.run_coroutine_threadsafe(model.close(), loop)
-                await asyncio.wrap_future(future)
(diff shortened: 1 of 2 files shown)


green3sf:
I found #4465 after opening this draft. The maintainer explanation there also applies here because my regression test replaces `OpenAIResponsesWSModel.close()` instead of reproducing an exception escaping its real implementation. Closing this as a duplicate—sorry for the noise.

PR #4542, opened 2026-08-20
fix(voice): accept async API-key providers

sylvesterkaczmarek:
### Summary

Allow `OpenAIVoiceModelProvider(api_key=...)` to accept the async API-key provider shape supported by `AsyncOpenAI`.

The current OpenAI Python client accepts `Callable[[], Awaitable[str]]` for `AsyncOpenAI(api_key=...)` and refreshes that credential before requests. The voice provider already passes its stored `api_key` through unchanged, and streamed STT now refreshes dynamic credentials before its manual WebSocket handshake, but the public provider constructor still types `api_key` as `str | None`. Valid rotating-credential configurations therefore require a type-ignore or a manually constructed client.

### Fix

- widen the voice provider's `api_key` parameter to `str | Callable[[], Awaitable[str]] | None`
- document that an async API-key provider is accepted
- leave client construction and static-key behavior unchanged

### Test plan

Added a focused regression that supplies an async key provider through the public `OpenAIVoiceModelProvider` constructor and verifies the exact callable is forwarded to `AsyncOpenAI`.

The regression intentionally uses the public constructor without a cast, so repository typechecking covers the contract in addition to the runtime assertion.

The branch is based directly on current `main` at `e26a7d8aed59141ee13fb0a1fa16445017b0ccf1` and is 0 commits behind it.

### Risk

Very low. String API keys and default-key resolution are unchanged. This widens only the accepted type for a value that the provider already forwards directly to `AsyncOpenAI`.

### Issue number

None. Found while checking dynamic-credential parity after streamed STT gained callable-key refresh support.

### DIFF
--- a/src/agents/voice/models/openai_model_provider.py
+++ b/src/agents/voice/models/openai_model_provider.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Awaitable, Callable
 from typing import Any
 
 import httpx2
@@ -38,7 +39,7 @@ class OpenAIVoiceModelProvider(VoiceModelProvider):
     def __init__(
         self,
         *,
-        api_key: str | None = None,
+        api_key: str | Callable[[], Awaitable[str]] | None = None,
         base_url: str | None = None,
         openai_client: AsyncOpenAI | None = None,
         organization: str | None = None,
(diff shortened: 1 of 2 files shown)


seratch:
Thanks, but we won't enable this for now.

sylvesterkaczmarek:
> Thanks, but we won't enable this for now.

Understood, thanks. To help me focus future contributions on work that is actually useful to the project, is there a maintained list of issues or a few current areas you would be happy to accept external PRs for? 

I’m happy to work on bugs, tests, compatibility issues, or other areas you consider worthwhile. A pointer to the highest-value open items would help me avoid spending time implementing changes that are technically valid but not currently within the project’s direction.

PR #4546, opened 2026-08-20
fix: validate realtime playback durations

hsusul:
### Summary

`RealtimePlaybackTracker.on_play_ms` accepted negative, non-finite, and overflowing cumulative playback durations. A later interruption could then fail while converting the invalid elapsed time to the integer `audio_end_ms` sent to Realtime. This change validates each duration and the prospective cumulative total before mutating tracker state.

### Test plan

- Added focused regressions for negative, `NaN`, positive/negative infinity, and cumulative float overflow.
- `tests/realtime/test_playback_tracker.py`: 26 passed.
- Broader realtime tests: 105 passed; 9 local websocket integration tests were blocked by the sandbox refusing local socket binding.
- `.agents/skills/code-change-verification/scripts/run.sh`: format, lint, typecheck, and full tests passed.
- `git diff --check`: passed.

### Issue number

Not linked to an issue; this is a directly reproducible public playback-tracker boundary bug.

### Checks

### DIFF
--- a/src/agents/realtime/model.py
+++ b/src/agents/realtime/model.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import abc
+import math
 from collections.abc import Callable
 
 from typing_extensions import NotRequired, TypedDict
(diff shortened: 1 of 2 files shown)


seratch:
`RealtimePlaybackTracker.on_play_ms()` accepts the amount of audio actually played. Negative values, non-finite values, and cumulative durations near `1e308` are outside that semantic contract, and the PR does not show a normal playback implementation producing them.

The added tests establish that validation can be implemented, but they do not demonstrate an unmet supported use case or exercise the claimed interruption failure through the public session path. I am going to close this PR for now. If a realistic playback integration can produce an invalid duration during ordinary use, please provide that code path and its observable session impact so we can reconsider the narrowest appropriate boundary.

PR #4547, opened 2026-08-20
fix: bound non-finite shell output limits

hsusul:
### Summary

Malformed shell tool payloads with non-finite `max_output_length` values (`NaN` or infinities) previously escaped coercion as `ValueError`/`OverflowError`, and negative infinity could disable output truncation. This change bounds all non-finite numeric limits to zero while preserving integer limits, including arbitrarily large integers.

### Test plan

- Added focused regressions for `NaN`, positive/negative infinity, and a 400-digit integer.
- `tests/test_shell_call_serialization.py` and `tests/test_shell_tool.py`: 49 passed.
- `.agents/skills/code-change-verification/scripts/run.sh`: format, lint, typecheck, and full tests passed.
- `git diff --check`: passed.

### Issue number

Not linked to an issue; this is a directly reproducible malformed model-payload bug.

### Checks

### DIFF
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -11,6 +11,7 @@ import dataclasses
 import functools
 import inspect
 import json
+import math
 from collections.abc import Awaitable, Callable, Mapping, Sequence
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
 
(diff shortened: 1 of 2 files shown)


seratch:
On the current base, these values fail during normalization before the shell executor runs. This patch instead converts them to zero, executes the command, and silently discards its output, so it weakens the failure behavior rather than fixing a supported case.

I am going to close this PR. If a supported provider is observed emitting this payload, please provide a sanitized real response and its provider/version. The appropriate SDK behavior would then be an actionable ModelBehaviorError before execution, not coercion to a valid zero limit.

PR #4556, opened 2026-08-21
fix(tracing): sanitize non-finite span payloads

### LINKED ISSUE #4551: tracing: non-finite span input/output bypasses JSON sanitization
## Problem

`BackendSpanExporter` sanitizes OpenAI tracing `span_data.input` and `span_data.output`, including dropping non-JSON-compatible values. However, `_truncate_span_field_value()` returns early whenever `_value_json_size_bytes(value)` is below the size limit.

`_value_json_size_bytes()` currently uses Python's permissive `json.dumps()` defaults. That means `NaN`, `Infinity`, and `-Infinity` are treated as small serializable values and bypass `_sanitize_json_compatible_value()`, even though the sanitizer itself explicitly rejects non-finite floats.

A small payload such as:

```python
{
    "object": "trace.span",
    "span_data": {
        "type": "generation",
        "input": {"score": float("nan")},
    },
}
```

therefore reaches the exporter with the non-finite value intact. Strict JSON encoders reject these values, so a single affected span can make the tracing request fail instead of being normalized at the existing ingest boundary.

The behavior is also size-dependent: the same kind of invalid nested value can be cleaned when the containing field is large enough to enter the sanitizer, but not when it is small.

## Expected behavior

OpenAI tracing sanitization should never pass non-finite floats through `input` or `output`. The size check should use strict JSON semantics so invalid values fall through to the existing sanitizer.

## Suggested fix

Use `allow_nan=False` in `_value_json_size_bytes()`. On `ValueError`, retain the existing sentinel size so `_truncate_span_field_value()` invokes `_sanitize_json_compatible_value()`. This keeps normal finite payloads unchanged and reuses the current sanitizer rather than adding another traversal.

Regression coverage should include nested `NaN`, positive infinity, and negative infinity, plus a finite control case.

sylvesterkaczmarek:
This pull request fixes OpenAI tracing normalization so non-finite float values cannot bypass the existing JSON sanitizer merely because the surrounding `input` or `output` field is small.

`BackendSpanExporter._value_json_size_bytes()` now measures payloads using strict JSON semantics with `allow_nan=False`. A field containing `NaN`, `Infinity`, or `-Infinity` therefore follows the existing unserializable-value path, where `_sanitize_json_compatible_value()` removes the invalid nested value before the tracing payload is sent. Finite JSON payloads keep the existing fast path and are unchanged.

The change deliberately reuses the current OpenAI-ingest sanitization boundary instead of introducing another recursive traversal. Focused regression coverage exercises both `input` and `output`, all three non-finite float forms, preservation of valid sibling data, and a finite control case.

This pull request resolves #4551.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -320,7 +320,12 @@ class BackendSpanExporter(TracingExporter):
 
     def _value_json_size_bytes(self, value: Any) -> int:
         try:
-            serialized = json.dumps(value, ensure_ascii=False, separators=(",", ":"))
+            serialized = json.dumps(
+                value,
+                ensure_ascii=False,
+                separators=(",", ":"),
+                allow_nan=False,
+            )
         except (TypeError, ValueError):
             return self._OPENAI_TRACING_MAX_FIELD_BYTES + 1
         return len(serialized.encode("utf-8"))
--- a/tests/test_trace_processor_nonfinite_fields.py
+++ b/tests/test_trace_processor_nonfinite_fields.py
@@ -1,0 +1,39 @@
+import math
+
+import pytest
+
+from agents.tracing.processors import BackendSpanExporter
+
+
+@pytest.mark.parametrize("non_finite", [float("nan"), float("inf"), float("-inf")])
+@pytest.mark.parametrize("field_name", ["input", "output"])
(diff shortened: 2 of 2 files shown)


seratch:
Closing this for the reason at https://github.com/openai/openai-agents-python/issues/4551#issuecomment-5365257830

PR #4565, opened 2026-08-21
fix(provider): reuse falsy websocket models

hsusul:
### Summary

`OpenAIProvider` uses truthiness to detect a cached websocket model. A valid `Model` instance with falsey truthiness is therefore recreated on every lookup in the same running loop. Check cache-key presence instead so all cached model instances are reused.

### Test plan

- Added regression coverage for a falsey websocket model and identity reuse.
- Focused: `2 passed` (`tests/test_config.py -k falsy_websocket_model or scopes_websocket_model_cache`).
- Full verification: `UV_CACHE_DIR=/tmp/uv-verify-pr5-b2 UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh` — all commands passed (format, lint, typecheck, tests).

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/models/openai_provider.py
+++ b/src/agents/models/openai_provider.py
@@ -259,8 +259,8 @@ class OpenAIProvider(ModelProvider):
                 if running_loop is not None
                 else None
             )
-            if loop_cache is not None and (cached_model := loop_cache.get(cache_key)):
-                return cached_model
+            if loop_cache is not None and cache_key in loop_cache:
+                return loop_cache[cache_key]
         client = self._get_client()
         model: Model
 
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. I traced this cache through OpenAIProvider.get_model(): the cached value is always the SDK-owned concrete OpenAIResponsesWSModel, which does not define falsey truthiness. The regression test reaches the branch only by replacing that internal class with a falsey test double, rather than through a supported provider or model configuration.

I am going to close this PR because the demonstrated case is outside the supported runtime contract. We can revisit it if a supported construction path or a real OpenAIResponsesWSModel instance can become falsey.

PR #4566, opened 2026-08-21
fix(memory): skip malformed encrypted session envelopes

hsusul:
### Summary

`EncryptedSession` documents that invalid or expired stored envelopes are skipped, but malformed payload types, invalid UTF-8, invalid JSON, and valid JSON with a non-object shape could abort session reads. Harden `_unwrap` to skip those malformed envelopes while preserving valid response-item mappings.

### Test plan

- Added regression coverage for non-string, lone-surrogate, and non-object encrypted payloads.
- Focused: `21 passed` (`tests/extensions/memory/test_encrypt_session.py`).
- Full verification: `UV_CACHE_DIR=/tmp/uv-verify-pr5-a2 UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh` — all commands passed (format, lint, typecheck, tests).

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/extensions/memory/encrypt_session.py
+++ b/src/agents/extensions/memory/encrypt_session.py
@@ -165,10 +165,22 @@ class EncryptedSession(SessionABC):
             return cast(TResponseInputItem, item)
 
         try:
-            token = item["payload"].encode("utf-8")
+            payload = item["payload"]
+            if not isinstance(payload, str):
+                return None
+            token = payload.encode("utf-8")
             plaintext = self.cipher.decrypt(token, ttl=self.ttl)
-            return cast(TResponseInputItem, _from_json_bytes(plaintext))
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this. Normal expiry, key mismatch, and ciphertext corruption already fail Fernet authentication and are skipped through the existing InvalidToken handling. The new branches require a structurally corrupted typed envelope or plaintext that was validly encrypted with the session key but was not produced by EncryptedSession._wrap().

I am going to close this PR for now because no supported backend, migration, or user report has been shown to produce those shapes. If you can provide a concrete stored record from a supported session backend and trace how ordinary SDK operation produced it, we can reconsider the narrow corruption boundary that needs handling.

PR #4567, opened 2026-08-21
fix: preserve native Modal directory snapshots with disjoint mounts

Filimoa:
### Summary

This pull request fixes Modal `workspace_persistence="snapshot_directory"` unnecessarily falling back to tar whenever any manifest mount cannot be detached, even when that mount is completely outside the directory being snapshotted.

For example, a workspace rooted at `/workspace` with a Modal cloud-bucket mount at `/mnt/remote` is a valid, disjoint topology. Modal's cloud-bucket strategy correctly reports that it cannot detach the mount, but the shared fallback check previously ignored the mount target and therefore downgraded `Sandbox.snapshot_directory("/workspace")` to tar. Because the mount cannot be included in that directory snapshot, its detach capability is irrelevant.

The fallback decision now accepts the resolved directory snapshot root and consults `supports_native_snapshot_detach()` only for mounts whose normalized POSIX targets overlap that root. Equality, descendants, and ancestors are treated as overlap. Filesystem-wide snapshot callers continue to inspect every mount.

The Modal adapter also preserves the no-copy boundary around mounted storage:

- Detachable overlapping mounts are detached and restored around persistence.
- Modal cloud-bucket mounts equal to, below, or containing the workspace root remain unsupported and are rejected before provider side effects, including trusted-manifest resume.
- Disjoint cloud-bucket mounts remain manifest-owned ephemeral storage and are recreated when a fresh sandbox restores the native workspace image.
- Mount contents, credentials, authority, native snapshot reference formats, restore behavior, and existing tar behavior are unchanged.

### Test plan

- Added generic fallback tests covering disjoint, equal, descendant, ancestor, normalized, double-slash, detachable, filesystem-wide, and Windows-flavored typed paths.
- Added Modal provider-fake tests covering native directory-snapshot references for disjoint cloud mounts, overlapping rejection, detach/restore behavior, trusted resume validation, and fresh restore recreation.
- Ran `uv run pytest -q tests/sandbox/test_session_utils.py tests/extensions/sandbox/test_modal.py` (`155 passed`).
- Ran `.agents/skills/code-change-verification/scripts/run.sh` with the Codex sandbox environment and a writable temporary uv cache; formatting, linting, type checking, and the full test suite passed.
- Completed the required independent final review with two clean reviewers on the final patch fingerprint.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/extensions/sandbox/modal/sandbox.py
+++ b/src/agents/extensions/sandbox/modal/sandbox.py
@@ -88,6 +88,7 @@ from ....sandbox.util.retry import (
 from ....sandbox.util.tar_utils import UnsafeTarMemberError, validate_tar_bytes
 from ....sandbox.workspace_paths import (
     coerce_posix_path,
+    normalize_posix_path,
     posix_path_as_path,
     posix_path_for_error,
     sandbox_path_str,
@@ -1421,7 +1422,7 @@ class ModalSandboxSession(BaseSandboxSession):
         assert self._sandbox is not None
         if not hasattr(self._sandbox, "snapshot_directory"):
             return await self._persist_workspace_via_tar()
-        if self._native_snapshot_requires_tar_fallback():
+        if self._native_snapshot_requires_tar_fallback(snapshot_root=root):
             return await self._persist_workspace_via_tar()
         plain_skip = self._modal_snapshot_plain_skip_relpaths(root)
         skip_abs = [root / rel for rel in sorted(plain_skip, key=lambda p: p.as_posix())]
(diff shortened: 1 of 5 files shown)


seratch:
Thanks for the substantial investigation. The current behavior falls back to tar when a non-detachable mount is present, so the workspace remains persistable and restorable in the reported disjoint-mount topology. The PR does not establish a user-visible failure or material operational cost from that fallback, while the change expands shared path normalization, native snapshot selection, mount restoration, resume validation, and provider-specific lifecycle behavior.

I am going to close this PR for now. We can revisit a narrower Modal change if there is a concrete supported workload where the tar fallback causes a measured failure or material latency, size, or resource problem.

PR #4568, opened 2026-08-21
fix(items): strip created_by when replaying RunItems as input

hsusul:
### Summary

This pull request fixes a gap left by #4308: `ModelResponse.to_input_items()` already strips output-only `created_by` before Responses replay, but public RunItem continuation still leaked it.

`RunItem.to_input_item()` dumped function-call and tool-output payloads with `created_by` intact. That same conversion feeds `run_item_to_input_item()`, session persistence, and `RunResult.to_input_list()`, so a later Responses request can include a field that is not part of the input schema.

The base converter now routes through `_output_item_to_input_item()`, and the dict path on `ToolCallOutputItem` pops `created_by` from the copied payload without mutating `raw_item`.

### Test plan

- [x] Added tests for `ToolCallItem.to_input_item`, `ToolCallOutputItem.to_input_item`, and `run_item_to_input_item` on function calls/outputs that carry `created_by`
- [x] `uv run pytest tests/test_items_helpers.py tests/test_run_internal_items.py`
- [x] `make format`, `make lint`, `make typecheck`
- [ ] Full `make tests` — one pre-existing local failure on this host, unrelated to this change: `tests/sandbox/test_run_cwd.py::test_python_skill_uses_absolute_root_from_nested_workdir` (`WorkspaceReadNotFoundError` for `tasks/task-a/nested/output.txt`). The same failure occurs on unpatched `upstream/main`.

### Issue number

N/A — small fix, no existing issue.

### Checks

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -484,8 +484,7 @@ class ToolCallOutputItem(RunItemBase[Any]):
                         if isinstance(outcome, dict):
                             if outcome.get("type") == "exit":
                                 entry["outcome"] = outcome
-            payload.pop("created_by", None)
-            return cast(TResponseInputItem, payload)
+            return _output_item_to_input_item(payload)
 
         return super().to_input_item()
 
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -150,14 +150,7 @@ class RunItemBase(Generic[T], abc.ABC):
 
     def to_input_item(self) -> TResponseInputItem:
         """Converts this item into an input item suitable for passing to the model."""
-        if isinstance(self.raw_item, dict):
-            # We know that input items are dicts, so we can ignore the type error
-            return self.raw_item  # type: ignore
-        elif isinstance(self.raw_item, BaseModel):
-            # All output items are Pydantic models that can be converted to input items.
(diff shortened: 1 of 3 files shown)


seratch (changes requested):
Thanks for addressing the RunItem replay path. The overall direction is right, but one dict-backed path still retains output-only metadata.

I reproduced this on the current head with a `ToolCallOutputItem` whose `raw_item` is a `shell_call_output` mapping. `to_input_item()` removes the top-level `created_by`, but `output[*].created_by` remains in the replayed payload because this branch returns its shallow-copied payload directly and bypasses the nested cleanup in `_output_item_to_input_item()`.

Please keep the existing dict-specific cleanup for `status`, `shell_output`, and `provider_data`, then route the cleaned payload through the shared output-item sanitizer. Please also add a regression test covering a dict-backed `ToolCallOutputItem` and asserting that:

- both the top-level and nested `created_by` fields are removed;
- the existing shell-output cleanup is preserved; and
- the original `raw_item`, including its nested chunk mapping, is not mutated.

Once that case is covered, this should be ready.

seratch (approved):

PR #4569, opened 2026-08-21
fix(tracing): reject zero batch sizes

hsusul:
## Summary
- reject `max_batch_size=0` during BatchTraceProcessor construction
- prevent force_flush from silently leaving queued spans unexported

## Test plan
- [x] `UV_CACHE_DIR=/tmp/uv-pr6-b uv run pytest tests/test_trace_processor.py -q` (58 passed)
- [x] changed-file Pyright and mypy
- [x] full suite: 9031 passed, 39 skipped; 38 environment-restricted failures (forkserver/native macOS sandbox)

## Issue number
N/A

## Checks
- [x] Focused regression tests added


seratch:
We'd like to hold off adding this type of obvious value error validation for now.

PR #4570, opened 2026-08-21
fix(sandbox): validate ephemeral directory child paths

hsusul:
## Summary
- validate every descendant path before applying an ephemeral directory
- prevent only_ephemeral manifest application from bypassing nested path safety

## Test plan
- [x] `UV_CACHE_DIR=/tmp/uv-pr6-a uv run pytest tests/sandbox/test_manifest_application.py -q` (12 passed)
- [x] changed-file Pyright and mypy
- [x] full suite: 9031 passed, 39 skipped; 38 environment-restricted failures (forkserver/native macOS sandbox)

## Issue number
N/A

## Checks
- [x] Focused regression test added

### DIFF
--- a/src/agents/sandbox/session/manifest_application.py
+++ b/src/agents/sandbox/session/manifest_application.py
@@ -100,6 +100,12 @@ class ManifestApplier:
         manifest_rel = Manifest._coerce_rel_path(rel_dest)
         Manifest._validate_rel_path(manifest_rel)
         if artifact.ephemeral:
+            if isinstance(artifact, Dir):
+                for child_name in artifact.children:
+                    self._validate_ephemeral_child_paths(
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this path. The helper does skip nested manifest validation when an entire ephemeral directory is selected, but the production application path does not let that escape the workspace. Each nested Dir or File operation reaches session.mkdir() or session.write(), which normalizes the destination through WorkspacePathPolicy and rejects paths outside the workspace. A path that normalizes elsewhere inside the workspace does not gain authority beyond what a top-level manifest entry already has.

I am going to close this PR because the test demonstrates a helper-level inconsistency, not a supported security-boundary violation or meaningful downstream effect. We can reconsider if a real sandbox backend can bypass the session path policy.

PR #4571, opened 2026-08-21
fix(voice): use monotonic STT event deadlines

hsusul:
## Summary
- use `time.monotonic()` for OpenAI STT event timeout accounting
- update the timeout regression test to control the monotonic clock

## Testing
- `python -m compileall -q` on changed source and test files
- focused pytest was attempted but this fresh worktree could not import the repository because optional dependency `httpx2` was unavailable; dependency sync was blocked by package-index DNS failures


seratch (changes requested):
Can you fix the CI errors?

seratch:
This issue will be resolved by https://github.com/openai/openai-agents-python/pull/4593

PR #4572, opened 2026-08-21
fix(sandbox): retain PTY output queued at timeout

hsusul:
## Summary
- drain shared, E2B, Cloudflare, and Modal PTY collectors after deadline/timeout
- preserve output queued between the final poll and timeout
- add a deterministic regression test for timeout-tail output

## Testing
- `python -m compileall -q` on changed source and test files
- focused pytest was attempted but this fresh worktree could not import the repository because optional dependency `httpx2` was unavailable; dependency sync was blocked by package-index DNS failures

### DIFF
--- a/src/agents/extensions/sandbox/cloudflare/sandbox.py
+++ b/src/agents/extensions/sandbox/cloudflare/sandbox.py
@@ -1059,6 +1059,11 @@ class CloudflareSandboxSession(BaseSandboxSession):
                 break
             entry.output_notify.clear()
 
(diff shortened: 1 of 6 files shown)


seratch (changes requested):
Can you fix all the review comments?

seratch:
The timeout-tail work is now consolidated in #4738 together with UTF-8 carry and output-settlement ownership. This older PR overlaps that implementation, so I am closing it as superseded.

The remaining verification and fixes will continue in #4738; closing this PR does not mean the fix has shipped.

PR #4587, opened 2026-08-22
fix: strip defaults from strict JSON schemas

sylvesterkaczmarek:
Fixes #4390.

## Summary

Strict schema conversion currently removes `default` only when the value is `None`. Non-null defaults such as enum, string, integer, or boolean values therefore remain in the generated strict schema even though the Structured Outputs schema subset rejects `default`.

This change:
- removes `default` from every schema node during strict conversion;
- removes the now-unused `NOT_GIVEN` import;
- preserves the existing behavior of marking object properties as required;
- adds regression coverage for `None`, string, integer, boolean, and Pydantic enum defaults.

## Validation

The branch is based directly on current upstream `main` at `5b8f6c71747e5feef143c7047a165ebadf0a021d` and contains one focused commit. Full repository tests were not run locally in this environment; upstream CI should run the strict-schema test suite.

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -3,8 +3,6 @@ from __future__ import annotations
 import copy
 from typing import Any, TypeGuard, cast
 
-from openai import NOT_GIVEN
-
 from .exceptions import UserError
 
 _EMPTY_SCHEMA = {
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for following up. This PR reintroduces the global strict-schema behavior change from #4529, but #4390 still does not include the requested default-isolated Azure comparison. The direct OpenAI comparison accepted both schemas, and omission of default from the documented subset does not by itself establish that stripping it across every SDK caller is the correct compatibility behavior. This helper also controls handoffs, structured outputs, Realtime, MCP, agent-as-tool, Codex tools, public parameter schemas, and serialized RunState metadata.

I am going to close this PR. Please avoid repeating this.

PR #4596, opened 2026-08-23
Remove defaults from strict output schemas

medisean:
Closes #4390

Summary:
- Remove property defaults from strict output schemas, including non-null defaults.
- Add regression coverage for the generated schema.

Tests:
- Python syntax compilation passed.


seratch:
Please check https://github.com/openai/openai-agents-python/pull/4587#issuecomment-5380594698

PR #4598, opened 2026-08-23
fix(realtime): preserve item status on retrieved conversation items

### LINKED ISSUE #4597: realtime: conversation.item.retrieved regresses completed history items to in_progress
### Please read this first

- **Have you read the docs?** [Agents SDK docs](https://openai.github.io/openai-agents-python/) — yes, including the Realtime sections.

### Describe the bug

`conversation_item_to_realtime_message_item` in `src/agents/realtime/openai_realtime.py` hardcodes `"status": "in_progress"` when converting a server `ConversationItem`, so the status carried by the server payload is dropped.

This is reachable in a default voice session. After the assistant produces audio, `_current_item_id` points at the assistant's audio item (set in `_handle_audio_delta`). When the user's next utterance finishes transcription, the handler for `conversation.item.input_audio_transcription.completed` sends `conversation.item.retrieve` for that item, and the server replies with `conversation.item.retrieved` carrying the item's actual status (`completed`). The hardcoded conversion rewrites that to `in_progress`, and the session history merge keeps the incoming status, so the completed assistant item regresses to `in_progress` in history after every user turn, with nothing to restore it. The same retrieve is also triggered by `conversation.item.truncated`.

The GA conversation item models do carry `status: Optional[Literal["in_progress", "completed", "incomplete"]]`, so the value is available on the parsed item and just needs to be passed through.

Effect: history items flip back to `in_progress` permanently, and consumers watching `RealtimeHistoryUpdated` see stale statuses plus a spurious extra update per user turn.

### Debug information

- Agents SDK version: 0.22.0 (also present on current main)
- Python version: 3.12
- Operating system: Linux
- Model and model provider: any realtime model; the repro below is offline with a stubbed websocket
- Does the issue occur consistently or intermittently? Consistently — it fires on every user utterance that follows assistant audio when input transcription is enabled.

### Repro steps

```python
import asyncio
import base64
import json
from unittest.mock import AsyncMock, Mock

from agents.realtime.agent import RealtimeAgent
from agents.realtime.items import AssistantMessageItem
from agents.realtime.openai_realtime import OpenAIRealtimeWebSocketModel


async def main() -> None:
    model = OpenAIRealtimeWebSocketModel()
    ws = Mock()
    ws.send = AsyncMock()
    model._websocket = ws

    session = RealtimeSession(model, RealtimeAgent(name="agent"), None)
    model.add_listener(session)

    async def feed(event: dict) -> None:
        await model._handle_ws_event(event)

    # Turn 1: assistant audio response completes.
    await feed({"type": "response.created", "event_id": "e1", "response": {"id": "resp_1"}})
    await feed({
        "type": "response.output_item.added",
        "response_id": "resp_1",
        "output_index": 0,
        "item": {
            "id": "item_A",
            "type": "message",
            "role": "assistant",
            "status": "in_progress",
            "content": [{"type": "output_audio", "transcript": None}],
        },
    })
    await feed({
        "type": "response.output_audio.delta",
        "event_id": "e3",
        "response_id": "resp_1",
        "item_id": "item_A",
        "content_index": 0,
        "output_index": 0,
        "delta": base64.b64encode(b"\x00\x00" * 2400).decode(),
    })
    await feed({
        "type": "response.output_item.done",
        "response_id": "resp_1",
        "output_index": 0,
        "item": {
            "id": "item_A",
            "type": "message",
            "role": "assistant",
            "status": "completed",
            "content": [{"type": "output_audio", "transcript": "hello there"}],
        },
    })
    await feed({"type": "response.done", "event_id": "e5", "response": {"id": "resp_1"}})

    item = session._history[0]
    assert isinstance(item, AssistantMessageItem)
    print("after turn 1:", item.status)  # completed

    # Turn 2: user speaks; transcription completing triggers a retrieve of item_A.
    await feed({
        "type": "conversation.item.input_audio_transcription.completed",
        "event_id": "e6",
        "item_id": "item_B",
        "content_index": 0,
        "transcript": "user speech",
        "usage": {"type": "tokens", "input_tokens": 1, "output_tokens": 1, "total_tokens": 2},
    })
    await feed({
        "type": "conversation.item.retrieved",
        "event_id": "e7",
        "item": {
            "id": "item_A",
            "type": "message",
            "role": "assistant",
            "status": "completed",
            "content": [{"type": "output_audio", "transcript": "hello there"}],
        },
    })

    item = session._history[0]
    print("after retrieve:", item.status)  # expected completed, actual in_progress


asyncio.run(main())
```

Output on 0.22.0:

```text
after turn 1: completed
after retrieve: in_progress
```

The server explicitly reported `status: "completed"` in the retrieved payload, but the history entry ends up `in_progress`.

shoutoutuoadi325:
### Summary

`conversation_item_to_realtime_message_item` hardcoded `status: "in_progress"`, discarding the status carried by the server item. When the SDK retrieves the last assistant audio item after `conversation.item.input_audio_transcription.completed` (or `conversation.item.truncated`), the server replies `conversation.item.retrieved` with the item's real status, usually `completed`. The conversion rewrote that to `in_progress`, and the session history merge keeps the incoming status, so a completed history entry regressed to `in_progress` after every user turn, with nothing to restore it.

The conversion now passes `item.status` through and falls back to `in_progress` when the server omits it, so streaming `conversation.item.created` events that carry no status behave exactly as before.

### Test plan

- [x] `uv run pytest tests/realtime/test_item_parsing.py` (status preserved; fallback to `in_progress` when absent)
- [x] `uv run pytest tests/realtime/test_openai_realtime.py::TestEventHandlingRobustness::test_retrieved_completed_item_keeps_status` (audio delta → transcription completed triggers the retrieve → retrieved item keeps `completed`)
- [x] Both new tests fail on unpatched `main` and pass with the fix
- [x] `uv run pytest tests/realtime/` (518 passed)
- [x] `.agents/skills/code-change-verification/scripts/run.sh` (format, lint, typecheck, full test suite)

### Issue number

Fixes #4597

### Checks


leadingproblemsolver (commented):
The converter fix looks right, but I think the regression test still stops one boundary too early. `test_retrieved_completed_item_keeps_status` asserts the emitted `RealtimeModelItemUpdatedEvent`; it never feeds that event through `RealtimeSession`, which is where the existing history item is actually reconciled/replaced.

That means this test could stay green while a later merge-path regression still turns a known terminal item into `in_progress` or drops content evidence.

I’d add one session-level fixture with an existing assistant history item (`status="completed"`, transcript present), then process a retrieved update and assert the final `session._history` item preserves both lifecycle state and transcript. Parameterizing `completed` / `incomplete` would cover both terminal statuses.

There is one subtle edge worth making intentional too: `item.status or "in_progress"` is safe for create-like events with no status, but `ConversationItem.status` is optional. If a *retrieved* item ever arrives with `status=None`, this fallback can still regress an already-known terminal history item. Either assert the protocol guarantees status on retrieve, or make the session reconciliation monotonic when incoming status is absent.

Stronger contract: a refresh may add information, but it should not erase already-known terminal lifecycle state unless the protocol explicitly reports a new state.

seratch (changes requested):
Thanks for the focused fix. The production change correctly preserves an explicit server status while retaining the released missing-status fallback.

Before merge, please extend the regression coverage through RealtimeSession: attach a session listener, seed its history with a completed assistant item containing a transcript, feed a conversation.item.retrieved event that reports completed while omitting the transcript, and assert that the final history item retains both status="completed" and the existing transcript. The current test stops at RealtimeModelItemUpdatedEvent, so it does not lock down the user-visible history outcome reported in #4597.

Please keep the current item.status or "in_progress" behavior. Changing omitted-status semantics or adding a new monotonic session-state rule would widen this fix beyond the demonstrated case.

shoutoutuoadi325:
Thanks @seratch! I've added the requested session-level coverage in `tests/realtime/test_openai_realtime.py` (`test_retrieved_completed_item_keeps_status_in_session_history`): it attaches a `RealtimeSession` listener to the real `OpenAIRealtimeWebSocketModel`, seeds its history with a completed assistant item containing a transcript, feeds a raw `conversation.item.retrieved` event that reports `status="completed"` while omitting the transcript, and asserts the final history item retains both `status="completed"` and the existing transcript.

I also verified the new test fails against the pre-fix converter (`'in_progress' != 'completed'`), so it locks down the #4597 outcome end to end. The `item.status or "in_progress"` behavior is unchanged, and I've intentionally not added any monotonic session-state rule.

seratch (approved):

PR #4604, opened 2026-08-23
Fix missing httpx2 dep due to unconditional import

timkpaine:
You have an unconditional import hoisted to the top level `agents` import here: https://github.com/openai/openai-agents-python/blob/233467994fac7e7dbd868931573cc9a4302c0a16/src/agents/tracing/processors.py#L14

xref:
https://github.com/conda-forge/openai-agents-feedstock/actions/runs/32274620397/job/96138993962#step:4:1314

added the dep and alphabetized your deps.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,13 +7,14 @@ requires-python = ">=3.10"
 license = "MIT"
 authors = [{ name = "OpenAI", email = "support@openai.com" }]
 dependencies = [
+    "griffelib>=2, <3",
+    "httpx2",
+    "mcp>=1.19.0, <3; python_version >= '3.10'",
     "openai>=3.0.0,<4",
     "pydantic>=2.12.2, <3",
-    "griffelib>=2, <3",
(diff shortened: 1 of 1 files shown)


seratch:
httpx2 is imported directly by the SDK, but the supported dependency graph already installs it through the required openai>=3,<4 dependency, which declares httpx2>=2.7.0,<3. Current MCP 2.x releases also depend on httpx2.

The linked conda-forge failure therefore does not establish that a normal installation from the SDK metadata is missing httpx2. Adding it directly would make the dependency ownership more explicit, but it would not fix a demonstrated user-facing gap.

I am going to close this PR for now. If there is a supported installer reproduction that installs the declared OpenAI and MCP dependencies but still omits httpx2, we can revisit the packaging boundary with that evidence.

timkpaine:
> httpx2 is imported directly by the SDK

In python, you usually specify your direct dependencies as direct dependencies. Relying on transitive dependency resolution is generally considered bad practice.

PR #4605, opened 2026-08-23
fix: strip non-null defaults from strict JSON schemas

AAliKKhan:
### Summary

Tools and output models built from Pydantic models with non-null field defaults currently produce strict schemas containing `"default"`. The OpenAI/Azure Structured Outputs API rejects these at call time:

```
'default' is not permitted within a property definition
```

This forces users to disable `strict_mode` entirely for any tool whose parameters include a non-null default (enum member, int, string, etc.), losing structured-output schema guarantees.

### Problem & root cause

`_ensure_strict_json_schema()` only stripped `"default"` when its value was exactly `None`. However, strict mode already forces every property into `required` (simulating optionality via nullability), so any remaining default is unreachable by construction — the model must always supply an explicit value. The `is None` check was an arbitrary distinction that left dead-weight keys in the schema that the API then rejects outright.

### Solution

Strip `default` unconditionally at each schema node (`json_schema.pop("default", None)`), and remove the now-unused `NOT_GIVEN` import. Python-side behavior is unchanged: defaults still apply locally via `to_call_args` / Pydantic validation; they are simply no longer emitted into wire schemas where the API forbids them.

### Test plan

- [x] New regression test `test_non_null_default_removal` verifying string/integer defaults are stripped while all properties stay required
- [x] Updated four function-schema tests that asserted the previous preservation behavior
- [x] Verified Python-side defaults still apply through `to_call_args`
- [x] Focused suites pass (`test_strict_schema.py`, `test_function_schema.py`, `test_strict_schema_oneof.py`, tool tests)
- [x] Full suite, `ruff format`/`ruff check`, `mypy src/agents`, and `pyright` all pass

Fixes #4390

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -3,8 +3,6 @@ from __future__ import annotations
 import copy
 from typing import Any, TypeGuard, cast
 
-from openai import NOT_GIVEN
-
 from .exceptions import UserError
 
 _EMPTY_SCHEMA = {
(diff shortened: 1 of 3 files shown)


seratch:
Please check the discussions in the past PRs. We don't plant to make this change for now.

PR #4608, opened 2026-08-23
fix(items): preserve replayable shell output status

hsusul:
### Summary

Preserve API-supported `shell_call_output.status` values when a dict-backed `ToolCallOutputItem` is replayed as model input.

The input schema now accepts `in_progress`, `completed`, and `incomplete`, and `ModelResponse.to_input_items()` already preserves those values. `ToolCallOutputItem.to_input_item()` still unconditionally removed the field, so replay behavior differed depending on which item representation reached the runner. Local-only values such as the shell executor's `failed` status remain stripped, along with `shell_output`, `provider_data`, and `created_by` metadata.

### Test plan

- Added regression coverage for all three input-supported statuses and retained coverage that strips the local-only `failed` status.
- Focused item/model/shell tests: 139 passed, then 6 targeted regression cases passed after the final assertion update.
- Verification wrapper: formatting and lint passed; mypy and Pyright passed; parallel tests passed (9,060 passed, 55 skipped).
- Non-review-optional serial tests passed (76 passed, 4 skipped, 63 deselected).
- The wrapper's Docker-backed Dapr/Redis review-optional phase stalled while starting its testcontainers and was terminated after more than six minutes. This integration path does not exercise item replay or shell output conversion.
- `git diff --check` passed.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -462,9 +462,8 @@ class ToolCallOutputItem(RunItemBase[Any]):
     def to_input_item(self) -> TResponseInputItem:
         """Converts the tool output into an input item for the next model turn.
 
-        Hosted tool outputs (e.g. shell/apply_patch) carry a `status` field for the SDK's
-        book-keeping, but the Responses API does not yet accept that parameter. Strip it from the
-        payload we send back to the model while keeping the original raw item intact.
+        Hosted tool outputs can carry SDK-only metadata that the Responses API does not accept.
+        Strip that metadata while keeping input-supported fields and the original raw item intact.
         """
 
         if isinstance(self.raw_item, dict):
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for investigating this. The current Responses input type now permits shell_call_output.status, but the field remains optional, and this PR currently demonstrates only a difference in replayed dictionary shape. It does not show a supported provider or runner path where preserving in_progress, completed, or incomplete changes a model decision, persisted interpretation, or user-visible result.

I am going to close this PR for now. If a real Responses or supported provider reproduction shows that omitting the status changes downstream behavior, we can revisit the narrow replay contract with that evidence.

PR #4609, opened 2026-08-23
fix: preserve annotated_types constraints in function schemas

AAliKKhan:
### Summary

`annotated_types` constraints declared via `typing.Annotated` (for example `Annotated[int, Gt(0)]`, `Annotated[str, MinLen(3)]`, or `StringConstraints`) are currently **silently dropped**: they appear neither in the advertised tool JSON schema nor in the runtime Pydantic validation of tool arguments. The equivalent `Field(gt=0)` spelling works, so behavior differs between two forms users reasonably expect to be identical.

### Problem & impact

Repro on current `main`:

```python
@function_tool
def half(x: Annotated[int, Gt(0)]) -> int:
    return x // 2
```

- `half.params_json_schema["properties"]["x"]` is just `{"type": "integer"}` - no `exclusiveMinimum`
- Validation accepts out-of-contract input: the args model happily validates `x=-5`

The model is shown an unconstrained parameter and invalid values reach the user's function body unvalidated - a silent contract violation rather than a loud failure.

### Root cause

`_strip_annotated()` removes all `Annotated` metadata up front. When rebuilding the arguments model with `create_model()`, only string descriptions and `FieldInfo` instances from that metadata are re-applied; every other metadata item (`annotated_types.Gt`, `Ge`, `MinLen`, `StringConstraints`, ...) is discarded entirely.

### Solution

In the normal-parameter branch of `function_schema()`, re-wrap any non-`str`/non-`FieldInfo` metadata back into the annotation (`Annotated[ann, *constraint_metadata]`) before building the field. Pydantic v2 applies `Annotated` constraint metadata natively on model fields, so both the schema and validation now honor it. All existing merging behavior for `FieldInfo`, defaults, and docstring descriptions is untouched.

This also matches the documented contract in `.agents/references/function-and-output-schema.md`: "Preserve `Field` constraints, aliases, and defaults when merging `Annotated` metadata."

### Test plan

- [x] New test: `Annotated[int, Gt(0)]` / `Annotated[str, MinLen(3)]` produce `exclusiveMinimum`/`minLength` in the schema and reject violating inputs at validation time; valid calls still reconstruct correctly
- [x] New test: `annotated_types` constraints combined with a `Field` in the same `Annotated` keep both description and constraint
- [x] Full focused suites pass (`test_function_schema.py`, `test_function_tool.py`, `test_strict_schema.py`, converter/output tests)
- [x] `ruff format`, `ruff check`, `mypy src/agents`, and `pyright` all clean

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -434,6 +434,15 @@ def function_schema(
             metadata = param_metadata.get(name, ())
             field_info_from_annotated = _extract_field_info_from_metadata(metadata)
 
+            # Preserve constraint metadata (for example `annotated_types.Gt`) by re-wrapping it
+            # into the annotation; Pydantic applies `Annotated` constraints natively on model
+            # fields, so dropping them would silently weaken both the schema and validation.
+            constraint_metadata = tuple(
+                item for item in metadata if not isinstance(item, str | FieldInfo)
+            )
+            if constraint_metadata:
+                ann = Annotated[(ann, *constraint_metadata)]
+
             if field_info_from_annotated is not None:
                 merged = FieldInfo.merge_field_infos(
                     field_info_from_annotated,
--- a/tests/test_function_schema.py
+++ b/tests/test_function_schema.py
@@ -3,6 +3,7 @@ from enum import Enum
 from typing import Annotated, Any, Literal
 
 import pytest
+from annotated_types import Gt, MinLen
 from pydantic import BaseModel, Field, ValidationError
 from pydantic.json_schema import PydanticJsonSchemaWarning
 from typing_extensions import TypedDict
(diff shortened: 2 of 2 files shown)


seratch:
I confirmed that the current function-schema path drops outer `Annotated` metadata other than strings and `FieldInfo`, and that this patch restores that metadata for ordinary parameters.

However, constrained tool arguments are already supported through the documented `Annotated[T, Field(...)]` form, and this PR does not provide a concrete scenario where that supported form is insufficient. The current patch also leaves the same behavior unchanged for `*args` and `**kwargs`, while its implementation would preserve arbitrary non-string Pydantic metadata beyond the stated `annotated_types` constraints.

I am going to close this PR for now. If there is a concrete supported use case where `Field(...)` cannot express the required behavior, we can revisit the narrowest metadata contract needed for that case.

PR #4612, opened 2026-08-24
fix: use structural matching for serialized approval checkpoint boundary

### LINKED ISSUE #4611: Serialized approval checkpoints cannot be resumed with output guardrails when the interruption follows an earlier model response
### Describe the bug

When an agent has output guardrails and a tool-approval interruption occurs on a *later* model response (i.e., after any earlier completed model response in the same run), every serialized checkpoint of that run fails to resume:

```
UserError: Cannot resume a serialized approval checkpoint with output guardrails because the current response boundary cannot be proven. Start a new run from safe input.
```

This fires deterministically on `RunState.from_string(...)` + `Runner.run(agent, state)` - non-streamed, streamed, with or without a session. Resuming the identical flow **in-memory** works fine, so this breaks exactly the durable cross-process pause/resume boundary that `RunState.to_string()` exists for.

### Debug information

- Agents SDK version: 0.22.0 (also present on current `main`)
- Python 3.13, any OS
- Repro is offline (`ScriptedModel`), no API access needed

### Repro

```python
import asyncio

from openai.types.responses import ResponseFunctionToolCall, ResponseOutputMessage, ResponseOutputText

from agents import Agent, RunContextWrapper, RunState, Runner, function_tool, guardrail
from agents.memory import SQLiteSession


def get_text_message(text):
    return ResponseOutputMessage(
        id="msg_x",
        content=[ResponseOutputText(text=text, type="output_text", annotations=[])],
        role="assistant",
        type="message",
        status="completed",
    )


def get_function_tool_call(name, arguments, call_id):
        id="fc_x", arguments=arguments, name=name, type="function_call", call_id=call_id
    )


async def scenario(name, *, serialize=True, streamed=False, session=False):
    model = ScriptedModel()

    @function_tool(name_override="normal_tool")
    def normal_tool() -> str:
        return "normal ok"

    @function_tool(name_override="approval_tool", needs_approval=True)
    def approval_tool() -> str:
        return "approved ok"

    @guardrail.output_guardrail
    def passthrough(ctx: RunContextWrapper, agent: Agent, output) -> guardrail.GuardrailFunctionOutput:
        return guardrail.GuardrailFunctionOutput(output_info=None, tripwire_triggered=False)

    agent = Agent(
        name="test",
        model=model,
        tools=[normal_tool, approval_tool],
        output_guardrails=[passthrough],
    )
    model.extend(
        [
            [get_function_tool_call("normal_tool", "{}", "call_normal")],  # turn 1
            [get_function_tool_call("approval_tool", "{}", "call_appr")],  # turn 2 -> interruption
            [get_text_message("done")],
        ]
    )

    sess = SQLiteSession(f"probe_{name}") if session else None

    first = await Runner.run(agent, input="hello", session=sess)
    assert len(first.interruptions) == 1
    state = first.to_state()
    state.approve(first.interruptions[0])

    restored = await RunState.from_string(agent, state.to_string()) if serialize else state

    try:
        if streamed:
            result = Runner.run_streamed(agent, restored, session=sess)
                pass
            resumed = result
        else:
            resumed = await Runner.run(agent, restored, session=sess)
        print(f"[{name}] OK final_output={resumed.final_output!r}")
    except Exception as e:
        print(f"[{name}] FAILED: {type(e).__name__}: {e}")


async def main():
    await scenario("inmem-nostream", serialize=False)                 # OK
    await scenario("serialized-nostream")                             # UserError
    await scenario("serialized-streamed", streamed=True)              # UserError
    await scenario("serialized-session", session=True)                # UserError
    await scenario("serialized-session-streamed", streamed=True, session=True)  # UserError


asyncio.run(main())
```

Observed output:

```
[inmem-nostream] OK final_output='done'
[serialized-nostream] FAILED: UserError: Cannot resume a serialized approval checkpoint ...
[serialized-streamed] FAILED: UserError: Cannot resume a serialized approval checkpoint ...
[serialized-session] FAILED: UserError: Cannot resume a serialized approval checkpoint ...
[serialized-session-streamed] FAILED: UserError: Cannot resume a serialized approval checkpoint ...
```

### Root cause

`_current_response_boundary()` in `src/agents/run_internal/blocked_output.py` proves which generated items belong to the current response using exact object identity (`_identity_sequence_start`). Serialization rebuilds every item, destroying that identity, so both the generated-items and session-items identity lookups fail.

The code already acknowledges this problem - there is a structural fallback for deserialized items - but it is gated on `_current_turn == 1`:

> `# Serialization rebuilds item identities, but turn one has no accepted prefix.`

A natural multi-turn run (approval requested on turn >= 2 after earlier tool calls) falls through to `proven=False`, so `_validate_resumed_session_output_guardrail_safety()` raises on every resume. The existing test coverage only asserts this error for a hand-mutated ambiguous state, not for a naturally serialized multi-turn checkpoint.

### Expected behavior

A serialized approval checkpoint taken mid-run should be resumable when the response boundary is reconstructible from the snapshot itself, regardless of turn number - matching the in-memory path.

### Suggested direction

Persist/restore enough information in the snapshot to prove the boundary positionally (for example, the index offset of `processed_response.new_items` within `_generated_items`) instead of relying on object identity alone.

szupzj18:
### Summary

When an approval checkpoint with output guardrails is serialized and deserialized, all `RunItem` objects are rebuilt, destroying object identity. `_current_response_boundary` relied on object identity (`_identity_sequence_start`) to locate the processed response's items within `_generated_items` and `_session_items`, so the boundary could not be proven for any turn after the first — raising `UserError: Cannot resume a serialized approval checkpoint with output guardrails`.

The existing turn-1 type-based prefix fallback did not help because it was gated on `_current_turn == 1`, and the processed items are not necessarily a prefix of `_generated_items` on later turns (interruption items like `ToolApprovalItem` are appended after the processed response's `new_items`).

### Change

Add `_structural_item_key` and `_structural_sequence_start` helpers that match items by `(item_id, item_type, call_id)` — the same identifiers used by `_merge_generated_items_with_processed` for deduplication. When identity matching fails, fall back to structural matching to locate the processed response's items within the generated and session item lists. The existing turn-1 type-based prefix fallback is retained as a last resort for items without identifiers.

### Test plan

- New test `test_serialized_approval_checkpoint_with_output_guardrail_resumes_after_earlier_turn` (parametrized non-streamed/streamed): runs a 3-turn ScriptedModel scenario (normal tool → approval tool interruption → final message), serializes and deserializes the `RunState`, approves the interruption, and verifies the run completes with `final_output == "done"`.
- Updated `test_ambiguous_serialized_approval_state_fails_before_tool_execution`: corrupts the processed response's item identifiers after deserialization so structural matching cannot locate them, verifying the safety check still raises `UserError` when the boundary genuinely cannot be proven.
- `make format`, `make lint`, `make typecheck` — all clean.
- `make tests` — parallel suite fully passes; serial suite: 77 passed, 4 skipped, 7 errors (all pre-existing Dapr Redis integration tests requiring Docker/testcontainers, unrelated to this change).

### Issue number

Fixes #4611

### Checks

### DIFF
--- a/src/agents/run_internal/blocked_output.py
+++ b/src/agents/run_internal/blocked_output.py
@@ -366,25 +366,39 @@ def _identity_sequence_start(
     return None
 
 
+def _structural_field(raw_item: Any, field: str) -> str | None:
+    """Read an exact string field without invoking payload equality or hashing hooks."""
+    if type(raw_item) is dict:
+        values = raw_item
+    else:
+        try:
+            values = object.__getattribute__(raw_item, "__dict__")
+        except AttributeError:
+            return None
+    if type(values) is not dict:
+        return None
+    value = _exact_dict_field(values, field)
+    if type(value) is not str:
+        return None
+    return value
+
+
 def _structural_item_key(item: RunItem) -> tuple[str | None, str | None, str | None]:
     """Return (item_id, item_type, call_id) for structural matching after deserialization."""
-    item_id: str | None = None
-    item_type: str | None = None
-    call_id: str | None = None
     raw = getattr(item, "raw_item", None)
-    if raw is not None:
-        if isinstance(raw, dict):
-            item_id = raw.get("id")
-            item_type = raw.get("type")
-            call_id = raw.get("call_id")
-        else:
-            item_id = getattr(raw, "id", None)
-            item_type = getattr(raw, "type", None)
-            call_id = getattr(raw, "call_id", None)
+    if raw is None:
+        return None, None, None
+    item_id = _structural_field(raw, "id")
+    item_type = _structural_field(raw, "type")
+    call_id = _structural_field(raw, "call_id")
     if item_id is None:
-        item_id = getattr(item, "id", None)
+        fallback_id = getattr(item, "id", None)
+        if type(fallback_id) is str:
+            item_id = fallback_id
     if item_type is None:
-        item_type = getattr(item, "type", None)
+        fallback_type = getattr(item, "type", None)
+        if type(fallback_type) is str:
+            item_type = fallback_type
     return item_id, item_type, call_id
 
 
@@ -454,8 +468,13 @@ def _current_response_boundary(
                                 sess_structural : sess_structural + len(processed_items)
                             ]
                         )
+                        suffixes.extend(run_state._session_items[sess_structural:])
+                    # When both owners match, they hold copies of the same
+                    # current response; the generated suffix extended above
+                    # is canonical so independently deserialized copies
+                    # cannot survive deduplication. The session start index
+                    # is still retained for owner cleanup.
                     session_start = sess_structural
-                    suffixes.extend(run_state._session_items[session_start:])
                 proven = True
             elif run_state._current_turn == 1:
                 # Last resort: turn one has no accepted prefix, so a type-based
--- a/src/agents/run_internal/blocked_output.py
+++ b/src/agents/run_internal/blocked_output.py
@@ -475,11 +475,15 @@ def _current_response_boundary(
                     proven = True
 
     current_items: list[RunItem] = []
-    seen: set[int] = set()
+    seen_ids: set[int] = set()
+    seen_keys: set[tuple[str | None, str | None, str | None]] = set()
     for item in (*processed_items, *suffixes, *response_items):
-        if id(item) in seen:
+        key = _structural_item_key(item)
+        if id(item) in seen_ids or (key != (None, None, None) and key in seen_keys):
             continue
-        seen.add(id(item))
+        seen_ids.add(id(item))
+        if key != (None, None, None):
+            seen_keys.add(key)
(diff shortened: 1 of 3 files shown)


szupzj18:
Good catch. Fixed in 9d1c551c: the deduplication loop now tracks structural keys (`item_id`, `item_type`, `call_id`) alongside `id(item)`. When `_generated_items` and `_session_items` contain independently deserialized copies of the same items (the post-deserialization case), the structural key recognizes them as equivalent and skips the duplicate. Items without identifiers (all-None key) fall back to identity-only deduplication, preserving the existing behavior for in-memory runs.

All 166 streamed + 464 run_state + 16 guardrail tests pass; lint and typecheck clean.

seratch:
Thanks for sharing this patch. The issue was resolved by https://github.com/openai/openai-agents-python/pull/4613

PR #4624, opened 2026-08-24
fix(extensions): bound trimmed string outputs to max_output_chars

### LINKED ISSUE #4622: ToolOutputTrimmer summaries can exceed max_output_chars for string outputs
### Describe the bug

ToolOutputTrimmer can emit a string summary longer than the configured max_output_chars budget. This happens when the existing rich summary is shorter than the original tool output but larger than the configured budget.

### Repro steps

With openai-agents 0.22.0:

```python
from agents.extensions import ToolOutputTrimmer
from agents.run_config import ModelInputData

items = [
    {"type": "function_call", "name": "search", "call_id": "call_1", "arguments": "{}"},
    {"type": "function_call_output", "call_id": "call_1", "output": "x" * 1000},
]
data = SimpleNamespace(
    model_data=ModelInputData(input=items, instructions=None),
)
result = ToolOutputTrimmer(max_output_chars=40)(data)
print(len(result.model_data.input[1]["output"]))
```

The emitted summary is longer than 40 characters.

### Expected behavior

Every supported string function-call output replacement should respect max_output_chars. Existing summaries should remain unchanged when they already fit, and outputs should remain untrimmed when the rich summary would not be shorter than the original.

### Debug information

- Agents SDK version: 0.22.0
- Python: 3.12.14
- OS: Ubuntu 22.04
- Reproducible: yes

abhay-codes07:
### Summary

`ToolOutputTrimmer` could emit a string summary **longer** than the configured `max_output_chars`. When trimming a string `function_call_output`, the summary was accepted as long as it was merely shorter than the *original* output — so a rich `[Trimmed: … ]\n<preview>` summary whose length fell between `max_output_chars` and the original size passed through unbounded. This defeats the purpose of the budget (bounding replayed context) and is inconsistent with the structured-output path, which already fits every replacement within `max_output_chars`.

Fixes #4622.

**Change:** route string `function_call_output` summaries and legacy free-text `tool_search_output` results through a shared `_string_output_summary` helper that:

1. keeps the full-fidelity preview summary verbatim when it already fits the budget (unchanged behavior for the common case);
2. otherwise shrinks the preview and falls back through progressively shorter headers (`[Trimmed: <tool> output — N chars → M char preview]` → `[Trimmed: <tool> output — N chars]` → `[Trimmed: <tool>]` → `[Trimmed]`), so the replacement never exceeds `max_output_chars`;
3. still leaves the output **untrimmed** when even the rich summary would not be shorter than the original — the existing documented escape hatch (`test_skips_trim_when_summary_would_exceed_original`), matching the structured path.

The header/preview cascade mirrors the one already used for structured outputs, so the two paths now share one budget contract. Docstrings for `max_output_chars`/`preview_chars` are updated to state that every replacement fits the budget and previews may shrink to fit.

### Test plan

- New tests in `tests/extensions/test_tool_output_trimmer.py`:
  - `test_string_output_respects_budget` — parametrized over `max_output_chars ∈ {1, 9, 15, 40, 100}` × `preview_chars ∈ {0, 20, 200, 990}` × `output_len ∈ {200, 501, 1000, 5000}`: a trimmed summary is always `≤ max_output_chars`, untrimmed outputs are left exactly as-is, and trimming is idempotent.
  - `test_string_output_matches_issue_repro` — the reported 1000-char output at budget 40 now yields a ≤40-char summary.
  - `test_legacy_tool_search_output_respects_budget` — free-text `tool_search_output` results are bounded too.
- 54 of the new parametrized cases (including the issue repro and the legacy path) fail on `main` without the source change.
- `uv run pytest tests/extensions tests/test_released_api_contract.py`: 1687 passed, 7 skipped.
- Full suite, `uv run ruff format --check`, `uv run ruff check` (on changed files), and `uv run mypy src/agents/extensions/tool_output_trimmer.py` pass locally.

### Issue number

Fixes #4622

### Checks

### DIFF
--- a/src/agents/extensions/tool_output_trimmer.py
+++ b/src/agents/extensions/tool_output_trimmer.py
@@ -97,12 +97,12 @@ class ToolOutputTrimmer:
         recent_turns: Number of recent user messages whose surrounding items are never
             trimmed. Defaults to 2.
         max_output_chars: Tool outputs above this character count are candidates for
-            trimming. Structured outputs count their model-facing string payloads without
-            Python or JSON representation overhead, and their replacements fit within this
-            budget. Defaults to 500.
+            trimming, and every replacement summary fits within this budget. Structured
+            outputs count their model-facing string payloads without Python or JSON
+            representation overhead. Defaults to 500.
         preview_chars: Maximum number of characters of a string output, or the text parts of
-            a structured output, to preserve as a preview when trimming. Structured previews
-            may be shorter when needed to fit ``max_output_chars``. Defaults to 200.
+            a structured output, to preserve as a preview when trimming. Previews may be
+            shorter when needed to fit ``max_output_chars``. Defaults to 200.
         trimmable_tools: Optional tool name or set of tool names whose outputs can be trimmed.
             For namespaced tools, both bare names and qualified ``namespace.name`` entries are
             supported. If ``None``, all tool outputs are eligible for trimming. Defaults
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for your interest here. Closing this PR for the reason I mentioned at https://github.com/openai/openai-agents-python/issues/4622#issuecomment-5401908081

PR #4628, opened 2026-08-24
fix(memory): compact the full stored history of a limited session

hsusul:
### Summary

`OpenAIResponsesCompactionSession` loads compaction input with `underlying_session.get_items()`, which applies `SessionSettings.limit`. After `responses.compact`, it clears the **entire** underlying store and writes the compact result.

For a SQLite (or similar) session that stores full history but retrieves only the last N items, successful compaction therefore deletes everything older than N. The restore path already uses `_get_all_underlying_session_items()`; candidate loading did not.

### Reproduction

```python
from agents.memory import SQLiteSession, SessionSettings, OpenAIResponsesCompactionSession

underlying = SQLiteSession("s", session_settings=SessionSettings(limit=2))
# add 3+ items, then force input-mode compaction
# responses.compact receives only the last 2 items
# the session is then replaced with that compact output
```

### Solution

Load candidates via the existing `_get_all_underlying_session_items()` helper (`limit=2_147_483_647`), matching the replacement/restore path.

This is a smaller version of the candidate-loading fix from closed stale #3827, without that PR's broader `previous_response_id` rewrite.

### Test plan

- [x] Added a SQLiteSession regression that sets `SessionSettings(limit=2)`, stores 3 items, and asserts `responses.compact` receives all 3
- [x] `uv run pytest tests/memory/test_openai_responses_compaction_session.py` (56 passed)
- [x] `uv run ruff format` / `ruff check` on the changed files
- [x] `uv run pyright` on the changed files

### Issue number

N/A. Related closed stale PR: #3827.

### Checks

- [x] I've run targeted format, lint, typecheck, and tests on the changed files
- [ ] I've run `.agents/skills/code-change-verification/scripts/run.sh` (full `make format` currently fails on an unrelated pre-existing E501 in `.agents/skills/implementation-final-review/scripts/test_skill_contract.py` on `main`)

### DIFF
--- a/src/agents/memory/openai_responses_compaction_session.py
+++ b/src/agents/memory/openai_responses_compaction_session.py
@@ -442,7 +442,11 @@ class OpenAIResponsesCompactionSession(SessionABC, OpenAIResponsesCompactionAwar
         if self._compaction_candidate_items is not None and self._session_items is not None:
             return (self._compaction_candidate_items[:], self._session_items[:])
 
-        history = _normalize_compaction_session_items(await self.underlying_session.get_items())
+        # Bypass SessionSettings.limit so compaction sees stored history, not just the
+        # retrieval window. Replacement still writes over the full store.
+        history = _normalize_compaction_session_items(
+            await self._get_all_underlying_session_items()
+        )
         candidates = select_compaction_candidate_items(history)
         self._compaction_candidate_items = candidates
         self._session_items = history
--- a/src/agents/memory/openai_responses_compaction_session.py
+++ b/src/agents/memory/openai_responses_compaction_session.py
@@ -156,8 +156,20 @@ class OpenAIResponsesCompactionSession(SessionABC, OpenAIResponsesCompactionAwar
         response_id: str | None,
         store: bool | None,
         requested_mode: OpenAIResponsesCompactionMode | None,
+        session_items: list[TResponseInputItem],
     ) -> _ResolvedCompactionMode:
         mode = requested_mode or self.compaction_mode
+        if mode != "input":
+            settings = self.underlying_session.session_settings
+            limit = settings.limit if settings is not None else None
+            if limit is not None and len(session_items) > max(limit, 0):
+                if mode == "previous_response_id":
+                    raise ValueError(
+                        "OpenAIResponsesCompactionSession cannot use previous_response_id "
+                        "compaction when the underlying session retrieval limit hides local "
+                        "history; use compaction_mode='input' instead."
+                    )
+                return "input"
         if (
             mode == "auto"
             and store is None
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The current full-history read makes the default auto mode unsafe. Hidden rows can now trigger compaction while auto selects previous_response_id; that response represents only the limited session window, but replacement clears the full local store.

Please make history selection mode-aware. Input compaction must use the full store, and auto must switch to input whenever the full store differs from the ordinary session view. An explicit previous_response_id request must fail before provider or storage side effects when hidden local rows make that response chain non-authoritative, with input mode identified as the supported alternative. Please add regressions for a visible count below the threshold with a full count above it, and for the explicit previous_response_id case, asserting both request shape and preserved history.

FU-max-boop:
Thanks for the mode-aware update at exact head `4142c699`. I independently reran the four event-controlled ownership probes against this head with a real SQLite session. The limited/full-history auto-authority case and the provider/storage-side explicit-previous fail-fast case no longer reproduce; I am keeping the already-reviewed hook-order issue separate.

Two deterministic concurrency windows remain (`2 passed, 2 failed`):

1. A wrapper `add_items(late-write)` that completes while the compaction provider is suspended is erased when the older provider result later performs unconditional replacement. The final store contains only the compacted snapshot.
2. Two gated `run_compaction` attempts called with `resp-first` and `resp-second` both send `resp-second`. Each attempt writes the shared `_response_id` before an await, then later rereads that shared field while constructing its request.

These appear to be one attempt-ownership boundary: response ID, resolved mode, full-history snapshot, provider request, and replacement authority need to belong to the same transaction. The narrow serialization option is to hold the existing mutation lock from snapshot through replacement. If the provider await must stay outside the lock, the equivalent safe shape needs a generation/history check after reacquiring it and must discard the provider output without mutating storage when the snapshot changed; taking a newer pre-clear snapshot does not make output computed from the older snapshot safe.

Current-head focused tests remain green (`60 passed`) and Ruff passes. I have compact cleanup-safe tests for the two remaining windows and have not opened a competing change.

rioyu123:
I checked this on current main (`fbd2dbc`) using the full SDK and a real SQLite session. With six stored items and `SessionSettings(limit=2)`, input-mode compaction receives only two items and then replaces all six. An offline compaction client that simply returns its input still leaves only two items; the unlimited-read control preserves all six.

I saw the request to pause #4293 while maintainers decide the approach, and the mode-selection feedback here. Is a follow-up focused on limited-read history loss welcome now, or should this remain on hold? I would keep the existing concurrency fixes separate and agree on the filter and response-ID behavior before implementing anything.

@hsusul, are you planning to resume this work? Happy to coordinate rather than duplicate it.

hsusul:
this specific pr is pretty out of date and from the looks of it osmeone made an adjacent fix although i havent dug too deep.

> I checked this on current main (`fbd2dbc`) using the full SDK and a real SQLite session. With six stored items and `SessionSettings(limit=2)`, input-mode compaction receives only two items and then replaces all six. An offline compaction client that simply returns its input still leaves only two items; the unlimited-read control preserves all six.
> 
> I saw the request to pause #4293 while maintainers decide the approach, and the mode-selection feedback here. Is a follow-up focused on limited-read history loss welcome now, or should this remain on hold? I would keep the existing concurrency fixes separate and agree on the filter and response-ID behavior before implementing anything.
> 
> @hsusul, are you planning to resume this work? Happy to coordinate rather than duplicate it.

rioyu123:
Thanks for the update. #4736 added protection for concurrent writes through the same compaction wrapper. The limited-read case I described was reproduced on `fbd2dbc`, which already includes that fix, so it looks like a separate issue. I'll wait for maintainer guidance before opening a follow-up.

PR #4629, opened 2026-08-24
feat(tracing): snapshot effective tools and handoffs per turn

### LINKED ISSUE #4626: Trace the effective tools and handoffs for each model turn
### Please read this first

- **Have you searched for related issues and PRs?** Yes.

Related but distinct:
- #4618 is a docs/enhancement request about deriving function-tool, MCP-tool, and handoff visibility from one runtime policy.
- #4620 is about input/output guardrail parity for locally executed MCP tools.

This issue is narrower: preserve the **effective model-visible capability set for each individual model turn** in tracing.

### Describe the feature

The SDK already resolves the effective tools and handoffs immediately before a model call and writes them to the current `AgentSpanData`:

https://github.com/openai/openai-agents-python/blob/main/src/agents/run_internal/run_loop.py#L2098-L2114

That is useful, but one agent span can cover multiple model turns. `AgentSpanData.tools` and `.handoffs` are mutable fields, so if runtime policy changes the effective capability set between turns, each update replaces the previous values.

As a result, after the run it may be impossible to answer a production-debugging question precisely:

> What tools and handoffs did this model invocation actually see?

This matters when `FunctionTool.is_enabled`, `Handoff.is_enabled`, or MCP `tool_filter` depend on request/run context such as tenant, workflow state, feature flags, experiment allocation, or state changed by an earlier tool call.

### Why `TurnSpanData` looks like a natural fit

Current `main` already has optional task/turn tracing. `TurnSpanData` represents one agent loop turn, but currently carries only the turn number, agent name, usage, and metadata:

https://github.com/openai/openai-agents-python/blob/main/src/agents/tracing/span_data.py

Conceptually, the trace could preserve:

```text
agent span: Agent A
  turn 1: tools=[search, lookup] handoffs=[billing]
  turn 2: tools=[lookup, submit] handoffs=[]
```

The agent span can keep its current behavior for compatibility; when turn spans are enabled, each turn span could additionally snapshot the already-resolved tool and handoff names for that model invocation.

I am not attached to `TurnSpanData` specifically if maintainers prefer the generation/response span or another existing model-invocation boundary. The important property is that the snapshot is associated with the exact model invocation rather than being overwritten later on a longer-lived agent span.

### Suggested semantics

1. Record the capability set **after** dynamic enablement/filtering and tool-name collision resolution, i.e. the same effective names that are about to be sent to the model.
2. Preserve one immutable snapshot per model turn/invocation.
3. Record names/identifiers only, matching the information already emitted on `AgentSpanData`; do not add tool arguments, schemas, credentials, or authorization data.
4. Keep this explicitly observational: a visible tool or handoff does **not** imply that a particular invocation is authorized.
5. Preserve existing tracing behavior when task/turn spans are disabled.

### Reproduction-shaped case

A single agent runs for two turns. A dynamic `is_enabled` policy exposes `draft_invoice` initially. After the first tool call changes workflow state, the next model turn exposes `send_invoice` instead.

Expected trace evidence:

```text
turn 1 -> [draft_invoice]
turn 2 -> [send_invoice]
```

The final `AgentSpanData.tools` may still contain the latest set, but the trace retains both historical snapshots.

A regression test could use one agent with a context-dependent tool/handoff predicate, force two model turns, and assert that the exported turn spans contain two distinct effective capability sets while the existing agent-span behavior remains unchanged.

### Non-goals

- no new capability-policy abstraction;
- no authorization framework;
- no tool registry;
- no raw tool schemas/arguments in tracing;
- no change to how capabilities are selected.

This is only about preserving the capability surface that the SDK has already resolved for each model invocation.

If this fits the tracing model, I would be happy to help with a focused implementation/tests around the preferred span.

sylvesterkaczmarek:
### Summary

This pull request adds per-turn tracing of the effective tools and handoffs visible to each model invocation.

Turn spans snapshot capability names after dynamic enablement, filtering, and tool-name collision resolution, while agent spans retain their existing latest-view and identity behaviour. The snapshots contain names only and work for both streaming and non-streaming runs. Runs with task/turn spans disabled retain the existing compact trace hierarchy.

The review follow-up also ensures that:
- turn spans record `Handoff.tool_name`, including `tool_name_override`
- namespaced `FunctionTool` entries use their canonical qualified trace identity
- existing agent-span handoff and tool identities remain backwards compatible
- regression coverage exercises both a handoff override and a namespaced function tool, including the exported turn-span payload

### Test plan

- Same-fork staging CI on the final review-fix head passed Ruff formatting/lint, typecheck, Windows mypy, native macOS sandbox, MCP v1 compatibility, prospective release contract, packaged Windows contract, and Python 3.10-3.13 test jobs.
- The temporary staging PR was closed after the relevant validation above; the remaining Python 3.14 and Windows/packaged matrix legs were still running with no failures reported at that point.
- Upstream OpenAI Actions is still `action_required` pending external-contributor approval.

### Issue number

Closes #4626

### Checks

### DIFF
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -337,17 +337,12 @@ def _record_effective_model_capabilities(
     agent_handoff_names = [handoff.agent_name for handoff in handoffs]
     turn_handoff_names = [handoff.tool_name for handoff in handoffs]
     agent_tool_names = [
-        tool_name
-        for tool in tools
-        if (tool_name := get_tool_trace_name_for_tool(tool)) is not None
+        tool_name for tool in tools if (tool_name := get_tool_trace_name_for_tool(tool)) is not None
     ]
     turn_tool_names = [
         trace_name
         for tool in tools
-        if (
-            trace_name := get_function_tool_trace_name(tool)
-            or get_tool_trace_name_for_tool(tool)
-        )
+        if (trace_name := get_function_tool_trace_name(tool) or get_tool_trace_name_for_tool(tool))
         is not None
     ]
 
@@ -1205,7 +1200,7 @@ async def start_streaming(
                 response_id=response_id,
                 reasoning_item_id_policy=streamed_result._reasoning_item_id_policy,
                 store=store_setting,
-                wrapper=context_wrapper,
+                wrapper=streamed_result.context_wrapper,
             )
             streamed_result._current_turn_persisted_item_count += saved_count
     except BaseException:
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -85,10 +85,18 @@ from ..tool import (
     dispose_resolved_computers,
 )
 from ..tool_guardrails import ToolInputGuardrailResult, ToolOutputGuardrailResult
-from ..tracing import Span, SpanError, agent_span, get_current_trace, task_span, turn_span
+from ..tracing import (
+    Span,
+    SpanError,
+    agent_span,
+    get_current_span,
+    get_current_trace,
+    task_span,
+    turn_span,
+)
 from ..tracing.config import include_task_and_turn_spans
 from ..tracing.model_tracing import get_model_tracing_impl
-from ..tracing.span_data import AgentSpanData, TaskSpanData
+from ..tracing.span_data import AgentSpanData, TaskSpanData, TurnSpanData
 from ..usage import (
     Usage,
     _extract_raw_usage_snapshot,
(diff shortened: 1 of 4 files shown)


seratch (changes requested):
Thanks for the focused contribution. The overall approach in #4629 is the implementation path we want to pursue: capture each turn's resolved capability set on `TurnSpanData` after dynamic enablement, filtering, and collision resolution.

I also verified the proposed payload with a live Traces ingest probe. The exact custom-span structure, including `list[str]` values for `tools` and `handoffs`, returned HTTP 204 and was retained and rendered correctly in the Traces UI. No tracing-schema redesign is needed.

Before merging, please make these focused identity corrections:

- Record `Handoff.tool_name` in `TurnSpanData` rather than `Handoff.agent_name`. With the supported `tool_name_override`, the current value can differ from the identifier offered to the model, and distinct handoff tools targeting the same agent become indistinguishable.
- Use the existing namespace-aware canonical trace identity for `FunctionTool` definitions. The current generic helper can reduce namespaced tools to the unqualified `tool.name`.
- Split the handoff projections in the helper: continue using `agent_name` for the released `AgentSpanData.handoffs` behavior, while using `tool_name` for the new per-turn capability snapshot.
- Add focused regression coverage for one handoff with `tool_name_override` and one namespaced `FunctionTool`, asserting that the exported turn-span identifiers match the resolved capability identifiers.

The existing streamed/non-streamed dynamic-enablement test already covers the lifecycle placement and does not need broader permutation coverage.

sylvesterkaczmarek:
Addressed the requested identity corrections: turn spans now use Handoff.tool_name and namespace-aware FunctionTool identities, while legacy agent-span identities remain unchanged. Added regression coverage for tool_name_override and a namespaced FunctionTool. The review threads are resolved; the latest Actions run is waiting for maintainer approval to start.

sylvesterkaczmarek:
@seratch  thanks again, addressed the review feedback! Turn spans now use the model-visible handoff tool name and namespace-aware FunctionTool identity, while agent spans preserve their existing identities.

I also added focused coverage for a tool_name_override handoff and a namespaced FunctionTool, including assertions against the exported turn-span payload. 

Same-fork validation is green across the relevant lint, typecheck, platform, and Python 3.10-3.13 test jobs; upstream Actions are still awaiting external-contributor approval.

seratch:
Thanks for the thoughtful implementation and for working through the review feedback. After validating the behavior with both scripted and live-model traces, I do not think we should add this duplication to the tracing schema.

For each model turn, the existing model-call span already records the exact functions sent to the model, including dynamically enabled tools and handoffs represented as function tools. Adding the same capability list to the turn span would create another source of truth and ongoing identity and maintenance coupling without a demonstrated supported scenario that the model-call trace cannot handle.

I am going to close this PR and #4626. If a concrete tracing workflow shows that the model-call span is insufficient, we can revisit that narrower gap. Thanks again for the careful work here.

PR #4631, opened 2026-08-24
feat(mcp): support tool guardrails on local servers

### LINKED ISSUE #4620: Support tool input/output guardrails for local MCP server tools
### Please read this first

Related but distinct:
- #4618 is about composing model-visible capability filtering across function tools, MCP tools, and handoffs.
- #2868 discusses generic per-tool authorization.
- #2970 discusses generic pre-execution validation.

This request is narrower: allow **locally executed MCP tools to opt into the existing tool input/output guardrail abstraction** that is already available for function tools.

### Describe the feature

The SDK has a useful `tool_input_guardrails` / `tool_output_guardrails` abstraction for function tools.

Today, however, the guardrails documentation explicitly scopes tool guardrails to function tools, while local MCP servers have a different set of controls:

- `tool_filter` for model-visible tool availability;
- `require_approval` for HITL approval;
- `tool_meta_resolver` for per-call MCP `_meta`;
- `failure_error_function` for MCP failure handling.

All of those are useful, but they solve different problems.

There does not appear to be a public configuration point for saying:

> For every invocation of this locally executed MCP tool, run these input guardrails immediately before execution and these output guardrails immediately after execution.

This is slightly surprising because local MCP tools are already converted internally into SDK `FunctionTool` objects by `MCPUtil`.

#### Production-shaped example

Consider an agent connected to an application-managed MCP server.

Before an MCP call executes, the application may need to deterministically check things such as:

- whether the model-generated arguments contain data that must not be sent to that tool;
- whether an argument violates an invariant derived from trusted run context;
- whether that particular invocation should be rejected even though the tool itself is visible and available.

After the MCP call returns, the application may need to check the result before it becomes model-visible, for example:

- reject a response containing sensitive/internal data;
- prevent an unexpected or unsafe result from entering the agent context;
- apply the same output-validation policy already used for local function tools.

Conceptually:

```text
model chooses MCP tool
        ↓
tool input guardrails
        ↓
local MCP server.call_tool(...)
        ↓
tool output guardrails
        ↓
result becomes model-visible
```

We ended up implementing this kind of wrapper downstream for MCP-backed agents. What started as a sample-specific concern became shared infrastructure because the same need recurred across tools and agents.

### Why the existing MCP controls are not equivalent

`tool_filter` is primarily about **visibility / eligibility**: should the model see this tool?

`require_approval` answers whether execution should pause for approval.

`tool_meta_resolver` is useful for trusted per-call metadata such as tenant or trace information.

Tool guardrails answer a separate question:

> Given the exact invocation/output at this execution boundary, should it be accepted?

These mechanisms are complementary rather than alternatives.

This also means I am **not** proposing that client-side guardrails replace authorization in the MCP server. Resource/action authorization should still be enforced at the protected operation.

### Possible API shape

The exact API is less important than parity with the existing tool-guardrail model.

One possible server-level shape would be:

```python
server = MCPServerStreamableHttp(
    name="Billing",
    params={"url": "..."},
    tool_input_guardrails=[validate_mcp_input],
    tool_output_guardrails=[validate_mcp_output],
)
```

A resolver/per-tool configuration could also make sense if maintainers prefer not to apply every guardrail to every tool.

Since local MCP tools are already converted to `FunctionTool`s internally, another possibility may simply be attaching the configured guardrails during MCP → FunctionTool conversion rather than introducing a separate execution system.

### Desired semantics

Ideally:

1. local MCP tools can opt into the existing input-tool-guardrail semantics before `call_tool()`;
2. local MCP tools can opt into the existing output-tool-guardrail semantics before the result is accepted/model-visible;
3. configuration can distinguish tools when necessary, rather than requiring one policy for every tool on a server;
4. guardrails receive normal run context plus enough MCP origin information to distinguish server/tool identity;
5. behavior is unchanged when no MCP guardrails are configured;
6. interaction with `require_approval` is documented and, where practical, follows the same ordering semantics as function-tool approvals + guardrails;
7. guardrail results remain observable/traced in the same way as existing tool-guardrail results.

### Non-goals

This issue is intentionally **not** asking for:

- a new authorization framework;
- replacing server-side MCP authorization;
- a generic middleware architecture for every SDK tool type;
- arbitrary input/output transformation semantics beyond what existing tool guardrails support;
- `HostedMCPTool` parity in the same change.

Hosted MCP has a materially different execution boundary because the Python process does not own the MCP round trip, so I think it should be considered separately.

### Why this seems like a useful SDK-level primitive

MCP tools are often precisely where an agent crosses from model reasoning into external systems.

Having the same lightweight pre/post execution checks for:

```text
FunctionTool
```

and:

```text
local MCP-backed FunctionTool
```

would make it easier to apply one production policy consistently without wrapping every MCP tool manually or reimplementing the MCP integration layer.

If maintainers agree this belongs in the SDK, I'd be happy to help with a focused implementation or docs/tests around the preferred API shape.

sylvesterkaczmarek:
### Summary

This pull request adds existing tool input/output guardrail support to locally executed MCP tools.

`MCPServer`, `MCPServerStdio`, `MCPServerSse`, and `MCPServerStreamableHttp` now accept optional `tool_input_guardrails` and `tool_output_guardrails`. A list applies to every tool on the server, while a mapping keyed by the original MCP tool name allows per-tool configuration.

`MCPUtil` attaches the selected guardrails to the converted `FunctionTool`, so local MCP tools use the existing tool-guardrail execution, approval, result-reporting, and tracing pipeline rather than introducing a separate MCP execution path. Existing `ToolExecutionConfig` pre-approval semantics are inherited unchanged. Hosted MCP behavior is unchanged, and local MCP behavior is unchanged when guardrails are omitted.

### Test plan

- `uv run pytest -q tests/mcp/test_mcp_tool_guardrails.py tests/mcp/test_mcp_util.py tests/mcp/test_runner_calls_mcp.py tests/mcp/test_mcp_approval.py` — 159 passed
- Changed-file Ruff formatting/lint — passed
- `uv run mypy src/agents/mcp/server.py src/agents/mcp/util.py` — passed
- Fork CI: lint, typecheck, MCP v1 compatibility, native macOS sandbox, Windows mypy, prospective release contract, and Windows packaged-contract checks passed on the clean rebased commit

### Issue number

Closes #4620

### Checks

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -44,6 +44,7 @@ from ..logger import (
 )
 from ..run_context import RunContextWrapper
 from ..tool import ToolErrorFunction
+from ..tool_guardrails import ToolInputGuardrail, ToolOutputGuardrail
 from ..util._types import MaybeAwaitable
 from . import _compat as mcp_compat
 from ._compat import (
@@ -549,6 +550,12 @@ class MCPServer(abc.ABC):
         failure_error_function: ToolErrorFunction | None | _UnsetType = _UNSET,
         tool_meta_resolver: MCPToolMetaResolver | None = None,
         custom_data_extractor: MCPToolCustomDataExtractor | None = None,
+        tool_input_guardrails: (
+            list[ToolInputGuardrail[Any]] | dict[str, list[ToolInputGuardrail[Any]]] | None
+        ) = None,
+        tool_output_guardrails: (
+            list[ToolOutputGuardrail[Any]] | dict[str, list[ToolOutputGuardrail[Any]]] | None
+        ) = None,
     ):
         """
         Args:
@@ -570,6 +577,10 @@ class MCPServer(abc.ABC):
                 tool calls. It is invoked by the Agents SDK before calling `call_tool`.
             custom_data_extractor: Optional callable that produces SDK-only custom data for
                 emitted MCP tool output items.
+            tool_input_guardrails: Input guardrails for converted local MCP tools. A list
+                applies to every tool; a dict applies guardrails by original MCP tool name.
+            tool_output_guardrails: Output guardrails for converted local MCP tools. A list
+                applies to every tool; a dict applies guardrails by original MCP tool name.
         """
         self.use_structured_content = use_structured_content
         self._needs_approval_policy = self._normalize_needs_approval(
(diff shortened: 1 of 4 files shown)


seratch:
Thanks for sending this. The issue was resolved by https://github.com/openai/openai-agents-python/pull/4632

PR #4633, opened 2026-08-24
docs: unify capability visibility guidance

### LINKED ISSUE #4618: docs: unified capability-visibility pattern across function tools, MCP tools, and handoffs
The Agents SDK now has several useful mechanisms for controlling what an agent can see at runtime:

- "FunctionTool.is_enabled";
- "Handoff.is_enabled";
- MCP "tool_filter";
- tool guardrails / approvals for execution-time checks;
- handoff input filtering.

Individually these APIs make sense.

I think there is an opportunity to document a single production pattern for using them together when capability availability comes from the same runtime policy.

Production-shaped use case

Consider an agent application where the capabilities valid for a request depend on runtime context such as:

- authenticated actor/client;
- channel or source;
- application variant;
- tenant;
- workflow state;
- feature flags.

The same application may expose capabilities through several SDK mechanisms:

- local function tools;
- MCP tools;
- specialist-agent handoffs.

Ideally the model-visible capability surface is derived consistently from one application policy rather than implementing slightly different conditions independently for each mechanism.

There is also an important distinction between visibility and authorization.

For example:

- a tool may be hidden entirely because it is irrelevant or unavailable for this client;
- a tool may be visible, but execution of a particular call may still require resource-level authorization based on its arguments;
- a handoff may be valid only during one workflow phase;
- after a handoff, the new agent should resolve its own effective capability set from the same run context.

Suggested documentation pattern

It would be useful to have one example showing something roughly like:

RunContext
    |
    +-- capability policy
           |
           +-- FunctionTool.is_enabled
           +-- Handoff.is_enabled
           +-- MCP tool_filter
           |
           +-- invocation-time authorization / guardrails

The example could explicitly establish that:

1. disabled capabilities are removed before the model sees them;
2. function tools, MCP tools, and handoffs can derive visibility from the same runtime context;
3. visibility filtering is not itself a security boundary;
4. argument/resource-level authorization belongs at invocation time;
5. capability policy is re-evaluated after agent transitions when appropriate.

Optional observability improvement

A related debugging capability would be useful: being able to inspect the effective model-visible capability set for a model invocation.

For production incidents, one of the first questions is often:

«What tools and handoffs did this agent actually see at this point in the run?»

If this information is already available through tracing/hooks, highlighting the recommended approach in the same example may be enough.

Scope

I'm not proposing another authorization framework or tool registry.

The existing primitives appear sufficient.

The main request is a canonical example showing how they compose into one coherent capability-visibility pattern.

I'd be happy to contribute that example/docs PR if maintainers think the pattern belongs in the SDK documentation.

sylvesterkaczmarek:
### Summary

This pull request documents a single request-scoped capability policy across function tools, local MCP tools, and handoffs.

The new section shows how one application policy can drive:

- `FunctionTool.is_enabled`
- MCP `tool_filter`
- `Handoff.is_enabled`

It also makes the security boundary explicit: model-visible capability filtering is not resource/action authorization. Function-tool application code and MCP servers remain responsible for authorizing protected operations.

The tracing note points to `AgentSpanData.tools` and `AgentSpanData.handoffs` for debugging while documenting the current agent-span limitation when one agent span covers multiple turns.

### Test plan

- Documented API-shape probe for `@tool(is_enabled=...)`, MCP `tool_filter`, and `handoff(..., is_enabled=...)` — passed
- Public `agents.mcp.ToolFilterContext` import — passed
- `git diff --check` — passed
- `make build-docs` — passed
- Focused Japanese/Korean/Chinese translation-safety review of the changed English prose — no concrete ambiguity findings

### Issue number

Closes #4618

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -838,6 +838,80 @@ Disabled tools are completely hidden from the LLM at runtime, making this useful
 -   A/B testing different tool configurations
 -   Dynamic tool filtering based on runtime state
 
+
+#### Use one capability policy across tools, MCP, and handoffs
+
+When function tools, local MCP tools, and handoffs depend on the same request state, keep the decision in one application policy and adapt that policy to each SDK hook. This avoids maintaining separate visibility rules that can drift apart.
+
+```python
+from dataclasses import dataclass
+
+from agents import Agent, RunContextWrapper, handoff
+from agents.decorators import tool
+from agents.mcp import MCPServerStreamableHttp, ToolFilterContext
+
+
+@dataclass
+class AppContext:
+    role: str
+    workflow_stage: str
+
+
+def capability_enabled(context: AppContext, capability: str) -> bool:
+    if capability == "refund_order":
+        return context.role == "billing" and context.workflow_stage == "review"
+    if capability == "lookup_invoice":
+        return context.role in {"support", "billing"}
+    if capability == "billing_handoff":
+        return context.role == "support"
+    return False
+
+
+def refund_enabled(ctx: RunContextWrapper[AppContext], _agent) -> bool:
+    return capability_enabled(ctx.context, "refund_order")
+
+
+@tool(is_enabled=refund_enabled)
+def refund_order(order_id: str) -> str:
+    """Request a refund for an order."""
+    return f"refund requested for {order_id}"
+
+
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for sharing the idea. We may update the docs for the purpose in some way, but these changes are not the ones we would like to pursue.

PR #4649, opened 2026-08-25
fix(sessions): chunk Conversations item creates at the API batch limit

hsusul:
### Summary

`OpenAIConversationsSession.add_items()` forwarded the entire list to `conversations.items.create`. The Conversations API allows **at most 20 items per create**, so a Runner turn with many parallel tool calls (11 tools → 11 calls + 11 outputs = 22 items) fails with HTTP 400. SQLite/Redis sessions have no such cap, so the same session workflow works there and breaks here.

### Reproduction

```python
session = OpenAIConversationsSession(conversation_id="conv", openai_client=client)
await session.add_items([{"role": "user", "content": str(i)} for i in range(21)])
# Before: one items.create(items=<21 items>) → API 400
# After: create(20) then create(1)
```

### Solution

- Split `add_items` into sequential creates of at most 20 items, preserving order.
- If a later create fails, delete the IDs from earlier successful creates (best-effort, reverse order) and re-raise, so the logical batch stays all-or-nothing per the session persistence contract.

### Test plan

- [x] 21 items → two creates (20 + 1)
- [x] 20 items → one create
- [x] later create failure deletes earlier chunk IDs
- [x] `uv run pytest tests/memory/test_openai_conversations_session.py tests/memory/test_session_limit.py` (56 passed)
- [x] `ruff` / `pyright` on the changed files
- [x] `make typecheck` passed

### Issue number

N/A. Related but different: closed #4228 was about `items.list` page size 1–100, not create batch size 20.

### Checks

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 from typing import Any
 
 from openai import AsyncOpenAI
@@ -10,6 +11,22 @@ from agents.models._openai_shared import get_default_openai_client
 from ..items import TResponseInputItem
 from .session import SessionABC
 from .session_settings import SessionSettings, coerce_session_settings, resolve_session_limit
+
+# Conversations items.create accepts at most 20 items per request.
+# See https://developers.openai.com/api/reference/resources/conversations/subresources/items/.
+_MAX_ITEMS_PER_CONVERSATION_CREATE = 20
+
+
+def _created_conversation_item_ids(result: object) -> list[str]:
+    data = getattr(result, "data", None)
+    if data is None:
+        return []
+    ids: list[str] = []
+    for item in data:
+        item_id = getattr(item, "id", None)
+        if isinstance(item_id, str) and item_id:
+            ids.append(item_id)
+    return ids
 
 
 async def start_openai_conversations_session(openai_client: AsyncOpenAI | None = None) -> str:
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Before merge, please make the multi-request append preserve the existing logical-batch semantics under cancellation and same-session concurrency.

`asyncio.CancelledError` bypasses the current `except Exception`, so cancellation after one chunk commits leaves a partial batch. Protect and await rollback before re-raising cancellation, and serialize the complete append/rollback phase against `get_items()`, `pop_item()`, `clear_session()`, and other `add_items()` calls on the same session instance. Please add controlled tests for cancellation and an overlapping surviving operation, asserting both operation results and final history.

hsusul:
Addressed the cancellation and same-session concurrency gaps.

- `add_items` now catches `asyncio.CancelledError` as well as `Exception`, shields and awaits rollback of already-created chunks, then re-raises cancellation.
- The full append/rollback holds `_session_lock`, which also serializes `get_items()`, `pop_item()`, `clear_session()`, and other `add_items()` calls on the same instance.
- Added controlled tests that cancel after the first 20-item create while a surviving `get_items` / `add_items` / `pop_item` is waiting, and assert both operation results and final history.

seratch:
Thanks for the follow-up. The cancellation handling and same-session serialization are improved, but the remaining failure mode is fundamental to the multi-request compensation approach.

After an earlier chunk commits, a rollback delete can fail. _delete_created_items() currently suppresses that failure, so add_items() raises the original create or cancellation error while the remote conversation retains a partial logical batch. Because the Conversations API does not provide an atomic transaction across these create requests, compensating deletes cannot satisfy the Session add_items() atomicity contract.

I do not think we should add a larger recovery protocol for this batch-limit workaround. I suggest closing this PR and keeping batches above the provider limit unsupported until we have an atomic provider primitive or deliberately define a different public contract.

Issue #4646, opened 2026-08-25
NextStepInterruption bypasses resumed Session-write recovery and loses tool guardrail results


FU-max-boop:
### Describe the bug

#4630 fixed failed resumed `Session` appends when resolving an approval advances the state to `NextStepRunAgain`. This report deliberately does not reopen that path.

A separate transition remains when one model response contains multiple approval-gated local tools and the caller resolves only a subset:

1. The first approved tool executes.
2. Its input/output tool guardrails and start/end hooks complete.
3. Another approval remains, so the advanced state is `NextStepInterruption`.
4. Persisting the completed tool output to a client-managed `Session` raises.
5. The state remains serializable and the second approval remains actionable, but the first output is not reconciled and its completed tool guardrail results are absent from durable state and every later public result.

This reproduces on both v0.22.0 and current `main` at `36976b1854534c0f184ed955afd542493abbb5d9`. It uses `ScriptedModel` and makes no provider or network call.

### Debug information

- Agents SDK version: v0.22.0 (`4df9ecfae1761ca6fea67cc5a20b383c1d492024`) and current main (`36976b1854534c0f184ed955afd542493abbb5d9`)
- Related library versions: none
- Python version: 3.12.13
- Operating system: macOS 26.5 arm64
- Model and model provider: `ScriptedModel`; no provider
- Does the issue reproduce with the latest Agents SDK release? Yes
- Does the issue occur consistently or intermittently? Consistently; 4/4 branch-cover rows reproduce on both refs

```text
The injected error is: RuntimeError: injected ambiguous renewed-interruption write failure
```

### Why #4630 does not cover this transition

The current pending-write implementation is gated to `NextStepRunAgain`:

- [`save_resumed_turn_items()` only supplies `resumed_write_state` for `NextStepRunAgain`](https://github.com/openai/openai-agents-python/blob/36976b1854534c0f184ed955afd542493abbb5d9/src/agents/run_internal/session_persistence.py#L717-L745).
- [`RunState` deserialization rejects a pending Session write unless the restored step is `NextStepRunAgain`](https://github.com/openai/openai-agents-python/blob/36976b1854534c0f184ed955afd542493abbb5d9/src/agents/run_state.py#L4356-L4383).
- In the non-streaming path, [the state advances before the resumed write, while tool guardrail results are published after it](https://github.com/openai/openai-agents-python/blob/36976b1854534c0f184ed955afd542493abbb5d9/src/agents/run.py#L1117-L1208).
[...]

PR #4651, opened 2026-08-25
fix(core): keep completed tool guardrail results when a resumed append fails

ayaangazali:
### Summary

On a resumed turn that resolves one approval while another stays pending, the run publishes that turn's tool guardrail results only inside the `NextStepInterruption` branch, and that branch sits after the resumed `Session` append:

```python
save_resumed_turn_items(...)            # can raise
...
if isinstance(turn_result.next_step, NextStepInterruption):
    tool_input_guardrail_results.extend(turn_result.tool_input_guardrail_results)
    tool_output_guardrail_results.extend(turn_result.tool_output_guardrail_results)
```

When that append raises, the approved tool has already executed and both its input and output guardrails have already returned, but nothing carries their results into `RunState`. The run-level lists die with the raised call, and `RunState` only ever receives guardrail results from a built result (`result.py`), which never happens on this path. Nothing republishes them later either, because the recovery resume does not re-execute the tool. So the results are gone permanently: absent from the checkpoint, from its serialization, and from the final `RunResult`.

Measured with an allow-only input guardrail and an allow-only output guardrail on the approved tool, across sync and streamed and across a live state and a JSON round trip. `ran` is how many times each guardrail actually executed:

```
before   ran=(1,1)   state=(0,0)  serialized=(0,0)  final=(0,0)     0/4 rows retain them
after    ran=(1,1)   state=(1,1)  serialized=(1,1)  final=(1,1)     4/4
```

The change publishes them next to the other resumed state updates, before the append, in both `run.py` and `run_loop.py`. Both paths already seed their run-level accumulators from `RunState` once at run start, before the turn loop, so writing the turn's results into state mid-turn cannot double count. Confirmed on the clean path, which still reports `(1, 1)` and not `(2, 2)` in both sync and streamed.

This is the second defect reported in #4646. It is independent of #4650: different root cause, different files, and the two diffs do not touch the same functions. #4650 makes the durable `Session` recover the dropped append; this one makes the completed guardrail results survive the same failure. Either can land without the other.

### Test plan

`tests/test_run_impl_resume_paths.py::test_failed_resumed_append_keeps_completed_tool_guardrail_results`, parametrized over streamed and non streamed and over a live state and a JSON round trip. It pauses on two approval-gated calls in one response, approves only the first, fails the resumed append, then asserts the checkpoint and its serialization each retain one input and one output guardrail result, and that the final result still reports them after the second approval is rejected.

Verified it fails without the source change by reverting `src/` and rerunning: all four cases fail.

`.agents/skills/code-change-verification/scripts/run.sh` passes end to end: format, lint, typecheck and the full test suite.

### Issue number

Partially addresses #4646

### Checks

---

Second half of the same issue, split out because the root cause is a different one and I did not want to bury two fixes in one diff. I'm a freshman in college and still learning this part of the runner, so the thing I'd most like checked is my claim that publishing before the append cannot double count. I reasoned it from the accumulators being seeded once at run start and then verified the clean path stays at one result each, but you know the ordering here far better than I do.

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1135,17 +1135,18 @@ class AgentRunner:
                                     generated_items=generated_items,
                                     session_items=session_items,
                                 )
-                                # Publish this turn's tool guardrail results before the
-                                # resumed Session append, which can raise and skip the
-                                # interruption branch below. The tool already ran.
-                                run_state._tool_input_guardrail_results = [
-                                    *tool_input_guardrail_results,
-                                    *turn_result.tool_input_guardrail_results,
-                                ]
-                                run_state._tool_output_guardrail_results = [
-                                    *tool_output_guardrail_results,
-                                    *turn_result.tool_output_guardrail_results,
-                                ]
+                                if isinstance(turn_result.next_step, NextStepInterruption):
+                                    # Publish this turn's tool guardrail results before the
+                                    # resumed Session append, which can raise and skip the
+                                    # interruption branch below. The tool already ran.
+                                    run_state._tool_input_guardrail_results = [
+                                        *tool_input_guardrail_results,
+                                        *turn_result.tool_input_guardrail_results,
+                                    ]
+                                    run_state._tool_output_guardrail_results = [
+                                        *tool_output_guardrail_results,
+                                        *turn_result.tool_output_guardrail_results,
+                                    ]
 
                             if (
                                 session_persistence_enabled
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1135,6 +1135,17 @@ class AgentRunner:
                                     generated_items=generated_items,
                                     session_items=session_items,
                                 )
+                                # Publish this turn's tool guardrail results before the
+                                # resumed Session append, which can raise and skip the
+                                # interruption branch below. The tool already ran.
+                                run_state._tool_input_guardrail_results = [
+                                    *tool_input_guardrail_results,
+                                    *turn_result.tool_input_guardrail_results,
+                                ]
+                                run_state._tool_output_guardrail_results = [
+                                    *tool_output_guardrail_results,
+                                    *turn_result.tool_output_guardrail_results,
+                                ]
 
                             if (
                                 session_persistence_enabled
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -1354,17 +1354,18 @@ async def start_streaming(
                             generated_items=generated_items,
                             session_items=streamed_result.new_items,
                         )
-                        # Publish this turn's tool guardrail results before the resumed
-                        # Session append, which can raise and skip the interruption
-                        # branch below. The tool already ran.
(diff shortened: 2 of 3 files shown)


ayaangazali:
Thanks, the streaming one was real and I have fixed it. The pre-approval one I could not reproduce, and the measurement disagrees with the premise, so I would rather show the numbers than quietly change the code.

**Copy resumed results into the streamed checkpoint: confirmed, fixed.**

Reproduced exactly as described. Streaming a sole pending approval makes the step `NextStepRunAgain`, and my publication ran on every resumed turn, so the source `RunState` and the detached `RunResultStreaming.to_state()` checkpoint disagreed after a failed append. That divergence did not exist before my change:

```
                 step                    source RunState   detached to_state()
main   tools=1   next_step_run_again     (0,0)             (0,0)   consistent
mine   tools=1   next_step_run_again     (1,1)             (0,0)   I introduced this
```

Rather than teach the streaming path to publish into both owners, I narrowed the change: both publications are now guarded on `NextStepInterruption`. Run-again accounting is byte for byte what it was on `main`, and the two checkpoints agree again:

```
tools=2  next_step_interruption   source (1,1)   detached (1,1)   MATCH   (the reported bug, fixed)
tools=1  next_step_run_again      source (0,0)   detached (0,0)   MATCH   (unchanged from main)
```

That is a smaller diff than the one you reviewed, not a larger one, and it removes the second owner problem instead of adding a branch for it.

**Avoid duplicating pre-approval guardrail results on retry: I do not think this one holds.**

The premise is that the established count is one. On `main`, with `pre_approval_tool_input_guardrails=True` and a renewed interruption, the guardrail genuinely runs twice and the clean path already stores both results:

```
main, pre_approval=True, renewed interruption, clean path:   ran=(2,1)   state=(2,1)
```

So two is the established count for that configuration, not one. My change makes the failure path match that clean path rather than diverge from it:

```
                              clean     failed append
main   pre_approval=False     (1,1)     (0,0)
main   pre_approval=True      (2,1)     (1,0)
mine   pre_approval=False     (1,1)     (1,1)
mine   pre_approval=True      (2,1)     (2,1)
```

Both modes now agree with their own clean path, which is the invariant I was aiming for. And after the narrowing above, the `NextStepRunAgain` resume you were protecting is not touched at all, so the run-again exclusion is preserved by construction rather than by my getting the arithmetic right.

Happy to be shown wrong on the second one if the intended contract really is one result per call regardless of the pre-approval recheck. In that case the clean path looks like it has the same problem, and I would rather fix that separately than special case the failure path.

seratch:
Thanks for the careful work on the second part of #4646. #4654 now covers the same remaining guardrail-result publication gap after #4650, but places the durable update immediately at the renewed-interruption persistence boundary and includes the stronger recovery matrix: streaming and non-streaming, live and JSON state, both ambiguous Session outcomes, and pre-approval guardrail accounting.

To keep one implementation and one regression suite for this invariant, we are going to proceed with #4654 and close this PR as superseded. The underlying issue and contribution are valid; this is only a choice between overlapping implementation paths.

PR #4653, opened 2026-08-25
fix(mcp): reject duplicate tools from one server

hsusul:
### Summary

This pull request fixes MCP tool discovery accepting duplicate tool names returned by a single server. Because both definitions share the same invocation identity, discovery now raises a deterministic `UserError` before conversion in both prefixed and unprefixed listing paths.

### Test plan

- Added deterministic coverage for same-server duplicates with server-name prefixing both disabled and enabled.
- Preserved existing cross-server collision and prefix behavior.
- Ran `.agents/skills/code-change-verification/scripts/run.sh` after `make sync`; formatting, lint, type checking, and the full test suite passed.

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/mcp/util.py
+++ b/src/agents/mcp/util.py
@@ -281,6 +281,7 @@ class MCPUtil:
             server_tool_batches = []
             for server_index, server in enumerate(servers):
                 listed_tools = await cls._list_tools_with_span(server, run_context, agent)
+                cls._validate_unique_tool_names(server, listed_tools)
                 server_tool_batches.append((server_index, server, listed_tools))
 
             prefixed_tool_name_overrides = cls._build_prefixed_tool_name_overrides(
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. I traced this through the released tool-collision path. Same-server duplicates without prefixing are already handled by `RunConfig.tool_name_collision_policy`: the default `"warn"` behavior exposes only the current dispatch winner, while `"error"` rejects the collision before the model is called.

The distinct prefixed case is currently demonstrated only with `FakeMCPServer`. MCP says tool names should be unique within a server, and the real-world report in #464 concerns collisions across different servers, so it does not establish this same-server need.

I am going to close this PR for now. We can reconsider a narrowly scoped fix if there is a real MCP server response with duplicate same-server names and a caller-visible incorrect result under the current behavior.

PR #4654, opened 2026-08-25
fix(sessions): persist resumed tool guardrail results

### LINKED ISSUE #4646: NextStepInterruption bypasses resumed Session-write recovery and loses tool guardrail results

FU-max-boop:
### Summary

This pull request fixes a recovery gap after resolving interrupted function-tool approvals with a client-managed Session. When an approved tool, its input and output guardrails, and its lifecycle hooks completed but the resumed Session append raised, the caller retained a `RunState` that omitted the completed guardrail audit results. A later retry could reconcile the tool call and output through #4650, but it could not reconstruct those results without rerunning user code.

- Publish the accepted guardrail-result prefix plus the current resumed step's results to the same `RunState` before the fallible Session append in both runner paths.
- Keep caller-mutated streaming result lists outside the durable ownership boundary by sourcing the streaming prefix from accepted state.
- Preserve the existing public result types, ordering, and RunState schema.

### Test plan

- Added a four-row public Runner matrix covering same-mode and cross-mode recovery, live and JSON state, and both before-commit and commit-then-raise Session outcomes.
- Enabled `pre_approval_tool_input_guardrails` in one cross-mode JSON row and verified that clean and failed-recovery paths both publish the established two input results and one output result without an extra tool effect, hook, or model call.
- Verified exactly-once tool effects, input/output guardrails, start/end hooks, Session history, public results, durable state, and final model input.
- Ran `.agents/skills/code-change-verification/scripts/run.sh` on the final commit and current `main`; formatting, lint, typecheck, and the full test suite all passed.

### Issue number

Closes #4646

### Checks

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1135,6 +1135,15 @@ class AgentRunner:
                                     generated_items=generated_items,
                                     session_items=session_items,
                                 )
+                                if isinstance(turn_result.next_step, NextStepInterruption):
+                                    run_state._tool_input_guardrail_results = [
+                                        *tool_input_guardrail_results,
+                                        *turn_result.tool_input_guardrail_results,
+                                    ]
+                                    run_state._tool_output_guardrail_results = [
+                                        *tool_output_guardrail_results,
+                                        *turn_result.tool_output_guardrail_results,
+                                    ]
 
                             if (
                                 session_persistence_enabled
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -1377,6 +1377,15 @@ async def start_streaming(
                         )
 
                     if isinstance(turn_result.next_step, NextStepInterruption):
+                        if run_state is not None:
+                            run_state._tool_input_guardrail_results = [
+                                *accepted_tool_input_guardrail_results,
+                                *turn_result.tool_input_guardrail_results,
+                            ]
+                            run_state._tool_output_guardrail_results = [
+                                *accepted_tool_output_guardrail_results,
+                                *turn_result.tool_output_guardrail_results,
+                            ]
                         await _finalize_streamed_interruption(
                             streamed_result=streamed_result,
                             save_items=_save_resumed_items,
--- a/tests/test_runner_guardrail_resume.py
+++ b/tests/test_runner_guardrail_resume.py
@@ -1,13 +1,22 @@
 from types import SimpleNamespace
-from typing import Any, cast
+from typing import Any, Literal, cast
 
 import pytest
 from openai.types.responses import ResponseFunctionToolCall
 
 import agents.run as run_module
-from agents import Agent, Runner
+from agents import (
+    Agent,
+    Runner,
+    ToolInputGuardrailData,
+    ToolOutputGuardrailData,
+    function_tool,
+)
 from agents.guardrail import GuardrailFunctionOutput, InputGuardrail, InputGuardrailResult
-from agents.items import ModelResponse, ToolApprovalItem
+from agents.items import ModelResponse, ToolApprovalItem, TResponseInputItem
+from agents.lifecycle import RunHooks
+from agents.memory import Session
+from agents.run import RunConfig
 from agents.run_context import RunContextWrapper
 from agents.run_internal.run_steps import (
     NextStepFinalOutput,
@@ -16,6 +25,7 @@ from agents.run_internal.run_steps import (
 )
 from agents.run_state import RunState
 from agents.testing import ScriptedModel
+from agents.tool import Tool
 from agents.tool_guardrails import (
     AllowBehavior,
     ToolGuardrailFunctionOutput,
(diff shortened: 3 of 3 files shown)


mohammedmessaoudene-cmd:
One focused coverage question, prompted by the P1 review on #4651:

Could the four-row matrix add a `ToolExecutionConfig(pre_approval_tool_input_guardrails=True)` case for the renewed-interruption path? In that supported mode the input guardrail runs before approval and is rechecked after approval. The current test uses the default post-approval-only path, so it does not pin whether a failed append plus recovery preserves the accepted/public guardrail count exactly once rather than duplicating the pre-approval result.

The new scoping to `NextStepInterruption` and the streaming use of accepted prefixes appear designed to avoid the two #4651 findings; this row would make that distinction executable across live/JSON and streaming/non-streaming recovery.

I reviewed the current diff and the related merged #4650 path, but I have not rerun the suite.

AI-assistance disclosure: ChatGPT assisted with review and drafting; Mohammed Messaoudene reviewed the claim and remains responsible for it.

FU-max-boop:
Thanks for the focused coverage question. I added it in 58e0e021 without expanding the four-row matrix: the existing run-to-streamed JSON commit-then-raise row now enables `pre_approval_tool_input_guardrails`.

One important oracle detail: the clean renewed-interruption path intentionally runs the input guardrail before approval and rechecks it after approval, so its accepted/public counts are 2 input and 1 output, not 1/1. The injected failure and recovery path now asserts the same 2/1 counts through the failed checkpoint, JSON round-trip, cross-mode recovery, final result, and `to_state()`, while the tool effect, hook pair, and model work remain exactly once.

The focused file is 6/6 green, and the repository format, lint, typecheck, and full test stack pass on the pushed head.

mohammedmessaoudene-cmd:
Thanks — the 2-input / 1-output oracle is the important correction. I reviewed the `58e0e021` diff, and the updated row now pins the property I was asking about: a failed checkpoint, JSON round-trip, and cross-mode recovery preserve the accepted/public guardrail counts while the tool effect, hook pair, and model work remain exactly once.

That resolves my coverage question. I have not independently rerun the suite, so I am treating the reported focused 6/6 and full-stack pass as the PR author's execution result rather than an external reproduction.

AI-assistance disclosure: ChatGPT assisted with delta review and drafting; Mohammed Messaoudene reviewed the bounded conclusion and remains responsible for it.

PR #4656, opened 2026-08-25
fix: strip empty provider data from Responses input

Excelius-Wang:
### Summary

Fix Responses input sanitization so an empty SDK-only `provider_data` field cannot reach the OpenAI wire payload.

The sanitizer now detects the field by key presence instead of value truthiness. This preserves real response IDs, keeps the existing placeholder-ID and provider-specific reasoning behavior, and sanitizes only the request copy rather than caller-owned input.

### Test plan

- `uv run pytest -q tests/models/test_remove_openai_responses_api_incompatible_fields.py` (`11 passed`)
- `make format`
- `make lint`
- `make typecheck`
- `make tests-review` (`9258 passed, 35 skipped` across parallel and serial suites)
- `.agents/skills/code-change-verification/scripts/run.sh` reached `9200 passed, 30 skipped` before five unrelated failures: four SQLite multiprocessing timing cases passed on focused sequential rerun, while one `review_optional` UnixLocal sandbox test still fails because this host resets `PATH` inside `sh -lc` and cannot see the fixture-provided mock `rclone`. The failure occurs during sandbox mount setup before the Responses code path is exercised.

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -1052,7 +1052,7 @@ class OpenAIResponsesModel(Model):
         # independently of it.
         needs_cleaning = any(
             isinstance(item, dict)
-            and (item.get("provider_data") or item.get("id") == FAKE_RESPONSES_ID)
+            and ("provider_data" in item or item.get("id") == FAKE_RESPONSES_ID)
             for item in list_input
         )
         if not needs_cleaning:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. The sanitizer correctly removes non-empty provider metadata today, but I could not find a supported SDK or provider-adapter path that emits provider_data={} into Responses input. The new coverage injects that value manually and does not show a normal producer or a failing provider request.

That proves the branch is constructible, but not that it represents a supported user-facing defect worth changing and maintaining. I am going to close this PR for now. If a supported adapter or replay path produces the empty field, please provide that source path and the resulting provider error so we can reconsider the focused fix.

PR #4658, opened 2026-08-25
append Orq.ai to tracing.md

maxhilding:
append Orq.ai and link to integration docs to tracing.md


seratch:
The service supports API proxy and OTel support, but it doesn't support OpenAI's tracing data interface. Thus, we won't add the link for now.

PR #4662, opened 2026-08-25
fix(items): preserve empty approval params

hsusul:
### Summary

This pull request fixes `ToolApprovalItem.arguments` treating empty `params` as absent and falling through to a different `input` payload. Explicit `None` checks now preserve present falsey argument values consistently for mapping-backed and attribute-backed provider items.

### Test plan

- Added mapping coverage for `params={}` alongside a different `input` value.
- Added attribute-backed coverage with `arguments=None` and empty `params`.
- Ran `.agents/skills/code-change-verification/scripts/run.sh`; formatting, lint, type checking, and the full test suite passed.

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/items.py
+++ b/src/agents/items.py
@@ -648,13 +648,15 @@ class ToolApprovalItem(RunItemBase[Any]):
         if isinstance(self.raw_item, dict):
             candidate = self.raw_item.get("arguments")
             if candidate is None:
-                candidate = self.raw_item.get("params") or self.raw_item.get("input")
-        elif hasattr(self.raw_item, "arguments"):
-            candidate = self.raw_item.arguments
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. I could not find a supported SDK producer that emits the decisive raw-item shape used by these tests: an empty params value together with a different input value. The added tests construct that collision manually, so they establish that the fallback can be changed, but not that normal approval handling currently returns an incorrect value.

Changing the fallback precedence would select a new semantic contract for an ambiguous synthetic shape without a demonstrated user outcome. I am going to close this PR for now. If a supported provider or SDK path produces this exact shape, please share the raw item and the caller-visible incorrect approval result so we can revisit the narrow contract.

PR #4664, opened 2026-08-25
examples: add HOL Guard command inspection input guardrail example

### LINKED ISSUE #4659: Docs/example: HOL Guard command inspection with FunctionTool input guardrails
### Please read this first

- **Have you read the docs?** Yes. The current guardrails docs say tool guardrails wrap `FunctionTool` instances, input tool guardrails run before execution, and they can skip a call, replace the output, or raise a tripwire.
- **Have you searched for related issues?** Yes. The only exact HOL Guard hit is #4581, which asked for a new provider-neutral contract across built-in execution tools and was closed. This proposal is intentionally different: it uses the already-shipped `FunctionTool` guardrail surface and asks only for a maintained runnable example/docs artifact.

### Describe the feature

Would maintainers be open to an official example showing how a command-bearing `FunctionTool` can invoke **HOL Guard** from `tool_input_guardrails` before the underlying function executes?

A small example could:

- install HOL Guard with `pipx install hol-guard` (or document an equivalent isolated install);
- define a custom function tool such as `run_command(command: str)`;
- attach an input tool guardrail that invokes `hol-guard command test '<command>' --json` before the function handler;
- permit execution only for an explicit safe classification and otherwise skip/tripwire so the command handler is called zero times;
- fail closed when HOL Guard is unavailable, times out, returns malformed output, or cannot classify the command;
- include a tiny test/counter demonstrating that a blocked command never reaches the wrapped tool.

HOL Guard documents `command test` as **side-effect-free command inspection**. It does not execute the command, evaluate final Guard policy, create approvals, or record receipts, so the example should state that boundary clearly rather than present it as comprehensive Guard runtime or Cloud enforcement.

This does **not** ask for a new SDK API and does not attempt to cover hosted tools or built-in `ShellTool` / `ApplyPatchTool` / `ComputerTool` execution paths from #4581. It is a concrete third-party integration example for the existing `FunctionTool` pre-execution boundary.

If this fits the examples/docs direction, I’m happy to contribute the smallest artifact at the placement maintainers prefer.

AAliKKhan:
### Summary

Adds `examples/tools/hol_guard_command_inspection.py`, a runnable example of the `FunctionTool` pre-execution boundary: an input tool guardrail invokes the external [HOL Guard](https://pypi.org/project/hol-guard/) CLI in its side-effect-free `command test` mode before a shell-running tool executes, fulfilling the concrete request in #4659.

### What the example shows

- A command-bearing `FunctionTool` (`run_command(command: str)`) built with `@tool`
- An input guardrail that calls `hol-guard command test '<command>' --json` and permits execution only for an explicit safe classification
- **Fail-closed semantics**: missing binary, timeout, non-zero exit, malformed JSON, or unrecognized verdicts all reject the call - inspection unavailability never opens the execution path
- An execution counter demonstrating that a blocked command never reaches the wrapped function
- Explicit boundary documentation: `command test` inspects the command string only; it does not execute anything or represent comprehensive Guard runtime/Cloud enforcement

No SDK API changes; purely additive example following the existing `examples/basic/tool_guardrails.py` conventions.

### Test plan

- [x] Example compiles and passes `ruff format` / `ruff check`
- [x] Verified offline with `ScriptedModel`: a scripted destructive `run_command` call is rejected by the guardrail (fail closed when HOL Guard is unavailable) and the execution counter confirms the wrapped function was never invoked
- [x] With HOL Guard installed, allowed commands pass through to the tool as shown in the example's expected-output walkthrough

Fixes #4659

### DIFF
--- a/examples/tools/hol_guard_command_inspection.py
+++ b/examples/tools/hol_guard_command_inspection.py
@@ -1,0 +1,165 @@
+import asyncio
+import json
+import shutil
+import subprocess
+
+from agents import (
+    Agent,
+    Runner,
+    ToolGuardrailFunctionOutput,
+    ToolInputGuardrailData,
+)
+from agents.decorators import tool, tool_input_guardrail
+
+"""Inspect shell commands with HOL Guard before a FunctionTool executes.
+
+This example demonstrates the pre-execution boundary of `FunctionTool` input tool
+guardrails. A guardrail invokes the external `hol-guard` CLI in its side-effect-free
+`command test` mode before the wrapped function runs, and only an explicitly safe
+classification lets the call through.
+
+Important boundary: `hol-guard command test` inspects the command string only. It does
+not execute the command, evaluate final Guard policy, create approvals, or record
+receipts. Treat it as one layer of defense, not as comprehensive runtime enforcement.
+
+Install HOL Guard before running this example:
+
+    pipx install hol-guard
+
+The example fails closed: if `hol-guard` is missing, times out, returns malformed JSON,
+or does not clearly classify the command as safe, the tool call is rejected and the
+wrapped function is never invoked.
+"""
+
+# Counts how many times the protected function body actually ran. The main() demo uses
+# this to prove that blocked commands never reach the tool implementation.
+EXECUTED_COMMANDS = 0
+
+HOL_GUARD_TIMEOUT_SECONDS = 10.0
+
+
+def _classify_with_hol_guard(command: str) -> dict[str, object] | None:
+    """Run `hol-guard command test` and return its parsed JSON verdict, or None.
+
+    Returning None means the inspection itself failed (missing binary, timeout,
+    non-zero exit, or malformed output), which the guardrail treats as fail closed.
+    """
+    binary = shutil.which("hol-guard")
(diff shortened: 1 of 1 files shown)


kantorcodes (commented):
HOL Guard contract correction before merge: current `hol-guard command test <command> --json` does **not** return `classification` as a string like `"safe"`. `classification` is an object. For this inspection-only example, the conservative allow condition should require both `classification.explicitly_benign == true` and `minimum_action == "allow"`; `policy_evaluation` remains `"not_run"` and should not be presented as a final Guard policy decision. As written, `_is_explicitly_safe()` will reject current HOL Guard output even for benign commands because it only accepts a string-valued `classification`/`verdict`. Please update the parser and its tests to the current JSON contract. Current source of truth: `src/codex_plugin_scanner/guard/runtime/command_inspection.py` on HOL Guard `release/3.0`. The separate Codex P2 about synchronous subprocess work on the runner event loop also looks worth addressing.

seratch:
Closing this for the reason I mentioned at https://github.com/openai/openai-agents-python/issues/4659#issuecomment-5419173869

PR #4671, opened 2026-08-26
fix(repl): read the demo loop prompt off the event loop

### LINKED ISSUE #4670: `run_demo_loop` blocks the asyncio event loop on `input()`
**Area:** `src/agents/repl.py:40`

### Description

`run_demo_loop` is an `async def`, but it reads the user's turn with the blocking builtin `input()`:

```python
async def run_demo_loop(agent, *, stream=True, context=None, max_turns=DEFAULT_MAX_TURNS) -> None:
    ...
    while True:
        try:
            user_input = input(" > ")   # blocks the whole event loop
```

While the prompt is waiting for a keystroke, the event loop is frozen. Nothing else scheduled on that loop can make progress:

- MCP servers connected over SSE / streamable-HTTP cannot service their read streams, so long idle periods at the prompt can let the server-side session time out or drop.
- Any user task launched alongside `run_demo_loop` (`asyncio.gather`, background `create_task`) is stalled for the entire time the human is thinking.
- `asyncio` signal handling is not serviced, so Ctrl-C behaves inconsistently across platforms while the loop sits in `input()`.

`ruff --select ASYNC` flags this directly:

```
src/agents/repl.py:40:26: ASYNC250 Blocking call to `input()` in async context
```

### Reproduction

```python
import asyncio
from agents import Agent, run_demo_loop

async def heartbeat():
    while True:
        print("tick", flush=True)
        await asyncio.sleep(1)

async def main():
    asyncio.create_task(heartbeat())
    await run_demo_loop(Agent(name="demo"))

asyncio.run(main())
```

`tick` stops printing as soon as the prompt appears and only resumes after Enter is pressed.

### Suggested fix

```python
user_input = await asyncio.to_thread(input, " > ")
```

`EOFError` / `KeyboardInterrupt` still propagate out of the thread and are caught by the existing `except` clause.

rajarshidattapy:
This pull request fixes `run_demo_loop` freezing the asyncio event loop while it waits for the user's next turn, without leaving a stdin reader the caller cannot stop.

### The bug

`run_demo_loop` is declared `async def` but read each turn with the blocking builtin `input()`:

```python
user_input = input(" > ")
```

Nothing else on that loop can run while the prompt waits for a keystroke. MCP servers connected over SSE / streamable-HTTP cannot service their read streams, so a long pause at the prompt can let the server-side session time out; any task the caller started alongside the loop is stalled for the whole time the human is thinking.

### Why this is not simply `asyncio.to_thread(input, ...)`

That was the first version of this PR, and it traded the stall for a hang. Ctrl+C cancels the loop task, but nothing can interrupt a worker blocked in `input()`, and the default executor is joined while the loop shuts down — so the process stays stuck until stdin delivers another line or EOF. Cancelling the read and measuring how long the process then takes to exit, with a stdin that never delivers:

```
old (await asyncio.to_thread(input, " > ")):  killed by a 20s timeout, rc=124
new (_read_user_input(" > ")):                asyncio.run returned at 0.31s,
                                              interpreter exit at 0.31s, rc=0
```

A daemon thread is not the answer either: one blocked in a buffered stdin read at interpreter shutdown raises CPython's `_enter_buffered_busy` fatal error and the process dies on a segfault.

### The change

The reader waits on the stdin *descriptor* in `_STDIN_POLL_INTERVAL_SECONDS` slices and checks a stop event between them, so it can be told to stop. On cancellation, `_read_user_input` sets that event and waits for the reader before re-raising, so when `run_demo_loop` is done no worker is left sitting on stdin — and that wait is bounded by one poll interval rather than by the user pressing Enter.

Waiting on the descriptor first is what makes the following `input()` safe: a terminal in canonical mode only reports readiness once Enter is pressed, so the read consumes an already-complete line instead of blocking. Verified against a pty — a partial line leaves `select` unready, and readiness arrives with the newline.

Where stdin cannot be waited on, the prompt is read on the event loop exactly as before rather than starting a worker that could not be stopped. That covers Windows, where `select` accepts only sockets and the console reports readiness per keystroke rather than per line, and any captured or synthetic stdin with no descriptor. Those platforms keep the original stall; they do not get a hang in exchange.

### Behavior

- `EOFError` and `KeyboardInterrupt` still reach the existing handler, so Ctrl+D and the `exit` / `quit` words end the loop as before.
- The public signature of `run_demo_loop` is unchanged.
- On the descriptor path the prompt is written before the wait, so `input()` is called without one. If the host program has imported `readline`, its history and arrow-key editing are not active for that read; the terminal's own canonical-mode editing still is. A plain script that has not imported `readline` sees no difference.
- `ruff --select ASYNC` reports `ASYNC250` on the deliberate on-loop fallback. No `# noqa` is added, because the pinned `ruff==0.9.2` predates that rule and `RUF100` would then reject the directive.

### Tests

*Event-loop responsiveness* — the bounded test from the first round stays, now backed by a real pipe so it exercises the reader path: it holds the prompt open until a coroutine scheduled alongside the loop has run, so reading on the loop deadlocks instead of passing.

*Cancellation and Ctrl+C exit* — two tests cancel the loop while the reader is waiting on a pipe with nothing in it. One asserts the reader function returned before the coroutine finished, which is the property that keeps a worker off stdin. The other asserts cancellation settles well inside the poll interval.

A third test pins that stdin without a usable descriptor reports as non-pollable, so the on-loop fallback is the documented path rather than an accident.

The descriptor-path tests are skipped on Windows, where that path does not exist. Verified on both: Linux 8 passed; Windows 5 passed, 3 skipped.

Resolves #4670.


seratch (changes requested):
Thanks. The event-loop stall is real, but I do not think the current `asyncio.to_thread(input, ...)` implementation is safe to merge.

With a normal `asyncio.run()` entry point, Ctrl+C cancels the event-loop task while input() can remain blocked in the default-executor worker. Executor shutdown then waits for that worker, so the process can remain stuck until stdin receives another line or EOF.

Please replace this with an input path that keeps the event loop responsive and has explicit cancellation and shutdown ownership: when `run_demo_loop()` exits, no worker may remain blocked on stdin. Please add deterministic coverage for both event-loop responsiveness and the cancellation/Ctrl+C exit path. The current bounded responsiveness test is useful and can remain.

rajarshidattapy:
I'll do the necessary changes.

rajarshidattapy:
You are right, and thank you for catching it — that hang is exactly the failure mode, and I should not have proposed shipping it with the caveat rather than fixing it. Pushed a replacement.

Measured with a stdin that never delivers, cancelling the read the way `Runner._on_sigint` does:

```
old (await asyncio.to_thread(input, " > ")):  killed by a 20s timeout, rc=124
new (_read_user_input(" > ")):                asyncio.run returned at 0.31s,
                                              interpreter exit at 0.31s, rc=0
```

**What changed.** The reader now waits on the stdin descriptor in 0.1s slices with a stop event between them, so it can be told to stop. On cancellation `_read_user_input` sets that event and waits for the reader before re-raising: when `run_demo_loop` is done no worker is left on stdin, and the wait is bounded by one poll interval instead of by the user pressing Enter.

Waiting on the descriptor first is what makes the following `input()` safe — a canonical-mode terminal only reports readiness once Enter is pressed, so the read consumes a complete line. I verified that against a pty rather than assuming it: a partial line leaves `select` unready, readiness arrives with the newline.

For completeness on the alternative: a daemon thread avoids the join but a daemon blocked in a buffered stdin read hits CPython's `_enter_buffered_busy` fatal error at interpreter shutdown and the process segfaults, so it is strictly worse.

**Where stdin cannot be waited on** — Windows, where `select` takes only sockets and the console signals per keystroke rather than per line, and any captured stdin with no descriptor — the prompt is read on the loop exactly as before. Those keep the original stall rather than trading it for a hang, and I would rather leave that visible than start a worker I cannot stop. Happy to follow up with a Windows console reader if you want the fix there too.

**Coverage.** The bounded responsiveness test stays, now backed by a real pipe so it takes the reader path. Two new tests cancel the loop while the reader is waiting: one asserts the reader returned before the coroutine finished, which is the property that keeps a worker off stdin, and one asserts cancellation settles inside the poll interval. A third pins that a descriptor-less stdin reports non-pollable, so the on-loop fallback is a documented path rather than an accident. Descriptor-path tests skip on Windows. Linux 8 passed; Windows 5 passed, 3 skipped; mypy clean on both.

One note: `ruff --select ASYNC` flags `ASYNC250` on the deliberate on-loop fallback. I did not add a `# noqa`, because the pinned `ruff==0.9.2` predates that rule and `RUF100` would reject the directive. Glad to handle it differently if you have a preference.

seratch:
https://github.com/openai/openai-agents-python/issues/4670#issuecomment-5446434515

PR #4672, opened 2026-08-26
docs: add SteadIO to external tracing processors list

groovefx:
Adds [SteadIO](https://www.steadio.ai/docs/sdk-integrations) to the external tracing processors list in `docs/tracing.md`.

SteadIO provides per-customer LLM cost visibility, hard spend caps, and usage rebilling. The `TracingProcessor` aggregates token usage across agent runs and reports cost and cap state per customer.

Setup docs: https://www.steadio.ai/docs/sdk-integrations


seratch:
Thanks, but the linked page documents routing model requests through SteadIO by replacing the default OpenAI client and forcing Chat Completions. It does not implement or document an Agents SDK TracingProcessor, add_trace_processor(), or set_trace_processors() integration.

This section is specifically for integrations that support the Agents SDK tracing API surface, so adding this link here would be misleading. I am closing this for now. We can revisit it once SteadIO has a documented tracing processor integration.

PR #4678, opened 2026-08-26
fix(sandbox): keep UnixLocal filesystem and tar work off the event loop

### LINKED ISSUE #4675: UnixLocalSandbox blocks the event loop with synchronous filesystem and tar I/O
### Please read this first

- **Have you searched for related issues?** Yes. Nothing covers blocking filesystem work in the sandbox backends.

### Describe the bug

`UnixLocalSandboxSession` calls synchronous `pathlib` and `tarfile` APIs directly from `async def` methods, so those methods block the event loop for the full duration of the I/O instead of yielding.

The two unbounded ones archive or extract the entire workspace tree:

```python
async def persist_workspace(self) -> io.IOBase:
    root = Path(self.state.manifest.root)
    if not root.exists():                        # blocking stat
        ...
    buf = io.BytesIO()
    with tarfile.open(fileobj=buf, mode="w") as tar:
        tar.add(root, arcname=".", filter=...)   # blocking walk + read of the whole tree

async def hydrate_workspace(self, data: io.IOBase) -> None:
    root = Path(self.state.manifest.root)
    root.mkdir(parents=True, exist_ok=True)
    with tarfile.open(fileobj=data, mode="r:*") as tar:
        safe_extract_tarfile(tar, root=root, ...)  # blocking extraction of the whole tree
```

Both sit directly on the snapshot path: `persist_snapshot()` awaits `session.persist_workspace()` and `restore_snapshot_into_workspace_on_resume()` awaits `session.hydrate_workspace()` (`src/agents/sandbox/session/snapshot_lifecycle.py`). A workspace holding a `node_modules`, a build directory, or a model checkpoint takes seconds to minutes to tar, and the whole application's loop is stalled for that window: concurrent agent runs, streamed responses, MCP sessions, and tracing exports all freeze. That rules the local sandbox out of any server process handling more than one run at a time.

**Scope note (edited after review feedback):** this report originally also listed the bounded `mkdir`, `exists`, and `resolve` calls in `_prepare_backend_workspace`, `_resolved_exec_context`, `_exec_internal`, `pty_exec_start` and `_write_stream_with_exec`. Those are possible cleanup candidates, but a static `ASYNC240` report on its own does not establish a user-visible defect for them, and I have no measurement showing one. The defect claimed here is limited to the two unbounded archive paths above.

### The pattern already exists in this file

This is a gap rather than a design decision: the same module already offloads its other blocking work, with `loop.run_in_executor(None, os.read, ...)` for PTY reads (line 517) and `asyncio.to_thread(_close_fd_quietly, fd)` for fd cleanup (line 616). `sandbox/session/sinks.py` uses `asyncio.to_thread` too. The filesystem and tar paths were simply missed.

### Why CI does not catch it

`ruff --select ASYNC` reports the two archive methods (along with five bounded sites that this report does not claim as a defect):

```
src/agents/sandbox/sandboxes/unix_local.py:1079:16: ASYNC240 Async functions should not use pathlib.Path methods
src/agents/sandbox/sandboxes/unix_local.py:1110:13: ASYNC240
```

`make lint` stays green because `pyproject.toml` pins `ruff==0.9.2` in the dev group, which predates the `ASYNC240` rule. Reproducing the report needs a newer ruff (checked with 0.16.3).

### Debug information

- Agents SDK version: `main` at e773b154, and v0.22.0
- Python version: 3.13
- Operating system: Linux (the module raises `ImportError` on Windows by design)
- Model and model provider: n/a — no model or network involved
- Does the issue reproduce with the latest Agents SDK release? Yes, and on `main`.

### Repro steps

Any coroutine sharing the loop stops making progress for the duration of the archive:

```python
import asyncio, time

async def heartbeat(stop):
    last = time.monotonic()
    worst = 0.0
    while not stop.is_set():
        await asyncio.sleep(0.05)
        now = time.monotonic()
        worst = max(worst, now - last)
        last = now
    print(f"worst loop stall: {worst:.2f}s")

async def main(session):          # a started UnixLocalSandboxSession whose
    stop = asyncio.Event()        # workspace holds a few hundred MB
    task = asyncio.create_task(heartbeat(stop))
    await session.persist_workspace()
    stop.set()
    await task
```

The reported stall tracks the tar duration, instead of staying near the 0.05s tick.

### Expected behavior

These methods should not hold the event loop. Move the blocking bodies onto a worker thread, matching what the rest of the module already does.

That needs explicit cancellation ownership to be correct. `asyncio.to_thread` cannot interrupt the thread it starts, so a cancelled caller would return while the worker is still walking or writing the workspace — and callers act on that return immediately: `restore_snapshot_into_workspace_on_resume` closes the archive stream in a `finally`, and resume clears the workspace root. So these methods must wait for the worker to settle before reporting cancellation.

I have a fix with regression tests and will open a PR shortly.

rajarshidattapy:
This pull request fixes `persist_workspace` and `hydrate_workspace` holding the asyncio event loop while they archive or extract the entire workspace tree.

### The bug

Both methods are `async def` but do their work with synchronous `tarfile` and `pathlib` calls, and both sit on the snapshot lifecycle — `persist_snapshot()` awaits `persist_workspace()`, and `restore_snapshot_into_workspace_on_resume()` awaits `hydrate_workspace()`:

```python
async def persist_workspace(self) -> io.IOBase:
    ...
    with tarfile.open(fileobj=buf, mode="w") as tar:
        tar.add(root, arcname=".", filter=...)   # walks and reads the whole tree
```

A workspace holding a `node_modules`, a build directory, or a model checkpoint takes seconds to minutes to archive, and for that whole window nothing else on the loop runs: concurrent agent runs, streamed responses, MCP sessions, and tracing exports all freeze.

### The change

The blocking bodies move into two module-level helpers, `_archive_workspace` and `_extract_workspace_archive`, run on a worker thread. Session state is read on the loop — `persist_workspace` reads `manifest.root` and `_persist_workspace_skip_relpaths()` itself — and only plain values cross into the thread.

**Cancellation ownership.** `asyncio.to_thread` cannot interrupt the thread it starts, so awaiting it plainly would report the operation as stopped while the worker is still walking or writing the workspace. Callers act on that return immediately: the snapshot lifecycle closes the archive stream in a `finally` as soon as the await returns, and resume clears the workspace root. Either would land on a live worker.

`_run_workspace_io` therefore records the cancellation, lets the worker settle, and only then propagates it. Repeated cancellation is absorbed the same way, and a worker failure that arrives after cancellation is consumed rather than surfacing instead of the cancellation. This follows `memory.sqlite_session._await_mutation`, which owns cancellation for session writes; if you would rather hoist that helper into a shared util than have a second copy, say the word.

### Scope

Per review, this is limited to the two operations with a demonstrated defect. The bounded `mkdir`, `exists`, and `resolve` calls in `_prepare_backend_workspace`, `_resolved_exec_context`, `_exec_internal`, `pty_exec_start`, and `_write_stream_with_exec` are unchanged, so `ruff --select ASYNC` still reports `ASYNC240` at those five sites. Leaving them alone also keeps this change off the exec confinement path.

### Behavior

- Error types, messages, and context payloads are unchanged. The helpers keep the same guards and raise the same `WorkspaceArchiveReadError` and `WorkspaceArchiveWriteError`, including the `workspace_root_not_found` reason.
- No public API change, and no change to any private signature outside these two methods.

### Tests

`tests/sandbox/test_unix_local.py` covers both properties the review asked for.

*Event-loop responsiveness* — two tests hold the archive/extract helper open until a coroutine scheduled alongside it has run, so doing that work on the loop deadlocks instead of passing.

*Cancellation* — three tests hold the worker open, cancel the task, and assert it has neither completed nor let the worker settle, then release the worker and assert `CancelledError` surfaces only afterwards. One covers repeated cancellation. All three fail against a plain `await asyncio.to_thread(...)`, on `assert not task.done()`. A fourth pins that a worker failure arriving after cancellation still reports as cancelled.

Two behavior guards round it out: a persist-then-hydrate round trip, and a missing workspace root still raising `WorkspaceArchiveReadError`.

Because this module refuses to import on Windows by design, verification ran on Linux: `tests/sandbox` is 1435 passed, 22 skipped, and `mypy` is clean on the module.

Resolves #4675.

### DIFF
--- a/src/agents/sandbox/sandboxes/unix_local.py
+++ b/src/agents/sandbox/sandboxes/unix_local.py
@@ -116,6 +116,65 @@ def _restore_pty_child_signal_defaults() -> None:
         signal.signal(signum, signal.SIG_DFL)
 
 
+# The helpers below do the blocking filesystem work for this backend. Each one is called
+# through ``asyncio.to_thread`` so the archive, extraction, and path syscalls stay off the
+# event loop; keeping them at module level keeps that boundary obvious and keeps session
+# state access on the loop side of it.
+
+
+def _resolve_existing_workspace(workspace: Path) -> Path:
+    """Confirm the workspace root exists and resolve it."""
+    if not workspace.exists():
+        raise WorkspaceRootNotFoundError(path=workspace)
+    return workspace.resolve()
+
+
+def _archive_workspace(root: Path, skip: set[Path]) -> io.BytesIO:
+    """Tar the whole workspace tree into an in-memory archive."""
+    if not root.exists():
+        raise WorkspaceArchiveReadError(path=root, context={"reason": "workspace_root_not_found"})
+
+    buf = io.BytesIO()
+    try:
+        with tarfile.open(fileobj=buf, mode="w") as tar:
+            tar.add(
+                root,
+                arcname=".",
+                filter=lambda ti: (
+                    None
+                    if should_skip_tar_member(
+                        ti.name,
+                        skip_rel_paths=skip,
+                        root_name=None,
+                    )
+                    else ti
+                ),
+            )
+    except (tarfile.TarError, OSError) as e:
+        raise WorkspaceArchiveReadError(path=root, cause=e) from e
+
+    buf.seek(0)
+    return buf
+
+
+def _extract_workspace_archive(root: Path, data: io.IOBase) -> None:
+    """Extract an archive over the workspace tree."""
+    try:
+        root.mkdir(parents=True, exist_ok=True)
+        with tarfile.open(fileobj=data, mode="r:*") as tar:
+            safe_extract_tarfile(
+                tar,
+                root=root,
+                allow_external_symlink_targets=False,
+            )
+    except UnsafeTarMemberError as e:
+        raise WorkspaceArchiveWriteError(
+            path=root, context={"reason": e.reason, "member": e.member}, cause=e
+        ) from e
+    except (tarfile.TarError, OSError) as e:
+        raise WorkspaceArchiveWriteError(path=root, cause=e) from e
+
+
 class UnixLocalSandboxSessionState(SandboxSessionState):
     type: Literal["unix_local"] = "unix_local"
     workspace_root_owned: bool = False
@@ -196,7 +255,7 @@ class UnixLocalSandboxSession(BaseSandboxSession):
     async def _prepare_backend_workspace(self) -> None:
         workspace = Path(self.state.manifest.root)
         try:
-            workspace.mkdir(parents=True, exist_ok=True)
+            await asyncio.to_thread(workspace.mkdir, parents=True, exist_ok=True)
         except OSError as e:
             raise WorkspaceStartError(path=workspace, cause=e) from e
 
@@ -265,8 +324,7 @@ class UnixLocalSandboxSession(BaseSandboxSession):
     async def _exec_internal(
         self, *command: str | Path, timeout: float | None = None
     ) -> ExecResult:
-        env, cwd = await self._resolved_exec_context()
-        workspace_root = Path(cwd).resolve()
+        env, cwd, workspace_root = await self._resolved_exec_context()
         command_parts = self._workspace_relative_command_parts(command, workspace_root)
         process_cwd, command_parts = self._shell_workspace_process_context(
             command_parts=command_parts,
@@ -318,8 +376,7 @@ class UnixLocalSandboxSession(BaseSandboxSession):
         max_output_tokens: int | None = None,
     ) -> PtyExecUpdate:
         _ = timeout
-        env, cwd = await self._resolved_exec_context()
-        workspace_root = Path(cwd).resolve()
+        env, cwd, workspace_root = await self._resolved_exec_context()
         sanitized_command = self._prepare_exec_command(*command, shell=shell, user=user)
         command_parts = self._workspace_relative_command_parts(sanitized_command, workspace_root)
         process_cwd, command_parts = self._shell_workspace_process_context(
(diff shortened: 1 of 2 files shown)


rajarshidattapy:
Thanks — both points taken, and the PR is updated in place.

**Scope narrowed.** The bounded `mkdir`, `exists`, and `resolve` sites are reverted to their original form; only `persist_workspace` and `hydrate_workspace` change now. You are right that an `ASYNC240` report alone did not establish a defect for them, and I have edited #4675 to say so rather than lump them in with the archive paths. Reverting also keeps this change off the exec confinement path, which is a better place for it not to be.

**Cancellation ownership.** This was a real hole in what I first pushed, not just a review preference — `restore_snapshot_into_workspace_on_resume` closes the archive stream in a `finally` as soon as the await returns, so a cancelled `hydrate_workspace` left a live worker reading a closed stream and writing into a workspace that resume may then clear.

`_run_workspace_io` now records the cancellation, waits for the worker to settle, and only then propagates it; repeated cancellation is absorbed the same way, and a worker failure arriving after cancellation is consumed so the caller still hears cancellation. It follows `memory.sqlite_session._await_mutation`. That does mean a second copy of that pattern in the tree — happy to hoist it into a shared util instead if you would prefer one home for it.

**Coverage** is deterministic for both properties. Responsiveness: two tests hold the helper open until a coroutine scheduled beside it has run. Cancellation: three tests hold the worker open, cancel, and assert the task has neither completed nor let the worker settle before it is released. All three fail against a plain `await asyncio.to_thread(...)` on `assert not task.done()`, so they pin the behavior rather than just exercising it.

Verified on Linux, since the module does not import on Windows: `tests/sandbox` is 1435 passed, 22 skipped, `mypy` clean.

seratch:
Thanks for putting this together. The underlying #4675 issue is real, but I do not think this PR is a safe basis for the fix. `asyncio.to_thread()` does not stop its worker when the awaiting task is cancelled. In `hydrate_workspace()`, snapshot cleanup can close the archive and proceed into session shutdown or deletion while the worker is still extracting into the same workspace. The tests do not cover this cancellation and cleanup ordering.

The scope is also driven by `ASYNC240` rather than the lifecycle boundary: it moves bounded `mkdir` and `resolve` calls but leaves other unbounded synchronous filesystem work, including recursive removal and streamed writes, on the event loop. Fixing this correctly requires a cancellation-safe blocking-I/O design rather than extending this patch. I am going to close #4678 while keeping #4675 open for that redesign.

PR #4682, opened 2026-08-26
fix(tracing): close the exporter when the batch processor shuts down

### LINKED ISSUE #4681: Tracing: the default `BackendSpanExporter`'s HTTP client is never closed
**Type:** bug / resource leak
**Severity:** low-medium

## Description

process lifetime:

```python
```

```python
def close(self):
    self._client.close()
```

## Repro

```python

processor.shutdown()
```

## Impact

  ResourceWarnings as errors).

## Suggested fix

```python
if callable(close):
    close()
```

rajarshidattapy:
Fixes #4681

## Problem

`BackendSpanExporter` opens a pooled `httpx2.Client` in its constructor and holds it for
the life of the process. It exposes a `close()`, but nothing in the shutdown chain called
it:

`_request_shutdown()`, and joins the worker thread — the exporter itself is never closed.
Before this change:

```python
processor.shutdown()
exporter._client.is_closed  # False
```

So the connection to the tracing endpoint is left for interpreter teardown to reap, which
surfaces as `ResourceWarning: unclosed <ssl.SSLSocket ...>` under `-W error::ResourceWarning`
or `PYTHONDEVMODE=1`, and applications that swap the global provider/processor at runtime
leak one client and its connection pool per replacement.

## Fix

`BatchTraceProcessor.shutdown()` now closes its exporter:

- **at the end**, after the join or the synchronous drain, so the final flush still has a
  usable client;
- **duck-typed** (`getattr(self._exporter, "close", None)`), mirroring the existing
  `_request_shutdown` lookup, because `TracingExporter` does not require `close()` and
  custom exporters may not have one;
- **best-effort** — a raising `close()` is logged as non-fatal rather than propagated out
  of application shutdown.

No public API changes.

## Tests

Three regression tests in `tests/test_trace_processor.py`, all failing before the change
and passing after:

- `test_batch_trace_processor_shutdown_closes_the_exporter` — pins that `close()` is called
  and that it happens *after* the final export.
- `test_batch_trace_processor_shutdown_closes_the_backend_exporter_client` — the reported
  leak: the default exporter's HTTP client is closed once the processor shuts down.
- `test_batch_trace_processor_shutdown_survives_exporter_close_failure` — a failing
  `close()` is logged, not raised.

`tests/test_trace_processor.py`, `tests/tracing/`, `tests/test_tracing.py` and
`tests/test_tracing_errors.py` pass (`tests/tracing/test_import_side_effects.py::test_core_imports_do_not_require_legacy_httpx`
fails identically on `main` in my environment). `ruff check`, `ruff format --check` and
`mypy` are clean on both changed files.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -644,6 +644,19 @@ class BatchTraceProcessor(TracingProcessor):
             # No background thread: process any remaining items synchronously.
             self._export_batches(deadline=deadline)
 
+        # Release whatever the exporter holds open -- the default one keeps a pooled HTTP
+        # client alive for the life of the process. Done last so the flush above still has
+        # it, and duck-typed like ``_request_shutdown`` because the interface does not
+        # require it.
+        close_exporter = getattr(self._exporter, "close", None)
+        if callable(close_exporter):
+            try:
+                close_exporter()
+            except Exception as exc:
+                log_model_and_tool_action_error(
+                    logger, "[non-fatal] Tracing: exporter close failed", exc
+                )
+
     def force_flush(self):
         """
         Forces an immediate flush of all queued spans.
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -580,7 +580,7 @@ class BatchTraceProcessor(TracingProcessor):
         self._thread_start_lock = threading.Lock()
         self._export_lock = threading.Lock()
         self._shutdown_deadline: float | None = None
-        self._close_lock = threading.Lock()
+        # Guarded by _export_lock so closing cannot overlap an export, in either order.
         self._exporter_closed = False
 
     def _ensure_thread_started(self) -> None:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for working through the shutdown races. I do not think BatchTraceProcessor can safely infer ownership of an arbitrary exporter it was handed.

TracingExporter does not require hashability, weak-reference support, or a close() ownership contract. The current fallback treats an unhashable or non-weak-referenceable exporter as solely owned, so two processors sharing such an exporter can still close it while the other processor is active. More generally, shared-resource ownership cannot be reconstructed from object identity here.

The default backend exporter's HTTP client leak should be fixed at the provider/setup boundary that creates and owns that exporter. Closing injected exporters would require an explicit ownership contract. I am closing this PR, while #4681 can remain open for that narrower fix.

rajarshidattapy:
got it.

PR #4684, opened 2026-08-26
fix(tracing): clear the exporter's shutdown request when a processor attaches

### LINKED ISSUE #4683: `BackendSpanExporter._shutdown_event` is never cleared, permanently disabling retries after one shutdown
**Type:** bug
**Severity:** low

## Description

shutdown:

```python
    self._shutdown_event.set()
```

```python
    if deadline is None:
            return False
    ...
```

## Repro

```python

    response = MagicMock()
    response.status_code = 504

                                   max_delay=0.02)

    first.shutdown(timeout=1.0)

    second.force_flush()

```

```
```

## Impact

## Suggested fix

rajarshidattapy:
Fixes #4683

## Problem

`BackendSpanExporter._shutdown_event` is one-way. `BatchTraceProcessor.shutdown(timeout=...)`
sets it through `_request_shutdown()` so an in-flight export abandons its retry backoff
instead of sleeping past the shutdown deadline — but nothing ever clears it, and
`default_exporter()` caches a module-level singleton, so the exporter outlives the processor
that shut it down.

Every export through that exporter afterwards gives up on the first 5xx or network error
rather than retrying:

```python
exporter = BackendSpanExporter(api_key="test_key", max_retries=3, ...)  # 504s from the mock client

first.shutdown(timeout=1.0)

second = BatchTraceProcessor(exporter=exporter)   # same singleton exporter
second.force_flush()
```

```
```

The batch is dropped, and the warning blames a shutdown that finished long ago.

## Fix

Make ownership of the request explicit: the processor that requests the exporter's shutdown
releases it, and only once its own worker has stopped.

- `BackendSpanExporter` counts outstanding requests: `_request_shutdown()` increments and
  sets the event, `_reset_shutdown()` decrements and clears it only when the count reaches
  zero, both under a lock. Processors share an exporter, so one processor's release must not
  speak for another's still-running worker.
- `BatchTraceProcessor` requests at most once — the request and the release that balances it
  are both taken under a per-processor lock, since `shutdown` may be called from several
  threads — and releases it once its worker is done: at the end of `_run`, or in `shutdown()`
  itself whenever the worker is not running by the time it returns. Both lookups stay
  duck-typed like the existing `_request_shutdown` call, since `TracingExporter` requires
  neither.
- The request is now made *before* the processor sets its own shutdown event, so a worker
  that exits immediately always sees that the request is owned.

Releasing on worker exit rather than when the next processor attaches is what keeps a
timed-out shutdown safe: a worker that outlived its `shutdown(timeout=...)` still owns the
cancellation, keeps abandoning its retries, and hands the exporter back only when it stops —
so a replacement processor can never resurrect the backoff of a worker that is supposed to be
going away. The existing
`test_batch_trace_processor_shutdown_interrupts_exporter_retry_backoff` continues to cover
the abort itself.

No public API changes.

## Tests

Two regression tests in `tests/test_trace_processor.py`, both failing on `main` and passing
with this change:

- `test_new_processor_restores_exporter_retries_after_a_previous_shutdown` — shuts a
  processor down with a timeout, attaches a new one to the same exporter, and asserts the
  next 504 batch is retried `max_retries` times instead of once (1 → 3).
- `test_worker_that_outlives_shutdown_keeps_the_exporter_cancelled` — pins the ownership
  rule: with the worker stuck inside a request, `shutdown(timeout=0.05)` returns while it is
  still running, a replacement processor attaches, and the abandoned worker must still find
  the exporter cancelled, give up without retrying, and release the request as it exits.
- `test_concurrent_shutdowns_leave_the_exporter_request_balanced` — pins the serialization:
  two threads shut the same processor down and rendezvous inside the window between the
  request check and the request itself, so an unsynchronized decision makes two requests
  against the single release the worker performs; the exporter's request count must be back
  to zero afterwards.
- `test_shared_exporter_stays_cancelled_until_every_shutdown_releases` — pins the counting:
  two processors share an exporter, the first times out with its worker inside a request and
  the second shuts down cleanly; the shared cancellation must survive the second processor's
  release and lift only once the blocked worker has stopped too.

`tests/test_trace_processor.py`, `tests/tracing/`, `tests/test_tracing.py` and
`tests/test_tracing_errors.py` pass (`tests/tracing/test_import_side_effects.py::test_core_imports_do_not_require_legacy_httpx`
fails identically on `main` in my environment). `ruff check`, `ruff format --check` and
`mypy` are clean on both changed files.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -567,15 +567,6 @@ class BatchTraceProcessor(TracingProcessor):
             export_trigger_ratio: The ratio of the queue size at which we will trigger an export.
         """
         self._exporter = exporter
-        # A processor that shut down earlier may have asked this exporter to abandon retry
-        # backoff. That request is one-way, and `default_exporter()` hands the same instance
-        # to every processor, so clear it as this one attaches: otherwise the first transient
-        # failure exported here gives up without retrying, blaming a shutdown that is over.
-        # Duck-typed like the `_request_shutdown` call in `shutdown()`, since the exporter
-        # interface does not require either.
-        reset_exporter_shutdown = getattr(exporter, "_reset_shutdown", None)
-        if callable(reset_exporter_shutdown):
-            reset_exporter_shutdown()
         self._queue: queue.Queue[Trace | Span[Any]] = queue.Queue(maxsize=max_queue_size)
         self._max_queue_size = max_queue_size
         self._max_batch_size = max_batch_size
@@ -593,6 +584,7 @@ class BatchTraceProcessor(TracingProcessor):
         self._thread_start_lock = threading.Lock()
         self._export_lock = threading.Lock()
         self._shutdown_deadline: float | None = None
+        self._requested_exporter_shutdown = False
 
     def _ensure_thread_started(self) -> None:
         # Fast path without holding the lock
@@ -637,11 +629,15 @@ class BatchTraceProcessor(TracingProcessor):
         """
         Called when the application stops. We signal our thread to stop, then join it.
         """
-        self._shutdown_event.set()
         if timeout is not None:
             request_exporter_shutdown = getattr(self._exporter, "_request_shutdown", None)
             if callable(request_exporter_shutdown):
+                # Flagged before the request so the worker, which may exit as soon as the
+                # event below is set, always sees that the request is ours to release.
+                self._requested_exporter_shutdown = True
                 request_exporter_shutdown()
+
+        self._shutdown_event.set()
 
         deadline = None if timeout is None else time.monotonic() + timeout
         self._shutdown_deadline = deadline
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the careful concurrency work. I do not think we should merge this lifecycle contract as-is.

The reproducer establishes that exporter reuse after `BatchTraceProcessor.shutdown(timeout=...)` is constructible, but that post-shutdown reuse is not currently an established supported lifecycle. #4712 is simultaneously taking the opposite ownership direction by closing the default exporter on shutdown. We should not land both contracts independently.

My preference is one explicit terminal ownership model: the owner closes the default exporter, and a later tracing initialization creates a fresh exporter and processor. That avoids adding a ref-counted cross-processor cancellation protocol for a reuse path we have not committed to support. I suggest closing this PR for now while keeping the underlying lifecycle question open.

PR #4689, opened 2026-08-26
fix(sessions): checkpoint a resumed handoff before the Session append

### LINKED ISSUE #4685: Resumed approval followed by handoff can leave Session history without tool outputs after append failure
### Describe the bug

When a resumed `RunState` resolves an approved local function tool and a queued handoff from the same model response, the handoff target is committed before the resumed turn is persisted. If the client-managed `Session.add_items()` call then fails atomically before storing the batch, the error propagates but no `pending_session_write` is retained.

Retrying the same live or JSON-restored state correctly continues with the target agent and does not re-execute the tool or handoff. However, it also never repairs the missing function-tool and handoff outputs. The target agent's final message is then appended to a Session that still contains two orphaned `function_call` items.

This creates two durable conversation views:

- the `Session` contains the `charge` and handoff calls, but neither output;
- `RunState` and `RunResult.to_input_list()` contain both complete call/output pairs.

The behavior is provider-neutral and deterministic with `ScriptedModel`; no API key or network call is involved. It reproduces on `main` at `a40ae9803e6b7a79faa246293f56adb100d5868b`.

### Reproduction outline

1. Configure a triage agent with:
   - an approval-gated `charge` function tool;
   - a handoff to a delegate agent.
2. Have one scripted model response emit both `charge(call_id="charge-1")` and `transfer_to_delegate(call_id="handoff-1")`.
3. Run until the approval interruption, approve `charge`, and configure the Session to raise before storing its next atomic batch.
4. Resume and observe the original Session error.
5. Retry the same live state or a `RunState.to_json()` / `RunState.from_json()` round trip.
6. Compare the Session with `RunResult.to_input_list()`.

After the failed resume:

```text
tool effects:       1
model calls:        1
target agent:       delegate
pending write:      None
Session charge:     [function_call]
Session handoff:    [function_call]
```

After retry:

```text
final output:       done
tool effects:       1
model calls:        2
last agent:         delegate
Session charge:     [function_call]
Session handoff:    [function_call]
replay charge:      [function_call, function_call_output]
replay handoff:     [function_call, function_call_output]
```

### Sync / streaming matrix

The matrix varies the failing runner, retry runner, live versus JSON-restored state, and whether the Session raises before commit or after committing but before acknowledging the append.

| Pause + failing resume | Retry | State | Atomic fail before commit | Commit then raise |
|---|---|---|---:|---:|
| `Runner.run()` | `Runner.run()` | live + JSON | 2/2 diverge | 0/2 diverge |
| `Runner.run()` | `Runner.run_streamed()` | live + JSON | 2/2 diverge | 0/2 diverge |
| `Runner.run_streamed()` | `Runner.run()` | live + JSON | 2/2 diverge | 0/2 diverge |
| `Runner.run_streamed()` | `Runner.run_streamed()` | live + JSON | 2/2 diverge | 0/2 diverge |

All 16 rows preserve the tool side effect exactly once and continue with the delegate. All eight atomic-failure rows leave the Session/replay split. The eight commit-then-raise rows already contain the batch in the Session, but the SDK has no pending checkpoint that explicitly reconciles that outcome.

### Root cause

The resumed handoff path crosses three ownership boundaries in the wrong durable order:

1. [`_commit_handoff_output()` records the target agent](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_internal/turn_resolution.py#L2620-L2653).
2. The non-streaming runner [copies `NextStepHandoff` and the completed items into `RunState`](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run.py#L1117-L1146), then [awaits the fallible Session write](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run.py#L1148-L1176). Its run-wide guardrail lists are extended only [after that write succeeds](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run.py#L1232-L1240).
3. The streaming runner has the same ordering: it [updates state and accumulates results](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_internal/run_loop.py#L1327-L1377), then [persists before converting the transition to `NextStepRunAgain`](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_internal/run_loop.py#L1407-L1429).

`save_resumed_turn_items()` only supplies `resumed_write_state` for [`NextStepRunAgain` or `NextStepInterruption`](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_internal/session_persistence.py#L717-L747), so a handoff append bypasses the existing pending-write checkpoint created at [`save_result_to_session()`](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_internal/session_persistence.py#L654-L667).

Persisting `NextStepHandoff` directly would widen the durable schema. `RunState` currently declares only [`NextStepInterruption | NextStepRunAgain | None`](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_state.py#L846-L874), serializes only those resumable states ([serialization](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_state.py#L1977-L2011)), and validates pending writes only with those steps ([deserialization](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_state.py#L4302-L4383)).

### Minimal design direction

One narrow design is to treat the handoff as committed once `_commit_handoff_output()` has recorded the target agent, then normalize the remaining resumable transition to the existing `NextStepRunAgain` checkpoint before the fallible Session append. That lets the current v1.17 `pending_session_write` mechanism carry the exact function-tool and handoff output batch without adding a new persisted step type.

At the same boundary, publish the completed tool input/output guardrail results to `RunState` before the append can raise. A retry can then use the existing pre-model reconciliation calls in the [non-streaming runner](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run.py#L625-L639) and [streaming runner](https://github.com/openai/openai-agents-python/blob/a40ae9803e6b7a79faa246293f56adb100d5868b/src/agents/run_internal/run_loop.py#L925-L929).

A no-schema prototype of that shape passes all 16 rows above. A separate JSON cross-mode, commit-then-raise probe also preserves one input guardrail result, one output guardrail result, one tool-start/tool-end hook pair, one tool side effect, and two total model calls.

I would like maintainer confirmation that `NextStepRunAgain` is the intended durable post-handoff checkpoint before preparing an implementation.

### Relationship to existing work

- #4630 introduced recoverable resumed Session appends for `NextStepRunAgain` and explicitly left handoff/terminal recovery and distributed exactly-once execution outside its scope. In [the review discussion](https://github.com/openai/openai-agents-python/pull/4630#discussion_r3847952752), the missing handoff outputs were confirmed as a real pre-existing persistence gap that should be tracked separately; the same comment notes that merely adding `NextStepHandoff` to the existing type gate is insufficient.
- #4650 extended the same checkpoint to renewed interruptions. This issue proposes completing the model-continuation side for handoffs, not changing the reconciliation algorithm.
- #2695 owns approval-time argument overrides and currently overlaps `run.py`, `run_loop.py`, `session_persistence.py`, `run_state.py`, and related tests. It does not address this failure mode, but any implementation here should coordinate with that maintainer-owned branch rather than introduce competing durable-state machinery.

### Acceptance criteria

For the original client-managed Session/backend and serialized access to its history:

- The first Session exception still propagates unchanged.
- The approved tool side effect, its input/output guardrails, tool start/end hooks, and the handoff callback each execute exactly once.
- The target agent is committed once and remains the current agent across live and JSON-restored recovery.
- The failed state retains the exact pending Session batch and completed guardrail results before any later model call.
- Retry reconciles both atomic-failure and commit-then-raise outcomes without duplicating either output.
- The Session, `RunResult.to_input_list()`, restored `RunState`, and next model input each contain exactly one ordered call/output pair for `charge-1` and `handoff-1`.
- The complete sync/stream, same-mode/cross-mode, live/JSON, fail-before/commit-then-raise matrix passes.
- A missing or different Session, changed tail, partial append, ambiguous tail, or concurrent resume fails closed before another model call or local side effect.
- If the implementation reuses `NextStepRunAgain`, no RunState schema change is introduced. If a different durable representation is chosen, its versioning and backward-read behavior are explicit and regression-tested.

### Non-goals

- Terminal `NextStepFinalOutput` recovery, including `tool_use_behavior="stop_on_first_tool"`; that needs a separate finalization/output-guardrail contract.
- Server-managed conversation recovery.
- Changing approval override behavior from #2695.
- Session compaction read-window or concurrent-replacement behavior.
- Repairing arbitrary partial writes or searching history for similar items.
- A distributed transaction or global exactly-once guarantee.

### Exactly-once boundary

The SDK can avoid re-executing already completed local work while reconciling one known pending batch when the caller supplies the same Session ID/backend, the append is atomic or its committed tail is detectable, and access to that history is serialized.

The Session protocol has no transaction ID, compare-and-swap operation, distributed lock, or idempotency key. Independent restored `RunState` copies can race, and a non-atomic partial append can be ambiguous. Those cases should fail closed before another model call; this issue does not claim distributed exactly-once tool effects or Session writes.

rajarshidattapy:
Fixes #4685.

### Problem

When a resumed `RunState` resolves an approved local function tool and a queued handoff from the same model response, the handoff target is committed before the resumed turn is persisted. If the client-managed `Session.add_items()` call then fails, the error propagates but no `pending_session_write` is retained.

Retrying the same live or JSON-restored state correctly continues with the target agent and does not re-execute the tool or the handoff, but it never repairs the missing outputs either. The result is two durable conversation views that never converge:

- the `Session` holds the `charge` and handoff `function_call` items with neither output;
- `RunState` and `RunResult.to_input_list()` hold both complete call/output pairs.

### Cause

`update_run_state_after_resume()` stores the raw transition, so `_current_step` becomes `NextStepHandoff`. `save_resumed_turn_items()` only supplies `resumed_write_state` for `NextStepRunAgain` or `NextStepInterruption`, so the handoff append bypasses the pending-write checkpoint introduced in #4630 and extended in #4650. `NextStepHandoff` is also outside the `NextStepInterruption | NextStepRunAgain | None` type `RunState` declares and serializes, so the failed state carries no resumable step at all.

The same boundary drops the turn's tool input/output guardrail results: both runners publish them to `RunState` only for `NextStepInterruption`.

### Fix

Treat the handoff as committed once the target agent is recorded, then normalize the remaining resumable transition to the existing `NextStepRunAgain` checkpoint before the fallible append. The current `pending_session_write` mechanism then carries the exact function-tool and handoff output batch, and `resume_pending_session_write()` settles it on the next resume — before any further model call or local side effect. No new persisted step type and no `RunState` schema change.

At the same boundary, both runners now publish the completed tool guardrail results to `RunState` before the append can raise, so a retry reports them through the existing pre-model reconciliation instead of losing them.

The streaming loop already set `NextStepRunAgain` after its handoff append; that assignment is now redundant and removed, since the shared helper normalizes it before the append.

### Behavior

- The first Session exception still propagates unchanged.
- The approved tool side effect, its guardrails, its hooks, and the handoff callback each run exactly once.
- The target agent is committed once and stays current across live and JSON-restored recovery.
- Retry reconciles both the atomic-failure and the commit-then-raise outcome without duplicating either output.
- Session, `RunResult.to_input_list()`, the restored `RunState`, and the next model input each contain exactly one ordered call/output pair for `charge-1` and `handoff-1`.
- A missing, different, or changed Session still fails closed through the existing `resume_pending_session_write()` checks.

### Tests

`test_resumed_handoff_session_append_is_recovered_before_next_model` covers the full sync/stream, same-mode/cross-mode, live/JSON, fail-before-commit/commit-then-raise matrix (16 rows). It fails on `main` and passes here.

### DIFF
--- a/src/agents/run.py
+++ b/src/agents/run.py
@@ -1135,7 +1135,12 @@ class AgentRunner:
                                     generated_items=generated_items,
                                     session_items=session_items,
                                 )
-                                if isinstance(turn_result.next_step, NextStepInterruption):
+                                if isinstance(
+                                    turn_result.next_step,
+                                    NextStepInterruption | NextStepHandoff,
+                                ):
+                                    # Publish before the fallible append so a retry does not
+                                    # lose guardrail results for work that already ran.
                                     run_state._tool_input_guardrail_results = [
                                         *tool_input_guardrail_results,
                                         *turn_result.tool_input_guardrail_results,
--- a/src/agents/run_internal/run_loop.py
+++ b/src/agents/run_internal/run_loop.py
@@ -1405,6 +1405,15 @@ async def start_streaming(
                         break
 
                     if isinstance(turn_result.next_step, NextStepHandoff):
+                        if run_state is not None:
+                            # `_accumulate_tool_guardrail_results` already folded this turn's
+                            # results in; publish them before the fallible append.
+                            run_state._tool_input_guardrail_results = list(
+                                accepted_tool_input_guardrail_results
+                            )
+                            run_state._tool_output_guardrail_results = list(
+                                accepted_tool_output_guardrail_results
+                            )
                         await _save_resumed_items(
                             list(turn_session_items),
                             turn_result.model_response.response_id,
@@ -1421,7 +1430,6 @@ async def start_streaming(
                         streamed_result._event_queue.put_nowait(
                             AgentUpdatedStreamEvent(new_agent=current_agent)
                         )
-                        run_state._current_step = NextStepRunAgain()
                         if await _wait_for_streamed_turn_events_and_stop_if_cancelled(
                             streamed_result
                         ):
--- a/src/agents/run_internal/session_persistence.py
+++ b/src/agents/run_internal/session_persistence.py
@@ -61,7 +61,13 @@ from .items import (
     strip_internal_input_item_metadata,
 )
 from .oai_conversation import OpenAIServerConversationTracker
-from .run_steps import NextStepInterruption, NextStepRunAgain, ProcessedResponse, SingleStepResult
+from .run_steps import (
+    NextStepHandoff,
+    NextStepInterruption,
+    NextStepRunAgain,
+    ProcessedResponse,
+    SingleStepResult,
+)
 
 __all__ = [
     "admit_pending_input",
@@ -541,7 +547,13 @@ def update_run_state_after_resume(
     run_state._generated_items = generated_items
     if session_items is not None:
         run_state._session_items = list(session_items)
-    run_state._current_step = turn_result.next_step  # type: ignore[assignment]
+    next_step = turn_result.next_step
+    if isinstance(next_step, NextStepHandoff):
+        # The target agent is already committed, so the rest of the turn is an ordinary
+        # "run again". Normalizing here, before the fallible Session append, lets the existing
+        # pending-write checkpoint carry the handoff batch without a new persisted step type.
+        next_step = NextStepRunAgain()
+    run_state._current_step = next_step  # type: ignore[assignment]
 
 
 async def save_result_to_session(
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -567,15 +567,6 @@ class BatchTraceProcessor(TracingProcessor):
             export_trigger_ratio: The ratio of the queue size at which we will trigger an export.
         """
         self._exporter = exporter
-        # A processor that shut down earlier may have asked this exporter to abandon retry
-        # backoff. That request is one-way, and `default_exporter()` hands the same instance
-        # to every processor, so clear it as this one attaches: otherwise the first transient
-        # failure exported here gives up without retrying, blaming a shutdown that is over.
-        # Duck-typed like the `_request_shutdown` call in `shutdown()`, since the exporter
-        # interface does not require either.
-        reset_exporter_shutdown = getattr(exporter, "_reset_shutdown", None)
-        if callable(reset_exporter_shutdown):
-            reset_exporter_shutdown()
         self._queue: queue.Queue[Trace | Span[Any]] = queue.Queue(maxsize=max_queue_size)
         self._max_queue_size = max_queue_size
         self._max_batch_size = max_batch_size
@@ -593,6 +584,7 @@ class BatchTraceProcessor(TracingProcessor):
         self._thread_start_lock = threading.Lock()
         self._export_lock = threading.Lock()
         self._shutdown_deadline: float | None = None
+        self._requested_exporter_shutdown = False
 
     def _ensure_thread_started(self) -> None:
         # Fast path without holding the lock
@@ -637,11 +629,15 @@ class BatchTraceProcessor(TracingProcessor):
         """
         Called when the application stops. We signal our thread to stop, then join it.
         """
-        self._shutdown_event.set()
         if timeout is not None:
             request_exporter_shutdown = getattr(self._exporter, "_request_shutdown", None)
             if callable(request_exporter_shutdown):
+                # Flagged before the request so the worker, which may exit as soon as the
+                # event below is set, always sees that the request is ours to release.
+                self._requested_exporter_shutdown = True
                 request_exporter_shutdown()
+
+        self._shutdown_event.set()
 
         deadline = None if timeout is None else time.monotonic() + timeout
         self._shutdown_deadline = deadline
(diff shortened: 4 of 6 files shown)


FU-max-boop (commented):
Thanks for taking on #4685. I reviewed the exact head `8f2985a9379b3a91ca024b8a62a6788c7d6c4a94` against its stated handoff scope and found two things to resolve before this is ready for behavioral review.

1. The branch currently includes the four tracing-exporter commits from #4684 (`4fb0dd57`, `8760e90e`, `560d2b1b`, and `73879e3d`) and therefore adds `src/agents/tracing/processors.py` plus 211 lines in `tests/test_trace_processor.py`. Those changes are unrelated to #4685 and are not disclosed in this PR body. The handoff commit alone (`73879e3d..8f2985a9`) applies cleanly to the stated base `a40ae980`, so please either rebuild this PR from that single commit or make the stacked dependency/base explicit. As currently shaped, merging this PR would also merge #4684.

2. The new 16-row resume matrix defines no tool input/output guardrails and explicitly asserts zero tool guardrail results. However, the runtime change also moves publication of `_tool_input_guardrail_results` and `_tool_output_guardrail_results` before the fallible Session append. That claimed durability behavior is therefore untested. Please add tool guardrails to at least one representative fail-before-commit row and assert that the completed results survive the failed live state, JSON round-trip, and cross-mode retry without re-running the guardrails. That would directly protect the new ordering rather than only the Session batch repair.

The core normalization of the committed `NextStepHandoff` to the existing `NextStepRunAgain` checkpoint is consistent with the narrow #4685 contract; this review is about isolating that change and covering the additional guardrail state it publishes.

seratch (changes requested):
Please remove unrelated commits

seratch:
@rajarshidattapy Can you remove the tracing changes from this PR?

seratch:
Resolved by https://github.com/openai/openai-agents-python/pull/4725

PR #4691, opened 2026-08-26
Keep trimmed tool outputs inside max_output_chars

### LINKED ISSUE #4686: ToolOutputTrimmer can emit a "trimmed" output many times larger than max_output_chars
### Describe the bug

`ToolOutputTrimmer` validates `max_output_chars` and `preview_chars` independently, but never checks that `preview_chars <= max_output_chars`:

https://github.com/openai/openai-agents-python/blob/main/src/agents/extensions/tool_output_trimmer.py#L117-L123

Two of the three trimming paths then slice `preview_chars` characters and only verify the result is smaller than the *original*, not that it fits the configured budget:

- `_trim_function_call_output` (plain-string outputs), `src/agents/extensions/tool_output_trimmer.py:260-266`
- `_trim_legacy_tool_search_results`, `src/agents/extensions/tool_output_trimmer.py:424-430`

```python
preview = output_str[: self.preview_chars]
summary = (
    f"[Trimmed: {display_name} output — {output_len} chars → "
    f"{self.preview_chars} char preview]\n{preview}..."
)
if len(summary) >= output_len:
    return None, 0
```

The structured path does the opposite — `_trim_structured_function_call_output` (lines 298-299) carefully picks a header and preview length so the replacement fits inside `max_output_chars`:

```python
preview_budget = self.max_output_chars - len(minimal_header) - 1
preview_len = min(len(text_content), self.preview_chars, max(0, preview_budget))
```

So the three paths disagree about whether `max_output_chars` is a budget or just a trigger threshold.

There is a second, cosmetic defect in the same f-string: the header always claims `{self.preview_chars} char preview`, even when the actual preview is shorter than `preview_chars` (whenever `preview_chars > output_len`).

### Debug information

- Agents SDK version: `main` (`src/agents/extensions/tool_output_trimmer.py`, unchanged since #4204)
- Python: any

### Repro steps

```python
from agents.extensions.tool_output_trimmer import ToolOutputTrimmer

t = ToolOutputTrimmer(max_output_chars=100, preview_chars=5000)   # accepted without complaint
item = {"type": "function_call_output", "call_id": "c1", "output": "x" * 6000}
trimmed, _ = t._trim_function_call_output(item, ("my_tool",))
len(trimmed["output"])   # 5062 — 50x the configured 100-char budget
```

### Expected behavior

A trimmed output should never exceed `max_output_chars`. Either:

1. Reject `preview_chars > max_output_chars` in `__post_init__`, or
2. Clamp the two string paths the way the structured path already does:

```python
preview_len = min(len(output_str), self.preview_chars, max(0, self.max_output_chars - len(header)))
```

and report `preview_len` in the header instead of `self.preview_chars`.

Option 1 is the smaller fix and makes the invariant explicit at the trust boundary; option 2 keeps existing configs working. Happy to send a PR for whichever you prefer.

rajarshidattapy:
Fixes #4686.

### Problem

`ToolOutputTrimmer` treated `max_output_chars` as a budget in one path and as a mere trigger threshold in the other two. `_trim_function_call_output` (plain-string outputs) and `_trim_legacy_tool_search_results` sliced `preview_chars` characters and then only checked the result against the *original* length, never against the configured budget:

```python
t = ToolOutputTrimmer(max_output_chars=100, preview_chars=5000)
item = {"type": "function_call_output", "call_id": "c1", "output": "x" * 6000}
trimmed, _ = t._trim_function_call_output(item, ("my_tool",))
len(trimmed["output"])   # 5062 — 50x the configured 100-char budget
```

`_trim_structured_function_call_output` already did the right thing, sizing the preview so the replacement fits inside `max_output_chars`.

The same f-string also always claimed `{preview_chars} char preview` in the header, even when the real preview was shorter.

### Fix

This takes option 2 from the issue: clamp rather than reject, so existing configs keep working.

Both free-text paths now go through one `_fit_preview_summary()` helper that shrinks the preview to whatever the header leaves inside `max_output_chars` and reports the actual preview length in the header. If not even the header fits, it returns `None` and the output is left untouched — no replacement would be an improvement at that point.

Option 1 (rejecting `preview_chars > max_output_chars` in `__post_init__`) would not have been sufficient on its own: the header is ~50 characters, so `max_output_chars=100, preview_chars=100` would still overflow the budget.

The header length depends on the preview length it reports, so the helper re-fits after shrinking; the loop terminates because the preview length strictly decreases on every retry.

### Behavior

- Default config (`max_output_chars=500, preview_chars=200`) produces byte-identical summaries to before.
- Previously overflowing configs now stay within budget: the repro above yields 98 characters instead of 5062.
- The `len(summary) >= output_len` guard is gone; it is unreachable once the summary is bounded by `max_output_chars`, which is itself below `output_len` whenever trimming runs.
- One existing test (`test_skips_trim_when_summary_would_exceed_original`) asserted the old overflow-then-skip behavior for `max_output_chars=500, preview_chars=490`. That input now trims to a 500-character summary instead of being skipped, so the test is rewritten as `test_summary_never_exceeds_max_output_chars`.

### Tests

- `test_trimmed_output_fits_budget_for_any_config` — parametrized over `max_output_chars` × `preview_chars`, asserts no configuration can produce a replacement larger than the budget.
- `test_summary_reports_the_actual_preview_length` — covers the header defect.
- `test_legacy_tool_search_results_fit_budget` — covers the second free-text path.

All fail on `main` and pass here; the rest of the trimmer suite is unchanged.


seratch:
Closing this PR because the referred issue was closed: https://github.com/openai/openai-agents-python/issues/4686#issuecomment-5435426058

PR #4694, opened 2026-08-26
fix(core): strictify schemas inside tuple prefixItems

XuQuanxin04:
### Summary

`ensure_strict_json_schema` recursed into `items`, `anyOf`, `oneOf`, `allOf`, `properties`, and `$defs`, but it never visited `prefixItems`. pydantic emits `prefixItems` for fixed-length tuple type hints, so a non-strict subschema hidden inside a tuple element was passed through unchanged.

Concretely, a `dict[str, int]` parameter is rejected at the top level because its `additionalProperties` is a schema rather than `false`:

```
additionalProperties should not be set for object types. This could be because you're using an older version of Pydantic
```

But the same type wrapped in a tuple — `def f(data: tuple[dict[str, int], str])` — silently produced:

```json
"prefixItems": [
  { "type": "object", "additionalProperties": { "type": "integer" } },
  { "type": "string" }
]
```

That schema is not valid for OpenAI Structured Outputs, so the tool/structured-output call is rejected by the API at request time.

### Fix

Recurse into each `prefixItems` entry exactly like array `items`, so tuple elements are strictified consistently. `$defs` are still strictified at the root, variable-length tuples (`tuple[int, ...]`) already use `items` and were already handled, and primitive/model tuples remain valid.

### Test plan

Added two tests to `tests/test_strict_schema.py`:
- `test_prefix_items_entries_are_strictified` — an object entry inside `prefixItems` gets `additionalProperties: false` and a `required` list.
- `test_open_object_inside_prefix_items_is_rejected_like_top_level` — an `additionalProperties` schema inside a tuple element raises the same `UserError` as the top-level case.

Verified directly with `function_schema`:
- `tuple[dict[str, int], str]` now raises the same `UserError` (before: silently non-strict).
- `tuple[Coord, str]`, `tuple[str, int, bool]`, `tuple[Pt, Pt]`, and `tuple[int, ...]` all produce valid strict schemas.

- `pytest tests/test_strict_schema.py tests/test_function_schema.py tests/test_function_tool.py` — 172 passed
- `pytest tests/models/test_openai_responses.py tests/models/test_openai_chatcompletions.py tests/models/test_openai_chatcompletions_stream.py` — 317 passed
- `ruff format --check` and `ruff check` clean on changed files.

No issue number; self-contained strict-schema correctness fix.

### Checks

- [ ] I've run `.agents/skills/code-change-verification/scripts/run.sh` — could not run the bash script end-to-end on this Windows/Git-Bash host (POSIX process-group management via `setsid`/`kill -PID`); I ran its constituent steps directly (ruff format, ruff check, and the relevant pytest suites above).

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -273,6 +273,23 @@ def _ensure_strict_json_schema(
             inside_nested_resource=inside_nested_resource,
         )
 
+    # tuples: { 'type': 'array', 'prefixItems': [{...}, ...] }
+    # pydantic emits prefixItems for fixed-length tuple type hints. Each element
+    # is an independent subschema that must be strictified just like array items.
+    prefix_items = json_schema.get("prefixItems")
+    if is_list(prefix_items):
+        json_schema["prefixItems"] = [
+            _ensure_strict_json_schema(
+                entry,
+                path=(*path, "prefixItems", str(i)),
+                root=root,
+                budget=budget,
+                depth=next_depth,
+                inside_nested_resource=inside_nested_resource,
+            )
+            for i, entry in enumerate(prefix_items)
+        ]
+
     # unions
     any_of = json_schema.get("anyOf")
     if is_list(any_of):
--- a/tests/test_strict_schema.py
+++ b/tests/test_strict_schema.py
@@ -990,3 +990,47 @@ def test_ref_with_anchor_is_still_expanded():
         "$anchor": "value",
         "type": "string",
     }
+
+
+def test_prefix_items_entries_are_strictified():
+    # pydantic emits prefixItems for fixed-length tuple type hints. Each element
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
The prefixItems recursion is the correct implementation direction. Please add one caller-visible regression that starts from an actual annotated function or Pydantic model, runs through function_schema(), and verifies that an object nested in a fixed tuple is strictified. The current hand-written schema tests prove the helper recursion but not the public schema-generation path that exposed the bug.

seratch:
Thanks for the focused strict-schema fix. #4696 now contains the same prefixItems traversal together with the originating fixed-tuple function-schema behavior and public-boundary regression coverage.

We should merge only one source of truth for this change, so I am closing this PR as superseded by #4696.

PR #4695, opened 2026-08-26
Tracing: wait on an event instead of polling in the batch worker

### LINKED ISSUE #4688: BatchTraceProcessor worker thread busy-polls with time.sleep(0.2) instead of waiting on its shutdown event
### Describe the bug

`BatchTraceProcessor`'s background export thread polls in a fixed 200 ms loop:

https://github.com/openai/openai-agents-python/blob/main/src/agents/tracing/processors.py#L653-L665

```python
def _run(self):
    while not self._shutdown_event.is_set():
        current_time = time.monotonic()
        queue_size = self._queue.qsize()

        if current_time >= self._next_export_time or queue_size >= self._export_trigger_size:
            self._export_batches()
            self._next_export_time = time.monotonic() + self._schedule_delay
        else:
            # Sleep a short interval so we don't busy-wait.
            time.sleep(0.2)
```

The thread is started lazily on the first trace/span (`_ensure_thread_started`) and then never stops, so for the rest of the process lifetime it wakes five times a second and does nothing. With the default `schedule_delay = 5.0` (`processors.py:572`), 24 of every 25 wakeups are pure overhead — constant background CPU in profiles of an otherwise idle agent service, and timer wakeups on battery-powered or heavily-multiplexed hosts.

`time.sleep()` is also not interruptible by the shutdown event, so `shutdown()` waits up to an extra 200 ms per call even when the queue is empty.

Notably, `BackendSpanExporter._sleep_before_retry` in this same file already does the right thing on its no-deadline path (`processors.py:236`):

```python
```

so the worker loop is the odd one out rather than a deliberate style choice.

### Debug information

- Agents SDK version: `main` (`src/agents/tracing/processors.py`, `_run` unchanged)
- Python: any

### Expected behavior

`self._shutdown_event` is already a `threading.Event`; waiting on it fixes both problems at once — no wakeups between scheduled flushes, and shutdown becomes immediate:

```python
timeout = max(0.0, min(0.2, self._next_export_time - time.monotonic()))
if self._shutdown_event.wait(timeout):
    break
```

The `min(0.2, ...)` preserves the existing responsiveness to the `_export_trigger_size` queue-size trigger; blocking on `queue.Queue.get` with a timeout would remove that last poll too, at the cost of a larger change to the drain logic. Happy to send a PR for either shape.

rajarshidattapy:
Fixes #4688.

### Problem

`BatchTraceProcessor`'s export thread polled in a fixed 200 ms loop:

```python
if current_time >= self._next_export_time or queue_size >= self._export_trigger_size:
    self._export_batches()
    self._next_export_time = time.monotonic() + self._schedule_delay
else:
    # Sleep a short interval so we don't busy-wait.
    time.sleep(0.2)
```

The thread starts lazily on the first trace/span and never stops, so for the rest of the process lifetime it woke five times a second and did nothing. With the default `schedule_delay = 5.0`, 24 of every 25 wakeups were pure overhead — constant background CPU in profiles of an idle agent service, and timer wakeups on battery-powered hosts.

`time.sleep()` is also not interruptible by the shutdown event, so `shutdown()` waited out the rest of the interval before the final drain even with an empty queue.

### Fix

The worker now blocks until it actually has something to do. It waits on a `threading.Event` with a timeout of "time until the next scheduled export", so an idle process wakes once per `schedule_delay` instead of 25 times.

Both of the things that previously needed polling to be noticed now signal the event explicitly:

- **the queue-size trigger** — `on_trace_start`/`on_span_end` set the event once the queue reaches `_export_trigger_size`, so the immediate-export path is as responsive as before (more so: it no longer waits for the next poll tick);
- **shutdown** — `shutdown()` sets it alongside `_shutdown_event`, so the worker leaves the loop and drains immediately.

This is the second shape offered in the issue rather than the `min(0.2, ...)` one-liner: capping the wait at 200 ms would have fixed the shutdown latency but left the idle wakeups — the main complaint — in place. It keeps the existing drain logic untouched, unlike blocking on `queue.Queue.get`.

There is no lost-wakeup window. A `set()` that lands while the worker is exporting, or between its `wait()` returning and its `clear()`, is covered by the queue-size check at the top of the next iteration — the condition that caused the `set()` is re-read there, not assumed.

As the issue notes, `BackendSpanExporter._sleep_before_retry` in this same file already waits on its shutdown event, so this brings the worker in line with it.

### Tests

- `test_batch_trace_processor_queue_trigger_wakes_the_idle_worker` — with `schedule_delay=60`, a sub-trigger queue leaves the worker parked and reaching the trigger wakes it. Hangs to failure if the producer-side signal is dropped.
- `test_batch_trace_processor_shutdown_wakes_the_idle_worker` — `shutdown()` on a parked worker joins promptly and still drains the queued span, instead of timing out and warning.
- `test_batch_trace_processor_schedule_uses_monotonic_clock` drove `_run` through its `time.sleep` calls, so it now drives it through the wake event's `wait` instead. Its `ControlledTime.sleep` raises, which makes the test fail if polling is ever reintroduced into the loop.

Each was verified to fail against a deliberately broken version of the corresponding change.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -568,6 +568,9 @@ class BatchTraceProcessor(TracingProcessor):
         self._max_batch_size = max_batch_size
         self._schedule_delay = schedule_delay
         self._shutdown_event = threading.Event()
+        # Set when the worker has something to do before its next scheduled export: the queue
+        # reached the trigger size, or we are shutting down.
+        self._wake_event = threading.Event()
 
         # The queue size threshold at which we export immediately.
         self._export_trigger_size = max(1, int(max_queue_size * export_trigger_ratio))
@@ -602,6 +605,8 @@ class BatchTraceProcessor(TracingProcessor):
             self._queue.put_nowait(trace)
         except queue.Full:
             logger.warning("Queue is full, dropping trace.")
+            return
+        self._wake_worker_if_queue_is_full_enough()
 
     def on_trace_end(self, trace: Trace) -> None:
         # We send traces via on_trace_start, so we don't need to do anything here.
@@ -619,12 +624,20 @@ class BatchTraceProcessor(TracingProcessor):
             self._queue.put_nowait(span)
         except queue.Full:
             logger.warning("Queue is full, dropping span.")
+            return
+        self._wake_worker_if_queue_is_full_enough()
+
+    def _wake_worker_if_queue_is_full_enough(self) -> None:
+        """Wake the worker early when the queue reaches the export trigger size."""
+        if self._queue.qsize() >= self._export_trigger_size:
+            self._wake_event.set()
 
     def shutdown(self, timeout: float | None = None):
         """
         Called when the application stops. We signal our thread to stop, then join it.
         """
         self._shutdown_event.set()
+        self._wake_event.set()
         if timeout is not None:
             request_exporter_shutdown = getattr(self._exporter, "_request_shutdown", None)
             if callable(request_exporter_shutdown):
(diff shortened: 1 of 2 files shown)


seratch:
Please refer to https://github.com/openai/openai-agents-python/issues/4688#issuecomment-5435446860

PR #4706, opened 2026-08-27
fix(core): make prompt input cancellable

### LINKED ISSUE #4670: `run_demo_loop` blocks the asyncio event loop on `input()`

seratch:
This pull request fixes #4670 by moving `run_demo_loop()` prompt reads off the event loop while giving the REPL deterministic ownership of input cancellation and shutdown.

It preserves normal interactive input, EOF, `quit`/`exit`, readline history, custom SIGINT handlers, and buffered multi-line redirected input across prompts. It also reaps the owned helper after Ctrl+C or programmatic cancellation and adds cross-platform deterministic regression coverage.

Callers that need to consume or frame `sys.stdin` before starting the interactive loop should drive `Runner` directly. This implementation takes over #4671 and preserves the original contributor's authorship with a co-author trailer.

### DIFF
--- a/src/agents/_repl_input_process.py
+++ b/src/agents/_repl_input_process.py
@@ -1,0 +1,100 @@
+from __future__ import annotations
+
+import json
+import os
+import signal
+import sys
+from typing import Any, BinaryIO, cast
+
+
+def _read_exactly(stream: BinaryIO, size: int) -> bytes:
+    chunks: list[bytes] = []
+    remaining = size
+    while remaining:
+        chunk = stream.read(remaining)
+        if not chunk:
+            raise EOFError
+        chunks.append(chunk)
+        remaining -= len(chunk)
+    return b"".join(chunks)
+
+
+def _open_control(value: str) -> BinaryIO:
+    kind, raw_value = value.split(":", 1)
+    if kind == "fd":
+        descriptor = int(raw_value)
+    elif kind == "handle" and sys.platform == "win32":
+        import msvcrt
+
+        descriptor = msvcrt.open_osfhandle(int(raw_value), os.O_RDONLY | getattr(os, "O_BINARY", 0))
+    else:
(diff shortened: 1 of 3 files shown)


seratch:
https://github.com/openai/openai-agents-python/issues/4670#issuecomment-5446434515

Issue #4669, opened 2026-08-26
to_call_args: **kwargs keys that collide with named parameters silently override them or crash with TypeError


mmjerge:
### Please read this first

- **Have you read the docs?** Yes.
- **Have you searched for related issues?** Yes. #3192 and #3200 cover `extra_args` collisions on the model-call path; nothing covers `**kwargs` collisions inside tool invocation.

### Describe the bug

A function tool that takes `**kwargs` accepts schema-valid input that its Python signature cannot express: a key inside the kwargs dict that names one of the function's own parameters. `FuncSchema.to_call_args` does not check for this. What happens next depends on the parameter kind, and both outcomes are wrong.

**1. Keyword-only parameter: the validated value is silently overwritten.**

```python
def g(*, opt: int = 1, **kw: Any) -> str: ...

# model sends {"opt": 5, "kw": {"opt": 9}}
# to_call_args -> args=[], kwargs={"opt": 9}
# the tool runs with opt=9, kw={}
```

The explicit `opt=5` the model supplied — and Pydantic validated — is replaced, and the `"opt"` key disappears from `kw`. No error, no log. The tool executes with arguments that differ from the validated input.

**2. Positional-or-keyword parameter: an unhandled TypeError escapes as an internal error.**

```python
def h(x: int, *rest: int, **kw: Any) -> str: ...

# model sends {"x": 1, "rest": [2, 3], "kw": {"x": 99}}
# to_call_args -> args=[1, 2, 3], kwargs={"x": 99}
# call is h(1, 2, 3, x=99)
# TypeError: h() got multiple values for argument 'x'
```

The invocation path (`src/agents/tool.py:2632-2645`) raises `ModelBehaviorError` for validation failures one line earlier, but the TypeError from the call itself is not converted. The run fails with an internal error instead of the model receiving actionable feedback.

### Mechanism

`FuncSchema.to_call_args` (src/agents/function_schema.py:46-81) binds POSITIONAL_OR_KEYWORD parameters positionally and KEYWORD_ONLY parameters by name, then splats the VAR_KEYWORD dict with `keyword_args.update(value or {})` — after the named parameters, with no intersection check. Dict update order produces the silent override; positional binding produces the TypeError.

Note the input is not nonsense the schema forbids: the generated Pydantic model has separate fields for `opt` and `kw`, and nothing constrains the keys inside `kw`. The schema says yes, then the call layer either lies or crashes.

[...]

PR #4709, opened 2026-08-27
Reject kwargs keys that collide with named tool parameters

mikemikimike:
## Summary

`FuncSchema.to_call_args` could silently overwrite a validated named argument, or produce an invalid Python call, when a `**kwargs` key collided with a named function parameter.

This change rejects those collisions with `ModelBehaviorError` before invocation. Non-conflicting `**kwargs` continue to be passed through unchanged.

## Testing

- `uv run --frozen pytest tests/test_function_schema.py -q` (41 passed)
- `uv run --frozen ruff check src/agents/function_schema.py tests/test_function_schema.py`
- `git diff --check`

Mypy was attempted but did not complete on the local Windows environment.

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -13,7 +13,7 @@ from griffe import Docstring, DocstringSectionKind  # type: ignore[import-untype
 from pydantic import BaseModel, Field, create_model
 from pydantic.fields import FieldInfo
 
-from .exceptions import UserError
+from .exceptions import ModelBehaviorError, UserError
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the fix. #4674 addresses the same #4669 bug and already preserves the valid positional-only case.

This implementation includes positional-only parameter names in the collision set, so it rejects valid Python calls such as `def f(value, /, **kwargs)` with `kwargs={"value": ...}`. Rather than maintain competing implementations, I am closing this in favor of #4674.

PR #4710, opened 2026-08-27
fix(tracing): close the default exporter at shutdown

### LINKED ISSUE #4681: Tracing: the default `BackendSpanExporter`'s HTTP client is never closed

mikemikimike:
Fixes #4681

`BackendSpanExporter` owns a pooled HTTP client, but the default tracing shutdown chain only stopped the processor and never closed that client. This can leave the tracing connection pool open until interpreter teardown and leak a client when the default provider is replaced.

This change adds an explicit setup-boundary cleanup for the lazily-created default exporter after the default provider has flushed and shut down. Injected exporters remain untouched because `BatchTraceProcessor` does not own arbitrary exporter instances.

Tests:
- `uv run --frozen pytest tests/test_trace_processor.py tests/test_tracing.py tests/test_tracing_errors.py tests/tracing -q` (159 passed)
- `uv run --frozen ruff format --check src/agents/tracing/processors.py src/agents/tracing/setup.py tests/test_trace_processor.py`
- `uv run --frozen ruff check src/agents/tracing/processors.py src/agents/tracing/setup.py tests/test_trace_processor.py`
- `uv run --frozen mypy src/agents/tracing/processors.py src/agents/tracing/setup.py`
- `git diff --check`

The focused suite reports one pre-existing Python 3.16 deprecation warning from `asyncio.get_event_loop_policy`.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -761,3 +761,10 @@ def default_processor() -> BatchTraceProcessor:
             _global_processor = processor
 
     return processor
+
+
+def shutdown_default_exporter() -> None:
+    """Close the lazily-created default exporter, if tracing initialized it."""
+    exporter = _global_exporter
+    if exporter is not None:
+        exporter.close()
--- a/src/agents/tracing/setup.py
+++ b/src/agents/tracing/setup.py
@@ -20,6 +20,9 @@ def _shutdown_global_trace_provider() -> None:
 
         if isinstance(provider, DefaultTraceProvider):
             provider.shutdown(timeout=_DEFAULT_SHUTDOWN_TIMEOUT)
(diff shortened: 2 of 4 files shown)


seratch:
Thanks for the implementation. #4712 has the more complete ownership shape: it closes only the module-owned default exporter and clears the associated module singletons.

Both versions still need the explicit public provider-shutdown path addressed, so I am closing this duplicate and continuing the review on #4712.

PR #4711, opened 2026-08-27
Fix tracing exporter reuse after shutdown

### LINKED ISSUE #4683: `BackendSpanExporter._shutdown_event` is never cleared, permanently disabling retries after one shutdown

mikemikimike:
This pull request fixes retry behavior when a `BackendSpanExporter` is reused after a timed shutdown.

`BatchTraceProcessor.shutdown(timeout=...)` signals the exporter to stop retry backoff, but the shared exporter retained that signal when attached to a later processor. The new private reset hook clears the signal when a processor attaches, while preserving the existing shutdown interruption behavior and compatibility with exporters that do not implement the hook.

Validation:
- `uv run pytest -q tests/test_trace_processor.py` (57 passed)
- `uv run ruff format --check .` (921 files already formatted)
- `uv run ruff check .` (passed)
- `git diff --check` (passed)

The full test suite could not be collected in this environment because optional dependencies including `sqlalchemy`, `httpx`, `numpy`, `litellm`, and `docker` are not installed. Repository-wide pyright also reports existing errors in sandbox extension files.

This pull request resolves #4683.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -537,6 +537,9 @@ class BackendSpanExporter(TracingExporter):
     def _request_shutdown(self) -> None:
         self._shutdown_event.set()
 
+    def _reset_shutdown(self) -> None:
+        self._shutdown_event.clear()
+
 
 class BatchTraceProcessor(TracingProcessor):
     """Some implementation notes:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for iterating on this. #4684 implements the same #4683 fix with an outstanding-request count and includes coverage for shared exporters, two processors shutting down independently, timed-out workers, and concurrent shutdown calls.

The generation approach here can allow a later processor to clear the shared event while an older processor still owns an active shutdown request. I am closing this in favor of the request-count implementation in #4684.

PR #4714, opened 2026-08-27
fix(core): apply **kwargs value annotation to each keyword value

abhay-codes07:
### Summary

`function_schema` builds the wrong type for a variadic-keyword parameter annotated with a generic, so the generated tool schema and the argument validation are one nesting level too shallow.

A `**kwargs: X` annotation applies to each keyword *value* (PEP 484), so the collected container is always `dict[str, X]`. For a scalar `X` the SDK already did this, but when `X` was itself a generic `dict[...]` a special case collapsed the value type:

```python
def tool(**kwargs: dict[str, int]): ...
```

- **Generated schema (before):** `kwargs` is `{additionalProperties: {type: integer}}` — i.e. `dict[str, int]`.
- **Correct schema:** `{additionalProperties: {additionalProperties: {type: integer}}}` — i.e. `dict[str, dict[str, int]]`.

The consequences are concrete: the model is told the wrong argument shape, a correctly-shaped call like `{"kwargs": {"group": {"x": 1}}}` is **rejected** with a `ValidationError`, and a wrong flat mapping `{"kwargs": {"a": 1}}` is **accepted**.

This is the exact bug that #4655 fixed for the symmetric variadic-positional case — there, `*args: tuple[int, ...]` was collapsed to `list[int]` instead of `list[tuple[int, ...]]`. This PR applies the same PEP 484 correction to `**kwargs`: the value annotation is preserved as-is and the container is always `dict[str, X]`. The buggy `dict`-origin special case is removed (a scalar `**kwargs: int` still yields `dict[str, int]`, and a bare `**kwargs` still yields `dict[str, Any]`).

### Test plan

- Updated `test_var_keyword_dict_annotation` to mirror `test_var_positional_tuple_annotation` (as revised by #4655): it now asserts the nested `dict[str, dict[str, int]]` schema, round-trips a correctly-shaped input through `to_call_args`, and asserts a flat `dict[str, int]` input raises `ValidationError`.
- Added `test_var_keyword_scalar_annotation` to lock in that `**kwargs: int` stays `dict[str, int]` and round-trips.
- `uv run pytest tests/test_function_schema.py` passes; `ruff format --check`, `ruff check`, and `mypy` on the changed files pass.

### Issue number

None (found while auditing variadic-annotation handling after #4655).

### Checks

- [ ] I've run `make lint`, `make format` and `make tests` (ran targeted lint/format/typecheck and the function-schema suite locally)

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -416,17 +416,12 @@ def function_schema(
             )
 
         elif param.kind == param.VAR_KEYWORD:
-            # **kwargs handling
-            if get_origin(ann) is dict:
-                # e.g. def foo(**kwargs: dict[str, int])
-                dict_args = get_args(ann)
-                if len(dict_args) == 2:
-                    ann = dict[dict_args[0], dict_args[1]]  # type: ignore
-                else:
-                    ann = dict[str, Any]
-            else:
-                # e.g. def foo(**kwargs: int) -> Dict[str, int]
-                ann = dict[str, ann]  # type: ignore
+            # **kwargs handling: a ``**kwargs: X`` annotation applies to each keyword *value*
+            # (PEP 484), so the collected container is always ``dict[str, X]``. Preserve the full
+            # annotation as the value type -- mirroring the variadic-positional handling above,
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@codex review

Issue #3549, opened 2026-06-01
function_schema() raises opaque ValueError for params named model_dump / model_validate / etc.


devteamaegis:
Calling `function_schema()` (or using `@function_tool`) on a function whose parameter is named `model_dump`, `model_dump_json`, `model_validate`, `model_validate_json`, or `model_validate_strings` raises an unhandled `ValueError` deep inside Pydantic rather than a clear SDK error.

**Trigger**

```python
from agents.function_schema import function_schema

def search_tool(model_validate: str, query: str) -> str:
    return f"search: {query}"

schema = function_schema(search_tool)
```

**Traceback**

```
File ".../src/agents/function_schema.py", line 407, in function_schema
    dynamic_model = create_model(f"{func_name}_args", __base__=BaseModel, **fields)
  ...
  File ".../pydantic/_internal/_fields.py", line 97, in _check_protected_namespaces
    raise ValueError(...)
ValueError: Field 'model_validate' conflicts with member <bound method BaseModel.model_validate ...>
  of protected namespace 'model_validate'.
```

Pydantic's `create_model()` refuses any field whose name matches a built-in `BaseModel` method in a protected namespace. The error message gives no hint that the issue is the parameter name or how to fix it. These five names are distinct from `model_config` / `model_fields` / `model_computed_fields` (covered by PR #3548, which surfaces a different `TypeError`).

The fix is to detect these names before calling `create_model()` and raise a `UserError` with an actionable message.

norika1207-lab:
One practical way to make this easier to evaluate is to turn the report into a small contract test around the runtime boundary, not only the user-facing symptom.

For `function_schema opaque ValueError for params named model_dump/model_validate`, I would check three things: first, the failing input is preserved as a minimal fixture; second, the SDK emits the same observable event/result shape before and after the fix except for the intended change; third, the trace or serialized session contains enough metadata to explain what happened after a resume or retry.

The part that matters for agent systems is durability: if a tool call, memory write, handoff, or structured-output step fails, downstream code should not have to infer state from natural-language output. A focused regression test around that boundary would make the issue much easier to close with confidence.

PR #4715, opened 2026-08-27
fix(core): support tool parameters named like BaseModel members

### LINKED ISSUE #3547: TypeError crash when function parameter is named 'model_config'
## What breaks

Calling `function_schema()` (or using `@function_tool`) on a Python function that has a parameter named `model_config` (or another Pydantic-reserved name like `model_fields`, `model_computed_fields`) causes an unhandled `TypeError` deep inside Pydantic internals.

```python

def my_tool(model_config: str) -> str:
    return model_config

schema = function_schema(my_tool)
# TypeError: 'FieldInfo' object is not iterable
```

Full traceback:

```
File "src/agents/function_schema.py", line 407, in function_schema
    dynamic_model = create_model(f"{func_name}_args", __base__=BaseModel, **fields)
  File "pydantic/main.py", line 1828, in create_model
  File "pydantic/_internal/_model_construction.py", line 131, in __new__
    config_wrapper = ConfigWrapper.for_model(bases, namespace, raw_annotations, kwargs)
  File "pydantic/_internal/_config.py", line 143, in for_model
    config_new.update(config_from_namespace)
TypeError: 'FieldInfo' object is not iterable
```

## Root cause

Pydantic's `create_model()` treats keyword arguments named `model_config` as a model-level configuration key, not a field. When `function_schema.py` passes a `FieldInfo` for that name, Pydantic's `ConfigWrapper.for_model()` calls `.update()` on it expecting a dict-like object, causing the crash. No guard exists to detect reserved names before handing them to `create_model()`.

abhay-codes07:
### Summary

A function tool whose parameter name shadows a Pydantic `BaseModel` member crashes at tool-definition time, deep inside `function_schema`'s `create_model` call, with an error that gives no hint about the real cause:

```python

@function_tool
def export_row(model_dump: bool) -> str:   # ValueError: Field 'model_dump' conflicts with member ...
    ...

@function_tool
def configure(model_config: str) -> str:    # TypeError: 'FieldInfo' object is not iterable
    ...
```

`model_dump`, `model_validate`, `model_dump_json`, `model_json_schema` and the other `model_*` methods raise a `ValueError` from Pydantic's protected-namespace guard; `model_config` raises a `TypeError` because `create_model` treats a `model_config` keyword as configuration.

This completes #4627, which taught the runtime call layer to read argument values from instance `__dict__` (rather than by attribute access) precisely so parameters can shadow `BaseModel` members like `model_extra` / `model_fields_set` — but those names still had to be *definable* first, and the `model_*` methods crash before that support is ever reached.

**Change:** build the generated args model on a small base whose `model_config` sets `protected_namespaces=()`, so method-shadowing parameter names no longer trip Pydantic's guard. Because the call layer already resolves values via `__dict__`, a field that shadows a model member is still passed to the tool correctly. `model_config` itself cannot be a Pydantic field name, so it is rejected with an actionable `UserError` ("Rename the parameter…") instead of an opaque `TypeError`.

### Test plan

- `test_parameter_named_like_a_basemodel_method_is_supported` (parametrized over `model_dump`, `model_dump_json`, `model_validate`, `model_validate_json`, `model_json_schema`): the tool is defined, validates input, and round-trips through `to_call_args` to the original function.
- `test_parameter_named_model_config_raises_actionable_error`: `model_config` raises a `UserError` mentioning the parameter, not a `TypeError`.
- 5 of the 6 new cases fail on `main` without the source change.
- The existing #4627 `model_extra` / `model_fields_set` tests still pass unchanged.
- `uv run pytest tests/test_function_schema.py tests/test_function_tool_decorator.py tests/test_tool_converter.py tests/test_released_api_contract.py` → 231 passed. `ruff format --check`, `ruff check`, and `mypy` on the changed file pass.

### Issue number

Follows up #4627 (closed #3547 / #3549 covered the same crash; the runtime shadowing was fixed there, the definition-time crash remained).

### Checks

- [ ] I've run `make lint`, `make format` and `make tests` (ran targeted lint/format/typecheck and the tool suites locally)

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -10,13 +10,26 @@ from typing import Annotated, Any, Literal, get_args, get_origin, get_type_hints
 
 # griffelib exposes the `griffe` package at runtime but currently does not ship typing markers.
 from griffe import Docstring, DocstringSectionKind  # type: ignore[import-untyped]
-from pydantic import BaseModel, Field, create_model
+from pydantic import BaseModel, ConfigDict, Field, create_model
 from pydantic.fields import FieldInfo
 
 from .exceptions import UserError
 from .run_context import RunContextWrapper
 from .strict_schema import ensure_strict_json_schema
 from .tool_context import ToolContext
+
+
+class _ToolArgsBaseModel(BaseModel):
+    """Base for dynamically generated tool-argument models.
+
+    ``protected_namespaces=()`` disables Pydantic's ``model_`` guard so a tool may declare
+    parameters whose names collide with ``BaseModel`` members (for example ``model_dump`` or
+    ``model_validate``) without raising at tool-definition time. The runtime call layer reads
+    argument values from instance ``__dict__`` rather than by attribute access, so a field that
+    shadows a model member is still resolved correctly.
+    """
+
+    model_config = ConfigDict(protected_namespaces=())
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating these Pydantic name collisions. The previously closed reports asked for an actionable failure, not for every BaseModel member name to become a supported tool parameter.

Disabling protected_namespaces for every generated argument model creates a broad new contract and leaves model instances with fields shadowing methods such as model_dump and model_validate. Renaming the parameter or using a wrapper function remains a reasonable supported alternative.

The narrower silent-value corruption for already constructible models was fixed separately in #4627. I am closing this broader support change.

PR #4718, opened 2026-08-27
Add turn-aware session history via get_items(turn_limit)

### LINKED ISSUE #3738: Consider run/turn-aware session history retrieval instead of item-only limit
this issue is similar to #1385

A single user turn may produce multiple persisted items:

1. user input
2. function calling
3. function calling output
4. assistant output

conversation strategy `session`, with item-based limit, the retrieved history can begin in the middle of that sequence, for example including a function calling output without its corresponding function calling, or an assistant output without the user request that triggered it.

can we do:

1. Persist a run_id / turn_id with each session item
2.  Add run_limit / turn_limit for `get_items`, returning complete runs messages

AAliKKhan:
### Summary

Adds whole-turn retrieval for every `SessionABC` backend through a new `turn_limit` keyword on `get_items`. A new `resolve_session_turn_limit` helper picks complete turns (and ignores any conflicting `limit`), and `slice_items_by_turn` returns whole turns so conversation history is never truncated mid-turn when replayed.

- `SessionSettings` gains a `turn_limit` field; `Runner` forwards it through `prepare_input_with_session`.
- All backends (sqlite, async/advanced sqlite, sqlalchemy, mongo, redis, dapr, encrypt, `OpenAIConversationsSession`, `OpenAIResponsesCompactionSession`) accept `turn_limit`.
- `slice_items_by_turn` re-exported from `agents.memory` for public use.

### Test plan

- Unit tests for `resolve_session_turn_limit`, `slice_items_by_turn`, and per-backend `get_items(turn_limit)` (sqlite, async sqlite, conversations, encrypt, compaction).
- Backend integration tests added for `AsyncSQLiteSession`, `OpenAIConversationsSession`, `EncryptedSession`, and `OpenAIResponsesCompactionSession`.
- `uv run ruff format`, `uv run ruff check`, and `uv run mypy` pass on the changed sources.
- Note: on this Windows dev box `tests/test_agent_as_tool.py` has a pre-existing `tee`-related failure unrelated to this change; redis/dapr/mongo live-service suites were not run locally.

### Issue number

Closes #3738

### Checks

### DIFF
--- a/docs/sessions/index.md
+++ b/docs/sessions/index.md
@@ -4,7 +4,7 @@ The Agents SDK provides built-in session memory to automatically maintain conver
 
 Sessions stores conversation history for a specific session, allowing agents to maintain context without requiring explicit manual memory management. This is particularly useful for building chat applications or multi-turn conversations where you want the agent to remember previous interactions.
 
-Use sessions when you want the SDK to manage client-side memory for you. In the same run, a session cannot be combined with the run-level continuation options `conversation_id`, `previous_response_id`, or `auto_previous_response_id`. If you want OpenAI server-managed continuation instead, choose one of those mechanisms rather than layering a session on top.
+Use sessions when you want the SDK to manage client-side memory for you. Sessions cannot be combined with `conversation_id`, `previous_response_id`, or `auto_previous_response_id` in the same run. If you want OpenAI server-managed continuation instead, choose one of those mechanisms rather than layering a session on top.
 
 ## Quick start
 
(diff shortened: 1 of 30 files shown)


seratch:
Thanks for the extensive implementation, but #3738 was deliberately closed after deciding not to add a durable run/turn grouping contract across every Session implementation. Applications needing semantic turn selection can use session_input_callback or a custom Session; the narrower orphaned-output invariant is tracked separately.

This PR changes the core Session protocol and nearly every backend, rewrites unrelated backend lifecycle code, and has already introduced concrete cancellation, transaction, ownership, configuration, and documentation regressions. That is not an acceptable implementation path for the original request.

I am closing this in line with the existing #3738 maintainer decision.

PR #4719, opened 2026-08-27
fix(core): resolve JSON Pointer array indexes in $ref paths

hsusul:
### Summary

\`resolve_ref()\` treated every JSON Pointer hop as a dict lookup. A valid pointer such as \`#/allOf/0\` therefore crashed with \`AssertionError: encountered non-dictionary entry\` because \`allOf\` is an array.

JSON Pointer (RFC 6901) uses decimal array indexes. OpenAPI and MCP schemas commonly \`$ref\` into \`allOf\` / \`anyOf\` / \`prefixItems\` members. Default \`FunctionTool\` construction (\`strict_json_schema=True\`) and MCP \`convert_schemas_to_strict=True\` run this path, so those tools fail at setup.

### Reproduction

```python
from agents.tool import FunctionTool

schema = {
    \"type\": \"object\",
    \"properties\": {
        \"value\": {\"\$ref\": \"#/allOf/0\", \"description\": \"first member\"},
    },
    \"allOf\": [
        {\"type\": \"object\", \"properties\": {\"a\": {\"type\": \"string\"}}, \"required\": [\"a\"], \"additionalProperties\": False},
        {\"type\": \"object\", \"properties\": {\"b\": {\"type\": \"integer\"}}, \"required\": [\"b\"], \"additionalProperties\": False},
    ],
}

FunctionTool(name=\"t\", description=\"d\", params_json_schema=schema, on_invoke_tool=lambda ctx, s: s)
# Before: AssertionError
# After: schema is inlined as a strict object
```

### Solution

Walk dict keys and RFC 6901 array indexes. Reject non-integer tokens, leading zeros, and out-of-range indexes with \`ValueError\`. Nested \`$id\` resource checks still apply only to dict nodes.

### Test plan

- [x] Direct \`resolve_ref(\"#/allOf/0\")\` and \`#/allOf/1\`
- [x] Invalid array token raises \`ValueError\`
- [x] \`ensure_strict_json_schema\` inlines a sibling \`$ref\` into an \`allOf\` member
- [x] \`uv run pytest tests/test_strict_schema.py\` (75 passed)
- [x] \`ruff\` / \`pyright\` on the changed files

### Issue number

N/A

### Checks

- [ ] If using Codex, I've run \`/review\` before submitting this PR

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -425,7 +425,12 @@ def _json_pointer_child(container: object, token: str, *, ref: str) -> object:
     if is_dict(container):
         return container[token]
     if is_list(container):
-        if not token.isdigit() or (len(token) > 1 and token.startswith("0")):
+        # RFC 6901 array indexes use JSON decimal digits (ASCII 0-9 only).
+        if (
+            not token
+            or not all("0" <= char <= "9" for char in token)
+            or (len(token) > 1 and token.startswith("0"))
+        ):
             raise ValueError(f"Invalid JSON Pointer array index {token!r} while resolving {ref}")
         index = int(token)
         if index >= len(container):
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -410,18 +410,30 @@ def resolve_ref(*, root: dict[str, object], ref: str) -> object:
         raise ValueError(f"Unexpected $ref format {ref!r}; Does not start with #/")
 
     path = ref[2:].split("/")
-    resolved = root
+    resolved: object = root
     for raw_key in path:
         key = raw_key.replace("~1", "/").replace("~0", "~")
-        value = resolved[key]
-        assert is_dict(value), (
-            f"encountered non-dictionary entry while resolving {ref} - {resolved}"
-        )
-        resolved = value
-        if _declares_schema_resource(resolved):
+        resolved = _json_pointer_child(resolved, key, ref=ref)
+        if is_dict(resolved) and _declares_schema_resource(resolved):
             raise UserError(_NESTED_RESOURCE_REF_ERROR)
 
     return resolved
+
+
+def _json_pointer_child(container: object, token: str, *, ref: str) -> object:
+    """Return the JSON Pointer child named by ``token`` (RFC 6901)."""
+    if is_dict(container):
+        return container[token]
+    if is_list(container):
+        if not token.isdigit() or (len(token) > 1 and token.startswith("0")):
+            raise ValueError(f"Invalid JSON Pointer array index {token!r} while resolving {ref}")
+        index = int(token)
+        if index >= len(container):
+            raise ValueError(
+                f"JSON Pointer array index {token} is out of range while resolving {ref}"
+            )
+        return container[index]
+    raise TypeError(f"encountered non-container entry while resolving {ref} - {container}")
 
 
 def _declares_schema_resource(schema: dict[str, object]) -> bool:
--- a/tests/test_strict_schema.py
+++ b/tests/test_strict_schema.py
@@ -709,6 +709,29 @@ def test_resolve_ref_rejects_non_integer_json_pointer_array_index():
         resolve_ref(root=root, ref="#/allOf/x")
 
 
+@pytest.mark.parametrize(
+    "index_token",
+    [
+        "\u0660",  # Arabic-Indic digit zero
+        "\u0661",  # Arabic-Indic digit one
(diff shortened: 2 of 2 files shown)


seratch (changes requested):
The array-index support is needed, but `str.isdigit()` accepts non-ASCII numerals such as Arabic-Indic and full-width digits. Those are not valid JSON Pointer array-index tokens and can cause a malformed third-party schema to resolve to a different array member instead of being rejected.

Please restrict the token to the ASCII array-index grammar, preserving the existing leading-zero and bounds checks, and add cases for Arabic-Indic, full-width, and mixed digit strings.

hsusul:
Addressed in 6b4ee865.

`_json_pointer_child` now validates array-index tokens with ASCII `0`-`9` only, instead of `str.isdigit()`. Leading-zero and out-of-range checks are unchanged.

Added regression coverage for Arabic-Indic (`\u0660`, `\u0661`), full-width (`\uff10`, `\uff11`), and mixed ASCII/non-ASCII digit strings.

seratch (changes requested):
Thanks, the resolver now follows RFC 6901's ASCII array-index grammar, and I do not see a remaining defect in the implementation itself. Before merge, please strengthen the regression coverage at the supported boundary.

The current `test_ensure_strict_json_schema_inlines_ref_to_allof_member` fixture is unsatisfiable: the root permits only `value`, while the two `allOf` branches respectively require and permit only `a` or `b`. It therefore proves traversal, but not that a usable strict schema is produced. Please replace it with a satisfiable valid JSON Schema and exercise `FunctionTool` construction or an equivalent public conversion boundary. Also add leading-zero and out-of-range cases, since regressions in those branches could accept or select the wrong array location.

With those focused tests and green CI, this should be ready.

src/agents/strict_schema.py
```diff
@@ -410,20 +410,37 @@ def resolve_ref(*, root: dict[str, object], ref: str) -> object:
...
+    resolved: object = root
     for raw_key in path:
         key = raw_key.replace("~1", "/").replace("~0", "~")
-        value = resolved[key]
-        assert is_dict(value), (
-            f"encountered non-dictionary entry while resolving {ref} - {resolved}"
-        )
-        resolved = value
-        if _declares_schema_resource(resolved):
+        resolved = _json_pointer_child(resolved, key, ref=ref)
```
Jokasa7, line 416: [P2] This lookup uses the same `root` dictionary that strict conversion is already mutating. Because `$defs` are processed before `properties`, a definition such as `{"allOf": [{"type": "string"}]}` has its singleton `allOf` collapsed and removed before a property-level `{"$ref": "#/$defs/T/allOf/0", "description": "value"}` is resolved here. The pointer is valid in the input schema, but `FunctionTool` construction now raises `KeyError: 'allOf'` instead of inlining its target. A target below `oneOf` is similarly erased when that key is renamed to `anyOf`. The current root-level `allOf` test passes only because properties are visited first at that level. Please resolve pointers against an unmodified source document (or preserve addressability until refs are expanded) and add a public-boundary regression with the target under `$defs`.

Jokasa7 (commented):
The array-index parser itself now follows the intended ASCII grammar, but the new valid index paths expose an ordering problem in strict conversion: references are resolved against the same schema object that earlier recursive passes mutate in place.

On this exact head, `$defs` are normalized before `properties`. A definition shaped as `{"allOf": [{"type": "string"}]}` therefore loses its `allOf` key during the singleton merge. A later property with `{"$ref": "#/$defs/T/allOf/0", "description": "value"}` then reaches the new array-aware resolver with a valid pointer whose path has already been erased, so public `FunctionTool` construction raises `KeyError: 'allOf'`. Targets below `oneOf` have the same ordering hazard because that key is renamed to `anyOf` first. The current root-level `allOf` fixture is visited after properties and does not expose this case.

Please keep an untouched source document for pointer resolution, or otherwise preserve pointer addressability until expansion is complete, and add a public-boundary regression whose target is nested under `$defs`.

seratch:
Thanks for iterating on this. After reconsidering the scope, I don't think we should continue expanding the shared strict-schema converter without a concrete producer or application that needs array-index references.

The constructed schemas demonstrate the failure, but they do not establish the claimed OpenAPI/MCP compatibility need. The current implementation also leaves interactions with schema normalization unresolved. Extending it further would add maintenance obligations before we have established the practical requirement.

My earlier review focused too much on completing the implementation before establishing that requirement. I'm closing this as not planned. We can revisit it with a real producer-generated schema and a concrete blocked workflow.

PR #4720, opened 2026-08-27
fix(util): keep memoryview opaque in dump-compatible conversion

hsusul:
### Summary

\`_to_dump_compatible()\` already leaves \`bytes\` and \`bytearray\` opaque, but it still walked \`memoryview\` as a generic iterable. Nested binary payloads (for example input audio in \`ItemHelpers.input_to_new_input_list\` and Chat Completions request conversion) were therefore rewritten to lists of integers such as \`[104, 101, 108, 108, 111]\`. \`json.dumps\` then succeeded, so the API could receive corrupted data instead of failing closed.

Tracing already treats \`memoryview\` as bytes-like. This change matches that rule.

### Reproduction

```python
from agents.util._json import _to_dump_compatible

print(_to_dump_compatible(memoryview(b\"hello\")))
# Before: [104, 101, 108, 108, 111]
# After: memoryview(b'hello')
```

### Solution

Exclude \`memoryview\` from the iterable conversion, next to \`str | bytes | bytearray\`.

### Test plan

- [x] Regression that a top-level and nested \`memoryview\` is left unchanged
- [x] \`uv run pytest tests/utils/test_json.py\` (3 passed)
- [x] \`ruff\` / \`pyright\` on the changed files

### Issue number

N/A. Related but different: #3700 preserved non-dict Mapping values in the same helper.

### Checks

- [ ] If using Codex, I've run \`/review\` before submitting this PR

### DIFF
--- a/src/agents/util/_json.py
+++ b/src/agents/util/_json.py
@@ -73,7 +73,7 @@ def _to_dump_compatible_internal(obj: Any) -> Any:
     if isinstance(obj, list | tuple):
         return [_to_dump_compatible_internal(x) for x in obj]
 
-    if isinstance(obj, Iterable) and not isinstance(obj, str | bytes | bytearray):
+    if isinstance(obj, Iterable) and not isinstance(obj, str | bytes | bytearray | memoryview):
         return [_to_dump_compatible_internal(x) for x in obj]
 
     return obj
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the defensive change. memoryview is not part of the public Responses input-item or input-audio contract in this SDK, and the PR does not identify an ordinary supported producer that can place one at this boundary.

Leaving the object opaque only changes the failure from integer-list conversion to a later serialization error; it does not establish supported binary handling. Without a concrete supported path and desired serialization contract, I am closing this as unnecessary edge-case handling.

PR #4721, opened 2026-08-27
fix(sessions): cap Conversations item pagination at the API limit

hsusul:
### Summary

Cap the Conversations items provider page size at 100 while preserving the caller's larger `get_items(limit=N)` cutoff. This lets session callers retrieve more than 100 recent items without sending an invalid page-size value to the API, while retaining chronological output ordering.

### Test plan

- `uv run pytest -q tests/memory/test_openai_conversations_session.py` — 39 passed
- `uv run ruff format --check` — 921 files already formatted
- `uv run ruff check` — passed
- `uv run python .github/scripts/check_optional_truthiness.py src/agents` — passed
- `uv run mypy src` — passed (307 source files)
- `uv run pyright --project pyrightconfig.json --threads "${PYRIGHT_THREADS:-4}"` — passed
- `make tests-serial` — 77 passed, 4 skipped, 55 deselected
- `make tests-parallel` — 9234 passed, 28 skipped; one unrelated failure in `tests/sandbox/test_run_cwd.py::test_python_skill_uses_absolute_root_from_nested_workdir`, reproduced unchanged on `upstream/main`

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -11,6 +11,8 @@ from ..items import TResponseInputItem
 from .session import SessionABC
 from .session_settings import SessionSettings, coerce_session_settings, resolve_session_limit
 
+_MAX_ITEMS_PAGE_SIZE = 100
+
 
 async def start_openai_conversations_session(openai_client: AsyncOpenAI | None = None) -> str:
     _maybe_openai_client = openai_client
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating and fixing the large-limit failure. The underlying bug is real, but I do not think we should copy the Conversations API's current page-size maximum into the Agents SDK. `_MAX_ITEMS_PAGE_SIZE = 100` creates a second source of truth for provider-owned behavior, and the added test makes that duplicated constraint part of this repository's implementation.

The simpler fix is to omit `limit` from `conversations.items.list()` and let the OpenAI Python client handle provider pagination, while the existing local cutoff continues to enforce the `Session.get_items(limit=N)` contract.

Since that replaces the central implementation and test in this PR rather than refining them, I am going to close this PR and pursue the narrower approach separately. Thank you for identifying the original issue.

PR #4722, opened 2026-08-27
Validate ToolOutputTrimmer integer settings

hsusul:
### Summary

Reject non-integer `ToolOutputTrimmer` limits during construction instead of allowing them to fail later during trimming. Adds regression coverage for all three integer settings with float and boolean inputs.

### Test plan

- `uv run pytest -q tests/extensions/test_tool_output_trimmer.py` — 66 passed
- `uv run ruff format --check` — 921 files already formatted
- `uv run ruff check` — passed
- `uv run python .github/scripts/check_optional_truthiness.py src/agents` — passed
- `uv run mypy src` — passed (307 source files)
- `uv run pyright --project pyrightconfig.json --threads "${PYRIGHT_THREADS:-4}"` — passed
- `make tests-serial` — 77 passed, 4 skipped, 55 deselected
- `make tests-parallel` — 9239 passed, 28 skipped; one unrelated failure in `tests/sandbox/test_run_cwd.py::test_python_skill_uses_absolute_root_from_nested_workdir`, reproduced unchanged on `upstream/main`

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/extensions/tool_output_trimmer.py
+++ b/src/agents/extensions/tool_output_trimmer.py
@@ -115,6 +115,10 @@ class ToolOutputTrimmer:
     trimmable_tools: str | Iterable[str] | None = field(default=None)
 
     def __post_init__(self) -> None:
+        for field_name in ("recent_turns", "max_output_chars", "preview_chars"):
+            value = getattr(self, field_name)
+            if not isinstance(value, int) or isinstance(value, bool):
+                raise ValueError(f"{field_name} must be an integer, got {value!r}")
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the validation improvement. These fields are publicly typed as int, and the proposed float and boolean inputs require callers to bypass that contract manually. Existing construction already fails before trimming or any irreversible side effect for the relevant invalid values.

There is no demonstrated SDK-owned risk that warrants duplicating the type contract as runtime validation, so I am closing this change.

PR #4724, opened 2026-08-27
fix(sandbox): keep split UTF-8 characters intact across PTY output windows

ayaangazali:
### Summary

`collect_pty_output` decodes each collection window with `errors="replace"`:

```python
text = output.decode("utf-8", errors="replace")
```

PTY output is not collected once. It is collected in repeated windows over one persistent chunk deque: `pty_exec_start` takes a window, then each `pty_write_stdin` takes another from the same `entry.output_chunks`. So a multi-byte character whose bytes land either side of a window boundary is decoded as two separate partial sequences, and both halves become U+FFFD. The bytes are destroyed at that first decode, so no later window can put the character back together.

Reproduced against `main` with one persistent deque, the producer still running for the first window:

```
window 1 out = b'h\xef\xbf\xbd'
window 2 out = b'\xef\xbf\xbdllo w\xc3\xb6rld \xe2\x9c\x85'
joined       : h??llo wörld ✅     <- two replacement characters
```

After the change the same sequence round trips, with the first window ending on `b"h"` and holding the partial byte back:

```
window 1 out = b'h'
window 2 out = b'\xc3\xa9llo w\xc3\xb6rld \xe2\x9c\x85'
joined       : héllo wörld ✅
```

The fix holds an unfinished trailing sequence back on the deque for the next window. It needs no new state, because the deque already outlives a single window and is owned by the PTY entry. When `is_done()` reports the producer has closed, nothing can complete the sequence, so the existing replacement behaviour is kept for that case.

This is not backend-specific. `collect_pty_output` is the shared helper behind the `unix_local`, `docker`, `blaxel`, `daytona` and `modal` PTY paths, so any of them silently mangles a character that straddles a window today. It is the same class as #4707, which fixed a UTF-8 boundary in the Cloudflare SSE reader; this is the boundary that the shared PTY collector has.

Note on overlap: #4572 also edits this function, adding a final drain immediately above the decode line. The two changes are independent and compose, but they will touch adjacent lines, so this may need a trivial rebase depending on merge order.

### Test plan

Two tests in `tests/sandbox/test_pty_output.py`:

- `test_collect_pty_output_holds_split_utf8_character_for_the_next_window` splits a two-byte character across two windows on one deque, asserts the first window ends on `b"h"` rather than emitting a replacement character, and asserts the two windows concatenate back to the original text.
- `test_collect_pty_output_replaces_partial_utf8_once_the_producer_is_done` pins the deliberate exception: after `is_done()` is true a truncated sequence still becomes U+FFFD and is not left stranded on the deque.

Verified the first fails without the source change by reverting `src/`: it fails on the `b"h"` assertion, which is the corruption itself. The second passes either way, since it pins existing behaviour.

### Issue number

None. Found while sibling-checking #4707.

### Checks

---

I went looking for this straight after #4707 landed, on the theory that a UTF-8 boundary bug in one reader usually has a sibling wherever else the SDK decodes bytes it did not accumulate itself. The judgement call I would most like checked is keeping `errors="replace"` once the producer is done, since holding bytes back forever would be worse than losing them. I'm a freshman in college and the PTY collector is a path I only learned recently, so please push back if the carry-back belongs in the backends rather than the shared helper.

### DIFF
--- a/src/agents/sandbox/session/pty_output.py
+++ b/src/agents/sandbox/session/pty_output.py
@@ -14,14 +14,18 @@ def _incomplete_utf8_suffix_length(data: bytes | bytearray) -> int:
         byte = data[-back]
         if byte < 0x80:
             return 0
-        if byte >= 0xC0:
-            if byte >= 0xF0:
-                needed = 4
-            elif byte >= 0xE0:
-                needed = 3
-            else:
-                needed = 2
-            return back if back < needed else 0
+        # Only real lead bytes can still be completed. 0xC0, 0xC1 and 0xF5 to 0xFF never
+        # start a valid sequence, so carrying them would withhold a byte that no later
+        # output can finish and would suppress the replacement character forever.
+        if 0xC2 <= byte <= 0xDF:
+            needed = 2
+        elif 0xE0 <= byte <= 0xEF:
+            needed = 3
+        elif 0xF0 <= byte <= 0xF4:
+            needed = 4
+        else:
+            return 0
+        return back if back < needed else 0
     return 0
 
 
--- a/src/agents/sandbox/session/pty_output.py
+++ b/src/agents/sandbox/session/pty_output.py
@@ -10,10 +10,15 @@ from .pty_types import truncate_text_by_tokens
 
 def _incomplete_utf8_suffix_length(data: bytes | bytearray) -> int:
     """Return how many trailing bytes start a UTF-8 sequence that is not finished yet."""
-    for back in range(1, min(3, len(data)) + 1):
+    # A sequence is at most four bytes, so the lead byte is within four of the end.
+    for back in range(1, min(4, len(data)) + 1):
         byte = data[-back]
         if byte < 0x80:
+            # ASCII cannot be part of a multi-byte sequence, so nothing is pending.
             return 0
+        if byte < 0xC0:
+            # Continuation byte. Keep walking back to find the lead byte it belongs to.
+            continue
         # Only real lead bytes can still be completed. 0xC0, 0xC1 and 0xF5 to 0xFF never
         # start a valid sequence, so carrying them would withhold a byte that no later
         # output can finish and would suppress the replacement character forever.
@@ -26,6 +31,7 @@ def _incomplete_utf8_suffix_length(data: bytes | bytearray) -> int:
         else:
             return 0
         return back if back < needed else 0
+    # Four trailing bytes with no lead byte cannot be completed either.
     return 0
 
 
--- a/src/agents/sandbox/session/pty_output.py
+++ b/src/agents/sandbox/session/pty_output.py
@@ -8,6 +8,23 @@ from collections.abc import Callable
 from .pty_types import truncate_text_by_tokens
 
 
+def _incomplete_utf8_suffix_length(data: bytes | bytearray) -> int:
+    """Return how many trailing bytes start a UTF-8 sequence that is not finished yet."""
+    for back in range(1, min(3, len(data)) + 1):
+        byte = data[-back]
+        if byte < 0x80:
+            return 0
+        if byte >= 0xC0:
+            if byte >= 0xF0:
+                needed = 4
+            elif byte >= 0xE0:
+                needed = 3
+            else:
+                needed = 2
+            return back if back < needed else 0
+    return 0
+
+
 async def collect_pty_output(
     *,
     output_chunks: deque[bytes],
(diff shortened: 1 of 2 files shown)


ayaangazali:
Good catch, this was a real regression in my change and it is fixed in `eb683f31`.

You are right that `0xC0`, `0xC1` and `0xF5` to `0xFF` never start a valid sequence, so treating anything at or above `0xC0` as a lead byte meant those bytes were requeued forever instead of being replaced. Reproduced it across repeated polls on a still-running producer:

```
before   invalid lead 0xC0   outs=[b'prompt', b'', b'']            still_queued=[b'\xc0']
         invalid lead 0xFF   outs=[b'prompt', b'', b'']            still_queued=[b'\xff']
after    invalid lead 0xC0   outs=[b'prompt\xef\xbf\xbd', b'', b''] still_queued=[]
         invalid lead 0xFF   outs=[b'prompt\xef\xbf\xbd', b'', b''] still_queued=[]
```

Your interactive-prompt example is the case that made this matter: the byte was withheld while the process sat waiting for input, so the prompt never reached the caller at all. That is worse than the replacement character it was meant to avoid.

Carry-over is now restricted to `C2` to `DF`, `E0` to `EF` and `F0` to `F4`, and everything else falls through to the existing `errors="replace"` path. A parametrized test over `0xC0`, `0xC1`, `0xF5` and `0xFF` asserts the byte is replaced in the same window and nothing is left queued; it fails on the previous commit for all four. The valid-but-incomplete case still carries as before.

Full verification stack passes.

ayaangazali:
Right again, and this one was worse than it looked: my previous commit only ever protected two-byte characters.

The scan returned zero the moment it saw a byte that was not a lead byte, and a continuation byte is exactly what a split three or four byte character leaves at the end of a window. So `E2 82 | AC` hit `0x82` first, returned zero, and both halves still became replacement characters.

Fixed in `65a333b3` by skipping continuation bytes and walking back to the lead byte, scanning up to four bytes since that is the longest sequence. Four trailing bytes with no lead byte still fall through to replacement.

The real lesson is that my test only covered a two-byte character split after one byte, which is the single case that happened to work. The helper is now tested directly across every split position of every sequence length:

```
ascii 0 | 2-byte: split-1 1, complete 0
3-byte: split-1 1, split-2 2, complete 0
4-byte: split-1 1, split-2 2, split-3 3, complete 0
invalid leads C0 C1 F5 FF 0 | stray continuation 0 | four continuations 0
```

Plus a window-level test parametrized over both split points of a three-byte character. On the previous commit the `split-2` and `split-3` rows and the three-byte window case all fail, which is the gap you described.

Full verification stack passes. One unrelated flake, `test_multiple_tool_calls_bound_cancelled_sibling_self_rescheduling_cleanup`, appeared once under xdist and passes in isolation and on a clean rerun; it is a cancellation test my diff does not touch.

ayaangazali:
Both findings are correct, and taken together they say my approach is the wrong shape rather than incomplete. I am closing this.

**On the malformed prefixes.** You are right that `E0 80`, `F0 80` and `F4 90` can never become valid and would be held forever. Working through it, the deeper problem is that I was hand-rolling incremental UTF-8 validation, which is why each round has found another case. The stdlib already does this correctly:

```
codecs.getincrementaldecoder("utf-8")("replace")
  a\xe0\x80  -> decodes 'a??'   pending=0    correctly rejected
  a\xf0\x80  -> decodes 'a??'   pending=0    correctly rejected
  a\xf4\x90  -> decodes 'a??'   pending=0    correctly rejected
  a\xed\xa0  -> decodes 'a'     pending=2    still buffered, even by CPython
```

So the right primitive is an incremental decoder, not my scan, and even that still buffers the surrogate prefix.

**On the exit race, which is the one that decided it.** I verified your reading. `_watch_process_exit` awaits `process.wait()`, then gathers the pump tasks, and only then sets `output_closed`. `_finalize_pty_update` pops and terminates the entry as soon as `process.returncode` is set. So there is a real window where my carry puts bytes into a deque that is about to be discarded, and the byte is lost instead of becoming U+FFFD. My change newly reaches that path, so it is mine to own.

Carrying bytes back into a shared deque cannot be made safe without coordinating with entry teardown, and the correct design is decoder state owned by the PTY entry and flushed on exit. That is per-entry state across the unix_local, docker, blaxel, daytona and modal wrappers, which is a much larger change than this PR, and I would rather not keep patching a shape I now think is wrong.

**The underlying bug is real and still there.** On `main`, a character split across two collection windows is destroyed:

```
window 1 = b'h\xef\xbf\xbd'   window 2 = b'\xef\xbf\xbdllo ...'   joined: h??llo ...
```

Happy to open an issue with that reproduction if you want it tracked, or to attempt the per-entry decoder version if you think the cross-backend spread is justified. I did not want to assume either.

Thanks for the three rounds of review. Every one of them was right, and the last one saved me from shipping data loss in place of a replacement character.

PR #4730, opened 2026-08-28
fix(extensions): adopt pre-upgrade SQLiteSession history instead of destroying it

### LINKED ISSUE #4729: Upgrading a SQLiteSession database to AdvancedSQLiteSession silently and permanently destroys the existing conversation history
### Please read this first

- **Have you read the docs?** Yes. The docs call `AdvancedSQLiteSession` "an enhanced version of the basic `SQLiteSession`."
- **Have you searched for related issues?** Yes. #3449/#3478/#3526 added orphan cleanup on branch deletion; #3766 made pop_item branch-aware. None covers pre-upgrade history.

### Describe the bug

Upgrading an existing `SQLiteSession` database to `AdvancedSQLiteSession` — same file, same session_id, the documented enhancement path — silently destroys the conversation.

Two stages. First, the old history is invisible: `get_items()` joins `message_structure`, and messages written by the plain session have no structure rows. Second, the first `pop_item()` or `delete_branch()` permanently deletes it: `_cleanup_orphaned_messages_sync` removes every message row for the session that has no structure reference, and the entire pre-upgrade history matches that predicate. One pop of a *new* item deletes every old row. Switching back to plain `SQLiteSession` afterwards shows nothing. There is no error and no warning; the data is gone.

### Debug information

- Agents SDK version: `main` at e773b154; also v0.22.0
- Python version: 3.10.16 (repo default via `uv`)
- Operating system: macOS (darwin, arm64)
- Model and model provider: n/a — pure SQLite, no model or network involved
- Does the issue reproduce with the latest Agents SDK release? Yes, and on `main`.
- Does the issue occur consistently or intermittently? Consistently, deterministically.

### Repro steps

Self-contained, no API key needed:

```python
import asyncio, sqlite3, tempfile
from pathlib import Path
from agents.memory.sqlite_session import SQLiteSession
from agents.extensions.memory.advanced_sqlite_session import AdvancedSQLiteSession

async def main():
    db = Path(tempfile.mkdtemp()) / "conv.db"
    plain = SQLiteSession("s1", db)
    await plain.add_items([{"role": "user", "content": f"legacy-{i}"} for i in range(5)])
    plain.close()

    adv = AdvancedSQLiteSession(session_id="s1", db_path=db, create_tables=True)
    print("advanced sees:", len(await adv.get_items()), "of 5 legacy items")   # 0
    await adv.add_items([{"role": "user", "content": "new-1"}])
    await adv.pop_item()                                                        # pops new-1
    rows = sqlite3.connect(db).execute("SELECT COUNT(*) FROM agent_messages").fetchone()[0]
    print("raw rows remaining:", rows)                                          # 0 — all 5 deleted
    adv.close()

asyncio.run(main())
```

### Expected behavior

The upgrade preserves the conversation: old items visible through `get_items()`, `pop_item()` removes one item, and moving back to the plain session shows the remaining history.

A structureless session with message rows is the unambiguous signature of an upgraded database, so adopting those rows into the `main` branch at initialization fixes both the invisibility and the destruction without changing the orphan-cleanup semantics that `tests` pin for genuine orphans of structured sessions.

PR with a fix and regression tests is up alongside this issue.

mmjerge:
### Summary

`AdvancedSQLiteSession` is documented as an enhanced version of the plain `SQLiteSession` over the same tables, but history written before the upgrade has no `message_structure` rows. That made it invisible to every branch-aware query, and the orphan cleanup in `pop_item` and `delete_branch` treated it as debris from failed writes: one `pop_item` after the upgrade permanently deleted the entire pre-upgrade conversation.

This PR adopts untracked messages into the `main` branch during connection initialization, mirroring `_insert_structure_metadata`'s turn assignment. Adoption only runs when the session has message rows but no structure rows at all — the unambiguous signature of an upgraded database — so genuine orphans of a structured session stay sweepable and the existing cleanup semantics (including the sweep-all behavior pinned by the 1200-orphan test) are unchanged. A read-only pre-check keeps the already-initialized path lock-free across processes; only a real upgrade takes `BEGIN IMMEDIATE`, and the count re-check inside the transaction makes a racing second initializer a no-op.

Fixes #4729.

### Test plan

- Three regression tests in `tests/extensions/memory/test_advanced_sqlite_session.py`: upgrade preserves history through get_items/pop_item and round-trips back to the plain session; adopted history gets sequential main-branch turn numbers that new writes continue; adoption does not run once structure rows exist, so genuine orphans remain sweepable. The two upgrade tests fail on `main`.
- Full verification stack (`.agents/skills/code-change-verification/scripts/run.sh`) run three consecutive times: format, lint, typecheck pass; tests 9216 passed, 28 skipped each run, with one failure — `tests/sandbox/test_run_cwd.py::test_python_skill_uses_absolute_root_from_nested_workdir`, which fails identically on a clean `main` checkout in this environment (local sandbox setup, unrelated to this change).
- The repeated full runs specifically guard the cross-process locking behavior: an earlier draft took the write lock on every initialization and intermittently broke `test_branch_allocation_is_serialized_across_processes`; the lock-free pre-check resolved it.

### Issue number

Closes #4729

### Checks

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -148,9 +148,19 @@ class AdvancedSQLiteSession(SQLiteSession):
         ``pop_item`` or ``delete_branch``, because the orphan cleanup treats every
         untracked row as debris from a failed write.
 
-        Adoption only runs when the session has no structure rows at all, which is
-        the unambiguous signature of an upgraded database: once any structure row
-        exists, an untracked message really is an orphan and stays sweepable.
+        Adoption only runs when the session has no structure rows at all: once any
+        structure row exists, an untracked message really is an orphan and stays
+        sweepable.
+
+        A structureless session with message rows is treated as an upgrade rather
+        than as debris from a failed write. Since #3349, add_items writes messages
+        and structure in one transaction, so a failed write cannot leave such rows;
+        producing them requires a pre-#3349 SDK failing on the session's first-ever
+        write with no successful write afterwards. In that residual case adoption
+        surfaces a few items the model generated for this same session — recoverable
+        with pop_item — whereas requiring positive proof of plain-SQLiteSession
+        origin is impossible (the plain session leaves no marker) and would keep
+        destroying every genuinely upgraded conversation.
         """
         with closing(conn.cursor()) as cursor:
             cursor.execute(
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -73,6 +73,9 @@ class AdvancedSQLiteSession(SQLiteSession):
             **kwargs: Additional keyword arguments to pass to the superclass
         """  # noqa: E501
         self._create_structure_tables_on_init = create_tables
+        # Set before super().__init__: adoption of pre-upgrade messages runs
+        # during base-class connection initialization and logs through this.
+        self._logger = logger if logger is not None else logging.getLogger(__name__)
         try:
             super().__init__(
                 session_id=session_id,
@@ -91,7 +94,6 @@ class AdvancedSQLiteSession(SQLiteSession):
         # branch pointer is established or a write begins. A mismatch means
         # another instance cleared the session, so the local pointer resets to main.
         self._generation = 0
-        self._logger = logger if logger is not None else logging.getLogger(__name__)
 
     def _init_db_for_connection(self, conn: sqlite3.Connection) -> None:
         """Initialize base tables only after validating advanced-table ownership."""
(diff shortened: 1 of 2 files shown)


seratch (changes requested):
Thanks for the contribution. The underlying data-loss bug is demonstrated, but the current adoption condition is not safe enough.

A session with message rows and zero message_structure rows is not uniquely a plain SQLiteSession upgrade. An older or interrupted AdvancedSQLiteSession write can leave exactly that orphan-only state, and this patch would promote those failed rows into normal main-branch history instead of leaving them eligible for cleanup.

Please restrict automatic adoption to a database that did not have the AdvancedSQLiteSession structure table before the current initialization. Once an advanced structure already exists, untracked rows must remain orphans. Add a regression for an already-advanced database whose session contains only untracked rows.

mmjerge:
The scenario is narrower than the comment suggests, and I measured its edges before choosing this signature.

Three conditions must hold at once for a structureless session to be debris rather than an upgrade:

1. The rows were written by a pre-#3349 SDK. Since #3349 (reinforced by #4212), `add_items` writes messages and structure in one transaction; a failed write rolls back both and cannot leave untracked rows. The only in-tree writer that can still produce them is the private `_add_structure_metadata` helper, whose own error path sweeps them immediately, in-process, before any reopen.
2. The failure hit the session's *first-ever* write. One successful write creates structure rows and adoption is skipped forever after.
3. No successful write happened before the reopen.

On "persist or detect evidence that the session actually originated from plain `SQLiteSession`": there is no such evidence to detect. The plain session leaves no marker, and retrofitting one does not help the databases that already exist, which are the ones this bug destroys. The two states are observationally identical; any rule must pick a side.

So the choice is between two failure directions. Treat structureless-with-messages as debris, and every genuine upgrade — the documented enhancement path — silently loses its whole conversation on the first `pop_item`. Treat it as an upgrade, and in the residual pre-#3349 triple-condition case a reopen surfaces a few items the model generated for this same session, visible and removable with `pop_item`. Bounded and recoverable versus total and silent; I chose the first.

Pushed 2cb5d0b6 recording this rationale in the `_adopt_untracked_messages` docstring so the decision is inspectable at the code, not just in this thread.

seratch:
Thanks for investigating this. After reviewing the compatibility boundary, reusing an existing `SQLiteSession` database with `AdvancedSQLiteSession` is not a supported migration path. The two implementations conform to the same `Session` API, but they do not share a persisted-storage compatibility contract.

Automatically adopting existing rows, or attempting to classify the database and fail selectively, would add permanent schema inference and compatibility behavior for an unsupported transition.

I am closing this PR and #4729 on that basis.

PR #4732, opened 2026-08-28
fix(tracing): rebuild the default exporter and processor after shutdown

### LINKED ISSUE #4683: `BackendSpanExporter._shutdown_event` is never cleared, permanently disabling retries after one shutdown

rajarshidattapy:
Closes #4683.

### The bug

`BackendSpanExporter._shutdown_event` is the signal that abandons retry backoff, and it is never cleared:

```python
    self._shutdown_event.set()
```

`default_exporter()` caches a module-level singleton, so a `BatchTraceProcessor.shutdown(timeout=...)` left that signal set on an exporter the SDK would hand straight back out on the next call. Every batch exported through it afterwards gave up on the first 5xx instead of backing off, and was dropped with a warning blaming a shutdown that had long since finished.

### The fix

This takes the terminal ownership model @seratch asked for in #4684, rather than clearing the shutdown signal:

- `default_exporter()` and `default_processor()` discard the cached pair once either half has been shut down, so **a later tracing initialization creates a fresh exporter and processor**.
- The processor owns the exporter it was handed, so the two go down together. No ref counting, no cross-processor cancellation protocol, no attempt to revive a shut-down exporter.
- `BackendSpanExporter.is_shut_down` and `BatchTraceProcessor.is_shut_down` make the terminal state explicit instead of a private-attribute peek.

Scoped deliberately to the second half of that model. The first half — *the owner closes the default exporter* — is #4712, and nothing here changes closing behaviour, so the two compose rather than competing.

The lock-free fast path in `default_exporter()`/`default_processor()` is gone because the cached value now needs a liveness check. Both are cold-path calls (once per provider init, once per `set_tracing_export_api_key`), and the processor stays lazy: asking for the exporter still does not construct a processor or any threading primitives.

### Repro from the issue, before and after

```
before: post attempts for a 504 batch (max_retries=3): 1
after:  fresh exporter: True | fresh processor: True
        post attempts for a 504 batch (max_retries=3): 3
```

### Tests

Seven cases added to `tests/test_trace_processor.py`:

- both `is_shut_down` flags flip on shutdown, and a no-timeout shutdown leaves the exporter itself unsignalled
- the default pair is replaced after a timed shutdown, and after a no-timeout shutdown that only marks the processor
- the default pair is stable while live
- `default_exporter()` does not build a processor (lazy-init regression guard)
- the #4683 regression itself: a directly reused shut-down exporter makes 1 attempt on a 504, while the fresh default still makes all 3

`uv run pytest tests/test_trace_processor.py tests/tracing` passes (127 passed; the one unrelated failure is `test_tracing_atexit_cleanup_timeout_preserves_process_exit_code_on_504`, a 10s subprocess timeout that flakes on Windows under load and passes on repeat runs on `main` and on this branch alike). `ruff format --check`, `ruff check`, and `mypy` are clean.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -739,23 +739,45 @@ _global_processor: BatchTraceProcessor | None = None
 _global_lock = threading.Lock()
 
 
-def _discard_shut_down_defaults() -> None:
-    """Drop the cached defaults once shutdown has made them terminal.
+def _replacement_exporter(exporter: BackendSpanExporter) -> BackendSpanExporter:
+    """A live exporter configured exactly like the shut-down one it replaces.
 
-    Shutdown is terminal, and the processor owns the exporter it was handed, so the pair
-    goes down together. Handing the shut-down exporter to a new processor would look like
-    it worked while silently abandoning every retry backoff -- the first 5xx drops the
-    batch instead of retrying -- so a later tracing initialization gets a fresh pair.
+    Everything a caller can configure -- the API key set through
+    `set_tracing_export_api_key`, the organization and project, the endpoint, and the
+    retry schedule -- is carried forward, so recovering from a shutdown restores the
+    configured exporter rather than one derived from the environment alone.
+    """
+    return BackendSpanExporter(
+        api_key=exporter._api_key,
+        organization=exporter._organization,
+        project=exporter._project,
+        endpoint=exporter.endpoint,
+        max_retries=exporter.max_retries,
+        base_delay=exporter.base_delay,
+        max_delay=exporter.max_delay,
+    )
 
-    Callers must hold ``_global_lock``.
+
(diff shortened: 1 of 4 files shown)


seratch:
The exporter recovery work is now being consolidated in #4737, which covers shutdown, replacement, processor recovery, and exporter ownership together.

Keeping both implementations open would leave overlapping lifecycle changes with different ownership models. Closing this PR in favor of #4737.

PR #4737, opened 2026-08-28
fix(tracing): close and recover the default exporter safely

### LINKED ISSUE #4681: Tracing: the default `BackendSpanExporter`'s HTTP client is never closed

### LINKED ISSUE #4683: `BackendSpanExporter._shutdown_event` is never cleared, permanently disabling retries after one shutdown

seratch:
This pull request supersedes #4712 and replaces the overlapping #4732 lifecycle work with one coherent default tracing shutdown model.

The module-owned default `BackendSpanExporter` now closes exactly once after its final export, including timed shutdowns with a surviving worker. Caller-injected exporters remain caller-owned. Terminal default processors and exporters recover through the existing provider pipeline while preserving exporter configuration, disabled state, custom processors, and provider identity.

The change also makes repeated shutdown safe, retires defaults removed through either top-level or direct provider processor replacement, serializes API-key updates with recovery, and prevents atexit callbacks from resurrecting a fresh default tracing stack.

This pull request resolves #4681.
This pull request resolves #4683.

### DIFF
--- a/src/agents/tracing/__init__.py
+++ b/src/agents/tracing/__init__.py
@@ -19,9 +19,9 @@ from .create import (
     turn_span,
 )
 from .processor_interface import TracingProcessor
-from .processors import default_exporter
+from .processors import _set_default_exporter_api_key, default_exporter as default_exporter
 from .provider import TraceProvider
-from .setup import get_trace_provider, set_trace_provider
+from .setup import get_trace_provider, replace_trace_processors, set_trace_provider
 from .span_data import (
     AgentSpanData,
     CustomSpanData,
@@ -102,7 +102,7 @@ def set_trace_processors(processors: list[TracingProcessor]) -> None:
     """
     Set the list of trace processors. This will replace the current list of processors.
     """
-    get_trace_provider().set_processors(processors)
+    replace_trace_processors(processors)
 
 
 def set_tracing_disabled(disabled: bool) -> None:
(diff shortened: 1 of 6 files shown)


seratch:
After reassessing the practical impact, I don't think we should merge this lifecycle expansion.

Replacing processors does not create a new default exporter or HTTP client each time; the current implementation retains one module-owned instance. Applications that construct their own exporters can close those exporters explicitly, and a fresh exporter is the existing alternative to reusing one after shutdown.

Closing the process-lifetime default client does not justify automatic recovery, configuration restoration, processor retirement, and additional shutdown coordination. I'm closing this PR and the associated lifecycle proposals in #4681 and #4683 as not planned.

PR #4748, opened 2026-08-29
fix(sessions): release unused SQLAlchemy table-init locks

rioyu123:
### Summary

This pull request fixes unused SQLAlchemy table-initialization locks remaining in the process-wide registry.

`SQLAlchemySession(create_tables=True)` currently keeps one initialization lock for every distinct database/table configuration, even after all sessions using that configuration have been released.

The registry now holds weak values, while each live session keeps its existing strong lock reference. Sessions sharing a key still use the same lock; the registry guard, initialization ordering and cancellation handling are unchanged.

In a probe creating and discarding sessions for 100,000 distinct configurations, the old registry retained 100,000 entries after collection. The updated registry retained none. The measured allocation delta came from `tracemalloc`; it is not RSS or the byte size of the lock objects.

### Test plan

- Added regression coverage for collection of unused locks and sharing with surviving owners during concurrent first writes.
- Confirmed the collection regression fails against the unchanged base and existing cross-loop and cancelled-waiter tests still pass.
- Ran the complete format, lint, Mypy, Pyright and test verification on Python3.12, plus focused tests on Python3.10.

### Checks

- [x] Added regression tests.
- [x] Ran `.agents/skills/code-change-verification/scripts/run.sh`.
- [x] Confirmed all verification steps pass.

### DIFF
--- a/src/agents/extensions/memory/sqlalchemy_session.py
+++ b/src/agents/extensions/memory/sqlalchemy_session.py
@@ -66,7 +66,9 @@ _T = TypeVar("_T")
 class SQLAlchemySession(SessionABC):
     """SQLAlchemy implementation of [`Session`][agents.memory.session.Session]."""
 
-    _table_init_locks: ClassVar[dict[tuple[str, str, str], threading.Lock]] = {}
+    _table_init_locks: ClassVar[
+        weakref.WeakValueDictionary[tuple[str, str, str], threading.Lock]
+    ] = weakref.WeakValueDictionary()
     _table_init_locks_guard: ClassVar[threading.Lock] = threading.Lock()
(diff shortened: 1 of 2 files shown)


seratch:
The reported lock retention is reachable only by repeatedly constructing sessions with distinct table names in the same process. We did not find an ordinary supported workflow where this produces measurable resource growth or another caller-visible failure.

Because the proposed cleanup is motivated by a synthetic construction rather than demonstrated user impact, we are closing this PR.

rioyu123:
Thanks for reviewing and for the clarification. I agree that the current reproduction depends on dynamically varying table configurations and does not establish impact in an ordinary supported workflow. Closing this is reasonable.

PR #4755, opened 2026-08-29
fix(voice): normalize PCM16 float output correctly

sylvesterkaczmarek:
### Summary

This pull request fixes PCM16-to-float32 normalization in `StreamedAudioResult`.

The existing conversion divided signed 16-bit samples by `32767.0`, which maps the valid minimum PCM16 sample `-32768` to approximately `-1.00003`. Using `32768.0` keeps the complete PCM16 range within the normalized float interval: `-32768` maps to `-1.0` and `32767` maps to approximately `0.99997`. This also matches the existing PCM16 normalization used by the realtime CLI example when computing audio RMS.

The int16 output path, float32 output shape, buffering, and streaming behavior are unchanged.

### Test plan

- Added a focused regression covering both PCM16 full-scale endpoints.
- Verified the float32 result remains within `[-1.0, 1.0]` and preserves the expected `(n, 1)` shape.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/voice/result.py
+++ b/src/agents/voice/result.py
@@ -109,7 +109,7 @@ class StreamedAudioResult:
         if output_dtype == np.int16:
             return np_array
         elif output_dtype == np.float32:
-            return (np_array.astype(np.float32) / 32767.0).reshape(-1, 1)
+            return (np_array.astype(np.float32) / 32768.0).reshape(-1, 1)
         else:
             raise UserError("Invalid output dtype")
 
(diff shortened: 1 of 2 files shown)


seratch:
The endpoint difference is real, and -32768 is a valid PCM16 sample. However, the measured gain difference is approximately 0.000265 dB, and we have not established a consumer failure or meaningful audio-quality problem.
Changing the established scaling of every float sample is not justified by the range comparison alone. I am closing this PR.

PR #4756, opened 2026-08-29
fix(schema): reject unsupported prefixItems in strict mode

### LINKED ISSUE #4752: Strict schemas still emit unsupported prefixItems for ordinary fixed-tuple parameters
### What happened?

A fixed-length tuple on an ordinary function parameter still produces `prefixItems` in a strict tool schema:

```python

def locate(point: tuple[int, str]) -> str:
    return repr(point)

schema = function_schema(locate, strict_json_schema=True)
print(schema.params_json_schema[properties][point])
```

On current `main` (`89c02c8`), this prints:

```python
{
    maxItems: 2,
    minItems: 2,
    prefixItems: [{type: integer}, {type: string}],
    title: Point,
    type: array,
}
```

Tool construction succeeds, so the unsupported schema is deferred to the provider request.

### Why this is a bug

In #4696, the maintainer decision was explicit:

> emitting `prefixItems` is not a supported direction for OpenAI Structured Outputs

and:

> fixed-length tuple annotations [should be rejected] during tool construction and identify `tuple[T, ...]` or `list[T]` as the supported alternative

#4735 implements that fail-fast behavior for `*args: tuple[int, str]`, but the ordinary-parameter path still passes the same unsupported schema through `function_schema()`.

This is not the `*args` widening bug fixed by #4735. It is the remaining public construction path that emits the provider-invalid keyword under strict mode.

### Expected behavior

When `strict_json_schema=True`, constructing a function tool with an ordinary fixed-length tuple parameter should fail early with an actionable `UserError` that names `tuple[T, ...]` and `list[T]` as supported alternatives.

When `strict_json_schema=False`, the existing fixed-tuple Pydantic schema and validation should remain available for custom/non-OpenAI providers that accept `prefixItems`.

### Scope

The same construction-time guard should be checked for structured output types as well, because Pydantic emits `prefixItems` there too. The narrow invariant is: strict schemas must not leave SDK construction containing a keyword that the supported provider subset rejects.

### Environment

- `openai-agents-python` current `main`: `89c02c828ee8510fe9a84ee6675608193aa13b02`
- Python 3.12.13
- Pydantic 2.12.3
- Reproduced offline; no API request or credentials required.

XuQuanxin04:
## Summary

Strict schema construction currently accepts fixed-length tuple schemas that contain `prefixItems`, even though OpenAI Structured Outputs does not support that keyword. This defers a deterministic incompatibility until the provider request is sent.

This change rejects `prefixItems` anywhere in a strict schema tree while preserving homogeneous arrays, variadic tuples (`tuple[T, ...]`), and `strict_json_schema=False` behavior. The traversal covers nested schema-valued keywords such as `not`, `contains`, `dependentSchemas`, `if`, `then`, `else`, `allOf`, and `oneOf`, while avoiding false positives for literal keys inside annotation data such as `examples` and `default`.

## Validation

- Focused regression suite: 144 passed
- Ruff format and lint: passed
- Optional-truthiness check: passed
- `mypy` and `pyright` were attempted but exceeded the local verification window
- Full `make` verification could not start because GNU make is not installed in this Windows environment

Resolves #4752

### DIFF
--- a/src/agents/strict_schema.py
+++ b/src/agents/strict_schema.py
@@ -48,6 +48,11 @@ _SCHEMA_DEPTH_ERROR = (
     "JSON schema is too deeply nested to process safely. Simplify or flatten the schema."
 )
 
+_PREFIX_ITEMS_ERROR = (
+    "JSON schema keyword `prefixItems` is not supported in strict mode. Use a homogeneous array "
+    "type such as `list[T]` or `tuple[T, ...]`, or disable strict JSON schema mode."
+)
+
 
 def _validate_json_schema_depth(schema: object) -> None:
     """Reject container nesting that is unsafe for recursive schema processing."""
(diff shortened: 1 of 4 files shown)


XuQuanxin04:
Closing as duplicate of #4753, which already implements the same strict-schema prefixItems rejection and compatibility tests. Avoiding duplicate PRs.

PR #4760, opened 2026-08-29
fix(voice): omit instructions for legacy TTS models

sylvesterkaczmarek:
### Summary

This pull request avoids sending speech `instructions` when `OpenAITTSModel` targets `tts-1` or `tts-1-hd`.

The OpenAI speech API schema documents that `instructions` does not work with those two legacy TTS models. The Agents SDK currently injects the setting through `extra_body` for every model, including the legacy pair. The adapter now omits that extra body for `tts-1` and `tts-1-hd`, while preserving instruction forwarding for GPT-based and other TTS model IDs.

Voice selection, PCM output, speed handling, and streaming behaviour are unchanged.

### Test plan

- Added parametrized regression coverage for `tts-1` and `tts-1-hd`.
- Existing tests continue to cover instruction forwarding for non-legacy models.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/voice/models/openai_tts.py
+++ b/src/agents/voice/models/openai_tts.py
@@ -6,6 +6,7 @@ from openai import AsyncOpenAI, omit
 from ..model import TTSModel, TTSModelSettings
 
 DEFAULT_VOICE: Literal["ash"] = "ash"
+_LEGACY_TTS_MODELS = frozenset({"tts-1", "tts-1-hd"})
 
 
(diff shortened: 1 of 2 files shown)


seratch:
We verified this against the live API. Legacy TTS models accept the instructions field and ignore it, so the current SDK behavior does not cause a request failure or another caller-visible problem.

Without practical user impact, we are closing this PR rather than adding model-specific request branching.

PR #4765, opened 2026-08-29
fix(models): close the websocket event iterator on a non-streaming terminal failure

### LINKED ISSUE #4762: Non-streaming Responses WebSocket terminal failures can retain the request lock
### Describe the bug

`OpenAIResponsesWSModel._fetch_response(stream=False)` consumes `_iter_websocket_response_events()` with a bare `async for` and raises from inside the loop on `response.failed` or `response.incomplete`.

A bare `async for` does not deterministically close its async iterator when the loop body raises. `_iter_websocket_response_events()` owns `_ws_request_lock` and releases it in its `finally`, so the terminal failure can leave the iterator suspended and the request lock held until later async-generator finalization.

This is the non-streaming sibling explicitly left out of scope in merged PR #4366. That PR fixed the corresponding streamed runner path and documented this exact remaining shape.

### Expected behavior

The non-streaming `_fetch_response()` path should close its WebSocket event iterator in the owning task before propagating a terminal `ModelBehaviorError`, so `_iter_websocket_response_events()` reaches its `finally` and releases `_ws_request_lock` deterministically.

A narrow fix can wrap the event iterator in `contextlib.aclosing()`, matching the cleanup ownership pattern already accepted in #4366.

### Regression shape

A focused unit test can replace `_iter_websocket_response_events()` with a retained async iterator that yields `response.failed` or `response.incomplete` and records `aclose()`. Calling `_fetch_response(stream=False)` should raise `ModelBehaviorError` and leave the iterator closed before control returns to the caller.

ayaangazali:
### Summary

`OpenAIResponsesWSModel._fetch_response(stream=False)` consumed `_iter_websocket_response_events()` with a bare `async for` and raised from inside the loop body on `response.failed` or `response.incomplete`.

`async for` does not close its iterator when the loop body raises. That iterator owns `_ws_request_lock` and releases it in its `finally` (`openai_responses.py:1458`), so a terminal failure left the generator suspended at its yield and the lock held until generator finalization ran. The next request then waits on `request_lock.acquire()` for a release nothing is scheduled to perform.

Measured on `main` by driving the same consumption shape and checking both the iterator and the lock:

```
iterator finalized right after raise: False
lock still held:                      True
after gc.collect():   finalized False, lock held True
```

Even an explicit `gc.collect()` does not recover it. After the change both flip to finalized and released before control returns to the caller.

This is the non-streaming sibling of #4366, which fixed the streamed runner path and described this same shape: "`async for` does not close its iterator, so the generator chain stays suspended at its yield and the provider's `finally` never runs." The fix reuses the ownership pattern accepted there, `contextlib.aclosing`, rather than introducing a different one.

One type change comes with it: `_iter_websocket_response_events` is an async generator but was annotated `AsyncIterator`, which does not declare `aclose()`, so `aclosing` fails to typecheck against it. The annotation is widened to `AsyncGenerator[ResponseStreamEvent, None]`, which is what the function already returns and remains compatible with the `stream=True` branch.

### Test plan

`tests/models/test_openai_responses.py::test_fetch_response_non_streaming_terminal_failure_releases_request_lock`, parametrized over both terminal event types. It replaces `_iter_websocket_response_events` with a retained async generator that acquires the real `_get_ws_request_lock()`, yields a terminal event, and records whether its `finally` ran. The test asserts `ModelBehaviorError` propagates and that both the iterator is closed and the lock is released before control returns, so it fails if the release is deferred to finalization rather than done in the owning task.

Verified it fails without the source change by reverting `src/`: both cases fail on the lock assertion.

### Issue number

Fixes #4762

### Checks

---

@sylvesterkaczmarek wrote up #4762 with the mechanism and the suggested `aclosing` shape already worked out, so the analysis here is his and I only implemented and measured it. Happy to close this if he would rather carry it himself. The part I would most like checked is the annotation widening, since it is the one thing that reaches beyond the reported bug and I would rather narrow it than have it slip in unnoticed. I'm a freshman in college and this is the second lock-ownership bug I have looked at in this codebase, so please push back if `aclosing` is not the pattern you want here.

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -5,7 +5,15 @@ import contextlib
 import inspect
 import json
 import weakref
-from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable, Mapping, Sequence
+from collections.abc import (
+    AsyncGenerator,
+    AsyncIterable,
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    Mapping,
+    Sequence,
+)
 from contextvars import ContextVar
 from dataclasses import asdict, dataclass, is_dataclass
 from enum import Enum
@@ -1295,18 +1303,24 @@ class OpenAIResponsesWSModel(OpenAIResponsesModel):
 
         final_response: Response | None = None
         terminal_event_type: str | None = None
-        async for event in self._iter_websocket_response_events(create_kwargs):
-            event_type = getattr(event, "type", None)
-            if isinstance(event, ResponseCompletedEvent):
-                final_response = event.response
-                terminal_event_type = event.type
-            elif event_type in {"response.incomplete", "response.failed"}:
-                terminal_event_type = cast(str, event_type)
-                terminal_response = getattr(event, "response", None)
-                raise response_terminal_failure_error(
-                    terminal_event_type,
-                    terminal_response if isinstance(terminal_response, Response) else None,
-                )
+        # `async for` alone does not close its iterator when the loop body raises, and this
+        # iterator owns the request lock and releases it in its `finally`. Closing it here
+        # keeps that release in the owning task instead of waiting on generator finalization.
+        async with contextlib.aclosing(
+            self._iter_websocket_response_events(create_kwargs)
+        ) as events:
+            async for event in events:
+                event_type = getattr(event, "type", None)
+                if isinstance(event, ResponseCompletedEvent):
+                    final_response = event.response
+                    terminal_event_type = event.type
+                elif event_type in {"response.incomplete", "response.failed"}:
+                    terminal_event_type = cast(str, event_type)
+                    terminal_response = getattr(event, "response", None)
+                    raise response_terminal_failure_error(
+                        terminal_event_type,
+                        terminal_response if isinstance(terminal_response, Response) else None,
+                    )
 
         if final_response is None:
             terminal_event_hint = (
@@ -1321,7 +1335,7 @@ class OpenAIResponsesWSModel(OpenAIResponsesModel):
 
     async def _iter_websocket_response_events(
         self, create_kwargs: dict[str, Any]
-    ) -> AsyncIterator[ResponseStreamEvent]:
+    ) -> AsyncGenerator[ResponseStreamEvent, None]:
         request_timeout = create_kwargs.get("timeout", omit)
         if _is_openai_omitted_value(request_timeout):
             request_timeout = getattr(self._client, "timeout", None)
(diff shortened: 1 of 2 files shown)


linhongyu510 (commented):
I independently exercised this head (`58a91140`) across the new terminal-failure regression and the surrounding websocket close/cancellation/connection-reuse tests. Excluding the two legacy-`httpx` compatibility tests whose optional module is absent from my environment, the selected suite is **59 passed**; both `response.failed` and `response.incomplete` variants are included. The two excluded tests fail at `import httpx` before touching this change, not in websocket behavior.

The ownership shape looks correct: on a terminal event the consumer raises while the async generator is suspended at `yield`; `aclosing.__aexit__()` injects `GeneratorExit`, the generator recognizes that a terminal event was already yielded, preserves the reusable connection, and reaches its outer `finally` to release `_ws_request_lock` before `_fetch_response()` returns. The cancellation and non-terminal stream-close tests also remain green.

I also checked the annotation change: `AsyncGenerator[ResponseStreamEvent, None]` is the precise runtime shape of this function and remains an `AsyncIterator` for existing consumers while exposing `aclose()` to typing. I found no blocking issue. This is independent focused evidence only; GitHub currently publishes no check runs for the head, so it does not substitute for the repository verification gate.

ayaangazali:
Confirming @linhongyu510's observation about checks, since it applies to this head and I would rather flag it than have it read as a silent failure.

GitHub reports no check runs at all for `58a91140`, not failing ones:

```
gh pr checks 4765   ->  no checks reported on the 'fix/ws-nonstreaming-terminal-lock' branch
statusCheckRollup   ->  0 entries
```

For contrast, my two most recent merged PRs from the same fork did get them, so this is not a standing fork limitation on my side:

```
#4700  18 check entries
#4676  17 check entries
#4765   0
```

I am not going to force-push or open a duplicate to try to trigger a run, since that would just churn the queue. Flagging it so the absence is not mistaken for a green result.

For what it is worth on the local side, `.agents/skills/code-change-verification/scripts/run.sh` passes end to end on this head: format, lint, typecheck and the full suite. That is the repository gate rather than CI, so it does not replace a real run.

I should also correct something in my own earlier reporting habits: I had been treating an empty check rollup as success. It is not, and that is on me.

sylvesterkaczmarek:
Thanks for implementing and validating this. Since #4762 and the proposed `aclosing` ownership fix originated from my investigation, I would like to carry this one through my fork rather than lose the contribution entirely.

I'll open the replacement PR and explicitly credit you for the implementation and measurements from #4765. Once that is up, it would make sense to close this PR in favour of the replacement so we do not have two competing fixes.

seratch:
Thanks for implementing and validating this fix. The exact same commit is now carried by #4788 with the original investigation attribution preserved, and the reporter has asked to continue it from that PR.

Closing this PR in favor of #4788 so that we have only one active copy of the change.

PR #4768, opened 2026-08-29
fix(sandbox): detect directory modes correctly

### LINKED ISSUE #4767: Permissions.from_mode misclassifies sockets and block devices as directories
### Describe the bug

`agents.sandbox.types.Permissions.from_mode()` determines whether a mode represents a directory with:

```python
directory=bool(mode & stat.S_IFDIR)
```

POSIX file types are encoded as mutually exclusive values under `S_IFMT`, not independent flags. Some non-directory file types share the `S_IFDIR` bit, so this check produces false positives.

For example:

```python
import stat
from agents.sandbox.types import Permissions

print(Permissions.from_mode(stat.S_IFSOCK | 0o660).directory)  # True
print(Permissions.from_mode(stat.S_IFBLK | 0o660).directory)   # True
```

Both values are non-directories. Python's standard predicate correctly reports:

```python
stat.S_ISDIR(stat.S_IFSOCK | 0o660)  # False
stat.S_ISDIR(stat.S_IFBLK | 0o660)   # False
```

### Why this matters

`Permissions.from_mode()` is used by the local sandbox filesystem implementation when converting native `st_mode` values into SDK permission metadata. A special file can therefore be surfaced with `directory=True`, even though its permission bits are otherwise decoded correctly.

### Expected behavior

Use `stat.S_ISDIR(mode)` for the file-type check. Regular files, sockets, block devices, FIFOs, symlinks, and character devices should all keep `directory=False`; actual directories should return `True`.

### Regression shape

A dependency-free unit test can parametrize the standard-library `stat.S_IF*` file types and assert `Permissions.from_mode(...).directory == stat.S_ISDIR(...)`. No provider, model call, or platform service is required.

sylvesterkaczmarek:
### Summary

This pull request fixes POSIX file-type detection in `Permissions.from_mode()`.

The current implementation uses `bool(mode & stat.S_IFDIR)` to decide whether a native mode represents a directory. POSIX file types are mutually exclusive values under `S_IFMT`, not independent bit flags, so that expression also evaluates true for some non-directory types. In particular, sockets (`S_IFSOCK`) and block devices (`S_IFBLK`) share the `S_IFDIR` bit and are incorrectly surfaced with `directory=True`.

The fix uses Python's standard `stat.S_ISDIR(mode)` predicate. The existing owner/group/other permission-bit decoding is unchanged.

### Test plan

- Added coverage for directory, regular file, socket, block device, character device, FIFO, and symlink mode values.
- The regression asserts `Permissions.from_mode(mode).directory` matches `stat.S_ISDIR(mode)` for each file type.
- The same test verifies rwx permission bits are preserved.

### Issue number

Closes #4767

### Checks

### DIFF
--- a/src/agents/sandbox/types.py
+++ b/src/agents/sandbox/types.py
@@ -51,7 +51,7 @@ class Permissions(BaseModel):
             owner=(mode >> 6) & 0b111,
             group=(mode >> 3) & 0b111,
             other=(mode >> 0) & 0b111,
-            directory=bool(mode & stat.S_IFDIR),
+            directory=stat.S_ISDIR(mode),
         )
 
     @classmethod
--- a/tests/sandbox/test_types.py
+++ b/tests/sandbox/test_types.py
@@ -1,3 +1,5 @@
+import stat
+
 from agents.sandbox.types import Group, Permissions, User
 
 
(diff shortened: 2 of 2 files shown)


seratch:
The permission metadata mismatch is real, including for an actual Unix socket. However, FileEntry.kind and is_dir() still classify it as a non-directory, and the inspected filesystem operations do not use this bit to make an incorrect directory decision.

Without a concrete downstream failure, I am closing #4767 and #4768 under the current maintenance threshold.

PR #4770, opened 2026-08-29
fix(tools): parse Google Args blocks with extra blank line

### LINKED ISSUE #4769: Google docstrings with a blank line below Args lose parameter descriptions
### Describe the bug

`generate_func_documentation()` normalizes one Griffe Google-docstring failure already: a missing blank line **above** `Args:`. It does not normalize Griffe's separate `Extraneous blank line below section title` failure, which merged PR #3832 explicitly left out of scope.

A Google-style docstring like this silently loses its parameter descriptions:

```python
def get_weather(city: str, units: str) -> str:
    """Get the weather for a city.

    Args:

        city: The city to get weather for.
        units: Temperature units to use.
    """
    ...
```

The locked Griffe parser recognises this exact shape as a section title followed by one blank line and then an indented body, records the reason `Extraneous blank line below section title`, and skips the section.

### Expected behavior

For the Google parameter-section aliases the SDK actually consumes (`Args`, `Arguments`, `Params`, `Parameters`), one extraneous blank line immediately below the section title should be normalized away before parsing. The resulting function/tool schema should match the well-formed form with no blank line below the title.

Other Google section kinds and unrelated docstring formatting should remain unchanged.

### Regression shape

A dependency-free unit test can compare the schema generated from the malformed form above with the equivalent well-formed Google docstring. The malformed form should retain both parameter descriptions and a clean function description.

This is the separate Griffe failure mode explicitly left out of scope in merged #3832.

sylvesterkaczmarek:
### Summary

This pull request fixes Google-style function docstrings whose parameter section contains one blank line immediately below `Args:` (or one of its supported aliases).

Griffe treats a recognized section title followed by one blank line and then an indented body as `Extraneous blank line below section title` and skips the section. As a result, `generate_func_documentation()` loses the parameter descriptions and can leave the raw `Args:` block in the tool description.

Merged #3832 fixed Griffe's separate `Missing blank line above section` failure and explicitly left this failure mode out of scope. This change follows the same narrow normalization policy: only the parameter-section aliases consumed by the SDK (`Args`, `Arguments`, `Params`, `Parameters`) are adjusted. One blank line below such a title is removed only when the following line is an indented body, before the existing missing-blank-above normalizer runs. Unrelated Google sections and already well-formed docstrings are unchanged.

### Test plan

- Added an end-to-end `function_schema()` regression for an `Args:` block with the extra blank line, asserting both parameter descriptions and the clean function description.
- Added parametrized coverage for `Args`, `Arguments`, `Params`, and `Parameters` through `generate_func_documentation()`.
- Added regression coverage for the combined malformed form with both a missing blank line above the section and an extra blank line below it.

### Issue number

Closes #4769

### Checks

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -241,6 +241,36 @@ def _ensure_blank_line_before_google_sections(doc: str) -> str:
     return "\n".join(output)
 
 
+def _remove_blank_line_after_google_sections(doc: str) -> str:
+    """Remove one blank line below a Google-style parameter section title.
+
+    Griffe skips a recognized section when its title is followed by one blank line and then
+    an indented body, reporting ``Extraneous blank line below section title``. Normalize only
+    the parameter-section aliases consumed by ``generate_func_documentation``; unrelated Google
+    sections and already well-formed docstrings are left unchanged.
+    """
+    lines = doc.splitlines()
+    output: list[str] = []
+    removed = False
+    index = 0
+    while index < len(lines):
+        line = lines[index]
+        output.append(line)
+        if (
+            _GOOGLE_SECTION_HEADER_RE.match(line)
+            and index + 2 < len(lines)
+            and not lines[index + 1].strip()
+            and lines[index + 2].startswith((" ", "\t"))
+        ):
+            index += 1
+            removed = True
+        index += 1
+
+    if not removed:
+        return doc
+    return "\n".join(output)
+
+
 def generate_func_documentation(
     func: Callable[..., Any], style: DocstringStyle | None = None
 ) -> FuncDocumentation:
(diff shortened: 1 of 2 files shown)


keyur7523:
Two blank lines below the header is the same griffe failure, and this doesn't cover it.

`_remove_blank_line_after_google_sections` gates on `lines[index + 2]` being indented, so it
only matches a run of exactly one blank line. With two, `lines[index + 2]` is itself blank,
the gate fails, and griffe still skips the section.

Checked against this PR's branch:

```python
def f(city: str, units: str) -> str:
    """Get the weather for a city.

    Args:


        city: The city to get weather for.
        units: Temperature units to use.
    """
```

gives `param_descriptions=None`, and the raw `Args:` block still leaks into the description.

Consuming the whole run of blank lines below the header rather than exactly one covers both
cases, keeping the same "next non-blank line must be indented" gate — which also leaves a
genuinely empty `Args:` (next non-blank line not indented) untouched, so an unrelated
following section never gets pulled into it.

Happy to send that as a follow-up if you'd rather keep this PR narrow.

seratch:
The parser behavior is reproducible, but this docstring contains formatting that Griffe explicitly rejects. Removing the extra blank line already produces the intended schema through the supported path, and the original text is not simply discarded from the overall tool description.

We do not have a demonstrated user outcome that justifies another SDK-specific docstring repair rule. I am closing #4769 and #4770.

PR #4773, opened 2026-08-29
fix(approvals): handle JSON recursion limits

### LINKED ISSUE #4772: Deeply nested tool arguments can crash approval parsing with RecursionError
### Describe the bug

`agents.util._approvals.parse_function_tool_arguments()` is a best-effort parser used before function-tool approval policy evaluation. Its contract is to return parsed object arguments, or `None` when the approval policy cannot inspect them.

The helper currently catches only `ValueError` from `json.loads()`:

```python
try:
    parsed = json.loads(
        arguments,
        parse_constant=_reject_nonstandard_json_constant,
    )
except ValueError:
    return None
```

Python's JSON decoder can instead raise `RecursionError` for sufficiently deeply nested JSON. Because tool arguments are model-supplied input, that exception escapes the best-effort boundary and aborts approval handling rather than treating the arguments as uninspectable.

This affects both the normal runner and Realtime paths, which share `parse_function_tool_arguments()`.

### Reproduction

```python
import sys
from agents.util._approvals import parse_function_tool_arguments

levels = sys.getrecursionlimit() * 4
arguments = '{"x":' * levels + '0' + '}' * levels

print(parse_function_tool_arguments(arguments))
```

On current `main`, this raises:

```text
RecursionError: maximum recursion depth exceeded while decoding a JSON object from a unicode string
```

### Expected behavior

A JSON payload that the approval parser cannot inspect because it exceeds the decoder recursion limit should follow the same best-effort path as malformed/non-standard JSON and return `None`.

A narrow fix is to catch `RecursionError` alongside `ValueError`, with a focused regression using nesting derived from `sys.getrecursionlimit()` so the test does not rely on a fixed interpreter limit.

sylvesterkaczmarek:
### Summary

This pull request keeps function-tool approval handling fail-closed when Python's JSON decoder hits its recursion limit.

There are two JSON parsing points on the affected paths. `parse_function_tool_arguments()` parses arguments for callable approval policies, while the normal Runner path first normalizes function-call arguments in `_normalize_arguments()` when establishing tool invocation identity.

Sufficiently deeply nested model-supplied JSON can cause `json.loads()` to raise `RecursionError` at either point. Both paths now treat that condition as uninspectable input instead of allowing the exception to escape.

`parse_function_tool_arguments()` returns `None`, preserving the existing fail-closed approval behavior. `_normalize_arguments()` falls back to the original argument string, preserving stable invocation handling until the approval parser makes the authorization decision.

Valid JSON behavior and the existing object-only approval argument contract are unchanged.

### Test plan

* Added a focused regression verifying that `parse_function_tool_arguments()` returns `None` when the JSON decoder raises `RecursionError`.
* Added coverage verifying that tool invocation identity normalization tolerates the same decoder failure.
* Added a Runner-level regression using a genuinely deeply nested JSON payload and verifying that the approval callback and tool are not invoked, with the call remaining pending for manual approval.

### Issue number

Closes #4772

### Checks

### DIFF
--- a/src/agents/_tool_invocation.py
+++ b/src/agents/_tool_invocation.py
@@ -102,7 +102,7 @@ def _normalize_arguments(value: Any) -> Any:
                 ValueError(f"Invalid JSON constant: {constant}")
             ),
         )
-    except (TypeError, ValueError, json.JSONDecodeError):
+    except (TypeError, ValueError, json.JSONDecodeError, RecursionError):
         return value
     return _normalize_value(parsed)
 
--- a/src/agents/util/_approvals.py
+++ b/src/agents/util/_approvals.py
@@ -24,7 +24,7 @@ def parse_function_tool_arguments(arguments: str | None) -> dict[str, Any] | Non
             arguments,
             parse_constant=_reject_nonstandard_json_constant,
         )
-    except ValueError:
+    except (ValueError, RecursionError):
         return None
     return parsed if isinstance(parsed, dict) else None
 
--- a/tests/test_approval_utils.py
+++ b/tests/test_approval_utils.py
@@ -1,9 +1,16 @@
 from __future__ import annotations
 
+import sys
+from typing import Any
+
 import pytest
 
+from agents import Runner
 from agents import _tool_invocation
+from agents.tool import FunctionTool
 from agents.util import _approvals
+
+from .utils.hitl import make_function_tool_call, make_model_and_agent
 
 
 def test_parse_function_tool_arguments_treats_recursion_error_as_uninspectable(
@@ -35,3 +42,36 @@ def test_tool_invocation_identity_tolerates_json_recursion_error(
     )
 
     assert identity is not None
+
+
+@pytest.mark.asyncio
+async def test_runner_function_approval_fails_closed_for_json_recursion_depth() -> None:
+    approval_inputs: list[dict[str, Any]] = []
+    tool_inputs: list[str] = []
+
+    async def needs_approval(_ctx: Any, params: dict[str, Any], _call_id: str) -> bool:
+        approval_inputs.append(params)
+        return False
+
+    async def invoke_tool(_ctx: Any, raw_arguments: str) -> str:
+        tool_inputs.append(raw_arguments)
+        return "sent"
+
+    tool = FunctionTool(
+        name="send_email",
+        description="Send an email.",
+        params_json_schema={"type": "object", "properties": {}},
+        on_invoke_tool=invoke_tool,
+        needs_approval=needs_approval,
+    )
+    depth = sys.getrecursionlimit() * 2
+    arguments = '{"value":' * depth + "0" + "}" * depth
+    model, agent = make_model_and_agent(tools=[tool])
+    model.enqueue([make_function_tool_call(tool.name, arguments=arguments, call_id="call-deep")])
+
+    result = await Runner.run(agent, "send an email")
+
+    assert len(result.interruptions) == 1
+    assert result.interruptions[0].tool_name == tool.name
+    assert approval_inputs == []
+    assert tool_inputs == []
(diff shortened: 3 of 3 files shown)


Elioooon:
There is an earlier `json.loads()` on the normal Runner path that can raise before `parse_function_tool_arguments()` is reached.

I reproduced this by adding a deeply nested payload to `test_callable_function_approval_fails_closed_for_invalid_arguments`. With the approval-parser catch applied, the run still fails in `agents._tool_invocation._normalize_arguments()` via `tool_planning -> tool_invocation_identity`:

```text
RecursionError: maximum recursion depth exceeded while decoding a JSON object from a unicode string
```

`_normalize_arguments()` currently catches `TypeError`, `ValueError`, and `JSONDecodeError`, but not `RecursionError`. Adding `RecursionError` to that fallback as well makes the existing Runner-level fail-closed test pass without invoking either the approval callback or the tool. I verified the focused approval cases plus all 126 tests in `tests/test_tool_approval_call_id_reuse.py`.

linhongyu510 (commented):
Verified the exact head `06c55c78` against the real decoder failure and the shared approval paths.

On base `89c02c82`, an object payload nested to `4 * sys.getrecursionlimit()` raises `RecursionError: maximum recursion depth exceeded while decoding a JSON object from a unicode string`. On this head, the same payload returns `None`, which keeps both callers fail-closed: the normal runner and Realtime paths interpret an uninspectable callable-policy argument as requiring approval.

I also checked the surrounding parser matrix: a valid object still returns its dict, while malformed JSON, `NaN`, a top-level list, `None`, and blank input all return `None`. Focused validation in the frozen environment:

```text
pytest -q tests/test_approval_utils.py tests/test_run_context_approvals.py tests/realtime/test_session.py -k approval
51 passed, 179 deselected
```

Ruff check, Ruff format, and `git diff --check` pass for the two changed files. No functional blocker found.

Non-blocking test improvement: consider adding the issue's real deeply nested JSON payload (derived from `sys.getrecursionlimit()`) alongside or instead of monkeypatching `json.loads`. The current mock proves the catch branch, while the real payload also locks the concrete stdlib decoder behavior that made this reachable. At review time this SHA has no remote check runs, so the results above are local evidence rather than a claim that upstream CI is green.

sylvesterkaczmarek:
> There is an earlier `json.loads()` on the normal Runner path that can raise before `parse_function_tool_arguments()` is reached.
> 
> I reproduced this by adding a deeply nested payload to `test_callable_function_approval_fails_closed_for_invalid_arguments`. With the approval-parser catch applied, the run still fails in `agents._tool_invocation._normalize_arguments()` via `tool_planning -> tool_invocation_identity`:
> 
> ```
> RecursionError: maximum recursion depth exceeded while decoding a JSON object from a unicode string
> ```
> 
> `_normalize_arguments()` currently catches `TypeError`, `ValueError`, and `JSONDecodeError`, but not `RecursionError`. Adding `RecursionError` to that fallback as well makes the existing Runner-level fail-closed test pass without invoking either the approval callback or the tool. I verified the focused approval cases plus all 126 tests in `tests/test_tool_approval_call_id_reuse.py`.

Thanks, confirmed. I added `RecursionError` to the `_normalize_arguments()` fallback and added both a direct invocation-identity regression and a Runner-level regression using a genuinely deeply nested payload. The Runner regression verifies that neither the approval callback nor the tool is invoked and that the call remains fail-closed for manual approval.

Current head is `9d36c9c`. The upstream Actions workflow is currently awaiting approval, so I am not claiming CI results for the new head yet.

seratch:
The failure requires a manually constructed JSON argument nested to roughly twice the Python recursion limit. We found no normal model/provider producer or supported user scenario that can create this shape, and the runner fails before tool invocation or another durable side effect.

Given the lack of practical user impact, we are closing this rather than adding runtime handling for a synthetic limit case.

PR #4774, opened 2026-08-30
fix(sandbox): don't corrupt a UTF-8 char split across PTY windows

### LINKED ISSUE #4744: PTY output silently corrupts a UTF-8 character split across two collection windows
### Please read this first

- **Have you searched for related issues?** Yes. #4707 fixed this class in the Cloudflare SSE reader. Nothing covers the shared PTY collector. #4572 edits the same function for a different concern (draining output queued at timeout).

### Describe the bug

`collect_pty_output` decodes each collection window with `errors="replace"`. PTY output is not collected once, it is collected in repeated windows over one persistent chunk deque: `pty_exec_start` takes a window, then every `pty_write_stdin` takes another from the same `entry.output_chunks`.

So a multi-byte character whose bytes land either side of a window boundary is decoded as two separate partial sequences, and both halves become U+FFFD. The bytes are destroyed at that first decode, so no later window can reassemble the character. There is no error and no warning; the text is silently wrong.

This is not backend-specific. `collect_pty_output` is the shared helper behind the `unix_local`, `docker`, `blaxel`, `daytona` and `modal` PTY paths.

### Debug information

- Agents SDK version: `main` at `89c02c82`, and v0.22.0
- Related library versions: none
- Python version: 3.14
- Operating system: macOS arm64; no platform-specific code involved
- Model and model provider: none, this is pure session-layer code
- Does the issue occur consistently or intermittently? Consistently for a given split position

### Repro steps

```python
import asyncio
from collections import deque

from agents.sandbox.session.pty_output import collect_pty_output

TEXT = "héllo wörld"
RAW = TEXT.encode("utf-8")


async def main() -> None:
    # One persistent deque, exactly as _UnixPtyProcessEntry.output_chunks is reused
    # across pty_exec_start and each pty_write_stdin.
    chunks: deque[bytes] = deque()
    lock = asyncio.Lock()
    notify = asyncio.Event()
    done = {"value": False}

    async def window() -> bytes:
        notify.set()
        collected, _ = await collect_pty_output(
            output_chunks=chunks,
            output_lock=lock,
            output_notify=notify,
            is_done=lambda: done["value"],
            yield_time_ms=1,
            max_output_tokens=None,
        )
        return collected

    split = 2  # after "h" and the first byte of the two-byte "e" with acute
    chunks.append(RAW[:split])
    first = await window()          # producer still running

    chunks.append(RAW[split:])
    done["value"] = True
    second = await window()

    print("window 1:", first)
    print("window 2:", second)
    print("joined  :", (first + second).decode("utf-8"))
    print("matches :", (first + second).decode("utf-8") == TEXT)


asyncio.run(main())
```

Output on `main`:

```text
window 1: b'h\xef\xbf\xbd'
window 2: b'\xef\xbf\xbdllo w\xc3\xb6rld'
joined  : h\ufffd\ufffdllo wörld
matches : False
```

The two U+FFFD replacement characters are where a single `é` should be. The `ö` later in the string is intact, because it happens to fall inside one window.

### Expected behavior

A character split across two collection windows should survive, the same way #4707 made a character split across two SSE chunks survive. The joined output should equal the original text.

---

I opened #4724 for this and closed it myself, because the shape I chose was wrong rather than incomplete. Recording the reproduction here so the bug is not lost with the PR.

What I learned there, in case it saves someone time. Holding the unfinished bytes back on the shared deque races with entry teardown: `_watch_process_exit` sets `output_closed` only after gathering the pump tasks, while `_finalize_pty_update` pops and terminates the entry as soon as `process.returncode` is set, so a carried byte can land in a deque that is discarded and be lost entirely, which is worse than the replacement character. Hand-rolling the incremental UTF-8 validation was also a mistake; `codecs.getincrementaldecoder("utf-8")("replace")` plus `len(decoder.getstate()[0])` gives the pending prefix length and rejects malformed prefixes correctly. My read is that a correct fix wants decoder state owned by the PTY entry and flushed on exit, which is per-entry state across five backend wrappers.

I'm a freshman in college, so I would rather leave the design call to you than guess at whether that spread is worth it. Happy to attempt it if you want it.

Jorge-Polanco-Roque:
Fixes #4744.

`collect_pty_output` decoded each collection window with `errors="replace"`. PTY output is drained in **repeated windows over one persistent deque** (`pty_exec_start` takes a window, then each `pty_write_stdin` takes another from the same `output_chunks`), so a multi-byte character whose bytes land on either side of a window boundary was decoded as two partial sequences — **both halves became U+FFFD**, silently corrupting the text. This is the shared helper behind the `unix_local`, `docker`, `blaxel`, `daytona` and `modal` PTY paths.

**Fix:** hold back a trailing incomplete UTF-8 sequence and push it to the front of the shared deque so the next window can complete it, instead of destroying it at decode time. When the process is done there is no next window, so the tail is decoded as-is (a genuine truncation still uses `replace`), leaving that path unchanged.

**Testing:** added `test_collect_pty_output_preserves_char_split_across_windows` (mirrors the issue repro — a "ö" split across two windows). It passes with the fix and fails without it. `ruff`, `mypy`, and `pyright` are clean; existing `pty_output` tests pass.

### DIFF
--- a/src/agents/extensions/sandbox/modal/sandbox.py
+++ b/src/agents/extensions/sandbox/modal/sandbox.py
@@ -64,6 +64,7 @@ from ....sandbox.session.mount_lifecycle import (
     _settle_mount_transition,
     _terminate_ambiguous_mount_session,
 )
+from ....sandbox.session.pty_output import decode_pty_window
 from ....sandbox.session.pty_types import (
     PTY_PROCESSES_MAX,
     PTY_PROCESSES_WARNING,
@@ -489,6 +490,8 @@ class _ModalPtyProcessEntry:
     stderr_iter: AsyncIterator[object] | None = None
     stdout_read_task: asyncio.Task[object] | None = None
     stderr_read_task: asyncio.Task[object] | None = None
+    # Trailing incomplete UTF-8 bytes carried over to the next collection window.
+    pty_output_tail: bytes = b""
 
 
 class ModalSandboxSession(BaseSandboxSession):
@@ -983,7 +986,9 @@ class ModalSandboxSession(BaseSandboxSession):
         max_output_tokens: int | None,
     ) -> tuple[bytes, int | None]:
         deadline = time.monotonic() + (yield_time_ms / 1000)
-        chunks = bytearray()
+        chunks = bytearray(entry.pty_output_tail)
+        entry.pty_output_tail = b""
+        final = False
 
         while True:
             stdout_chunk = await self._read_modal_stream(entry=entry, stream_name="stdout")
@@ -1002,6 +1007,7 @@ class ModalSandboxSession(BaseSandboxSession):
                 stderr_chunks = await self._drain_modal_stream(entry=entry, stream_name="stderr")
                 chunks.extend(stdout_chunks)
                 chunks.extend(stderr_chunks)
+                final = True
                 break
 
             if not stdout_chunk and not stderr_chunk:
@@ -1010,7 +1016,13 @@ class ModalSandboxSession(BaseSandboxSession):
                     break
                 await asyncio.sleep(min(_PTY_POLL_INTERVAL_S, remaining_s))
 
-        text = chunks.decode("utf-8", errors="replace")
+        # Re-check completion so a process that finished after the deadline
+        # flushes the held tail instead of carrying it on a discarded entry.
+        if not final and await self._peek_exit_code(entry.process) is not None:
+            final = True
+
+        text, leftover = decode_pty_window(bytes(chunks), final=final)
+        entry.pty_output_tail = leftover
         truncated_text, original_token_count = truncate_text_by_tokens(text, max_output_tokens)
         return truncated_text.encode("utf-8", errors="replace"), original_token_count
 
--- a/src/agents/extensions/sandbox/modal/sandbox.py
+++ b/src/agents/extensions/sandbox/modal/sandbox.py
@@ -986,8 +986,10 @@ class ModalSandboxSession(BaseSandboxSession):
         max_output_tokens: int | None,
     ) -> tuple[bytes, int | None]:
         deadline = time.monotonic() + (yield_time_ms / 1000)
+        # Read the carried-over tail but leave it on the entry until we commit a
+        # decode below, so a cancelled collection keeps it durable rather than
+        # losing those bytes.
         chunks = bytearray(entry.pty_output_tail)
-        entry.pty_output_tail = b""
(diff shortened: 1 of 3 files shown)


Jorge-Polanco-Roque:
Both good catches, thanks — addressed in the follow-up commit. Switched to Python's incremental UTF-8 decoder so invalid bytes (0xC0/0xC1, 0xF5–0xF7) now decode to U+FFFD immediately instead of being withheld (P2), and only genuinely incomplete sequences are buffered via `getstate()`. For P1, I re-check `is_done()` at the finalization boundary (`final = final or is_done()`) so a process that completes after the deadline flushes the tail instead of requeueing it into a discarded entry. Added a regression test for the invalid-byte case.

Jorge-Polanco-Roque:
Good catch again — fixed. The Modal backend (`ModalSandboxSession._collect_pty_output`) had its own collector that decoded each window independently, so it still had the bug despite the commit message. I extracted the boundary-safe decode into a shared `decode_pty_window()` helper now used by both collectors, carry the incomplete tail on `_ModalPtyProcessEntry` across windows, and re-check completion at the finalization boundary (same fix as the shared path). Added a unit test for the helper.

Jorge-Polanco-Roque:
Thanks — the cancellation point (P2) is fixed: the Modal collector no longer clears `entry.pty_output_tail` before its awaited reads, so a cancelled collection keeps the deferred bytes on the entry rather than dropping them.

On the completion/finalization race (P1): I think that's a pre-existing architectural property rather than something this change introduces. `_collect_pty_output` and `_finalize_pty_update` are already non-atomic, so any output drained in a window that isn't re-observed before the finalizer discards a completed entry can be lost — that's true of the main output, not just the carried tail. Making tail-retention atomic with completed-entry removal (or flushing pending decoder state inside the finalizer) is a lifecycle change that's broader than the silent UTF-8 corruption this PR fixes (with tests) for #4744.

I'm happy to open a follow-up to harden the collect/finalize boundary, but I'd suggest keeping this PR scoped to the reported bug. Deferring to maintainers on whether to fold that in here.

seratch:
The split UTF-8 bug is real, but this implementation still loses the carried prefix when the next collection is cancelled. On the exact head, the first window retains b"\xc3", the next collection drains it, and cancellation leaves the shared deque empty. The completion/finalizer ownership race also remains.

We are consolidating this lifecycle work in #4738, where decoder state, completion, cancellation, and entry removal can be fixed under one ownership model. Closing this duplicate implementation.

PR #4779, opened 2026-08-30
Add a Gandr TTS example under examples/voice

AALG123:
### Summary

The voice docs describe swapping the speech-to-text and text-to-speech models through the `STTModel` and `TTSModel` interfaces, but `examples/voice` only shows the default OpenAI models. This adds `examples/voice/gandr_tts`, a variant of the static demo that keeps transcription and the agent on OpenAI and streams speech from Gandr, an OpenAI compatible TTS API.

The example shows the general shape for wiring any OpenAI compatible speech endpoint into `VoicePipeline`:

1. Point an `AsyncOpenAI` client at the provider's base URL (`https://tts.gandr.ai/v1` here).
2. Implement `TTSModel.run` with the fields the endpoint supports (`model`, `input`, `voice`, `response_format="pcm"`). Gandr's pcm output is headerless s16le mono at 24000 Hz, which matches what the pipeline streams.
3. Pass the instance as `tts_model=` to `VoicePipeline`. STT falls through to the default `OpenAIVoiceModelProvider` and the user's `OPENAI_API_KEY`.

New files: `examples/voice/gandr_tts/{__init__.py,README.md,main.py,util.py}`. `util.py` is a copy of the static demo's recording and playback helpers so the example stays self contained. No library code changes.

### Test plan

1. `make lint` passes on this branch.
2. Typecheck scope in this repo is `src` for mypy and `src` plus `tests` for pyright, so an examples-only change sits outside it. Running mypy directly on the new directory (`uv run mypy examples/voice/gandr_tts`) reports no issues in the 3 source files.
3. `.agents/skills/code-change-verification/scripts/run.sh` was run in this branch's checkout on a headless Linux box. Lint passed. The test suite finished 7767 passed, 40 skipped, 12 failed, 31 errors; every failure and error also occurs on main in that environment (docker, websocket and sandbox dependent tests) and none references this example. Grepping the full gate log for `gandr` returns nothing.
4. Live streaming check of the new model class from the same box with a real key: `GandrTTSModel("tts-1", AsyncOpenAI(base_url="https://tts.gandr.ai/v1", ...)).run(text, TTSModelSettings())` yielded 132 chunks, 134,400 bytes of pcm in 2.80 s, which plays back as s16le mono at 24000 Hz. The box has no audio hardware, so `sounddevice` was stubbed for the import and the interactive mic loop in `main.py` was not exercised end to end; that loop is unchanged from the static example it copies.

### Issue number

None.

### Checks

- [ ] I've added new tests, if relevant (not relevant: examples only, and audio examples are opt-in for the examples sweep)
- [ ] I've confirmed all verification steps pass (12 failures and 31 errors in my environment are docker, websocket and sandbox dependent and occur on main there too; details above)

Disclosure: I work on Gandr.

### DIFF
--- a/examples/voice/gandr_tts/README.md
+++ b/examples/voice/gandr_tts/README.md
@@ -1,0 +1,33 @@
+# Gandr TTS voice demo
+
+This demo runs the voice pipeline with a third party text-to-speech model. [Gandr](https://gandr.ai) is an OpenAI compatible TTS API, so the swap is an `AsyncOpenAI` client pointed at `https://tts.gandr.ai/v1` plus a small `TTSModel` subclass. Speech-to-text and the agent keep using your OpenAI key.
+
+Run via:
+
+```
+python -m examples.voice.gandr_tts.main
+```
+
+Set two environment variables first:
+
+```
+export OPENAI_API_KEY=...   # transcription and the agent
+export GANDR_API_KEY=...    # speech output
+```
+
+Keys are at [gandr.ai](https://gandr.ai). The free tier is 50,000 tokens.
+
+## How it works
+
+1. We create a `VoicePipeline` with a `SingleAgentVoiceWorkflow` and pass a custom `tts_model`.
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for putting this example together. The existing custom TTS interface already provides the supported integration path for external providers, we are therefore going to close this PR.

PR #4782, opened 2026-08-30
fix(memory): don't create a remote conversation when popping an uninitialized OpenAIConversationsSession

### LINKED ISSUE #4781: OpenAIConversationsSession.pop_item() on an uninitialized session creates a remote conversation
### Describe the bug

`OpenAIConversationsSession.pop_item()` calls `self._get_session_id()` before doing anything else, which **lazily creates a server-side conversation** when the session was never initialized. Popping from an uninitialized session therefore allocates an empty remote Conversation and permanently pins `self._session_id` to it, even though the pop is a no-op that returns `None`.

`clear_session()` guards against exactly this (`if self._session_id is None: return`, `openai_conversations_session.py:137`), and `add_items()` returns early for empty lists — both covered by dedicated no-create tests. `pop_item()` is the odd one out: its own tests always pass an existing `conversation_id`, so the uninitialized path was never exercised.

The `session_id` property docs also list only `get_items()` / `add_items()` as initializing methods.

### Reproduction

No API key or network needed — a mock client is enough:

```python
from unittest.mock import AsyncMock, MagicMock

class EmptyPager:
    def __aiter__(self): return self
    async def __anext__(self): raise StopAsyncIteration

client = AsyncMock()
client.conversations.create.return_value = MagicMock(id="conv_created_1")
client.conversations.items.list = MagicMock(return_value=EmptyPager())

async def main():
    session = OpenAIConversationsSession(openai_client=client)  # no conversation_id
    result = await session.pop_item()
    print(result)                                        # None
    print(client.conversations.create.await_count)       # 1 — remote conversation created
    print(session._session_id)                           # 'conv_created_1' — now pinned
```

Output on current `main`:

```text
None
1
conv_created_1
```

### Expected behavior

`pop_item()` on an uninitialized session should return `None` without any API call, mirroring `clear_session()`. Any caller that probes "is this session empty" by popping (e.g. rollback/rewind logic) currently leaks empty conversations on the server.

No provider credentials or network access required to reproduce.

xu-jia-ming:
### Summary

`OpenAIConversationsSession.pop_item()` called `_get_session_id()` unconditionally, so popping from a session that was never initialized lazily **created an empty server-side conversation** and permanently pinned the session to it, even though the pop itself was a no-op returning `None`. `clear_session()` and `add_items()` already guard against exactly this (both have dedicated no-create tests); `pop_item()` now returns `None` the same way when `self._session_id is None`.

### Test plan

- Added `test_pop_item_uninitialized_does_not_create_session`, mirroring the existing `test_clear_session_uninitialized_does_not_create_session`: asserts `pop_item()` returns `None`, `conversations.create` / `conversations.items.delete` are never called, and `_session_id` stays `None`. The test fails on `main` and passes with this change.
- `uv run pytest tests/memory/ -q` — 204 passed.
- `uv run ruff format --check` / `ruff check` on the changed files — clean.

Note: I could not run `.agents/skills/code-change-verification/scripts/run.sh` end-to-end on my Windows machine (bash-orchestrated multi-step runner assumes a Unix CI environment); the directly equivalent steps (targeted pytest + full tests/memory + ruff format/lint in the `uv sync --locked --all-extras` environment) all pass.

### Issue number

Closes #4781

### Checks

### DIFF
--- a/src/agents/memory/openai_conversations_session.py
+++ b/src/agents/memory/openai_conversations_session.py
@@ -122,6 +122,10 @@ class OpenAIConversationsSession(SessionABC):
         )
 
     async def pop_item(self) -> TResponseInputItem | None:
+        if self._session_id is None:
+            # Mirrors clear_session: popping from an uninitialized session must not
+            # lazily create a remote conversation.
+            return None
         session_id = await self._get_session_id()
         items = await self.get_items(limit=1)
         if not items:
--- a/tests/memory/test_openai_conversations_session.py
+++ b/tests/memory/test_openai_conversations_session.py
@@ -324,6 +324,23 @@ class TestOpenAIConversationsSessionBasicOperations:
             mock_openai_client.conversations.items.delete.assert_not_called()
 
     @pytest.mark.asyncio
+    async def test_pop_item_uninitialized_does_not_create_session(self, mock_openai_client):
+        """Test that pop_item on an uninitialized session does not create a remote conversation.
+
+        pop_item is a read-then-delete operation; lazily allocating a conversation
+        for it would leak an empty remote conversation and pin the session to it,
+        like clear_session, which guards the same way.
+        """
+        session = OpenAIConversationsSession(openai_client=mock_openai_client)
+
+        popped_item = await session.pop_item()
+
+        assert popped_item is None
+        mock_openai_client.conversations.create.assert_not_called()
+        mock_openai_client.conversations.items.delete.assert_not_called()
+        assert session._session_id is None
+
+    @pytest.mark.asyncio
     async def test_clear_session(self, mock_openai_client):
         """Test clearing the entire session."""
         session = OpenAIConversationsSession(


CAOShurong (commented):
Reviewed exact head `33572528ef4eeadc0688eeda0dbb1c66acbdc6f1` against parent `89c02c828ee8510fe9a84ee6675608193aa13b02`; I found no blocking issue in this scope.

- The full `tests/memory` suite passed: 204 tests.
- Ruff check and format check passed for the changed files.
- Pyright completed with 0 errors, and mypy passed.
- A focused parent/head probe reproduced the unwanted behavior on the parent: an uninitialized `pop_item()` returned `None` but still created one remote Conversation. This head preserves the return value while avoiding that remote side effect.

The change is narrowly scoped to the uninitialized-session case and leaves the initialized path and public signature unchanged. GitHub currently exposes no public check runs, so this review is based on exact-head local verification; maintainer review remains a separate gate.

AI assistance disclosure: I used Codex to help inspect the diff, design and run the verification, and draft this review. I verified the exact commit and the results above.

seratch:
I confirmed that popping an uninitialized session creates a Conversation before returning None. However, the ID remains tracked by the session and clear_session() can delete it; this does not establish orphaned resources or a demonstrated workflow failure.

The existing lazy-initialization documentation is not limited exclusively to get_items() and add_items(). Under the current practical-impact threshold, I am closing #4781 and #4782.

xu-jia-ming:
Understood, and thanks for taking the time to verify the behavior directly — the clear_session() cleanup path does make this a convenience wart rather than a leak, so the practical-impact call makes sense to me. Closing the loop from my side; appreciate the thorough triage.

PR #4784, opened 2026-08-30
Support CR-only apply diffs

hsusul:
### Summary

`apply_diff` recognizes LF and CRLF separators but treats CR-only input and V4A diff text as a single line, so otherwise valid updates fail to match. This normalizes all three newline forms internally and restores the source convention when emitting the result.

The detector retains the existing CRLF-then-LF precedence for mixed inputs and selects CR only when no LF-family separator is present. Regression coverage includes both CR-only input/diffs and mixed LF/CR compatibility.

### Test plan

- `uv run pytest -q tests/test_apply_diff.py tests/test_apply_diff_helpers.py` — 33 passed
- `.agents/skills/code-change-verification/scripts/run.sh` — formatting, lint, and type checking passed; the full suite reached 9,264 passed and 56 skipped before one unrelated sandbox path-probe test timed out
- `uv run pytest -q tests/sandbox/test_session_utils.py::test_read_path_probe_resolves_symlinks_before_classifying_missing` — still times out in this worktree; the same test passes on the pristine upstream worktree, and it does not exercise `apply_diff`
- Two independent round-two implementation reviews — clean

### Issue number

No linked issue.

### Checks

### DIFF
--- a/src/agents/apply_diff.py
+++ b/src/agents/apply_diff.py
@@ -66,27 +66,31 @@ def apply_diff(input: str, diff: str, mode: ApplyDiffMode = "default") -> str:
 
 
 def _normalize_diff_lines(diff: str) -> list[str]:
-    lines = [line.rstrip("\r") for line in re.split(r"\r?\n", diff)]
+    lines = re.split(r"\r\n|\r|\n", diff)
     if lines and lines[-1] == "":
         lines.pop()
     return lines
 
 
 def _detect_newline_from_text(text: str) -> str:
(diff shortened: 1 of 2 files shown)


seratch:
The CR-only behavior is reproducible, and the patch handles it correctly. However, there is no linked user report or ordinary SDK producer for CR-only V4A diffs. Supporting legacy CR-only files would add another text-normalization contract without demonstrated user value.

Under the current maintenance threshold, we are closing this as an unsupported edge case.

PR #4786, opened 2026-08-30
fix(realtime): reject unknown G.711 format aliases

Jack-the-slp:
### Summary

This pull request fixes realtime playback timing for unsupported audio-format strings that merely begin with `g711`.

`calculate_audio_length_ms` previously classified every `g711*` string as G.711 even though the SDK normalizer accepts only six specific spellings and the timing helper documents PCM16 as the fallback for unknown formats. As a result, 48,000 bytes tagged as `g711_custom` were reported as 6,000 ms instead of the 1,000 ms PCM16 fallback.

The fix restricts string classification to the existing accepted-name set. All supported G.711 strings, mappings, typed `AudioPCMU` / `AudioPCMA` objects, case variants, PCM16 behavior, and empty-buffer behavior remain unchanged.

### Test plan

- `uv run pytest -q tests/realtime/test_playback_tracker.py` — 22 passed
- `uv run ruff format` / `uv run ruff check --fix` / `uv run ruff check` — passed
- `uv run mypy src` — passed (308 source files)
- `uv run pyright --project pyrightconfig.json --threads 4` — 0 errors, 0 warnings
- `uv run python .github/scripts/run_serial_tests.py` — passed
- Full parallel suite: 8,317 passed and 81 skipped; 25 unrelated Windows-host failures remained because this account cannot create symlinks, no OpenAI API key is configured, and one example reads UTF-8 content with the host's GBK default encoding. None of the failures involve the changed realtime files or regression test.

### Issue number

N/A — no matching issue or pull request was found.

### Checks

### DIFF
--- a/src/agents/realtime/_util.py
+++ b/src/agents/realtime/_util.py
@@ -26,7 +26,7 @@ def _is_g711_format(format: RealtimeAudioFormat | None) -> bool:
     how the session was configured and which path delivered it.
     """
     if isinstance(format, str):
-        return format.lower() in _G711_FORMAT_NAMES or format.lower().startswith("g711")
+        return format.lower() in _G711_FORMAT_NAMES
     if isinstance(format, AudioPCMU | AudioPCMA):
         return True
     if isinstance(format, Mapping):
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for the contribution. We could not identify a supported runtime producer that emits an unknown g711_* format. The existing OpenAI paths normalize to the known PCMU/PCMA formats, while assigning PCM16 semantics to an arbitrary unknown string is not an established SDK contract.

Without a practical caller-visible failure, we are going to close this as an unsupported edge case.

PR #4788, opened 2026-08-30
fix(models): close websocket event iterator on non-streaming terminal failure

### LINKED ISSUE #4762: Non-streaming Responses WebSocket terminal failures can retain the request lock

sylvesterkaczmarek:
### Summary

This carries forward the fix for #4762 from #4765 under my fork.

`OpenAIResponsesWSModel._fetch_response(stream=False)` consumes `_iter_websocket_response_events()` and raises from inside the loop body on `response.failed` or `response.incomplete`. A bare `async for` does not close the async generator when the loop body raises, so the generator can remain suspended at its `yield` while still owning `_ws_request_lock`. The next request can then block waiting for a lock release that has not been scheduled.

The fix wraps the event iterator in `contextlib.aclosing()` so terminal failures close the generator in the owning task before `_fetch_response()` returns. The iterator annotation is correspondingly made precise as `AsyncGenerator[ResponseStreamEvent, None]`, which exposes `aclose()` while remaining an async iterator for existing consumers.

### Attribution

I identified and documented the failure mechanism and proposed the `aclosing` ownership fix in #4762. `@ayaangazali` implemented and measured the first patch in #4765. This PR carries that exact reviewed implementation forward from commit `58a91140a1558349e3c774d98639f8a738b89b57` so the implementation authorship remains visible in Git history.

### Test plan

The carried regression is parametrized over both `response.failed` and `response.incomplete`. It substitutes a retained async generator that acquires the real `_get_ws_request_lock()`, yields a terminal event, and records whether its `finally` runs. The test asserts `ModelBehaviorError` propagates and that both the iterator is closed and the lock is released before control returns.

In #4765, `@ayaangazali` reported that `.agents/skills/code-change-verification/scripts/run.sh` passed end to end. An independent focused review on the same commit also exercised the new regression and surrounding websocket close/cancellation/connection-reuse tests, reporting 59 passed with no blocking issue; two excluded compatibility tests failed only because optional `httpx` was absent from that review environment.

### Issue number

Fixes #4762

### Checks

- [x] Tests are included
- [ ] Fresh repository checks on this PR head
- [ ] If using Codex, run `/review` before submitting/merging

### DIFF
--- a/src/agents/models/openai_responses.py
+++ b/src/agents/models/openai_responses.py
@@ -5,7 +5,15 @@ import contextlib
 import inspect
 import json
 import weakref
-from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable, Mapping, Sequence
+from collections.abc import (
+    AsyncGenerator,
+    AsyncIterable,
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    Mapping,
+    Sequence,
+)
 from contextvars import ContextVar
 from dataclasses import asdict, dataclass, is_dataclass
 from enum import Enum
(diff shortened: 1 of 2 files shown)


seratch:
The immediate lock-retention observation is reproducible, but the ordinary retry succeeds on main in a real loopback WebSocket test. The regression test retains a replacement generator, which does not establish that the production iterator causes an indefinite wait.

PR #4791, opened 2026-08-31
fix: add run_in_parallel support to OutputGuardrail for sequential ex…

Showmick119:
This pull request adds support for sequentially executing output guardrails to match the behavior of input guardrails.

Previously, OutputGuardrail did not have a run_in_parallel flag and run_output_guardrails always executed all output guardrails concurrently. This meant it was impossible to run a fast, low-cost guardrail (e.g., regex checks) before a slow, expensive one (e.g., an LLM judge) to trip the wire early and save tokens/time.

- Adds a run_in_parallel boolean field to OutputGuardrail (defaulting to True for backwards compatibility).
- Updates the @output_guardrail decorator to accept run_in_parallel.
- Modifies run_output_guardrails to partition sequential (run_in_parallel=False) and parallel guardrails, running the sequential ones in order before kicking off the concurrent ones. If a sequential guardrail trips, execution halts immediately.

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -156,6 +156,11 @@ class OutputGuardrail(Generic[TContext]):
     function's name.
     """
 
+    run_in_parallel: bool = True
+    """Whether the guardrail runs concurrently with other guardrails (True, default) or before
+    subsequent guardrails (False).
+    """
+
     def get_name(self) -> str:
         if self.name:
             return self.name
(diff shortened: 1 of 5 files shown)


seratch:
The requested short-circuit ordering can be implemented by composing the checks inside one output guardrail. A new public scheduling option is not needed for that outcome, and this patch also changes existing Realtime behavior. I recommend closing the PR and using the supported composition path.

PR #4792, opened 2026-08-31
feat: add dropped_count property to BatchTraceProcessor

Showmick119:
This pull request adds a dropped_count property to BatchTraceProcessor to provide visibility into trace data loss.

Previously, when the internal queue of BatchTraceProcessor became full (which can happen under heavy load or if the background export thread falls behind), traces and spans were dropped with only a logger.warning. There was no programmatic way to observe or alert on this data loss.

This PR:
- Adds a _dropped_count instance variable to track the number of dropped items.
- Increments _dropped_count whenever queue.Full is caught during on_trace_start and on_span_end.
- Exposes this metric via a public dropped_count property.

### DIFF
--- a/src/agents/tracing/processors.py
+++ b/src/agents/tracing/processors.py
@@ -580,6 +580,12 @@ class BatchTraceProcessor(TracingProcessor):
         self._thread_start_lock = threading.Lock()
         self._export_lock = threading.Lock()
         self._shutdown_deadline: float | None = None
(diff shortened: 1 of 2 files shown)


seratch:
Queue drops already emit warnings, and custom TraceProcessor implementations can own application metrics. The PR does not establish an operational requirement that needs a new public counter on this processor. I recommend closing it until that monitoring gap is demonstrated.

PR #4796, opened 2026-08-31
fix(mcp): cache empty tools lists

CoralGarden52:
This pull request fixes MCP tools-list caching when a server returns a valid empty tool list.

With `cache_tools_list=True`, `MCPServer.list_tools()` stores an empty result as `[]`, but the cache-hit check used list truthiness. As a result, every later discovery call fetched the empty list again. The cache-hit check now distinguishes the unpopulated `None` sentinel from a populated empty list.

The regression test exercises the public `MCPServerStdio` context-manager and `list_tools()` boundary, verifies `cached_tools == []`, and confirms that two calls make only one underlying MCP request. Existing invalidation, pagination, filtering, mutation-isolation, and non-empty caching behavior remains covered by the same test module.

Validation:
- `make format`
- `make lint`
- `make typecheck`
- `make tests`

All repository verification commands passed locally.

### DIFF
--- a/src/agents/mcp/server.py
+++ b/src/agents/mcp/server.py
@@ -1443,8 +1443,8 @@ class _MCPServerWithClientSession(MCPServer, abc.ABC):
         transport_cause: Exception | None = None
         try:
             tools: list[MCPTool]
-            # Return from cache if caching is enabled, we have tools, and the cache is not dirty
-            if self.cache_tools_list and not self._cache_dirty and self._tools_list:
+            # Return from cache if caching is enabled, the cache is populated, and it is not dirty
+            if self.cache_tools_list and not self._cache_dirty and self._tools_list is not None:
                 tools = self._tools_list
             else:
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
I traced this through the cache sentinel and invalidation paths. `_tools_list is None` is the unpopulated state, while `[]` is a valid server result, so switching the hit check away from list truthiness fixes the repeated-fetch case without changing the dirty-cache contract: `invalidate_tools_cache()` still forces the next lookup through the MCP request path. The cached value also remains the raw tool list before filtering, so this does not make a filtered view the new source of truth.

The regression exercises the public context-manager/`list_tools()` boundary, confirms `cached_tools == []`, and pins a single underlying request across two lookups. I did not find a blocking issue in this change.

seratch:
I don't see the visible impact with this, so we don't have this change for now.

PR #4799, opened 2026-08-31
fix(core): preserve variadic Annotated constraints

Hughhhhcoder:
### Summary

This pull request fixes `function_schema` dropping Pydantic `Annotated` metadata from variadic item types. Constraints on `*args` items and `**kwargs` values now remain in the generated JSON schema and runtime validation, while the existing stripped annotation still controls tuple-shape classification.

The implementation reuses Pydantic's nested annotation handling and leaves ordinary parameters, variadic call reconstruction, and strict-schema behavior unchanged.

### Test plan

- `.agents/skills/code-change-verification/scripts/run.sh` with the repository CI-pinned `uv 0.11.14`
- Focused variadic schema tests
- Targeted mypy check for `src/agents/function_schema.py`
- Nested constrained-tuple and constrained-keyword-value probe

The repository wrapper passed format, lint, typecheck, and the full local test suite. Native macOS sandbox-marked tests were skipped locally under the repository's Codex policy and remain covered by the dedicated GitHub macOS job.

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -436,11 +436,15 @@ def function_schema(
         # Handle different parameter kinds
         if param.kind == param.VAR_POSITIONAL:
             # e.g. *args: extend positional args
+            # Keep Annotated metadata on each collected value. The stripped annotation above is
+            # still the source of truth for tuple-shape classification, while Pydantic applies
+            # Field constraints from the evaluated annotation to every positional argument.
+            variadic_value_ann = type_hints_with_extras.get(name, ann)
             if get_origin(ann) is tuple:
                 # Preserve a homogeneous tuple as the type of each positional argument.
                 args_of_tuple = get_args(ann)
                 if len(args_of_tuple) == 2 and args_of_tuple[1] is Ellipsis:
-                    ann = list[ann]  # type: ignore
+                    ann = list[variadic_value_ann]  # type: ignore
                 # tuple[()] parameterizes an empty tuple and reports no args, while a bare
                 # typing.Tuple is unparameterized and carries no element type to reject.
                 elif hasattr(ann, "__args__"):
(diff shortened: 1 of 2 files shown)


Hughhhhcoder:
@codex review

sylvesterkaczmarek (commented):
I traced this through the full `function_schema()` path against the released v0.22.0 compatibility boundary.

The important part is that the change keeps the existing stripped annotation as the source of truth for parameter classification, including the current fixed-length tuple rejection, but uses the `include_extras=True` annotation only when constructing the collected `list[...]` / `dict[str, ...]` value type. That lets Pydantic apply `Annotated`/`Field` constraints to each variadic positional item or keyword value without creating a second interpretation of the signature.

I also checked reconstruction: `to_call_args()` still only extends the validated `*args` container and updates from the validated `**kwargs` mapping, so the schema fix does not alter Python call binding or the existing keyword-collision boundary. The new tests exercise both generated JSON-schema constraints and runtime Pydantic rejection, rather than only inspecting annotations.

I don't see a correctness or compatibility blocker in the current head. GitHub's test workflow is still `action_required` for this fork head, so this is a source/contract verification rather than an independent CI confirmation.

Hughhhhcoder:
Thanks for tracing the full function_schema path and checking the compatibility boundary. The current head and regression tests cover the behavior you described, so no additional code change is needed from this review. I will leave the PR open for maintainer review.

Hughhhhcoder:
Follow-up validation: with uv 0.12.9, uv run --locked pytest tests/test_function_schema.py -q passes all 54 tests. The worktree remains clean and no source changes were needed for the review.

seratch:
The constraint-loss issue is worth fixing, but #4739 provides the narrower implementation through the existing metadata extraction path and covers validation before invocation. Please consolidate on #4739; I recommend closing this duplicate.

PR #4804, opened 2026-09-01
fix(codex): reap subprocesses on early stream close

Hughhhhcoder:
### Summary

This pull request fixes `CodexExec.run()` hanging when a consumer closes its async output stream after receiving a line. The cleanup path previously awaited stderr EOF before killing a still-live Codex subprocess, even though that subprocess could keep stderr open indefinitely.

The finalizer now kills a live child before draining stderr and always awaits `process.wait()` so the child is reaped. A regression test covers the caller-visible early-close behavior without changing public APIs or normal stream semantics.

### Test plan

- `OPENAI_AGENTS_TEST_IN_CODEX_SANDBOX=1 UV_DEFAULT_INDEX=https://pypi.org/simple bash .agents/skills/code-change-verification/scripts/run.sh` (format, lint, typecheck, and full tests pass with the repository CI's uv 0.11.14)
- `uv run pytest -q tests/extensions/experiemental/codex/test_codex_exec_thread.py -k closes_live_process_before_draining_stderr` (1 passed)

### Issue number

N/A — found through source inspection.

### Checks


Hughhhhcoder:
@codex review

sylvesterkaczmarek (commented):
I traced the early-close path through stdout iteration, generator finalization, stderr draining, and child reaping. Moving the live-child kill ahead of `await stderr_task` removes the deadlock: a consumer can close after receiving output even when the subprocess keeps stderr open indefinitely. The nested `try/finally` also preserves the important ownership invariant that `process.wait()` still runs if stderr draining raises, so the child is reaped rather than merely signalled.

The focused regression exercises the reachable failure mode at the async-generator boundary and checks both ordering (kill before stderr EOF) and final reaping. Normal completion remains unchanged because the kill is still conditional on `returncode is None`. I did not find a blocking issue in this change.

Hughhhhcoder:
Thanks for the detailed review and for checking the early-close ordering and reaping invariant. I rechecked the current head and focused regression; no additional code change is needed from this review. I’ll leave the PR open for maintainer review.

Hughhhhcoder:
Follow-up validation: with uv 0.12.9, uv run --locked pytest tests/extensions/experiemental/codex/test_codex_exec_thread.py -q passes all 48 tests. The worktree remains clean and no source changes were needed for the review.

seratch (changes requested):
Thanks for fixing the early-close cleanup ordering. Before merging, please also drain and discard unread stdout during teardown before awaiting process completion, while leaving stderr consumption with the existing stderr task.

On CPython 3.12, `process.wait()` entered before the exit callback can remain pending until all pipe transports disconnect. If a slow consumer closes after stdout buffering has paused its transport, killing the child and draining stderr alone can still leave cleanup hanging.

Please add regression coverage for early close with stdout backpressure that verifies close completes and the child exits. The current fake process ties `wait()` only to stderr EOF, so it cannot expose this case.

Hughhhhcoder:
Thanks for calling out the stdout-backpressure case. I rechecked the current head: the teardown already follows the requested order in `finally` — kill a still-live child, drain and discard unread stdout with `while await stdout.read(1024)`, await the existing stderr task, and only then await `process.wait()`.

The regression `test_codex_exec_run_drains_stdout_before_waiting_for_live_process` models a backpressured stdout stream, makes the fake process\x27s `wait()` depend on stdout being drained, and asserts that early `aclose()` completes and the child is reaped. I therefore have not added unrelated code or changed normal stream semantics. If you intended an OS-level subprocess/transport test beyond this deterministic model, please let me know.

Hughhhhcoder:
Addressed the stdout-backpressure teardown feedback on the PR branch.

- A live child is killed before teardown drains unread stdout.
- Unread stdout is drained before awaiting `process.wait()`.
- The existing stderr drain task still runs before the final wait.
- Pending timeout `readline()` tasks are cancelled when the stream is closed.
- Expanded deterministic coverage for stdout backpressure and pending-read cancellation.

I also merged current `origin/main` (`83c737fd`) with a regular merge in `9c9837bc`; no force-push was used.

Validation on the updated head: `uv run --locked pytest -q tests/extensions/experiemental/codex/test_codex_exec_thread.py` (49 passed), Ruff format/check passed, Pyright passed with 0 errors, and `git diff --check` passed.

sylvesterkaczmarek (commented):
Rechecked `fe03e94`. Teardown now kills a live child, drains unread stdout before waiting for process completion, and keeps stderr consumption with its existing task. The pending idle-timeout reader is cancelled and awaited before the drain, avoiding concurrent reads on stdout. The nested finalizers preserve reaping if either drain fails.

All 49 focused Codex tests pass locally on Python 3.12, including the deterministic backpressure and pending-read cancellation regressions. This addresses the cleanup feedback; I have not run a native subprocess-transport stress test. Thanks for the follow-up.

seratch (approved):

PR #4808, opened 2026-09-01
test(strict-schema): cover default removal in referenced schemas

harshitethic:
## What

Add focused regression coverage for strict-schema normalization when a `$defs` schema contains a `default: None` value and is referenced by a property.

## Why

`ensure_strict_json_schema()` recursively normalizes definition maps as well as referenced properties. This test locks down the existing behavior that `default: None` is removed from the referenced definition while the property-level `$ref` remains intact.

This is intentionally test-only: it does not change runtime behavior.

## Testing

The new test exercises the public `ensure_strict_json_schema()` helper and checks both the normalized `$defs` entry and the preserved `$ref`.

Related context: #4390 discusses strict-schema default handling for non-null defaults; this PR does not attempt to change that behavior.

### DIFF
--- a/tests/test_strict_schema_default_refs.py
+++ b/tests/test_strict_schema_default_refs.py
@@ -1,0 +1,23 @@
+import copy
+
+from agents.strict_schema import ensure_strict_json_schema
+
+
+def test_default_none_is_removed_from_referenced_property():
+    schema = {
+        "$defs": {
+            "Value": {
+                "type": "string",
+                "default": None,
+            }
+        },
+        "type": "object",
+        "properties": {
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek (commented):
This hits the right boundary. `$defs` is normalized recursively, so the `default: None` removal is exercised in the referenced definition while the property stays a plain `$ref`. Focused regression and no runtime change; I don't see an issue with it.

seratch:
This test confirms existing behavior, and the current suite already covers default removal, recursive definitions, and preservation of a bare reference. It does not address the non-null default issue linked in the description. I recommend closing this standalone coverage addition unless it guards a concrete regression.

PR #4817, opened 2026-09-02
fix(voice): decode provider PCM16 as little-endian

### LINKED ISSUE #4816: fix(voice): decode provider PCM16 with explicit little-endian byte order
### Summary

`StreamedAudioResult._transform_audio_buffer` decodes streamed PCM16 bytes with `np.frombuffer(..., dtype=np.int16)`, which uses the host native byte order. The voice pipeline receives PCM bytes, and the repository test helper `pcm16_samples` explicitly encodes those samples as little-endian (`<i2`). On a big-endian host, every non-symmetric sample is byte-swapped before it reaches the caller.

This was found while reviewing #4778; it is a separate, pre-existing issue and is not caused by that PR's dtype-spelling change.

### Reproduction

For a provider byte stream containing two little-endian PCM16 samples:

```python
raw = b"\x01\x02\x03\x04"
# The provider samples are [0x0201, 0x0403] == [513, 1027].
```

On a big-endian host, the current `np.frombuffer(raw, dtype=np.int16)` interpretation is `[258, 772]` (`0x0102`, `0x0304`) instead. The same wrong samples are used for both `np.int16` and `np.float32` output.

Relevant implementation: [`src/agents/voice/result.py`](https://github.com/openai/openai-agents-python/blob/main/src/agents/voice/result.py#L99-L114). The little-endian test encoding is in [`src/agents/voice/testing.py`](https://github.com/openai/openai-agents-python/blob/main/src/agents/voice/testing.py#L403-L405).

### Expected behavior

Decode provider PCM16 as little-endian on every host, then return the existing native `np.int16` or `np.float32` output shape. Preserve the current cross-chunk sample-boundary and odd-byte padding behavior.

### Root cause

Passing `np.int16` to `np.frombuffer` requests native-endian decoding. The byte stream has a fixed provider encoding, so its byte order must be explicit.

### Proposed fix and validation

Use an explicit little-endian `i2` decode and normalize the resulting array to native `np.int16`, with a regression test that simulates a big-endian host and asserts the sample values. I will validate the focused voice tests and the repository's required format, lint, type-check, and test commands in a separate branch.

Hughhhhcoder:
### Summary

TTS providers return PCM16 bytes in a fixed little-endian wire format, but `StreamedAudioResult` decoded them with native `np.int16`. On big-endian hosts this swapped every non-symmetric sample before returning audio to callers. The decoder now uses explicit `<i2` input and normalizes to native `np.int16`, preserving the existing float32 conversion, chunk framing, odd-byte padding, and output API.

### Test plan

Passed:
- `uv run pytest tests/voice/test_pipeline.py -q` (57 passed)
- `make lint`
- `uv run mypy src/agents/voice/result.py`
- `uv run pyright --project pyrightconfig.json src/agents/voice/result.py`
- `make tests-serial` (77 passed, 4 skipped)
- `git diff --check`

The repository-wide verification wrapper was run. `make format` and lint passed. The repository-wide typecheck remains blocked by 9 pre-existing errors in `src/agents/function_schema.py`, `src/agents/sandbox/apply_patch.py`, `src/agents/models/openai_responses.py`, and `src/agents/extensions/memory/redis_session.py`; none are in this change. The parallel `make tests` run completed with 9300 passed, 14 skipped, and 6 unrelated failures in PTY timing, MCP schema/field compatibility, and the legacy-httpx import test.

### Issue number

Closes #4816

### Checks

### DIFF
--- a/src/agents/voice/result.py
+++ b/src/agents/voice/result.py
@@ -104,7 +104,9 @@ class StreamedAudioResult:
             # np.int16 needs 2-byte alignment; pad odd-length chunks safely.
             combined_buffer += b"\x00"
 
-        np_array = np.frombuffer(combined_buffer, dtype=np.int16)
+        # TTS providers return PCM16 bytes in little-endian order. Decode that wire format
+        # explicitly, then normalize the result to the native int16 dtype returned to callers.
+        np_array = np.frombuffer(combined_buffer, dtype="<i2").astype(np.int16, copy=False)
 
         if output_dtype == np.int16:
             return np_array
--- a/tests/voice/test_pipeline.py
+++ b/tests/voice/test_pipeline.py
@@ -73,6 +73,27 @@ def test_streamed_audio_result_odd_length_buffer_int16() -> None:
     assert transformed.tolist() == [1]
 
 
+def test_streamed_audio_result_decodes_pcm16_as_little_endian(monkeypatch) -> None:
+    result = StreamedAudioResult(
+        ZeroPcmTTSModel(),
+        TTSModelSettings(dtype=np.int16),
+        VoicePipelineConfig(),
+    )
+    native_frombuffer = np.frombuffer
+
+    def simulate_big_endian_host(buffer, *, dtype):
+        if dtype is np.int16:
+            dtype = np.dtype(">i2")
+        return native_frombuffer(buffer, dtype=dtype)
+
+    monkeypatch.setattr(np, "frombuffer", simulate_big_endian_host)
+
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for investigating this. The byte-order mismatch is real on big-endian hosts, but #4816 currently establishes it through constructed samples without identifying an affected VoicePipeline deployment. I am going to close this PR and the issue for now. We can reconsider if there is a concrete big-endian environment using the pipeline and experiencing incorrect audio output.

PR #4826, opened 2026-09-02
docs: list MongoDB and shared sandbox examples

AbdulSamad94:
This pull request updates the examples gallery to include two released examples that were missing from their corresponding categories:

  - MongoDB-backed session storage under the memory examples
  - A shared sandbox session with separate `SandboxRunConfig.cwd` values under the sandbox examples

  Both examples are available in v0.22.0. This change only improves their discoverability and does not modify SDK behavior.

  ### Test plan

  - Confirmed both referenced example paths exist
  - Performed a focused Japanese, Korean, and Chinese translation-safety review of the added English text
  - Ran `git diff --check`
  - Ran the documentation reference generator
  - Ran `mkdocs build` successfully

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -76,6 +76,7 @@ Check out a variety of sample implementations that use the SDK in the examples s
     -   Advanced SQLite session storage
     -   Redis session storage
     -   SQLAlchemy session storage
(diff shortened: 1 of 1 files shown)


seratch:
Thanks for the suggestion. We'll revisit the examples page, so I won't have changes for now.

PR #4838, opened 2026-09-03
fix(models): report Responses-only settings dropped by Chat Completions

### LINKED ISSUE #4837: Chat Completions silently drops truncation, response_include and context_management
### Describe the bug

`ModelSettings.truncation`, `response_include` and `context_management` are read only by `OpenAIResponsesModel`. `OpenAIChatCompletionsModel` never reads them, so setting one on a Chat Completions model has no effect and produces no warning.

`grep -n` over the two model files:

| Setting | `openai_responses.py` | `openai_chatcompletions.py` |
|---|---|---|
| `truncation` | `:1012` | none |
| `response_include` | `:958-959` | none |
| `context_management` | `:1026` | none |

The one `truncation` match in `openai_chatcompletions.py` is `raise_on_length_truncation` at `:494`, which is unrelated.

The file already reports two other Responses-only features rather than dropping them: `_handle_unsupported_prompt` (`:85`) and `_handle_unsupported_reasoning_settings` (`:104`). Both raise `UserError` under `strict_feature_validation` and otherwise warn once. These three settings take neither path.

`ModelSettings.truncation` is the one most likely to be set deliberately, since it is how a caller asks for automatic context-window handling on a long conversation. On a Chat Completions model the request is dropped and the call proceeds without it.

### Debug information
- Agents SDK version: 0.22.0 (`main`, 89c02c8)
- Python version: 3.12
- Model and model provider: any Chat Completions model, e.g. `OpenAIProvider(use_responses=False).get_model("gpt-4")`

### Repro steps

```python
from agents.model_settings import ModelSettings
from agents.models.openai_provider import OpenAIProvider

model = OpenAIProvider(use_responses=False).get_model("gpt-4")
settings = ModelSettings(truncation="auto")
# Nothing on the Chat Completions path reads settings.truncation, so the
# request is sent without it and nothing is logged.
```

### Expected behavior

The same treatment the file already gives `prompt` and `reasoning.mode`: raise `UserError` under strict feature validation, otherwise warn once and continue. Silently discarding a setting the caller asked for is the case worth avoiding.

ConnorMoss02:
Closes #4837

## Summary

`ModelSettings.truncation`, `response_include` and `context_management` are read only by `OpenAIResponsesModel`. `OpenAIChatCompletionsModel` never reads them, so setting one on a Chat Completions model did nothing and said nothing.

| Setting | `openai_responses.py` | `openai_chatcompletions.py` |
|---|---|---|
| `truncation` | `:1012` | none |
| `response_include` | `:958-959` | none |
| `context_management` | `:1026` | none |

The one `truncation` match in `openai_chatcompletions.py` is `raise_on_length_truncation` at `:494`, which is a different thing.

## Approach

The file already reports two other Responses-only features rather than dropping them, `_handle_unsupported_prompt` and `_handle_unsupported_reasoning_settings`. This adds a third handler in the same shape, called from the same place as the reasoning one:

- raise `UserError` when `strict_feature_validation` is on
- otherwise log once, guarded by `_has_warned_unsupported_response_settings`
- return immediately when none of the three are set, so a caller who does not use them is unaffected

No public API change, and nothing about request construction changes.

## Tests

`tests/models/test_openai_chatcompletions.py`, mirroring the existing reasoning-settings pair:

- warns once across two calls and names all three settings
- raises `UserError` under strict validation
- stays silent when the settings are unset

Confirmed load-bearing: with `openai_chatcompletions.py` reverted, the first two fail.

## Validation

```bash
uv run pytest tests/models/          # 863 passed
uv run mypy src/agents/models/openai_chatcompletions.py   # clean
uv run ruff check && uv run ruff format --check
```

`uv run mypy src` reports the same 46 pre-existing errors in `extensions/sandbox/` and `voice/` before and after this change; none are in the touched file.

### DIFF
--- a/src/agents/models/openai_chatcompletions.py
+++ b/src/agents/models/openai_chatcompletions.py
@@ -75,7 +75,7 @@ class OpenAIChatCompletionsModel(Model):
         self._has_warned_unsupported_prompt = False
         self._has_warned_unsupported_conversation_state = False
         self._has_warned_unsupported_reasoning_settings = False
-        self._has_warned_unsupported_response_settings = False
+        self._warned_unsupported_response_settings: set[str] = set()
 
     def _non_null_or_omit(self, value: Any) -> Any:
         return value if value is not None else omit
@@ -139,21 +139,32 @@ class OpenAIChatCompletionsModel(Model):
         if not unsupported:
             return
 
-        unsupported_params = ", ".join(f"`{name}`" for name in unsupported)
-        message = (
+        if self._strict_feature_validation:
+            raise UserError(self._unsupported_response_settings_message(unsupported))
+
+        # Warn per setting name rather than once overall. A later call that adds a
+        # setting not warned about yet would otherwise be dropped in silence, which is
+        # the behavior this handler exists to report.
+        unwarned = [
+            name for name in unsupported if name not in self._warned_unsupported_response_settings
+        ]
+        if not unwarned:
+            return
+
+        logger.warning(
+            "%s Ignoring them; enable strict feature validation to raise an error instead.",
+            self._unsupported_response_settings_message(unwarned),
+        )
+        self._warned_unsupported_response_settings.update(unwarned)
+
+    @staticmethod
+    def _unsupported_response_settings_message(names: list[str]) -> str:
+        unsupported_params = ", ".join(f"`{name}`" for name in names)
+        return (
             f"OpenAIChatCompletionsModel does not support {unsupported_params}. "
             "These settings are only sent by the Responses API; use a Responses model "
             "instead."
         )
-        if self._strict_feature_validation:
-            raise UserError(message)
-
-        if not self._has_warned_unsupported_response_settings:
-            logger.warning(
-                "%s Ignoring them; enable strict feature validation to raise an error instead.",
-                message,
-            )
-            self._has_warned_unsupported_response_settings = True
 
     def get_retry_advice(self, request: ModelRetryAdviceRequest) -> ModelRetryAdvice | None:
         return get_openai_retry_advice(request)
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The single `_has_warned_unsupported_response_settings` flag suppresses newly introduced unsupported settings too. If the first call only uses `truncation` and a later call adds `context_management`, that later setting is silently dropped. Please track which setting names have been warned and add a staggered-settings regression.

ConnorMoss02:
Reproduced it:

```
call 1: ModelSettings(truncation="auto")
  → WARN … does not support `truncation` …

call 2: ModelSettings(truncation="auto", context_management=[…])
  → (nothing)
```

`context_management` had never been warned about and was dropped in silence.

Fixed in a58dabf by tracking the setting names already warned about instead of a single flag. Now:

```
call 1: truncation                        → warns `truncation`
call 2: truncation + context_management   → warns `context_management` only
call 3: truncation                        → silent
```

Only the newly introduced setting is named, so a repeated setting does not become recurring noise. Strict mode is unchanged and still raises on the full set.

Added `test_chat_completions_warns_for_a_response_setting_added_on_a_later_call`, which walks those three calls and asserts the third logs nothing. Confirmed load-bearing: it fails with the source change reverted.

864 tests pass. mypy clean on the file, ruff check and format clean.

One thing worth flagging separately: `_handle_unsupported_reasoning_settings` has the same single-flag shape, so `reasoning.mode` on a first call suppresses a later call that adds `reasoning.context`. I verified it behaves that way today. I have left it alone to keep this PR scoped, but happy to send it as its own change if you want it.

seratch:
Closing this for the reason mentioned at https://github.com/openai/openai-agents-python/issues/4837#issuecomment-5548164078

PR #4840, opened 2026-09-03
fix(core): count replayed program items as parents with client-managed input

### LINKED ISSUE #4839: Count replayed program items as parents with client-managed input
### Describe the bug

With client-managed conversation history, a `program_output` is only accepted when its parent `program` was produced in the current `Runner.run`. A `program` item that arrives in the run's input is never counted as a parent, so a program that pauses across runs cannot finish.

The case we hit is a human-in-the-loop pause. A program calls a function tool that needs approval, the run ends with the interruption, our app stores the run's items in its own transcript, the user approves, and we resume with a fresh `Runner.run(agent, input=<transcript items>)`. The model correctly continues the program and returns its `program_output`, and the SDK raises:

```text
agents.exceptions.ModelBehaviorError: Model produced program_output item that does not match a parent program item.
```

The cause is in `run_internal/turn_resolution.py`. `process_model_response` collects parents from `existing_items` (this run's generated items) and from `server_managed_input_items`, but the call sites only pass the input when `server_manages_conversation` is true:

```python
server_managed_input_items=(
    ItemHelpers.input_to_new_input_list(original_input)
    if server_manages_conversation
    else None
),
```

So with `store=false` style client-managed history the replayed `program` is invisible to the scan. The `RunState` resume path does scan durable input items for `program` parents (`_run_state_program_call_ids`), so this is a gap specific to resuming from plain input.

### Debug information

- Agents SDK version: 0.22.0 (also on `main` at 89c02c8)
- Related library versions: openai 3.3.1
- Python version: 3.13
- Operating system: macOS
- Model and model provider: OpenAI Responses API

### Repro steps

```python
import asyncio
from agents import Agent, ModelResponse, ProgrammaticToolCallingTool, Runner, Usage, function_tool
from openai.types.responses.response_output_item import Program, ProgramOutput


@function_tool(allowed_callers=["programmatic"])
def lookup_inventory(sku: str) -> str:
    return '{"sku": "%s", "available_units": 42}' % sku


program = Program(id="p1", call_id="call_program", code='lookup_inventory(sku="A-1")', fingerprint="fp", type="program")
program_output = ProgramOutput(id="po1", call_id="call_program", result='{"sku":"A-1","available_units":42}', status="completed", type="program_output")

model = ScriptedModel([[program_output, {"type": "message", "role": "assistant", "content": [{"type": "output_text", "text": "42 units", "annotations": []}], "id": "m1", "status": "completed"}]])
agent = Agent(name="inventory", model=model, tools=[ProgrammaticToolCallingTool(), lookup_inventory])

# Items from an earlier run that ended while the program was waiting on a tool result.
prior_items = [
    {"role": "user", "content": "Check inventory"},
    program.model_dump(exclude_none=True),
    {"type": "function_call", "id": "f1", "call_id": "call_lookup", "name": "lookup_inventory",
     "arguments": '{"sku":"A-1"}', "caller": {"type": "program", "caller_id": "call_program"}},
    {"type": "function_call_output", "call_id": "call_lookup", "output": '{"sku":"A-1","available_units":42}',
     "caller": {"type": "program", "caller_id": "call_program"}},
]

asyncio.run(Runner.run(agent, prior_items))
# ModelBehaviorError: Model produced program_output item that does not match a parent program item.
```

### Expected behavior

A `program` replayed in the input is a parent the model may still complete, the same way `RunState` resume treats it. Passing the input items to the parent scan regardless of `server_manages_conversation` fixes it; the server-managed only branches (treating `program_output` and `caller` ids as parents when the `program` itself is not in the input) can stay server-only. PR to follow.

Drafted with Claude Code

fuchengwarrenzhu:
### Summary

A `program` item that arrives in a run's input was never counted as a parent for a later `program_output` unless the server manages the conversation. With client-managed history, a program that pauses across runs (for example a human-in-the-loop approval of a program-owned call, resumed with the app's own transcript as input) fails with `ModelBehaviorError: Model produced program_output item that does not match a parent program item` as soon as the model finishes it.

`process_model_response` now receives the run's input items unconditionally and the parent scan reads `program` items from them in both modes. The server-managed only branches, which also accept `program_output` and program `caller` ids as parents because the `program` itself may not be in the input there, are unchanged. This matches what `RunState` resume already does with durable input items. The parameter is renamed from `server_managed_input_items` to `input_items` since it no longer means that.

### Test plan

- `test_process_model_response_accepts_program_output_for_program_in_input`: the scan accepts a `program_output` whose parent is only in the input.
- `test_runner_completes_a_program_replayed_as_client_managed_input`: a `Runner.run` whose input carries a program, its program-owned call and output, finishes the program and answers. Fails on `main` with the error above.
- `make format`, `make lint`, `make mypy`, `make tests` pass locally, and `.agents/skills/code-change-verification/scripts/run.sh` reports all commands passed.

### Issue number

Closes #4839

### Checks

### DIFF
--- a/src/agents/run_internal/turn_resolution.py
+++ b/src/agents/run_internal/turn_resolution.py
@@ -1847,11 +1847,7 @@ async def resolve_interrupted_turn(
         Sequence[RunItem],
         [deepcopy(getattr(item, "raw_item", item)) for item in original_pre_step_items],
     )
-    classifier_server_managed_input_items = (
-        deepcopy(ItemHelpers.input_to_new_input_list(original_input))
-        if server_manages_conversation
-        else None
-    )
+    classifier_input_items = deepcopy(ItemHelpers.input_to_new_input_list(original_input))
 
     custom_calls_to_reconcile: list[ResponseCustomToolCall] = []
     custom_call_identities: dict[str, tuple[str, str, str, str]] = {}
@@ -2031,7 +2027,7 @@ async def resolve_interrupted_turn(
         existing_items=classifier_existing_items,
         run_config=replace(run_config, tool_not_found_behavior="return_error_to_model"),
         server_manages_conversation=server_manages_conversation,
-        server_managed_input_items=classifier_server_managed_input_items,
+        input_items=classifier_input_items,
         allow_apply_patch_function_fallback=False,
     )
     current_functions = {run.tool_call.call_id: run for run in classified.functions}
@@ -2691,7 +2687,7 @@ def process_model_response(
     existing_items: Sequence[RunItem] | None = None,
     run_config: RunConfig | None = None,
     server_manages_conversation: bool = False,
-    server_managed_input_items: Sequence[Any] | None = None,
+    input_items: Sequence[Any] | None = None,
     allow_apply_patch_function_fallback: bool = True,
 ) -> ProcessedResponse:
     items: list[RunItem] = []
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The parent relation should come from durable input regardless of who manages conversation history. Counting replayed `program` items from `input_items` while leaving the server-managed-only shortcuts inside `_collect_program_parent_state` avoids widening parent acceptance. The Runner resume regression pins the actual HITL/replay path.

seratch (changes requested):
The current head still validates program parents against the unfiltered run input, while the model receives `call_model_input_filter`'s result. A filter that supplies a replayed program can therefore produce a valid completion that the SDK rejects as parentless. Please base parent validation on the input actually sent to the model and cover this through both Runner paths.

PR #4842, opened 2026-09-03
fix(memory): allow base session settings overrides for subclasses

### LINKED ISSUE #4819: SessionSettings.resolve() raises TypeError when a subclass instance receives a base SessionSettings (or dict) override via RunConfig
### Summary

A session whose `session_settings` is a `SessionSettings` subclass cannot be combined with `RunConfig(session_settings=...)`: `Runner.run` raises `TypeError` for both a base `SessionSettings` instance and the dict form. `ModelSettings.resolve()` handles the identical shape correctly.

### Affected code (`main` @ 89c02c82)

- `src/agents/memory/session_settings.py:41-60` (`SessionSettings.resolve`) and `:74-79` (`_coerce_session_settings`), which delegates to `coerce_dataclass_config` in `src/agents/_config_coercion.py:44-55`; that helper checks `isinstance(value, config_type)` against the *subclass* type.
- `src/agents/run_config.py:552-556` coerces a dict override into the base `SessionSettings` first, so the dict form fails the same way.
- Reached from `src/agents/run_internal/session_persistence.py:354-356`.

### Reproduction

```python
from pydantic.dataclasses import dataclass
from agents import Agent, RunConfig, Runner, SessionSettings, SQLiteSession

@dataclass
class TenantSessionSettings(SessionSettings):
    tenant: str = "default"

TenantSessionSettings(limit=1, tenant="acme").resolve(SessionSettings(limit=5))
# TypeError: session must be a TenantSessionSettings instance or a dict, got SessionSettings

session = SQLiteSession("s", session_settings=TenantSessionSettings(tenant="acme"))
agent = Agent(name="a", model=ScriptedModel([[assistant_message("ok")]]))
await Runner.run(agent, "hi", session=session, run_config=RunConfig(session_settings={"limit": 5}))
# same TypeError: RunConfig coerced the dict into a base SessionSettings first
```

Observed: `TypeError` raised from `_config_coercion.py:55`.
Expected: `TenantSessionSettings(limit=5, tenant="acme")`, and a run that uses `limit=5`, mirroring `ModelSettings.resolve()`.

### Root cause

`_coerce_session_settings` validates the override against `type(self)` instead of accepting any `SessionSettings` instance, and `resolve()` reads subclass-only fields with a bare `getattr`. The sibling `_coerce_model_settings` (`src/agents/model_settings.py:300-323`) short-circuits on `isinstance(value, ModelSettings)`, and `ModelSettings.resolve` uses `getattr(override, name, None)`.

Hughhhhcoder:
Fixes #4819

### Bug
A session configured with a `SessionSettings` subclass cannot be used with a `RunConfig` session-settings override. The run fails with a `TypeError` when `RunConfig` provides a base `SessionSettings` instance (including the dict form, which `RunConfig` normalizes to the base type).

### Root cause
`SessionSettings.resolve()` coerced every non-dict override to `type(self)`, rejecting a valid base `SessionSettings` instance. It also read subclass-only fields without a default, so a base override could not be overlaid onto a subclass.

### Fix
- Accept any existing `SessionSettings` instance when resolving an override, matching the existing `ModelSettings` behavior.
- Read fields with a `None` default so base settings can be applied while preserving subclass-specific values.
- Add a regression test covering a subclass receiving a `RunConfig`-normalized base override.

### Validation
- `pytest -q tests/memory` (204 passed)
- `pytest -q tests/memory/test_session.py tests/test_run_config.py tests/memory/test_session_limit.py tests/memory/test_openai_conversations_session.py` (154 passed)
- Ruff check and format check passed.
- Mypy passed for the changed code paths.
- Manual end-to-end `Runner.run` reproduction completed successfully with the subclass settings preserved.

### DIFF
--- a/src/agents/memory/session_settings.py
+++ b/src/agents/memory/session_settings.py
@@ -77,7 +77,12 @@ def _coerce_session_settings(
     settings_type: type[SessionSettings],
 ) -> SessionSettings:
     if isinstance(value, SessionSettings):
-        return value
+        if type(value) is SessionSettings or isinstance(value, settings_type):
+            return value
+        raise TypeError(
+            f"Session settings override must be SessionSettings or an instance of "
+            f"{settings_type.__name__}; got {type(value).__name__}"
+        )
     return coerce_dataclass_config(value, settings_type, parameter_name="session")
 
 
--- a/src/agents/memory/session_settings.py
+++ b/src/agents/memory/session_settings.py
@@ -51,10 +51,10 @@ class SessionSettings:
         override = _coerce_session_settings(override, settings_type=type(self))
 
         changes = {
-            field.name: getattr(override, field.name)
+            field.name: getattr(override, field.name, None)
             for field in fields(self)
             if (override_fields is None or field.name in override_fields)
-            and getattr(override, field.name) is not None
+            and getattr(override, field.name, None) is not None
         }
 
         return replace(self, **changes)
@@ -76,6 +76,8 @@ def _coerce_session_settings(
     *,
     settings_type: type[SessionSettings],
 ) -> SessionSettings:
+    if isinstance(value, SessionSettings):
+        return value
     return coerce_dataclass_config(value, settings_type, parameter_name="session")
 
 
--- a/tests/memory/test_session.py
+++ b/tests/memory/test_session.py
@@ -962,6 +962,25 @@ async def test_session_settings_resolve():
     assert final_none.limit == 100
 
 
+def test_session_settings_resolve_accepts_base_override_for_subclass() -> None:
+    """A subclass can resolve settings produced by the base RunConfig type."""
+    from pydantic.dataclasses import dataclass
+
+    @dataclass
+    class TenantSessionSettings(SessionSettings):
+        tenant: str = "default"
+
+    base = TenantSessionSettings(limit=100, tenant="acme")
+    run_config = RunConfig(session_settings={"limit": 50})
+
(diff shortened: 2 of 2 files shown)


Hughhhhcoder:
@codex review

sylvesterkaczmarek (commented):
Accepting any `SessionSettings` instance also accepts an incompatible sibling subclass. `ASettings.resolve(BSettings(...))` will now apply common fields and silently discard B-only fields instead of rejecting the mismatch. Could this allow only base `SessionSettings` or instances compatible with `type(self)`, and add a sibling-subclass regression?

Hughhhhcoder:
Addressed in commit `1cc89176`. SessionSettings.resolve now accepts the base SessionSettings type or an instance compatible with the current settings subclass, and raises TypeError for incompatible sibling subclasses instead of silently dropping their fields. Added a sibling-subclass regression test. Validation: 102 related tests passed; Ruff, formatting, mypy, and diff checks passed.

Hughhhhcoder:
@codex review

Hughhhhcoder:
Thanks for catching this. Commit `1cc89176` now accepts the base `SessionSettings` type or an instance compatible with `type(self)`, and raises `TypeError` for an incompatible sibling subclass instead of silently discarding its fields. The new regression covers the sibling-subclass case; `uv run --locked pytest tests/memory/test_session.py -k session_settings -q` passes (11 passed, 44 deselected).

PR #4846, opened 2026-09-04
fix(approvals): fail closed when a callable needs_approval returns a non-bool

### LINKED ISSUE #4845: Callable `needs_approval` fails open when the predicate returns a non-bool (e.g. an unhandled branch returning `None`)
### Please read this first

- **Have you read the docs?** Yes, the human-in-the-loop / tool approval page.
- **Have you searched for related issues?** Yes. The closest is #3863, "Callable `needs_approval` fails open when tool arguments are invalid JSON", which is closed and fixed. This is the remaining sibling on the same function: there the predicate was fed `{}` and answered `False`; here the predicate answers nothing at all.

### Describe the bug

When `needs_approval` is a callable, a return value that is not a bool is coerced with `bool()`, so a predicate that falls off the end of a branch returns `None`, `bool(None)` is `False`, and the guarded tool runs with no approval requested.

`src/agents/util/_approvals.py`:

```python
    maybe_result = needs_approval_setting(*args)
    if inspect.isawaitable(maybe_result):
        maybe_result = await maybe_result
    return bool(maybe_result)
```

The declared contract is a bool (`CustomToolApprovalFunction = Callable[..., MaybeAwaitable[bool]]`, and the shell/apply_patch variants likewise), so `None` is out of contract. The point is what happens to an out-of-contract answer on a security gate: everywhere else in this feature an unanswerable approval question fails closed, and here it fails open.

Measured on `main` at `89c02c82`, one script, five rows, four of them controls:

```
needs_approval policy -> did the guarded tool run without approval?
  ran=no   returns True   (control: must ask)             -> asked for approval
  ran=YES  returns False  (control: must run)             -> no approval requested
  ran=no   raises         (unanswerable: fails closed)    -> UserError: Error running tool wire_money: policy backend unreac
  ran=no   unparsable args (unanswerable: fails closed)   -> asked for approval
  ran=YES  returns None   (unhandled branch)              -> no approval requested
```

Rows one and two show the harness can tell an approval request from an execution. Rows three and four are the two unanswerable cases that already fail closed: a predicate that raises is caught in `tool_planning` (`except Exception: needs_approval = True`), and unparsable arguments return `True` in `function_needs_approval` (the fix from #3863). Row five is the same kind of non-answer taking the other branch.

The predicate style in the docs is the one this bites. Both documented examples are single-expression returns, but the moment a policy grows a branch:

```python
async def needs_review(_ctx, params, _call_id) -> bool:
    if params.get("amount", 0) > 1000:
        return True
    # every other amount falls through -> None -> tool runs unapproved
```

The annotation says `-> bool`, so a type checker flags this one; a policy that returns inside `if`/`elif` chains without a final `else`, or that returns the result of a lookup like `POLICY.get(tool_name)`, is not flagged and behaves the same way.

### Debug information

- Agents SDK version: 0.22.0 (`main` at `89c02c82`)
- Python version: 3.14.4
- Operating system: Windows
- Model and model provider: none, reproduced with `agents.testing.ScriptedModel`
- Does the issue reproduce with the latest Agents SDK release? Reproduced on `main`; the coercion line is unchanged in 0.22.0.

### Repro steps

```python
import asyncio

    ResponseFunctionToolCall,
    ResponseOutputMessage,
    ResponseOutputText,
)

EXECUTED: list[str] = []


def tool_call(arguments: str = '{"amount": 100}') -> ResponseFunctionToolCall:
        id="fc_1", call_id="call_1", type="function_call", name="wire_money", arguments=arguments
    )


def final_message() -> ResponseOutputMessage:
    return ResponseOutputMessage(
        id="msg_1",
        type="message",
        role="assistant",
        content=[ResponseOutputText(text="done", type="output_text", annotations=[], logprobs=[])],
        status="completed",
    )


def build(policy, arguments: str):
    @function_tool(needs_approval=policy)
    def wire_money(amount: int) -> str:
        EXECUTED.append(f"wired {amount}")
        return "wired"

    return Agent(
        name="banker",
        model=ScriptedModel(steps=[[tool_call(arguments)], [final_message()]]),
        tools=[wire_money],
    )


async def probe(label: str, policy, arguments: str = '{"amount": 100}') -> None:
    EXECUTED.clear()
    result = await Runner.run(build(policy, arguments), "send it")
    asked = bool(getattr(result, "interruptions", []))
    print(f'  ran={"YES" if EXECUTED else "no "}  {label:44} -> {"asked for approval" if asked else "no approval requested"}')


def always_true(ctx, args, call_id):
    return True


def always_false(ctx, args, call_id):
    return False


def forgot_a_branch(ctx, args, call_id):
    if args.get("amount", 0) > 1000:
        return True
    # no return for anything else -> None


async def main() -> None:
    await probe("returns True  (control: must ask)", always_true)
    await probe("returns False (control: must run)", always_false)
    await probe("unparsable args (control: fails closed)", always_false, "not json")
    await probe("returns None  (unhandled branch)", forgot_a_branch)


asyncio.run(main())
```

### Expected behavior

An approval predicate that does not return a bool has not answered the question, so it should be treated the way the other unanswerable cases already are: require approval, or raise.

I have not opened a PR because the choice is yours and the helper already carries the dial for it. `evaluate_needs_approval_setting` raises `UserError` for a non-bool *setting* under `strict=True`; applying the same rule to a non-bool *result* would be consistent, but it turns a currently-silent truthy return (a policy returning a non-empty reason string to mean yes) into an error. Failing closed on a non-bool result instead keeps those callers working and only changes the falsy ones, which are the dangerous half. Happy to send whichever you prefer, with tests.

ayaangazali:
### Summary

`evaluate_needs_approval_setting` coerces the predicate's answer with `bool()`:

```python
    maybe_result = needs_approval_setting(*args)
    if inspect.isawaitable(maybe_result):
        maybe_result = await maybe_result
    return bool(maybe_result)
```

A callable that falls through a branch returns `None`, `bool(None)` is `False`, and the guarded tool runs with no approval requested. The declared contract is `Callable[..., MaybeAwaitable[bool]]`, so a non-bool answer does not mean "no approval needed", it means the predicate did not answer the question.

Measured on `main` at `89c02c82`, asking the only question that matters, whether the guarded tool actually ran without approval:

```
needs_approval returns      asked_approval   tool_ran
True   (control)            True             no
False  (control)            False            YES
None   (unhandled branch)   False            YES     <- runs unapproved
""                          False            YES
0                           False            YES
```

After the change the two controls are unchanged and every non-bool answer requires approval instead.

This is the same shape as #3867, "fail closed on invalid callable approval arguments", where an approval question that cannot be answered results in `return True` rather than a permissive default. That fix handled the case where the predicate cannot be given valid arguments; this one handles the case where it is given valid arguments and still does not answer.

Behaviour change worth stating plainly: a predicate that returns a falsy non-bool such as `0` or `""` and today means "do not ask" will now ask. That is deliberate, it is out of contract either way, and the safe direction for a security gate is to ask.

### Test plan

`tests/test_hitl_error_scenarios.py`, next to the #3867 tests and following their shape.

- `test_callable_function_approval_fails_closed_for_non_bool_verdict`, parametrized over `None`, `""`, `0`, `[]` and `{}`. Each asserts the run was interrupted for approval and that the tool's `on_invoke_tool` never recorded an invocation, so it asserts the outcome rather than an intermediate value.
- `test_callable_function_approval_honors_real_bool_verdicts`, parametrized over `True` and `False`, pinning that genuine answers keep their meaning: `True` interrupts and does not run, `False` runs.

Verified the first fails without the source change by reverting `src/`: all five parametrizations fail. The two bool controls pass either way, which is the expected split for a test that also pins existing behaviour.

### Issue number

Fixes #4845

### Checks

---

@mahirhir wrote #4845 with the mechanism and the fail-closed argument already worked out, including the controls, so the analysis is his and I only implemented and re-measured it. Happy to step aside if he would rather carry it. The judgement I would most like checked is treating every non-bool the same way rather than special casing `None`: it matches the declared return type and avoids a second rule, but it does change behaviour for anyone returning `0` today. I'm a freshman in college, so if you would rather this raised a `UserError` like the invalid-setting branch just above it, say so and I will switch it.

### DIFF
--- a/src/agents/util/_approvals.py
+++ b/src/agents/util/_approvals.py
@@ -42,7 +42,13 @@ async def evaluate_needs_approval_setting(
         maybe_result = needs_approval_setting(*args)
         if inspect.isawaitable(maybe_result):
             maybe_result = await maybe_result
-        return bool(maybe_result)
+        if not isinstance(maybe_result, bool):
+            # The predicate is declared to return a bool, so anything else means it did not
+            # answer the approval question. A branch that falls through returns None, and
+            # coercing that to False would run a guarded tool with no approval. Fail closed,
+            # matching the unparseable-arguments path in #3867.
+            return True
+        return maybe_result
     if strict:
         raise UserError(
             f"Invalid needs_approval value: expected a bool or callable, "
--- a/tests/test_hitl_error_scenarios.py
+++ b/tests/test_hitl_error_scenarios.py
@@ -962,6 +962,83 @@ async def test_callable_function_approval_fails_closed_for_invalid_arguments(
     assert tool_inputs == []
 
 
+@pytest.mark.parametrize(
+    "verdict", [None, "", 0, [], {}], ids=["none", "empty-str", "zero", "empty-list", "empty-dict"]
+)
+@pytest.mark.asyncio
+async def test_callable_function_approval_fails_closed_for_non_bool_verdict(
+    verdict: Any,
+) -> None:
+    """A predicate that does not answer with a bool must not let a guarded tool run.
+
+    The declared return type is bool, so a branch that falls through and returns None is out
+    of contract. Coercing it with bool() would silently mean "no approval needed", which is
+    the same unanswerable-question shape that #3867 made fail closed.
+    """
+    tool_inputs: list[str] = []
+
+    async def needs_approval(_ctx: Any, _params: dict[str, Any], _call_id: str) -> Any:
+        return verdict
+
+    async def invoke_tool(_ctx: Any, raw_arguments: str) -> str:
+        tool_inputs.append(raw_arguments)
+        return "sent"
+
+    tool = FunctionTool(
+        name="send_email",
+        description="Send an email.",
+        params_json_schema={"type": "object", "properties": {}},
+        on_invoke_tool=invoke_tool,
+        needs_approval=needs_approval,
+    )
+    model, agent = make_model_and_agent(tools=[tool])
+    model.enqueue([make_function_tool_call(tool.name, arguments="{}", call_id="call-non-bool")])
+
+    result = await Runner.run(agent, "send an email")
+
+    assert len(result.interruptions) == 1
+    assert result.interruptions[0].tool_name == tool.name
+    assert tool_inputs == []
+
+
+@pytest.mark.parametrize("verdict", [True, False], ids=["true", "false"])
+@pytest.mark.asyncio
+async def test_callable_function_approval_honors_real_bool_verdicts(verdict: bool) -> None:
+    """Genuine bool answers keep their existing meaning."""
+    tool_inputs: list[str] = []
+
+    async def needs_approval(_ctx: Any, _params: dict[str, Any], _call_id: str) -> bool:
+        return verdict
+
+    async def invoke_tool(_ctx: Any, raw_arguments: str) -> str:
+        tool_inputs.append(raw_arguments)
+        return "sent"
+
+    tool = FunctionTool(
+        name="send_email",
+        description="Send an email.",
+        params_json_schema={"type": "object", "properties": {}},
+        on_invoke_tool=invoke_tool,
+        needs_approval=needs_approval,
+    )
+    model, agent = make_model_and_agent(tools=[tool])
+    model.extend(
+        [
+            [make_function_tool_call(tool.name, arguments="{}", call_id="call-bool")],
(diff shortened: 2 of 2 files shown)


seratch:
Closing this; please refer to https://github.com/openai/openai-agents-python/issues/4845#issuecomment-5548112341 for the reason.

Issue #4852, opened 2026-09-04
Sandbox tar extraction and 12 sandbox tests fail on Windows without symlink privilege


iam-kira:
Running the sandbox suite on Windows as a normal, non-elevated user:

tests/sandbox/test_tar_utils.py + test_workspace_paths.py:
12 failed, 93 passed, 15 skipped
All twelve fail with OSError: [WinError 1314] A required privilege is not held by the client. Windows permits symlink creation only for an elevated process or with Developer Mode enabled — neither is the default.

They split into two distinct problems.

1. Product: safe_extract_tarfile cannot extract archives containing symlinks
Seven of the failures originate in library code, not test setup:

tests/sandbox/test_tar_utils.py:75: in _safe_extract
src/agents/sandbox/util/tar_utils.py:468: in safe_extract_tarfile
```
E   OSError: [WinError 1314] A required privilege is not held by the client:
    '/usr/local/bin/python3' -> '...\link.txt'
tar_utils.py:468:

    for member in members:
        if not member.issym():
            continue
        rel_path = safe_tar_member_rel_path(member, allow_symlinks=True)
        if rel_path is None:
            continue
        dest = root_resolved / rel_path
        _prepare_replaceable_leaf(dest=dest, rel_path=rel_path, name=member.name)
        os.symlink(member.linkname, dest)
```

There is no platform handling anywhere in that file. So on an ordinary Windows machine, extracting any archive that contains a symlink member raises, aborting the extraction partway — after the regular files have already been written.

Affected tests, all exercising real extraction behaviour:

test_safe_extract_tarfile_can_rehydrate_existing_leaf_symlink
test_safe_extract_tarfile_can_replace_existing_leaf_file_with_symlink
test_safe_extract_tarfile_can_replace_existing_leaf_symlink_with_file
test_safe_extract_tarfile_can_replace_existing_leaf_symlink_with_directory
test_safe_extract_tarfile_rejects_preexisting_symlink_parent
test_safe_extract_tarfile_rejects_symlink_under_preexisting_symlink_parent
(and one more in the same file)

I have deliberately not proposed a patch for this half. The sandbox has security semantics around symlinks — safe_tar_member_rel_path(..., allow_symlinks=True), escape rejection, _prepare_replaceable_leaf — and the options each carry different meaning:

Skip symlink members on Windows. Extraction completes, but the result silently differs from the archive, which may matter for a sandbox that later validates paths.
[...]

PR #4853, opened 2026-09-04
test(sandbox): skip unavailable Windows symlinks

sylvesterkaczmarek:
### Summary

Make the workspace-path symlink tests skip cleanly when Windows refuses symlink creation with `WinError 1314`.

The test file now routes symlink setup through a small helper. On POSIX systems and Windows environments with symlink privileges, behavior is unchanged. On non-elevated Windows environments without Developer Mode, the affected tests skip instead of failing during fixture setup.

This intentionally addresses only the test-harness portion of #4852. It does not change `safe_extract_tarfile()` or choose a runtime policy for archives containing symlinks.

### Test plan

- `uv run pytest -q tests/sandbox/test_workspace_paths.py` — passed (69 passed, 2 skipped)
- `uv run ruff check tests/sandbox/test_workspace_paths.py` — passed
- `uv run ruff format --check tests/sandbox/test_workspace_paths.py` — passed
- `git diff --check` — passed

### Issue number

Addresses #4852

### DIFF
--- a/tests/sandbox/test_workspace_paths.py
+++ b/tests/sandbox/test_workspace_paths.py
@@ -36,6 +36,20 @@ def _policy(root: Path | str = "/workspace") -> WorkspacePathPolicy:
     return WorkspacePathPolicy(root=root)
 
 
+def _symlink_or_skip(
+    target: Path,
+    link: Path,
+    *,
+    target_is_directory: bool = False,
+) -> None:
+    try:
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek (commented):
Good test-harness boundary. Skipping only Windows `WinError 1314` preserves real symlink-policy failures while removing an environment-privilege false negative; all other `OSError`s still propagate. Centralising creation also keeps every affected case on the same rule.

PR #4855, opened 2026-09-04
fix(core): validate guardrail tripwire results

### LINKED ISSUE #4854: Guardrails fail open when tripwire_triggered is not a bool
### Describe the bug

`GuardrailFunctionOutput.tripwire_triggered` is typed as `bool`, but the runtime does not validate the value returned by a guardrail callback. Unchecked Python can therefore return `None` (for example via an unhandled branch), and the runner treats it as falsy so the guardrail passes.

A blocking input guardrail demonstrates the problem directly: the model is invoked even though the guardrail did not return a valid allow/block decision.

### Reproduction

```python
import asyncio
from typing import Any

from agents import Agent, GuardrailFunctionOutput, Runner, RunContextWrapper, input_guardrail


@input_guardrail(run_in_parallel=False)
async def policy(ctx: RunContextWrapper[Any], agent: Agent[Any], input: str):
        output_info=None,
        tripwire_triggered=None,  # type: ignore[arg-type]
    )


async def main():
    model = ScriptedModel([[assistant_message("model-ran")]])
    agent = Agent(name="guardrail-repro", model=model, input_guardrails=[policy])
    result = await Runner.run(agent, "test")
    print(result.final_output)
    print(len(model.calls))


asyncio.run(main())
```

On current `main`, this prints `model-ran` and `1`.

### Expected behavior

A guardrail decision must be an actual boolean. If `tripwire_triggered` is not a `bool`, the SDK should reject the guardrail result before downstream execution instead of interpreting a falsy non-boolean as an allow decision.

The same validation should apply to both input and output guardrails. Explicit `True` and `False` behavior should remain unchanged.

sylvesterkaczmarek:
### Summary

Reject non-boolean `tripwire_triggered` values returned by input and output guardrails.

`GuardrailFunctionOutput.tripwire_triggered` is typed as `bool`, but unchecked runtime values were previously consumed by truthiness checks. A guardrail returning `None` could therefore be treated as passing and allow downstream execution.

Validation now happens at the shared `InputGuardrail.run()` / `OutputGuardrail.run()` boundary before the result reaches runner logic. Explicit `True` and `False` behavior is unchanged.

### Test plan

- `uv run pytest -q tests/test_guardrails.py` — 63 passed
- `uv run ruff check src/agents/guardrail.py tests/test_guardrails.py` — passed
- `uv run ruff format --check src/agents/guardrail.py tests/test_guardrails.py` — passed
- targeted mypy — passed
- `git diff --check` — passed
- repository verification wrapper was attempted; its global typecheck stopped on existing optional-dependency and unrelated sandbox typing errors in untouched files (`litellm`, `temporalio`, SQLAlchemy, Vercel/Runloop sandbox typing, and related baseline errors)

Fixes #4854.

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -68,6 +68,11 @@ class OutputGuardrailResult:
     """The output of the guardrail function."""
 
 
+def _validate_tripwire_triggered(output: GuardrailFunctionOutput, guardrail_name: str) -> None:
+    if not isinstance(output.tripwire_triggered, bool):
+        raise UserError(f"Guardrail {guardrail_name} must return a boolean for tripwire_triggered.")
+
+
 @dataclass
 class InputGuardrail(Generic[TContext]):
     """Input guardrails are checks that run either in parallel with the agent or before it starts.
@@ -119,15 +124,9 @@ class InputGuardrail(Generic[TContext]):
 
         output = self.guardrail_function(context, agent, input)
         if inspect.isawaitable(output):
-            return InputGuardrailResult(
-                guardrail=self,
-                output=await output,
-            )
-
-        return InputGuardrailResult(
-            guardrail=self,
-            output=output,
-        )
+            output = await output
+        _validate_tripwire_triggered(output, self.get_name())
+        return InputGuardrailResult(guardrail=self, output=output)
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


seratch:
please see https://github.com/openai/openai-agents-python/issues/4854#issuecomment-5548123049

PR #4857, opened 2026-09-04
fix(core): validate tool guardrail behaviors

### LINKED ISSUE #4856: Tool guardrails fail open on unknown behavior types
### Describe the bug

`ToolGuardrailFunctionOutput.behavior` is typed as one of `allow`, `reject_content`, or `raise_exception`, but runtime behavior is not validated. If unchecked Python returns an unknown behavior such as `{"type": "unknown"}`, the tool guardrail execution path falls through without blocking and treats the result as effectively allowed.

For input guardrails this means the guarded tool can execute even though the policy did not return a valid allow/block decision. For output guardrails an invalid behavior can leave the original tool output accepted.

### Reproduction

```python
from typing import Any, cast

from agents import Agent, ToolGuardrailFunctionOutput, ToolInputGuardrail, ToolInputGuardrailData

async def invalid(_data: ToolInputGuardrailData) -> ToolGuardrailFunctionOutput:
    return ToolGuardrailFunctionOutput(
        output_info=None,
        behavior=cast(Any, {"type": "unknown"}),
    )

guardrail: ToolInputGuardrail[Any] = ToolInputGuardrail(invalid, name="policy")
result = await guardrail.run(
    ToolInputGuardrailData(
        context=ToolContext(
            context=None,
            tool_name="dangerous",
            tool_call_id="call_1",
            tool_arguments="{}",
        ),
        agent=Agent(name="test"),
    )
)
```

On current `main`, the invalid result is returned and the runner's tool-guardrail executor only branches for `raise_exception` and `reject_content`; an unknown string therefore falls through as allow.

### Expected behavior

Tool guardrail behavior must match one of the documented variants. Unknown or malformed behavior values should raise an SDK `UserError` before tool execution/output acceptance.

The existing default `ToolGuardrailFunctionOutput(output_info=...)` behavior should remain `allow`, and valid `allow`, `reject_content`, and `raise_exception` values should be unchanged.

sylvesterkaczmarek:
### Summary

Validate tool guardrail behavior values at the shared `ToolInputGuardrail.run()` / `ToolOutputGuardrail.run()` boundary.

Previously, an unchecked callback could return an unknown behavior type and the runner would fall through as if the guardrail allowed execution. The shared boundary now accepts only the documented `allow`, `reject_content`, and `raise_exception` variants, and requires a string message for `reject_content`.

The existing default `ToolGuardrailFunctionOutput(output_info=...)` behavior remains `allow`.

### Test plan

- `uv run pytest -q tests/test_tool_guardrails.py` — 25 passed
- `uv run ruff check src/agents/tool_guardrails.py tests/test_tool_guardrails.py` — passed
- `uv run ruff format --check src/agents/tool_guardrails.py tests/test_tool_guardrails.py` — passed
- targeted mypy — passed
- `git diff --check` — passed
- repository verification wrapper was also run: lint passed and the test job reached 7,515 passed / 69 skipped before unrelated existing optional-dependency, voice, sandbox and release-contract environment failures; global typechecking likewise depends on optional packages not installed in this environment

Fixes #4856.

### DIFF
--- a/src/agents/tool_guardrails.py
+++ b/src/agents/tool_guardrails.py
@@ -117,6 +117,22 @@ class ToolGuardrailFunctionOutput:
         return cls(output_info=output_info, behavior=RaiseExceptionBehavior(type="raise_exception"))
 
 
+def _validate_guardrail_behavior(
+    output: ToolGuardrailFunctionOutput, guardrail_name: str
+) -> ToolGuardrailFunctionOutput:
+    behavior = output.behavior
+    if not isinstance(behavior, dict):
+        raise UserError(f"Tool guardrail {guardrail_name} returned an invalid behavior")
+
+    behavior_type = behavior.get("type")
+    if behavior_type in {"allow", "raise_exception"}:
+        return output
+    if behavior_type == "reject_content" and isinstance(behavior.get("message"), str):
+        return output
+
+    raise UserError(f"Tool guardrail {guardrail_name} returned an invalid behavior")
+
+
 @dataclass
 class ToolInputGuardrailData:
     """Input data passed to a tool input guardrail function."""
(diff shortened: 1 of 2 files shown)


seratch:
The issue was closed, so I am closing this PR as well

PR #4861, opened 2026-09-04
fix(core): raise a clear UserError when a tool parameter is named model_config

coderdailyone:
### Summary

A tool parameter named `model_config` made `function_schema()` fail with an unrelated-looking error from inside Pydantic:

```
TypeError: 'FieldInfo' object is not iterable
```

`function_schema()` builds the argument model with `create_model(f"{name}_args", __base__=BaseModel, **fields)`, and Pydantic interprets a `model_config` keyword to `create_model()` as the model's configuration dict rather than as a field, so it tries to iterate the `FieldInfo`. Other reserved names already produce readable Pydantic errors (`model_dump` / `model_validate` → "conflicts with member … of protected namespace", leading underscores → "Fields must not use names with leading underscores"); `model_config` was the one that failed opaquely.

This PR checks for that name before calling `create_model()` and raises a `UserError` that names the function and the parameter:

```
Parameter `model_config` in function configure is reserved by Pydantic and cannot be a tool argument. Rename the parameter.
```

No behavior changes for any other parameter name.

### Test plan

- `test_model_config_parameter_raises_a_clear_user_error` in `tests/test_function_schema.py`: on `main` it fails with the Pydantic `TypeError`; with this change it gets the `UserError`.
- `tests/test_function_schema.py`, `tests/test_function_tool.py`, `tests/test_function_tool_decorator.py`: 177 passed. `ruff format`/`ruff check` clean, `mypy` and `pyright` clean on the changed files (Linux, Python 3.10). The repository-wide `mypy src` reports pre-existing Python 3.10 errors in `archive_ops.py` / `run_loop.py` unrelated to this change, so the full-stack checkbox is left unchecked.

### Issue number

None (found while probing `function_schema()` with reserved parameter names).

### Checks

- [x] I've run the verification steps from `.agents/skills/code-change-verification` individually (format, lint, typecheck on changed files, tests)
- [ ] I've confirmed all verification steps pass (see Test plan)

https://claude.ai/code/session_01DFSMrLZZ3oq1dKmJFAofz6

### DIFF
--- a/src/agents/function_schema.py
+++ b/src/agents/function_schema.py
@@ -511,6 +511,13 @@ def function_schema(
                 )
 
     # 3. Dynamically build a Pydantic model
+    if "model_config" in fields:
+        # Pydantic reads a ``model_config`` keyword as the model's configuration, not as a
+        # field, so create_model() fails deep inside Pydantic with an unhelpful TypeError.
+        raise UserError(
+            f"Parameter `model_config` in function {func_name} is reserved by Pydantic and cannot"
+            " be a tool argument. Rename the parameter."
+        )
     dynamic_model = create_model(f"{func_name}_args", __base__=BaseModel, **fields)
 
     # 4. Build JSON schema from that model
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Catching `model_config` before `create_model()` is the correct ownership boundary: this is a Pydantic metaclass/config namespace collision, not a malformed tool schema, so allowing the opaque `FieldInfo` TypeError to escape misdiagnoses the user's function. The exact-name check is intentionally narrow and leaves all other parameter handling unchanged. The regression also pins the public error to the function and offending parameter.

seratch:
We consider #4800 as the primary candidate for the same purpose.

PR #4866, opened 2026-09-04
fix(tracing): record the tool output on approval-gated function spans

dfedoryshchev:
### Summary

When a function tool is gated by `needs_approval`, the function span exports the internal `FunctionToolResult` wrapper into `span_data.output` instead of the tool's output.

`_run_single_tool` returns either a raw tool output or a `FunctionToolResult`, because the approval path short-circuits. Every other consumer discriminates the two with `isinstance`, including `_build_function_tool_results` in the same file; the span write does not. `FunctionSpanData.export()` then calls `str()` on the wrapper, so the trace carries a repr of the whole `Agent` object in the field documented to hold the tool output, including `agent.instructions`, the tool inventory with JSON schemas, the model settings and the raw arguments. On the repo's own test agent that is 2506 characters, under the default `trace_include_sensitive_data=True`.

The rejection message leaked by two separate routes: it was written to `span_data.output` from outside the sensitive-data gate, and it went onto `SpanError.message`, which `Span.export()` sends through unchanged. Both reached the backend with the flag off.

The fix reads the output off the wrapper the way the rest of the file already does, drops the ungated write, and routes the error message through `_error_tracing.get_trace_error` under the same flag.

### Test plan

- Three tests in `tests/test_run_step_execution.py`, all failing on `main`: the `FunctionToolResult(...)` repr where `None` is expected, `assert error["message"] == "Tool execution rejected"` against a leaked `secret-denial-456`, and the include-sensitive-data path keeping the message intact.
- The redaction test asserts against the whole exported span rather than `span_data.output` alone, so a leak through any other field fails it too.
- Suite 7656 -> 7659 passed, with the same 25 pre-existing environment failures as an unpatched run. `ruff check`, `ruff format --check` and `mypy` on the changed module are clean.

The behaviour change worth a second look: with `trace_include_sensitive_data=False` the exported rejection output goes from the message text to `None`, and the span error message to a fixed string. That is what the flag asks for and it matches the error paths, but it is visible to anyone reading rejection reasons out of a redacted trace.

`redacted_message` is passed inline at the call site, where `_tool_errors.py` hoists its equivalent to a module constant. Say the word and I will match that shape.

### Issue number

None.

### DIFF
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -1971,7 +1971,13 @@ class _FunctionToolBatchExecutor:
         )
         span_fn.set_error(
             SpanError(
-                message=rejection_message,
+                # The rejection message is app-supplied text, so it reaches the exported span
+                # under the same sensitive-data gate as the tool output above.
+                message=_error_tracing.get_trace_error(
+                    trace_include_sensitive_data=self.config.trace_include_sensitive_data,
+                    error_message=rejection_message,
+                    redacted_message="Tool execution rejected",
+                ),
                 data={
                     "tool_name": func_tool.name,
                     "error": (
--- a/src/agents/run_internal/tool_execution.py
+++ b/src/agents/run_internal/tool_execution.py
@@ -1852,7 +1852,12 @@ class _FunctionToolBatchExecutor:
                 raise UserError(f"Error running tool {func_tool.name}: {e}") from e
 
             if self.config.trace_include_sensitive_data:
-                span_fn.span_data.output = result
+                # Approval short-circuits return the FunctionToolResult wrapper rather than the
+                # tool's own output, so read the output off it the way every other consumer of
+                # this value does (see `_build_function_tool_results`).
+                span_fn.span_data.output = (
+                    result.output if isinstance(result, FunctionToolResult) else result
+                )
             return result
 
     async def _maybe_execute_tool_approval(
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The unwrap is at the right owner: `_run_single_tool` already has the sensitive-data gate and is the common point for raw outputs and `FunctionToolResult`. Removing the earlier rejection write is important because it otherwise bypasses the redaction policy even though the later gated write is correct. The pending-approval regression also catches the larger leak surface in the wrapper repr (Agent/instructions/tool schemas), not just a cosmetic type mismatch. No blocker.

seratch (changes requested):
The wrapper-output fix is useful, but the rejection message still reaches the exported span through `SpanError.message` with sensitive-data tracing disabled. This leak also exists on the base; the issue is that the rejection-redaction part of this PR remains incomplete. Please gate the error message at the same approval boundary and assert against the complete exported span, or split this into the wrapper-output fix alone.

dfedoryshchev:
@seratch Updated - the error message is gated the same way now. Have a look.

sylvesterkaczmarek (commented):
The rejection message now follows the same sensitive-data tracing gate as tool output, so the exported span no longer leaks app-supplied rejection text when sensitive tracing is disabled. This resolves the remaining blocker.

seratch:
@sylvesterkaczmarek Could you please refrain from posting this type of comment on other people's PRs?

seratch (approved):

PR #4869, opened 2026-09-04
fix(sessions): reset compaction state on pop

### LINKED ISSUE #4867: Popping compaction session history retains the previous response chain
### Describe the bug

`OpenAIResponsesCompactionSession.pop_item()` removes the latest item from the client-managed session history, but it leaves the wrapper's retained Responses chain state unchanged: `_response_id`, `_deferred_response_id`, and `_last_unstored_response_id`.

After a successful pop, the local history and the retained server-managed response chain can therefore describe different conversations. A later compaction in `previous_response_id` mode can reuse a response ID whose server-side history still contains the item the caller explicitly removed locally.

### Reproduction

```python
from unittest.mock import AsyncMock, MagicMock

from agents.memory import OpenAIResponsesCompactionSession
from tests.utils.simple_session import SimpleListSession

item = {"type": "message", "role": "assistant", "content": "old"}
underlying = SimpleListSession(history=[item])
client = MagicMock()
client.responses.compact = AsyncMock(
    return_value=SimpleNamespace(output=[], usage=None)
)

session = OpenAIResponsesCompactionSession(
    session_id="test",
    underlying_session=underlying,
    client=client,
    compaction_mode="previous_response_id",
)

# Below the threshold, this records the response ID without making a compact call.
await session.run_compaction({"response_id": "resp-old"})

# The local history no longer matches the response chain behind resp-old.
await session.pop_item()

# Current main still reuses resp-old here instead of requiring fresh chain state.
await session.run_compaction({"force": True})
```

On current `main`, the final call proceeds with `previous_response_id="resp-old"`.

### Expected behavior

A successful `pop_item()` that actually removes an item should invalidate all retained response-chain state owned by the compaction wrapper. A later `previous_response_id` compaction should require a newly supplied response ID before making a client call.

If `pop_item()` is a no-op because the session is empty, the existing response-chain state should remain unchanged. If the underlying pop raises, state should likewise remain unchanged because the mutation did not complete.

This is the `pop_item()` counterpart to #4864: both operations mutate local history, but `pop_item()` has its own successful-mutation versus no-op boundary.

sylvesterkaczmarek:
### Summary

Reset retained Responses response-chain state after `OpenAIResponsesCompactionSession.pop_item()` actually removes an item from local session history.

Previously, a successful pop invalidated the cached session/compaction items but retained `_response_id`, `_deferred_response_id`, and `_last_unstored_response_id`. A later `previous_response_id` compaction could therefore reuse server-side history that still contained the item the caller had removed locally.

The response-chain fields now reset at the same successful-mutation boundary as the local caches. A no-op pop preserves the existing chain, and an underlying pop failure still leaves state untouched.

### Test plan

- `uv run pytest -q tests/memory/test_openai_responses_compaction_session.py` — 57 passed
- Ruff check/format on changed files — passed
- targeted mypy — passed
- `git diff --check` — passed
- repository lint — passed
- full runtime test job reached 7,487 passed / 96 skipped; the remaining failures/errors are existing optional-integration/environment gaps involving packages such as httpx, Docker, voice/numpy, LiteLLM, SQLAlchemy and sandbox/release-contract fixtures
- repository-wide typecheck remains blocked by the same missing optional dependencies and unrelated sandbox/voice baseline errors in untouched files

Fixes #4867.

### DIFF
--- a/src/agents/memory/openai_responses_compaction_session.py
+++ b/src/agents/memory/openai_responses_compaction_session.py
@@ -431,6 +431,9 @@ class OpenAIResponsesCompactionSession(SessionABC, OpenAIResponsesCompactionAwar
         async with self._mutation_lock:
             popped = await self.underlying_session.pop_item()
             if popped:
+                self._response_id = None
+                self._deferred_response_id = None
+                self._last_unstored_response_id = None
                 self._compaction_candidate_items = None
                 self._session_items = None
             return popped
--- a/tests/memory/test_openai_responses_compaction_session.py
+++ b/tests/memory/test_openai_responses_compaction_session.py
@@ -166,6 +166,53 @@ class TestOpenAIResponsesCompactionSession:
         mock_session.get_items.assert_called_once()
 
     @pytest.mark.asyncio
+    async def test_pop_item_invalidates_response_chain_state(self) -> None:
+        item = cast(
+            TResponseInputItem,
+            {"type": "message", "role": "assistant", "content": "old"},
+        )
+        underlying = SimpleListSession(history=[item])
+        mock_client = MagicMock()
+        mock_client.responses.compact = AsyncMock(
+            return_value=SimpleNamespace(output=[], usage=None)
+        )
+        session = OpenAIResponsesCompactionSession(
+            session_id="test",
+            underlying_session=underlying,
+            client=mock_client,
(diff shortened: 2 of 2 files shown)


seratch:
We'll consider #4868 for this

PR #4871, opened 2026-09-05
tests(sandbox): skip workspace path tests when symlinks are unavailable

iam-kira:
Five tests in test_workspace_paths.py create symlinks in their setup with no guard, so they fail on Windows for any user who is not elevated and does not have Developer Mode enabled:

    OSError: [WinError 1314] A required privilege is not held by the client

CI does not catch this: tests-windows runs the full suite on windows-latest and its runner holds the privilege, so these pass there and fail only locally.

Route the module's os.symlink calls through a small helper that skips with a clear reason when the platform refuses. Where symlinks work the helper is a passthrough, so behaviour on CI and on POSIX is unchanged.

test_workspace_paths.py on Windows: 5 failed / 65 passed -> 0 failed / 65 passed.

Note: this covers only the test-side half of the problem. The seven failures in test_tar_utils.py come from safe_extract_tarfile itself calling os.symlink unguarded, which needs a decision about symlink semantics on Windows and is left for maintainers.

### DIFF
--- a/tests/sandbox/test_workspace_paths.py
+++ b/tests/sandbox/test_workspace_paths.py
@@ -33,16 +33,23 @@ class WorkspacePathCase:
 
 
 def _symlink_or_skip(*args: Any, **kwargs: Any) -> None:
-    """Create a symlink, or skip the test where the platform forbids one.
+    """Create a symlink, or skip the test where Windows forbids one.
 
     Windows refuses symlink creation unless the process is elevated or Developer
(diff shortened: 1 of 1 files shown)


iam-kira:
Partially addresses #4852

seratch:
Thanks for addressing the Windows symlink setup failures. #4853 covers the same five test sites with a narrower typed helper and the same WinError 1314-only skip policy. I recommend closing this duplicate and keeping #4853 as the single implementation. Other symlink errors should continue to fail the tests.

PR #4873, opened 2026-09-05
fix(memory): let the client paginate conversation items

### LINKED ISSUE #4757: OpenAIConversationsSession get_items(limit>provider page size) forwards the session limit as provider page size
`OpenAIConversationsSession.get_items(limit=N)` currently forwards `N` directly to `conversations.items.list(limit=N)`. For large session limits this couples the SDK's local history cutoff to the provider's page-size constraints and can make an otherwise valid `Session.get_items(limit=N)` call fail before client pagination occurs.

A previous implementation attempt in #4721 confirmed the underlying bug. Maintainer guidance there was to avoid copying provider page-size constants into the SDK and instead omit `limit` from `conversations.items.list()`, letting the OpenAI client handle provider pagination while preserving the SDK's existing local cutoff.

Expected behavior: for a finite session limit, request items in descending order without forwarding the SDK limit as a provider page size, stop locally once N items are collected, then reverse them into chronological order.

Hughhhhcoder:
### Summary

This pull request fixes a mismatch between the Agents SDK session limit and the Conversations API page size. `OpenAIConversationsSession.get_items(limit=N)` currently forwards the SDK's total-result limit as the provider request limit, so a valid large session limit can be rejected before client pagination begins.

The finite-history path now omits the provider `limit`, as requested in #4721, lets the OpenAI Python async paginator own provider page sizing and cursor traversal, and keeps the existing local cutoff and final reversal. A regression requests 101 items, verifies that pagination can cross a provider page boundary without forwarding `limit`, stops at exactly 101 items, and returns the latest items in chronological order.

### Test plan

- Python 3.10 focused session suite: 39 passed
- Python 3.12 focused session suite: 39 passed
- Independent final review: two clean verdicts; no findings
- `.agents/skills/code-change-verification/scripts/run.sh`: format, lint, typecheck, and complete tests passed

### Issue number

Closes #4757

### Checks


Hughhhhcoder:
The only failing check is `tests/test_guardrails.py::test_parallel_guardrail_trip_cancels_model_task` on Windows/Python 3.10. The PR diff only touches `openai_conversations_session.py` pagination and its focused tests, so this failure is outside the changed path. I ran that exact guardrail test 20 consecutive times with Python 3.10.16 on the PR head and all 20 passed; every other CI job, including Windows/Python 3.13 and the focused memory tests within the full suite, passed. No PR code change appears justified for this result; a rerun of the failed job would be appropriate.

seratch (changes requested):
The tests with Python 3.10 + Windows fail even with retries. Can you check what's wrong?

Hughhhhcoder:
I checked both Windows/Python 3.10 failures rather than treating the rerun as the same result:

- The original job `101304163924` failed `tests/test_guardrails.py::test_parallel_guardrail_trip_cancels_model_task`.
- The latest rerun job `101770004198` passed that test and instead failed `tests/test_trace_processor.py::test_backend_span_exporter_deadline_stops_during_5xx_retry_backoff` because the 10 ms deadline elapsed before the mocked retry sleep.

Neither file differs in this PR; the complete diff remains limited to `openai_conversations_session.py` and its memory tests. On the exact PR head with Python 3.10.16, I reran both failing tests together repeatedly (including a fresh 5/5 sequence) and every run passed. The fact that the retry moved to a different timing-sensitive test is also evidence against the pagination change being causal.

I do not see a scoped PR code change that would address either result without mixing unrelated test behavior into this fix. Another rerun of the failed Windows job should be the appropriate next step; if you prefer, I can separately investigate hardening the tracing deadline test against scheduler delay.

Hughhhhcoder:
I attempted to rerun the remaining Windows/Python 3.10 failure from the contributor account, but GitHub requires repository admin rights for this historical workflow run. The failing tests remain outside this PR's diff, and the exact tests pass locally; no unrelated code change has been pushed.

seratch:
Thanks for suggesting this. https://github.com/openai/openai-agents-python/pull/4961 will resolve the issue.

PR #4882, opened 2026-09-05
fix(sandbox): preserve empty Mountpoint prefixes

Hughhhhcoder:
### Summary

This pull request fixes explicit empty prefixes being replaced by a `MountpointMountPattern` default for S3 and GCS mounts.

Both provider adapters used a truthiness fallback, so `prefix=""` behaved like `None` and unexpectedly mounted the configured default subtree instead of the bucket root. The adapters now use `None` as the sole inheritance sentinel, preserving explicit empty prefixes while keeping nonempty overrides and existing `None` behavior unchanged.

The regression test exercises the public `Mount.apply` path for both providers and verifies the final command omits `--prefix` for an explicit empty value while still forwarding the pattern default for `None`.

### Test plan

- `uv run --locked pytest tests/sandbox/test_mounts.py::test_mountpoint_prefix_uses_none_as_inheritance_sentinel tests/sandbox/test_mounts.py::test_s3_mount_direct_mountpoint_fields_override_pattern_options -q`
- `/usr/bin/env -u OPENAI_API_KEY OPENAI_AGENTS_TEST_IN_CODEX_SANDBOX=1 UV_DEFAULT_INDEX=https://pypi.org/simple PYTEST_XDIST_AUTO_NUM_WORKERS=4 bash .agents/skills/code-change-verification/scripts/run.sh`

### Issue number

N/A

### Checks

### DIFF
--- a/src/agents/sandbox/entries/mounts/providers/gcs.py
+++ b/src/agents/sandbox/entries/mounts/providers/gcs.py
@@ -142,7 +142,7 @@ class GCSMount(_ConfiguredMount):
                 access_key_id=self.access_id,
                 secret_access_key=self.secret_access_key,
                 session_token=None,
-                prefix=self.prefix or options.prefix,
+                prefix=self.prefix if self.prefix is not None else options.prefix,
                 region=self.region or options.region,
                 endpoint_url=(
                     self.endpoint_url or options.endpoint_url or "https://storage.googleapis.com"
--- a/src/agents/sandbox/entries/mounts/providers/s3.py
+++ b/src/agents/sandbox/entries/mounts/providers/s3.py
@@ -101,7 +101,7 @@ class S3Mount(_ConfiguredMount):
                 access_key_id=self.access_key_id,
                 secret_access_key=self.secret_access_key,
                 session_token=self.session_token,
-                prefix=self.prefix or options.prefix,
+                prefix=self.prefix if self.prefix is not None else options.prefix,
                 region=self.region or options.region,
                 endpoint_url=self.endpoint_url or options.endpoint_url,
                 mount_type=self.type,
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek (commented):
Using `None` as the inheritance sentinel is the right distinction here because an empty prefix has real semantics. I also like that the regression exercises `Mount.apply` and the final argv for both S3 and GCS rather than only checking the dataclass merge.

One consistency question: region/endpoint still use truthiness-based fallback. That is fine if empty strings are invalid for those fields; if an empty value is ever meant to explicitly clear an inherited setting, they would have the same ambiguity this PR fixes for `prefix`. For `prefix` itself, the change looks correct.

Hughhhhcoder:
Thanks for raising this. I checked the downstream semantics before keeping the change scoped to `prefix`: `MountpointMountPattern.apply()` only emits `--region` and `--endpoint-url` for non-empty values, and the provider paths likewise treat an empty endpoint as absent/default rather than as a supported explicit region or URL. An empty prefix is different: it has a concrete meaning (mount the bucket root), so it must be able to suppress an inherited `--prefix`. If explicit clearing of inherited region/endpoint defaults is desired later, that would need a separate configuration contract rather than treating an invalid empty region or URL as a value.

seratch:
Thanks for investigating this. The empty-string fallback is real, and the patch implements the proposed distinction. However, bucket-root mounts already work by giving the mount a pattern with `options.prefix=None`, while retaining its region and endpoint settings.

I would close this PR for now because the report does not establish a concrete workflow where that configuration is insufficient. Changing the fallback also changes released behavior for existing configurations. We can revisit this if there is an application scenario that requires clearing an inherited prefix through the mount entry and cannot reasonably use a separate pattern.

PR #4894, opened 2026-09-06
fix(sandbox): reject commas in S3 Files mount -o options

Oskii:
## Summary

`S3FilesMountPattern.apply()` turns helper options into one `mount -t s3files -o …` argument by comma-joining every key/value:

```python
rendered_options = ",".join(
    key if value is None else f"{key}={value}" for key, value in options.items()
)
cmd.extend(["-o", rendered_options])
```

`extra_options`, `mount_target_ip`, `access_point`, and `region` all go into that string. A comma in any of them is a new `-o` token.

```python
S3FilesMount(
    file_system_id="fs-123",
    extra_options={"mounttargetip": "10.0.0.1,ro"},
    mount_strategy=InContainerMountStrategy(pattern=S3FilesMountPattern()),
)
```

renders (after the usual join):

```
mount -t s3files -o mounttargetip=10.0.0.1,ro … fs-123 /workspace/data
```

`ro` was not requested. The same form works with a key (`"foo,ro": "1"`) or with `mount_target_ip="10.0.0.1,subdir=/"`.

Blaxel s3fs already rejects this class of injection (`cloud mount configuration values must not contain s3fs option delimiters` / `test_rejects_blaxel_s3fs_endpoint_option_injection`). S3 Files has the same join and no check. Broad-credential acknowledgement still lets the comma through, because that acknowledgement is about ambient IAM, not about `-o` syntax.

## Why

I was reading the in-container mount helpers next to the Blaxel delimiter tests. The S3 Files path builds the same comma-joined `-o` list and never calls `_value_contains_s3fs_option_delimiter`. I put a comma in `extra_options` after acknowledging broad IAM and `validate_manifest_mount_credential_boundaries` accepted it.

This matters when a manifest is assembled from more than one source: a shared template, a restored session, or a third-party example. The person who runs the sandbox may not have written every option field.

## How

Reuse `_value_contains_s3fs_option_delimiter` on S3 Files joined fields and on `extra_options` keys/values (mount object and pattern options). The existing Blaxel error path now includes those fields. No new helper type.

## Testing

Added `test_rejects_s3_files_mount_option_delimiter_injection`.

```
uv run pytest tests/sandbox/test_mount_security.py::test_rejects_s3_files_mount_option_delimiter_injection tests/sandbox/test_mount_security.py::test_s3_files_require_broad_acknowledgement_before_ambient_iam_can_be_used tests/sandbox/test_mount_security.py::test_rejects_blaxel_s3fs_endpoint_option_injection -q
```

9 passed.

Safe `extra_options={"tlsport": "4049"}` still works after acknowledgement.

### DIFF
--- a/src/agents/sandbox/_mount_security.py
+++ b/src/agents/sandbox/_mount_security.py
@@ -109,6 +109,8 @@ _BLAXEL_S3FS_OPTION_FIELDS_BY_MOUNT_TYPE: dict[str, tuple[str, ...]] = {
     "r2_mount": ("custom_domain", "account_id"),
     "s3_mount": ("endpoint_url", "region"),
 }
+# S3 Files renders these fields (plus extra_options) into a single comma-joined `mount -o` string.
+_S3_FILES_JOINED_OPTION_FIELDS: tuple[str, ...] = ("mount_target_ip", "access_point", "region")
 # Every free-form value interpolated into an rclone configuration line must remain a single line.
 # Keep this table aligned with the built-in providers' ``_rclone_required_lines`` methods.
 _RCLONE_CONFIG_VALUE_FIELDS_BY_MOUNT_TYPE: dict[str, tuple[str, ...]] = {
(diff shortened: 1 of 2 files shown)


seratch:
The comma-separated option parsing is real, but the current report does not establish a concrete operational failure or a bypass of a protected boundary. `extra_options` already permits explicit helper options, and this S3 Files path requires a runtime acknowledgement of broad credential exposure. The template and restore examples do not show which restricted input gains authority it did not already have. I am closing this standalone validation change on the current evidence.

PR #4905, opened 2026-09-08
fix(sandbox): distinguish directory modes from sockets and devices

SeanKudrna:
### Summary

This pull request fixes directory classification in `Permissions.from_mode`. The current bitwise check treats `stat.S_IFDIR` as an independent flag, but Unix file types share bits: socket mode `0o140000` and block-device mode `0o060000` also contain `0o040000`.

As a result, `UnixLocalSandboxSession.ls()` can return a socket with `kind=OTHER` and directory permissions such as `drwxr-xr--`. Use `stat.S_ISDIR(mode)` to classify the complete file-type field while preserving rwx extraction. No public signature, permission enforcement, or stored schema changes.

### Test plan

- Before the fix, the new socket and block-device regression cases fail.
- After the fix, `uv run pytest tests/sandbox/test_types.py tests/sandbox/test_parse_utils.py -q`: 23 passed.
- Regression coverage includes directories, regular files, symlinks, sockets, block/character devices, FIFOs, and permission-only modes.
- Independent review completed with no actionable findings.
- Confirmed `UnixLocalSandboxSession.ls()` reports `directory=False` for a real Unix socket.
- `make format`, `make lint`, `make mypy`, and `make pyright`: passed.
- Full parallel tests on macOS/Python 3.14: 9,476 passed, 55 skipped, 1 failed. Serial tests: 77 passed, 4 skipped.
- The failing test, `tests/sandbox/test_runtime.py::test_runner_merges_sandbox_instructions_and_tools`, also fails with the unchanged upstream runtime. The workspace sandbox prevents creating the default snapshot directory under the user's Library/Application Support directory, so runtime falls back to `NoopSnapshotSpec` while the test expects `LocalSnapshotSpec`. A direct probe confirms `PermissionError`. This is unrelated to file-type classification.
- Used `OPENAI_AGENTS_TEST_IN_CODEX_SANDBOX=1` as prescribed. Native macOS sandbox tests remain for the dedicated CI runner. The verification wrapper was invoked, but restricted process inspection made its status accounting unreliable; results above come from the documented direct commands, with the serial suite run separately after the environment failure.

### Issue number

No existing issue found for this metadata-classification bug.

### Checks

- [x] I've added new tests, if relevant.
- [x] I've run `.agents/skills/code-change-verification/scripts/run.sh`.
- [ ] I've confirmed all verification steps pass. One reproduced upstream environment failure remains, as detailed above.
- [x] Independent Codex review completed before submission.

### DIFF
--- a/src/agents/sandbox/types.py
+++ b/src/agents/sandbox/types.py
@@ -51,7 +51,8 @@ class Permissions(BaseModel):
             owner=(mode >> 6) & 0b111,
             group=(mode >> 3) & 0b111,
             other=(mode >> 0) & 0b111,
-            directory=bool(mode & stat.S_IFDIR),
+            # File types share bits; sockets and block devices also contain S_IFDIR's bit.
+            directory=stat.S_ISDIR(mode),
         )
 
     @classmethod
--- a/tests/sandbox/test_types.py
+++ b/tests/sandbox/test_types.py
@@ -1,4 +1,29 @@
+import stat
+
+import pytest
+
 from agents.sandbox.types import Group, Permissions, User
+
+
+@pytest.mark.parametrize(
+    ("file_type", "is_directory"),
+    [
+        pytest.param(stat.S_IFDIR, True, id="directory"),
+        pytest.param(stat.S_IFREG, False, id="regular-file"),
+        pytest.param(stat.S_IFLNK, False, id="symlink"),
+        pytest.param(stat.S_IFSOCK, False, id="socket"),
+        pytest.param(stat.S_IFBLK, False, id="block-device"),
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for investigating this. The `stat.S_ISDIR` correction is technically sound, and the socket example demonstrates the metadata inconsistency. However, `ls()` already returns the correct `FileEntry.kind`, and `FileEntry.is_dir()` and the SDK's listing consumers use that classification. The PR does not show a downstream failure caused by the permissions flag.

I'm going to close this for now because the demonstrated impact is limited to metadata. We can reconsider if there is a concrete supported consumer whose behavior is incorrect because it relies on `permissions.directory`.

PR #4908, opened 2026-09-08
fix(memory): close SQLite connections when setup fails

Jokasa7:
This pull request fixes SQLiteSession connection cleanup when configuration or initialization fails, preserving the original exception and allowing file-backed sessions to retry after a transient lock clears.

This is a draft while the full verification gap described below remains open. Implementation and regression tests were developed with Codex.

### Summary

Each new file-backed worker connection is configured before it is registered for normal cleanup. If enabling WAL fails because another connection holds an exclusive lock, that connection remains open and untracked by session.close(). In-memory construction similarly leaves its shared connection open when initialization raises.

Both failure paths now use the existing connection invalidation policy. Public APIs, table schemas, WAL retry timing, and successful operations remain unchanged. Two regression tests exercise real SQLite errors, verify connection closure, and cover recovery and normal close behavior.

### Test plan

- Both new regressions fail against the original runtime source and pass with the fix.
- CPython 3.10.20: 57 passed, covering the full SQLiteSession test file and two adjacent AdvancedSQLiteSession tests. Only optional numeric error-code assertions are restricted to Python 3.11+; exception class/message, cleanup, and retry checks run on Python 3.10.
- CPython 3.12.13: focused SQLiteSession tests 55 passed.
- Installed all locked development dependencies and extras with uv sync --frozen --all-extras --all-packages --group dev.
- Ran the documented manual workflow using the unchanged Makefile: make format, make lint, make typecheck, then make tests. Format and lint passed; Mypy passed for 310 source files; Pyright reported zero errors and warnings.
- Full local test verification remains blocked by the Windows environment. The parallel phase finished with 14 failed, 8,539 passed, 91 skipped. Twelve unchanged sandbox Docker/tar tests fail at os.symlink with WinError 1314 because this process lacks the required Windows privilege. The other two failures are unchanged Daytona example tests reading UTF-8 text with the local GBK default; both pass in a separate PYTHONUTF8=1 rerun. The complete suite was not rerun, and the serial phase was not reached because the first test phase failed. No privilege escalation, test exclusions, or repository verification changes were applied.
- Two rounds of independent review used two fresh-context reviewers per round. The first round identified the Python 3.10 test-metadata compatibility issue; after correction, both second-round reviewers returned clean results. Content and repository fingerprints remained unchanged through final verification.

### Issue number

No existing issue; found while inspecting SQLiteSession resource ownership.

### Checks

- [ ] I've run .agents/skills/code-change-verification/scripts/run.sh
- [ ] If using Codex, I've run /review before submitting this PR

The script and literal /review checkboxes remain unchecked: verification used the documented manual workflow with locally installed GNU Make and Windows Python, and review used the repository's independent-review procedure. The full-pass checkbox remains unchecked because of the environment failures above.

### DIFF
--- a/src/agents/memory/sqlite_session.py
+++ b/src/agents/memory/sqlite_session.py
@@ -85,7 +85,9 @@ class SQLiteSession(SessionABC):
                     ) as init_conn:
                         self._configure_connection(init_conn)
                         self._init_db_for_connection(init_conn)
-        except Exception:
+        except BaseException:
+            if self._is_memory_db and hasattr(self, "_shared_connection"):
+                self._invalidate_connection(self._shared_connection)
             if self._lock_path is not None and not self._lock_released:
                 self._release_file_lock(self._lock_path)
                 self._lock_released = True
@@ -142,7 +144,7 @@ class SQLiteSession(SessionABC):
                 raise
 
     def _invalidate_connection(self, conn: sqlite3.Connection) -> None:
-        """Close and evict a connection that could not roll back safely."""
+        """Close and evict a connection that cannot be used safely."""
         try:
             conn.close()
         except BaseException:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this. The missing explicit cleanup is real, but failed connections are not cached, and subsequent calls already create fresh connections. The tests demonstrate immediate closure, but do not establish a practical resource-retention problem or show that recovery requires this change.

I'm going to close this PR for now. We can revisit the underlying problem if a realistic workload demonstrates material resource growth or blocked recovery caused by these failed connections.

PR #4910, opened 2026-09-08
fix(sessions): use LONGTEXT for MySQL message storage

rioyu123:
### Summary

This pull request changes the `message_data` column to `LONGTEXT` for newly created MySQL and MariaDB message tables. Other dialects retain `TEXT`; serialization, session identifiers, transactions, and public APIs are unchanged. Existing tables are not migrated.

MySQL-family `TEXT` storage is too small for some serialized session items. In a live example, Chinese/emoji content serialized to 1,200,028 bytes: strict mode rejected the `TEXT` insert with error 1406, while non-strict mode logged truncation and stored 65,535 bytes of invalid JSON that the SDK could not return as the original item. A `LONGTEXT` column preserved the full item through SDK add/get/pop in both modes.

**Draft dependency:** current `main` cannot automatically create these MySQL/MariaDB tables because the `session_id` columns use an unbounded `String`. #4741, an open PR from another contributor, proposes fixing that separate problem. This PR contains none of its changes and does not independently restore automatic table creation. It is open as a draft to discuss the column-capacity change while the schema-creation work is reviewed. Users who already manage their own `LONGTEXT` schema do not need this patch.

Before marking this ready, I will refresh against the accepted schema-creation baseline, verify the combined automatic-create and large-item path, and finish the final review and verification gates.

### Test plan

- `tests/extensions/memory/test_sqlalchemy_session.py`: 47 passed, including four new column-compilation cases for MySQL, MariaDB, PostgreSQL, and SQLite.
- Ruff check and format check passed for both changed files; `git diff --check` passed.
- Real MySQL 8.0.46 and MariaDB 11.8.9, using caller-managed tables and `SQLAlchemySession(create_tables=False)`: small-item controls passed; strict `TEXT` raised error 1406; non-strict `TEXT` produced truncated, invalid JSON; `LONGTEXT` preserved the complete 1,200,028-byte serialized value and exact SDK get/pop results in both modes.
- Offline complete-table compilation: current main plus this change still fails on unbounded `session_id`; #4741's schema plus the proposed column type compiles for both MySQL and MariaDB. This was a disposable metadata composition check, not a combined-source live automatic-creation test.
- The full repository verification stack and final schema-change review are still pending for this early draft. No full-stack pass is claimed.

### Issue number

N/A. Related schema-creation work: #4741.

### Checks

### DIFF
--- a/src/agents/extensions/memory/sqlalchemy_session.py
+++ b/src/agents/extensions/memory/sqlalchemy_session.py
@@ -48,6 +48,7 @@ from sqlalchemy import (
     text as sql_text,
     update,
 )
+from sqlalchemy.dialects.mysql import LONGTEXT
 from sqlalchemy.exc import IntegrityError, OperationalError
 from sqlalchemy.ext.asyncio import AsyncEngine, async_sessionmaker, create_async_engine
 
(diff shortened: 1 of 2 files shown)


rioyu123:
Still planned, and intentionally kept as a draft while #4741 (or an equivalent MySQL/MariaDB schema-creation fix) is reviewed. That dependency has not merged yet. Once the accepted fix reaches main, I will update this branch and verify automatic table creation plus small- and large-item round trips on both databases before requesting review.

Could a maintainer add `skip-stale` while that dependency is pending? I would like to keep the column-capacity fix separate from the schema-creation work.

Issue #4823, opened 2026-09-02
buffer_streamed_tool_calls=True crashes when a Chat Completions provider omits the tool call index


rioyu123:
### Summary

The buffered tool-call path assumes every streamed `tool_calls[]` delta carries an `index`. The OpenAI SDK's lenient `construct_type` parsing leaves `index=None` when a provider omits it; the buffered replay then raises `pydantic_core.ValidationError` (`ChoiceDeltaToolCall.index` must be an int) and, when indexed and index-less calls coexist, `TypeError: '<' not supported between instances of 'NoneType' and 'int'`. The unbuffered path handles the same stream correctly, and the option's documented audience (`src/agents/models/openai_provider.py:85-88`) is exactly OpenAI-compatible providers whose chunk semantics are not reliable.

### Affected code (`main` @ 89c02c82), `src/agents/models/chatcmpl_stream_handler.py`

- `_accumulate_tool_call_delta` (`:309-316`) keys buffered calls on `tool_call_delta.index`, which may be `None`.
- `_buffered_tool_call_delta` (`:355-356`): `ChoiceDeltaToolCall(index=buffered_call.index, ...)` raises `ValidationError` for `None`.
- `_buffered_tool_calls_chunk` (`:379-383`): `sorted(buffered_calls.items())` raises `TypeError` with mixed `None`/`int` keys.

### Reproduction

```python
from openai._models import construct_type
from openai.types.chat.chat_completion_chunk import ChatCompletionChunk
from agents.models.chatcmpl_stream_handler import ChatCmplStreamHandler

def chunk(delta, finish_reason=None):
    return construct_type(
        type_=ChatCompletionChunk,
        value={"id": "c", "object": "chat.completion.chunk", "created": 1, "model": "m",
               "choices": [{"index": 0, "delta": delta, "finish_reason": finish_reason}]},
    )

chunks = [
    chunk({"role": "assistant", "tool_calls": [{"id": "call_1", "type": "function",
                                                "function": {"name": "f", "arguments": '{"a":'}}]}),
    chunk({"tool_calls": [{"function": {"arguments": "1}"}}]}),
    chunk({}, "tool_calls"),
]

async def gen():
    for c in chunks:
        yield c

events = [e async for e in ChatCmplStreamHandler.handle_stream(
    response, ChatCmplStreamHandler.buffer_tool_call_stream(gen()))]
[...]

PR #4918, opened 2026-09-08
fix(chat-completions): handle missing streamed tool call indexes

Hughhhhcoder:
### Summary

This pull request fixes buffered Chat Completions streams from OpenAI-compatible providers that omit `tool_calls[].index`. The OpenAI client preserves the missing field as `None`, and the buffering path previously replayed that value into a strongly typed tool-call delta, causing a Pydantic validation failure.

The buffering path now accepts the unambiguous single-call wire shape, normalizes its start and arguments-only continuation deltas to index `0`, and reuses the existing accumulator. Ambiguous multi-call, repeated-start, and mixed indexed/unindexed shapes fail before accumulation instead of silently merging tool calls.

### Test plan

- `uv run pytest -q tests/models/test_openai_chatcompletions_stream.py` — 114 passed
- `.agents/skills/code-change-verification/scripts/run.sh`
  - format passed
  - lint passed
  - mypy and Pyright passed
  - parallel tests: 9496 passed, 74 skipped
  - serial tests: 77 passed, 4 skipped

### Issue number

N/A

### Checks

- [x] I have added new tests, if relevant
- [x] I have run `.agents/skills/code-change-verification/scripts/run.sh`
- [x] I have confirmed all verification steps pass
- [x] If using Codex, I have run `/review` before submitting this PR


seratch:
Thanks for the contribution. I am closing this for the same reason recorded in #4823 and #4824. The constructed chunks demonstrate the buffered replay failure, but they do not identify an affected provider or show why the existing `buffer_streamed_tool_calls=False` path is insufficient for that stream.

A sanitized raw stream from an identified provider, together with the concrete behavior that requires buffering, would justify reconsidering this. Without that evidence, we should avoid adding permanent missing-index correlation rules to the shared handler.

PR #4923, opened 2026-09-08
fix(chat-completions): drop unrepresentable tool output parts instead of failing the turn

dfedoryshchev:
### Summary

A function tool that returns `ToolOutputImage(file_id=...)` kills the run on the Chat Completions path, in the branch whose job is to discard non-text tool output.

With `preserve_tool_output_all_content` false (the default), `items_to_messages` converts the whole tool output through `extract_all_content` first and only then filters down to text parts. An `input_image` carrying `file_id` instead of `image_url` has no Chat Completions representation, so it raises at `chatcmpl_converter.py:450` on its way to being thrown away. In the mixed case the accompanying text is destroyed with it, and the turn after the tool executed fails.

That is the opposite of what this branch is documented to do. #3312 built it specifically to convert non-text-only output "to a non-empty placeholder string instead of raising", gated the `UserError` behind `strict_feature_validation=True`, and kept "mixed text/non-text outputs preserving only the text parts by default". A URL image already takes that path; only `file_id` fails, and `ToolOutputImage` accepts `file_id` alone by its own validator.

The fix drops the part types that can never become a text part before conversion, rather than converting them and filtering them out afterwards.

### Test plan

- Two cases added to `tests/models/test_openai_chatcompletions_converter.py`, mirroring the existing `..._non_text_only_...` and `..._mixed_function_output_...` tests with `file_id` in place of `image_url`. Both fail on `main` with `UserError: Only image URLs are supported for input_image {'type': 'input_image', 'file_id': 'file-abc123'}`.
- `tests/models/test_openai_chatcompletions_converter.py`: 58 passed, including the existing strict-mode and placeholder cases.
- Full suite before: 20 failed, 7837 passed, 31 errors. After: 20 failed, 7839 passed, 31 errors. Same failure and error sets; the remaining ones are environmental on this machine.
- `ruff check`, `ruff format --check` and `mypy` clean on both files.

One behaviour change worth naming rather than leaving to be found: a malformed non-text part that previously raised while being discarded now gets discarded quietly, since it is dropped before conversion. Unknown part types still reach `extract_all_content` and still raise `Unknown content`. If you would rather keep the strict read there, I am happy to narrow this to `input_image` alone.

`litellm_model.py` and `any_llm_model.py` both pass `preserve_tool_output_all_content=True`, so they take the other branch and are unaffected.

### DIFF
--- a/src/agents/models/chatcmpl_converter.py
+++ b/src/agents/models/chatcmpl_converter.py
@@ -75,6 +75,12 @@ ResponseInputContentWithAudioParam = (
 
 _OMITTED_TOOL_OUTPUT_PLACEHOLDER = "[tool output omitted]"
 
+# Content part types that never convert to a text part, so they are dropped from a tool
+# output before conversion when only the text is being kept.
+_NON_TEXT_TOOL_OUTPUT_CONTENT_TYPES = frozenset(
+    {"input_image", "image_url", "input_file", "input_audio", "video_url"}
+)
+
 
 class Converter:
     @classmethod
@@ -872,7 +878,20 @@ class Converter:
                 if preserve_tool_output_all_content:
                     tool_result_content = cls.extract_all_content(output_content)
                 else:
-                    all_output_content = cls.extract_all_content(output_content)
+                    keepable_output_content = output_content
+                    if not isinstance(output_content, str):
+                        # Parts that cannot become a text part are dropped here rather than
+                        # converted and then filtered out, so an unrepresentable one (an image
+                        # carrying file_id, say) omits itself instead of failing the turn.
+                        keepable_output_content = [
+                            c
+                            for c in output_content
+                            if not (
+                                isinstance(c, dict)
+                                and c.get("type") in _NON_TEXT_TOOL_OUTPUT_CONTENT_TYPES
+                            )
+                        ]
+                    all_output_content = cls.extract_all_content(keepable_output_content)
                     if isinstance(all_output_content, str):
                         tool_result_content = all_output_content
                     else:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this. The conversion failure is clear, and file-ID image outputs are used by downstream libraries. What is still missing is a concrete workflow where those outputs reach the default Chat Completions path and omitting the image is the desired outcome.

Could you share that workflow, including why it uses Chat Completions and what useful result remains after the image is omitted? The related image-tool reports require the model to actually see the image, which this change would not address.

dfedoryshchev:
Good point - I do not have one for the image-only case, so I have narrowed this rather than argue it.

Checking first: Chat Completions has no file-id image content part at all, only `image_url` with a URL or data URI. So nothing here can make the model see that image, and you are right that the image-tool reports need a different answer.

What I can point at is text sitting alongside the image. The docs sanction returning "one or many" outputs from a tool, so a tool that returns a short summary plus a rendered chart is ordinary usage - and today the conversion dies on the image and takes the summary with it. The model loses something it could have used, because of a part that was going to be dropped anyway. That is the useful result that remains.

Image-only output now keeps raising, unchanged. For a file id specifically that seems better than a placeholder: the id is an uploaded-file reference, so the error points the caller at a model that can read it instead of quietly degrading the run. There is a test pinning that case so the asymmetry is deliberate rather than incidental.

`566de96`, now 23 lines of source. One side effect worth naming: in mixed output a malformed non-text part is dropped rather than raising, since it is discarded either way - unknown part types still raise.

seratch:
Thanks for clarifying and narrowing the scope. The conversion failure is clear, but the summary-plus-chart example still does not establish a concrete workflow that needs this behavior: why it uses Chat Completions, why omitting the image is acceptable, and why returning text from the tool is insufficient.

I am going to close this PR for now. We can revisit it with a concrete workflow demonstrating those constraints and the useful outcome this change enables.

PR #4938, opened 2026-09-09
fix(sessions): skip malformed encrypted envelopes

### LINKED ISSUE #4936: EncryptedSession crashes on malformed envelope payloads
## Bug

`EncryptedSession` identifies an encrypted envelope from its marker and required keys, then unconditionally calls `.encode()` on `payload`. A malformed persisted envelope whose payload is not a string raises `AttributeError`; a string containing an unpaired Unicode surrogate raises `UnicodeEncodeError`.

Both failures abort `get_items()` and hide otherwise valid history. In `pop_item()`, the corrupt tail is removed by the underlying session, but the exception prevents the wrapper from continuing to the next valid item.

## Reproduction

```python
import asyncio
from typing import cast
from cryptography.fernet import Fernet
from agents import SQLiteSession, TResponseInputItem
from agents.extensions.memory.encrypt_session import EncryptedSession

async def main():
    backend = SQLiteSession("repro", ":memory:")
    session = EncryptedSession("repro", backend, Fernet.generate_key().decode())
    malformed = cast(TResponseInputItem, {
        "__enc__": 1,
        "v": 1,
        "kid": "hkdf-v1",
        "payload": 7,
    })
    await backend.add_items([malformed])
    await session.add_items([{"role": "user", "content": "valid"}])
    print(await session.get_items())

asyncio.run(main())
```

Current result:

```text
AttributeError: 'int' object has no attribute 'encode'
```

Expected result: the malformed envelope is isolated like an invalid or expired token, and the valid item is returned.

## Impact

A single malformed row in an underlying database can make an encrypted conversation unreadable. The session persistence contract says corrupt records must not hide valid history, and the wrapper already skips invalid Fernet tokens; malformed envelope payloads currently bypass that recovery path.

hsusul:
This pull request fixes encrypted session reads when a persisted envelope has a malformed payload. It resolves #4936.

## Root cause

`EncryptedSession._unwrap()` recognized an envelope by its marker and keys, then called `.encode()` on `payload` without validating its stored type. Non-string values raised `AttributeError`, and strings containing invalid Unicode scalar data raised `UnicodeEncodeError`. Either exception aborted `get_items()` and hid valid history; `pop_item()` removed the corrupt tail but stopped before returning the next valid item.

## Solution

Validate that the envelope payload is a string before decryption, and treat Unicode encoding/decoding failures as record corruption alongside invalid Fernet tokens and invalid decrypted JSON. The exception boundary stays narrow so configuration errors such as an invalid TTL still propagate.

## Regression coverage

The new tests exercise both integer and unpaired-surrogate payloads through:

- `get_items(limit=1)`, including backfill to a later valid record
- `pop_item()`, including continuation to the preceding valid record
- invalid TTL propagation, proving configuration `TypeError` is not mistaken for corrupt data

The tests fail on `upstream/main` and pass with this patch.

## Validation

- `ruff format --check` on changed files: passed
- `ruff check` on changed files: passed
- `pytest -q tests/extensions/memory/test_encrypt_session.py`: 43 passed
- repository formatting and lint: passed
- mypy: passed (312 source files)
- pyright: passed (0 errors, 0 warnings)
- full parallel tests: 9,564 passed, 56 skipped; 14 unrelated `tests/test_code_change_verification_runner.py` process-harness cases timed out under xdist. The affected test file passes independently.
- `git diff --check`: passed

## Risk assessment

Low. Valid encrypted envelopes, plaintext passthrough items, TTL handling, and the persisted envelope format are unchanged. The behavior change is limited to malformed marked envelopes, which are now skipped consistently with invalid or expired encrypted tokens.

### DIFF
--- a/src/agents/extensions/memory/encrypt_session.py
+++ b/src/agents/extensions/memory/encrypt_session.py
@@ -208,11 +208,15 @@ class EncryptedSession(SessionABC):
         if not _is_encrypted_envelope(item):
             return cast(TResponseInputItem, item)
 
+        payload = item["payload"]
+        if not isinstance(payload, str):
+            return None
+
         try:
-            token = item["payload"].encode("utf-8")
+            token = payload.encode("utf-8")
             plaintext = self.cipher.decrypt(token, ttl=self.ttl)
             return cast(TResponseInputItem, _from_json_bytes(plaintext))
-        except (InvalidToken, KeyError):
+        except (InvalidToken, UnicodeError, json.JSONDecodeError):
             return None
 
     def _unwrap_valid_items(
--- a/tests/extensions/memory/test_encrypt_session.py
+++ b/tests/extensions/memory/test_encrypt_session.py
@@ -42,6 +42,13 @@ def _invalid_encrypted_envelope() -> TResponseInputItem:
     )
 
 
+def _malformed_encrypted_envelope(payload: Any = 7) -> TResponseInputItem:
+    return cast(
+        TResponseInputItem,
+        {"__enc__": 1, "v": 1, "kid": "hkdf-v1", "payload": payload},
+    )
+
+
 @pytest.fixture
 def agent() -> Agent:
     """Fixture for a basic agent with a scripted model."""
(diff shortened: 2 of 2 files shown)


seratch:
Thanks for the patch. The additional handling addresses the constructed payloads, but #4936 does not establish how they arise in supported use. The tests demonstrate the proposed behavior without resolving that evidence gap.

I'm closing this PR for now. We can revisit a focused fix if a concrete storage failure or migration demonstrates the need.

PR #4940, opened 2026-09-09
fix(sessions): reject SQLite JSON scalar items

### LINKED ISSUE #4939: SQLite sessions accept JSON scalars as history items
## Bug

The SQLite session implementations treat any syntactically valid JSON value as a `TResponseInputItem`. Persisted JSON scalars and arrays are therefore returned as model-facing history even though response input items are JSON objects.

A particularly damaging case is JSON `null`:

- `get_items(limit=1)` returns `[None]`, which fails when the runner or application handles the item as a response input object.
- `pop_item()` deletes the `null` row and returns `None`, incorrectly signaling that the session is empty instead of continuing to the preceding valid item.

The behavior affects `SQLiteSession`, `AsyncSQLiteSession`, and `AdvancedSQLiteSession`.

## Reproduction

```python
import asyncio

async def main():
    session = SQLiteSession("repro", ":memory:")
    await session.add_items([{"role": "user", "content": "valid"}])

        conn.execute(
            f"INSERT INTO {session.messages_table} (session_id, message_data) VALUES (?, ?)",
            (session.session_id, "null"),
        )
        conn.commit()

    print(await session.get_items(limit=1))  # [None]
    print(await session.pop_item())          # None; valid predecessor was not returned

asyncio.run(main())
```

## Expected behavior

A deserialized session item must be a JSON object. Other valid JSON shapes should follow the existing corrupt-record path: limited reads should backfill with valid objects, and pops should discard the corrupt tail and continue to the next valid item.

## Impact

A scalar row introduced by corruption, a manual repair, or a migration can poison model input or make rollback logic stop early. This violates the session persistence rule that a corrupt record must not hide valid history.

hsusul:
This pull request fixes SQLite-backed sessions that accepted JSON scalars and arrays as model-facing history items. It resolves #4939.

## Root cause

The SQLite deserialization paths checked only whether `json.loads()` succeeded. Valid JSON values such as `null`, strings, numbers, booleans, and arrays therefore passed through as `TResponseInputItem` even though that public type is object-shaped.

For `null`, `get_items(limit=1)` returned `[None]`. `pop_item()` deleted the scalar tail and returned `None`, incorrectly signaling an empty session instead of continuing to the next valid object.

## Solution

After JSON decoding, require the result to be a dictionary in all `get_items()` and `pop_item()` paths for:

- `SQLiteSession`
- `AsyncSQLiteSession`
- `AdvancedSQLiteSession`

Non-object JSON now uses each backend's existing corrupt-row handling. Storage schemas, serialization, transaction boundaries, limits, ordering, and AdvancedSQLiteSession branch cleanup are unchanged.

## Regression coverage

Existing corrupt-row tests now run against both malformed JSON and valid JSON `null`. They cover limited backfill and pop recovery in the synchronous and asynchronous implementations, plus implicit/explicit branch reads and pop cleanup in `AdvancedSQLiteSession`.

The five `null` cases fail on `upstream/main` and pass with this patch.

## Validation

- SQLite-family test files: 226 passed
- changed-file Ruff format and lint: passed
- repository formatting and lint: passed
- mypy: passed (312 source files)
- pyright: passed (0 errors, 0 warnings)
- full parallel tests: 9,564 passed, 56 skipped; 14 unrelated `tests/test_code_change_verification_runner.py` process-harness cases timed out under xdist, matching the same baseline failure seen on the independent preceding branch
- `git diff --check`: passed

## Risk assessment

Low. Valid persisted JSON objects are returned exactly as before. Only values outside the public response-input-item shape are reclassified as corrupt, using the existing skip and cleanup paths. No database migration or persisted format change is required.

### DIFF
--- a/src/agents/extensions/memory/advanced_sqlite_session.py
+++ b/src/agents/extensions/memory/advanced_sqlite_session.py
@@ -391,8 +391,10 @@ class AdvancedSQLiteSession(SQLiteSession):
             for (message_data,) in rows:
                 try:
                     item = json.loads(message_data)
-                    items.append(item)
-                except json.JSONDecodeError:
+                    if not isinstance(item, dict):
+                        raise TypeError("Session item must be a JSON object")
+                    items.append(cast(TResponseInputItem, item))
+                except (json.JSONDecodeError, TypeError):
                     continue
             return items
 
@@ -545,7 +547,10 @@ class AdvancedSQLiteSession(SQLiteSession):
                             continue
 
                         try:
-                            return json.loads(message_row[0])
+                            item = json.loads(message_row[0])
+                            if not isinstance(item, dict):
+                                raise TypeError("Session item must be a JSON object")
+                            return cast(TResponseInputItem, item)
                         except (json.JSONDecodeError, TypeError):
                             # Drop corrupted JSON entries and keep looking for a valid item.
                             continue
--- a/src/agents/extensions/memory/async_sqlite_session.py
+++ b/src/agents/extensions/memory/async_sqlite_session.py
@@ -243,8 +243,10 @@ class AsyncSQLiteSession(SessionABC):
             for (message_data,) in rows:
                 try:
                     item = json.loads(message_data)
-                    items.append(item)
-                except json.JSONDecodeError:
+                    if not isinstance(item, dict):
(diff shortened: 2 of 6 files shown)


seratch:
https://github.com/openai/openai-agents-python/issues/4939#issuecomment-5609487444

PR #4943, opened 2026-09-10
fix(sessions): expand tilde in SQLite db paths

### LINKED ISSUE #4942: SQLiteSession crashes on ~/... db paths instead of expanding the home directory
Passing a `~/...` path as `db_path` to `SQLiteSession` (or `AsyncSQLiteSession`) crashes during init with `sqlite3.OperationalError: unable to open database file`.

The file-lock helper already expands the path (`Path.expanduser().resolve()`), but the actual sqlite connection uses the raw string, so `~/sessions.db` is treated as a literal relative path instead of a home-directory path.

Repro (no network needed):

```python
session = SQLiteSession("s1", db_path="~/sessions.db")
```

Expected: the database is created under the home directory. Actual: init raises `OperationalError`.

soroush5:
### Summary

`SQLiteSession` and `AsyncSQLiteSession` crashed on `~/...` db paths with `OperationalError`, because only the file-lock path was expanded. This normalizes `db_path` with `expanduser()` (except `:memory:`) so the lock and the connection agree.

### Test plan

- New `tests/memory/test_sqlite_db_path.py` covers both sessions with a fake `HOME` (failed before, passes now).
- Neighbors pass: `test_session.py`, `test_async_sqlite_session.py`, `test_advanced_sqlite_session.py` (223 passed).
- `ruff check` clean, `ruff format` applied.

### Issue number

Fixes #4942

### Checks

### DIFF
--- a/src/agents/extensions/memory/async_sqlite_session.py
+++ b/src/agents/extensions/memory/async_sqlite_session.py
@@ -54,7 +54,7 @@ class AsyncSQLiteSession(SessionABC):
             if session_settings is not None
             else SessionSettings()
         )
-        self.db_path = db_path
+        self.db_path = db_path if str(db_path) == ":memory:" else Path(db_path).expanduser()
         self.sessions_table = sessions_table
         self.messages_table = messages_table
         self._connection: aiosqlite.Connection | None = None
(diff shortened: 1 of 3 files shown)


seratch:
Thanks for the contribution. The outcome in #4942 is already supported by passing an explicitly expanded `Path` to either session class. We will keep path expansion caller-controlled and close this PR rather than introduce automatic filename reinterpretation.

PR #4945, opened 2026-09-10
fix(core): fall back to type name for nameless guardrail callables

### LINKED ISSUE #4944: Runner.run crashes with AttributeError when a guardrail function has no __name__

```python
import functools

```

soroush5:
Fixes #4944.

Guardrail span naming assumed every guardrail function has `__name__`, so a `functools.partial` or callable instance crashed the run with `AttributeError` before the guardrail even ran. This adds a small shared helper that falls back to the callable type name (same approach function tools already use) and applies it to all four guardrail classes plus the four guardrail decorators. Plain functions and explicit names behave exactly as before.

Test plan:
- Added 5 tests (4 in `tests/test_guardrails.py`, 1 in `tests/test_tool_guardrails.py`) covering `get_name`, the decorators, and an end-to-end `Runner.run` with a partial guardrail. All fail on the parent commit with `AttributeError` and pass with this change.
- `tests/test_guardrails.py` + `tests/test_tool_guardrails.py`: 87 passed.
- Neighbors `test_output_guardrail_cancellation.py`, `test_runner_guardrail_resume.py`, `test_stream_input_guardrail_timing.py`: 16 passed.
- `ruff check` and `ruff format --check` clean on all touched files.

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -10,7 +10,7 @@ from typing_extensions import TypeVar
 from .exceptions import UserError
 from .items import TResponseInputItem
 from .run_context import RunContextWrapper, TContext
-from .util._types import MaybeAwaitable
+from .util._types import MaybeAwaitable, _callable_name
 
 if TYPE_CHECKING:
     from .agent import Agent
@@ -106,7 +106,7 @@ class InputGuardrail(Generic[TContext]):
         if self.name:
             return self.name
 
-        return self.guardrail_function.__name__
+        return _callable_name(self.guardrail_function)
 
     async def run(
         self,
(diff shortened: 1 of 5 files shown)


seratch:
Thanks for investigating this. The missing-`__name__` failure is real, but partials and callable instances already work when the guardrail is given an explicit `name`. This applies to all four guardrail constructors and decorators in the current release.

For #4944's example, adding `name="threshold_check"` to `InputGuardrail(...)` avoids the failing lookup without changing execution behavior.

I'm going to close this PR because the existing naming option covers the reported use case. We can reconsider if there is a concrete workflow where supplying a name is insufficient.

PR #4946, opened 2026-09-10
fix(core): fall back to type name when guardrail function has no __name__

### LINKED ISSUE #4944: Runner.run crashes with AttributeError when a guardrail function has no __name__

shoemoney:
### Summary

Passing a `functools.partial` or a plain callable instance as a guardrail function crashes the run with `AttributeError: 'functools.partial' object has no attribute '__name__'`. `get_name()` is evaluated on the tracing span path before the guardrail runs (`with guardrail_span(guardrail.get_name())` in `src/agents/run_internal/guardrails.py:37` and `:49`), so the run dies before the guardrail is ever evaluated.

I verified all four surfaces the issue names, with different call sites for each:

- `InputGuardrail.get_name()` (`src/agents/guardrail.py:105-109`) and `OutputGuardrail.get_name()` (`:159-163`) had no fallback and are called before `guardrail.run()` at `src/agents/run_internal/guardrails.py:37` and `:49`. Reproduced end to end with `Runner.run()`.
- The `input_guardrail`/`output_guardrail` decorators (`src/agents/guardrail.py:261`, `:335`) accessed `f.__name__` directly, so calling them on a partial or callable instance crashed at construction time, not at run time.
- `ToolInputGuardrail.get_name()`/`ToolOutputGuardrail.get_name()` (`src/agents/tool_guardrails.py:168`, `:197` before this change) are not on the tool-call hot path, but they are called during run-state serialization (`src/agents/run_state.py:2891`, `_serialize_tool_guardrail_results`), which round-trips through `RunState` persistence, and its existing `hasattr(result.guardrail, "get_name")` guard doesn't help since the guardrail wrapper always has `get_name`.
- The `tool_input_guardrail`/`tool_output_guardrail` decorators (`src/agents/tool_guardrails.py:239`, `:275` before this change) had the identical construction-time crash.

### What changed

Applied the same `getattr(fn, "__name__", type(fn).__name__)` fallback used elsewhere in this codebase (`src/agents/tool.py:2521`, `fallback_name = type(func).__name__`, used by `function_tool()` for callable instances; also `src/agents/run_internal/tool_use_tracker.py` and `src/agents/agent_output.py`) to all six sites above: both `get_name()` methods on `InputGuardrail`/`OutputGuardrail`, both decorators in `src/agents/guardrail.py`, both `get_name()` methods on `ToolInputGuardrail`/`ToolOutputGuardrail`, and both decorators in `src/agents/tool_guardrails.py`.

One existing test, `tests/realtime/test_session_guardrails.py::test_output_guardrail_failure_tolerates_missing_callable_name`, asserted the old broken behavior (that `get_name()` raises for a nameless callable, forcing `_guardrail_diagnostic_extra`'s except-branch fallback). Updated its assertion to match the corrected behavior: `get_name()` now succeeds and returns the type name directly, so the diagnostic context carries `guardrail_name` instead of `guardrail_type`.

### Test plan

RED (against the unfixed code, on branch before the fix): added 13 new tests reproducing the crash and ran them, all failing with the exact reported `AttributeError`:

```
uv run pytest tests/test_guardrails.py tests/test_tool_guardrails.py -k "functools_partial or callable_instance" -v
```
7 failed in test_guardrails.py, 6 failed in test_tool_guardrails.py, all `AttributeError: '...' object has no attribute '__name__'`, including `test_run_with_functools_partial_input_guardrail_does_not_crash` failing inside `Runner.run()` at `run_internal/guardrails.py:37`, matching the issue's own repro.

GREEN (after the fix):
```
uv run pytest tests/test_guardrails.py tests/test_tool_guardrails.py -k "functools_partial or callable_instance" -v
```
`13 passed`

Full suite:
```
make tests
```
`9613 passed, 29 skipped` (parallel) and `77 passed, 4 skipped` (serial), including the one existing test I updated.

Lint and types:
```
make lint
make mypy
```
Both clean. Also ran the repo's own `.agents/skills/code-change-verification/scripts/run.sh`, which reported "all commands passed."

### Issue number

Fixes #4944

### Checks

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -106,7 +106,7 @@ class InputGuardrail(Generic[TContext]):
         if self.name:
             return self.name
 
-        return self.guardrail_function.__name__
+        return getattr(self.guardrail_function, "__name__", type(self.guardrail_function).__name__)
 
     async def run(
         self,
@@ -160,7 +160,7 @@ class OutputGuardrail(Generic[TContext]):
         if self.name:
             return self.name
 
-        return self.guardrail_function.__name__
+        return getattr(self.guardrail_function, "__name__", type(self.guardrail_function).__name__)
 
     async def run(
         self, context: RunContextWrapper[TContext], agent: Agent[Any], agent_output: Any
@@ -258,7 +258,7 @@ def input_guardrail(
         return InputGuardrail(
             guardrail_function=f,
             # If not set, guardrail name uses the function’s name by default.
-            name=name if name else f.__name__,
+            name=name if name else getattr(f, "__name__", type(f).__name__),
             run_in_parallel=run_in_parallel,
         )
 
@@ -332,7 +332,7 @@ def output_guardrail(
         return OutputGuardrail(
             guardrail_function=f,
             # Guardrail name defaults to function's name when not specified (None).
-            name=name if name else f.__name__,
+            name=name if name else getattr(f, "__name__", type(f).__name__),
         )
 
     if func is not None:
--- a/src/agents/tool_guardrails.py
+++ b/src/agents/tool_guardrails.py
@@ -165,7 +165,9 @@ class ToolInputGuardrail(Generic[TContext_co]):
     """
 
     def get_name(self) -> str:
-        return self.name or self.guardrail_function.__name__
+        if self.name:
+            return self.name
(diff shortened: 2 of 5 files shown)


seratch:
@codex review

seratch:
https://github.com/openai/openai-agents-python/issues/4944#issuecomment-5625439150

PR #4947, opened 2026-09-10
fix(core): deliver tool-not-found output on server-managed resume

hyeonsang010716:
### Summary

A single model response can pair a call that needs approval with a call to a tool that is not
registered. With `RunConfig(tool_not_found_behavior="return_error_to_model")` the runner answers the
unknown call locally and then interrupts for the approval, so that error output exists but has never
been sent. Resuming under `conversation_id`, `previous_response_id`, or `auto_previous_response_id`
never delivers it, and the server keeps a function call that has no output. The same run without
server-managed continuation delivers both outputs, so the two paths disagreed.

`get_unsent_tool_call_ids_for_interrupted_state` enumerated only the pending tool-run groups on the
last `ProcessedResponse`. Those groups never include `function_tools_not_found`, so
`hydrate_from_state` recorded the locally built output as already delivered and `prepare_input`
dropped it from the delta.

Adding `function_tools_not_found` to that enumeration is not sufficient: `_serialize_tool_action_groups`
does not persist the group, so it would fix only an in-memory resume and still lose the output after a
`RunState` round trip. The interruption branch is now seeded from the tool call IDs of the latest model
response, which is the rule the `NextStepRunAgain` branch already applies and the response
`resolve_interrupted_turn` already treats as authoritative.

Two properties worth confirming during review. The set only grows, so no output can be dropped, and
nothing server-owned can be resent: `hydrate_from_state` records every item the response reported into
`server_item_ids` and `server_tool_call_ids` before it consults this set, and `prepare_input` checks
those first. `RunState` serialization is unchanged, so states written by earlier versions resume as
before.

The existing tool-run enumeration is retained. It is redundant for every state the tests reach, since
`ProcessedResponse` is classified from the same response, but removing it would narrow released
behavior for pending runs without evidence, so it stays out of this fix. Happy to drop it in a
follow-up if you would rather have the single source of truth.

### Test plan

- `tests/test_run_impl_resume_paths.py`: `test_resumed_server_managed_run_sends_tool_not_found_output`
  covers streamed and non-streamed runs, live and JSON round-tripped `RunState`, and both
  `auto_previous_response_id` and `conversation_id`, for eight cases in total. It asserts the delivered
  call IDs as a list so a duplicate delivery fails it, and asserts the payload names the missing tool.
- `tests/test_run_impl_resume_paths.py`: `test_resumed_server_managed_run_sends_each_tool_output_once`
  stages two approvals, so the same model response stays current across both resumes, and asserts each
  output reaches the model exactly once. It also asserts the middle resume issues no model request.
- `tests/test_server_conversation_tracker.py`: a tracker-level case where the interrupted
  `ProcessedResponse` carries no pending tool run for the missing call, which is the shape a resumed
  state actually has. The pre-existing interrupted-turn test keeps pinning the enumeration path; its
  stand-in state gained `_model_responses`, which every real interrupted `RunState` has.
- All 10 new cases fail on `main` and pass with the change.
- Parallel suite 8557 -> 8567 passed, 178 skipped, with the same 12 pre-existing sandbox failures an
  unpatched tree produces on Windows without symlink privilege (#4852). Serial suite: 77 passed.
- `ruff format --check`, `ruff check`, `check_optional_truthiness.py`, `mypy src` (312 files) and
  `pyright` are all clean.
- `tests/test_run_state.py` is in the Windows `collect_ignore` list in `tests/conftest.py`, so it does
  not run on this host. With its Unix sandbox import stubbed, its 445 tests pass locally, and CI covers
  it on Linux.
- The verification script itself could not run: `make` is not installed on this host, and `make tests`
  would not pass here anyway while those symlink cases fail. The `Makefile` targets were run directly
  in the required order instead: format, lint, typecheck, then the parallel and serial suites.

### Issue number

None.

### Checks

### DIFF
--- a/src/agents/run_internal/agent_runner_helpers.py
+++ b/src/agents/run_internal/agent_runner_helpers.py
@@ -205,19 +205,24 @@ def _extract_tool_call_id(raw: Any) -> str | None:
     return candidate if isinstance(candidate, str) else None
 
 
+def _latest_response_tool_call_ids(run_state: RunState[Any]) -> set[str]:
+    """Return the call or item identifiers carried by the most recent model response."""
+    if not run_state._model_responses:
+        return set()
+    return {
+        call_id
+        for item in run_state._model_responses[-1].output
+        if (call_id := _extract_tool_call_id(item)) is not None
+    }
+
+
 def get_unsent_tool_call_ids_for_interrupted_state(run_state: RunState[Any] | None) -> set[str]:
     """Return tool call IDs whose local outputs have not reached a server conversation."""
     if run_state is None:
         return set()
 
     if isinstance(run_state._current_step, NextStepRunAgain):
-        if not run_state._model_responses:
-            return set()
-        return {
-            call_id
-            for item in run_state._model_responses[-1].output
-            if (call_id := _extract_tool_call_id(item)) is not None
-        }
+        return _latest_response_tool_call_ids(run_state)
 
     if not isinstance(run_state._current_step, NextStepInterruption):
         return set()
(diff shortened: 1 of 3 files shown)


joinwell52-AI:
@hyeonsang010716 This fix inspired a small engineering experiment on what to check when a resumed run appears to finish normally. We reran the ten added regressions at `3f9397f`: all passed. With the rest of the runtime and tests unchanged, substituting only the unsent-call helper from `83c737f` made all ten fail. In the eight Runner cases, the final `done` assertion still passed before the call-ID assertion exposed the missing output. That contrast helped us understand why inspecting the next model input matters, beyond checking the final response.

These were local ScriptedModel experiments, so our observation stops at request construction. The staged-approval case establishes one inclusion per output in that input; it does not tell us what happens if a server accepts a request but its acknowledgement is lost. For a follow-up fault-injection experiment, is there an existing SDK test boundary you would recommend for that ambiguous-acknowledgement case, or is it best exercised separately at the transport/server boundary?

[Reproduction scripts, passing/failing logs, and limitations](https://github.com/joinwell52-AI/joinwell52/tree/e09cf423/research/manual-runs/2026-09-11-current-authority).

PR #4948, opened 2026-09-10
fix(sandbox): reject archives needing unavailable symlink privilege

### LINKED ISSUE #4852: Sandbox tar extraction and 12 sandbox tests fail on Windows without symlink privilege

MohammedAlkindi:
**Summary**

`safe_extract_tarfile()` creates symlink members last, but `os.symlink()` was unguarded: on a non-elevated Windows host without Developer Mode it raises `OSError: [WinError 1314]` after files are already written. This adds a Windows-only capability probe before any member is written, so an archive whose symlinks this process cannot create is rejected with nothing written. An unsafe-archive error still outranks it, so such a host reports why the archive is unsafe instead of masking it. Archives without symlinks are unaffected; symlink creation still runs last and every containment check is untouched.

**Test plan**

Non-elevated Windows 11, Developer Mode off. Before: `WinError 1314` mid-extraction, files already on disk. After: rejected up front, nothing written.

- `pytest tests/sandbox/test_tar_utils.py`: 4 new tests pass. Failing set identical to base by node id, not count: 7 before, 7 after, none added or fixed. Those 7 create real symlinks, which this host cannot.
- `ruff format --check`, `ruff check`, `mypy src`, `mypy --platform win32 src`: pass

**Issue number**

Fixes #4852 (product half only). The test-harness half landed in #4853.

**Checks**

- [ ] `run.sh` not run: it is POSIX-only (`os.setsid`). Ran its ruff, mypy and pytest steps directly; `make tests` not run.

### DIFF
--- a/src/agents/sandbox/util/tar_utils.py
+++ b/src/agents/sandbox/util/tar_utils.py
@@ -365,6 +365,70 @@ def validate_tar_bytes(
         raise UnsafeTarMemberError(member="<tar>", reason="invalid tar stream") from e
 
 
+def _reject_symlink_leaves_over_existing_dirs(
+    members: list[tarfile.TarInfo], *, root: Path
+) -> None:
+    """Reject symlink members whose destination is an existing directory.
+
+    ``_prepare_replaceable_leaf`` already raises this, but only once the symlink
+    pass reaches that member - after :func:`_raise_if_symlinks_unsupported` has
+    run. This read-only pre-scan keeps the archive-level diagnostic ahead of the
+    host-capability one, so a host that cannot create symlinks still reports why
+    the archive is unsafe rather than masking it. It writes nothing.
+    """
+
+    for member in members:
+        if not member.issym():
+            continue
+        rel_path = safe_tar_member_rel_path(member, allow_symlinks=True)
+        if rel_path is None:
+            continue
+        dest = root / rel_path
+        if dest.is_dir() and not dest.is_symlink():
+            raise UnsafeTarMemberError(
+                member=member.name,
+                reason=f"destination directory already exists: {rel_path.as_posix()}",
+            )
+
+
+def _raise_if_symlinks_unsupported(members: list[tarfile.TarInfo], *, root: Path) -> None:
+    """Fail before anything is written when the archive needs symlinks this process
+    cannot create.
+
+    A non-elevated Windows process without Developer Mode enabled cannot call
+    ``os.symlink``: it raises ``OSError`` with ``winerror == 1314`` ("A required
+    privilege is not held by the client"). Symlink members are deliberately the last
+    thing ``safe_extract_tarfile`` writes, so probing for this once up front - before
+    any directory or file has been restored - means a host that cannot satisfy the
+    archive rejects it outright instead of aborting partway through extraction.
+    """
+
+    if os.name != "nt":
+        # The privilege this guards against is Windows-only, and the probe costs a
+        # temp directory plus a symlink on every extraction of a symlink-bearing
+        # archive (venv-style symlinks are the common case here).
+        return
+    if not any(member.issym() for member in members):
+        return
+
+    # Probe inside *root* rather than the system temp directory: the privilege is
+    # process-wide, but a probe on a different filesystem could pass while the real
+    # extraction target rejects symlinks, reintroducing the partial state this
+    # guards against.
+    with tempfile.TemporaryDirectory(dir=root, prefix=".agents-symlink-probe-") as probe_dir:
+        try:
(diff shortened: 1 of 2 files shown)


seratch:
Thanks for investigating this. The partial extraction behavior is real when calling `safe_extract_tarfile()` directly, but its only SDK runtime caller is UnixLocal workspace hydration, and that backend explicitly rejects Windows. The documented Windows alternatives use Docker or hosted providers; Docker extracts inside the container and does not use this host-side helper.

I recommend closing this PR because the current evidence does not establish a supported Windows workflow that needs the added capability probe. The remaining test portability work can be handled separately. We can revisit the runtime change if a concrete application workflow demonstrates reliance on this utility on Windows and a meaningful consequence from partial extraction.

PR #4949, opened 2026-09-10
fix guardrails: support partial callables

### LINKED ISSUE #4944: Runner.run crashes with AttributeError when a guardrail function has no __name__

fszcd:
Closes #4944.

`InputGuardrail`, `OutputGuardrail`, `ToolInputGuardrail`, and `ToolOutputGuardrail` assumed their callable always had `__name__`. That made `functools.partial` and callable instances crash before the guardrail could run.

Guardrail names now fall back to the callable type name, matching the existing function-tool behavior. Tests cover both direct name lookup and a real `Runner.run()` call with a partial input guardrail.

### DIFF
--- a/src/agents/guardrail.py
+++ b/src/agents/guardrail.py
@@ -106,7 +106,7 @@ class InputGuardrail(Generic[TContext]):
         if self.name:
             return self.name
 
-        return self.guardrail_function.__name__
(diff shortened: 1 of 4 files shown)


seratch:
https://github.com/openai/openai-agents-python/issues/4944#issuecomment-5625439150

PR #4953, opened 2026-09-11
test: cover config coercion, custom-data, and provider cleanup lifecycle paths

GodBlf:
### Summary

Adds regression tests for caller-visible error and lifecycle paths in core modules that had no coverage: `src/agents/_config_coercion.py`, `src/agents/util/_custom_data.py`, `src/agents/run_internal/model_provider_lifecycle.py`, and `src/agents/_httpx_compat.py` (85% / 89% / 86% / 71% before; three of the four now at 100%). All tests exercise public boundaries (`RunConfig`, `Runner.run`, `@function_tool`) and use `ScriptedModel` where possible; the provider-cancellation test uses a minimal blocking `Model` double because scripted steps cannot suspend a non-streamed run at a controlled await point, with that boundary documented in the test.

Covered behaviors:

- Unknown and non-dictionary typed settings are rejected with actionable `TypeError` messages across `RunConfig`, `Runner.run`, and sandbox options.
- Custom-data extractor results that are non-mappings or have non-string keys are rejected, empty mappings normalize to `None`, and async extractors are awaited.
- A non-streamed run cancelled twice while the runner-owned model provider close is in flight still completes the close and re-raises `CancelledError`.
- The legacy-httpx compat helpers degrade gracefully when httpx is missing, without importing it as a side effect.
- `SessionSettings` dictionary merge semantics: an empty dictionary passed through `RunConfig(session_settings={})` overlays rather than replaces the session's configured limit.

### Test plan

- `make format`, `make lint`, and `make pyright` pass with the change.
- `make tests-parallel`: 9583 passed, 63 skipped.
- `make tests-serial`: 77 passed, 4 skipped.
- `make coverage`: total coverage stays at 92% while the four target modules reach 100% / 100% / 100% / 89%.
- `make mypy` reports 5 pre-existing errors in `src/agents/sandbox/session/archive_ops.py` and `src/agents/run_internal/run_loop.py`; the same errors reproduce on a clean checkout of `main` with this change stashed, so they are unrelated to this PR.

### Issue number

N/A

### Checks

### DIFF
--- a/tests/test_agent_runner.py
+++ b/tests/test_agent_runner.py
@@ -2142,6 +2142,42 @@ class _RevokedReasoningIdModel(ScriptedModel):
 
 
 @pytest.mark.asyncio
+async def test_session_settings_dictionary_merge_overrides_runner_session_limit() -> None:
+    """A `SessionSettings` dictionary passed to `Runner.run` overlays the per-session
+    default instead of silently replacing it: fields the dictionary omits keep the
+    session's configured value, and omitted fields do not reset the base to None.
+    """
+    session = SQLiteSession("session-settings-merge", ":memory:")
+    await session.add_items(
+        [
+            {"role": "user", "content": "first"},
+            {"role": "assistant", "content": "first answer"},
+            {"role": "user", "content": "second"},
+            {"role": "assistant", "content": "second answer"},
+        ]
+    )
+    session.session_settings = SessionSettings(limit=1)
+    model = ScriptedModel()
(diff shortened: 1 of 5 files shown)


seratch:
Thanks for adding these tests. We are not taking coverage-only PRs unless they address a concrete regression or a demonstrated risk to an important supported behavior. This PR does not establish that need, so I am going to close it rather than ask for further revisions.

PR #4960, opened 2026-09-11
fix(models): treat blank OPENAI_DEFAULT_MODEL as unconfigured

sbguangha:
### Summary

`get_default_model()` used `os.getenv("OPENAI_DEFAULT_MODEL", "gpt-5.6-luna")`. When the variable is **set but blank or whitespace-only** (common in dotenv / CI), getenv does not fall back. The empty string then:

- becomes the model name sent to Responses / Chat Completions
- makes `is_gpt_5_default()` false
- drops the GPT-5 default `ModelSettings` (`reasoning.effort="none"`, `verbosity="low"`)

Docs say: if you do not set a custom model, you get `gpt-5.6-luna`. Same package already treats blank strings as unset in `_resolve_str`.

### Test plan

- New tests for `""` and `"   "` → `gpt-5.6-luna` and GPT-5 default settings.
- `uv run pytest tests/models/test_default_models.py tests/test_run_config.py -q` → 82 passed.
- `uv run ruff check` / `ruff format --check` on the touched files → passed.

Full `make tests` was not run (large suite). Focused model/default lanes cover this contract.

### Issue number

None. Existing-behavior bug; no `CONTRIBUTING.md`, `AGENTS.md` allows a focused fix + test.

### Checks

### DIFF
--- a/src/agents/models/default_models.py
+++ b/src/agents/models/default_models.py
@@ -97,10 +97,14 @@ def is_gpt_5_default() -> bool:
 
 
 def get_default_model() -> str:
+    """Return the default model name.
+
+    ``OPENAI_DEFAULT_MODEL`` overrides the built-in default only when it is set to a
+    non-empty value. Blank or whitespace-only values are treated as unconfigured.
     """
-    Returns the default model name.
-    """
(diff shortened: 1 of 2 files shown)


seratch:
For this env variable, we recommend doing `unset`.

PR #4979, opened 2026-09-12
fix: honor zero history limits in the FileSession example

JiangLLM:
### Summary

`FileSession.get_items(limit=0)` returns all saved messages instead of an empty list. The cause is `items[-limit:]`: when the limit is zero, Python treats it as `items[0:]`.

Return an empty list for zero. Positive limits still return the most recent messages in their original order. The saved history stays intact.

happy to discuss this fix!
### Test plan

- Add five cases for `0`, `1`, `2`, `5`, and `None`. Each test opens a saved session, checks the result, and confirms that the full history is still intact.
- The zero-limit case fails before the fix. All 20 focused tests pass after it.
- Format, lint, and type checks passed in the prior verification run.

### Issue number

None.

### Checks

A separate Codex agent reviewed the diff. The `/review` command was not run.

### DIFF
--- a/examples/memory/file_session.py
+++ b/examples/memory/file_session.py
@@ -47,7 +47,7 @@ class FileSession(Session):
         session_id = await self._ensure_session_id()
         items = await self._read_items(session_id)
         if limit is not None and limit >= 0:
-            return items[-limit:]
+            return items[-limit:] if limit > 0 else []
         return items
 
     async def add_items(self, items: list[Any]) -> None:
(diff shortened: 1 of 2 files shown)


seratch:
As mention at https://github.com/openai/openai-agents-python/pull/4904, we won't accept this change,

JiangLLM:
> As mention at #4904, we won't accept this change,

ok cool. thanks for point out!

PR #4989, opened 2026-09-13
fix(sessions): serialize Conversations mutations

### LINKED ISSUE #4988: OpenAIConversationsSession.pop_item is not safe for concurrent calls
### Describe the bug

`OpenAIConversationsSession.pop_item()` performs a non-atomic read-then-delete sequence without serializing concurrent calls on the same session instance. When two tasks call `pop_item()` concurrently, both can read the same newest item and both return that item. The next item remains in the conversation instead of being popped by the second call.

### Debug information
- Agents SDK version: `main` at `fbd2dbca`
- Related library versions: `openai 3.0.0`
- Python version: `3.13.14`
- Operating system: macOS
- Model and model provider: Not applicable; this is session storage behavior
- Does the issue reproduce with the latest Agents SDK release? The code path is present on `main`; release verification is not required for the deterministic repro.
- Does the issue occur consistently or intermittently? Consistently when both list requests overlap.

### Repro steps

The following fake Conversations client forces two list calls to observe the same snapshot before either delete completes:

```python
import asyncio


class Items:
    def __init__(self):
        self.rows = [
            {"id": "old", "role": "user", "content": "old"},
            {"id": "new", "role": "assistant", "content": "new"},
        ]
        self.reads_started = 0
        self.release_reads = asyncio.Event()
        self.deleted = []

    async def list(self, *, conversation_id, order, **kwargs):
        snapshot = list(reversed(self.rows))
        self.reads_started += 1
        if self.reads_started == 2:
            self.release_reads.set()
        await self.release_reads.wait()
        for row in snapshot:
            yield SimpleNamespace(model_dump=lambda exclude_unset=True, row=row: dict(row))

    async def delete(self, *, conversation_id, item_id):
        self.deleted.append(item_id)
        self.rows = [row for row in self.rows if row["id"] != item_id]


async def main():
    items = Items()
    client = SimpleNamespace(conversations=SimpleNamespace(items=items))
    session = OpenAIConversationsSession(conversation_id="conversation", openai_client=client)
    first, second = await asyncio.gather(session.pop_item(), session.pop_item())
    print({"first": first, "second": second, "deleted": items.deleted, "remaining": items.rows})


asyncio.run(main())
```

Observed output:

```text
{'first': {'id': 'new', 'role': 'assistant', 'content': 'new'}, 'second': {'id': 'new', 'role': 'assistant', 'content': 'new'}, 'deleted': ['new', 'new'], 'remaining': [{'id': 'old', 'role': 'user', 'content': 'old'}]}
```

### Expected behavior

Concurrent `pop_item()` calls on the same `OpenAIConversationsSession` instance should return distinct items in newest-first order, just like the local session implementations. The read-and-delete sequence should be serialized per session instance so each call observes the result of the previous deletion.

kaluli123123:
### Summary

This pull request resolves #4988 by serializing remote mutations within each `OpenAIConversationsSession` instance. Previously, concurrent `pop_item()` calls could read the same newest item before either delete completed, causing both callers to receive the same item and leaving the next item in history.

The change uses one instance-level mutation lock for conversation initialization, item creation, item deletion, and conversation deletion. All remote mutations use `_await_mutation()` so cancellation waits for the remote operation to settle before propagating. The class documentation explicitly limits this guarantee to one session instance; callers sharing a conversation ID across instances must coordinate externally.

### Test plan

- `uv run pytest tests/memory/test_openai_conversations_session.py -q` — 43 passed
- `uv run pytest tests/memory/test_session.py -q` — 53 passed
- `uv run ruff format` on changed files — passed
- `uv run ruff check` on changed files — passed
- `uv run pyright src/agents/memory/openai_conversations_session.py` — passed
- `uv run mypy src/agents/memory/openai_conversations_session.py` — passed
- `git diff --check` — passed
- `make tests` — 7831 passed, 52 skipped, 21 failed, 42 errors; the failures/errors are pre-existing environment and optional-dependency issues involving Docker, httpx, Runloop, voice extras, and release-contract baselines, not the changed files
- `.agents/skills/code-change-verification/scripts/run.sh` — format and lint passed; typecheck stopped on the same unrelated missing optional dependencies and existing Runloop/Vercel typing errors before the test step

### Issue number

Closes #4988

### Checks

- [x] If using Codex, I've run independent final review before submitting this PR


seratch:
Thanks for investigating this. The read-then-delete sequence can overlap, but #4988 currently demonstrates the result only with a fake client and does not identify an application workflow that requires concurrent pops. Ordered removals can already use sequential awaits, while multi-step corrections need coordination around the whole operation.

I would close this PR and #4988 for now. We can reconsider if there is a concrete workflow showing meaningful impact and why caller-side sequencing is insufficient. That would help establish the required concurrency boundary before expanding mutation locking and cancellation behavior.

PR #4991, opened 2026-09-13
fix(dapr): settle session clear cancellation

### LINKED ISSUE #4990: DaprSession.clear_session can leave metadata after cancellation
### Describe the bug

`DaprSession.clear_session()` deletes the messages state and metadata state in two sequential awaits, but does not settle the multi-step mutation when the caller is cancelled. If cancellation arrives after the messages key has been deleted and before the metadata delete runs, the method exits with a partially cleared session and leaves stale metadata behind.

This is inconsistent with the other session backends, which use the SDK's mutation-settling helper so a remote mutation completes before cancellation is propagated.

### Debug information
- Agents SDK version: `main` at `fbd2dbca`
- Related library versions: `dapr 1.16.0`
- Python version: `3.13.14`
- Operating system: macOS
- Model and model provider: Not applicable; this is session storage behavior
- Does the issue reproduce with the latest Agents SDK release? The code path is present on `main`.
- Does the issue occur consistently or intermittently? Consistently when cancellation is delivered between the two state deletes.

### Repro steps

```python
import asyncio
from agents.extensions.memory.dapr_session import DaprSession


class FakeClient:
    def __init__(self):
        self.state = {
            "s:messages": b"[]",
            "s:metadata": b'{"created_at":"old"}',
        }
        self.messages_deleted = asyncio.Event()
        self.release_messages = asyncio.Event()
        self.deleted = []

    async def delete_state(self, *, store_name, key, options=None):
        self.deleted.append(key)
        self.state.pop(key, None)
        if key == "s:messages":
            self.messages_deleted.set()
            await self.release_messages.wait()

    async def get_state(self, **kwargs):
        key = kwargs["key"]
        return SimpleNamespace(data=self.state.get(key, b""), etag=None)


async def main():
    client = FakeClient()
    session = DaprSession("s", state_store_name="store", dapr_client=client)
    clear_task = asyncio.create_task(session.clear_session())
    await client.messages_deleted.wait()
    clear_task.cancel("caller-cancelled")
    await asyncio.sleep(0)
    client.release_messages.set()
    try:
        await clear_task
        pass
    print({"deleted": client.deleted, "remaining_keys": sorted(client.state)})


asyncio.run(main())
```

Observed output:

```text
{'deleted': ['s:messages'], 'remaining_keys': ['s:metadata']}
```

### Expected behavior

`clear_session()` should settle both remote deletes before propagating caller cancellation. After the call completes, neither the messages key nor the metadata key should remain.

kaluli123123:
### Summary

This pull request resolves #4990 by making `DaprSession.clear_session()` cancellation-settled. Previously, cancellation after the messages state was deleted but before the metadata delete began left a stale metadata key behind.

The two state deletes now run inside the existing `_await_mutation()` helper. Caller cancellation is still propagated, but only after both deletes have completed. The change is limited to cancellation settling; Dapr state stores do not provide a transaction across the two independent keys, so ordinary provider failures can still require a separate recovery policy.

### Test plan

- `uv run pytest tests/extensions/memory/test_dapr_session.py -q` — 53 passed
- `uv run ruff format` on changed files — passed
- `uv run ruff check` on changed files — passed
- `uv run pyright src/agents/extensions/memory/dapr_session.py` — passed
- `uv run mypy src/agents/extensions/memory/dapr_session.py` — passed
- `git diff --check` — passed

### Issue number

Closes #4990

### Checks

- [x] If using Codex, I've run independent final review before submitting this PR


seratch:
Thanks for the focused patch. The added test demonstrates that both deletes settle after cancellation, but #4990 does not establish a meaningful downstream problem from the remaining metadata. Consistency with other backends alone does not justify extending the cancellation behavior here.

I'm closing this PR alongside #4990. We can reconsider a focused fix if a concrete affected workflow establishes the need.

PR #4992, opened 2026-09-13
perf(voice): release completed synthesis tasks between turns

rioyu123:
### Summary

This pull request releases completed TTS task references between turns in a voice pipeline.

`StreamedAudioResult` currently keeps every synthesis task until the result is discarded and passes the growing list to `asyncio.gather()` at each turn boundary. Long conversations repeatedly wait on tasks that have already finished.

Clear the list after a turn's synthesis succeeds. Failed or cancelled synthesis keeps the existing error and cleanup path. Audio ordering, lifecycle events, output text, and public APIs are unchanged.

### Test plan

- Added multi-turn regressions with tracing enabled and disabled, plus later-turn synthesis failure and consumer cancellation controls. Task completion is checked through independently retained task references.
- Full Voice suite: 228 passed on Linux / Python 3.14.4 with the full SDK and locked dependencies; 228 also passed on Windows.
- Repeated a synthetic full-SDK benchmark at 100, 500, 1,000, and 2,000 turns, with five segments per turn. At 2,000 turns, orchestration took about 3.3 s before and 0.28 s after; PCM digests, TTS calls, and lifecycle-event counts matched. This excludes model inference, network, and playback time, so it is not a claim about speech latency.
- Full repository verification passed: formatting, lint, Mypy, Pyright, and 9,691 tests; 33 skipped.

### Issue number

N/A

### Checks


seratch:
Thanks for the contribution. The completed-task accumulation and repeated gathering are real, and the proposed change is narrowly scoped. However, the reported improvement is cumulative orchestration time across 2,000 synthetic turns; it does not yet establish meaningful latency or memory impact in a realistic voice workload.

I would close this PR for now. We can revisit it with a representative session length and segment count showing material memory growth, event-loop delay, or CPU cost attributable to retained synthesis tasks. That would establish the practical need for this optimization.

PR #4995, opened 2026-09-13
fix(voice): finish transcription iterators when sessions close during setup

rioyu123:
### Summary

This pull request fixes a hang when an application closes a streamed transcription session during setup and then waits for its transcript consumer to finish.

After the WebSocket connects, the SDK sends `session.update` and waits for the response. Calling `close()` at this point stops the connection and background tasks, but `transcribe_turns()` can remain blocked because the event processor that normally signals completion has not started yet.

Signal completion from the close path as well as normal event processing, using one shared guard to avoid duplicate completion markers. Task cleanup, transcript delivery, and error selection remain in their existing paths. Public APIs and WebSocket messages are unchanged.

This addresses application-initiated `close()`, not the server-initiated setup disconnect discussed in #4585. It does not include that PR's setup error propagation changes.

### Test plan

- Added real loopback WebSocket regressions that wait until the server receives `session.update`, then close the session before `session.updated`. Both tracing modes reproduce the hang on the base and finish with the fix.
- Covered repeated close, active-session close, server close, provider errors, and consumer cancellation, including transcript delivery, owned-task completion, and tracing cleanup.
- Full Voice suite: 230 passed on Windows and Linux. The STT tests also passed with four parallel workers: 32 passed.
- Full repository verification passed on Linux / Python 3.14.4: formatting, lint, Mypy, Pyright, and 9,693 tests; 33 skipped. No live OpenAI speech calls were made.

### Issue number

N/A

### Checks


rioyu123:
Still relevant: current main still doesn't signal completion on the close path. I reran `tests/voice/test_openai_stt.py` (32 tests) against this PR's head and all pass. The change remains scoped to application-initiated close during setup. Could this stay open for review?

PR #4999, opened 2026-09-13
fix(core): read keyword-only docstring sections for tool parameter descriptions

### LINKED ISSUE #4985: Keyword Args and Other Parameters docstring descriptions are dropped from tool schemas
### Describe the bug

If a tool's docstring documents a parameter under a Google `Keyword Args:` (or `Keyword Arguments:`) section, or a NumPy `Other Parameters` section, that description never makes it into `params_json_schema`, and nothing warns. The sphinx `:keyword limit:` form works fine.

griffe puts those sections under `DocstringSectionKind.other_parameters`, but `generate_func_documentation` only reads `DocstringSectionKind.parameters`.

### Debug information

- Agents SDK version: 0.22.2 (also main at fbd2dbca)
- Related library versions: griffelib 2.0.1, pydantic 2.12.3
- Python version: 3.14.3
- Operating system: macOS
- Model and model provider: None, this only affects schema generation
- Does the issue reproduce with the latest Agents SDK release? Yes, 0.22.2

### Repro steps

```python
import json

@function_tool
def search_google(query: str, *, limit: int = 10) -> str:
    """Search the catalog.

    Args:
        query: What to search for.

    Keyword Args:
        limit: Max number of results to return.
    """
    return ""

@function_tool(docstring_style="numpy")
def search_numpy(query: str, *, limit: int = 10) -> str:
    """Search the catalog.

    Parameters
    ----------
    query : str
        What to search for.

    Other Parameters
    ----------------
    limit : int
        Max number of results to return.
    """
    return ""

@function_tool
def search_sphinx(query: str, *, limit: int = 10) -> str:
    """Search the catalog.

    :param query: What to search for.
    :keyword limit: Max number of results to return.
    """
    return ""

for t in (search_google, search_numpy, search_sphinx):
    print(t.name, "limit ->", json.dumps(t.params_json_schema["properties"]["limit"]))
```

Output:

```text
search_google limit -> {"default": 10, "title": "Limit", "type": "integer"}
search_numpy limit -> {"default": 10, "title": "Limit", "type": "integer"}
search_sphinx limit -> {"default": 10, "description": "Max number of results to return.", "title": "Limit", "type": "integer"}
```

### Expected behavior

All three tools should get `"description": "Max number of results to return."` on `limit`, the same way the sphinx one does.

Ayushraj06-bit:
### Summary

This is the change from #4993, which was closed together with #4985 because the report did not establish a workflow where the existing ways to supply parameter descriptions (an `Args:` / `Parameters` entry, or `Annotated` metadata) are insufficient. The stated bar for revisiting was an integration that needs to consume existing docstrings unchanged, where those alternatives impose a meaningful limitation. This PR is opened to meet that bar with a concrete case, not to relitigate the close. If it still does not meet it, I will leave it there.

**The case is third-party functions.** The tools guide says "You can use any Python function as a tool." When the function comes from a library, its docstring is consumed as-is: the caller cannot add an `Args:` entry or an `Annotated` hint without vendoring the source. Both alternatives require owning the code.

A concrete example from this repository's own `make sync` environment: `sounddevice.rec`, a direct dev dependency here, documents its `out` parameter under a NumPy `Other Parameters` section:

```
Other Parameters
----------------
out : numpy.ndarray or subclass, optional
    If *out* is specified, the recorded data is written into the
    given array instead of creating a new array.
    In this case, the arguments *frames*, *channels* and *dtype* are
    silently ignored!
```

That is precisely the caveat a model needs in order to use the parameter correctly. Passing the function to `generate_func_documentation` with no `docstring_style`, which is the default `function_tool` path:

```
main:         'out' -> None
this branch:  'out' -> 'If *out* is specified, the recorded data is written into the given array instead of creating a new array. In this case, the arguments *frames*, *channels* and *dtype* are silently ignored! ...'
```

Nothing warns. A user following the numpydoc or napoleon conventions ships a tool with a degraded schema and gets no signal that a workaround was needed. `Other Parameters` is a normative section in the numpydoc standard, and `Keyword Args:` / `Keyword Arguments:` are the Google-style sections napoleon documents, so this is an ordinary way to write a docstring: the same environment contains 37 such section headers across 12 packages (`google.protobuf.message`, `numpy`, `markdown`, `sounddevice`, and others).

A hand-written wrapper with its own docstring is the workaround, but it means re-documenting every parameter of every wrapped function by hand, which is the work docstring parsing exists to remove. For users who already write `Args:`, this change is a no-op.

**What changes.** griffe routes those sections to `DocstringSectionKind.other_parameters`; `generate_func_documentation` read only `parameters` and dropped them. Sphinx `:keyword x:` was unaffected because griffe's sphinx parser folds it into `parameters`, which is why only that form worked.

- Read both section kinds through the existing comprehension, so variadic star-stripping, `Annotated` precedence, and empty-description handling stay on one code path. `DocstringSectionOtherParameters` subclasses `DocstringSectionParameters`, so its `value` is a `list[DocstringParameter]` and the existing access is correct by construction.
- Teach `_detect_docstring_style` to count `:keyword` / `:key` for sphinx and `Other Parameters` for numpy. Without this, a tool whose parameters are all keyword-only scores zero, falls back to the google parser, and still loses the description on the default path. Google needs no new marker because it is already the zero-score fallback. Codex found this on #4993; it is not on the reporter's branch.
- Correct two comments that claimed `parameters` was the only kind consumed.

One intentional behavior change: a parameter documented in a keyword-only section *and* carrying an `Annotated` description now takes the docstring description, matching how `Args:` already resolves against `Annotated`. Covered by a test.

Credit for finding and diagnosing this belongs to @Om-singhaI (#4985); their branch carries the equivalent section-kind change.

### Test plan

`tests/test_function_schema.py` gains four tests:

- `test_keyword_only_param_descriptions_preserved` — parametrized over Google `Keyword Args:`, Google `Keyword Arguments:`, NumPy `Other Parameters`, and Sphinx `:keyword:`. The sphinx entry passes before the fix and is the control.
- `test_starred_keyword_only_param_descriptions_preserved` — griffe returns `**options` verbatim from keyword-only sections too, so the existing `.lstrip("*")` is exercised on the new path.
- `test_keyword_only_docstring_description_takes_precedence_over_annotated`.
- `test_keyword_only_sections_detected_without_style_override` — parametrized over all three styles with docstrings that contain nothing but the keyword-only section and **no** `docstring_style`. With the detection change reverted, the numpy and sphinx entries fail.

Locally, on this branch rebased onto current `main`:

- `make format-check`, `make lint` — clean
- `uv run mypy src` and `uv run mypy --platform win32 src` — no issues found in 312 source files
- `make tests` — 8588 passed, 178 skipped; serial 77 passed, 4 skipped

The upstream run on this PR has not executed any jobs (first-contributor approval gate). The identical workflow on my fork, same two commits, is here: https://github.com/Ayushraj06-bit/openai-agents-python/actions/runs/34793304556

One local caveat, confirmed identical on unmodified `main`: 12 `tests/sandbox/` tests fail on my Windows machine with `WinError 1314` because creating symlinks needs elevated privileges there. The `tests-windows` job on the fork run above covers those on a properly configured runner.

### Issue number

Fixes #4985

### Checks


seratch:
Thanks for the contribution. This carries forward the changes from #4993, but it does not provide new evidence addressing the reason we closed that PR and #4985. Parameter descriptions can already be supplied through ordinary Args/Parameters sections, including for keyword-only arguments, or through Annotated metadata.

I am going to close this PR for the same reason. Please avoid sending the same PRs further.

PR #5000, opened 2026-09-13
fix(sessions): reset compaction response chain on clear_session

sbguangha:
### Summary

`OpenAIResponsesCompactionSession.clear_session` cleared the underlying history but left `_response_id` and `_last_unstored_response_id` in place. A later `run_compaction` then reused a response id that pointed at the cleared conversation: in `previous_response_id` mode it forwarded the stale id to `responses.compact` instead of raising the documented "requires a response_id" error, and in `auto` mode a stale `_response_id == _last_unstored_response_id` match could still resolve to `input` from cleared state.

`pop_item` already invalidates the whole response chain after a deletion (#4934). `clear_session` deletes every item, a strictly stronger reset, so it now clears the same three fields (`_response_id`, `_deferred_response_id`, `_last_unstored_response_id`) on both the success and the failed-acknowledgement paths, matching `pop_item`.

### Test plan

- Added `test_clear_session_invalidates_response_chain`: it seeds a response id plus deferred compaction, clears the session, and asserts the next `previous_response_id` compaction raises "requires a response_id" instead of reusing the stale id. The test fails on `main` (`DID NOT RAISE`) and passes with the fix.
- `uv run pytest tests/memory/test_openai_responses_compaction_session.py` → 65 passed.
- `uv run ruff check`, `uv run ruff format --check`, and `uv run mypy` on the two changed files are clean.

### Issue number

N/A

### Checks


seratch (changes requested):
Thanks for the fix. Resetting the response chain after successful clears and uncertain backend outcomes is the right approach.

Before merging, please add regressions for a clear that commits before cancellation or an acknowledgement error reaches the caller. Assert that the original failure propagates, subsequent compaction cannot reuse the old response ID, and a subsequent write survives. Please also cover default `auto` mode using current local input after clearing a previously stored response. The new test currently covers only successful clearing; the existing settled-failure tests exercise `pop_item()`.

sbguangha:
Thanks for the review! Added the requested regressions in bc03568:

- `test_settled_clear_failure_invalidates_response_chain` (parametrized over cancellation and an acknowledgement error): the underlying deletion commits before the failure reaches the caller; the test asserts the original failure propagates, a forced compaction can no longer reuse the old response id, and a subsequent write survives — the session stays usable and a new response chain works.
- `test_run_compaction_auto_uses_input_after_clear`: default `auto` mode first compacts via `previous_response_id` for a stored response, then after `clear_session()` plus new local items it compacts from the current local input instead of reusing the cleared chain.

All three fail on `main` without the fix and pass with it.

sbguangha:
@seratch The regressions from the review are in bc03568:

- `test_settled_clear_failure_invalidates_response_chain` covers a clear that commits before cancellation or an acknowledgement error. The original failure still propagates, later compaction cannot reuse the old response id, and a subsequent write survives.
- `test_run_compaction_auto_uses_input_after_clear` covers default `auto` mode compacting from current local input after a previously stored response is cleared.

Please take another look when you have a chance.

PR #5001, opened 2026-09-14
fix(codex): clean up subprocesses after startup cancellation

rioyu123:
### Summary

This pull request fixes subprocess cleanup when prompt delivery is cancelled or fails during `CodexExec.run()` startup.

The subprocess and stderr reader are started before the prompt is written to stdin. That write and its `drain()` currently happen outside the cleanup block, so cancelling a run while stdin is backpressured can leave the child and stderr reader running after the caller has finished.

Move prompt delivery into the existing protected lifecycle. This builds on #4804 and leaves its teardown implementation unchanged. Output handling, public arguments, and the timing of the optional signal watcher remain unchanged.

### Test plan

- Added write/drain error regressions and a real Python subprocess test with a normal-completion control. The cancellation case checks that stdin remains buffered above its high-water mark before cancelling, then checks child exit, completed process wait, settled stderr reading, and closed pipes before fixture cleanup.
- On current main, the two error cases and real startup cancellation case fail while the normal control passes. With the fix, all 53 focused tests pass on Windows and Linux.
- The subprocess test uses a fixed local helper and synthetic input, with no Codex API calls or API keys.
- Full repository verification passed on Linux CPython 3.14.4 via `.agents/skills/code-change-verification/scripts/run.sh`: formatting, lint, Mypy, Pyright, and 9,717 tests passed; 33 skipped.
- Completed `codex review --base fbf59a40e9da5adb88d370fefaeaae0478376d4a` before submission with no actionable findings. Its read-only sandbox could not access the uv cache, so the test results above come from the separate verification run.

### Issue number

N/A

### Checks


rioyu123:
Thanks for taking a look. Is the startup stdin-cancellation cleanup handled by another change, or would you prefer a different approach? This PR moved prompt delivery into the protected lifecycle block from #4804. I will leave it closed unless you would like a follow-up.

Issue #5008, opened 2026-09-14
Tracing exporter honours OPENAI_API_KEY/ORG/PROJECT from the environment but not the endpoint, so OPENAI_BASE_URL does not redirect traces


xizhuomengcontin:
### Summary

`OPENAI_BASE_URL` redirects model traffic but not tracing. The tracing exporter reads `OPENAI_API_KEY`, `OPENAI_ORG_ID` and `OPENAI_PROJECT_ID` from the environment but **has no environment override for its endpoint** — so a deployment pointed at a self-hosted or gateway endpoint keeps posting spans, with the configured key, to `https://api.openai.com/v1/traces/ingest`. It fails non-fatally, so the only sign is one stderr line.

### Reproduction

```python
from agents import Agent, Runner
a = Agent(name="Demo", instructions="You are terse.", model="gpt-4o-mini")
r = await Runner.run(a, "Say hello in three words.")
```

Run with `OPENAI_BASE_URL` pointing at a local OpenAI-compatible server and `OPENAI_API_KEY=sk-test`:

```text
agents-sdk replied: the stub answered          ← model call went to the local endpoint
[non-fatal] Tracing client error 401. Response data is redacted.   ← this did not
```

The 401 is from OpenAI: the model call was redirected, the trace export was not.

### Where it comes from

`agents/tracing/processors.py`:

```python
_OPENAI_TRACING_INGEST_ENDPOINT = "https://api.openai.com/v1/traces/ingest"
...
endpoint: str = _OPENAI_TRACING_INGEST_ENDPOINT,
```

and the credential accessors right below it:

```python
return self._api_key or os.environ.get("OPENAI_API_KEY")
return self._organization or os.environ.get("OPENAI_ORG_ID")
return self._project or os.environ.get("OPENAI_PROJECT_ID")
```

So three OpenAI settings are environment-aware and the endpoint is not. `grep -rn "TRACING" agents/ | grep -i environ` finds only the **disable** switch in `provider.py`, nothing for the endpoint.

### Why this is worth more than a docs line

The asymmetry is the problem, not the default. Someone who sets `OPENAI_BASE_URL` has usually done it for a reason — a gateway, an on-prem model, a data-residency requirement. In every one of those cases the current behaviour sends span data **and the configured API key** to a host they just deliberately routed away from, and reports it as a non-fatal line that is easy to miss in a container log.

[...]

PR #5009, opened 2026-09-14
docs: explain custom tracing endpoint configuration

00200200:
### Summary

`OPENAI_BASE_URL` can redirect model requests while traces still go to the default OpenAI ingest endpoint. This pull request updates the tracing guide with the existing way to configure a separate destination using `BackendSpanExporter` and replace the default processor during application startup.

The example uses separate application-defined tracing credentials and explains receiver compatibility, per-run key precedence, organization/project header fallbacks, and why adding a processor keeps the default destination active. SDK behavior is unchanged.

### Test plan

- Executed the exact documentation snippet against this checkout and the `v0.22.2` source, with all HTTP calls mocked. Verified that the model base URL leaves the default trace endpoint unchanged, while the example exports one synthetic trace to the selected receiver with its own key and configured routing headers.
- `uv run --frozen pytest tests/test_trace_processor.py tests/tracing/test_processor_api_key.py -k 'custom_endpoint or test_processor' -q`: 6 passed, 53 deselected.
- `make build-docs`.
- Independent review of the complete diff, runtime claims, and translation ambiguity completed with no findings.

No live model or tracing API calls were made. The full SDK verification stack is not applicable to this documentation-only change under the contributor guide.

### Issue number

Related to #5008. This addresses the documentation portion; it does not add the suggested automatic environment override or warning.

### Checks

- [x] Verified the runnable example and affected existing tests.
- [ ] Ran `.agents/skills/code-change-verification/scripts/run.sh` (not applicable: documentation-only change).
- [x] Confirmed all applicable verification steps pass.
- [x] Completed independent review before submission.


seratch:
When a developer checks the arguments for `BackendSpanExporter` initialization, they will figure the way to configure the endpoint URL out.

PR #5012, opened 2026-09-14
Keep SQLite sessions on the original database

1fanwang:
### Summary

After changing directories, a SQLite session created with a relative filename can read or clear another database, or raise `sqlite3.OperationalError: no such table: agent_messages`.

Bind the filename at construction. In-memory sessions and literal `~` filenames keep their behavior.

### Test plan

Ran this against both versions with real SQLite files.

| Source | Command | Result |
| --- | --- | --- |
| Base | Inline program below | Missing-table error |
| Fix | Same program | Original database retained |

<details><summary>Reproducer and raw logs</summary>

```bash
PYTHONPATH=src python - <<'PY'
import asyncio
import os
import sqlite3
from contextlib import closing
from tempfile import TemporaryDirectory

original_cwd: str = os.getcwd()
try:
    with TemporaryDirectory() as root:
        try:
            os.chdir(root)
            os.mkdir("other")
            with closing(SQLiteSession(session_id="example", db_path="history.db")) as session:
                os.chdir("other")
                print("history:", asyncio.run(session.get_items()))
                print("second database created:", os.path.exists("history.db"))
        finally:
            os.chdir(original_cwd)
except sqlite3.OperationalError as error:
    print(f"sqlite3.OperationalError: {error}")
    raise SystemExit(1) from None
PY
```

Before:
```text
sqlite3.OperationalError: no such table: agent_messages
exit=1
```

After:
```text
history: []
second database created: False
exit=0
```

</details>

The full check has one unrelated docs-publisher failure on both versions: Git's `safe.bareRepository=explicit` rejects its bare-repo fixture. The SQLite and serial tests pass.

### Issue number

No issue filed.

### Checks

- [ ] I've confirmed all verification steps pass (Docs fixture failure above.)
- [ ] If using Codex, I've run `/review` before submitting this PR (Not applicable.)


seratch:
I don't think changing the working directory after referring to relative path to database file is a common use case. So, we won't add this additional logic to convert a relative path to its absolute path.

PR #5016, opened 2026-09-14
fix(core): preserve structured guardrail diagnostics and agent output during persistence

### LINKED ISSUE #5006: RunState stringifies structured guardrail diagnostics and agent output during persistence
### Please read this first

- **Have you read the docs?** Yes: the guardrail guide uses a Pydantic result as `GuardrailFunctionOutput.output_info`, and RunState is the durable checkpoint boundary.
- **Have you searched for related issues?** Yes. #3288 addressed non-JSON values making serialization fail. This report concerns structured models becoming opaque strings while serialization succeeds, not that original datetime exception.

### Describe the bug

Guardrail diagnostics containing a normal Pydantic model lose their structure when a run is saved and restored. The same happens to structured `OutputGuardrailResult.agent_output`. An audit consumer can read fields such as `allowed` and `reason` before persistence, but afterward only a Python repr string remains; JSON decoding does not recover the fields.

This occurs with the documented shape `GuardrailFunctionOutput(output_info=structured_result, ...)`, not only manually constructed RunState internals. The reproduction runs actual input/output guardrails and an Agent with structured output through `Runner.run()`, then saves `result.to_state().to_string()` and restores it with `RunState.from_json()`.

A checkpoint used for approval/resume has the same diagnostic persistence path. This report is about retaining audit data, not a claim that guardrail decisions stop enforcing policy.

### Debug information

- Checkout baseline: `fbf59a40`, SDK 0.22.2.
- Python 3.13.14, macOS arm64.
- Uses the SDK's documented `ScriptedModel` for deterministic SDK-owned orchestration; no model service, API credentials or network required.
- Reproduced consistently on the checkout. The release package was not separately installed for this report.

### Reproduction

```python
import asyncio
import json
from agents import Agent, Runner, RunConfig, RunState, GuardrailFunctionOutput
from agents.guardrail import InputGuardrail, OutputGuardrail

class Verdict(BaseModel):
    allowed: bool
    reason: str

class Answer(BaseModel):
    text: str

async def check(*args):
    return GuardrailFunctionOutput(output_info=Verdict(allowed=True, reason='approved'), tripwire_triggered=False)

async def main():
    agent = Agent(name='Audit', model=ScriptedModel([[assistant_message('{"text":"hello"}')]]), output_type=Answer, input_guardrails=[InputGuardrail(check)], output_guardrails=[OutputGuardrail(check)])
    result = await Runner.run(agent, 'hello', run_config=RunConfig(tracing_disabled=True))
    snapshot = result.to_state().to_string()
    restored = await RunState.from_json(agent, json.loads(snapshot))
    print('input verdict:', restored._input_guardrail_results[0].output.output_info)
    print('output verdict:', restored._output_guardrail_results[0].output.output_info)
    print('answer:', restored._output_guardrail_results[0].agent_output)
    assert restored._input_guardrail_results[0].output.output_info == {'allowed': True, 'reason': 'approved'}
    assert restored._output_guardrail_results[0].agent_output == {'text': 'hello'}

asyncio.run(main())
```

Before the fix, restored values are strings:

```text
input verdict: allowed=True reason='approved'
output verdict: allowed=True reason='approved'
answer: text='hello'
AssertionError
```

Expected plain JSON data:

```text
input verdict: {'allowed': True, 'reason': 'approved'}
output verdict: {'allowed': True, 'reason': 'approved'}
answer: {'text': 'hello'}
```

### Cause and expected behavior

The agent and tool guardrail serializers pass payloads directly to `_ensure_json_compatible`, which uses `json.dumps(..., default=str)`. Pydantic models and dataclasses therefore become repr strings. Tool result serialization already has `_serialize_output_value` to preserve these values as plain data.

Preserve normal model/dataclass fields in guardrail snapshots using the same conversion pipeline, including structured values nested in containers. There is no requirement to reconstruct the original Python classes. Existing JSON-native payloads, old snapshots, and best-effort fallback for values whose custom serialization fails should remain supported.

rksharma-owg:
### Summary

When serializing `RunState` to JSON, guardrail results (`GuardrailFunctionOutput.output_info` and `OutputGuardrailResult.agent_output`) were passed directly to `_ensure_json_compatible()`, which relies on `json.dumps(..., default=str)` for non-JSON native values. Consequently, structured Pydantic models and dataclasses degraded into opaque Python repr strings like `allowed=True reason='approved'`. Upon restoring via `RunState.from_json()`, these values remained raw strings instead of structured dictionaries.

This PR reuses the structured conversion pipeline (`_serialize_output_value`) used by tool outputs, converting models and dataclasses into plain JSON data while retaining a safe fallback if custom serialization raises.

Closes #5006

### Changes

- `src/agents/run_state.py`:
  - Added `_serialize_guardrail_payload(value: Any) -> Any` to normalize structured output and diagnostic values using `_serialize_output_value` before passing to `_ensure_json_compatible`.
  - Updated `_serialize_guardrail_results` and `_serialize_tool_guardrail_results` to use `_serialize_guardrail_payload` for `output_info` and `agent_output`.
- `tests/test_run_state.py`:
  - Added `test_runner_guardrail_models_survive_state_serialization` asserting Pydantic models in input/output guardrails and structured agent output round-trip as JSON dictionaries.
  - Added `test_tool_guardrail_dataclasses_survive_state_serialization` verifying dataclasses in tool guardrail diagnostics round-trip.
  - Added `test_guardrail_state_keeps_fallback_when_model_serializer_raises` verifying best-effort string fallback when a model serializer raises.

### Verification

- Passed all 449 tests in `tests/test_run_state.py`.
- Verified offline with `ScriptedModel` against the reproduction in #5006.
- Code linting and formatting checked with `ruff`.


seratch (approved):

rksharma-owg:
👏 thanks

PR #5029, opened 2026-09-14
Fail closed on invalid realtime approval settings

### LINKED ISSUE #5024: needs_approval misconfiguration raises in the core runner and silently skips approval in realtime sessions
### What happens

`evaluate_needs_approval_setting` in `src/agents/util/_approvals.py` is shared by the core runner and realtime sessions, and the two call sites disagree about what to do with a `needs_approval` value that is neither a bool nor a callable. The core runner raises. A realtime session treats it as "no approval needed" and runs the tool.

`_approvals.py:32-51`:

```python
if isinstance(needs_approval_setting, bool):
    return needs_approval_setting
    ...
if strict:
    raise UserError(
        f"Invalid needs_approval value: expected a bool or callable, "
        f"got {type(needs_approval_setting).__name__}."
    )
return default
```

`default` is `False`. `strict` defaults to `True`.

`run_internal/tool_execution.py:1312` calls it without `strict`, so it raises.

`realtime/session.py:662-667` passes `strict=False`, so it returns `False`. And at `realtime/session.py:730-731`:

```python
if approval_status is None and not needs_approval:
    return True
```

which proceeds with the call.

### Reproduction

Running the shipped helper, with only the `UserError` import stubbed:

```
needs_approval value               core runner (strict)   realtime (strict=False)
True                               needs_approval=True    needs_approval=True
False                              needs_approval=False   needs_approval=False
'always'                           raises UserError       needs_approval=False   <-- executes, no approval
'never'                            raises UserError       needs_approval=False   <-- executes, no approval
1                                  raises UserError       needs_approval=False   <-- executes, no approval
None                               raises UserError       needs_approval=False   <-- executes, no approval
{'tool': 'x'}                      raises UserError       needs_approval=False   <-- executes, no approval
```

Bool and callable agree, as they should. Everything else diverges.

### Why the misconfiguration is plausible

`FunctionTool` is a plain dataclass (`tool.py:449`) and `needs_approval` (`tool.py:494`) is not validated at construction, so nothing rejects the value where it is written. The only `isinstance` check on this path is inside `_approvals.py` itself.

`"always"` is the specific value worth worrying about. It reads like a valid setting, it is what several other libraries use for this exact option, and it is truthy, so a reader glancing at it would expect approval to be required. In a realtime session it means the opposite.

### Why it is worth reporting even though it needs a misconfiguration

The library already decided this input is an error. That is what the `UserError` is. The issue is that the same error is fatal on one path and silent on the other, so which of two behaviours you get depends on which runner you happen to use, and the quieter one is the permissive one.

The asymmetry also removes the feedback that would fix it. On the core runner the mistake is unmissable. In realtime it produces tool calls that run without approval and look exactly like tool calls that were correctly configured not to need it.

### Suggested shape

Either raise on both paths, or if realtime genuinely cannot raise there, default to `True` rather than `False` so an uninterpretable approval policy means approval is required. The second is a one-word change and matches how the same file already treats an unreadable argument list.

That last comparison is worth making explicitly, because the file gets it right immediately above: `parse_function_tool_arguments` returns `None` when arguments cannot be inspected, and both call sites treat `None` as "approval required". The unreadable-arguments case fails closed. The uninterpretable-policy case fails open. Same file, same question, opposite answers.

### Verification

Read against `main` on 2026-09-14. The transcript is the shipped `evaluate_needs_approval_setting` executed directly, not a retyped copy. I did not drive a live realtime session, so the call-site behaviour at `session.py:730` is read rather than executed.

DeepanshuPal:
### Summary

The realtime session passed `strict=False` to the shared `needs_approval` evaluator, so an invalid setting such as `"always"` became `False` and the tool ran without approval. The core runner already raises `UserError` for the same input.

Use the shared strict behavior in realtime too. Invalid approval settings now raise before the tool is invoked, matching the core runner and avoiding a fail-open path.

### Test plan

- Added a realtime regression test for an invalid `needs_approval` value.
- `uv run pytest tests/realtime/test_session_approvals.py -q` (31 passed)
- `uv run pytest tests/realtime -q` (520 passed)
- `uv run ruff check src/agents/realtime/session.py tests/realtime/test_session_approvals.py`
- `uv run pyright src/agents/realtime/session.py tests/realtime/test_session_approvals.py`

### Issue number

Closes #5024

### Checks


joinwell52-AI:
Thanks for this change. It prompted a small approval-boundary experiment for our agent runtime work. Using the real RealtimeSession tool-call entry point at b6c2cef, a recording model, and a local list-writing tool, we compared ten inputs with the predecessor approval method substituted into the same candidate source. Invalid top-level settings ("always", 1, None, {}) went from one invocation each to UserError with zero invocations. The candidate's original approval test file also passed all 31 tests; the method ablation failed the invalid-setting test.

One boundary the experiment clarified for us: a callback that returns None still leads to execution in both modes. That is deliberately outside the Boolean return contract, so we are not calling it a supported-input bug. Is coercion of callback results intentional here, or would validating that result at runtime be useful too? It seems worth distinguishing valid configuration type from valid computed approval result.

[Probe, pinned versions, and saved observations](https://github.com/joinwell52-AI/joinwell52/blob/0b7b0ae594b55b00ecb97f3d61c2b503d41f8821/research/manual-runs/2026-09-15-execution-facts/README.en.md)

seratch:
The behavior difference for values such as "always" is reproducible, but those values are outside the public needs_approval contract. The report and follow-up probe do not identify a supported producer that supplies them or an untrusted-input boundary that allows an attacker to change approval policy. Valid Boolean values and approval callbacks already express the supported behavior.

I would close this PR and #5024. Matching the core runner's handling of invalid typed values is not enough to establish a security fix. We can revisit this if a concrete configuration-ingestion path produces those values and the resulting approval consequence is demonstrated.

joinwell52-AI:
Understood. Our probe establishes the behavior difference for out-of-contract inputs, not a supported-input security issue. We haven't identified a configuration-ingestion path that supplies those values, so I agree the probe isn't grounds for a security fix. Thanks for making that boundary explicit.

PR #5039, opened 2026-09-15
docs: add OrcaReplay to the external tracing processors list

xizhuomengcontin:
## What

One line appended to **External tracing processors list** in `docs/tracing.md`.

**Disclosure: I maintain OrcaReplay.** Apache-2.0, free, no paid tier, no account, no hosted service.

## It is a tracing processor, which is what this list is for

`orcareplay-openai-agents` registers an `OrcaTracingProcessor` through the SDK's own API:

```python
from agents import add_trace_processor, set_trace_processors
from orcareplay_openai_agents import OrcaTracingProcessor

set_trace_processors([OrcaTracingProcessor()])   # what install() does
add_trace_processor(OrcaTracingProcessor())      # with ORCA_AGENT_SPANS_KEEP_EXPORT=1
```

It implements the [`TracingProcessor`](https://openai.github.io/openai-agents-python/ref/tracing/processor_interface/) surface — `on_trace_start`, `on_trace_end`, `on_span_start`, `on_span_end`, `shutdown`, `force_flush` — and writes agent, handoff and guardrail spans to disk. It is not a subclass of `agents.tracing.TracingProcessor`, and deliberately so: subclassing would force an SDK import at definition time, and the package has to stay importable, and inert, on a machine that has no SDK. It satisfies the protocol.

> **Correction to the first version of this description.** It said `add_trace_processor` was used rather than `set_trace_processors`, and that installing the package does not disable the OpenAI traces dashboard. Both were wrong, and I would rather say so here than have someone find it in the source. `install()` calls `set_trace_processors`, which **does** remove the SDK's own exporter. Measured, with the registration functions stubbed: one call, `('set', 1)`.

**Why it replaces rather than appends.** `orca record` installs this through a `sitecustomize` on `PYTHONPATH`, so it runs before the user's first statement — at which point the only registered processor is the SDK's own exporter. Leaving it in place means a second egress out of a run that is being recorded offline, and on a machine with no real tracing key it also fills the output with `Tracing client error 401`. Anything the user registers afterwards is appended and keeps working, and a user who calls `set_trace_processors` themselves replaces ours, which is their explicit choice.

**To keep the OpenAI dashboard as well**, one variable switches it to `add_trace_processor`:

```console
ORCA_AGENT_SPANS_KEEP_EXPORT=1 orca record generic-openai -- python your_agent.py
```

## What is written, and what is not

Bounded, metadata-only spans — three span types, and from each only the fields the reader consumes:

| span | fields kept |
| --- | --- |
| `AgentSpanData` | `name`, `handoffs`, `tools`, `output_type` |
| `HandoffSpanData` | `from_agent`, `to_agent` |
| `GuardrailSpanData` | `name`, `triggered` |

Everything else is dropped, including every other span type. Raw prompts, model output, tool arguments, tool results, `mcp_data`, `CustomSpanData` payloads, transcription and speech audio and `SpanError.data` never reach the file — an allow-list rather than a deny-list, so a field a future SDK version adds stays out without a code change. `ResponseSpanData` and `GenerationSpanData` are dropped for a second reason too: the proxy below already holds those exchanges byte for byte, so a second copy would be a lossier duplicate.

```console
pip install orcareplay-openai-agents
```

## What it is for, stated plainly

The rest of OrcaReplay records model traffic at a proxy and replays it later with the provider unreachable, so a past run reproduces without a key. That proxy sees HTTP and nothing else — which agent was speaking, which handoff fired, whether a guardrail ran are not on the wire. A handoff reaches it as a `transfer_to_<agent>` function tool, which names the destination and never the source; a guardrail that passes need make no request at all. This package supplies exactly that structure, so the recording carries the run's shape alongside its bytes.

Verified against openai-agents **0.20.0** and **0.22.2** — a clean `pip install` of the published package on an interpreter that had never seen the repository, then a two-agent run with a handoff and a guardrail, which produced `agent.start` ×2, `agent.handoff` and `agent.guardrail`.


seratch:
Thanks for the suggestion. Running an agent app using the record command is not the way we support, so let us hold off adding the link for now.

Issue #2540, opened 2026-02-24
Getting two traces with Human-in-the-loop


gemue-parndt:
### Question

Hey everyone, first of all, I love the HITL feature!

I've implemented HITL and use Phoenix as an open-source alternative to the OpenAI traces dashboard for viewing the traces. :)

**Here’s my workflow:**
1. The LLM generates text.
2. The LLM attempts to call the "write_page" tool, which requires approval.
3. The user clicks "approve".
4. The LLM executes the tool call.
5. The LLM returns its response.

Everything works well, but I notice that instead of a single trace, I end up with two traces:
- The first trace covers steps 1–2 and is ending with the tool call .
- The second trace includes steps 4–5 (step 3 isn’t shown, which makes sense since it’s not an LLM action) and is starting with the tool call.

Does anyone have suggestions or tips on how to combine these traces, or is this split the expected behavior?

hshahrokni2:
Tool call failures in Continue often come from model capability mismatches -- not all models support function calling, and the ones that do have different schemas for tool definitions.

The routing issue: if Continue is sending tool-formatted requests to a model that does not support them, you get silent failures or malformed responses instead of a useful error.

A routing layer can solve this by ensuring tool-call requests only go to models with confirmed tool support. Komilion routes tool-heavy requests to capable models automatically, falling back gracefully if needed. komilion.com if useful.

seratch:
Thanks for sharing this. This seems to be a bug, so we'll fix it in the upcoming release.

PR #5045, opened 2026-09-15
fix(tracing): keep trace state on completed non-streamed runs

BlueX888:
### Summary

A non-streamed run that finishes normally produces a `RunResult` whose `to_state()` carries no trace. Resuming from that state with `Runner.run(agent, state)` opens a new trace instead of reattaching to the original one, so a single logical run ends up split across two traces.

Root cause: `AgentRunner._run_impl` (`src/agents/run.py:584`) has several `RunResult` construction sites, and the completion path at `src/agents/run.py:2067-2070` was the only one that did not assign `result._trace_state = run_state._trace_state`. `RunResult` has no `trace` attribute, so the fallback in `result_to_run_state` (`src/agents/result.py:166-169`) resolves to `None` and `to_state()` writes a state without trace information. On resume, `create_trace_for_run(..., trace_state=run_state._trace_state, reattach_resumed_trace=is_resumed_state)` (`src/agents/run.py:2508`) then has nothing to reattach to and starts a fresh trace. The other construction sites (`src/agents/run.py:1426`, `src/agents/run.py:1607`, `src/agents/run.py:2590`) already copy it.

This is the remaining case of #2540, whose fix (#2547) covered the interrupted-run path but not the completed-run path. The symptom is not limited to human-in-the-loop: `Runner.run(agent, input=...)` followed by `Runner.run(agent, result.to_state())` also starts two traces on the current tree.

### Changes

- `src/agents/run.py:2070` — carry `run_state._trace_state` onto the result in the completion path, matching the other `RunResult` construction sites in the module.
- `tests/test_agent_tracing.py` — add `test_resumed_run_after_completed_turn_reuses_original_trace`.

### Test plan

The new test runs an agent with a `needs_approval` tool, approves the interruption, lets the run finish, then resumes from `approved.to_state()`. It asserts exactly one trace and one `trace_start`/`trace_end` pair.

With the fix reverted (test present, `src/agents/run.py` at the base revision) it fails:

```
    traces = fetch_traces()
>       assert len(traces) == 1
E       assert 2 == 1
E        +  where 2 = len([<agents.tracing.traces.TraceImpl object at 0x10c53ff50>, <agents.tracing.traces.TraceImpl object at 0x10c53fb60>])

tests/test_agent_tracing.py:450: AssertionError
=========================== short test summary info ============================
FAILED tests/test_agent_tracing.py::test_resumed_run_after_completed_turn_reuses_original_trace
============================== 1 failed in 0.28s ==============================
```

With the fix applied:

- `uv run pytest tests/test_agent_tracing.py::test_resumed_run_after_completed_turn_reuses_original_trace` -> `1 passed in 0.20s`
- `uv run pytest tests/test_agent_tracing.py` -> `34 passed in 0.52s`
- `uv run pytest tests/test_agent_tracing.py tests/test_tracing.py tests/test_tracing_errors.py tests/test_tracing_errors_streamed.py tests/test_run.py tests/test_run_state.py tests/test_run_impl_resume_paths.py tests/test_responses_tracing.py` -> `642 passed in 3.24s`
- `bash .agents/skills/code-change-verification/scripts/run.sh` -> `code-change-verification: all commands passed.` (`make format`, `make lint`, `make typecheck`; `9637 passed, 29 skipped in 47.92s` parallel, `77 passed, 4 skipped, 55 deselected in 1.70s` serial)

### Issue number

N/A — no issue was filed for this. Related: #2540, fixed by #2547.

### Checks


karaaslanz (commented):
There looks to be one remaining gap in the non-HITL path described in the PR body.

The new assignment is still inside `if run_state is not None`, so it only copies trace state when this completion itself came from a resumed `RunState`. For a fresh `Runner.run(agent, input=...)`, `run_state` is `None`, this line is skipped, and I don't see another completion/finalization assignment that puts the current trace onto the resulting `RunResult`. `result_to_run_state()` then falls back from a missing `_trace_state` to `getattr(result, "trace", None)`, but `RunResult` does not expose that trace surface.

The regression test currently starts with an approval interruption, so `approved` is already a resumed run and exercises exactly the branch this patch fixes. It does not cover the separate scenario called out in the description: `Runner.run(agent, input=...) -> result.to_state() -> Runner.run(agent, state)`.

Could we add a direct fresh-completion regression for that path and either preserve the active trace state for fresh results as well, or narrow the stated behavior if continuing a fully completed fresh result is intentionally unsupported? That would keep the implementation/test aligned with the broader claim in the PR.

BlueX888:
Checked this against the code and a repro rather than taking it at face value — the fresh path turns out to be covered already, so no behavior change is needed here.

By the time execution reaches that assignment, `run_state` is not the caller's resume argument: for a fresh run it is rebound to a newly created `RunState` at `src/agents/run.py:787`, so `run_state is not None` holds and the line does run. The guard is not a "resumed only" gate at that point; it matches the two pre-existing copy sites (lines 1426 and 1607), which are written the same way.

Verified empirically with a fresh run that completes and is then resumed from `to_state()`, on `main` (fbf59a40) and on this branch:

| | `first._trace_state` | traces after resuming |
|---|---|---|
| `main` | `None` | 2 (`assert 2 == 1`, 2x `trace_start`/`trace_end`) |
| this branch | set | 1 (every span shares one `trace_id`) |

`run_state` is only `None` at that point for a run that never entered the normal turn loop, so no completion result is produced there.

I did add the regression test you asked for, since nothing pinned the non-HITL path: `tests/test_agent_tracing.py::test_fresh_completed_run_state_reuses_original_trace` (f77cd2f7). It fails on `main` with `assert 2 == 1` and passes here. `tests/test_agent_tracing.py` 35 passed, the wider run/trace subset 182 passed, `ruff check` / `ruff format --check` clean.

seratch:
Thanks for tracing the missing assignment. The original #2540 workflow concerns resuming an interrupted approval, which is already covered; resubmitting a fully completed result through to_state() is a separate case. The new tests establish that the trace count changes, but do not show a completed-run workflow that needs this identity reuse.

I would close this for now. Applications can correlate completed runs with group_id or place multiple Runner calls inside an explicitly owned trace. A concrete workflow that cannot use either supported approach would justify revisiting the completed-state contract, rather than extending it for symmetry with the interrupted paths.

PR #5046, opened 2026-09-15
fix(run-state): preserve agent-tool owner provenance across restore

namtran1812:
## Summary

Persist `Agent.as_tool()` owner provenance for interrupted nested runs so owner replacement can still be detected after `RunState` serialization and restoration.

Previously, replacing an interrupted agent tool before `RunState.from_json()` could cause the restored action to resolve directly to the replacement wrapper. This loses the original Python object identity before interrupted-turn reconciliation, allowing the existing replacement check to be bypassed when the replacement uses the same agent and tool names.

This change:

- persists the existing stable agent identity signature with interrupted `Agent.as_tool()` state
- carries that provenance through restoration
- validates it during interrupted-turn reconciliation before relying on tool object identity
- preserves resume behavior when an equivalent agent graph is reconstructed with new Python objects
- keeps states without owner provenance readable and rejects malformed provenance

The new field is included in the current unreleased RunState schema version (`1.17`).

## Testing

- Added regression coverage for owner replacement before restoration and failure before side effects
- Added coverage for equivalent reconstructed owners
- Added compatibility coverage for missing and malformed owner provenance
- `tests/test_tool_name_collision_policy.py`: 40 passed
- `tests/test_run_state.py`: 446 passed
- `tests/test_run_state_compatibility_corpus.py`: 117 passed
- Ruff, mypy, and Pyright pass


seratch:
This PR still needs many changes, and resolving the TODO comment is not an urgent need for us. Thus, let us close this PR for now.

PR #5049, opened 2026-09-16
fix(tests): restore any_llm_model in sys.modules after stubbed imports

Jinzhengxu:
### Summary

`_import_any_llm_module` in `tests/models/test_any_llm_model.py` removes
`agents.extensions.models.any_llm_model` from `sys.modules` with a bare
`sys.modules.pop(...)` before re-importing it against a stubbed `any_llm`.
Nothing puts the original module object back, so every test that uses the
helper leaves a *different* `any_llm_model` module in `sys.modules` than the
one `agents.extensions.models.any_llm_provider` already imported
`AnyLLMModel` from.

`test_any_llm_provider_passes_api_override` then imports `AnyLLMModel` from
the replacement module while `AnyLLMProvider.get_model()` still instantiates
the original class, and `isinstance(model, AnyLLMModel)` fails:

```
E   AssertionError: assert False
E    +  where False = isinstance(<agents.extensions.models.any_llm_model.AnyLLMModel object at ...>, <class 'agents.extensions.models.any_llm_model.AnyLLMModel'>)
```

Whether this happens depends on which tests an xdist worker ran earlier, so it
shows up as an intermittent failure under `make tests`. Deterministic
reproduction on `main` (fbf59a4), with `any-llm-sdk` installed:

```
uv run pytest -p no:randomly \
  "tests/models/test_any_llm_model.py::test_any_llm_provider_reads_default_model_at_call_time" \
  "tests/models/test_any_llm_model.py::test_any_llm_chat_path_is_used_when_responses_are_unsupported" \
  "tests/models/test_any_llm_model.py::test_any_llm_provider_passes_api_override"
```

fails on the third test; the same three tests pass in the reverse order.

### Change

Remove the module through `monkeypatch.delitem(..., raising=False)` instead of
`sys.modules.pop(...)`. `MonkeyPatch` records the original module object and
restores it at teardown, so tests that run after the helper see the same
`any_llm_model` module the provider module is bound to. The stubbed
`any_llm` was already installed through `monkeypatch.setitem`; this makes the
two halves of the helper symmetric.

### Verification

- The ordered reproducer above passes after the change; the whole file passes
  in default order, under `pytest-randomly` seeds 1, 2 and 3, and under
  `-n 4`.
- `ruff format --check`, `ruff check` and `pyright` on the file report no
  issues.


karaaslanz (commented):
One test-isolation edge case remains in the module restore path.

tests/models/test_any_llm_model.py
```diff
@@ -108,7 +108,7 @@ def create(provider_name: str, api_key: str | None = None, api_base: str | None
     fake_any_llm: Any = pytypes.ModuleType("any_llm")
     fake_any_llm.AnyLLM = FakeAnyLLMFactory
 
-    sys.modules.pop("agents.extensions.models.any_llm_model", None)
+    monkeypatch.delitem(sys.modules, "agents.extensions.models.any_llm_model", raising=False)
```
karaaslanz, line 117: `MonkeyPatch.delitem` restores the `sys.modules` entry, but the re-import also overwrites the parent package attribute `agents.extensions.models.any_llm_model`. Pytest does not track that assignment, so teardown can leave `sys.modules["agents.extensions.models.any_llm_model"]` pointing at the original module while `agents.extensions.models.any_llm_model` still points at the stub-imported replacement. That preserves a second module-identity split for code/tests importing through the parent package. Could this helper also restore the parent package attribute (or otherwise make both bindings participate in teardown)? A small regression assertion that both bindings are identical after teardown would pin the isolation guarantee.

Jinzhengxu: Good catch, confirmed: `importlib.import_module` rebinds `agents.extensions.models.any_llm_model` on the parent package, and `MonkeyPatch.delitem` only covers the `sys.modules` entry, so after teardown `from agents.extensions.models import any_llm_model` still resolved to the stub-backed module. Pushed a follow-up (3933700): the helper now routes the parent attribute through `monkeypatch.delattr(..., raising=False)` as well, and `test_import_any_llm_module_restores_module_bindings_on_teardown` asserts both bindings are the original object after the context exits (it fails against the previous revision). The file passes in default order, three `pytest-randomly` seeds and `-n 4`; ruff and pyright are clean.

jbeckwith-oai (approved):
Reviewed `393370058dccd323e468040eb3c2e55664b13092` in its pinned review worktree.

The reported test-order identity split is real. Restoring both the existing sys.modules entry and the parent-package attribute keeps later imports consistent with the class retained by the provider. The added regression checks both bindings after teardown.

PR #5050, opened 2026-09-16
fix(litellm): apply the Chat Completions feature validation to LitellmModel

L4XB:
## What does this PR do?

The SDK has two Chat Completions adapters. Only one of them is treated as one.

`OpenAIChatCompletionsModel` refuses, or warns about, the three things a Chat
Completions call cannot carry — server-managed conversation state
(`previous_response_id` / `conversation_id`), a reusable `prompt`, and the
Responses-only reasoning settings (`reasoning.mode`, `reasoning.context`) —
gated on `strict_feature_validation` (#3298, #3309, #3312, #3314).

`LitellmModel` speaks the same API and has none of it. It accepted all three and
dropped them in silence, marked only with a comment:

```python
previous_response_id: str | None = None,  # unused
conversation_id: str | None = None,       # unused
```

It also never passed `strict_feature_validation` into the `Converter` and
`ChatCmplStreamHandler` it shares with the other adapter, so a LiteLLM user could
not turn on the feature at all.

## The workflow that breaks

Not a dropped setting nobody notices. Once `previous_response_id` or
`conversation_id` is set, `Runner` installs `OpenAIServerConversationTracker`
(`run.py:700-712`) purely on the presence of the id, whatever model is in use.
The tracker then sends only the items the server is not assumed to hold. So on
the second model call of a tool-using run the request is:

One agent, one function tool, `Runner.run(..., previous_response_id="resp_abc")`,
recording what reaches `litellm.acompletion`:

```
--- request 1 (2 messages) ---
    {"content": "be brief", "role": "system"}
    {"role": "user", "content": "What is the weather in Paris?"}
--- request 2 (2 messages) ---
    {"content": "be brief", "role": "system"}
    {"role": "tool", "tool_call_id": "call_1", "content": "sunny in Paris"}
```

The second request has the tool result and neither the user turn nor the
assistant tool call that produced it. Every Chat Completions provider rejects a
`tool` message that does not follow an `assistant` message carrying the matching
`tool_calls`.

The same script against `OpenAIChatCompletionsModel` produces the same two
requests — the trimming is the runner's server-managed behaviour and this PR does
not change it — but it says so first:

```
WARNING openai.agents: OpenAIChatCompletionsModel does not support server-managed
conversation state (previous_response_id). Chat Completions requires callers to
pass the full conversation history; use a Responses API model for
previous_response_id or a conversation-capable model for conversation_id.
Ignoring unsupported server-managed conversation state; enable strict feature
validation to raise an error instead.
```

With `strict_feature_validation=True` it raises a `UserError` before the first
call. On LiteLLM there was nothing at all.

I know #4837 was closed because it showed settings that are not forwarded rather
than an application that fails. This is the other case: the guard already exists
and is already released, one of the two adapters that needs it does not have it,
and the request that goes out without it is malformed.

## Second half of the same defect

`validate_output_guardrails_with_server_managed_conversation`
(`agent_runner_helpers.py:279-296`) rejects output guardrails combined with
either id, "because rejected output cannot be removed from server-managed
conversation history" — and exempts `OpenAIChatCompletionsModel` by class, since
"Chat Completions owns its released warn-and-ignore or strict rejection
behavior".

LiteLLM has no server-managed conversation history either, so the rejection names
a store that does not exist. Measured on `main`:

| model | `Runner.run(..., previous_response_id=…)` with an output guardrail |
|---|---|
| `LitellmModel` | `UserError: Output guardrails cannot be combined with conversation_id, previous_response_id, …` |
| `OpenAIChatCompletionsModel` | allowed; warns, then calls the provider |

## Change

- `ChatCmplUnsupportedFeatures` in `models/chatcmpl_helpers.py` holds the three
  checks and the warn-once / raise decision. Both adapters call it, so they
  cannot drift apart again. Messages are unchanged for `OpenAIChatCompletionsModel`:
  the class name comes from `type(self).__name__`, and its three
  `_handle_unsupported_*` methods stay as one-line delegates.
- `LitellmModel` gains `strict_feature_validation: bool = False` (appended last,
  per the parameter-order contract in `AGENTS.md`), runs the three checks in both
  `get_response` and `stream_response`, and threads the flag into
  `Converter.items_to_messages`, `Converter.message_to_output_items` and
  `ChatCmplStreamHandler.handle_stream` the way `OpenAIChatCompletionsModel` does.
- The guardrail validation exempts any adapter that owns this validation instead
  of one named class. Duck-typed via `owns_chatcmpl_feature_validation` so
  `run_internal` does not import the optional `litellm` extra.

Deliberately not in scope: `LitellmProvider` gets no flag (the documented entry
point is constructing `LitellmModel` directly), and no `docs/` change — the
behaviour is not in the latest release yet, which per `AGENTS.md`
"Documentation Release Timing" belongs in a separate docs-only PR. Happy to open
that one whenever it suits the release.

## Tests

`tests/models/test_litellm_feature_validation.py`, 13 tests: warn-and-ignore and
strict rejection for each of the three features, warn-once per model instance,
the streaming path, the two-call `Runner` workflow above, and the guardrail
validation for all three model kinds.

```
$ pytest tests/models/test_litellm_feature_validation.py -q
13 passed in 1.18s

# same tests, src reverted to main
12 failed, 1 passed
```

The one that passes on `main` is
`test_output_guardrails_are_still_rejected_on_a_responses_model` — it pins the
behaviour this PR must not change.

Full suite, same interpreter, `main` vs this branch:

```
main:        48 failed, 9511 passed, 48 skipped, 9 errors
this branch: 48 failed, 9524 passed, 48 skipped, 9 errors
```

The 57 failures/errors are identical line for line (`comm` on the sorted
`FAILED`/`ERROR` lists is empty in both directions); they are optional
dependencies missing in my local env, not this diff. +13 is exactly the new file.

Gates: `ruff format --check` clean, `ruff check` clean,
`check_optional_truthiness.py` clean, `mypy src` 54 errors before and after with
none in the changed files, `pyright --project pyrightconfig.json` 29 errors with
none in the changed files.

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/openai/openai-agents-python/blob/main/AGENTS.md)?
- [ ] Was this discussed/approved via a GitHub issue? No — happy to open one if
      you would rather discuss it there first.
- [x] Did you make sure to update the documentation with your changes? See the
      release-timing note above.
- [x] Did you write any new necessary tests?


seratch:
Thanks for tracing the request construction and adding the Runner-level example. The history trimming is real, but the example passes a Responses-only conversation ID to LiteLLM. The supported LiteLLM workflow omits those IDs: Runner retains tool-call history within a run, and sessions or result.to_input_list() provide history between runs.

I am closing this PR. The default behavior still sends the same incomplete request, and the example does not establish a need for the broader validation API or guardrail exemption. A concrete application case showing why the supported history workflow is insufficient would justify revisiting a narrower change.

PR #5051, opened 2026-09-16
fix(tracing): retry rate-limited trace exports (Fixes #5023)

### LINKED ISSUE #5023: BackendSpanExporter drops the whole batch on HTTP 429 instead of retrying with Retry-After
### Please read this first

- **Have you read the docs?** Yes: [Tracing](https://openai.github.io/openai-agents-python/tracing/) and the `BackendSpanExporter` reference.
- **Have you searched for related issues?** Yes. #4683 / #4711 cover retries being disabled after shutdown, #4983 covers a non-serializable value dropping a batch, #5008 covers the endpoint. None covers how the exporter classifies HTTP status codes.

### Describe the bug

`BackendSpanExporter._export_with_deadline()` treats every `4xx` response as final:

```python
# If the response is a client error (4xx), we won't retry
if 400 <= response.status_code < 500:
    logger.error("[non-fatal] Tracing client error %s: %s", ...)
    break
```

That includes `429 Too Many Requests`. A single rate-limited response from `/v1/traces/ingest` discards the whole batch, up to `max_batch_size` (128) traces and spans, with no retry and without looking at `Retry-After`. The run itself succeeds, so the only symptom is a gap in the dashboard and one `[non-fatal]` log line. The same applies to `408` and `409`, which the OpenAI Python client (`_base_client._should_retry`) retries together with `429` and `5xx`; the exporter only mirrors the `5xx` half of that policy.

The export path already has everything needed to retry: `max_retries`, `base_delay`, `max_delay`, the backoff loop, and `agents.models._retry_runtime.parse_retry_after_value` / `parse_retry_after_ms` for the header.

### Debug information

- Agents SDK version: `main` at `fbf59a40` (also 0.22.0)
- Python version: 3.10.12
- Operating system: Linux
- Model and model provider: none needed; `httpx2.Client` is patched
- Does the issue occur consistently or intermittently? Consistently for any 429

```text
ERROR:openai.agents:[non-fatal] Tracing client error 429. Response data is redacted.
post calls: 1
```

### Repro steps

```python
import logging

from agents.tracing.processors import BackendSpanExporter
from agents.tracing.span_data import CustomSpanData

logging.basicConfig(level=logging.WARNING)

with patch("httpx2.Client") as client:
    limited = MagicMock(status_code=429, headers={"retry-after": "1"})
    limited.text = '{"error": {"message": "Rate limit reached"}}'
    ok = MagicMock(status_code=200)
    client.return_value.post.side_effect = [limited, ok]

    exporter = BackendSpanExporter(api_key="sk-test", max_retries=3, base_delay=0.1)
    span = SpanImpl(
        trace_id="trace_1",
        span_id="span_1",
        parent_id=None,
        processor=MagicMock(),
        span_data=CustomSpanData(name="probe", data={}),
        tracing_api_key=None,
    )
    exporter.export([span])
    print("post calls:", client.return_value.post.call_count)  # 1: the batch is gone
```

### Expected behavior

`408`, `409` and `429` are retried with the existing backoff up to `max_retries`, waiting at least the advertised `Retry-After` / `retry-after-ms` (bounded by `max_delay`), the way the OpenAI client and the SDK's own model retry helpers already treat these statuses. Other `4xx` responses stay non-retried. In the script above the second call succeeds and `post calls` is 2.

A fix with regression tests is ready and I will attach it right away.

jaideeppyne:
## Summary
- `BackendSpanExporter` treated every `4xx` as final, so an HTTP `429` from `/v1/traces/ingest` discarded the whole batch (up to `max_batch_size`) with no retry and no `Retry-After` handling.
- Retry `408`, `409`, and `429` through the existing backoff loop (up to `max_retries`), honouring `Retry-After` / `retry-after-ms` via the shared `_retry_runtime` parsers (lazy import), bounded by `max_delay`.
- Honour an explicit `x-should-retry: false` veto so classification matches `get_openai_retry_advice` / the OpenAI client's `_should_retry`. Other `4xx` responses stay non-retried.

Fixes #5023.

## Notes for reviewers
The fork is behind `openai/main` (cannot sync: OAuth token lacks `workflow` scope after upstream added CodeQL workflows). The branch updates only `processors.py` and `test_trace_processor.py` to match current `main` plus this fix. The substantive change is the retry classification + tests around `_export_with_deadline`; please focus the review there.

## Test plan
- [x] Regression tests for `408`/`409`/`429` retries, recovery on second attempt, `Retry-After` (seconds / ms / over `max_delay` / unparsable / negative), and `x-should-retry: false`
- [ ] `pytest tests/test_trace_processor.py tests/tracing`
- [ ] Existing non-retry `4xx` and `5xx` retry tests still pass


jaideeppyne:
Closing: branch was based on a stale fork tip and conflicts with main (could not sync fork — OAuth token lacks `workflow` scope). Will reopen with a minimal mergeable patch.

PR #5059, opened 2026-09-16
fix(core): keep str(RunResult) working after release_agents plus GC

MohammadHijjawi97:
### Summary

`release_agents()` is meant to drop strong agent refs so the agent graph can be collected. `repr(result)` and `dataclasses.asdict(result)` already survive that path. `str(result)` still went through `pretty_print_result`, which reads `last_agent.name` and raises `AgentsException` once the weakref is gone.

Pretty-print now treats a released agent as `<released>` instead of crashing, so logging or printing a result after cleanup still works. Streaming results use the same helper.

### Test plan

- [x] Added `test_run_result_str_after_release_agents_and_gc`
- [x] Existing release_agents repr/asdict test still passes
- [x] `python -m pytest tests/test_result_cast.py::test_run_result_str_after_release_agents_and_gc tests/test_result_cast.py::test_run_result_repr_and_asdict_after_release_agents`

### Checks

- [x] I've confirmed the focused tests above pass


seratch:
Thanks for the contribution. I confirmed the exception after release_agents() and garbage collection, but on reconsideration I do not think the report establishes a need to extend the post-release display contract. Callers can capture the formatted result before releasing the agent graph, log final_output, or use repr(result) for diagnostics.

I am going to close this PR for now, and withdraw the earlier request to narrow the helper as a condition for merging. Matching repr/asdict behavior alone does not demonstrate a workflow those existing options cannot support. We can reconsider if a concrete integration requires formatting the complete result after the graph has been collected.

MohammadHijjawi97:
Thanks for looking at this and for confirming the exception. Agreed that this does not need to extend the post-release display contract. Callers can format the result before release_agents(), log final_output, or use repr(result). Happy to leave it closed.

PR #5060, opened 2026-09-17
fix(extensions): close owned event streams when runs fail

WenJing95:
### Summary

Ensure `Thread.run()` closes its event iterator on `turn.failed` or `error` events so the Codex subprocess and temporary output schema are cleaned up before raising. Successful aggregation remains unchanged.

### Test plan

* Full verification suite passes (formatting, lint, type checks, and tests).
* Verified focused Codex tests pass.
* Offline checks confirm cleanup triggers on error and doesn't affect normal completion.

### Issue number

N/A

### Checks

* [x] I've added new tests, if relevant
* [x] I've run `.agents/skills/code-change-verification/scripts/run.sh`
* [x] I've confirmed all verification steps pass
* [x] If using Codex, I've run `/review` before submitting this PR


jbeckwith-oai (approved):
Reviewed `5e093917aa2ae06f528e25ef617c5adb202d446f` in its pinned review worktree.

The early turn-failure and stream-error exits leave the owned iterator open. Using contextlib.aclosing at Thread.run reaches the existing per-run subprocess and schema-file cleanup deterministically. The regression retains the iterator explicitly, so garbage collection cannot substitute for correct ownership.

PR #5062, opened 2026-09-17
fix(tracing): preserve batches with unserializable values

### LINKED ISSUE #4983: BackendSpanExporter: one metadata or custom span value that isn't JSON serializable drops the whole export batch, including unrelated traces
### Describe the bug

`BackendSpanExporter` posts a whole batch of traces and spans in one request. If any one item in that batch has a value `json` can't encode, the request body can't be built, `export()` raises, and `BatchTraceProcessor` drops the entire batch. So one bad value in one run's `trace_metadata` also throws away every other trace and span that happened to be queued with it, including runs that had nothing wrong with them.

Things that trigger it: a `uuid.UUID` or `datetime` in `RunConfig(trace_metadata=...)`, a `datetime` in `custom_span(data=...)`, or a NaN or inf float (httpx2 encodes with `allow_nan=False`). It happens on the default OpenAI ingest endpoint and on custom endpoints.

### Debug information
- Agents SDK version: 0.22.2 (also main at fbd2dbca)
- Python version: 3.14.3
- Operating system: macOS
- Model and model provider: none needed, the repro uses `ScriptedModel`

```text
ERROR [non-fatal] Tracing exporter failed; dropping batch of 8 items
items received: 0 []
```

Calling the exporter directly shows the cause:

```text
exporter.export([trace("x", metadata={"request_id": uuid.uuid4()})])
TypeError: Object of type UUID is not JSON serializable
```

### Repro steps

The repro doesn't touch the network. `httpx2.MockTransport` stands in for the ingest endpoint.

```python
import asyncio, json, uuid, logging
import httpx2
from agents import Agent, Runner, RunConfig
from agents.testing.model import ScriptedModel, assistant_message
from agents.tracing import set_trace_processors

logging.basicConfig(level=logging.ERROR, format="%(levelname)s %(message)s")
received = []

def handler(request):
    received.extend(json.loads(request.content)["data"])
    return httpx2.Response(200)

exporter = BackendSpanExporter(api_key="sk-fake")  # default OpenAI ingest endpoint
exporter._client = httpx2.Client(transport=httpx2.MockTransport(handler))
processor = BatchTraceProcessor(exporter, schedule_delay=3600)
set_trace_processors([processor])

async def main():
    agent = Agent(name="assistant", model=ScriptedModel([[assistant_message("hi")], [assistant_message("hi")]]))
    await Runner.run(agent, "hello", run_config=RunConfig(workflow_name="run_A"))  # nothing wrong with this run
    await Runner.run(agent, "hello", run_config=RunConfig(
        workflow_name="run_B", trace_metadata={"request_id": uuid.uuid4()}))

asyncio.run(main())
processor.force_flush()
print("items received:", len(received), [d.get("workflow_name") for d in received if d.get("object") == "trace"])
```

### Expected behavior

run_A's trace and spans get exported no matter what's in run_B's metadata. At most, the value that can't be encoded gets dropped from run_B, and run_B's trace and spans still arrive.

AnvitDevadiga:
Fixes #4983

BackendSpanExporter now sanitizes each exported trace or span independently before JSON encoding, so an unserializable metadata value such as UUID or datetime cannot discard unrelated items from the same batch. The invalid value is omitted while the rest of the trace remains exportable. Added regression coverage for mixed valid and invalid trace metadata; 17 tracing exporter tests and Ruff pass.


seratch:
Thanks for addressing the batch-loss issue in #4983. I compared this with #4984 and checked the actual request body. Applying the existing sanitizer unconditionally removes valid integer dictionary keys: metadata such as status_counts={200: 3} reaches a custom exporter as an empty mapping on this head, while main preserves it.

I would close this duplicate and continue with #4984, which preserves JSON-compatible keys and limits repair to items that cannot be encoded. That candidate still needs the previously requested removal of the unsupported extreme-integer handling. This keeps one implementation for the demonstrated batch-loss problem without discarding valid sibling data.

PR #5057, opened 2026-09-16
docs: fix 30 dead relative links in the ja/ko/zh translated docs

simpleqt:
## Summary

The ja/ko/zh pages were copied verbatim from the English docs but live one directory deeper under the mkdocs i18n `folder` layout (`mkdocs.yml:44-45`), so 30 relative targets resolved under `docs/<lang>/` — which has no `ref/` or `assets/` trees — and 404 on the built site:

- `{ja,ko,zh}/agents.md:303` → `ref/lifecycle.md`
- `{ja,ko,zh}/mcp.md` → `ref/mcp/manager.md`
- `{ja,ko,zh}/testing.md` → `ref/testing.md`, `ref/realtime/testing.md`, `ref/voice/testing.md`
- `{ja,ko,zh}/sessions.md` → `ref/memory/openai_conversations_session.md`
- `{ja,ko,zh}/tools.md` → 3 codex extension ref pages
- `{ja,ko,zh}/sandbox/guide.md:17` → `../assets/images/harness_with_compute.png`

All 30 targets prefixed so they resolve to the shared `docs/ref/` and `docs/assets/` trees (every English target verified with `test -e`).

## Test plan

Repo-wide link walk before/after: 30 MISSING → 0; English targets re-verified by file existence.

PR #5063, opened 2026-09-17
docs: add English examples guide and nav entry

transparent777:
## Summary
- Expand `docs/examples.md` into a topic map of the `examples/` tree and document the supported `make examples-run` workflow for clone-based development.
- Add the page to the English MkDocs nav and link it from the intro "Choose your path" table.

## Why
The examples directory is large; the published docs only had a single GitHub link. This gives AI-application developers a stable on-site entry point aligned with Quickstart, Tools, MCP, Sessions, and Sandbox guides.

## Verification
- `uv run docs/scripts/generate_ref_files.py`
- `uv run mkdocs build`

## Scope note
Does not modify ja/ko/zh translated docs (separate from #5057).


seratch:
Thanks for the suggestion, but we've intentionally simplified this for lower maintenance cost. Thus, let us close this PR now.

PR #5065, opened 2026-09-17
docs: update translated pages

seratch:
This pull request applies the translation changes based on the latest main branch.


seratch:
@codex review again

markstuart-oai (approved):
Reviewed commit `3bef0e3bace2b1ad84fbdb82949c4c0f6d6c3ecd` against the current English pages.

The Japanese, Korean, and Chinese updates preserve the distinction between workspace separation and command confinement, including Linux's lack of added OS isolation, macOS's filesystem-only restrictions, trusted path grants, and the shared-workspace behavior of a custom `Manifest.root`. The session lifecycle, ownership, and credential-restoration guidance remains consistent with the source.

Validation included an independent review of each language, unchanged contents of all 84 fenced code blocks, matching inline API identifiers and reference targets, valid anchors for the six added links, and `git diff --check`. CI ran `make build-docs` successfully on a merge tree identical to this head; all reported checks are green. Runtime tests were not rerun locally for this translation-only change.

PR #5076, opened 2026-09-17
fix(tests): ignore GC-time ResourceWarnings in redaction warning assertions

Jinzhengxu:
### Summary

Three tests in `tests/test_error_logging_redaction.py` record warnings with
`warnings.catch_warnings(record=True)` + `warnings.simplefilter("always")` and
then assert `not caught_warnings` in the redacted branch:

- `test_invalid_final_output_handler_invalid_fallback_preserves_redaction`
- `test_invalid_final_output_handler_fallback_serialization_follows_redaction_policy`
- `test_empty_final_output_handler_fallback_serialization_follows_redaction_policy`

`simplefilter("always")` also un-ignores `ResourceWarning`, which the interpreter
emits while garbage-collecting objects that earlier tests in the same xdist worker
left unclosed. When a collection happens to run inside that capture window, the
assertion fails although nothing about redaction changed. Under `make tests`
(`-n auto --dist worksteal`) the window moves with scheduling, so the failure is
intermittent and lands on a different parametrization each time.

### Evidence

On `main` with `pytest-randomly`, seeds 11–14 over six full-suite runs failed
three times, always in one of those three tests and always with exactly one
recorded warning:

```
assert not caught_warnings
E   assert not [<warnings.WarningMessage object at 0x...>]
```

The failing worker's full 246-test prefix plus the victim passes serially, so it
is not order pollution but GC timing. Running the suite with
`-W always::ResourceWarning` shows the sources that fire during collection:
an unclosed event loop (`tests/mcp/test_caching.py`), unclosed
`BufferedReader`/`BufferedRandom` handles (`tests/sandbox/test_docker.py`,
`test_session_sinks.py`, `test_runtime.py`, `test_tar_utils.py`) and the
`scandir` iterator in `_list_local_dir_files_from_dir_fd` (`test_entries.py`,
see #4436).

### Change

Add `_warnings_under_test()` in the test module and pass the recorded warnings
through it before the three assertions. It drops `ResourceWarning` records only;
every other category, including the pydantic serializer `UserWarning` the
non-redacted branch asserts on, still reaches the assertion unchanged. No
behaviour under test is touched and no global warning filter is added, in line
with the guidance in `tests/README.md` to keep warning handling local to the
test.

### Verification

- `tests/test_error_logging_redaction.py`: 168 passed in default order, under
  `pytest-randomly` seeds 1 and 2, and with `-n 4`.
- Full suite with `pytest-randomly` seeds 12 and 13, twice each: 4 of 4 runs green (9622 passed, 47 skipped each); on `main` the same seeds failed 3 of 6 runs in these tests.
- `ruff format --check`, `ruff check` and `pyright` on the file report no issues.


jbeckwith-oai (approved):
Reviewed `deda3ef43f80d396dfcc372937637574601d50b1` in its pinned review worktree.

The broad warning capture can include unrelated GC ResourceWarnings. Filtering that category only at these three assertions retains serializer warnings and the sensitive-data assertions, without changing runtime redaction or global warning policy.

seratch:
@Jinzhengxu Can you change the draft mode to be ready for review? We're unable to merge draft ones, plus we cannot change the mode for this PR on our side.

Jinzhengxu:
I tried, but the "Ready for review" button is disabled for me: "Author has reached the open pull request limit" — #5049 (also approved, workflows still awaiting approval) counts against this repo's limit for users without write access. I've resolved the open thread on #5049; as soon as it's merged (or if you add me to the bypass list), I'll mark this one ready immediately.

Jinzhengxu:
@seratch #5049 has merged, so this account is no longer at the open pull request limit and could mark this one ready for review right away. Since it was closed from your side I can't reopen it myself (authors can only reopen PRs they closed themselves). If you're still open to this change, could you reopen it? I'll switch it to Ready immediately. The branch (`deda3ef`) still merges cleanly onto `main`, and the three `simplefilter("always")` assertion sites in `tests/test_error_logging_redaction.py` are unchanged there. If you closed it on purpose and would rather not take it, that's fine too, just let me know.

PR #5079, opened 2026-09-17
docs(examples): drop the duplicated word in the Temporal trace name

simpleqt:
## Summary

The Temporal sandbox example's trace name was "Temporal Sandbox Sandbox Agent" — duplicated "Sandbox" (the README titles the example "Temporal Sandbox Agent"). The name is surfaced in the Temporal UI and trace backends.

## Test plan

String-only change in an example; no tests reference the literal.


jbeckwith-oai (approved):
Reviewed `20473e1e21135b1a348d1175b59d4fa08f9249fd` in its pinned review worktree.

The duplicated word is present in the original Temporal trace label. The one-string correction matches the example name and leaves workflow behavior unchanged.

PR #5080, opened 2026-09-17
fix(sessions): EncryptedSession.pop_item drains recoverable ciphertext when configure

### LINKED ISSUE #5005: EncryptedSession.pop_item drains recoverable ciphertext when configured with the wrong key
### Please read this first

- **Have you read the docs?** Yes, the encrypted-session guide and session correction examples.
- **Have you searched for related issues?** Yes, searched EncryptedSession and pop_item. This report excludes TTL cleanup, which is covered intentionally by existing tests.

### Describe the bug

With an incorrect encryption key, a single `EncryptedSession.pop_item()` deletes every ciphertext item in the underlying session and returns `None`. Restoring the correct key afterward cannot recover the conversation. Before the pop, all messages remain readable with the correct key.

A concrete application scenario is an undo/regenerate action after a deployment has loaded the wrong encryption secret. The configuration mistake initially hides history without deleting it (`get_items()` returns an empty list); one correction action then removes the entire recoverable conversation.

This is **not** a claim that replacing the key transparently rotates existing ciphertext, or that expired items must remain indefinitely. The reproduction uses freshly written, unexpired items and intentionally demonstrates a configuration error. I am asking whether a destructive operation should distinguish authentication failure from ordinary TTL expiry rather than drain both identically.

### Debug information

- SDK checkout: `fbf59a40` (0.22.2).
- Python: 3.13.14; macOS arm64.
- Backend: real SQLiteSession in a temporary on-disk database.
- No model, API key, network, concurrency, timing manipulation, or fabricated backend.
- Frequency: deterministic.
- Latest packaged release was not separately installed for this reproduction.

### Repro steps

```python
import asyncio
import tempfile
from pathlib import Path
from agents.extensions.memory import EncryptedSession

async def main():
        store = SQLiteSession('conversation', str(Path(tmp) / 'history.db'))
        try:
            correct = EncryptedSession(session_id='conversation', underlying_session=store, encryption_key='example-correct-key', ttl=3600)
            wrong = EncryptedSession(session_id='conversation', underlying_session=store, encryption_key='example-wrong-key', ttl=3600)
            messages = [{'role':'user','content':'hello'}, {'role':'assistant','content':'hi'}, {'role':'user','content':'follow-up'}]
            await correct.add_items(messages)
            assert await correct.get_items() == messages
            print('recoverable with correct key before:', len(await correct.get_items()))
            print('wrong-key get_items:', await wrong.get_items())
            print('stored ciphertext after get_items:', len(await store.get_items()))
            print('wrong-key pop_item:', await wrong.pop_item())
            print('stored ciphertext after one pop:', len(await store.get_items()))
            print('recoverable with correct key after:', len(await correct.get_items()))
        finally:
            store.close()

asyncio.run(main())
```

Observed output:

```text
recoverable with correct key before: 3
wrong-key get_items: []
stored ciphertext after get_items: 3
wrong-key pop_item: None
stored ciphertext after one pop: 0
recoverable with correct key after: 0
```

### Expected behavior / scope clarification

I would expect an incorrect key not to turn a one-item correction into deletion of all recoverable history. An explicit authentication error or a non-destructive refusal would make the configuration problem recoverable. The intended retention policy for this case is not explicitly established in the current docs, so maintainer confirmation of that policy would be useful before choosing a patch.

`_unwrap()` catches Fernet `InvalidToken` and returns `None` for both expiry and authentication failure. `pop_item()` destructively pops first, then loops whenever `_unwrap()` returns `None` (`src/agents/extensions/memory/encrypt_session.py:207-216,280-295`).

A naive `get_items(1)` followed by `pop_item()` is not proposed as a fix: those are separate backend operations and can address different items if another caller mutates the session. A fix should preserve the existing expiry behavior and the underlying pop operation's concurrency guarantees.

nandanadileep:
Fixes #5005

## What changed
Changed `pop_item()` to pop exactly one item instead of looping. When decryption fails (wrong key or expired), the item is pushed back to the underlying session and `None` is returned, preventing destructive draining of recoverable ciphertext.

## Verification
`python -m pytest -q -x tests/extensions/memory/test_encrypt_session.py` passed


karaaslanz (commented):
This preserves wrong-key ciphertext, but it also changes the existing TTL cleanup behavior because `_unwrap()` intentionally returns `None` for both authentication failure and expiration. On an expired tail, this branch pops the token, gets `None`, appends the same encrypted envelope back, and returns. The next `pop_item()` sees the same expired tail again, so expired ciphertext is retained indefinitely instead of being drained/skipped as before. This is especially visible in the existing `test_encrypted_session_pop_expired` / mixed expired+valid behavior: the public return value may still be `None`, but the underlying expired record is no longer removed.

The issue itself explicitly calls out preserving the existing expiry behavior. I think the fix needs to distinguish “wrong key / invalid authentication” from “valid token, expired by TTL” before deciding whether to restore the popped envelope. One possible direction is authenticating the Fernet token without TTL before the destructive pop (or otherwise separating those two failure modes), then keeping the current loop/cleanup semantics for authentic-but-expired items.

Source-level review only; I did not run the branch locally.

seratch:
Thanks for the contribution. The data-loss problem in #5005 warrants a fix, but this implementation also restores authenticated expired tokens, changing the existing TTL cleanup behavior. Its separate pop and append operations can reorder history under concurrent writes or restore a record after the session has been cleared.

I’m going to close this PR and consolidate the work in #5083, which already separates authentication failure from expiry. That PR still needs an atomicity correction before it can merge.

PR #5084, opened 2026-09-18
Fix multiline skill frontmatter descriptions

### LINKED ISSUE #5026: Skills frontmatter parser mangles multi-line descriptions (folded >, literal |, wrapped lines) in the skill index

harshitethic:
## Summary

Fix skill metadata parsing for multi-line YAML frontmatter descriptions.

The lightweight frontmatter parser previously treated every line containing `:` as a new top-level field, so common skill metadata such as folded (`>`), literal (`|`), and wrapped plain descriptions was either reduced to the scalar marker or split into bogus keys.

The parser now:
- gathers indented continuation lines for a top-level field
- folds `>` descriptions into space-separated text while preserving paragraph breaks
- preserves line breaks for `|` descriptions
- joins wrapped plain scalar lines
- avoids treating colons inside indented description text as new metadata keys

## Tests

Added focused coverage through `LocalDirLazySkillSource.list_skill_metadata()` for:
- folded block descriptions
- literal block descriptions
- wrapped plain descriptions

Fixes #5026


seratch:
We'll go with the complete version of https://github.com/openai/openai-agents-python/pull/5028

PR #5087, opened 2026-09-18
fix(local-shell): surface executor errors to the model instead of aborting the run

### LINKED ISSUE #5086: Local shell executor failure aborts the whole run instead of reaching the model
Hey! Found that when the local shell executor itself fails (missing binary, sandbox denial, failing command with check), the raw error propagates out of Runner.run and kills the whole run — the model never sees the failure and can't retry. The sibling executors degrade to model-visible failed output instead. Happy to PR a fix that surfaces the failure as output like the others if that helps.

soroush5:
When the local shell executor itself raised, the raw error escaped Runner.run and killed the whole run, so the model never saw the failure. It now returns the failure as output like the sibling executors do, and the run continues.

Verified: new tests fail before (2 failed) and pass after (10 green), runner suites green.

Fixes #5086


seratch:
Thanks for reporting this. The local shell tool is deprecated, and using shell tool is now recommended. Thus, we won't fix the pattern.

PR #5090, opened 2026-09-18
fix(sessions): close SQLiteSession connections owned by exited worker threads

Luccacvb:
### Summary

A long-lived file-backed `SQLiteSession` keeps one open sqlite connection for every worker thread that has ever served it, including threads that have since exited. An application that runs one event loop per turn accumulates connections and file descriptors until the process hits its descriptor limit and every further request fails.

`_get_connection` opens one connection per thread, stores it in `self._local`, and also registers it in `self._connections` so `close()` can release connections it does not own. Nothing ever removes an entry when the owning thread exits, and the thread-local slot dies with the thread, so that registry becomes the only reference to a connection that no caller can reach or use again.

The natural trigger is not a thread the caller created. `asyncio.run` shuts down its default executor when it returns, so the worker that `asyncio.to_thread` used inside the session exits after every turn:

```python
session = SQLiteSession("conversation_123", "conversation_history.db")

for turn in range(30):                      # a sync CLI, GUI, or Streamlit-style loop
    asyncio.run(Runner.run(agent, message, session=session))
```

| after N turns | `main` | with this PR |
|---|---|---|
| 1 | 1 tracked connection, +3 fds | 1 tracked connection, +3 fds |
| 10 | 10 tracked connections, +21 fds | 1 tracked connection, +4 fds |
| 30 | 30 tracked connections, +61 fds | 1 tracked connection, +4 fds |

On `main` all 30 retained handles still answer `SELECT 1`, and none of them is reachable any more: every owning worker has exited, including the newest.

The same accumulation happens in a threaded sync server calling `Runner.run_sync` per request, where it reaches a hard failure. Lowering the soft `RLIMIT_NOFILE` to 120 so the ceiling arrives in seconds, with one session shared across 60 requests on `main`:

```text
request= 45 tracked_connections= 45
RESULT: first failure at request 47: OSError: [Errno 24] Too many open files
        failures after it: 13
```

With this change the same run serves all 60 requests. The real ceiling scales with the configured descriptor limit; recovering from it on `main` means closing the session and building a new one against the same database, since `close()` is terminal and nothing else releases those handles.

This is a regression rather than a design choice. Before #2963 the only reference to a worker connection was the thread-local slot, so a connection became reclaimable once its thread exited; that PR added `self._connections` so `close()` could reach other threads' connections for Windows temp-directory cleanup, and the strong reference made dead threads' connections permanent. The two shapes, isolated from the SDK and measured after `gc.collect()`:

```text
thread-local reference only              30 exited threads -> fd_delta=+0
thread-local + tracking-set strong ref   30 exited threads -> fd_delta=+30  (all still usable)
```

The fix records which thread owns each tracked connection and closes the ones whose owner has exited, at the moment a new worker registers a connection. That is the only point where the set can grow. For the sequential workload above, each new registration closes the preceding worker's connection; more generally, each registration reaps connections whose owners have exited, while retaining connections owned by live threads.

Notes for review:

- The sweep runs under the existing `_connections_lock` inside `_get_connection`, which already holds `self._lock`; the lock order matches `close()` and `_invalidate_connection`. Closing another thread's connection is what `close()` already does, and connections are opened with `check_same_thread=False`.
- A `WeakSet` would express this directly but is not available: `sqlite3.Connection` does not support weak references (`TypeError: cannot create weak reference to 'sqlite3.Connection' object`).
- Reaping is keyed on `threading.Thread` identity rather than on a thread identifier, so a recycled identifier cannot be mistaken for a live owner.
- A connection whose `close()` fails during the sweep is quarantined, the same handling `close()` applies, so it still blocks the file-lock release and is not silently forgotten.
- The sweep closes a connection before dropping it from the registries, so an interrupt cannot leave an open connection with no remaining reference.
- Reaping runs before the replacement connection is opened, so the retired handles are released first. That ordering also matters for how completely they are released: closing a WAL connection while another connection to the same database is open leaves part of its descriptors held. Measured in plain `sqlite3` with no SDK involved, closing 30 WAL connections releases 61 descriptors when no other connection is open and 28 when one is.
- `AdvancedSQLiteSession` inherits the fix; the in-memory shared connection never registers per-thread entries and is unaffected; `AsyncSQLiteSession` uses a single aiosqlite connection. No public API, schema, or call shape changes.

Known limitations, unchanged by this change:

- A connection whose `_configure_connection` fails is still left open and untracked, exactly as on `main`. #4908 proposed closing it and was closed with the note that failed connections are not cached and later calls create fresh ones; nothing here alters that path, and the sweep does not see such a connection because it was never registered.
- A session whose workers all retire and that then never allocates again keeps its last retired connections until `close()`, because reaping is driven by allocation rather than by a background task.
- A connection owned by a thread that `threading` did not create is never reaped: `_DummyThread` always reports itself alive. That is the conservative direction, and it stays tracked for `close()`.
- A child process created with `os.fork()` sees every inherited thread as stopped, so its first allocation closes the inherited connections. Carrying a sqlite connection across `fork()` is already unsupported, and the parent's descriptors are unaffected.

### Test plan

Three tests in `tests/memory/test_session.py`:

- `test_sqlite_session_reaps_connections_from_exited_worker_threads` drives five turns through `asyncio.run(session.add_items(...))`, captures the first turn's connection, and asserts it is closed rather than merely untracked while the stored history stays complete. On `main` it fails at the second turn with `assert 2 == 1`.
- `test_sqlite_session_closes_retired_connections_before_the_next_allocation` covers the burst case: three workers register connections and hold at a barrier so they retire together, then a later turn must find every retired connection already closed at the moment it allocates its own. It records how many retired connections are still open when `sqlite3.connect` is called, which is `0` only when the sweep precedes the allocation. On `main`, and with the sweep placed after the allocation, the ordering assertion fails with `assert [3] == [0]`. Session history stays complete across the burst.
- `test_sqlite_session_keeps_connections_of_live_worker_threads` pins the other direction. A second running event loop has its own default executor, so its worker is a distinct thread that stays alive; the test captures that connection before this loop's worker registers one and triggers the sweep, then asserts it is still tracked and still usable. Sequential awaits on one loop would reuse a single worker and never reach the sweep, so the second loop is what makes the case deterministic. Removing the `is_alive()` guard fails it.

The tests run in about 10 ms each; 20 consecutive focused runs and 6 `tests/memory` runs under `-n 4` passed without a flake. `tests/memory` and `tests/extensions/memory` pass (679 passed, 4 skipped). `.agents/skills/code-change-verification/scripts/run.sh` passes.

### Issue number

N/A

### Checks


sylvesterkaczmarek (commented):
The sweep happens after the replacement sqlite connection has already been opened and configured. If a burst leaves enough dead-thread connections to hit the process fd limit, the next sqlite3.connect() can fail before the cleanup ever gets a chance to free them. Could dead owners be reaped before opening a new per-thread connection, with a regression where several workers exit before the next registration?

seratch (changes requested):
Thanks for investigating this and adding coverage for both exited and live workers. The retained connections are a real resource leak, and tracking their owning threads is a reasonable fix.

Please move the dead-owner sweep before opening and configuring the new per-thread connection. When several workers exit after registering their connections, their retained handles can prevent the next allocation from succeeding, so the current ordering never reaches cleanup. Please add a regression test covering that burst-then-exit sequence and verifying that abandoned connections are closed before the next allocation, with session history remaining usable.

Luccacvb:
Done, moved the sweep before `sqlite3.connect()` and `_configure_connection()`, and added the burst regression test.

The test keeps three workers alive until all have registered connections, then waits for them to exit. At the next `sqlite3.connect()` call, it checks that all three abandoned connections are already closed. Moving the sweep back after allocation makes the assertion fail with `[3] != [0]`. The test also verifies that the previous history and the new message remain readable.

Connections are now removed from the registries after closing, so an interrupted close leaves the connection tracked for subsequent cleanup.

The regression verifies cleanup ordering directly; it does not reproduce an actual allocation failure under descriptor exhaustion.

PR #5095, opened 2026-09-19
fix(core): give Agent.as_tool() nested runs a stable prompt cache key

### LINKED ISSUE #5085: Agent.as_tool() nested subagent runs never hit prompt cache, even with byte-identical instructions/tools/prompt_cache_key
**Setup:** A top-level `Agent` with several subagents attached via `sub_agent.as_tool(tool_name=..., tool_description=..., session=read_only_session_view)`, where `read_only_session_view` is a custom `SessionABC` implementation wrapping the parent's own session (read-only — `add_items`/`pop_item`/`clear_session` are no-ops). Model: `gpt-5.6-luna`. We explicitly set a stable `session_id` on the wrapper so the SDK's own cache-key grouping (`resolve_run_grouping`) resolves to `("session", <stable-hash>)` instead of falling back to a random per-run key — confirmed this works (see below).

**Observed:** Across 42 real subagent calls that clear the 1,024-token minimum, **0 ever returned any `cached_tokens`** — a flat 0% hit rate, including repeat calls to the *same* subagent, seconds apart, with verified byte-identical prefixes.

**Reproduction — two consecutive calls to the same subagent, ~12 seconds apart:**

Captured the literal kwargs passed to `AsyncResponses.create()` (patched at that layer to rule out any SDK-side transformation):

```
Call 1: instructions="Collect \n1. User first name\n2. User last name\n3. User email\n\n...
         ## Known conversation context\n\nToday's date: Fri, 18 Sep 2026 (UTC)."
        prompt_cache_key="agents-sdk:session:<same-hash-both-calls>"
        input: [54 items]
        usage: input_tokens=695, cached_tokens=0

Call 2: instructions=<byte-identical to Call 1 — diff returns nothing>
        prompt_cache_key="agents-sdk:session:<same-hash-both-calls>"
        input: [58 items — first 53 byte-equal to Call 1's, remainder is
                the new turn appended, i.e. a normal growing conversation]
        usage: input_tokens=838, cached_tokens=0
```

Also confirmed identical between the two calls: `tools` (empty in this pair), `model`, and every other kwarg we captured (`store`, `previous_response_id`, `conversation`, `truncation`, `parallel_tool_calls`, `text`, `reasoning`, `extra_body`, `extra_headers` — all consistently omitted/default on both).

**What we ruled out before filing:**
- Unstable `prompt_cache_key` (confirmed identical both calls, via a raw capture at the `responses.create()` call site)
- Tool definition drift
- Instructions drift (confirmed byte-identical via diff, not just visual inspection)
- Input prefix drift (53/54 leading items are dict-equal)
- Cache TTL expiry (still 0% at gaps as short as 8.5s)
- Being under the 1,024-token minimum (42/42 tested calls are above it)

**For comparison:** the *parent* agent in the same run, calling the same model via a normal (non-nested) `Runner.run()` with its own stable `session`, gets 78-90% cache hit rates over the same kind of growing conversation. The only structural difference we can identify is that these are nested calls originating from `Agent.as_tool()`, using a custom read-only `Session` implementation rather than the SDK's own session type.

**Question:** is there a known limitation where nested/`as_tool()`-originated `Runner.run()` calls are excluded from prompt caching regardless of `session`/`prompt_cache_key`, or is there something about a custom `SessionABC` subclass that would prevent this (e.g. is `session_id` alone insufficient, and something else internal is expected)?

Related but distinct: #2784 (base64 file inputs breaking cache) — different root cause, but confirms `prompt_cache_key` having "zero effect" on unexpected 0% hit rates is a recurring pattern in this SDK.

mrchatam:
## Summary

- Fixes #5085: nested `Agent.as_tool()` runs without a session/conversation previously got a random per-run `prompt_cache_key`
- Derive a stable `group_id` (`agent-as-tool:{agent}:{tool}`), namespaced under a parent `group_id` when inherited, so consecutive identical `as_tool()` calls share the same cache boundary as a normal `Runner.run`
- Session / conversation paths are unchanged

## Test plan

- [x] `pytest tests/test_prompt_cache_key.py` (17 passed), including:
  - consecutive `as_tool()` calls without session share a stable group cache key
  - nested run does not share the parent's raw group_id partition
  - with session, `as_tool()` matches `Runner.run` session cache key


seratch:
Thanks for investigating this. This patch does not address the scenario in #5085: the reporter already uses a session with a stable ID and confirmed identical outgoing cache keys, while the new helper leaves that path unchanged.

For runs without a session, callers can already select a stable key through `ModelSettings(extra_args={"prompt_cache_key": "..."})`. The added tests establish key equality but do not demonstrate that the reported cache misses are resolved. Generating or rewriting tracing `group_id` values therefore adds behavior without an established need.

I'm going to close this PR and keep #5085 focused on explaining the zero-hit behavior with its existing stable key.

PR #5098, opened 2026-09-20
fix(sandbox): preserve multiline skill frontmatter descriptions

### LINKED ISSUE #5026: Skills frontmatter parser mangles multi-line descriptions (folded >, literal |, wrapped lines) in the skill index

seratch:
This pull request supersedes #5028 w/ the original author's credit in co-authored-by section in the commit and fixes truncated multiline SKILL.md descriptions in both lazy local discovery and sandbox discovery without adding a YAML dependency.

Folded and literal blocks now handle header comments, indentation indicators, paragraph breaks, more-indented content, and indented document markers. Wrapped plain values retain their continuation text, while nested metadata and comment-only lines cannot overwrite the skill name.

The parser preserves existing single-line behavior and documents its supported string-metadata subset. Block values retain internal formatting while outer whitespace is trimmed for the skill index.

Validation includes 172 focused tests, independent implementation review, and passing formatting, lint, type checking, and full test suites.

Closes #5026


markstuart-oai (approved):
Reviewed `f3a947b7ad005bc21996c94bb18877ebea6e839c`. The shared parser now carries folded/literal blocks and wrapped plain descriptions through both local lazy discovery and sandbox discovery. The regressions cover field boundaries, comments, indentation, paragraphs, and preservation of the existing single-line string behavior. The earlier dedented-comment issue is fixed; I found no actionable correctness or structural issue within the documented metadata subset.

Validation: reviewed both changed files and the parser in the latest release (`v0.22.3`), with an independent source review. All 22 reported checks on this head are successful, including the Python 3.10–3.14 test matrix, Windows tests, type checking, lint, and packaged compatibility checks. I did not rerun tests locally.

PR #5099, opened 2026-09-20
fix(sandbox): validate host grants against normalized workspace roots

seratch:
This pull request fixes Docker host-path grant validation for workspace roots containing parent-directory segments. Validation previously compared mount targets against the unnormalized root, allowing host bind mounts to overlap the effective workspace and be included in snapshot persistence or restore cleanup.

Use the existing workspace path policy to normalize the root before checking overlap. Reject overlapping mounts before container creation or resume, while preserving disjoint mounts and read-only settings. Add regression coverage for creation, serialized-state resume, and valid disjoint grants.


markstuart-oai (approved):
Reviewed the complete diff at 928ce9972e308c703c7d4b59ab49d161fb826824. No actionable findings.

Workspace roots, grant targets and volume destinations now use one lexical Linux path identity, including leading-slash aliases. The existing overlap and collision checks therefore compare the same normalized representation without probing the container root on the host. Creation and resumed trusted-grant rebinding reject invalid overlaps before provider access or mutation of saved session state. The earlier review concerns are addressed.

Validation: independent source review and all 22 exact-head CI checks passed, with actual test, type-check and container steps verified. No local test or Docker execution was performed.

PR #5100, opened 2026-09-20
fix(sandbox): reject host sources in dictionary sandbox manifests

seratch:
This pull request fixes host file access through dictionary sandbox manifests by rejecting `LocalFile` and `LocalDir` entries with a source, including nested directory entries, during configuration construction. Previously, these entries could copy files from the SDK process working directory into the sandbox.

Applications that need to copy local files must validate their paths and provide a trusted `Manifest` instance. Inline files and source-free directories remain supported.


markstuart-oai (approved):
Reviewed all four changed files at `28c22db4d9cc661d80bf89b3aaebd21df17d91f8`, including an independent review. No actionable findings.

The shared coercion check covers all three public configuration paths and rejects nested or typed host-source entries before session creation. Trusted `Manifest` instances retain their identity and host-source configuration, while inline entries and source-free directories remain supported. The regression tests exercise those public boundaries directly, and the API docstrings explain the supported alternative.

Validation: all 22 reported CI checks passed. Verified actual execution of formatting/lint, type checking, Linux tests, Windows tests, and native macOS sandbox tests on this head. Source and whitespace checks passed; no local runtime tests were run.

dpiet-oai (approved):

PR #5102, opened 2026-09-20
fix(realtime): bound incoming WebSocket message size by default

seratch:
This pull request fixes unbounded incoming WebSocket message buffering by setting an 8 MiB default receive limit for Realtime WebSocket and SIP connections. Oversized messages close the connection through the existing transport error path before SDK JSON parsing.

Applications that need larger messages can set `transport_config.max_size` to a higher byte limit or explicitly use `None` to retain unlimited reception. Explicit limits remain unchanged.


src/agents/realtime/openai_realtime.py
```diff
@@ -680,7 +680,7 @@ async def _create_websocket_connection(
         connect_kwargs: dict[str, Any] = {
             "user_agent_header": _USER_AGENT,
             "additional_headers": headers,
-            "max_size": None,  # Allow any size of message
+            "max_size": 8 * 1024 * 1024,
```
dpiet-oai, line 683: **[Medium] Keep the public transport guide aligned with the new default**

`src/agents/realtime/openai_realtime.py:683`

This changes the runtime default to 8 MiB, but `docs/realtime/transport.md` still says the SDK default is `None` and incoming messages are unlimited. A user following that guide can leave `max_size` unset expecting no limit, then have a response larger than 8 MiB close the connection with code 1009.

**Suggested fix:** Update the transport guide to state the 8 MiB default and that `max_size=None` explicitly restores unlimited reception.

dpiet-oai (approved):

markstuart-oai (approved):
Reviewed the complete diff at 23137d5efa8239711f201f74c20be94ba05a29c3. No actionable findings.

The shared connection helper applies the defensive 8 MiB default to WebSocket and SIP connections while preserving explicit byte limits and explicit None. Oversized messages reach the existing transport-error and cleanup path before JSON processing. The real WebSocket tests cover the assembled/decompressed message boundary, overrides and peer close behavior. The transport-guide update can follow the repository's separate release-timed documentation policy.

Validation: independent source review and all 22 exact-head CI checks passed, with actual test, teardown, type-check and container steps verified. No local tests or live Realtime calls were run.

src/agents/realtime/openai_realtime.py
seratch: yeah, the docs will be updated once the change is landed in the next release.

PR #5103, opened 2026-09-20
fix(extensions): preserve filtered history during automatic compaction

seratch:
This pull request fixes automatic compaction uploading or replacing session history that was excluded from the latest model exchange. Automatic compaction now checks the complete stored history against the actual model input and response. If any history falls outside that exchange, compaction is skipped and the history is retained.

The check covers filtered inputs, handoffs, resumed runs, encrypted sessions, and backend read limits. Explicit manual compaction retains its existing behavior.


src/agents/memory/openai_responses_compaction_session.py
```diff
@@ -270,6 +289,45 @@ async def _run_compaction_locked(
...
+            model_items: tuple[str, ...] = getattr(wrapper, "_session_compaction_model_items", ())
+            # Read one occurrence beyond what the model saw. If more history exists,
+            # coverage fails without loading the entire backend. Do this only after
+            # the decision hook approves, and bypass any cached partial snapshot.
+            approved_session_items = session_items
+            (
+                compaction_candidate_items,
+                session_items,
+                complete,
+            ) = await self._ensure_compaction_candidates(read_items, limit=len(model_items) + 1)
```
dpiet-oai: **[Medium] Preserve model-input ordering during visibility checks**

This `Counter` check proves multiplicity but not order. A supported `call_model_input_filter` or `session_input_callback` can reorder stored history, so stored `[A, B]` and model input `[B, A]` both pass and automatic compaction can summarize a sequence the model never saw (or replace stored history with a summary from a differently ordered `previous_response_id` chain).

**Suggested fix:** Require the stored digests to form an ordered subsequence of the recorded model digests, and add a reorder-only filter regression test.

dpiet-oai (commented):
One actionable finding is attached inline.

markstuart-oai (dismissed):
Reviewed 55000b2. The automatic compaction gate accounts for each retained occurrence using the actual filtered model input and successful response, preserves backing-store ID normalization, and skips mutation when the bounded snapshot cannot prove complete coverage. The reloaded snapshot is checked by the decision hook, while explicit manual compaction remains available. I also checked the streaming, deferred-tool, encrypted-session and resumed-run paths and the four earlier bot findings; no actionable issues remain.

Validation: inspected the complete patch and supporting persistence/lifecycle code, with an independent source review. Exact-head hosted CI passed all 22 checks, including Python 3.10–3.14, Windows, type checks and lint. I did not execute tests locally or exercise a live compaction API.

src/agents/memory/openai_responses_compaction_session.py
dpiet-oai: Fixed on the current head. The coverage check now consumes `model_items` through a single iterator and matches each stored digest (or its ID-policy-normalized digest) in order, so reordered stored history cannot pass as the same model exchange. `test_automatic_compaction_preserves_reordered_history` covers the reorder-only filter for both streaming modes: https://github.com/openai/openai-agents-python/blob/ccf18694e897a8f2918c60cd2fc2371f376025e4/tests/memory/test_compaction_model_visibility.py#L499.

dpiet-oai (dismissed):

dpiet-oai (dismissed):

dpiet-oai (approved):

PR #5104, opened 2026-09-20
fix(extensions): validate MongoDB session identifiers

seratch:
This pull request fixes MongoDB session filters accepting non-string identifiers that can be interpreted as query operators, allowing operations to target other sessions. Both `MongoDBSession(...)` and `MongoDBSession.from_uri(...)` now raise `TypeError` for non-string IDs before accessing or creating a client. String identifiers retain their exact values and existing behavior.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed the complete diff at 966282980dff43c2316b6048bd20af039f0ae0bf. No actionable findings.

Both construction paths reject non-string session IDs before touching a supplied client or allocating one. Supported strings retain their exact identity and CRUD isolation. The two small guards protect different side-effect boundaries without adding a separate validation framework; the tests check failure timing, unchanged client arguments and another session's preserved history.

Validation: independent source review and all 22 exact-head CI checks passed, including verified test, type-check and container steps. No local tests or live MongoDB checks were run.

PR #5106, opened 2026-09-20
fix(core): bound agent tool streaming callback backlogs

seratch:
This pull request fixes unbounded event accumulation when an `Agent.as_tool(on_stream=...)` callback cannot keep up with a nested run.

`on_stream_max_pending_events` defaults to 1024, accepts a positive custom limit, and supports `None` for unlimited buffering. Overflow cancels nested generation and callback dispatch before awaiting cleanup, then applies the existing `failure_error_function` policy. Successful calls preserve ordered event delivery, and ordinary stream errors still drain pending callbacks.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed the complete diff at 79a23528491f12eb46a90eba34cb43fc2acffa19. No actionable findings. The bounded callback queue preserves ordered delivery, reserves completion capacity, and stops nested generation before awaiting callback cancellation cleanup. Ordinary stream failures still drain pending callbacks, while iterator-close failures preserve the primary overflow or cancellation.

The default limit is 1024 pending events; callers needing unlimited buffering can explicitly pass on_stream_max_pending_events=None. Reviewed the default/custom/unlimited cases, error policies, full-queue completion and cancellation regressions.

Validation: all 22 checks passed at this head, including the actual Linux/Windows test, type-check and container-test steps. Source review and hosted CI verification only; no local or live API tests were run.

PR #5107, opened 2026-09-20
fix(examples): require interactive approval for host shell commands

seratch:
This pull request fixes unattended host command execution in the shell and local skill examples by requiring explicit interactive approval for every command batch, even when automatic approval environment variables are set. Child processes receive only `PATH` and `SystemRoot`, with closed standard input, to avoid inheriting API credentials, shell startup hooks, or operator input.

The example runner skips these demos in automatic mode and streams approval prompts during manual runs. The examples clarify that approved commands retain host filesystem and network access, and point to existing isolated execution examples.


examples/run_examples.py
```diff
@@ -93,6 +93,9 @@
     "examples/sandbox/tutorials/repo_code_review/main.py",
     "examples/sandbox/tutorials/vision_website_clone/main.py",
     "examples/tools/codex_same_thread.py",
+    # Host shell execution requires a person to review every command batch.
+    "examples/tools/local_shell_skill.py",
+    "examples/tools/shell.py",
```
dpiet-oai, line 98: **[High] Auto mode still approves host commands in the HITL example**

`examples/run_examples.py:98`

Adding only these two paths leaves `examples/tools/shell_human_in_the_loop.py` eligible in auto mode. The runner automatically allows the `interactive` tag in auto mode, and that example calls `confirm_with_fallback(..., default=is_auto_mode())`, so non-interactive runs approve both the current and all future model-generated shell commands. Its executor also passes `os.environ.copy()`, so `.github/scripts/run_examples.sh start` can still execute unattended model output on the host with repository credentials.

**Suggested fix:** Also auto-skip that example, and preferably make its approval path reject in auto/non-TTY mode; add a regression covering it alongside these two paths.

dpiet-oai (changes requested):
One blocking security issue is noted inline.

markstuart-oai (dismissed):
Reviewed the complete diff at 657f7fe6ff41744c6c8f3481e1bc9cad79099ab8. No actionable findings.

Both host-shell examples require a per-call interactive decision, with non-terminal input and EOF denying execution. The shared callback removes the environment-variable bypass; command rendering escapes terminal controls, while child processes receive only PATH/SystemRoot and closed stdin. The existing runner handles visible manual prompts and skips these demos in automatic mode. The examples correctly distinguish approval from host isolation.

Validation: independent source review and all 22 exact-head CI checks passed, with actual test, type-check and container steps verified. No local tests, host shell commands or live API calls were run.

examples/run_examples.py
seratch: @dpiet-oai Fixed. The HITL example is now excluded from auto runs and shares the approval prompt and restricted executor with the other shell examples. Auto/non-TTY runs are rejected, and each command batch requires a fresh approval. Added regression coverage for all three examples, including a check that approving one HITL call does not approve subsequent calls.

Can you take a look again?

dpiet-oai (approved):

PR #5109, opened 2026-09-20
docs: demonstrate fail-closed trace redaction

seratch:
This pull request adds a tracing exporter example that redacts copied payloads before delivery and drops the entire batch if redaction fails, without logging sensitive exception details. It explains why separately registered processors cannot enforce redaction before export and shows how to replace the default processor with a single redacting exporter. Existing SDK behavior is unchanged; the example uses a local console destination and documents its policy and ID-handling limits.


markstuart-oai (approved):
Reviewed the complete diff at 7740c5720f438517f9c7a62756b4edf5f2f9abf9. No actionable findings.

The exporter copies payloads, completes redaction for the entire batch before calling its destination, and drops the batch on serialization, copying or redaction failure. The fixed warning is emitted after leaving the caught redaction exception. Processor replacement, trusted callback assumptions, caller-controlled IDs and previously buffered data are described accurately. The small exporter reuses the existing batch lifecycle without a parallel tracing pipeline.

Validation: independent source review and all 22 exact-head CI checks passed, with actual test, type-check and container steps verified. No local tests or live tracing-backend calls were run.

PR #5111, opened 2026-09-20
fix(examples): restrict the Realtime demo to local bounded sessions

seratch:
This pull request fixes unrestricted access and resource consumption in the Realtime browser demo. The server now binds to loopback, validates the browser Origin against the local Host before opening an upstream session, and connects the UI to the page's host.

Bound concurrent sessions, client messages, audio batches, and image uploads. Assign each connection a server-owned identity so duplicate URL labels cannot overwrite another session. Invalid input closes the connection through the existing cleanup path. Document the local-only boundary and the authentication and usage controls required for public deployments.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed the complete diff at b14088e58857332dd47ced8eb92e35a8bcb9d80e. No actionable findings.

Origin and Host checks reject unrelated browsers before opening an upstream session. Server-owned connection IDs prevent client labels from sharing resources, and session capacity is reserved before the first await. Message, audio and image-assembly limits remain bounded through the existing cleanup path. The launch uses loopback and an explicitly bounded WebSocket backend; the README clearly limits this to trusted local use.

Validation: independent source review and all 22 exact-head CI checks passed, with actual test, type-check and container steps verified. No local tests, browser/microphone session or live Realtime API calls were run.

PR #5114, opened 2026-09-20
fix(voice): stop forcing the STT session logging header

seratch:
This pull request fixes streamed speech-to-text connections unconditionally sending `OpenAI-Log-Session: 1` and overwriting application-provided values. It removes the SDK-injected header and preserves explicit client settings through the existing header merge path. SDK tracing defaults and controls remain unchanged.


markstuart-oai (approved):
Reviewed the complete diff at b43234d. Removing the forced session-logging header correctly reuses the shared client-header merger: explicit values and omission survive, while authentication, endpoint selection, session setup, and SDK tracing controls remain unchanged. The real-client handshake regression covers the affected boundary without a new configuration path. No actionable correctness or structural findings.

All 22 hosted checks passed at this exact head, including the Python test matrix, Windows, typecheck and lint. Source review only; I did not run local tests or a live STT connection. Header omission alone is not a claim about server-side retention.

PR #5115, opened 2026-09-20
fix(core): preserve guarded final outputs in agent tools

seratch:
This pull request fixes default `Agent.as_tool()` extraction so empty strings and `None` validated by nested output guardrails remain the exported final output. Previously, the adapter could replace these values with intermediate message or tool output that the guardrails had not checked.

Unguarded runs retain the existing fallback behavior, and custom output extractors retain control over their output. Nonempty and structured outputs and pending approval handling remain unchanged.

Regression tests cover agent-level and run-level guardrails, handoffs, nested streaming, nullable structured output, refusal propagation, and serialized approval resume, checking both parent tool results and subsequent model input.


markstuart-oai (approved):
Reviewed 2f4399eeb1389dc596135c7bb7fabc1d2188ff74 against the actual merge base. No actionable correctness or structural findings.

The default extractor now preserves the final output that actually completed output guardrails, including empty strings and None, instead of substituting an unchecked intermediate message or tool result. The existing result field covers terminal-agent handoffs and run-level guardrails without duplicating policy resolution. Custom extraction, unguarded fallback, nonempty/structured results, and interrupted approval handling retain their existing behavior. The regressions check both the exported tool output and the next parent-model input, including streamed execution and serialized approval resume.

Validation: independent source review and all 22 hosted checks passed on this head, including Python 3.10–3.14, Windows tests, type checks, lint, packaged contracts, and CodeQL. No local tests or live-provider calls were run.

PR #5116, opened 2026-09-20
fix(sandbox): preserve read-only grants during recursive sandbox removal

seratch:
This pull request fixes recursive sandbox removal bypassing nested read-only path grants. UnixLocal and shared remote sessions reject affected directory removals before deleting any contents, preserving protected files and writable siblings. Remote checks resolve aliases while preserving safe leaf-symlink removal and existing grant precedence.


dpiet-oai (approved):

tests/sandbox/test_docker_removal.py
```diff
@@ -0,0 +1,1835 @@
...
+    DockerSandboxClient,
+    DockerSandboxSession,
+    DockerSandboxSessionState,
+)
+from agents.sandbox.sandboxes.docker_removal import _Worker
+from agents.sandbox.snapshot import NoopSnapshot
+
+worker_code = pytest.importorskip(
+    "agents.sandbox.sandboxes._docker_removal_worker", exc_type=ImportError
+)
```
markstuart-oai: [P2] Split removal tests by service and worker ownership

This new 1,835-line module combines Docker service/container lifecycle tests with the independent trusted worker traversal, account parsing, namespace, descriptor and process-protocol tests. These tests use different authorities and doubles; most worker cases do not need RecordingContainer, RecordingWorker or the service fixture, while service/client cases should not depend on module-wide worker import state. Split worker tests into a dedicated test_docker_removal_worker.py and keep service/session integration tests together, sharing only the small recording fixtures genuinely needed across files. This preserves coverage while making each safety boundary independently reviewable and keeps a brand-new test module from starting above the 1,000-line structural threshold.

src/agents/sandbox/session/base_sandbox_session.py
```diff
@@ -1134,7 +1134,18 @@ async def rm(
...
+        :raises WorkspaceArchiveWriteError: If recursive removal is requested while
+            the session has any read-only extra path grant.
         """
+        if recursive:
+            policy = self._workspace_path_policy()
+            if any(read_only for _, read_only in policy.extra_path_grant_rules()):
+                workspace_path = policy.normalize_sandbox_path(path, for_write=True)
+                raise WorkspaceArchiveWriteError(
+                    path=posix_path_for_error(workspace_path),
+                    context={"reason": "recursive_remove_with_read_only_grants"},
```
markstuart-oai, line 1151: [P2] Preserve snapshot cleanup with unrelated read-only grants

A released supported manifest can grant /opt/toolchain read-only while retaining a writable workspace. This unconditional check now prevents every recursive removal in that workspace, not just protected subtrees; the matching UnixLocal guard does the same. Snapshot restore calls clear_workspace_dir_on_resume_pruned(), which invokes rm(child, recursive=True) for each workspace child before restoring the archive, so restoring an existing nonempty workspace now fails even when the protected grant is unrelated. The new opt-in Docker service only provides an alternative on a restricted rootful Linux daemon-host setup and does not repair UnixLocal or other remote sessions. Preserve a verified safe cleanup path for the supported backend/grant combination, and keep unsupported-operation rejection at a capability boundary with an explicit migration, instead of applying a global read-only-grant ban inside the shared rm path. Do not restore the old racy check-then-delete behavior. This also corroborates the remaining compatibility concern in the earlier public review comment; the current Docker service does not cover these other backends.

markstuart-oai (commented):
Reviewed 3d75e8eac820a7c9bce44aa921356cc07e1a5c5a. Two issues remain: the new 1,835-line test module combines independent service and worker boundaries, and the shared/UnixLocal fallback blocks snapshot cleanup for supported configurations with unrelated read-only grants. The inline comments describe a concrete decomposition and the affected released contract.

All 22 hosted checks passed at this head. Source review included an independent pass and comparison with v0.22.3. I did not run local tests or Docker operations; the new service tests use recording workers and syscall doubles, so passing CI does not establish live execution of that privileged service.

dpiet-oai (changes requested):
**[High] Preserve supported cleanup with unrelated read-only grants**

The existing current-head cleanup finding is verified. `BaseSandboxSession.rm()` now rejects every recursive removal whenever any read-only extra grant exists, even when the requested workspace subtree cannot contain that grant. A supported manifest with an unrelated read-only toolchain mount therefore makes `clear_workspace_dir_on_resume_pruned()` fail on an ordinary nonempty workspace, and the opt-in Docker service does not cover UnixLocal or the other fallback paths.

**Suggested fix:** retain an atomic, backend-appropriate removal path that proves the selected subtree cannot cross a protected grant, or reject the unsupported backend/grant combination at its capability boundary with an explicit migration. I have not duplicated the existing inline thread.

dpiet-oai (changes requested):
**[High] Preserve snapshot cleanup with unrelated read-only grants**

`src/agents/sandbox/session/base_sandbox_session.py:1146`

The existing current-head finding remains independently verified. This unconditional guard rejects every recursive removal whenever any read-only grant exists, even when the target is an unrelated writable workspace child. Snapshot resume clears each existing child through `rm(..., recursive=True)`, so a supported manifest such as a writable workspace plus read-only `/opt/toolchain` can no longer restore a nonempty snapshot on UnixLocal or ordinary remote backends; the opt-in Docker removal service does not cover those paths.

**Suggested fix:** Preserve an atomic, verified cleanup path for supported backends when the protected grant is unrelated, while keeping racy check-then-delete behavior disabled.

src/agents/sandbox/workspace_paths.py
```diff
@@ -285,6 +290,24 @@ def _validate_split_path_grant(self) -> SandboxPathGrant:
...
+def _validate_read_only_grant_capability(
+    grants: tuple[SandboxPathGrant, ...], *, atomic_recursive_remove: bool
+) -> None:
+    """Reject grants a backend cannot preserve through cleanup and snapshot restore.
+
+    Capability comes from trusted backend code, never from the manifest. A backend
+    advertising support must implement recursive removal with authorization that
+    remains valid through every destructive mutation.
+    """
+    if not atomic_recursive_remove and any(grant.read_only for grant in grants):
```
dpiet-oai: **[High] Preserve supported read-only Docker grants**

`src/agents/sandbox/workspace_paths.py:302`

This unconditional capability rejection now prevents every default `DockerSandboxClient` from creating or resuming a manifest with a read-only grant, including the existing credential-isolation and compatible host-grant paths. The current-head `containers` job shows both `test_runner_owned_local_sandbox_cannot_inspect_trusted_client_credential` cases aborting here before the runner starts, and the Python matrix fails the existing ordinary read-only bind, path-only grant, external-storage, and privileged-storage cases. Merging this head would therefore break supported secure sandbox configurations rather than only rejecting unsafe recursive removal.

**Suggested fix:** Preserve those supported grant paths by wiring an atomic removal capability into their client construction (and keeping the integration coverage green), or introduce a separately reviewed compatibility/migration boundary that does not make the default secure credential path fail at creation.

dpiet-oai (changes requested):
One blocking current-head regression is included inline.

src/agents/sandbox/workspace_paths.py
dpiet-oai: **[High] Preserve released read-only grant configurations**

`src/agents/sandbox/workspace_paths.py:302`

`SandboxPathGrant(read_only=True)` is part of the v0.22.3 public contract, but this unconditional capability check now runs during create and resume for every built-in backend except Docker with an optional live `DockerRemovalService`. Applications upgrading with existing UnixLocal, remote-provider, or ordinary Docker manifests therefore fail before connecting; even kernel-enforced read-only host mounts are rejected. This is reachable now: the exact-head Tests workflow fails the 3.10-3.13 matrices in `test_docker_create_preserves_compatible_host_grants_and_storage`, and the container security integration fails because its released read-only credential grant can no longer start.

**Suggested fix:** Preserve configurations whose read-only boundary is already enforced safely (including Docker read-only mounts), and provide a race-safe cleanup path for other supported backends. If any released backend must stop supporting these grants, require and document the minor-version compatibility decision and migration at the release gate, with the existing security and integration contracts updated deliberately.

dpiet-oai (changes requested):
One current-head blocking compatibility finding is included inline.

dpiet-oai (changes requested):
**[High] Preserve released read-only grant configurations**

`src/agents/sandbox/workspace_paths.py:302`

The current-head compatibility regression remains independently verified. This unconditional capability rejection makes every built-in backend except Docker with an optional live `DockerRemovalService` reject released `SandboxPathGrant(read_only=True)` manifests during create or resume, even when callers never request recursive removal. The exact-head Tests workflow fails all Linux and Windows Python lanes, including ordinary read-only bind, path-only grant, external-storage, and privileged-storage coverage; the container security integration also fails because its read-only credential grant cannot start.

**Suggested fix:** Preserve supported configurations whose read-only boundary is already enforced safely, and provide a race-safe cleanup path for the remaining backends. If a released backend must drop this capability, require and document the compatibility and migration decision at the release gate rather than making secure existing manifests fail on upgrade.

seratch:
@dpiet-oai Can you review again?

PR #5117, opened 2026-09-20
fix(sandbox): reject privileged storage with read-only host grants

seratch:
This pull request fixes Docker sandbox configurations that combine read-only host bind grants with in-container storage mounts requiring SYS_ADMIN. On Docker runtimes that permit remounting, those privileges can allow sandbox processes to make the host bind writable.

Create and resume now reject this combination before Docker resource access or environment resolution, including when mount credential exposure has been acknowledged. The error directs callers to remove the host grant or use storage that does not require container mount privileges. Ordinary read-only binds, writable grants, path-only grants, and external Docker-volume storage retain their existing behavior.


markstuart-oai (approved):
Reviewed at `86592de7df7dc82d6219d777eb7144fa1c1cc953`. The guard uses the same effective mount-strategy predicates as Docker capability allocation and rejects incompatible create/resume configurations before provider access, environment resolution, or saved-state changes. Credential-exposure acknowledgement cannot waive the host-bind restriction; ordinary read-only binds, writable grants, path-only grants, and external Docker-volume mounts retain their behavior. No actionable findings after independent review.

Validation: inspected both changed files and surrounding lifecycle/mount validation; all 22 reported checks passed, including the exact-head Tests workflow. No local tests or live Docker/remount operations were run.

dpiet-oai (approved):

PR #5118, opened 2026-09-20
feat(sessions): add an opt-in encrypted history scan budget

seratch:
This pull request adds an optional `max_scan_items` budget to `EncryptedSession` so applications can bound history retrieval and decryption attempts when expired or invalid records remain in storage. The budget counts all items returned by the underlying store during one `get_items` call, including overlapping backfill windows.

Budgeted reads preserve latest-valid-item ordering and raise `RuntimeError` when a complete result cannot be established within the budget. A full final window may raise even when it contains the entire store. The default remains unlimited and backward compatible; encryption, TTL checks, and stored envelopes are unchanged. This item-count budget requires the underlying store to honor requested limits and does not bound item size, backend-internal work, elapsed time, or `pop_item`.


dpiet-oai (commented):
One actionable compatibility issue is noted inline.

src/agents/extensions/memory/encrypt_session.py
```diff
@@ -233,8 +251,13 @@ async def get_items(
     ) -> list[TResponseInputItem]:
         wrapper = _get_session_wrapper(self.underlying_session, wrapper)
         effective_limit = resolve_session_limit(limit, self.session_settings)
-        if effective_limit is not None and effective_limit > 0:
-            window = effective_limit
+        remaining = self.max_scan_items
+        positive_limit = effective_limit is not None and effective_limit > 0
```
dpiet-oai, line 259: **[Medium] Preserve non-positive limit behavior**

With `max_scan_items` enabled, a negative effective limit enters this branch and is replaced by the positive scan budget. That changes reachable existing behavior: `RedisSession` and `MongoDBSession` deliberately return no items for non-positive limits, while SQLite preserves its historical `-1` unlimited behavior; the wrapper now instead returns decrypted history or raises `RuntimeError`. For example, an existing `SessionSettings(limit=-1)` begins feeding history to a runner after the application opts into the scan budget.

**Suggested fix:** Enter the budgeted unlimited scan only when `effective_limit is None`; delegate all `<= 0` limits unchanged (or make a separate, explicit global validation change), and add a negative-limit regression.

markstuart-oai (commented):
Reviewed at `37084a56d3936ca4a839e5645dd1350eb6f40930`. One P2 compatibility issue remains, already captured in [the negative-limit thread](https://github.com/openai/openai-agents-python/pull/5118#discussion_r4058419370): RedisSession and MongoDBSession return no items for a negative effective limit, including in v0.22.3, but the new budgeted branch replaces that limit with a positive scan window. For example, a session with two valid records, `SessionSettings(limit=-1)`, and `max_scan_items=10` now supplies those records instead of an empty history. Please make bounded mode handle negative limits explicitly rather than silently changing the model input; rejecting unsupported negative limits before retrieval is one bounded option. Passing SQLite -1 through unchanged would also need care because it means an unlimited read there.

The positive/None-limit scan accounting, repeated-window charging, chronological ordering, TTL filtering, and documented full-window exhaustion behavior otherwise look consistent. Independent review completed; all 22 reported checks passed, including the exact-head Tests workflow. Source review only; no local tests were run. Referencing the existing inline finding to avoid a duplicate.

dpiet-oai (approved):

PR #5119, opened 2026-09-20
fix(mcp): bind resumed MCP calls to their original recipients

seratch:
This pull request fixes pending MCP approvals being redirected when changed tool listings assign an existing public tool name to a different server or raw tool. It persists the original recipient in RunState schema 1.18 and checks that binding during restore and final resume reconciliation before invoking an MCP tool.

Unchanged MCP calls, completed sibling calls, and explicit application-owned local tool overrides retain their existing behavior. Older snapshots without recipient bindings cannot resume pending local MCP calls and require a new run. The application's MCP server configuration and ordering must remain unchanged across resume.


markstuart-oai (approved):
Reviewed `c261f1d01015c1a861a66bbd42580c31b6227ca2`. No blocking correctness or maintainability findings.

The per-call recipient binding remains independent of restore-time tool discovery, and final reconciliation prevents MCP approvals from authorizing changed recipients. I checked missing/reappearing tools, local and handoff replacements, rejection provenance, completed siblings, and the schema 1.18/legacy boundary across streamed and nonstreamed resume. The documented requirement to keep application MCP configuration and ordering unchanged is clear.

Validation: independent runtime and RunState review passes; all 22 checks pass on this head. The Python 3.11 CI log reports 10,426 passed plus 77 serial tests passed. Review used pinned source and hosted CI; no local tests were run.

dpiet-oai (approved):

PR #5120, opened 2026-09-20
fix(voice): honor pipeline tracing opt-out inside caller traces

seratch:
This pull request fixes `VoicePipelineConfig(tracing_disabled=True)` being ignored when a voice pipeline runs inside an existing caller trace. Both static and streamed input now suppress implicit pipeline spans while preserving the caller's trace and sibling spans, including on failure or cancellation.

Default tracing and independent text/audio capture settings remain unchanged. Speech service requests and application-visible text and audio are preserved. Regression tests cover the public pipeline and outgoing trace payloads using synthetic data and fake transports.


markstuart-oai (approved):
Reviewed 2070a61ffa23e942a38840bdcb16bddbd4cfa55f. No actionable findings. The disabled producer context now obtains both trace and span from the configured provider, addressing the inherited-span and provider-bypass feedback. Static and streamed input keep caller/sibling tracing intact, including error and cancellation cleanup, while the normal tracing path and independent text/audio flags retain their existing behavior.

Validation: independent source review and hosted CI. All 22 checks are successful; verified the test steps ran on Python 3.10–3.14 and Windows 3.10/3.13, along with type checks, lint and packaged contracts. The new tests exercise the real speech adapters and export boundary with synthetic transports, provider customization, and event-controlled cancellation. No local tests or live audio/API calls were run for this review.

dpiet-oai (approved):

PR #5122, opened 2026-09-21
docs: secure local Dapr database provisioning

seratch:
This pull request fixes the Dapr memory example's opt-in local setup, which published database ports on all interfaces and used a fixed PostgreSQL password. PostgreSQL and Redis now bind to `127.0.0.1`, and setup requires a non-default, single-line `POSTGRES_PASSWORD`. The helper passes the password through Docker's environment, encodes it in the PostgreSQL connection URI, and writes owner-only component files without printing credentials.

Setup checks existing containers and component files before making changes. Existing named containers require manual inspection or migration, while differing component files require explicit `--overwrite`. Help and printed instructions use the same setup helper, and Docker failures stop with sanitized recovery guidance. Normal Dapr session behavior is unchanged.


markstuart-oai (approved):
Reviewed `0c9925f624a53fabea26e052108d315cedd92574`, including an independent source review. No actionable findings.

The opt-in setup keeps the database ports on loopback, uses an explicitly supplied password consistently, and protects generated credentials at default and custom component paths. The directory lock covers preflight, component replacement, and container creation; existing containers remain untouched, with manual recovery required for partial provisioning. Normal Dapr session behavior is unchanged.

Validation: 21 of 22 checks passed on this head; `tests-windows (3.10)` is still running. Reviewed the new setup/CLI/concurrency coverage and Docker endpoint-selection semantics. Source review and hosted CI only; no local tests or live Docker/Dapr setup were run.

dpiet-oai (approved):

PR #5123, opened 2026-09-21
fix(core): isolate nested agent tool state across fresh runs

seratch:
This pull request fixes independent fresh runs reusing another run's pending `Agent.as_tool()` approval when provider-generated tool call fields match. Each fresh streamed and nonstreamed run now receives a unique private cache scope before nested tool lookup, so its nested arguments and approvals stay attached to that run.


markstuart-oai (approved):
Reviewed aaffb483e320a4b042159a4cbb198af0f401265d against a2fdb94627776efc9b3e893afe22f28e078c647a. The fresh-run scope uses the existing nested-tool cache identity in both entrypoints, while checkpoint copying and restoration preserve independent approval state. The regression exercises matching provider calls, out-of-order approval, and the original context and usage ownership through both streaming modes. No actionable findings from source and independent review.

All 22 checks passed on the reviewed head, including the completed Tests workflow. Validation used pinned source and hosted CI; no local tests or live API execution.

PR #5125, opened 2026-09-21
fix(core): validate agent tool stream callbacks before execution

seratch:
This pull request fixes silent loss of nested run configuration when an older positional `Agent.as_tool()` call binds `RunConfig` to `on_stream`. Such calls could start the nested run without the explicitly intended controls and only log a callback error afterward.

Reject non-callable `on_stream` values during tool construction with an actionable error directing callers to `run_config=...`. Preserve the current parameter order, valid synchronous and asynchronous callbacks, callable instances, keyword configuration, and existing callback-error handling.


markstuart-oai (approved):
Reviewed 1533cb26b893de322eb5fd991e754eab9c946908 against a2fdb94627776efc9b3e893afe22f28e078c647a. The construction-time guard prevents an invalid stream callback from starting a nested run and gives an actionable run_config keyword fix. It preserves the existing parameter order, valid synchronous/asynchronous and callable-instance handlers, explicit configuration precedence, and callback error handling. The new tests check actual guardrail, input-filter and tracing behavior through the generated tool. No actionable findings from source and independent review.

All 22 checks and the Tests workflow passed on the reviewed head. Validation used pinned source, the v0.22.3 signature, and hosted CI; no local tests or live API execution.

PR #5126, opened 2026-09-21
docs: clarify encrypted session key guidance

seratch:
This pull request updates encrypted-session examples and documentation to recommend randomly generated, high-entropy keys. The previous password-like sample values and guidance accepting any string could imply that per-session HKDF derivation makes weak secrets suitable for encryption.

The in-memory example and quick start now generate a fresh Fernet key. Persistent-storage examples load a securely provisioned key, and the guidance explains retaining the same key and session ID across restarts. It also clarifies that HKDF does not harden passwords, the session ID salt adds no secret entropy, and replacing the configured key does not re-encrypt existing records. Runtime key acceptance, derivation, and ciphertext compatibility are unchanged.


dpiet-oai (approved):

markstuart-oai (approved):
Reviewed cf9834f7f2edafc6a21123f7b3f58cd12c25c4e1 against merge base 082daee3d243d7ddebe4564d48e4a410171b818c. The examples now use generated Fernet keys for ephemeral storage and load a retained, application-provisioned key when reopening persistent storage. The guidance accurately distinguishes deterministic session-key derivation from password hardening and explains key retention and rotation without changing runtime behavior.

Independent source review found no actionable correctness or structural issues. All 22 hosted checks passed, including the documentation build, Python 3.10–3.14 and Windows 3.10/3.13 test steps, type/lint and packaged-contract checks. Source review and hosted CI only; no local tests, examples or documentation builds were run.

PR #5129, opened 2026-09-21
fix(core/extensions): respect sensitive trace capture for model metadata

seratch:
This pull request fixes model-request metadata being copied into generation traces when `trace_include_sensitive_data=False`. The shared trace configuration helper now applies the existing capture policy across Chat Completions, LiteLLM, and AnyLLM chat adapters, including streaming calls.

Provider requests continue receiving `ModelSettings.metadata`, and enabled sensitive-data capture retains its current behavior. Operational trace settings, explicit `RunConfig.trace_metadata`, and caller-authored trace metadata remain unchanged.

Regression tests cover all three adapters in streaming and non-streaming runs, checking exported spans and provider requests. Formatting, lint, both type checkers, and the full parallel and serial test suites pass.


markstuart-oai (approved):
Reviewed e5fed5ef863d524f207de9d29bdde2e780eb53f7 against merge base 518b1f2d6393e4aca80b37b1e84a4a3256fc1b9f. No actionable findings in the full diff or independent review.

The existing shared trace-config helper is the right boundary for removing SDK-copied model metadata when sensitive capture is disabled. All six streaming/non-streaming adapter call sites pass the tracing mode; filtering a fresh dictionary preserves provider metadata, original settings, operational trace fields and explicitly caller-owned trace metadata. This keeps the policy in one place without a new abstraction or public API. No file crosses 1,000 lines.

The regression exercises the actual OpenAI Chat Completions, LiteLLM and AnyLLM adapters through Runner, checking provider kwargs and exported spans with capture enabled and disabled. All 22 hosted checks passed. The Python 3.11 job installed AnyLLM and reports 10,379 tests passed plus 77 serial tests passed. Validation here was pinned-source review and hosted CI evidence only; no local tests or live provider calls were run.

dpiet-oai (approved):

PR #5130, opened 2026-09-21
perf(core): avoid list shifts in function-tool scheduling

seratch:
This pull request improves function-tool scheduling by consuming pending calls with a deque. Repeatedly removing the first element of a list shifts every remaining entry, making queue consumption quadratic in the number of calls, including with default unlimited concurrency. Using `popleft()` makes queue consumption linear while preserving FIFO scheduling, concurrency limits, model-order outputs, and failure and cancellation behavior.

A controlled Runner regression verifies FIFO scheduling across multiple queue refills and model-order outputs when calls finish out of order.


markstuart-oai (approved):
Reviewed 4d9f76ffb5ff8f1794bc8f64c3216f659cb5831b against merge base 518b1f2d6393e4aca80b37b1e84a4a3256fc1b9f. No actionable findings in the full diff or independent review.

The deque directly removes repeated list shifts while preserving FIFO consumption, original order IDs, available-slot accounting, and the existing failure/cancellation paths. The Runner regression controls completion with events and checks multiple refills plus model-order outputs. This is a focused change in the existing scheduler; no new abstraction or scheduling mode is needed, and no file newly crosses 1,000 lines.

All 22 hosted checks passed, including Python 3.10–3.14 tests, Windows tests, lint/type checks, and native macOS sandbox coverage. Validation here was pinned-source review and hosted CI evidence only; I did not run local tests or a performance benchmark.

dpiet-oai (approved):

PR #5131, opened 2026-09-21
fix(core): preserve closed parent constraints in singleton allOf

seratch:
This pull request fixes singleton `allOf` normalization that could introduce properties forbidden by an explicit `additionalProperties: false` parent with empty or missing `properties`. Incompatible merges now raise `UserError`; synthesized closure is applied after nested wrappers have been flattened. Existing reference and annotation normalization, non-strict construction, MCP fallback, and caller-owned schemas are preserved.


markstuart-oai (approved):
Reviewed f667618ab0bf208f000d71ffa7ab8e897cd26625 against merge base 518b1f2d6393e4aca80b37b1e84a4a3256fc1b9f. No actionable findings in the full diff or independent review.

The fix preserves author-supplied closure and defers synthesized closure until compatible singleton-allOf wrappers are flattened. Rejecting an empty closed parent that would acquire new properties matches the constraint's subschema-local meaning ([JSON Schema reference](https://json-schema.org/understanding-json-schema/reference/object#extending-closed-schemas)). The FunctionTool and MCP regressions exercise rejection, nested wrappers, annotations, caller-schema preservation and the existing non-strict alternatives. The changed MCP fixture now tests a valid wrapper instead of relying on the broadening bug.

The v0.22.3 source contains the permissive merge being corrected; the patch preserves supported normalization and keeps non-strict construction/fallback available. The 14-line runtime change stays in the existing normalization/merge owners; no new abstraction or file crossing 1,000 lines is introduced.

All 22 hosted checks passed, including Python 3.10–3.14, Windows, lint/type checks, packaged contracts and native macOS sandbox coverage. The Python 3.11 log reports 10,376 tests passed plus 77 serial tests passed. Validation here was pinned-source review and hosted CI evidence only; no local tests or live provider calls were run.

dpiet-oai (approved):

PR #5132, opened 2026-09-21
fix(realtime): redact exception details from Realtime tool error events

seratch:
This pull request fixes SDK-generated Realtime async tool failure events that included raw exception messages and could expose sensitive details when applications forwarded those events to clients. Tool execution failures and cached-output delivery failures now emit fixed messages regardless of diagnostic logging settings.

Original Python exception propagation, cancellation, cached-output retries, and opt-in diagnostic logging are preserved. Provider error events and intentionally returned tool outputs are unchanged. Regression tests cover both failure paths through session event consumption, logging policies, exception propagation, and retry without repeating tool execution.


markstuart-oai (approved):
Reviewed at 41931bd. The two SDK-generated Realtime error events now use fixed messages while retaining the original exception propagation, diagnostic logging policy, cancellation handling and cached-output retry behavior. The regression tests exercise event consumption, exception/cause identity and resend without reexecuting the tool. No correctness or structural findings.

Independent source review is clean; all 22 current checks pass, including the exact-head CI workflow. No local tests were run.

dpiet-oai (approved):

PR #5133, opened 2026-09-21
feat(mcp): add configurable MCP listing page limits

seratch:
This pull request adds an optional `max_list_pages` setting to stdio, SSE, and Streamable HTTP MCP servers. Automatic tool and prompt listing can otherwise keep fetching successful pages indefinitely when a server returns fresh continuation cursors, including during tool discovery before model execution.

A positive limit bounds successful pages per listing call. Completion on the final permitted page succeeds; further continuation raises an actionable `UserError` without returning or caching partial results or exposing cursor and response data. The budget survives failed-page retries.

`None` preserves the existing unlimited default, and the new keyword-only parameter preserves positional compatibility. Constructor docstrings explain the option. Explicit resource pagination and existing retry policies remain unchanged.


markstuart-oai (approved):
Reviewed at 42d7fa4. The keyword-only limit is forwarded consistently by all three transports, counts successful pages without resetting after tool-request retries, permits completion on the last allowed page, and fails before returning or caching partial discovery. Existing unlimited defaults, positional arguments, cancellation and explicit resource pagination remain intact. The cache, redaction and public-runner tests cover the consequential boundaries. No correctness or structural findings.

Independent source review is clean. At review time 21 checks pass and the Windows Python 3.10 job is still running. No local tests were run.

dpiet-oai (approved):

PR #5135, opened 2026-09-21
feat(sandbox): allow configuring memory consolidation turns

seratch:
This pull request adds `MemoryGenerateConfig.phase_two_max_turns` so applications can bound the separate phase-two memory consolidation run. The outer `Runner` turn limit does not apply to consolidation, which previously always used 500 turns.

The new option preserves the default of 500 and existing positional arguments, and forwards directly to the existing Runner turn limit. Exhaustion retains the memory manager's existing failure handling: it logs the failure, does not record a successful consolidation selection, and continues cleanup. Earlier tool writes are not rolled back. The limit counts model turns; it is not a wall-clock, token, or monetary quota.


markstuart-oai (approved):
Reviewed at b1ed2cea. No actionable correctness or structural findings. The appended configuration field preserves the 500-turn default and existing positional arguments, and reuses Runner turn accounting directly. The exhaustion path and regression coverage preserve earlier writes, avoid recording a successful consolidation selection, and complete session cleanup.

Validation: complete diff and independent source review; 21/22 hosted checks pass; still running: tests-windows (3.10). No local tests were run.

dpiet-oai (approved):

PR #5136, opened 2026-09-21
fix(core): preserve tool identity during asynchronous enablement

seratch:
This pull request fixes tool enablement results being applied to different tools when a callback or concurrent application update changes `agent.tools` during asynchronous enablement. An enabled tool's result could previously expose a different, disabled tool after the list was reordered or replaced.

Snapshot the configured tool list before evaluating enablement and filter that same snapshot, preserving tool identity and order. Later resolutions still use the current configuration, and MCP fetching, callback errors, cancellation, and dispatch behavior remain unchanged.

Regression tests cover direct resolution and streaming/non-streaming Runner model exposure, including configuration changes taking effect on later calls.


markstuart-oai (approved):
Reviewed the complete diff at `650811f6`. The shallow snapshot keeps enablement results paired with the evaluated tool objects and preserves their order, while later resolutions observe the current configuration. Existing MCP resolution, cancellation/error propagation, pruning, and dispatch remain on their established paths.

The focused regression tests exercise reordering and replacement through direct resolution and streaming/non-streaming Runner model exposure, with deterministic synchronization and task cleanup. The change stays within the canonical resolver without adding new state or abstractions. Independent review found no actionable issues.

All 22 hosted checks passed on this commit. Validation was source review and exact-commit CI evidence; no local tests were run.

dpiet-oai (approved):

PR #5137, opened 2026-09-21
fix(sessions): add an optional compaction rollback item budget

seratch:
This pull request adds `max_rollback_items` to `OpenAIResponsesCompactionSession`. Manual compaction and legacy whole-history replacement can load the entire stored history for rollback even when the session's default retrieval window is small. A configured budget rejects oversized rollback history with a bounded read before the compaction API call or destructive writes, reuses the complete accepted snapshot, and bounds inspection after a failed clear.

`None` preserves the existing default behavior. The budget counts items, not bytes, and is independent of `SessionSettings.limit`. Complete-history restoration and cancellation handling remain intact; native automatic suffix replacement continues to use its existing bounded snapshots and atomic replacement.


markstuart-oai (approved):
Reviewed the complete diff at `091de916`. The opt-in budget uses the existing full-history reader, rejects overflow before compaction or destructive replacement, and preserves unlimited behavior by default. Native automatic suffix replacement remains on its existing atomic path. Refreshing the rollback snapshot after the API call preserves the existing behavior for expiring session data.

The regression tests cover retrieval-default independence, exact-budget acceptance, failure recovery, and cancellation. The change stays within the established compaction and restore machinery; independent review found no actionable issues.

All 22 hosted checks passed on this commit. Validation was source review and exact-commit CI evidence; no local tests were run.

dpiet-oai (approved):

PR #5138, opened 2026-09-21
fix(extensions): honor run tracing policy in Codex command spans

seratch:
This pull request fixes Codex command spans capturing command text and output when `RunConfig.trace_include_sensitive_data=False`. It applies the existing run policy before span creation and throughout command updates and completion, including failed-command error data, while retaining status and exit codes.

`tracing_disabled=True` also suppresses command spans inside a caller-owned trace. Enabled capture, truncation, direct calls without a run configuration, application stream callbacks, and tool results retain their existing behavior. No public option is added.


markstuart-oai (approved):
Reviewed the complete diff at `54e6c088a27ea3a358d3887d0814ce1d0fa90d71`. Command spans now honor the owning run policy at creation, through updates, and in failure data; disabling tracing also suppresses them under a caller-owned trace. Safe status/exit-code metadata, tool results, stream callbacks, and the existing capture-enabled truncation behavior are preserved.

The change uses the existing RunConfig and span handlers without introducing a second policy layer. The regression coverage checks processor-start data as well as subsequent updates and final exports. No actionable correctness or structural findings after source review and independent review.

Validation: all 22 hosted checks passed on this head, including Tests run 35568311015. No local tests were run.

dpiet-oai (approved):

PR #5139, opened 2026-09-21
fix(sandbox): limit Darwin host PATH read grants to the child environment

seratch:
This pull request fixes Unix-local macOS profiles granting read access to host PATH entries that were removed from the child environment. For example, replacing the child PATH with `/usr/bin:/bin` no longer implicitly grants access to an unrelated host virtual environment.

Only host PATH entries retained in the effective child PATH can expand read access to a Python virtual environment root. Existing interpreter handling, retained host virtual environments, manifest-only PATH restrictions, and explicit path grants are preserved. Regression coverage checks excluded host paths and retained virtual environments with named and absolute Python commands.


markstuart-oai (approved):
Reviewed the full diff at `253b8bd8ad34766485eb65b984cf996a6c5d8c93`. The Darwin grant calculation now follows the effective child PATH and the current client's inheritance policy, including resumed sessions. Retained virtual environments, resolved interpreters and explicit path grants continue through the existing grant helpers; Linux execution is unchanged.

The change stays local and adds no parallel policy state. The regression tests cover removed host entries, named and absolute Python commands, and create/resume with matching manifest PATH under both client policies. No actionable correctness or structural findings after source and independent review.

Validation: all 22 hosted checks passed on this commit. Tests run 35570683251 passed, including the native macOS sandbox job. No local tests were run.

dpiet-oai (approved):

PR #5140, opened 2026-09-21
docs: cap financial research at 15 planned searches

seratch:
This pull request fixes unbounded search scheduling in the financial research example. The planner asks for 5–15 searches, but a longer model-generated plan previously started a search agent for every item.

The manager now schedules only the first 15 searches, sharing the limit with the planner prompt and progress count. Normal plans and the output schema remain unchanged. Regression tests verify search calls, report evidence, and progress totals for normal and oversized plans.


markstuart-oai (approved):
Reviewed the complete diff at `60a1ebb7430c644953e4865707dca6be5b4ea729`. The cap is enforced before task creation, so oversized plans launch only the first 15 searches. Planning and completion counts agree with the actual work, and only successful performed searches flow into report evidence. Normal plans and the output schema retain their existing behavior.

The shared constant and list slice keep the implementation direct. The regression tests exercise the manager workflow and verify launched queries, evidence, progress totals and cleanup for normal, exact-limit and oversized plans. No actionable correctness or structural findings after source and independent review.

Validation: all 22 hosted checks passed on this commit, including Tests run 35569115302. No local tests were run.

dpiet-oai (approved):

